PR #140, opened 2018-06-18
Updates for forceRefresh

shoatman:


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -739,7 +744,7 @@ private AuthenticationConstants() {
          * Signature info for Azure authenticator app that installs authenticator
          * component.
          */
-        public static final String AZURE_AUTHENTICATOR_APP_SIGNATURE = "ho040S3ffZkmxqtQrSwpTVOn9r0=";
+        public static final String AZURE_AUTHENTICATOR_APP_SIGNATURE = "N1jdcbbnKDr0LaFZlqdhXgm2luE=";//
```
heidijinxujia: This is only for testing, right? Would it be better not to merge it?

iambmelt: +1

shoatman: I'll fix this.  We should probably revisit how this value gets set as part of the release process.  We effectively have the production signature checked in.

shoatman:
@iambmelt @heidijinxujia - Thanks guys.  I've updated.

PR #141, opened 2018-06-18
merge the change deprecating unity-fork

piqiums:
merge the change deprecating unity-fork


common/build.gradle
```diff
@@ -11,7 +11,7 @@ apply plugin: 'maven-publish'
 
 android {
     compileSdkVersion 25
-    buildToolsVersion "26.0.2"
+    buildToolsVersion '27.0.3'
```
iambmelt: The `.travis.yml` needs to be updated to support this -- good to see this getting rev'd though 😁

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -148,6 +151,7 @@ private AuthenticationConstants() {
          * Waiting for broker package to be installed, and resume request in broker.
          */
         public static final int BROKER_REQUEST_RESUME = 2006;
+        public static final int BROWSER_CODE_DEVICE_REGISTER = 2006;
```
iambmelt: Is this constant value dictated by a spec? It matches the same code as `BROKER_REQUEST_RESUME`

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -461,13 +478,21 @@ private AuthenticationConstants() {
          * String for request id returned from Evo.
          **/
         public static final String REQUEST_ID_HEADER = "x-ms-request-id";
+
+        /** String for the broker version. */
+        public final static String ADAL_BROKER_VERSION = "x-client-brkrver";
```
iambmelt: Misspelled constant

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -97,6 +115,30 @@ public void setSecretKey(byte[] rawKey) {
         mSecretKeyData.set(rawKey);
     }
 
+    /**
+     * set two raw bytes to derive secretKey to use in encrypt/decrypt. KeySpec
+     * algorithm is AES. used by broker only.
+     *
+     * @param secretKeys App related keys to use in encrypt/decrypt. Should contain two secret keys. 
+     */
+    public void setSecretKeys(final Map<String, byte[]> secretKeys) {
```
iambmelt: This function already has a method called `setSecretKey` -- this is confusing.

This is for the broker only, right? I think this needs to be in another class or named differently. This is too similar to the existing method

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -97,6 +115,30 @@ public void setSecretKey(byte[] rawKey) {
...
+        if (secretKeys == null) {
+            throw new IllegalArgumentException("The passed in secret key map is null.");
+        }
+        
+        if (secretKeys.size() != 2) {
+            throw new IllegalArgumentException("Expect two keys are passed in.");
+        }
+
+        for (Map.Entry<String, byte[]> entry : secretKeys.entrySet()) {
```
iambmelt: nit: could declare `final`

```java
for (final Map.Entry<String, byte[]> entry : secretKeys.entrySet()) {
    // ...
}
```

common/src/main/java/com/microsoft/identity/common/adal/internal/net/WebRequestHandler.java
```diff
@@ -53,6 +54,8 @@
 
     private UUID mRequestCorrelationId = null;
 
+    private String mCurrentAdalClientVersion = "";
```
iambmelt: This doesn't seem to be used anywhere; why is this needed?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
piqiums: this is a constant used in ad-accounts, it also has the value of '2006', to not break the build of ad-accounts, it is added here. Let's check if we could use the same variable name after Unity is deprecated.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
piqiums: This is a merge from the Unity, so I would like to keep the "backward compatibility"

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -760,6 +835,7 @@ private AuthenticationConstants() {
          * The value for pkeyauth redirect.
          */
         public static final String PKEYAUTH_REDIRECT = "urn:http-auth:PKeyAuth";
+        public static final String CLIENT_TLS_REDIRECT = "urn:http-auth:PKeyAuth";
```
iambmelt: This matches the constant declared directly above -- should we just use one of these?

common/src/main/java/com/microsoft/identity/common/adal/internal/net/WebRequestHandler.java
```diff
@@ -94,4 +97,13 @@ public HttpWebResponse sendPost(URL url, Map<String, String> headers, byte[] con
     public void setRequestCorrelationId(final UUID requestCorrelationId) {
         mRequestCorrelationId = requestCorrelationId;
     }
+
+    @Override
+    public void setAdalClientVersion(String adalClientVersion) {
```
iambmelt: Let's rename this to `setClientVersion` -- [MSAL uses this header as well](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/f3861311f0ea1ee32cd30da7b548f8f701515775/msal/src/internal/java/com/microsoft/identity/client/PlatformIdHelper.java#L76) and ideally we will share all HTTP logic in the future.

Please update the parameter of this method to be `final clientVersion` -- don't forget to update the field name as well

common/src/main/java/com/microsoft/identity/common/adal/internal/net/IWebRequestHandler.java
```diff
@@ -60,4 +60,10 @@ HttpWebResponse sendPost(URL url, Map<String, String> headers, byte[] content,
      * @param requestCorrelationId {@link UUID} of the correlation id to set in the web request.
      */
     void setRequestCorrelationId(final UUID requestCorrelationId);
+
+    /**
+     * Set the version name of ADAL client
+     * @param adalClientVersion the version name of ADAL client
+     */
+    void setAdalClientVersion(String adalClientVersion);
```
iambmelt: Rename to `setClientVersion` for eventual ADAL/MSAL sharing

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -57,6 +57,9 @@ private AuthenticationConstants() {
      */
     public static final int DEFAULT_EXPIRATION_TIME_SEC = 3600;
 
+    /** The authentication file directory */
+    public static final String AUTHENTICATION_FILE_DIRECTORY = "com.microsoft.aad.adal.authentication";
```
iambmelt: Unused?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
piqiums: removed the "CLIENT_TLS_REDIRECT "

common/src/main/java/com/microsoft/identity/common/adal/internal/net/IWebRequestHandler.java
piqiums: renamed. good catch

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
piqiums: removed.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/WebRequestHandler.java
piqiums: updated. thanks

PR #145, opened 2018-06-21
Added State validation changes to Authorization Request

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -1,126 +1,128 @@
...
+
+    /**
+     * Perform the authorization request in the embedded web view and return the authorization result.
+     *
+     * @param request authorization request
+     * @return AuthorizationResult
+     */
+    public AuthorizationResult requestAuthorization(@NonNull final AuthorizationRequest request) {
+        loadStartUrl();
+        // TODO : Add state parameter from the AuthorizationRequest to the Intent parameter like below
```
kreedula: Only TODO is added here, Not sure why github diff is showing deletion/addition of all the lines here. Locally it's fine.

PR #153, opened 2018-06-27
Correct the contents of the token cache item for MRRT and FRT token c…

shoatman:
Correct the contents of the token cache item for MRRT and FRT entries.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/ADALTokenCacheItem.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/ADALTokenCacheItem.java
@@ -104,6 +104,23 @@ public class ADALTokenCacheItem {
     }
 
 
+    public static ADALTokenCacheItem getAsMRRTTokenCacheItem(ADALTokenCacheItem originalCacheItem){
+        ADALTokenCacheItem newItem = new ADALTokenCacheItem(originalCacheItem);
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/ADALTokenCacheItem.java
```diff
@@ -104,6 +104,23 @@
...
+        newItem.setResource(null);
+        newItem.setAccessToken(null);
+        return newItem;
+    }
+
+    public static ADALTokenCacheItem getAsFRTTokenCacheItem(ADALTokenCacheItem originalCacheItem){
+        ADALTokenCacheItem newItem = new ADALTokenCacheItem(originalCacheItem);
+        newItem.setResource(null);
+        newItem.setAccessToken(null);
+        newItem.setClientId(null);
```
heidijinxujia: Should the client id be "foci-1"?

shoatman: I think that FOCI-1 is used in the cache key, but not in the tokencacheitem itself.  I couldn't find any code setting the family id or the client id in the token to be FOCI-1.   I'll be writing a new test to dump the cache contents tomorrow so if i'm mistaken i'll find out soon.

PR #158, opened 2018-07-04
Implement the OAuth2WebViewClient related classes

heidijinxujia:
* Implement the OAuth2WebViewClient #155 
* Implement the NtlmChallengeHandler #156 
* Add the ChallengeCompletionCallback interface #157

PR #163, opened 2018-07-06
Implement the AzureActiveDiretoryWebViewClient and related handlers

### LINKED ISSUE #143: [Webview] Implement the AzureActiveDiretoryWebViewClient

heidijinxujia:
Close #143 #160 #161


common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+        final Uri requestUrl = request.getUrl();
+        return handleUrl(view, requestUrl.toString());
+    }
+
+    private boolean handleUrl(final WebView view, final String url) {
+        final String formattedURL = url.toLowerCase(Locale.US);
+
+        if (isPkeyAuthUrl(formattedURL)) {
+            Logger.verbose(TAG, "WebView detected request for pkeyauth challenge.");
+            new PKeyAuthChallengeHandler(url, view, getRequest(), getCompletionCallback()).process();
```
iambmelt: My feedback for this line is quite similar to [this feedback about `NtlmChallengeHandler`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/158#discussion_r200783221) -- mainly, that the construction and processing of the challenge should be the responsibility of separate objects

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+     * @return return true means the host application handles the url, while return false means the current WebView handles the url.
+     */
+    @Override
+    @RequiresApi(Build.VERSION_CODES.N)
+    public boolean shouldOverrideUrlLoading(final WebView view, final WebResourceRequest request) {
+        final Uri requestUrl = request.getUrl();
+        return handleUrl(view, requestUrl.toString());
+    }
+
+    private boolean handleUrl(final WebView view, final String url) {
```
iambmelt: Aside from the critique about the challenge handlers, overall this method is pretty clean and readable 👍

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -69,7 +70,7 @@ public AuthorizationResult requestAuthorization(@NonNull final AuthorizationRequ
      * @throws ClientException
      */
     public EmbeddedWebViewAuthorizationStrategy(@NonNull final Activity activity,
```
iambmelt: We're increasing our coupling here between `EmbeddedWebViewAuthorizationStrategy` and Microsoft's services. I recall giving the feedback that the `EmbeddedWebViewAuthorizationStrategy` really ought to accept the `WebViewClient` as a parameter and this is exactly why.

If we're ever going to add support for other IdP's (and it is our goal to do this over a long enough timespan), this will have to get refactored out.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/OAuth2WebViewClient.java
```diff
@@ -47,7 +48,7 @@
 
     private final IChallengeCompletionCallback mCompletionCallback;
     private final AuthorizationRequest mRequest;
-    private final Context mContext;
+    private final Activity mActivity;
```
iambmelt: This field appears to be used only in the creation of the `NtlmChallengeHandler` -- I suspect you can use [getContext()](https://developer.android.com/reference/android/view/View.html#getContext()) and avoid this entirely

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/OAuth2WebViewClient.java
```diff
@@ -96,7 +97,7 @@ public void onReceivedHttpAuthRequest(WebView view, final HttpAuthHandler handle
 
         //TODO TelemetryEvent.setNTLM(true); after the Telemetry is finished in common.
         final NtlmChallengeHandler ntlmChallengeHandler
-                = new NtlmChallengeHandler(view, handler, host, realm, mContext, mCompletionCallback);
+                = new NtlmChallengeHandler(view, handler, host, realm, mActivity, mCompletionCallback);
```
iambmelt: Consider whether [`getContext()`](https://developer.android.com/reference/android/view/View.html#getContext()) will work for you purposes and consider removing the `Activity` here

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -22,5 +22,79 @@
...
+import java.security.Principal;
+import java.security.PrivateKey;
+import java.security.cert.X509Certificate;
+
+public final class ClientCertAuthChallengeHandler implements IChallengeHandler {
+    private static final String TAG = ClientCertAuthChallengeHandler.class.getSimpleName();
+    private ClientCertRequest mClientCertRequest;
+    private Activity mActivity;
+
+    public ClientCertAuthChallengeHandler(@NonNull final ClientCertRequest request,
```
iambmelt: The `ClientCertRequest` is basically "the challenge", right? Unlike the other `ChallengeHandler` classes, this one appears to accept the Challenge and not construct it (which I like).

Consider the suggestion [from here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/158#discussion_r200779395) --  you could declare this class using `IChallengeHandler<ClientCertRequest, Void>` and users of this class could conceivably swap out its implementation if they wanted to

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+                                    @NonNull final WebView view,
+                                    @NonNull final AuthorizationRequest request,
+                                    @NonNull IChallengeCompletionCallback completionCallback) {
+        mWebView = view;
+        mRedirectUri = redirectUri;
+        mRequest = request;
+        mChallengeCallback = completionCallback;
+    }
+
+    public void process() {
```
iambmelt: Needs `@Override` but also I really don't see the value of this interface without accepting a parameter or return type.... 🤔

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -22,5 +22,79 @@
...
+                                          @NonNull final Activity activity) {
+        mClientCertRequest = request;
+        mActivity = activity;
+    }
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public void process() {
```
iambmelt: Needs `@Override`

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+            deviceCertProxy = (IDeviceCertificate) constructor.newInstance((Object[]) null);
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException
+                | IllegalArgumentException | InvocationTargetException e) {
+            throw new ClientException(ErrorStrings.DEVICE_CERTIFICATE_API_EXCEPTION,
+                    "WPJ Api constructor is not defined", e);
+        }
+        return deviceCertProxy;
+    }
+
+    class PkeyAuthChallengeRequest {
```
iambmelt: This should be a top-level class. Not a child of its handler

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+    /**
+     * @param redirectUri        Location: urn:http-auth:CertAuth?Nonce=<noncevalue>
+     *                           &CertAuthorities=<distinguished names of CAs>&Version=1.0
+     *                           &SubmitUrl=<URL to submit response>&Context=<server state that
+     *                           client must convey back>
+     * @param view
+     * @param completionCallback
+     * @throws ClientException
+     */
+    public PKeyAuthChallengeHandler(@NonNull final String redirectUri,
```
iambmelt: This class really doesn't expose itself to be tested -- I think the `IChallengeHandler` interface needs some rethinking such that it provides some swappability/testability.

The `private` and `static` methods on this class do a TON of complex logic that we have no way to test.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+        }).start();
+    }
+
+    /**
+     * This parses the redirectURI for challenge components and produces
+     * response object.
+     *
+     * @return Return Device challenge response
+     */
+    private PkeyAuthChallengeResponse getChallengeResponseFromUri() throws ClientException {
```
iambmelt: This is basically a factory method but because it's `private` we cannot test it -- how do you know this works??

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+     * This parses the redirectURI for challenge components and produces
+     * response object.
+     *
+     * @param redirectUri Location: urn:http-auth:CertAuth?Nonce=<noncevalue>
+     *                    &CertAuthorities=<distinguished names of CAs>&Version=1.0
+     *                    &SubmitUrl=<URL to submit response>&Context=<server state that
+     *                    client must convey back>
+     * @return Return PKeyAuth challenge response
+     */
+    private PkeyAuthChallengeRequest getChallengeRequestFromUri(
```
iambmelt: This is also a factory method but is not testable

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+            mNonce = nonce;
+            mContext = context;
+            mCertAuthorities = certAuthorities;
+            mThumbprint = thumbprint;
+            mVersion = version;
+            mSubmitUrl = submitUrl;
+        }
+    }
+
+    class PkeyAuthChallengeResponse {
```
iambmelt: Another top-level class

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+        }
+
+        void setAuthorizationHeaderValue(String authorizationHeaderValue) {
+            mAuthorizationHeaderValue = authorizationHeaderValue;
+        }
     }
 
-    void processPkeyAuthChallenge() {
-        //TODO
+    enum RequestField {
```
iambmelt: _Maybe_ a top-level `enum` -- maybe just `Strings` (seems to only be used in this way)

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+ * For web view client, we do not distinguish V1 from V2.
+ * Thus we name V1 and V2 webview client as AADWebViewClient, synced with the naming in the iOS common library.
+ * <p>
+ * The only differences between V1 and V2 is
+ * 1. on the start url construction, which is handled in the Authorization request classes.
+ * 2. the auth result is handled in the Authorization result classes.
+ */
+public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
+    private static final String TAG = AzureActiveDirectoryWebViewClient.class.getSimpleName();
+    private static final String INSTALL_URL_KEY = "app_link";
```
piqiums: this constant is also defined in "BasicWebViewClient", could we move it to AuthenticationConstants to keep only one definition?

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+                .getStringExtra(AuthenticationConstants.Broker.BROKER_REQUEST));
+    }
+
+    private boolean processRedirectUrl(@NonNull final WebView view, @NonNull final String url) {
+        final Map<String, String> parameters = StringExtensions.getUrlParameters(url);
+        if (!StringExtensions.isNullOrBlank(parameters.get("error"))) {
+            Logger.info(TAG, "Sending intent to cancel authentication activity");
+            Intent resultIntent = new Intent();
+            resultIntent.putExtra(AuthenticationConstants.Browser.RESPONSE_ERROR_CODE, parameters.get("error"));
+            resultIntent.putExtra(AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE, parameters.get("error_description"));
```
piqiums: please make these strings constants

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+        getCompletionCallback().sendResponse(AuthenticationConstants.UIResponse.BROKER_REQUEST_RESUME, resultIntent);
+
+        // Having thread sleep for 1 second for calling activity to receive the result from
+        // prepareForBrokerResumeRequest, thus the receiver for listening broker result return
+        // can be registered. openLinkInBrowser will launch activity for going to
+        // playstore and broker app download page which brought the calling activity down
+        // in the activity stack.
+        final int threadSleepForCallingActivity = 1000;
+        try {
+            Thread.sleep(threadSleepForCallingActivity);
```
piqiums: Usually, we should avoid using 'Thread.sleep' since it might cause ANR. Could we try some other methods, like "Handler.postDelayed"?

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -0,0 +1,261 @@
...
+        return false;
+    }
+
+    private void returnError(final String errorCode, final String errorMessage) {
+        Intent resultIntent = new Intent();
+        resultIntent.putExtra(AuthenticationConstants.Browser.RESPONSE_ERROR_CODE, errorCode);
+        resultIntent.putExtra(AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE, errorMessage);
+
+        if (getRequest() != null) {
+            if (getRequest() instanceof AzureActiveDirectoryAuthorizationRequest) {
```
piqiums: Please correct me if I was wrong, since MicrosoftStsAuthorizationRequest extends AzureActiveDirectoryAuthorizationRequest, these 2 case could be merged into one. Otherwise, the case of "if (getRequest() instanceof MicrosoftStsAuthorizationRequest)" won't be reached.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -22,5 +22,79 @@
...
+        final Principal[] acceptableCertIssuers = mClientCertRequest.getPrincipals();
+
+        // When ADFS server sends null or empty issuers, we'll continue with cert prompt.
+        if (acceptableCertIssuers != null) {
+            for (Principal issuer : acceptableCertIssuers) {
+                if (issuer.getName().contains("CN=MS-Organization-Access")) {
```
piqiums: please make this string a constant

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -22,21 +22,259 @@
...
+        mWebView = view;
+        mRedirectUri = redirectUri;
+        mRequest = request;
+        mChallengeCallback = completionCallback;
+    }
+
+    public void process() {
+        mWebView.stopLoading();
+        mChallengeCallback.setPKeyAuthStatus(true);
+        new Thread(new Runnable() {
```
piqiums: since you're posting a Runnable to the MessageQueue in the UI thread, you don't need to start a background thread to post the Runnable.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/OAuth2WebViewClient.java
heidijinxujia: Actually, activity is required to construct the webview. And it is also used in the client cert request and other requests, like open link in browser and install brokers.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallengeHandler.java
heidijinxujia: enum looks much cleaner to me. Not sure the benefits to change them into Strings constants.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
heidijinxujia: This is only used in this class. So I keep it inside the scope of this class.

heidijinxujia:
@iambmelt  @piqiums  Thanks a lot for code reviews. I have addressed the comments. Have a look when you avail, as it still shows "requested changes". Thanks!

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/AzureActiveDirectoryWebViewClient.java
piqiums: I found it defined in 'BasicWebViewClient' on line 50  :  https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/45db7dddf3ea20d5f558aca67203546924a1dd2b/adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java#L50

iambmelt:
Oops. Didn't see that there were conflicts in with the base branch -- let me know once you've resolved and I can reapprove 👍

heidijinxujia:
Due to the change of OAuth2ClientWebView PR. Need to resolve the conflict and keep a clean git commit history by creating a new PR #170

PR #167, opened 2018-07-11
Merging release branch 0.0.1 to master

kreedula:


kreedula:
Thanks @heidijinxujia , Looks like these .idea files are in dev for a while. @iambmelt Can you let us know if the .idea files are needed ? I see a commit from you adding these https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/9ccc0f18f867a51822ba61bee118cca0b712326c

iambmelt:
@kreedula @heidijinxujia -- I used the [GitHub reference `.gitignore`](https://github.com/github/gitignore/blob/master/Android.gitignore) to create this. The reference `.gitignore` exclude _some_ `.idea/` files, but not all of them -- `vcs.xml` is not in their blacklist.

The `vcs.xml` files indicates to Android Studio that this project is under source control from `git` and prevents Android Studio from prompting users to _add_ version control to the project.

I don't have strong feelings one way or another because the reality is this:
- If we ignore this file and users open this project in Android Studio after checking out from git, their environment will generate this file which will subsequently _also_ be ignored
- If we do _not_ ignore this file, users open this project in another IDE will be unaffected by its presence, as it's only used by Android Studio

All this to say, if we remove it users' environments will generate their own identical file anyway -- I can remove this from `dev` and we can bring the change into this branch: that's no problem.

kreedula:
Thanks @iambmelt . Per our discussion and recommendation by IntelliJ on .idea files [here ](https://www.jetbrains.com/help/idea/2017.1/about-projects.html), I think we are doing the right thing here by ignoring all the user specific .idea files and having project specific ones in the repo. @heidijinxujia , please let us know us what you think ?

heidijinxujia:
I think we should remove 
image
 Like you said, they are all user specific. As a library, we should not include these local generated files into our lib.

heidijinxujia:
@iambmelt  Thank you for add the changelog.

iambmelt:
@kreedula Please don't merge this PR until we're signed off and ready for prod release 👍 

/cc @addev-ashish @shoatman @heidijinxujia @piqiums

kreedula:
Closing the PR. Will submit a new PR from release/0.0.2 to master during prod release

PR #171, opened 2018-07-16
Fix the bugs for Embedded WebView Intergration

heidijinxujia:
* Add default constructor for authorization request
* Add serialVersionUID for auth request.
* Fix the encoding for the auth start url
* Fix the finding webview. As the webview is defined in MSAL.AuthenticationActivity. MSAL should be responsible to find the webview by id and passes it to common's webviewclient.
* Fix the bug where auth code is case-sensitive
* Add null check to PKCE challenge
* Update the constants for auth request.
* Fix the unit tests.
* Fix PMD issue
* Resolve the lint warning
* Rename the createWebView to setUpWebView
* Add javadoc to serialVersionUID

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
@@ -62,6 +62,7 @@ public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest
     public static final String LIB_ID_CPU = "x-client-CPU";
     public static final String LIB_ID_OS_VER = "x-client-OS";
     public static final String LIB_ID_DM = "x-client-DM";
+    private static final long serialVersionUID = 6873634931996113294L;
 
     /**
      * Required.
@@ -127,6 +128,13 @@ public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest
         mPkceChallenge = pkceChallenge;
         mExtraQueryParam = extraQueryParam;
         mLibraryVersion = libraryVersion;
+    }
+
+    /**
+     * Default constructor of {@link MicrosoftAuthorizationRequest}
+     */
+    public MicrosoftAuthorizationRequest() {
+        super();
     }
 
     public URL getAuthority() {
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
@@ -37,6 +37,7 @@ import com.microsoft.identity.common.internal.util.StringUtil;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
+import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -111,7 +112,6 @@ public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest
                                          final String extraQueryParam,
                                          final String libraryVersion) {
         super(responseType, clientId, redirectUri, state, scope);
-        setState(generateState());
 
         if (StringUtil.isEmpty(redirectUri)) {
             throw new IllegalArgumentException("redirect Uri is empty");
@@ -210,10 +210,20 @@ public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest
         }
     }
 
-    protected String generateState() {
+    public String generateEncodedState() throws UnsupportedEncodingException {
         final UUID stateUUID1 = UUID.randomUUID();
         final UUID stateUUID2 = UUID.randomUUID();
-        return stateUUID1.toString() + "-" + stateUUID2.toString();
+        final String state = stateUUID1.toString() + "-" + stateUUID2.toString();
+        return Base64.encodeToString(state.getBytes("UTF-8"), Base64.NO_PADDING | Base64.URL_SAFE);
+    }
+
+    public String decodeState(final String encodedState) {
+        if (StringUtil.isEmpty(encodedState)) {
+            return null;
+        }
+
+        final byte[] stateBytes = Base64.decode(encodedState, Base64.NO_PADDING | Base64.URL_SAFE);
+        return new String(stateBytes, Charset.defaultCharset());
     }
 
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
@@ -49,6 +49,11 @@ import static com.microsoft.identity.common.adal.internal.util.StringExtensions.
 import static com.microsoft.identity.common.adal.internal.util.StringExtensions.urlFormDecode;
 
 public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest {
+    /**
+     * Serial version id.
+     */
+    private static final long serialVersionUID = 6873634931996113294L;
+
     /* Constants */
     private static final String TAG = MicrosoftAuthorizationRequest.class.getSimpleName();
     public static final String ENCODING_UTF8 = "UTF_8";
@@ -63,7 +68,6 @@ public abstract class MicrosoftAuthorizationRequest extends AuthorizationRequest
     public static final String LIB_ID_CPU = "x-client-CPU";
     public static final String LIB_ID_OS_VER = "x-client-OS";
     public static final String LIB_ID_DM = "x-client-DM";
-    private static final long serialVersionUID = 6873634931996113294L;
 
     /**
      * Required.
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -71,25 +75,24 @@ public GenericAuthorizationResult requestAuthorization(final GenericAuthorizatio
...
     }
 
     /**
      * Set up the web view configurations.
-     * @param activity  AuthenticationActivity
      * @param webViewClient AzureActiveDirectoryWebViewClient
      */
-    private void createWebView(final GenericWebViewClient webViewClient) {
-        final Activity activity = webViewClient.getActivity();
+    private void createWebView(final GenericWebViewClient webViewClient, final WebView webView) {
```
iambmelt: Given that this method accepts a `WebView` as a parameter, the name `createWebView()` seems a bit of a misnomer.... maybe.... `setUpWebView()` or `initializeWebView()`? Open to other thoughts :-)

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -62,6 +63,7 @@
     public static final String LIB_ID_CPU = "x-client-CPU";
     public static final String LIB_ID_OS_VER = "x-client-OS";
     public static final String LIB_ID_DM = "x-client-DM";
+    private static final long serialVersionUID = 6873634931996113294L;
```
piqiums: please use all upper case letters for constants.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -46,14 +46,15 @@
...
-    private static final String CORRELATION_ID = "correlation_id";
+    private static final String CORRELATION_ID = "client-request-id";
     private static final String LOGIN_REQ = "login_req";
     private static final String DOMAIN_REQ = "domain_req";
     private static final String SCOPE_PROFILE = "profile";
     private static final String[] RESERVED_SCOPES = {"openid", SCOPE_PROFILE, "offline_access"};
     private static final String PLATFORM_VALUE = "MSAL.Android";
     private static final String PROMPT_SELECT_ACCOUNT = "select_account";
     private static final String PROMPT_CONSENT = "consent";
+    private static final long serialVersionUID = 6545759826515911472L;
```
piqiums: please use all upper case letters for constants.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -39,7 +39,7 @@
  * This should provide an extension point for additional parameters to be set
  */
 public abstract class AuthorizationRequest implements Serializable {
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 6171895895590170062L;
```
piqiums: ditto

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
iambmelt: This value should be left as-is: [it is used by the Android Runtime to compute when the classes change between versions](https://developer.android.com/reference/java/io/Serializable). Its value is inspected through reflection.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -202,10 +210,20 @@ protected void appendExtraQueryParameters(final String queryParams, final Map<St
...
+    public String generateEncodedState() throws UnsupportedEncodingException {
         final UUID stateUUID1 = UUID.randomUUID();
         final UUID stateUUID2 = UUID.randomUUID();
-        return stateUUID1.toString() + "-" + stateUUID2.toString();
+        final String state = stateUUID1.toString() + "-" + stateUUID2.toString();
+        return Base64.encodeToString(state.getBytes("UTF-8"), Base64.NO_PADDING | Base64.URL_SAFE);
+    }
+
+    public String decodeState(final String encodedState) {
+        if (StringUtil.isEmpty(encodedState)) {
```
iambmelt: Could consider logging a warning here

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -46,14 +46,15 @@
     private static final String TAG = MicrosoftStsAuthorizationRequest.class.getSimpleName();
 
     /* Constants */
-    private static final String CORRELATION_ID = "correlation_id";
+    private static final String CORRELATION_ID = "client-request-id";
```
iambmelt: `correlationId` and `client-request-id` are not the same value

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
iambmelt: Leave as-is. This is a special system value.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -96,6 +96,12 @@ public AuthorizationRequest(final String responseType,
         mScope = scope;
     }
 
+    /**
+     * Default constructor of AuthorizationRequest.
+     */
+    public AuthorizationRequest() {
```
piqiums: there's a default ctor without any args, so this ctor is unnecessary if it's empty.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -58,10 +60,12 @@
      * @return generic AuthorizationResult
      */
     public GenericAuthorizationResult requestAuthorization(final GenericAuthorizationRequest request) {
```
iambmelt: Consider declaring this method [`@Nullable`](https://developer.android.com/reference/android/support/annotation/Nullable) to convey its potential return value

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
addev-ashish: unfortunately, correlation_id goes under client-request-id field in the header, the code is correct if that is the usage.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
heidijinxujia: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
heidijinxujia: https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/f3861311f0ea1ee32cd30da7b548f8f701515775/msal/src/internal/java/com/microsoft/identity/client/OauthConstants.java#L79

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -57,11 +61,13 @@
...
+    public GenericAuthorizationResult requestAuthorization(@Nullable final GenericAuthorizationRequest request) {
+        Logger.verbose(TAG, "Perform the authorization request with embedded webView.");
         loadStartUrl();
-        // TODO : Add state parameter from the AuthorizationRequest to the Intent parameter like below
-        // intent.putExtra(MicrosoftAuthorizationResult.REQUEST_STATE_PARAMETER, request.getState());
-        throw new UnsupportedOperationException("Not implemented yet.");
+        // requestAuthorization could not return the authorization result
+        // The activity result is set in AuthenticationActivity.setResult()
+        // And AuthenticationActivity in ADAL/MSAL is not moved into common
+        return null;
```
iambmelt: If this cannot be returned from here, I suggest setting `Void` as the return type in:
`... extends AuthorizationStrategy<Void, GenericAuthorizationResult>` although this also means that inheritors of this class must `@Override` this implementation if they want a return value....

Does this method need to be abstract?

PR #175, opened 2018-07-23
add a new constants into AuthenticationConstants

piqiums:
add a new constants into AuthenticationConstants

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -151,6 +151,9 @@ public final class AuthenticationConstants {
          */
         public static final int BROKER_REQUEST_RESUME = 2006;
 
+        /**
+         * Device registration in broker apps.
+         */
         public static final int BROWSER_CODE_DEVICE_REGISTER = 2007;
     }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -150,6 +150,8 @@ private AuthenticationConstants() {
          * Waiting for broker package to be installed, and resume request in broker.
          */
         public static final int BROKER_REQUEST_RESUME = 2006;
+
+        public static final int BROWSER_CODE_DEVICE_REGISTER = 2007;
```
iambmelt: Is this value [changing](https://github.com/AzureAD/azure-activedirectory-library-for-android-unity-fork/blob/a0f8be665339992162f624a67eba18bb2e89856e/adal-unity-fork/src/main/java/com/microsoft/aad/adal/unity/AuthenticationConstants.java#L89)?

piqiums: this a result code merged from Unity-fork. i.e. Activity A is trying to start Activity B to do something, A will call startActivityForResult ( https://developer.android.com/reference/android/app/Activity.html#startActivityForResult(android.content.Intent,%20int) ) with a request code, and after B has done the job and return the result to A, it will call onActivityResult ( https://developer.android.com/reference/android/app/Activity.html#onActivityResult(int,%20int,%20android.content.Intent) ) with a result value, here it's defined in UIResponse as constants.

PR #176, opened 2018-07-30
Merge ui/dev into dev

heidijinxujia:
ui/dev includes
1. Embedded Webview
2. Authorization Request
3. Authorization Result


common/src/androidTest/java/com/microsoft/identity/common/MicrosoftStsAuthorizationRequestTests.java
```diff
@@ -0,0 +1,180 @@
...
+    private static final String DEFAULT_TEST_DISPLAYABLEID = "user@contoso.com";
+    private static final String DEFAULT_TEST_SLICE_PARAMETER = "slice=myslice";
+    private static final Set<String> DEFAULT_TEST_EXTRA_SCOPE = new HashSet<>(Arrays.asList(DEFAULT_TEST_EXTRA_SCOPE_ARRAY));
+    private static final String DEFAULT_TEST_AUTHORITY_STRING = "https://login.microsoftonline.com/common";
+
+    static URL getValidRequestUrl() throws MalformedURLException {
+        return new URL(DEFAULT_TEST_AUTHORITY_STRING);
+    }
+
+    static MicrosoftStsAuthorizationRequest createAuthenticationRequest(final String responseType,
```
shoatman: Why is a method called createAuthenticationRequest returning an Authorization Request?  Any reason to prefer a static method over an actual constructor?  Should we consider having the MicrosoftSTSOAuth2Strategy provide methods for creating the authorization request.

common/src/main/java/com/microsoft/identity/common/adal/internal/JWSBuilder.java
```diff
@@ -0,0 +1,209 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package com.microsoft.identity.common.adal.internal;
```
shoatman: I'd be fine with using the existing nimbus dependency to help create the signed JWT token.

common/src/main/java/com/microsoft/identity/common/internal/cache/ADALOAuth2TokenCache.java
```diff
@@ -134,7 +135,7 @@ public void saveTokens(
         Logger.info(TAG + ":" + methodName, "Setting items to cache for user...");
         for (final String cacheIdentifier : account.getCacheIdentifiers()) {
             //Azure AD Uses Resource and Not Scope... but we didn't override... heads up
-            final String scope = request.getScope();
+            final String scope = StringUtil.convertSetToString(request.getScope(), " ");
```
shoatman: Did we change scope string in request to a set?  If so, why?  The intention of the Authorization request is to be as close to the wire as possible.  It's a space delimited string in the spec... not a set.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationErrorResponse.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final String NULL_INTENT = "Received null intent";
+
+    /**  Error string to return for invalid response from Authorization server. */
+    public static final String AUTHORIZATION_SERVER_INVALID_RESPONSE = "The authorization server returned an invalid response.";
+
+    /**  Error description string to return if the user cancelled the flow.*/
+    public static final String USER_CANCELLED_FLOW = "User pressed device back button to cancel the flow.";
+
+    /**  Error string to return if the state parameter from authorization endpoint doesn't match with the request state. */
+    public static final String STATE_NOT_THE_SAME = "Returned state from authorize endpoint is not the same as the one sent";
```
shoatman: Is this returned from the server?  We should only have things here that are returned from the server.  Please review the rest with this in mind.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationRequest.java
```diff
@@ -22,34 +22,175 @@
...
+        final String authorizationUrl = StringExtensions.appendQueryParameterToUrl(
+                getAuthorizationEndpoint(),
+                createAuthorizationRequestParameters());
+        Logger.infoPII(TAG, getCorrelationId().toString(), "Request uri to authorize endpoint is: " + authorizationUrl);
+        return authorizationUrl;
+    }
+
+    private void addDiagnosticParameters(@NonNull final Map<String, String> requestParameters) {
+        // append device and platform info in the query parameters
+        requestParameters.put(LIB_ID_PLATFORM, PLATFORM_VALUE);
```
shoatman: I'd rather just serialize the authorization request... rather than constructing it with methods.  I'm fine with checking part in as in and then refactoring.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
```diff
@@ -0,0 +1,193 @@
...
+ * {@link AzureActiveDirectoryAuthorizationResult}
+ */
+public class AzureActiveDirectoryAuthorizationResultFactory extends AuthorizationResultFactory<AzureActiveDirectoryAuthorizationResult> {
+
+    private static final String TAG = AzureActiveDirectoryAuthorizationResultFactory.class.getSimpleName();
+
+    private static final String ERROR_CODES = "error_codes";
+
+    @Override
+    public AzureActiveDirectoryAuthorizationResult createAuthorizationResult(final int resultCode, final Intent data) {
```
shoatman: I would want to refactor this as well... would be better for the authorizationStrategy to be the only class that knows about the resultcodes used by that strategy.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryPromptBehavior.java
```diff
@@ -0,0 +1,60 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory;
+
+/**
+ * Prompt Behaviors that sets the behavior for authentication activity launch.
+ */
+
+public enum AzureActiveDirectoryPromptBehavior {
```
shoatman: Prompt is part of the OIDC spec.  We should likely put this in the authorization request....  I believe that both the AAD v1 and V2 endpoints will have extensions to what's in the spec, but let's start with what's in the spec.  http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/challengehandlers/PKeyAuthChallenge.java
```diff
@@ -0,0 +1,112 @@
+package com.microsoft.identity.common.internal.ui.embeddedwebview.challengehandlers;
```
iambmelt: Missing license info

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    public void setLibraryVersion(final String libraryVersion) {
+        mLibraryVersion = libraryVersion;
+    }
+
+    /**
+     * Convert the query parameters string to pair and add into the request parameters map.
+     */
+    protected void appendExtraQueryParameters(final String queryParams, final Map<String, String> requestParams) throws ClientException {
```
iambmelt: The behavior of this method is....odd because its second parameter is functioning as the return value. Why not just return a new `Map`?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -0,0 +1,279 @@
...
+                                         @NonNull final String clientId,
+                                         @NonNull final String redirectUri,
+                                         final String state,
+                                         final Set<String> scope,
+                                         @NonNull final URL authority,
+                                         @NonNull final String authorizationEndpoint,
+                                         final String loginHint,
+                                         final UUID correlationId,
+                                         final PkceChallenge pkceChallenge,
+                                         final String extraQueryParam,
```
iambmelt: No javadoc expressing the expectations around format for the extra qp.... I would more intuitively expect this parameter to be a `Map` than a `String`....

are they encoded?
space delimited?
more than one are acceptable?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -0,0 +1,279 @@
...
+
+    public void setPkceChallenge(final PkceChallenge pkceChallenge) {
+        mPkceChallenge = pkceChallenge;
+    }
+
+    public String getExtraQueryParam() {
+        return mExtraQueryParam;
+    }
+
+    public void setExtraQueryParam(final String extraQueryParam) {
```
iambmelt: No notes about the expected format....does this need to be encoded? Can I set multiple query params at once? Or do I call this in a loop?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -0,0 +1,279 @@
...
+            Logger.warn(TAG, "Decode state failed because the input state is empty.");
+            return null;
+        }
+
+        final byte[] stateBytes = Base64.decode(encodedState, Base64.NO_PADDING | Base64.URL_SAFE);
+        return new String(stateBytes, Charset.defaultCharset());
+    }
+
+
+    protected void addExtraQueryParameter(final String key, final String value, final Map<String, String> requestParams) {
```
iambmelt: More strange behavior where parameters mutate other parameters

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+
+    /**
+     * Sanitizes the query portion of a URL and convert the query parameters into a map.
+     *
+     * @param url The url to decode for.
+     * @return A map of all the query parameter pairs in the URL's query portion.
+     */
+    private static Map<String, String> constructQueryParamsMap(final String url) {
```
iambmelt: This implementation makes a lot of assumptions about the formatting of query params....

Consider the following _valid_ URL and how this code would treat it:
`http://foo.bar?list_a=1&list_a=2&list_a=3&list_b[]=1&list_b[]=2&list_b[]=3&list_c=1,2,3`

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -22,7 +22,276 @@
...
+        //TODO can wrap the user info into User class object.
+        addExtraQueryParameter(LOGIN_REQ, mUid, requestParameters);
+        addExtraQueryParameter(DOMAIN_REQ, mUtid, requestParameters);
+        addExtraQueryParameter(LOGIN_HINT, mDisplayableId, requestParameters);
+    }
+
+    /**
+     * Add the extra query parameters and slice parameters into the request parameter map.
+     */
+    private void addExtraQueryParameter(@NonNull final Map<String, String> requestParameters) throws ClientException {
```
iambmelt: More parameters mutating parameters

heidijinxujia:
I will create another PR out of dev to address the comments. Thanks.

PR #178, opened 2018-07-31
Shoatman get issuer bug

shoatman:


iambmelt:
This fix must be cherry picked into release

PR #181, opened 2018-07-31
Removing @VisibleForTesting annot

iambmelt:
This method is now called outside of tests, removing annot (and thus, the warning it generates)


iambmelt:
This will be in RC3

PR #183, opened 2018-08-02
APIcontroller updates

shoatman:
Merged changes.  These likely won't pass all existing unit tests... but the code does build.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectoryb2c/AzureActiveDirectoryB2CAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectoryb2c/AzureActiveDirectoryB2CAuthorizationRequest.java
@@ -59,4 +59,10 @@ public class AzureActiveDirectoryB2CAuthorizationRequest extends MicrosoftAuthor
         super(responseType, clientId, redirectUri, state, scope, authority, authorizationEndpoint,
                 loginHint, correlationId, pkceChallenge, extraQueryParam, libraryVersion);
     }
+
+    @Override
+    public String getAuthorizationEndpoint() {
+        return null;
+
+    }
 }
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectoryb2c/AzureActiveDirectoryB2CAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectoryb2c/AzureActiveDirectoryB2CAuthorizationRequest.java
@@ -48,15 +48,14 @@ public class AzureActiveDirectoryB2CAuthorizationRequest extends MicrosoftAuthor
                                                        @NonNull final String clientId,
                                                        final String redirectUri,
                                                        final String state,
-                                                       final Set<String> scope,
+                                                       final String scope,
                                                        @NonNull final URL authority,
-                                                       @NonNull final String authorizationEndpoint,
                                                        final String loginHint,
                                                        final UUID correlationId,
                                                        final PkceChallenge pkceChallenge,
                                                        final String extraQueryParam,
                                                        final String libraryVersion) {
-        super(responseType, clientId, redirectUri, state, scope, authority, authorizationEndpoint,
+        super(responseType, clientId, redirectUri, state, scope, authority,
                 loginHint, correlationId, pkceChallenge, extraQueryParam, libraryVersion);
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
@@ -25,6 +25,7 @@ package com.microsoft.identity.common.internal.providers.microsoft.microsoftsts;
 import android.os.Build;
 import android.support.annotation.NonNull;
 
+
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.exception.ClientException;
@@ -293,5 +294,12 @@ public class MicrosoftStsAuthorizationRequest extends MicrosoftAuthorizationRequ
         // If excluded, code_challenge is assumed to be plaintext if code_challenge is included.
         // Azure AAD v2.0 supports both plain and S256.
         requestParameters.put(CODE_CHALLENGE_METHOD, getPkceChallenge().getCodeChallengeMethod());
+
+    }
+
+    @Override
+    public String getAuthorizationEndpoint(){
+        //TODO: Need to take authority aliasing via instance discovery into account here
+        return "https://login.microsoftonline.com/common/oauth2/v2.0/authorize";
     }
 }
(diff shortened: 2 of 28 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectoryb2c/AzureActiveDirectoryB2CAuthorizationRequest.java
```diff
@@ -48,15 +48,20 @@ public AzureActiveDirectoryB2CAuthorizationRequest(final String responseType,
...
                                                        final String extraQueryParam,
                                                        final String libraryVersion) {
-        super(responseType, clientId, redirectUri, state, scope, authority, authorizationEndpoint,
+        super(responseType, clientId, redirectUri, state, scope, authority,
                 loginHint, correlationId, pkceChallenge, extraQueryParam, libraryVersion);
     }
+
+    @Override
+    public String getAuthorizationEndpoint() {
+        return null;
```
iambmelt: Is this a `//TODO` ?

shoatman: Yes.  We haven't fully implemented anything for B2C yet.

common/src/main/java/com/microsoft/identity/common/internal/ui/embeddedwebview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -130,4 +122,19 @@ public void run() {
             }
         });
     }
+
+    @Override
+    public Future<AuthorizationResult> requestAuthorization(AuthorizationRequest request) throws UnsupportedEncodingException {
```
iambmelt: Not seeing where this `UnsupportedEncodingException` is thrown...

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -54,8 +55,8 @@
     private static final String CORRELATION_ID = "client-request-id";
     private static final String LOGIN_REQ = "login_req";
     private static final String DOMAIN_REQ = "domain_req";
-    private static final String SCOPE_PROFILE = "profile";
-    private static final String[] RESERVED_SCOPES = {"openid", SCOPE_PROFILE, "offline_access"};
+    //TODO: Should this be in the request or in the oAuth strategy?
```
heidijinxujia: Is there any reason why we want to put it into oauth strategy? It is easier to understand if we keep the reserved scopes inside the request. Because RESERVED_SCOPES is part of scope. And scope belongs to authorization request.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -293,5 +255,12 @@ private void addPkceChallengeToRequestParameters(@NonNull final Map<String, Stri
         // If excluded, code_challenge is assumed to be plaintext if code_challenge is included.
         // Azure AAD v2.0 supports both plain and S256.
         requestParameters.put(CODE_CHALLENGE_METHOD, getPkceChallenge().getCodeChallengeMethod());
+
+    }
+
+    @Override
+    public String getAuthorizationEndpoint(){
+        //TODO: Need to take authority aliasing via instance discovery into account here
+        return "https://login.microsoftonline.com/common/oauth2/v2.0/authorize";
```
heidijinxujia: It seems the AuthorizationEndpoint is not a property of request anymore. I'm confused on the getter here. Where should it set the AuthorizationEndpoint value? Are we going to move the instance discovery functions into the getAuthorizationEndpoint()?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -72,12 +83,17 @@
      * encode information about the user's state in the app before the authentication request
      * occurred, such as the page or view they were on.
      */
+    @SerializedName("state")
     private String mState;
 
+    //Marking as transient to avoid these values being serialized by GSON or other Java Serialization
+    private transient Activity mActivity;
```
heidijinxujia: Why we need these two properties in the authorization request? Shouldn't it belong to the authorization strategy where handling the UI flow?

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -67,4 +70,15 @@ public static String convertSetToString(final Set<String> inputSet, final String
 
         return stringBuilder.toString();
     }
+
```
heidijinxujia: Feel confused on the usage between `com.microsoft.identity.common.internal.util.StringUtil` and `com.microsoft.identity.common.adal.internal.util.StringExtensions`.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
shoatman: Maybe... i'd like to ponder this a bit.  Trying to keep plain old java objects used for serialization as simple as possible.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
shoatman: Let's talk about this.  Instance discovery should be part of AAD... not part of a request.

iambmelt:
@shoatman In #192 I added the generics to the `Future` / `AuthorizationResult` code paths such that consumers (internal) can specify the exact return type -- I generally think this degree of type-safety is nice to have to cut down on big `if`/`else` blocks for the various subclasses. I'll leave it to your discretion whether or not you'd like to see those suggestions merged into this PR

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
shoatman: I tend to agree.  @iambmelt gave the same feedback.  I'll take a look.

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
shoatman: ADAL is for ADAL... longer term we should deprecate the use of the ADAL version.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
heidijinxujia: To keep AuthorizationRequest "plain", do you also suggest we should move the authorization url constructor to somewhere else?

shoatman: Yes.  I think we should do that.  We can continue to make changes as we go though

shoatman:
@iambmelt - Ok to re-approve?  Then i'll merge.

PR #184, opened 2018-08-02
VSTS Build automation changes in common

kreedula:


common/versioning/version.properties
```diff
@@ -0,0 +1,3 @@
+#Thu Aug 02 12:03:16 PDT 2018
+versionName=0.0.1
```
iambmelt: I suspect the 'correct' thing to do here would be to start this value at `0.0.2` and rev to `0.0.3` when `release/0.0.3` merges to `dev`

kreedula:
Thank you @iambmelt . Makes sense to update dev to 0.0.2 as release is at 0.0.3. Made changes as needed.

kreedula:
Thank you for the review @iambmelt

PR #186, opened 2018-08-08
Commented out Redirect URI validation check and added a TODO to refactor

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -117,9 +117,11 @@ public MicrosoftAuthorizationRequest(final String responseType,
                                          final String libraryVersion) {
         super(responseType, clientId, redirectUri, state, scope);
 
+        //TODO : Refactor to move the validation logic of all NonNull values before the request is constructed.
```
iambmelt: @kreedula @addev-ashish @heidijinxujia Will we _actually_ ever uncomment this? Or must this be wholesale removed to support the silent auth case?

kreedula: Per my understanding this check should be removed from here and performed at a higher level as the Silent auth case will fail. I just commented it out so that during the refactor we would miss moving/add this at a higher level where the request is created. I happy to remove this completely and add specific comment to add this later

kreedula:
Thanks @iambmelt Removed the commented out code and added a comment

PR #194, opened 2018-08-14
Changes to OAuth2TokenCache internals and api surface to support MSAL getAccounts()

iambmelt:


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -161,6 +162,182 @@ public void saveTokens(
...
+        // Get all of the Accounts for this environment
+        final List<Account> accountsForEnvironment =
+                mAccountCredentialCache.getAccountsFilteredBy(
+                        null, // wildcard (*) homeAccountId
+                        environment,
+                        null // wildcard (*) realm
+                );
+
+        // Declare a List to hold the MicrosoftStsAccounts
+        final List<Account> microsoftStsAccounts = new ArrayList<>();
```
shoatman: We probably shouldn't assume these accounts are MSTS accounts.  What about ADFS, B2C?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -161,6 +162,182 @@ public void saveTokens(
...
+
+        // Grab all of the Accounts supplied by either the v1 or v2 endpoint
+        for (final Account account : accountsForEnvironment) {
+            if (MicrosoftAccount.AUTHORITY_TYPE_V1_V2.equals(account.getAuthorityType())) {
+                microsoftStsAccounts.add(account);
+            }
+        }
+
+        // Grab the Credentials for this app...
+        final List<Credential> appCredentials =
```
shoatman: appCredentials is a potentially misleading name.... do you mean get all credentials currently in the cache?  If so... perhaps... allCredentials?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -161,6 +162,182 @@ public void saveTokens(
...
+                        environment,
+                        CredentialType.RefreshToken,
+                        clientId,
+                        null, // realm
+                        null // target
+                );
+
+        // For each Account with an associated RT, add it to the result List...
+        for (final Account account : microsoftStsAccounts) {
+            if (accountHasToken(account, appCredentials)) {
```
shoatman: rather than accountHasToken....accountHasCredential

iambmelt: Resolved via c8cffe58b5905762221cd6ccce2d5a95db888404

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Resolved via ea7e9ee4b7eb23135ad0223021b7dcfa5175e817

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Going to leave as-is; these credentials are loaded only for the current clientId (see param to `getCredentialsFilteredBy`)

PR #196, opened 2018-08-15
Implement authorization flow with device browser

heidijinxujia:
Add multi-browser support for authorization flow, including Browser, BrowserBlockList and BrowserSelector implemention.

Add BrowserAuthorizationStrategy to handle the initialization and completion of browser auth flow.

Add AuthorizationStrategyFactory to generate BrowserAuthorizationStrategy or EmbeddedWebViewAuthorizationStrategy according to the pass-in configuration class.

Add the AuthorizationActivity to perform the authorization activity flow.

If the dev wants to specify the ui flow to use embedded webview, you need to call
`AuthorizationConfiguration.getInstance().setAuthorizationAgent(AuthorizationAgent.WEBVIEW);`
before initializing the authorization strategy. 

The dev can create the AuthorizationStrategy by calling `authorizationStrategy = AuthorizationStrategyFactory.getAuthorizationStrategy(Activity, AuthorizationConfiguration)`

And start the auth flow by calling
`authorizationStrategy.requestAuthorization(final Uri requestUrl)`

The dev can add the ref in the app activity.onActivityResult with authorizationStrategy.completeAuthorization() which will assign the auth result to authorizationStrategy.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/ResponseType.java
```diff
@@ -0,0 +1,33 @@
+package com.microsoft.identity.common.internal.providers.oauth2;
+
+/**
+ * The response type values defined by the
+ * [OAuth 2.0](https://tools.ietf.org/html/rfc6749) and
+ * [OpenID Connect Core 1.0](http://openid.net/specs/openid-connect-core-1_0.html)
+ **/
+
+public final class ResponseType {
```
shoatman: This is fine for now.  Let's discuss where to put these constants... I took a somewhat different approach.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/Browser.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param version         The version name of the browser.
+     */
+    public Browser(@NonNull String packageName, @NonNull Set<String> signatureHashes, @NonNull String version, boolean isCustomTabsServiceSupported) {
+        mPackageName = packageName;
+        mSignatureHashes = signatureHashes;
+        mVersion = version;
+        mIsCustomTabsServiceSupported = isCustomTabsServiceSupported;
+    }
+
+    public Browser setCustomTabsServiceSupported(final Boolean isCustomTabsServiceSupported) {
```
shoatman: Not sure why this method is returning a new Browser object.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -0,0 +1,88 @@
...
+                null,
+                requestUrl.toString(),
+                mConfiguration);
+        mReferencedActivity.get().startActivityForResult(authIntent, BROWSER_FLOW);
+        return mAuthorizationResultFuture;
+    }
+
+    @Override
+    public void completeAuthorization(int requestCode, int resultCode, Intent data) {
+        if (requestCode != BROWSER_FLOW) {
```
shoatman: I'm wondering whether we should just ignore.... in this case... rather than throwing an exception?  Meaning that perhaps we just return and not throw an exception.

PR #197, opened 2018-08-16
First draft api for OAuth2TokenCache extensions

iambmelt:
Extensions for the `OAuth2TokenCache` interface to support cool new stuff like:

- Get out of the cache exactly what goes in when tokens are saved
- Get the `AccessToken`, `RefreshToken`, and `Idtoken` of a supplied `Account`
- Delete a provided `Credential` (This functionality already exists, but is only exposed through `IAccountCredentialCache`)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/ADALOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/ADALOAuth2TokenCache.java
@@ -34,6 +34,7 @@ import com.microsoft.identity.common.adal.internal.cache.CacheKey;
 import com.microsoft.identity.common.adal.internal.cache.DateTimeAdapter;
 import com.microsoft.identity.common.adal.internal.cache.StorageHelper;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
+import com.microsoft.identity.common.internal.dto.Credential;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftAccount;
 import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftRefreshToken;
@@ -56,6 +57,7 @@ import java.util.List;
 public class ADALOAuth2TokenCache
         extends OAuth2TokenCache<AzureActiveDirectoryOAuth2Strategy, AzureActiveDirectoryAuthorizationRequest, AzureActiveDirectoryTokenResponse>
         implements IShareSingleSignOnState {
+    private static final String ERR_UNSUPPORTED_OPERATION = "This method is unsupported by the ADALOAuth2TokenCache";
     private ISharedPreferencesFileManager mISharedPreferencesFileManager;
 
     private static final String TAG = ADALOAuth2TokenCache.class.getSimpleName();
@@ -115,7 +117,7 @@ public class ADALOAuth2TokenCache
      * @param response
      */
     @Override
-    public void saveTokens(
+    public ISaveTokenResult saveTokens(
             final AzureActiveDirectoryOAuth2Strategy strategy,
             final AzureActiveDirectoryAuthorizationRequest request,
             final AzureActiveDirectoryTokenResponse response) {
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/ADALOAuth2TokenCache.java
```diff
@@ -155,22 +157,38 @@ public void saveTokens(
...
             sharedSsoCache.setSingleSignOnState(account, refreshToken);
         }
+
+        return null; // Returning null, since the ADAL cache's schema doesn't support this return type.
+    }
+
+    @Override
+    public ISaveTokenResult loadTokens(com.microsoft.identity.common.internal.dto.Account account) {
+        throw new UnsupportedOperationException(
+                ERR_UNSUPPORTED_OPERATION
```
iambmelt: I'm not thrilled about to do this a ton, but I don't see a fantastic way to avoid it other that hardwiring the dependency in ADAL to `AdalOAuth2TokenCache` and having it no longer implement this `interface`.

^^ This may not be a half-bad idea, come to think of it. +1 if you agree

common/src/main/java/com/microsoft/identity/common/internal/cache/ISaveTokenResult.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.microsoft.identity.common.internal.dto.AccessToken;
+import com.microsoft.identity.common.internal.dto.Account;
+import com.microsoft.identity.common.internal.dto.IdToken;
+import com.microsoft.identity.common.internal.dto.RefreshToken;
+
+/**
+ * Result container for Account and Credential - usually the result of a save or load operation.
+ */
+public interface ISaveTokenResult {
```
iambmelt: I'm not thrilled about this interface's name, tbh -- open to suggestions.

common/src/main/java/com/microsoft/identity/common/internal/cache/ISaveTokenResult.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return The Account to get.
+     */
+    Account getAccount();
+
+    /**
+     * Gets the {@link AccessToken}.
+     *
+     * @return The AccessToken to get.
+     */
+    AccessToken getAccessToken();
```
iambmelt: @shoatman Would it be preferable to just return a `List<Credential>` and forget the various `Credential` subtypes entirely? (Thinking about x-IdP compat, sanity)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2TokenCache.java
```diff
@@ -52,11 +54,28 @@ public OAuth2TokenCache(Context context) {
...
+                                                final U request,
+                                                final V response) throws ClientException;
+
+    /**
+     * Loads the tokens for the supplied Account into the result {@link ISaveTokenResult}.
+     *
+     * @param account The Account whose Credentials should be loaded.
+     * @return The resulting ISaveTokenResult. Entries may be empty if not present in the cache.
+     */
+    public abstract ISaveTokenResult loadTokens(final Account account);
```
iambmelt: This method's signature is likely going to change to also accept a `clientId` so that these transactions are app-specific.

common/src/main/java/com/microsoft/identity/common/internal/cache/ISaveTokenResult.java
iambmelt: Discussed offline, going to leave as-is

common/src/main/java/com/microsoft/identity/common/internal/cache/ISaveTokenResult.java
kreedula: True as this is being returned on loadTokens() as well. That said I don't have recommendation either.  It would have been more intuitive if save tokens api just saved the tokens and returned success/false and loadTokens actually returned the result with tokens

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -160,6 +161,101 @@ public void saveTokens(
...
+        final List<Credential> idTokens = mAccountCredentialCache.getCredentialsFilteredBy(
+                account.getHomeAccountId(),
+                account.getEnvironment(),
+                CredentialType.IdToken,
+                clientId,
+                account.getRealm(),
+                null // wildcard (*)
+        );
+
+        // Check that there was only 1 result for each of these
```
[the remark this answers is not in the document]
iambmelt: Yeah - I'll yank this out. This can be written better

common/src/main/java/com/microsoft/identity/common/internal/cache/ISaveTokenResult.java
iambmelt: Offline w/ @shoatman `ICacheResult`

PR #200, opened 2018-08-22
Bugfix for scopes ordering, completeness + tests

iambmelt:


common/src/main/java/com/microsoft/identity/common/internal/cache/AccountCredentialCache.java
```diff
@@ -334,6 +334,30 @@ public synchronized Credential getCredential(@NonNull final String cacheKey) {
         return matchingCredentials;
     }
 
+    private boolean targetsIntersect(final String soughtTarget, final String credentialTarget) {
```
shoatman: targetToMatch instead of soughtTarget?

PR #202, opened 2018-08-22
Implements isExpired() for access tokens

iambmelt:
Matches the [soon-to-be-removed implementation in `AccessTokenCacheItem`](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/cache/java/com/microsoft/identity/client/AccessTokenCacheItem.java#L133-L143).

Credential types other than AT return `false`, as expiry is not tracked for these types.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
@@ -23,6 +23,10 @@
 package com.microsoft.identity.common.internal.dto;
 
 import com.google.gson.annotations.SerializedName;
+import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
+
+import java.util.Calendar;
+import java.util.Date;
 
 import static com.microsoft.identity.common.internal.dto.AccessToken.SerializedNames.ACCESS_TOKEN_TYPE;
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
```diff
@@ -241,4 +245,16 @@ public String getExpiresOn() {
...
+    @Override
+    public boolean isExpired() {
+        // Init a Calendar for the current time/date
+        final Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.SECOND, AuthenticationSettings.INSTANCE.getExpirationBuffer());
+        final Date validity = calendar.getTime();
+        // Init a Date for the accessToken's expiry
+        long epoch = Long.valueOf(getExpiresOn());
+        final Date expiresOn = new Date(epoch * 1000);
+        return expiresOn.before(validity);
```
addev-ashish: we need 5 minutes for clock skew, is that handled somewhere else?

iambmelt: Line 253

addev-ashish: oops

PR #206, opened 2018-08-24
Refactor AuthorizationRequest and its subclasses

### LINKED ISSUE #207: Refactor the AuthorizationRequest

### LINKED ISSUE #211: Fix the nested json string bug for authorizaiton url construction
Encounter the illegal argument exception when trying to serialize the pkcechallenge property of authorization request.
```Expected a string but was BEGIN_OBJECT at line 1 column 254 path $.```

### PR DESCRIPTION
**Pull request by heidijinxujia**
Apply generic builder pattern to auth request and its subclasses.
Close #207 
Close #211

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
@@ -24,11 +24,17 @@ package com.microsoft.identity.common.internal.net;
 
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import com.microsoft.identity.common.internal.logging.Logger;
+import com.microsoft.identity.common.internal.util.StringUtil;
 
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
+import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
+import java.util.StringTokenizer;
 import java.util.TreeMap;
 
 public final class ObjectMapper {
@@ -37,6 +43,7 @@ public final class ObjectMapper {
      * Encoding scheme.
      */
     public static final String ENCODING_SCHEME = "UTF-8";
+    public static final String TAG = ObjectMapper.class.getSimpleName();
 
     private ObjectMapper() {
         // Utility class.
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java [hunk 1]
```diff
@@ -96,4 +103,34 @@ public static String serializeObjectToFormUrlEncoded(Object object) throws Unsup
         return builder.toString();
     }
 
+    public static Map<String, String> extraQueryString(final String extraQueryString) {
```
shoatman: deserializeQueryStringToMap for method name?  queryString for parameter name?  Rather than this lower level parsing... do you want to use this instead: https://developer.android.com/reference/android/net/Uri.html#getQueryParameterNames()

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -22,65 +22,46 @@
...
-    public static final String LIB_ID_VERSION = "x-client-Ver";
-    public static final String LIB_ID_CPU = "x-client-CPU";
-    public static final String LIB_ID_OS_VER = "x-client-OS";
-    public static final String LIB_ID_DM = "x-client-DM";
 
     /**
      * Required.
      */
-    private URL mAuthority;
+    private URL mAuthority; //Not going to be serialized into request url parameters
```
shoatman: we should like apply the transient modifier here to ensure that it's not serialized.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -211,49 +249,16 @@ public String decodeState(final String encodedState) {
         return new String(stateBytes, Charset.defaultCharset());
     }
 
+    @Override
+    public Uri getAuthorizationRequestAsHttpRequest() throws UnsupportedEncodingException {
+        String queryStringParameters = ObjectMapper.serializeObjectToFormUrlEncoded(this);
+        Uri.Builder uriBuilder = Uri.parse(getAuthorizationEndpoint()).buildUpon()
```
shoatman: When i ran this code yesterday.... there was a problem here since the "?" denoting the beginning of the query string is not added.  

I modifed to this:

HashMap<String, String> params = ObjectMapper.serializeObjectHashMap(this);
        Uri.Builder uriBuilder = Uri.parse(getAuthorizationEndpoint()).buildUpon();

        Iterator it = params.entrySet().iterator();

        while(it.hasNext()){
            Map.Entry<String, String> param = (Map.Entry<String, String>)it.next();
            uriBuilder.appendQueryParameter(param.getKey(), param.getValue());
        }

        return uriBuilder.build();

heidijinxujia: Thanks a lot, Shane. I will work on it.

heidijinxujia:
@shoatman  I've fixed the nested json string bug. And merged dev into this PR. Please let me know whether it looks good to you. Thanks.

heidijinxujia:
Close #210

heidijinxujia:
Yes, the default for response_mode is query in the implementation. Do we need to provide the api to set the response_mode?

heidijinxujia:
Thank you, Shane.

PR #208, opened 2018-08-27
Adding code to support authorities in MSAL and authority discovery

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryInstanceResponse.java
```diff
@@ -0,0 +1,40 @@
+package com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory;
```
iambmelt: license 😏

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -109,6 +126,39 @@ public static void initializeCloudMetadata(final String authorityHost, final Map
...
+                AzureActiveDirectoryInstanceResponse instanceResponse =  ObjectMapper.deserializeJsonStringToObject(response.getBody(), AzureActiveDirectoryInstanceResponse.class);
+
+                for (final AzureActiveDirectoryCloud cloud : instanceResponse.getClouds()) {
+                    cloud.setIsValidated(true); // Mark the deserialized Clouds as validated
+                    for (final String alias : cloud.getHostAliases()) {
+                        sAadClouds.put(alias.toLowerCase(Locale.US), cloud);
+                    }
+                }
+            }
+        } catch (IOException e) {
```
iambmelt: This failure case, if it happens, seems like a good candidate for telemetry as well

PR #214, opened 2018-09-01
Fix bugs for auth flow E2E testing

heidijinxujia:
1. Fix the null pointer exception for nested json deserialization 
2. Fix auth intent issue, replace intent.getStringExtra with intent.getExtras().getString()
3. Set the query flag into MATCH_DEFAULT_ONLY to make sure the default browser is returned as the best choice


common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -133,7 +133,10 @@ public static String serializeObjectToFormUrlEncoded(Object object) throws Unsup
                     && !StringUtil.isEmpty((String)entry.getValue())) {
                 objectMap.put(URLDecoder.decode(entry.getKey(), ENCODING_SCHEME), URLDecoder.decode((String)entry.getValue(), ENCODING_SCHEME));
             } else {
-                objectMap.putAll(serializeNestedJsonToMap(entry.getValue().toString()));
+                final Map<String, String> hashMap = serializeNestedJsonToMap(entry.getValue().toString());
```
iambmelt: Do these values need to be URL-decoded too?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResultFactory.java
```diff
@@ -60,14 +60,24 @@ public MicrosoftStsAuthorizationResult createAuthorizationResult(final int resul
...
-                Logger.verbose(TAG, null, "User cancel the request in webview.");
+                Logger.verbose(TAG, null, "User cancel the authorization request in UI.");
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.USER_CANCEL,
                         MicrosoftAuthorizationErrorResponse.USER_CANCEL, MicrosoftAuthorizationErrorResponse.USER_CANCELLED_FLOW);
                 break;
 
             case AuthenticationConstants.UIResponse.BROWSER_CODE_COMPLETE:
-                final String url = data.getStringExtra(MSSTS_AUTHORIZATION_FINAL_URL);
-                result = parseUrlAndCreateAuthorizationResponse(url, data.getStringExtra(MicrosoftAuthorizationResult.REQUEST_STATE_PARAMETER));
+
```
iambmelt: I'm not seeing how this change differs from the previous behavior.... I can see that we check for a `null` URL inside of `parseUrlAndCreateAuthorizationResponse`; is this check necessary here?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -137,9 +138,10 @@ public void run() {
             return;
         }
 
-        if (getIntent().getDataString() != null) {
+        if (getIntent().getExtras() != null
```
iambmelt: Would the below check be insufficient here?
```java
if (!StringUtil.isEmpty(getIntent().getStringExtra(AuthorizationStrategy.CUSTOM_TAB_REDIRECT))) {
    ...
```

iambmelt:
Couple of questions

heidijinxujia:
Thank you, Brian. I will address them today.

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
heidijinxujia: No, this will call serializeNestedJsonToMap iteratively where the values are URL-decoded in Line 134.

heidijinxujia:
@iambmelt thanks a lot for the review. I've addressed the comments. Please have a look.

PR #218, opened 2018-09-05
Shoatman integrate

shoatman:
Updates to AuthorizationRequest and other items in support of integration with refactored MSAL.


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftTokenRequest.java
```diff
@@ -0,0 +1,34 @@
+package com.microsoft.identity.common.internal.providers.microsoft;
```
iambmelt: License

PR #220, opened 2018-09-05
Token endpoint authority and PPE support

titorodi:
Updating version to 0.0.4-RC
Adding Token Endpoint Authority configuration support for Sovereign Clouds and PPE
Adding support for null cloud from AAD discovery with authority validation disabled

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -16,7 +16,7 @@ android {
         minSdkVersion 14
         targetSdkVersion 27
         versionCode 1
-        versionName "0.0.3"
+        versionName "0.0.4-cp"
         project.archivesBaseName = "common"
         project.version = android.defaultConfig.versionName
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -16,7 +16,7 @@ android {
         minSdkVersion 14
         targetSdkVersion 27
         versionCode 1
-        versionName "0.0.4-cp"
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryOAuth2Configuration.java
```diff
@@ -42,5 +42,19 @@ public void setAuthorityHostValidationEnabled(boolean authorityHostValidationEna
...
+    }
+    
+    /**
+     * Sets an authority to use as a base for URIs as appropriate.
+     */
+    public void setAuthority(String authority) {
+        mAuthority = authority;
+    }
+    
+    private String mAuthority = "";
```
iambmelt: This change looks awfully similar to work @shoatman has done in this PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/218/files#diff-9e0be2be55bd29586c7ece6853cbb8d6

I think we should reconcile which of these implementations we go with, since there's clearly some value here

common/build.gradle
```diff
@@ -16,7 +16,7 @@ android {
         minSdkVersion 14
         targetSdkVersion 27
         versionCode 1
-        versionName "0.0.3"
+        versionName "0.0.4-cp"
```
shoatman: @addev-ashish - Is this something that we've done before?

PR #228, opened 2018-09-13
Use the issuer cache identifier as environment, rather than the value from the idtoken

iambmelt:


shoatman:
I'm not sure I'm bought into the iOS plan based.  The whole point of leveraging preferred_cache is to avoid islands of SSO.  Not using it seems incorrect to me, but perhaps i'm missing something.

iambmelt:
PR updated to support `@Nullable environment` see https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/358

common/src/main/java/com/microsoft/identity/common/internal/cache/ADALTokenCacheItem.java
```diff
@@ -83,8 +83,9 @@
     ADALTokenCacheItem(final AzureActiveDirectoryOAuth2Strategy strategy,
                        final AzureActiveDirectoryAuthorizationRequest request,
                        final AzureActiveDirectoryTokenResponse response) {
-        AzureActiveDirectoryAccount account = strategy.createAccount(response);
         String issuerCacheIdentifier = strategy.getIssuerCacheIdentifier(request);
+        AzureActiveDirectoryAccount account = strategy.createAccount(response);
```
shoatman: Future opportunity: To have createAccount actually deal with the issuer cache identifier.

PR #234, opened 2018-09-18
Adding support for ext_expires_on in isExpired() calculation

iambmelt:
See title. Adds tests.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
@@ -26,6 +26,7 @@ import com.google.gson.annotations.SerializedName;
 
 import java.util.Calendar;
 import java.util.Date;
+import java.util.concurrent.TimeUnit;
 
 import static com.microsoft.identity.common.internal.dto.AccessToken.SerializedNames.ACCESS_TOKEN_TYPE;
 import static com.microsoft.identity.common.internal.dto.AccessToken.SerializedNames.AUTHORITY;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
```diff
@@ -246,15 +246,27 @@ public void setExpiresOn(final String expiresOn) {
...
     }
+
+    @Override
+    public boolean isExpired() {
+        boolean isExpired = isExpired(getExpiresOn());
+
+        if (isExpired
+                && null != getExtendedExpiresOn()
+                && 0 < Long.valueOf(getExtendedExpiresOn())) {
+            isExpired = isExpired(getExtendedExpiresOn());
```
addev-ashish: This was in discussion recently, what if extended expires on is smaller than expired, lets check the max of 2 and set it.

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessToken.java
```diff
@@ -246,15 +245,26 @@ public void setExpiresOn(final String expiresOn) {
         mExpiresOn = expiresOn;
     }
 
-    @Override
-    public boolean isExpired() {
+    private boolean isExpired(final String expires) {
         // Init a Calendar for the current time/date
         final Calendar calendar = Calendar.getInstance();
-        calendar.add(Calendar.SECOND, AuthenticationSettings.INSTANCE.getExpirationBuffer());
```
iambmelt: @addev-ashish Removed per discussion in standup

PR #246, opened 2018-09-26
Update 0.0.5-RC2 -> 0.0.5-RC3

iambmelt:
Picking up fixes in #244, #245 -- will need to cut another RC


iambmelt:
@kreedula -- Just FYI, I'm merging this change but we are **_not_** ready to cut the next RC

PR #247, opened 2018-09-27
Fix the auth flow bug for Huawei browser and add webview fallback

heidijinxujia:
Fix the bug for Huawei Browser.
Add logger and fix the error code.
Add the null check for auth intent.

### DIFF
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:launchMode="singleTask"/>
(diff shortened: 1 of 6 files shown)


common/src/main/AndroidManifest.xml
```diff
@@ -13,7 +13,8 @@
         android:theme="@style/AppTheme">
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize" />
+            android:configChanges="orientation|keyboardHidden|screenSize"
+            android:launchMode="singleTask"/>
```
shoatman: Would be good to include a comment about using singleTask in the actual activity.  Feels like we should make a note of this for future maintainers.

heidijinxujia:
#248

PR #250, opened 2018-09-28
Adds support for MRRTs in common cache on v1/v2 endpoints

iambmelt:
Also updates the save() logic to delete RTs for the current account if there are some already in the cache.


iambmelt:
See corollary pr here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/387

PR #253, opened 2018-10-03
Fix incorrect constants to read error code and error description

kreedula:
Fixes : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/254


iambmelt:
Can you post a sanitized (free of pii/oii, or secrets) sample response here showing the format of these errors?

kreedula:
@iambmelt looks like in the specific case I am testing, error code and error message seems same. I'll  fix this . But in general the format is error code ; error message

Sample response of this format is

The redirectUri for broker is invalid ; The RedirectUri is not as expected. Received [url1] and expected [url]

PR #257, opened 2018-10-03
[B2C] Workaround for "realm" and "username" missing in IdToken

heidijinxujia:
• Preferred_username is mandatory filed. Use placeholder if it is missing in the id_token.

• Tid id required according to the spec. The fix is in the serve's test slice. Add the workaround with place holder string. Will remove it after the test slice becoming production.

• Fix the lib versioning for MSAL and ADAL.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
@@ -32,6 +32,7 @@ import com.microsoft.identity.common.internal.cache.SchemaUtil;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectoryIdToken;
 import com.microsoft.identity.common.internal.providers.oauth2.IDToken;
+import com.microsoft.identity.common.internal.util.StringUtil;
 
 import java.util.ArrayList;
 import java.util.Calendar;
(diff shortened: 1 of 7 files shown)


heidijinxujia:
Passed the E2E testing for interactive B2C test. But silent auth failed. Investigating it now.

iambmelt:
Reviewing this PR now...

heidijinxujia:
@iambmelt What's iOS's value? Could you share what you find? Thanks.

heidijinxujia:
Replace the username placeholders with the same string (based on the Schema spec)

heidijinxujia:
Need more discussion on the tid fix:
According to the spec, for Account, AccessToken, IdToken, the realm filed is required. The value of it would be full tenant or organizational identifier that account belongs to. It could also be an empty string for non-AAD scenarios. This introduced issue for the current realm validation.
@iambmelt and I are trying to find a solution...
@addev-ashish @shoatman Any thoughts?

iambmelt:
PS don't forget to update the MSAL/ADAL submodule pointers

We can do this task in the morning (10/4) since it's getting late in the eve -- we just can't ship without ADAL/MSAL aligned on the common SHA

/cc @kreedula

iambmelt:
> Need more discussion on the tid fix:
> According to the spec, for Account, AccessToken, IdToken, the realm filed is required. The value of it would be full tenant or organizational identifier that account belongs to. It could also be an empty string for non-AAD scenarios. This introduced issue for the current realm validation.
> @iambmelt and I are trying to find a solution...
> @addev-ashish @shoatman Any thoughts?

@addev-ashish @shoatman -- Chatted on Teams; going to comment-out the `realm` validation for now and deliver a better fix in `0.2.1+` or a later release. For now, we're going to write `""` to match the iOS implementation

iambmelt:
@heidijinxujia Can you make sure that #260 is not a regression from this change?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
```diff
@@ -90,7 +91,14 @@ public MicrosoftAccount(@NonNull final IDToken idToken,
...
         mFamilyName = claims.get(AzureActiveDirectoryIdToken.FAMILY_NAME);
         mMiddleName = claims.get(AzureActiveDirectoryIdToken.MIDDLE_NAME);
-        mTenantId = claims.get(AzureActiveDirectoryIdToken.TENANT_ID);
+        if (!StringUtil.isEmpty(claims.get(AzureActiveDirectoryIdToken.TENANT_ID))) {
+            mTenantId = claims.get(AzureActiveDirectoryIdToken.TENANT_ID);
+        } else {
+            // According to the spec, full tenant or organizational identifier that account belongs to.
+            // Can be an empty string for non-AAD scenarios.
+            Logger.warn(TAG, "TenantID is not returned from server.");
+            mTenantId = "";
```
iambmelt: I think we can do (for the `else` case)

```java
mTenantId = utid;
```

PR #280, opened 2018-10-13
Constants to accomodate LLT changes

kreedula:


kreedula:
Thank you @addev-ashish

PR #282, opened 2018-10-15
Shoatman multicloud

### LINKED ISSUE #281: Update performTokenRequest in oAuth2Strategy to use correlationId from diagnostic context

shoatman:
Add support for an application declaring itself aware of multiple clouds.  Doing so means that the application is aware that a user from a cloud other than world wide could be authorized.  Token returned will only be usable in the cloud from which they are issued.

Also closes #281


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -127,8 +128,7 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         );
         String requestBody = ObjectMapper.serializeObjectToFormUrlEncoded(request);
         Map<String, String> headers = new TreeMap<>();
-        String correlationId = UUID.randomUUID().toString();
-        headers.put("client-request-id", correlationId);
+        headers.put("client-request-id", DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
```
iambmelt: I stole this change to use in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/285 for a `0.0.7` hotfix

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -109,37 +109,9 @@ public static String serializeObjectToFormUrlEncoded(Object object) throws Unsup
...
-            } else {
-                final Map<String, String> hashMap = serializeNestedJsonToMap(entry.getValue().toString());
-                if (hashMap != null) {
-                    objectMap.putAll(hashMap);
-                }
-            }
-        }
-
-        return objectMap;
+        return new Gson().fromJson(json, Map.class);
```
iambmelt: 👍

Issue #291, opened 2018-10-18
[Webview] Fix the back button pressed issue


heidijinxujia:
In the flow with Webview auth, we should cancel the request when the back button is pressed.

PR #292, opened 2018-10-19
Add PendingIntent Support for authorization flow | Part 2/2

heidijinxujia:
1. Add PendingIntent property to AuthorizationStrategy (#294 )
2. Add PendingIntent support to handle the result in AuthorizationActivity. (#294 )
3. To reduce the confusion, rename the ChallengeCompletionCallback into AuthorizationCompletionCallback. (#293 )
4. Fix the webview back button pressed issue. (#291 )
5. Update the gradle version 
6. Update the manifest files.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -18,7 +20,7 @@
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.ui.AuthorizationAgent;
 import com.microsoft.identity.common.internal.ui.webview.AzureActiveDirectoryWebViewClient;
-import com.microsoft.identity.common.internal.ui.webview.challengehandlers.IChallengeCompletionCallback;
+import com.microsoft.identity.common.internal.ui.webview.challengehandlers.IAuthorizationCompletionCallback;
```
addev-ashish: its ok for now, but remember we are taking dependencies on these names in the broker

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -178,13 +227,31 @@ protected void onStop() {
...
         outState.putBoolean(KEY_AUTHORIZATION_STARTED, mAuthorizationStarted);
         outState.putBoolean(KEY_PKEYAUTH_STATUS, mPkeyAuthStatus);
         outState.putSerializable(KEY_AUTH_AUTHORIZATION_AGENT, mAuthorizationAgent);
         outState.putString(KEY_AUTH_REDIRECT_URI, mRedirectUri);
         outState.putString(KEY_AUTH_REQUEST_URL, mAuthorizationRequestUrl);
     }
 
+    @Override
+    public void onBackPressed() {
+        Logger.verbose(TAG, "Back button is pressed");
```
addev-ashish: what happens if they press back on the MFA screen?

heidijinxujia: @addev-ashish 
If the prompt behavior is SELECT_ACCOUNT or CONSENT, it will back to the select account page.
If the prompt behavior is FORCE_PROMPT, it will back to the login page.

PR #298, opened 2018-10-25
Broker client work - First PR

shoatman:
This has not been tested at all yet.  Getting the PR out such that the service side of the equation can potentially leverage what's been introduced for request, token and broker error responses.


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResult.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.identity.common.internal.broker;
```
iambmelt: Minor: This class will need license info

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -0,0 +1,125 @@
...
+    public MicrosoftAuthServiceFuture connect() {
+
+        MicrosoftAuthServiceFuture future = new MicrosoftAuthServiceFuture();
+        mMicrosoftAuthServiceConnection = new MicrosoftAuthServiceConnection(future);
+
+        final boolean serviceBound = mContext.bindService(mMicrosoftAuthServiceIntent, mMicrosoftAuthServiceConnection, Context.BIND_AUTO_CREATE);
+        Logger.verbose(TAG + "connect", "The status for MicrosoftAuthService bindService call is: " + Boolean.valueOf(serviceBound));
+
+        if (!serviceBound) {
+            disconnect();
```
iambmelt: Judging from [the docs](https://developer.android.com/reference/android/content/Context.html#bindService(android.content.Intent,%20android.content.ServiceConnection,%20int)), we may not need to take this step if `false == serviceBound`

>`true` if the system is in the process of bringing up a service that your client has permission to bind to; `false` if the system couldn't find the service or if your client doesn't have permission to bind to it. If this value is `true`, you should later call unbindService(ServiceConnection) to release the connection.`

shoatman: Good catch.  I'll update this in a future PR.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new RuntimeException("Unable to bind to service");
```
addev-ashish: based on our current struggles with boundService, crashing the app here might be problematic, we should look at throwing a client exception saying broker not reponding

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthServiceConnection.java
```diff
@@ -0,0 +1,70 @@
...
+        Logger.verbose(TAG, "MicrosoftAuthService is connected.");
+        mMicrosoftAuthService = IMicrosoftAuthService.Stub.asInterface(service);
+        mMicrosoftAuthServiceFuture.setMicrosoftAuthService(mMicrosoftAuthService);
+    }
+
+    /*
+
+    //Needed for API level 26 and above
+    @Override
+    public void onBindingDied(ComponentName name){
```
addev-ashish: this is interesting, can we add a log in these and run it with the Vivo device we have that has binding issues

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
shoatman: This is work in progress relative to exceptions.  The handler upstream can convert this to whatever i like.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthServiceConnection.java
shoatman: I'll take a look.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
shoatman: Fixed it now.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResult.java
shoatman: Fixed it

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
shoatman: I fixed this with clientException for now....

PR #303, opened 2018-11-05
Changes to IMicrosoftAuthService.aidl

kreedula:
- Added `BrokerRequest.aidl` and `BrokerResult.aidl` and made them implement parcelable
- Refactored BrokerTokenResponse and `BrokerErrorResponse`


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -927,6 +927,15 @@ private AuthenticationConstants() {
          */
         public static final String CALLER_INFO_PACKAGE = "caller.info.package";
 
+        /**
+         * String to send Msal V2 Request params.
+         */
+        public static final String BROKER_REQUEST_V2 = "broker.request.v2";
```
addev-ashish: Add a todo to say that we need to find a common place for this that does not say ADAL in the path

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResponse.java
```diff
@@ -0,0 +1,148 @@
...
+        setOAuthError(in.readString());
+        setOAuthSubError(in.readString());
+        setOAuthErrorMetadata(in.readString());
+        setOAuthErrorDescription(in.readString());
+        setHttpResponseBody(in.readString());
+        setHttpResponseHeaders(in.readString());
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
```
addev-ashish: should we do something with the flags to make sure we are future proof here?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
```diff
@@ -0,0 +1,90 @@
...
+        setClientInfo(in.readString());
+        setClientId(in.readString());
+        setExtExpiresIn(in.readLong());
+        setFamilyId(in.readString());
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
+        dest.writeLong(getExpiresIn());
+        dest.writeString(getAccessToken());
```
addev-ashish: what happens if one of these is null?

heidijinxujia: Parcel allows null to write. And it will return null when calling read.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResponse.java
```diff
@@ -0,0 +1,148 @@
...
+    private String mOAuthErrorDescription;//NOPMD
+    private String mHttpResponseBody;//NOPMD
+    private String mHttpResponseHeaders;//NOPMD
+
+    public BrokerErrorResponse(){
+
+    }
+
+    protected BrokerErrorResponse(Parcel in) {
+        setErrorCode(in.readInt());
```
heidijinxujia: Should we add `null` check here for `Parcel in`? Otherwise, it will throw `java.lang.NullPointerException`. Same for the other functions.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
```diff
@@ -0,0 +1,90 @@
...
+        public BrokerTokenResponse createFromParcel(Parcel in) {
+            return new BrokerTokenResponse(in);
+        }
+
+        @Override
+        public BrokerTokenResponse[] newArray(int size) {
+            return new BrokerTokenResponse[size];
+        }
+    };
+}
```
heidijinxujia: Please check out com.microsoft.identity.common.internal.broker.BrokerTokenResult#isIntunePolicyRequiredError

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResponse.java
kreedula: No flags is about how the object should be written , default value if good for us here. This is used to avoid leaks if FileDescriptors are used to write

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
kreedula: Thanks @heidijinxujia , I have added a To-do in ad-accounts, will sync with ashish about it and add as needed

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResponse.java
kreedula: Thanks @heidijinxujia . This should never be the case per documentation but I have added it to be on the safer side

Issue #300, opened 2018-10-30
CustomTabs not working with AndroidX


heidijinxujia:
After migrating project to AndroidX AppAuth-Android stop showing custom tabs and always open auth forms in standalone browser.

[Ref](https://github.com/openid/AppAuth-Android/issues/395)

heidijinxujia:
Need investigation and repro

PR #304, opened 2018-11-09
Dereferencing constant from CustomTabService, no reimplementing

iambmelt:
Replaces hard coded `android.support.customtabs.action.CustomTabsService` value with [`CustomTabsService.ACTION_CUSTOM_TABS_CONNECTION`](https://developer.android.com/reference/android/support/customtabs/CustomTabsService.html#ACTION_CUSTOM_TABS_CONNECTION)


iambmelt:
Related: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/300

PR #309, opened 2018-11-20
If items fail to deserialize, neither remove them nor return them.

### LINKED ISSUE #308: SharedPreferencesAccountCredentialCache should skip AccountRecords/Credentials it cannot deserialize
Currently, if [deserialization of an `AccountRecord`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java#L105-L116) or [`Credential`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java#L139-L153) fails, it is removed from the cache.

Instead of assuming that it is corrupt, the `SharedPreferencesAccountCredentialCache` should assume that the JSON payload may have been encrypted by another `SharedPreferencesFileManager` whose encryption keys are different.

iambmelt:
Closes #308 

@addev-ashish I'd like to review this change with you before we merge to make sure we're on the same page re expected behavior

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesAccountCredentialCacheTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesAccountCredentialCacheTest.java
@@ -28,8 +28,8 @@ import android.support.test.runner.AndroidJUnit4;
 
 import com.microsoft.identity.common.adal.internal.AndroidSecretKeyEnabledHelper;
 import com.microsoft.identity.common.adal.internal.cache.StorageHelper;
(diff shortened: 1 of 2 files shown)


iambmelt:
Need to update unit tests.

iambmelt:
Thanks @addev-ashish -- @heidijinxujia you may want to review this, even if only for FYI's sake

PR #310, opened 2018-11-22
Changes to add SDK Type to broker request and broker result

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -30,7 +30,12 @@
 /**
  * Represents the broker request
  */
-public class BrokerRequest implements Parcelable{
+public class BrokerRequest implements Parcelable {
+
+    public enum SDK_TYPE {
```
iambmelt: Minor nit:
Java's enums typically follow class naming conventions with uppercased _values_ -- for an example of this, see Oracle's docs on enums [here](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).

So this could be:
`SdkType.ADAL`, `SdkType.MSAL`

iambmelt: Come to think of it, what does this value ultimately get used for? Is there some property _about_ the upstream sdk we could switch on instead of on the sdk type itself?

For instance, if we're deciding which service to use to send the response, perhaps this value expresses service preference... Just a thought

kreedula: Thanks @iambmelt . Renamed as suggested.

Yes within the broker code, we would know based on the service which SDK invoked it, we use that to set a value for this SdkType in ADAL's case. This value is used to distinguish between some v1 and v2 request/result construction. Since `BrokerRequest` and `BrokerResult` are the entry and exit classes , it's useful to have this value in these classes

PR #311, opened 2018-11-27
Disambiguates AccountRecords lookups in guest scenarios

iambmelt:
The following method signatures have been updated. The methods which now accept a `realm` parameter do so such that they may disambiguate guest scenarios vs. not. The method which accepts a `local_account_id` is a convenience function to support broker lookups when only this value is present.

```java
/**
     * Returns the AccountRecord matching the supplied criteria.
     *
     * @param environment   The environment to which the sought AccountRecord is associated.
     * @param clientId      The clientId to which the sought AccountRecord is associated.
     * @param homeAccountId The homeAccountId of the sought AccountRecord.
     * @param realm         The tenant id of the targeted account (if applicable).
     * @return The sought AccountRecord or null if it cannot be found.
     */
    public abstract AccountRecord getAccount(final String environment,
                                             final String clientId,
                                             final String homeAccountId,
                                             final String realm
    );

    /**
     * Returns the AccountRecord matching the supplied criteria.
     *
     * @param environment    The environment to which the sought IAccount is associated.
     * @param clientId       The clientId to which the sought IAccount is associated.
     * @param localAccountId The local account id of the targeted account.
     * @return The sought AccountRecord or null if it cannot be found.
     */
    public abstract AccountRecord getAccountWithLocalAccountId(final String environment,
                                                               final String clientId,
                                                               final String localAccountId
    );

    /**
     * Removes the Account (and its associated Credentials) matching the supplied criteria.
     *
     * @param environment   The environment to which the targeted Account is associated.
     * @param clientId      The clientId of this current app.
     * @param homeAccountId The homeAccountId of the Account targeted for deletion.
     * @param realm         The tenant id of the targeted Account (if applicable).
     * @return True, if the Account was deleted. False otherwise.
     */
    public abstract boolean removeAccount(final String environment,
                                          final String clientId,
                                          final String homeAccountId,
                                          final String realm
    );
```

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
@@ -73,8 +73,6 @@ public class MsalOAuth2TokenCache
             GenericAccount,
             GenericRefreshToken> mAccountCredentialAdapter;
 
-    private final boolean mSecondaryMode;
-
     /**
      * Constructor of MsalOAuth2TokenCache.
      *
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -350,7 +426,7 @@ public AccountRecord getAccount(@Nullable final String environment,
                 mAccountCredentialCache.getCredentialsFilteredBy(
                         null, // homeAccountId
                         environment,
-                        CredentialType.RefreshToken,
+                        CredentialType.IdToken,
```
iambmelt: Note to self: need to get this change blessed.

Since broker may be hosting AT/RT pairs for the current Account, we can no longer define the presence of an Account based on whether or not we have an RT for it. Instead, we will say we have an Account if we have an IdToken for it which matches the current apps' client Id.

iambmelt: @shoatman @kreedula @heidijinxujia Let's plan to make a follow-up item out of this in tomorrow's scrum

PR #314, opened 2018-11-30
Required changes to common to remove ADAL depenency from ad-accounts

kreedula:


common/src/main/java/com/microsoft/identity/common/adal/internal/ADALError.java
```diff
@@ -0,0 +1,729 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package com.microsoft.identity.common.adal.internal;
```
iambmelt: Are there going to be two versions of this file? One in ADAL and another in common?

kreedula: Yes since we don't have any ADAL release planned. I think we should add a task to Adal Backlog for the next release to remove it. @shoatman Let us know what you think?

iambmelt: JFYI, `ADALError` is `public` surface for us, so any relocation of it needs to be framed as a major-break from a versioning perspective.

kreedula: Yes that's true, May be we can do what we did with `AuthenticationConstants` , point all the constants to the values to common's `ADALError`

kreedula:
Had an offline discussion with Shane. Created an issue to change here : https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1370

PR #318, opened 2018-12-07
Optionally remove all Accounts/Credentials associated to a home_account_id, adds broker token cache sequestration

iambmelt:
The easiest summary of these changes can be seen in the javadoc comments of `MsalOauth2TokenCache#removeAccount`

```java
    /**
     * Removes the specified Account or Accounts from the cache.
     * <p>
     * Note: if realm is passed as null, all tokens and AccountRecords associated to the
     * provided homeAccountId will be deleted. If a realm is provided, then the deletion is
     * restricted to only those AccountRecords and Credentials in that realm (tenant).
     *
     * @param environment   The environment to which the targeted Account is associated.
     * @param clientId      The clientId of this current app.
     * @param homeAccountId The homeAccountId of the Account targeted for deletion.
     * @param realm         The tenant id of the targeted Account (if applicable).
     * @return
     */
    @Override
    public AccountRecordDeletionResult removeAccount(final String environment,
                                                     final String clientId,
                                                     final String homeAccountId,
                                                     @Nullable final String realm) {
        // ...
    }
```

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/AccountRecordDeletionResult.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/AccountRecordDeletionResult.java
@@ -1,0 +1,175 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/AccountRecordDeletionResult.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+
+/**
+ * Functions as a 'receipt' when deleting AccountRecords: AccountRecords which have been removed
+ * from the cache are returned in this container.
+ */
+public class AccountRecordDeletionResult implements List<AccountRecord> {
```
iambmelt: Note: This gets a dedicated `class` and not just a `List` to be really specific about the result purpose.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -524,36 +538,69 @@ public boolean removeAccount(final String environment,
                         homeAccountId,
                         realm
                 ))) {
-            return false;
+            return new AccountRecordDeletionResult(null);
         }
 
+        // If no realm is provided, remove the Account/Credetials from all realms.
+        final boolean isRealmAgnostic = null == realm;
```
shoatman: this may be a little too cute.  I'd consider using parentheses around (null==realm) to improve readability.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2TokenCache.java
```diff
@@ -139,12 +140,12 @@ public abstract AccountRecord getAccountWithLocalAccountId(final String environm
...
      * @param homeAccountId The homeAccountId of the Account targeted for deletion.
      * @param realm         The tenant id of the targeted Account (if applicable).
-     * @return True, if the Account was deleted. False otherwise.
+     * @return The {@link AccountRecordDeletionResult} containing the removed AccountRecords.
      */
-    public abstract boolean removeAccount(final String environment,
-                                          final String clientId,
-                                          final String homeAccountId,
-                                          final String realm
+    public abstract AccountRecordDeletionResult removeAccount(final String environment,
```
shoatman: I'd consider AccountDeletionRecord instead of Result... since I don't believe that's it's possible for the deletion of a specific account to fail.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Sounds good

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2TokenCache.java
iambmelt: 👍

PR #320, opened 2018-12-10
Refactor/Move classes  to common from Msal to call acquireToken and acquireTokenSilent from common directly

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -0,0 +1,94 @@
...
+package com.microsoft.identity.common.internal.authorities;
+
+import com.google.gson.annotations.SerializedName;
+import com.microsoft.identity.common.internal.logging.Logger;
+
+public abstract class AzureActiveDirectoryAudience {
+
+    private static final String TAG = AzureActiveDirectoryAudience.class.getSimpleName();
+
+    private String cloudUrl;
```
heidijinxujia: mCloudUrl

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -0,0 +1,94 @@
...
+    @SerializedName("tenant_id")
+    private String tenantId;
```
heidijinxujia: mTenantId

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryB2CAuthority.java
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            return new URL(this.getAuthorityUri().toString());
+        } catch (MalformedURLException e) {
+            throw new IllegalArgumentException("Authority URL is not a URL.", e);
+        }
+    }
+
+    @Override
+    public OAuth2Strategy createOAuth2Strategy() {
+        MicrosoftStsOAuth2Configuration config = new MicrosoftStsOAuth2Configuration();
```
heidijinxujia: Could you add a Logger here?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -0,0 +1,346 @@
...
+                        }
+                    }
+                }
+            });
+        }
+    }
+
+    private static void logInteractiveRequestParameters(final String methodName,
+                                                        final AcquireTokenOperationParameters params) {
+        com.microsoft.identity.common.internal.logging.Logger.verbose(
```
heidijinxujia: Logger.verbose would be fine.

common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
```diff
@@ -0,0 +1,158 @@
...
+                if(result.getSucceeded()){
+                    com.microsoft.identity.common.internal.logging.Logger.verbose(
+                            TAG + methodName,
+                            "Executing with controller: " + controller.getClass().getSimpleName() + ": Succeeded"
+                    );
+                    return result;
+                }
+            }catch(UiRequiredException e){
+                if(e.getErrorCode() == UiRequiredException.INVALID_GRANT){
+                    continue;
```
heidijinxujia: Could you clarify the reason why we continue the acquireTokenSilent when getting INVALID_GRANT here?

kreedula: Yes, since we are iterating over a list of controllers, we will continue and try silent call with next controller in this case. I just saw there was a bug here with `equals` vs `==` for String here. Fixed it.

kreedula:
Thanks @heidijinxujia . Addressed your comments, have a look when you get a chance

common/src/main/java/com/microsoft/identity/common/internal/result/IBaseAuthenticationResult.java
```diff
@@ -0,0 +1,76 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import java.util.Date;
+
+/**
+ * Interface to wrap successful authentication result. When auth succeeds, token will be wrapped into the
+ * {@link IBaseAuthenticationResult}
+ */
+public interface IBaseAuthenticationResult {
```
iambmelt: The more I look at this interface, the more it scares me....

It seems like it would be all too easy to look at this file and see:
1. That it is in `common`
2. That is is namespaced to `internal`

And then [incorrectly] assume that we can make changes to this file willy-nilly: that wouldn't be the case, however since changes made here would be reflected in public API.

I'm mulling-over the thought that this interface and `IAuthenticationResult` should be separate entirely....

iambmelt: Let's take a backlog item to separate these interfaces and provide an adapter between them. Yes, it's extra paperwork, but I really think we're going to want this separation going forward.

Not going to 'flunk' on this, but we shouldn't deploy until its resolved in more future-proof way.

common/src/main/java/com/microsoft/identity/common/internal/request/IAuthenticationCallback.java
```diff
@@ -0,0 +1,58 @@
...
+
+package com.microsoft.identity.common.internal.request;
+
+import com.microsoft.identity.common.exception.BaseException;
+import com.microsoft.identity.common.internal.result.IBaseAuthenticationResult;
+
+/**
+ * Callback passed with token acquisition. {@link IBaseAuthenticationResult} or {@link Exception} will be returned back via callback.
+ */
+public interface IAuthenticationCallback {
```
iambmelt: Same as how the the internal `IAuthenticationCallback` is separate from `AuthenticationCallback`, the `IBaseAuthenticationResult` and `IAuthenticationResult` should be separate.

iambmelt: The name of this class seems like a possible source of confusion.... since we have `IAuthenticationCallback` and `AuthenticationCallback`.

The public API should probably not change at this point so maybe something in this class' name to signal that it's internal?

iambmelt: Issue opened here: https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/478

common/src/main/java/com/microsoft/identity/common/internal/result/IBaseAuthenticationResult.java
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/478

kreedula: I agree with the concerns raised w.r.t to the design here. I have taken the easier way out yesterday after incorporating to changes to get it working. I have made the changes via https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/320/commits/8b577226de00ecb30f01a110c1e7386a064dc870 to use an adapter instead.

kreedula:
Thanks @iambmelt for the review and feedback, made changes as appropriate. Also addressed the issue : https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/478

iambmelt:
Thanks @kreedula -- starting second pass review

iambmelt:
LGTM, thanks @kreedula!

PR #323, opened 2018-12-14
Fix for resource reuse | Fix #322 

### LINKED ISSUE #322: Shared resource is broken
MSAL and ad-accounts cannot get the shared resource from common.

heidijinxujia:


kreedula:
I'm taking the liberty to merge it as I can use it to test my changes

PR #324, opened 2018-12-20
Added UserInfo class to common

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -74,6 +81,53 @@ protected void throwIfNetworkNotAvailable(final Context context) throws ClientEx
         );
     }
 
+
+    protected AuthorizationRequest getAuthorizationRequest(final OAuth2Strategy strategy,
```
heidijinxujia: Both input parameters should be "notnull", right? We need to have a null check here.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
```diff
@@ -134,9 +134,9 @@ private String getUniqueId(final Map<String, String> claims) {
 
         String uniqueId = null;
 
-        if (!StringExtensions.isNullOrBlank(claims.get(AzureActiveDirectoryIdToken.OJBECT_ID))) {
+        if (!StringExtensions.isNullOrBlank(claims.get(AzureActiveDirectoryIdToken.OBJECT_ID))) {
```
heidijinxujia: 👍

kreedula:
Thanks @heidijinxujia and @shoatman

PR #328, opened 2019-01-03
Fix the onResume for webview flow

heidijinxujia:
Fix for customer report bug.
https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/484


iambmelt:
@heidijinxujia Rather than put all of this protective logic into `onResume()` why not move this logic up into `onCreate()` or `onStart()` -- this is what you're really trying to detect, isn't it?

heidijinxujia:
@iambmelt It's a good question.

I did a test on moving the loading the start url into onCreate(). 
As a result, 
1. for the webview-based flow, we could move the loading start url into the onCreate().
2. for the browser-based flow, we cannot start browser activity inside OnCreate(). Because the life cycle of the current activity will continue. They are async. And onResume of the current activity will be called and got canceled before finishing the login in browser activity.

Previously, I put all the logic into the same onResume function, because launching the browser/customTabs activity has to be inside onResume().  While putting the loading url of webview inside onCreate is do-able.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -119,6 +119,22 @@ protected void onCreate(final Bundle savedInstanceState) {
             // If activity is killed by the os, savedInstance will be the saved bundle.
             extractState(savedInstanceState);
         }
+
+        if (mAuthorizationAgent == AuthorizationAgent.WEBVIEW) {
+            mAuthorizationStarted = true;
```
iambmelt: I don't see this value ever get set to `false` unless `extractState()` is called and that happens _before_ the value is set to `true`

iambmelt: Or does it only get set to true if we're using the WebView? In that case, should the variable be renamed to communicate that specifically?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -134,57 +150,43 @@ protected void onNewIntent(Intent intent) {
...
-                    startActivity(mAuthIntent);
-                } else {
-                    final Intent resultIntent = new Intent();
-                    resultIntent.putExtra(AuthenticationConstants.Browser.RESPONSE_AUTHENTICATION_EXCEPTION, new ClientException(ErrorStrings.AUTHORIZATION_INTENT_IS_NULL));
-                    sendResult(AuthorizationStrategy.UIResponse.BROWSER_CODE_AUTHENTICATION_EXCEPTION, resultIntent);
-                    finish();
-                }
-            }
-
+        if (mAuthorizationAgent == AuthorizationAgent.WEBVIEW) {
```
iambmelt: Can we remove this side of the conditional and instead gate this logic on...
```java
if (mAuthorizationAgent == AuthorizationAgent.DEFAULT || mAuthorizationAgent.BROWSER) {
    // ...
}
```

 I think this code is easier to understand out-of-context if the logic happens based on what "_is_" rather than "_what isn't_"

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
heidijinxujia: Addressed. Thank you, Brian.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
heidijinxujia: Addressed. Thanks.

iambmelt:
@heidijinxujia Updates look good. Looks like this branch is out of date - once you pull in dev latest I think we can :shipit:

Issue #326, opened 2018-12-28
java.lang.ClassCastException: AccountRecord cannot be cast to MicrosoftStsAccount


heidijinxujia:
```java
2018-12-28 13:59:14.316 10030-10820/com.microsoft.identity.client.sample.local E/ApiDispatcher:beginInteractive:  [2018-12-28 21:59:14 - {"thread_id":"5283","correlation_id":"32d17199-a073-4452-8e1e-26bfe8d14769"}] Interactive request failed with Exception Android 24
    java.lang.ClassCastException: com.microsoft.identity.common.internal.dto.AccountRecord cannot be cast to com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsAccount
        at com.microsoft.identity.common.internal.result.MicrosoftStsAuthenticationResult.<init>(MicrosoftStsAuthenticationResult.java:44)
        at com.microsoft.identity.client.internal.controllers.LocalMSALController.acquireToken(LocalMSALController.java:118)
        at com.microsoft.identity.common.internal.controllers.InteractiveTokenCommand.execute(InteractiveTokenCommand.java:64)
        at com.microsoft.identity.common.internal.controllers.ApiDispatcher$1.run(ApiDispatcher.java:73)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:762)
```

heidijinxujia:
@kreedula  Could you have a look at this issue? This seems to be caused by [commit ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/c344d201eccc8edf1df09b63153f59f5d02d4ba9)

PR #329, opened 2019-01-04
Revert MicrosoftStsResult changes and changes to use ADALUserInfo for WPJ

kreedula:
- Reverted commit : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/6ecc51f44848299cafcff71a9c1a27240e9fe0b6
- Fixes : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/326
- Moved ADALUserInfo to adal.internal package
- Added couple of util constructors to ADALUserInfo

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SchemaUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SchemaUtil.java
@@ -150,7 +150,7 @@ public final class SchemaUtil {
                 final Map<String, String> idTokenClaims = idToken.getTokenClaims();
 
                 if (null != idTokenClaims) {
-                    idp = idTokenClaims.get("idp");
+                    idp = idTokenClaims.get(MicrosoftIdToken.ISSUER);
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SchemaUtil.java
```diff
@@ -137,6 +138,38 @@ public static String getAlternativeAccountId(final IDToken idToken) {
...
+        String idp = null;
+
+        if (null != idTokenString) {
+            IDToken idToken;
+            try {
+                idToken = new IDToken(idTokenString);
+                final Map<String, String> idTokenClaims = idToken.getTokenClaims();
+
+                if (null != idTokenClaims) {
+                    idp = idTokenClaims.get("idp");
```
iambmelt: There should be constant somewhere for this, probably inside of `AuthenticationConstants` or one of the [many] `IdToken` classes

PR #331, opened 2019-01-08
Moved IBrokerAccountService.aidl to com.microsoft.aad.adal

kreedula:
- Both client and server AIDL files need to be in the package, so moved it to be backwards compat with ADAL


iambmelt:
@kreedula It _truly_ makes sense to have the `IBrokerAccountService.aidl` in this namespace inside of common, right? For future reference, can you briefly summarize why?

PR #333, opened 2019-01-13
Reverting fix of auth flow bug for Huawei browser

scscgit:
Using singleTask launchMode causes other compatibility issues observed on low Android API,
specifically tested on Sony Xperia SP using API 18. The consensus is that startActivityForResult()
mustn't be used if an activity is being launched as a singleInstance or singleTask - StackOverflow:
https://stackoverflow.com/questions/7910840/android-startactivityforresult-immediately-triggering-onactivityresult
The developer can always override this in app manifest for example using:
\<activity android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
android:launchMode="standard" tools:replace="android:launchMode" />

### DIFF
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -7,9 +7,8 @@
     <application>
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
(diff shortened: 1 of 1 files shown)


iambmelt:
Closing as this no longer applies to the v1.0.x library

PR #339, opened 2019-01-22
Add additional build config entries

shoatman:


shoatman:
@iambmelt - Given that there are no changes in this PR to .gitignore. this is clearly an existing issue.  I'm not cleaning up everything related to the build.  I would suggest filing an issue and we can revisit this later.  See no reason to block this PR.

shoatman:
Thanks Brian!

PR #341, opened 2019-01-24
Second set of ad-accounts refactor changes

kreedula:
- Added additional constants and error strings as needed
- Added a new constructor to `BrokerTokenResponse` and unused remove primary refresh token param
- Removed redundant `IllegalArgumentException` from AcquireTokenParameters
- Added missing ResponseCodes to `MicrosoftStsAuthorizationFactory`

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
@@ -37,7 +37,6 @@ public class BrokerTokenResponse extends MicrosoftStsTokenResponse implements Pa
 
     private String mAuthority;
     private String mTenantId;
-    private String mPrimaryRefreshToken;
 
     @SerializedName("not_before")
     private Date mExpiresNotBefore;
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
```diff
@@ -37,16 +37,27 @@
...
 
     @SerializedName("not_before")
     private Date mExpiresNotBefore;
 
     @SerializedName("cloud_instance_host_name")
     private String mCloudInstanceHostName;
 
-    public BrokerTokenResponse(){
-
+    public BrokerTokenResponse(MicrosoftStsTokenResponse copy){
```
shoatman: I'd suggest an alternate parameter name.  tokenResponse?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerTokenResponse.java
```diff
@@ -75,7 +85,7 @@ protected BrokerTokenResponse(Parcel in) {
     @Override
     public void writeToParcel(Parcel dest, int flags) {
         if (dest != null) {
-            dest.writeLong(getExpiresIn() != null ? getExpiresIn() : 0);
+            dest.writeLong(getExpiresIn());
```
shoatman: I'm assuming that getExpiresIn has been refactored to no longer return null?

kreedula: Thanks shane ,, this is an unintentional change, reverting it back

kreedula:
Thanks @shoatman for the review. Addressed your comments

PR #336, opened 2019-01-16
First set of refactor changes relevant to ad-accounts

kreedula:
- Added new params to `BrokerRequest`
- Fixed UTF_8 typo to UTF-8
- Added  util methods as appropriate
- Added new fields to `ServiceException` and `UIRequiredException`

PR #347, opened 2019-01-28
Merge refactor to dev

kreedula:
Previous PR's https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/336 , https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/341

Current changes
- Added refresh token to ILocalAuthenticationResult
- Added common methods in sub class controllers to `BaseController`
- Minor formatting and code clean up


common/src/main/java/com/microsoft/identity/common/adal/internal/util/HashMapExtensions.java
```diff
@@ -133,6 +134,26 @@ private HashMapExtensions() {
...
+    /**
+     * get key value pairs from response.
+     *
+     * @param webResponse {@link HttpResponse} to convert to a map
+     * @return HashMap
+     * @throws JSONException
+     */
+    public static HashMap<String, String> getJsonResponse(HttpResponse webResponse) throws JSONException {
+        final HashMap<String, String> response = new HashMap<>();
+        if (webResponse != null && !TextUtils.isEmpty(webResponse.getBody())) {
```
iambmelt: This looks like a good opportunity for code sharing with the _other_ `getJsonResponse` method inside of this class - looks like both methods could accept a `String` parameter.

[Opened an issue here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/349), since we can clean this up later.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -146,6 +158,102 @@ protected TokenResult performTokenRequest(final OAuth2Strategy strategy,
...
+            );
+
+            // Create a new AuthenticationResult to hold the saved record
+            final LocalAuthenticationResult authenticationResult = new LocalAuthenticationResult(savedRecord);
+
+            // Set the AuthenticationResult on the final result object
+            acquireTokenSilentResult.setLocalAuthenticationResult(authenticationResult);
+        } else {
+            // Log all the particulars...
+            if (null != tokenResult.getErrorResponse()) {
```
iambmelt: Should `L#193` be an `else if(null != tokenResult.getErrorResponse())`?

What code path would be go down if the server returned an empty response body? How would we detect this?

What if there were an error response that was **not** `invalid_grant`? Would we see it? Would we log it?

PR #348, opened 2019-01-28
Update the version number and changelog

heidijinxujia:


kreedula:
@heidijinxujia we can name the release branch as `release/0.08` instead of `release/0.0.8-rc` as RC flag will only be temporary until testing

iambmelt:
> @heidijinxujia we can name the release branch as `release/0.0.8` instead of `release/0.0.8-rc` as RC flag will only be temporary until testing

Fixt :-)

PR #352, opened 2019-01-30
App/FOCI/UID Metadata Caching

iambmelt:
Implements and integrates a FOCI, application, and uid metadata caching for `BrokerOAuth2TokenCache`

**Usage with cache:**
When ADAL is calling broker:
```java
        final BrokerOAuth2TokenCache brokerOAuth2TokenCache2 = new BrokerOAuth2TokenCache(
                context,
                TEST_APP_UID,
                new SharedPreferencesBrokerApplicationMetadataCache(context),
                true // allow vis into other apps' caches
        );
```

When MSAL is calling broker:
```java
        final BrokerOAuth2TokenCache brokerOAuth2TokenCache2 = new BrokerOAuth2TokenCache(
                context,
                TEST_APP_UID,
                new SharedPreferencesBrokerApplicationMetadataCache(context),
                false // disallow vis into other apps' caches
        );
```

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
@@ -94,28 +94,50 @@ public class BrokerOAuth2TokenCache
 
     private static final String UNCHECKED = "unchecked";
 
+    private final IBrokerApplicationMetadataCache mApplicationMetadataCache;
     private final MicrosoftFamilyOAuth2TokenCache mFociCache;
     private MsalOAuth2TokenCache mAppUidCache;
     private List<MsalOAuth2TokenCache> mOptionalCaches;
+    private final int mCallingAppUid;
 
     /**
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -166,11 +194,49 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
...
                 oAuth2Strategy,
                 request,
                 response
         );
+
+        updateApplicationMetadataCache(
+                result.getIdToken().getClientId(),
+                result.getIdToken().getEnvironment(),
+                result.getRefreshToken().getFamilyId(),
+                mCallingAppUid
```
shoatman: Let's add the word "Process" to ensure that future developer understand that this is the process identifier as provided by the OS.

iambmelt:
TODO: If two apps share the same `clientId` but have different `callingProcessUids`, make sure SSO state can be picked up...

PR #354, opened 2019-01-30
Used single instance to monitor Service connection and used Applicati…

MichalDanielDobrzanski:
Used single instance to monitor Service connection and used Application Context instead of Activity Context as for some devices and scenarios Activity can be destroyed, and when returning, a new one is created, hence there is IllegalArgumentException: Service not registered as new Activity with its new Context has come.
Basically, Services have different lifecycle than Activites. Activities' instances can be garbage collected, and therefore we weren't able to unbind from service. 

Reference:
- https://stackoverflow.com/questions/8614335/android-how-to-safely-unbind-a-service
- https://stackoverflow.com/questions/22079909/android-java-lang-illegalargumentexception-service-not-registered

To test that it occurs without these changes:
1. use old phone <=6.0 or set 'Dont keep activities' flag
2. open MSAL Sample App or Testing app
3. navigate to WebView
4. press 'X'
Repeat 3. and 4. continuously. For me it happens on second try.


MichalDanielDobrzanski:
@shoatman  @heidijinxujia  @kreedula  @iambmelt

annakocheshkova:
@iambmelt @heidijinxujia can you take a look at this? It's affecting us, too. Thanks!

iambmelt:
Merged. Thanks for your contribution, @MichalDanielDobrzanski!

MichalDanielDobrzanski:
Thank you! :)

PR #358, opened 2019-02-01
Fix the fragment parameter when extracting the redirect url

heidijinxujia:
During the interactive auth code flow, MSAL extracts the auth code from the redirect URI's query parameters. And the fragment parameter is expected to be null. However, the B2C bug we encountered is, Facebook returns both query parameter and fragment parameter “_=_”.

For AAD V1(Section: OAuth 2.0 auth code grant) and V2 (Section: request an authorization code), the query is set as default for “response_mode”. And in MSAL, at least in Android platform, we actually did not have the property to set the “response_mode” value of the auth request. That is we always use the default value of “response_mode” set by the server. 

And the B2C service cannot delete the fragment para sent by Facebook. To fix this issue, we can simply skip the fragment para when extracting the auth response from redirect_uri

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/MicrosoftStsAuthorizationResultFactoryTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/MicrosoftStsAuthorizationResultFactoryTest.java
@@ -52,6 +52,7 @@ public class MicrosoftStsAuthorizationResultFactoryTest {
 
     private static final String REDIRECT_URI = "msauth-clientid://packagename/";
     private static final String AUTH_CODE_AND_STATE = "code=authorization_code&state=state";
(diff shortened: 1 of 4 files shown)


heidijinxujia:
@iambmelt Addressed. Unit tests added.

common/src/androidTest/java/com/microsoft/identity/common/MicrosoftStsAuthorizationResultFactoryTest.java
```diff
@@ -196,4 +197,33 @@ public void testUrlWithErrorInParams() {
         assertEquals(errorResponse.getErrorDescription(), ERROR_DESCRIPTION);
     }
 
+    @Test
```
iambmelt: 💪

PR #360, opened 2019-02-04
Changes corresponding to ad-accounts PR

kreedula:
PR : https://github.com/AzureAD/ad-accounts-for-android/pull/803


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerErrorResponse.java
```diff
@@ -52,6 +52,16 @@ public BrokerErrorResponse(){
 
     }
 
+    public BrokerErrorResponse(MicrosoftTokenErrorResponse microsoftTokenErrorResponse){
```
iambmelt: Minor nit, param could be `final` -- looks like `SchemaUtil` could use to have a formatter run over it too. Rest LGTM

PR #364, opened 2019-02-05
Merge Master back to dev

kreedula:


kreedula:
@heidijinxujia , I think the conflict is due to some of your changes have already being merged to dev. Can you take a look?

PR #369, opened 2019-02-07
Implements and tests the new BrokerOAuth2TokenCache 3-arg save() API

iambmelt:
This is a feature request on the `BrokerOAuth2TokenCache`: allow it to save WPJ'd Accounts, IdTokens, AccessTokens (but RefreshTokens will not be present).

API Design:
The BrokerOAuth2TokenCache will get a new API with approx. the following functionality

```java
public ICacheRecord save(@NonNull AccountRecord account,
                         @NonNull IdTokenRecord idToken,
                         @NonNull AccessTokenRecord accessToken,
                         @Nullable String familyId) {
    if (null != familyId) {
        // Save the passed-in Account/Credentials to the FOCI cache
    } else {
        // Save the passed-in Account/Credentials to the processUid cache
    }

    // Update the BrokerApplicationMetadataCache with the new record
}
```

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
@@ -94,6 +94,53 @@ public class MsalOAuth2TokenCache
         mAccountCredentialAdapter = accountCredentialAdapter;
     }
 
+    /**
+     * @param accountRecord     The {@link AccountRecord} to store.
+     * @param idTokenRecord     The {@link IdTokenRecord} to store.
+     * @param accessTokenRecord The {@link AccessTokenRecord} to store.
+     * @return The {@link ICacheRecord} result of this save action.
+     * @throws ClientException If the supplied Accounts or Credentials are schema invalid.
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -94,6 +94,53 @@ public MsalOAuth2TokenCache(final Context context,
...
+            throw new ClientException(ACCOUNT_IS_SCHEMA_NONCOMPLIANT);
+        }
+
+        if (!isIdTokenValid) {
+            throw new ClientException(CREDENTIAL_IS_SCHEMA_NONCOMPLIANT, "[(ID)]");
+        }
+
+        if (!isAccessTokenValid) {
+            throw new ClientException(CREDENTIAL_IS_SCHEMA_NONCOMPLIANT, "[(AT)]");
+
```
iambmelt: Remove extra newline

iambmelt: Resolved in 53f0e850f09e3a3af0d15082a7a30634c1fc7142

PR #370, opened 2019-02-08
Adds support for SPE Ring / x-ms-clitelem parsing for /token requests

iambmelt:
Work in progress PR.

This PR contains formatting fixes, import scrubbing.

Implements support for SPE Ring/x-ms-clitelem header parsing. The result is set to the `TokenResult` and later packaged up for broker consumers to parse out.

PR #371, opened 2019-02-11
Remove INVALID_GRANT omission in TokenCommand.execute()

rapong:
When INVALID_GRANT was received, it was silently consumed hence there is no exception thrown, and the null result object is returned.

This crashes ad-accounts as it's expecting result object to not be null.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
@@ -81,25 +81,17 @@ public class TokenCommand implements TokenOperation {
         final String methodName = ":execute";
 
         for(BaseController controller : mControllers) {
-            try {
+            com.microsoft.identity.common.internal.logging.Logger.verbose(
+                    TAG + methodName,
+                    "Executing with controller: " + controller.getClass().getSimpleName()
+            );
+            result = controller.acquireTokenSilent((AcquireTokenSilentOperationParameters) getParameters());
+            if(result.getSucceeded()){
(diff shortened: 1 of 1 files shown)


shoatman:
@rapong - Do you understand why the code was there to swallow this in the first place?

rapong:
> @rapong - Do you understand why the code was there to swallow this in the first place?

No. To me, it doesn't make sense to swallow this particular error here. It's crashing the flow I'm testing, hence I removed it.

@kreedula - Is there any particular flow I should be aware of?

iambmelt:
I have an issue tracking this here: https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1394

With a different proposed fix (only swallowing the `Exception` when this is _not_ the last controller to try...

See:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/8b5f61c7412b27c24d0d1246dcb0ab877f6f30c5

rapong:
> I have an issue tracking this here: [AzureAD/azure-activedirectory-library-for-android#1394](https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1394)
> 
> With a different proposed fix: only swallowing the `Exception` when this is _not_ the last controller to try...
> 
> See:
> [8b5f61c](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/8b5f61c7412b27c24d0d1246dcb0ab877f6f30c5)

That fix preserves the current behavior better, so I'd suggest using that one instead. Although we might still want to figure out why this particular exception was consumed here to begin with.

iambmelt:
Thanks @rapong -- my fix has merged with #370. We should now be able to close this.

PR #372, opened 2019-02-12
 Changes to accommodate Joined Account Caching changes in ad-accounts

kreedula:
- Also fixed issue where default scopes could be added multiple times.
- Added Additional Extra query params related to Broker to `MicrosoftTokenRequest`
- (Temporary hack) TBD : added to parse V1 idtoken `upn` in `MicrosoftStsRequest` to store the right value of displayable id in cache

Related ad-accounts PR : https://github.com/AzureAD/ad-accounts-for-android/pull/811


kreedula:
Thanks @shoatman

PR #378, opened 2019-02-15
Add sovereign cloud support for V2 broker

heidijinxujia:
1. Fixed the cloud authority for the sovereign token request
2. Fixed the broken back compatibility on BrokerAuthenticationResult.authority between V1 broker and V2 broker
3. Add the extraqp filtering and mapping
4. Add cloud host name authority into broker token result.
5. Fix the unit tests

TODO.

2. Add unit tests

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
@@ -93,7 +93,11 @@ public class MicrosoftStsAccountCredentialAdapter
 
             // Optional fields
             accessToken.setExtendedExpiresOn(getExtendedExpiresOn(response));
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -76,7 +80,11 @@ public AccessTokenRecord createAccessToken(
             accessToken.setCredentialType(CredentialType.AccessToken.name());
             accessToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
             accessToken.setRealm(getRealm(strategy, response));
-            accessToken.setEnvironment(strategy.getIssuerCacheIdentifier(request));
+            if (!StringUtil.isEmpty(response.getAuthority())) {
```
iambmelt: Is there a case where the authority is expected to be null on the response? Or just being defensive here?

PR #381, opened 2019-02-15
Shoatman logging improvements

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/IErrorResponse.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.identity.common.internal.providers.oauth2;
```
iambmelt: We should include the customary license info boilerplate at the top of this file and `IResult`

PR #382, opened 2019-02-19
Introduce subError in TokenErrorResponse.

rapong:
- Introduce subError field.
- Introduce bad_token subError.
- remove redundant fields from BrokerErrorResponse.
- replace redundant constant and make them use the one in AuthenticationConstants instead.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
@@ -44,7 +44,7 @@ public final class Device {
         final Map<String, String> platformParameters = new HashMap<>();
 
         platformParameters.put(PlatformIdParameters.PRODUCT, PlatformIdParameters.PRODUCT_NAME);
-        platformParameters.put(PlatformIdParameters.VERSION, "0.1.3");
+        platformParameters.put(PlatformIdParameters.VERSION, PlatformIdParameters.PRODUCT_VERSION);
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
```diff
@@ -44,7 +44,7 @@ private Device() {
         final Map<String, String> platformParameters = new HashMap<>();
 
         platformParameters.put(PlatformIdParameters.PRODUCT, PlatformIdParameters.PRODUCT_NAME);
-        platformParameters.put(PlatformIdParameters.VERSION, "0.1.3");
+        platformParameters.put(PlatformIdParameters.VERSION, PlatformIdParameters.PRODUCT_VERSION);
```
iambmelt: 👍

common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
```diff
@@ -70,6 +70,11 @@ private Device() {
          */
         static final String PRODUCT_NAME = "MSAL.Android";
 
+        /**
+         * The String representing the sdk platform version.
+         */
+        static final String PRODUCT_VERSION = "0.2.2";
```
iambmelt: Let's pull this from `BuildConfig` - this looks like something we've been forgetting to update if it's still @ `0.1.3`.

iambmelt: Issue filed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/385

rapong: Let's do that separately.

PR #383, opened 2019-02-20
Fix issue where default scopes are not set

kreedula:
- Adding default scopes to `Operation parameters` o initialization
issue : https://github.com/AzureAD/ad-accounts-for-android/issues/827

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
@@ -27,8 +27,8 @@ import android.content.Intent;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.support.annotation.NonNull;
+import android.text.TextUtils;
 
-import com.google.gson.Gson;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.exception.ArgumentException;
 import com.microsoft.identity.common.exception.ClientException;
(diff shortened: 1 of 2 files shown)


iambmelt:
Should we consider using a Set to hold the scopes? How do we ensure that scopes don't get added multiple times?

kreedula:
Set would have been nicer, but needs some API changes w.r.t Msal and also probably public facing API. For now I have added logic in the set method here to avoid duplication here : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/383/files#diff-37c9340cd26ba11cb1d207ed29a72f9bR67

iambmelt:
Yuck - we should make a backlog item for that. As for that check, you'll want to trim() and do the check as case insensitive -- the case for a Set (with a proper Comparator) getting stronger already 😅

iambmelt:
I can make it in the morning, currently on a mobile

kreedula:
Thanks @iambmelt , I will add a backlog item, scopes are case sensitive per rfc [here ](https://tools.ietf.org/html/rfc6749#section-3.3). Trimming is good idea.

shoatman:
I added a comment to your other PR on this topic.  FYI - In AAD scopes are not case sensitive.  One of the reasons that we moved validation of requests to the strategy associated with the issuer.  Some IDPs will follow the spec and others will not.

kreedula:
Closing this , opened a new one here : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/386

Issue #384, opened 2019-02-20
OperationParameters  should declare mScopes as Set instead of ArrayList to avoid duplicates


kreedula:
This would avoid additional checks for duplicate scopes

Related PR for more context : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/383

PR #386, opened 2019-02-20
 Changes to make scopes type as Set

kreedula:
- Also added `addDefaultScopes` util method to base controller
- Setting accesstoken target from request scopes.
- Fixes : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/384

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
@@ -27,8 +27,8 @@ import android.content.Intent;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.support.annotation.NonNull;
+import android.text.TextUtils;
 
-import com.google.gson.Gson;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.exception.ArgumentException;
 import com.microsoft.identity.common.exception.ClientException;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -368,6 +356,16 @@ protected boolean accessTokenIsNull(ICacheRecord cacheRecord) {
         return null == cacheRecord.getAccessToken();
     }
 
+    protected void addDefaultScopes(final OperationParameters operationParameters){
+        final Set<String> requestScopes = operationParameters.getScopes();
+        requestScopes.add(OPEN_ID_SCOPE);
+        requestScopes.add(OFFLINE_ACCESS_SCOPE);
+        requestScopes.add(PROFILE_SCOPE);
+        // sanatize empty and null scopes
```
iambmelt: nit: `sanitize`

kreedula:
Thanks @iambmelt , Fixed the typo

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -77,6 +75,10 @@
 
     private static final String TAG = BaseController.class.getSimpleName();
 
+    private static final String OPEN_ID_SCOPE = "openid";
```
rapong: nit: I suggest putting these values in AuthenticatorConstants, as these values is also being used somewhere else (PRTController)

kreedula: Thanks for the suggestion @rapong . Addressed it

kreedula:
Thanks @iambmelt

PR #388, opened 2019-02-21
Adding @Expose attribute to requests, responses and error responses in support of logging

shoatman:
Added additional non-PII logging of requests and response objects.  Controlling access to PII data via the Expose annotation.  Data that is considered PII is not marked with the attribute.  Then Gson is configured to only serialize fields that are marked with the annotation.  Regular Gson instances are unaffected by this attribute.


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -58,12 +59,14 @@
     /**
      * Can be used to pre-fill the username/email address field of the sign-in page for the user, if you know their username ahead of time.
      */
+    @Expose()
```
iambmelt: PII?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/ISuccessResponse.java
```diff
@@ -0,0 +1,26 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.providers.oauth2;
+
+public interface ISuccessResponse {
```
iambmelt: This looks like a marker interface but it's not immediately clear to me how it's being used - is this being consumed somewhere?

iambmelt:
As a minor syntactic note, we should be able to omit the empty parens and tag fields as `@Expose` (instead of `@Expose()`)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/ISuccessResponse.java
shoatman: Yes.  IResult was updated.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
shoatman: Wasn't sure about this myself.  Am expecting a document that may shed some additional light.

heidijinxujia: upn and login hint are pii.

common/src/main/java/com/microsoft/identity/common/internal/request/AcquireTokenOperationParameters.java
```diff
@@ -33,9 +34,13 @@
 public class AcquireTokenOperationParameters extends OperationParameters {
 
     private transient Activity mActivity;
+    @Expose()
```
heidijinxujia: pii

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -66,6 +69,7 @@
      * application, the authorization server will redirect the user back to the application with
      * either an authorization code or access token in the URL.
      */
+    @Expose()
```
heidijinxujia: @shoatman redirect_uri is OII. Not sure if we should include it into logging.

PR #397, opened 2019-02-26
Broker Cache Upgrade

iambmelt:
Work-in-progress: (testing, code cleanup, parallelization)

These changes increase the visibility of access modifiers to support migration. Additionally, new API is added to support syncing SSO state to the broker cache.


common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -895,4 +896,66 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
 
         return targetCache;
     }
+
+    public boolean setSingleSignOnState(@NonNull final String uidStr,
```
iambmelt: Javadoc

iambmelt: Resolved in 4cf8055b1036217a1508690997c7f2d0f83ee83b

PR #398, opened 2019-02-28
Device Registration changes for a MAM resource

kreedula:
- Added new exception to track this specific scenerio.
- Also changed MSAL version to 0.3.0

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
@@ -66,17 +66,17 @@ public class ExceptionAdapter {
                 switch (result.getAuthorizationResult().getAuthorizationStatus()) {
                     case FAIL:
                         // Check if the error is to register device and throw DEVICE_REGISTRATION_NEEDED exception
-                        if(authorizationErrorResponse instanceof MicrosoftAuthorizationErrorResponse){
+                        if (authorizationErrorResponse instanceof MicrosoftAuthorizationErrorResponse) {
                             MicrosoftAuthorizationErrorResponse microsoftAuthorizationErrorResponse =
                                     (MicrosoftAuthorizationErrorResponse) authorizationErrorResponse;
 
-                            if(microsoftAuthorizationErrorResponse.getError().equals(
-                                    MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED)){
+                            if (microsoftAuthorizationErrorResponse.getError().equals(
+                                    MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED)) {
 
                                 return new DeviceRegistrationRequiredException(
                                         microsoftAuthorizationErrorResponse.getError(),
-                                        microsoftAuthorizationErrorResponse.getErrorDescription()
-                                        , microsoftAuthorizationErrorResponse.getUserName()
+                                        microsoftAuthorizationErrorResponse.getErrorDescription(),
+                                        microsoftAuthorizationErrorResponse.getUserName()
                                 );
                             }
                         }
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -63,6 +65,22 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
...
+                            MicrosoftAuthorizationErrorResponse microsoftAuthorizationErrorResponse =
+                                    (MicrosoftAuthorizationErrorResponse) authorizationErrorResponse;
+
+                            if(microsoftAuthorizationErrorResponse.getError().equals(
+                                    MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED)){
+
+                                return new DeviceRegistrationRequiredException(
+                                        microsoftAuthorizationErrorResponse.getError(),
+                                        microsoftAuthorizationErrorResponse.getErrorDescription()
+                                        , microsoftAuthorizationErrorResponse.getUserName()
```
iambmelt: Nit: Just some strange formatting here

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -65,7 +65,7 @@
 public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     private static final String TAG = AzureActiveDirectoryWebViewClient.class.getSimpleName();
     private static final String INSTALL_URL_KEY = "app_link";
-    private static final String INSTALL_UPN_KEY = "username";
+    public static final String INSTALL_UPN_KEY = "username";
```
iambmelt: Should we consider graduating these constants out of this file? Not necessarily that it doesn't belong here (it's set here, after all) but, I notice that now the `MicrosoftStsAuthorizationResultFactory` is reaching-into this class to grab this constant which is maybe a little weird organizationally speaking.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResultFactory.java
```diff
@@ -93,11 +113,18 @@ public MicrosoftStsAuthorizationResult createAuthorizationResult(final int resul
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
                         MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED,
                         MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED);
+                // Set username returned from the service
+                result.getAuthorizationErrorResponse().setUserName(data.getStringExtra(
```
iambmelt: As one minor additional note, something here looks 'unsafe' -- will the error response always be here? will it always have a username? Is there no need at all to check that these values `!= null` ?

Chalk my questions up to paranoia 😅

kreedula: Thanks for the raising this. I have double checked, `
createAuthorizationResultWithErrorResponse` will return you a non-null error response, so we are good here.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
kreedula: Good call, I was being lazy here and just changed the constant to public. Moved to `AuthenticationConstants` as appropriate

PR #399, opened 2019-03-01
Move SubError to BaseException. Refactoring some code.

rpdome:
- UIRequiredInteraction will only represent any error that requires, well, UI Interaction.
- Some sub error can't be fixed by UI interaction (i.e. WPJ Device is disabled by admin), therefore I moved all of the tokenResult-based exception in the lower layer to ServiceException (and let its caller, namely acquireToken/acquireToken silent) decide if it should wrap those exception with UiRequiredException - this will be in another card)
- Remove redundant code.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
+++ b/common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
@@ -42,6 +42,8 @@ public class BaseException extends Exception {
 
     private String mErrorCode;
 
+    private String mSubErrorCode;
+
     /**
      * Default constructor.
      */
@@ -87,6 +89,22 @@ public class BaseException extends Exception {
      */
     public String getErrorCode() {
         return mErrorCode;
+    }
+
+    /**
+     * @return The sub error code for the exception, could be null. {@link BaseException} is the top level base exception, for the
+     * constants value of all the error code.
+     */
+    public String getSubErrorCode() {
+        return mSubErrorCode;
+    }
+
+    /**
+     * @@param subErrorCode - The sub error code for the exception, could be null. {@link BaseException} is the top level base exception, for the
+     * constants value of all the error code.
+     */
+    public void setSubErrorCode(@Nullable final String subErrorCode){
+        mSubErrorCode = subErrorCode;
     }
 
     /**
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
@@ -99,28 +99,18 @@ public class ExceptionAdapter {
             );
         }
 
-        final TokenResult tokenResult = result.getTokenResult();
+        return exceptionFromTokenResult(result.getTokenResult());
+    }
+
+    /**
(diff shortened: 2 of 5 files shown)


kreedula:
@rpdome I believed This would impact both Msal and ad-accounts controllers. So chnages need to me made there as well if we are getting to this. See `LocalMsalController` [here ](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java)

kreedula:
Also recommend to get a approval from @shoatman for these changes

common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
```diff
@@ -42,6 +42,8 @@
 
     private String mErrorCode;
 
+    private String mSubErrorCode;
```
iambmelt: One thing that jumps out to me here is the somewhat confusing API `BaseException` has developed - looking at this class we see `errorCode` and `subErrorCode` and would naturally assume "well, those must be error response payloads". In actuality though, we populate `errorCode` with our own non-service related codes too ([example](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/exception/ArgumentException.java#L39)).

Just thinking out loud.... it would be helpful to have our errors _and_ their origins be clear. 

For the purposes of this PR, would it make sense to beef-up `ServiceException` to provide access to error and suberror payloads and use that to signal instances where popping UI won't resolve the issue? If yes, could we roll specific errors (which may been returned to us by the service in subErrors) up to a series of single error codes we surface? `WPJ_DEVICE_IS_DISABLED` etc etc?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -99,29 +99,19 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
...
+     * Extract exception from a given TokenResult object.
+     * */
+    public static ServiceException exceptionFromTokenResult(TokenResult tokenResult) {
+        final String methodName = ":exceptionFromTokenResult";
         final TokenErrorResponse tokenErrorResponse;
 
         if (tokenResult != null && !tokenResult.getSuccess()) {
             tokenErrorResponse = tokenResult.getErrorResponse();
 
-            if (tokenErrorResponse.getError().equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)) {
```
kreedula: @rpdome As synced offline, this can be added back which means no changes will be required in Msal or ad-accounts controllers. 

In addition any parsing to look and map for UIRequiredException can be done here, like looking of "interaction_required" here as discussed offline

common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
rpdome: Feedback taken. Will push out a change regarding the error code series in ad_accounts at some point

rpdome:
Closing this as the scope changed. Will open a new one.

Issue #393, opened 2019-02-23
generateSignedJWT failed, audience is missing.


rpdome:
generateSignedJWT throws an exception because "audience" is missing.

However, in getChallengeHeader(final PKeyAuthChallenge pKeyAuthChallenge), we're passing audience with pKeyAuthChallenge.getThumbprint(). and mThumbprint is never set. (In PKeyAuthChallenge constructor, it is always null)

@heidijinxujia FYI

kreedula:
Even I have encountered this 
`java.lang.IllegalArgumentException: audience
        at com.microsoft.identity.common.adal.internal.JWSBuilder.generateSignedJWT(JWSBuilder.java:124)
        at com.microsoft.identity.common.internal.ui.webview.challengehandlers.PKeyAuthChallengeHandler.getChallengeHeader(PKeyAuthChallengeHandler.java:108)
        at com.microsoft.identity.common.internal.ui.webview.challengehandlers.PKeyAuthChallengeHandler.processChallenge(PKeyAuthChallengeHandler.java:66)
        at com.microsoft.identity.common.internal.ui.webview.AzureActiveDirectoryWebViewClient.handleUrl(AzureActiveDirectoryWebViewClient.java:120)
        at com.microsoft.identity.common.internal.ui.webview.AzureActiveDirectoryWebViewClient.shouldOverrideUrlLoading(AzureActiveDirectoryWebViewClient.java:109)
        at aCt.a(SourceFile:1)
        at xH.b(SourceFile:78)
        at org.chromium.android_webview.AwContentsClientBridge.shouldOverrideUrlLoading(SourceFile:150)
        at android.os.MessageQueue.nativePollOnce(Native Method)
        at android.os.MessageQueue.next(MessageQueue.java:323)
        at android.os.Looper.loop(Looper.java:136)
        at android.app.ActivityThread.main(ActivityThread.java:6776)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410)`

kreedula:
[...]

PR #403, opened 2019-03-06
Fix the pkeyauth flow

heidijinxujia:
[passed] test case:[ Test Case 652091](https://identitydivision.visualstudio.com/IDDP/_workitems/edit/652091)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/PKeyAuthChallengeHandler.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/PKeyAuthChallengeHandler.java
@@ -60,35 +60,30 @@ public final class PKeyAuthChallengeHandler implements IChallengeHandler<PKeyAut
     public Void processChallenge(final PKeyAuthChallenge pKeyAuthChallenge) {
         mWebView.stopLoading();
         mChallengeCallback.setPKeyAuthStatus(true);
-        new Thread(new Runnable() {
-            @Override
-            public void run() {
 
-                try {
-                    //Get no device cert response
-                    final Map<String, String> header = getChallengeHeader(pKeyAuthChallenge);
+        try {
+            //Get no device cert response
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -60,30 +60,35 @@ public PKeyAuthChallengeHandler(@NonNull final WebView view,
     public Void processChallenge(final PKeyAuthChallenge pKeyAuthChallenge) {
         mWebView.stopLoading();
         mChallengeCallback.setPKeyAuthStatus(true);
+        new Thread(new Runnable() {
```
kreedula: Why a new Thread ? ` mWebView.post(new Runnable() { ` is again executed on UI thread anyways right ?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/PKeyAuthChallengeHandler.java
```diff
@@ -105,8 +110,11 @@ public void run() {
                 if (privateKey == null) {
```
kreedula: Wouldn't the if statement always be false ? This still needs to be fixed i believe , where we need to set the right thumbprint from the challenge headers and compare with the one in the device certificate

heidijinxujia: "thumbprint from the challenge headers" this only possibly happen when talking to :token endpoint.  This happens when broker receive device certificate challenge request during the token request. 

And currently, the code is missing this part. https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/406 I'm going to fix this in other PR. This PR is focusing on fixing the pkeyauth challenge during the :auth flow.

heidijinxujia:
#393 Fix

PR #405, opened 2019-03-06
Centralize the Exception conversion logic

rpdome:
- Move subError to ServiceException (since subError is OAuth2 specific)
- extract getExceptionFromOAuthError() function so that it can also be called from ad_accounts (primarily in PRTController).

another card in ad_accounts will follow.


common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -137,111 +130,131 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
...
+    /**
+     * Get an exception object from the given oAuth values.
+     *
+     * @param oAuthError
+     * @param oAuthSubError
+     * @param oAuthErrorDescription
+     * @param response
+     * @return ServiceException, UiRequiredException
+     * */
+    public static ServiceException getExceptionFromOAuthError(final String oAuthError,
```
shoatman: Is there a reason that this method doesn't just accept the TokenErrorResponse?  Since it's using all of the information from that object?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -137,111 +130,131 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
...
+     *
+     * @param oAuthError
+     * @param oAuthSubError
+     * @return boolean
+     * */
+    private static boolean shouldBeConvertedToUiRequiredException(final String oAuthError,
+                                                                  final String oAuthSubError){
+        // Invalid_grant doesn't necessarily requires UI protocol-wise.
+        // We simplify our logic because this layer is also used by MSAL.
+        if (oAuthError.equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)){
```
shoatman: Is this where we plan to add invalid_resource for now?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -137,111 +130,131 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
...
             outErr.setCliTelemErrorCode(cliTelemInfo.getServerErrorCode());
             outErr.setCliTelemSubErrorCode(cliTelemInfo.getServerSubErrorCode());
         }
     }
 
-    private static void applyHttpErrorResponseData(@NonNull final ServiceException targetException,
-                                                   final int statusCode,
-                                                   @Nullable String responseHeadersJson,
-                                                   @Nullable String responseBody) {
+    private static HttpResponse synthesizeHttpResponse(final int statusCode,
```
shoatman: What's consuming HTTP Response above that could not consume the TokenErrorResponse?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
rpdome: This will also be called by PRTController (so that we have one place that does the exception conversion, which is ExceptionAdapter)

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
rpdome: I'm not sure what that is. This is only for converting Exception type. 

For error-handling (inside Broker), we should still use the error code + subError code.

common/src/main/java/com/microsoft/identity/common/exception/ServiceException.java
```diff
@@ -139,6 +156,7 @@ public void setHttpResponse(final HttpResponse response) throws JSONException {
     public ServiceException(final String errorCode, final String errorMessage, final Throwable throwable) {
         super(errorCode, errorMessage, throwable);
         mHttpStatusCode = DEFAULT_STATUS_CODE;
+        mSubErrorCode = "";
```
kreedula: Any reason we are initializing here to empty string , can't just be left null ?

common/src/main/java/com/microsoft/identity/common/exception/ServiceException.java
```diff
@@ -152,5 +170,6 @@ public ServiceException(final String errorCode, final String errorMessage, final
     public ServiceException(final String errorCode, final String errorMessage, final int httpStatusCode, final Throwable throwable) {
         super(errorCode, errorMessage, throwable);
         mHttpStatusCode = httpStatusCode;
+        mSubErrorCode = "";
```
kreedula: Same here

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
rpdome: None. It's just that getExceptionFromOAuthError is also consumed in PRTController, and over there it parses httpResponse along with the error codes. The only TokenErrorResponse have are strings, so we have to convert them back.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
kreedula: We should add `AuthenticationConstants.OAuth2ErrorCode.INTERACTION_REQUIRED` here in addition to Invalid grant to fix the issue which Shane encountered, he opened a PR to fix : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/400 but as discussed offline this should be fixed here

rpdome: I totally missed that. thanks.

common/src/main/java/com/microsoft/identity/common/exception/ServiceException.java
rpdome: I could leave it null yes.

kreedula:
@rpdome in addition, I think `pmd` is falling on this PR, might be a minor issue , you can run `gradlew common:pmd`  locally to see what's the issue

rpdome:
> @rpdome in addition, I think `pmd` is falling on this PR, might be a minor issue , you can run `gradlew common:pmd` locally to see what's the issue

@kreedula resolved. There was an unused param. Thanks!

rpdome:
Thanks @kreedula  @shoatman

PR #414, opened 2019-03-13
Added additional constant

shoatman:


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -969,9 +969,15 @@ private AuthenticationConstants() {
...
         /**
-         * String to send Msal V2 Request params.
+         * String to return Msal V2 response.
          */
         public static final String BROKER_RESULT_V2 = "broker.result.v2";
+
+        /**
+         * String to return Msal V2 response.
+         */
+        public static final String BROKER_USERINFO_RESULT_V2 = "broker.userinfo.v2";
```
iambmelt: I think the naming here implies too much that this will contain a `UserInfo` when this will really contain and `AccountRecord` instead - can we adjust naming to steer us away from that assumption?

shoatman: Renamed to AccountRecord.  Thanks Brian.

PR #423, opened 2019-03-19
Update common to support device_authentication_failed changes.

rpdome:
https://github.com/AzureAD/ad-accounts-for-android/pull/889


rpdome:
That was really fast Thanks @iambmelt

PR #437, opened 2019-03-26
Do not send select_account if login_hint is set

rpdome:
If we send both, SELECT_ACCOUNT wins and the user will always be asked to select account in the webView.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
@@ -125,8 +125,7 @@ public abstract class BaseController {
             );
 
             // We don't want to show the SELECT_ACCOUNT page if login_hint is set.
-            if (acquireTokenOperationParameters.getLoginHint() != null &&
-                    acquireTokenOperationParameters.getLoginHint().length() > 0 &&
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -123,6 +124,13 @@ public abstract AcquireTokenResult acquireTokenSilent(final AcquireTokenSilentOp
                     acquireTokenOperationParameters.getRequestHeaders()
             );
 
+            // We don't want to show the SELECT_ACCOUNT page if login_hint is set.
+            if (acquireTokenOperationParameters.getLoginHint() != null &&
```
iambmelt: We have a handy convenience function for checking non-`null`ness as well as length (see `StringExtensions.isNullOrBlank()`)

kreedula: or we can use Android's in build lib, `TextUtils.isEmpty()`

iambmelt: @kreedula +1 we should transition to this (low pri)

PR #442, opened 2019-03-27
Breaking out v1 resource -> scope conversion to a shared method

iambmelt:
Behavioral changes are still TBD on this. For now, moving to a shared method with broker.

See:
- https://github.com/AzureAD/ad-accounts-for-android/pull/915

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
@@ -433,25 +433,31 @@ public class TokenCacheItemMigrationAdapter {
      */
     @NonNull
     public static String getScopesForTokenRequest(@NonNull final String v1Resource) {
-        final String methodName = ":getScopesForTokenRequest";
-
-        String scopes = v1Resource;
-
-        if (!v1Resource.endsWith("/")) {
-            scopes += "/.default";
-        } else {
-            Logger.warn(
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -433,27 +433,33 @@ public static ADALTokenCacheItem findFrt(
...
     }
 
+    /**
+     * Given a v1 resource uri, append '/.default' to convert it to a v2 scope.
+     *
+     * @param resource The v1 resource uri.
+     * @return The v1 resource uri as a scope.
+     */
+    @NonNull
+    public static String getScopeFromResource(@NonNull final String resource) {
```
kreedula: We can another duplicate here : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java#L309

We can remove either of them and use one.

iambmelt: Thanks for calling this out -- this change will need to happen in a separate PR, since this file doesn't exist in `common` in this branch

iambmelt:
Closing and moving to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/444

PR #448, opened 2019-03-28
Updates to support PPE configuration setting

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/authorities/Environment.java
```diff
@@ -0,0 +1,28 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.authorities;
+
+public enum Environment {
+    PreProduction,
```
iambmelt: In the event that sovereign ppe environments come along in the future, is it worth naming this something like `PreProductionWorldwide`, `ProductionWorldwide`, etc?

shoatman: It looks like clouds belong to an environment (PPE, Production).  At least based on the instance disocovery document which is environment specific.

common/src/test/java/com/microsoft/identity/common/unit/AudienceDefaultPreProductionCloudUrlTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test
+    public void test_AnyOrganizationAccountDefaultCloudUrl() {
+
+        AzureActiveDirectoryAudience audience = new AllAccounts();
+        Assert.assertEquals(AzureActiveDirectoryEnvironment.PREPRODUCTION_CLOUD_URL, audience.getCloudUrl());
+
+    }
+
+    @Test
+    public void test_AnyPerosnalAccountDefaultCloudUrl()  {
```
kreedula: Nit : typo "Personal"

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -38,7 +39,11 @@
     public static final String ALL = "common";
 
     public String getCloudUrl() {
-        return mCloudUrl;
+        if(mCloudUrl == null){
```
kreedula: Should we check for empty here too?

PR #449, opened 2019-03-28
[WIP] Resolves #554 (Adds public API for setting MSAL SecretKey byte data)

iambmelt:
Work in progress.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -27,6 +27,7 @@ import android.annotation.TargetApi;
 import android.content.Context;
 import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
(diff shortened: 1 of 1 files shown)


iambmelt:
In my testing, ADAL is less resilient than MSAL in catching and presenting this error to the user.

Will investigate tomorrow whether this would be more appropriately surfaced as a checked `Exception` or if ADAL's logic can be shored-up in some intelligent way

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -130,18 +132,28 @@ public final void putString(
         if (null == mStorageHelper) {
             editor.putString(key, value);
         } else {
-            editor.putString(key, encrypt(value));
+            final String encryptedValue = encrypt(value);
```
iambmelt: In an upcoming PR, `null` may be used to signal an Exception case. For now, preserving existing behavior

PR #459, opened 2019-04-05
Removed single task from Authorization Activity

kreedula:
- This resulted in AuthorizationActivity being created in Broker app instead of Client app activity stack

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
@@ -95,15 +95,16 @@ public class BrowserAuthorizationStrategy<GenericOAuth2Strategy extends OAuth2St
         final Uri requestUrl = authorizationRequest.getAuthorizationRequestAsHttpRequest();
         authIntent.setData(requestUrl);
 
-        mReferencedActivity.get().startActivity(
-                AuthorizationActivity.createStartIntent(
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -95,15 +95,16 @@ public BrowserAuthorizationStrategy(@NonNull Activity activity, @NonNull Intent
...
-                        AuthorizationAgent.BROWSER));
+        final Intent intent = AuthorizationActivity.createStartIntent(
+                mReferencedActivity.get().getApplicationContext(),
+                authIntent,
+                mResultIntent,
+                requestUrl.toString(),
+                mAuthorizationRequest.getRedirectUri(),
+                mAuthorizationRequest.getRequestHeaders(),
+                AuthorizationAgent.BROWSER);
+        intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
heidijinxujia: Could you add a comment on this? like " singleTask launchMode is required for the authorization redirect in the browser authorization flow". Thanks.

iambmelt: +1 This feels easy for someone months or years down the line to inadvertently break

kreedula:
Thanks @heidijinxujia , @iambmelt , added the comment as suggested

PR #461, opened 2019-04-08
TenantProfile Implementation (Common Changes)

iambmelt:
Supports `TenantProfile` internals


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -571,6 +760,76 @@ public AccountRecord getAccountWithLocalAccountId(@Nullable final String environ
...
+        );
+
+        // Iterate and populate
+        for (final AccountRecord acct : allAppAccounts) {
+            if (allMatchingAccountsByHomeId.contains(acct) && !accountRecord.equals(acct)) {
+                corollaryAccounts.add(acct);
+            }
+        }
+
+        return corollaryAccounts;
```
iambmelt: Make this List unmodifiable

iambmelt: Resolved in f4fb5b6042e1b678cbb49142cb110b54afa47106

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -211,16 +214,18 @@ protected void renewAccessToken(@NonNull final AcquireTokenSilentOperationParame
...
                     getAuthorizationRequest(strategy, parameters),
                     tokenResult.getTokenResponse()
             );
+            final ICacheRecord savedRecord = savedRecords.get(0);
 
             // Create a new AuthenticationResult to hold the saved record
             final LocalAuthenticationResult authenticationResult = new LocalAuthenticationResult(
                     savedRecord,
-                    parameters.getSdkType()
+                    savedRecords
```
shoatman: What happened to SDK type here?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -140,6 +140,52 @@ ICacheRecord save(@NonNull AccountRecord accountRecord,
...
+
+    @NonNull
+    private List<ICacheRecord> mergeCacheRecordWithCorollaryAccounts(
+            @NonNull final ICacheRecord savedCacheRecord) {
+        final List<ICacheRecord> result = new ArrayList<>();
+        // Whatever ICacheRecord you provide will _always_ be the first element in the result List.
+        result.add(savedCacheRecord);
+
+        final List<AccountRecord> corollaryAccounts = new ArrayList<>(
+                getCorollaryAccounts(
```
shoatman: Naming.... we probably need to simplify this....

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -140,6 +140,52 @@ ICacheRecord save(@NonNull AccountRecord accountRecord,
...
+                                .getRefreshToken()
+                                .getClientId(),
+                        savedCacheRecord
+                                .getAccount() // This account wil be the 0th element in the result.
+                )
+        );
+
+        if (!corollaryAccounts.isEmpty()) {
+            // Remove the first element from the List since it is already contained in the result List
+            corollaryAccounts.remove(0);
```
shoatman: This entry was added in getCorollaryAccounts... and then removed here.  I think this may be due to the same method being used from 2 different paths.  Just confirming that this is necessary.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -235,14 +281,58 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
...
+
+    /**
+     * Given an AccountRecord and associated client_id, load a sparse ICacheRecord containing
+     * the provided AccountRecord and its accompanying IdTokens.
+     *
+     * @param clientId The client_id relative to which IdTokens should be loaded.
+     * @param acct     The target AccountRecord.
+     * @return A sparse ICacheRecord containing the provided AccountRecord and its IdTokens.
+     */
+    ICacheRecord getSparseCacheRecordForAccount(@NonNull final String clientId,
```
shoatman: Not sure I followed what "sparse" was meant to convey here.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -235,14 +281,58 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
...
+                                                @NonNull final AccountRecord acct) {
+        final String methodName = ":getSparseCacheRecordForAccount";
+
+        final List<IdTokenRecord> acctIdTokens = getIdTokensForAccount(
+                clientId,
+                acct
+        );
+
+        if (acctIdTokens.size() > new CredentialType[]{
```
shoatman: I think it might be simpler to have a constant here....

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -395,37 +485,92 @@ public ICacheRecord load(@NonNull final String clientId,
...
+            if (!account.equals(cacheRecord.getAccount())) {
+                result.add(cacheRecord);
+            }
+        }
+
+        return result;
+    }
+
+    @Override
+    public List<IdTokenRecord> getIdTokensForAccount(@Nullable String clientId,
```
shoatman: I think Account in the internal sense vs Account in the public sense is likely to get confusing for us going foward.  We may want to align these two models of the world in our naming at some point to improve our chances of avoiding errors.

iambmelt: Resolved in be3f9ffcba3e3f70e72b7b792c79f0e64039f0a7

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
iambmelt: I added [another constructor that specifies its type as MSAL by default](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/tenant-profile/common/src/main/java/com/microsoft/identity/common/internal/result/LocalAuthenticationResult.java#L58-L61).

iambmelt: I'll probably remove this additional constructor and pass-in the param, though. In an upcoming commit...

iambmelt: Resolved in 70dbc62076909d5d18aa4471540145288ab000a8

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Resolved in 1e724f6ec32bb12e997da4af53a223817cf15a1f

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: I added a clarifying sentence to the javadoc in 1cebf6e0deab478263b0e35a0fd6ef34218956b7 - I am borrowing sparse here to signal that elements inside this structure are going to be `null`

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Yep - this is an artifact of the code paths... I wanted the behavior of both `mergeCacheRecordWithCorollaryAccounts` and `getCorollaryAccounts` to be the same: to always return their input as the 0th element.

Since `mergeCacheRecordWithCorollaryAccounts` calls `getCorollaryAccounts` and they both return their input as the first element, I need to remove the first element of the result so I don't end up with the same record twice in the final result.

This could be factored out[ by only calling `result.add()` if `corollaryAccounts.isEmpty()`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/tenant-profile/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java#L159-L160) but I thought that would be less obvious.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Took a pass at simplifying the language in 659d973a3ce00ffe8db67ed32d119eed05c36b60

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -617,6 +842,92 @@ public AccountRecord getAccountWithLocalAccountId(@Nullable final String environ
...
+                    mCallingProcessUid
+            );
+
+            result = cache.getIdTokensForAccountRecord(
+                    clientId,
+                    accountRecord
+            );
+        } else {
+            // If the client id was null... then presumably we want to aggregate the IdTokens across
+            // apps... why would you want that? For now, throw an Exception and see if anyone requests
```
rpdome: Should we make it @NonNull in that case?

iambmelt:
This PR should gate on https://github.com/AzureAD/ad-accounts-for-android/pull/1008, which needs to pick up the new API signature in `BaseController`

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -774,7 +1085,7 @@ public boolean isClientIdKnownToCache(@NonNull final String clientId) {
      *
      * @return A List of ICacheRecords for the FoCI accounts.
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings(UNCHECKED)
     public List<ICacheRecord> getFociCacheRecords() {
```
iambmelt: TODO This should be updated to return the `IdToken`

iambmelt: Resolved in 874474577f25a6c533ea388751dccbece21e2d77

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
iambmelt: Resolved in ad4e15dde76044b232e91be74420dad5e3aab778

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -571,6 +799,76 @@ public AccountRecord getAccountWithLocalAccountId(@Nullable final String environ
...
+
+        // Grab all of the AccountRecords associated with this clientId
+        final List<AccountRecord> allAppAccounts = getAccounts(
+                accountRecord.getEnvironment(),
+                clientId
+        );
+
+        // Iterate and populate
+        for (final AccountRecord acct : allAppAccounts) {
+            if (allMatchingAccountsByHomeId.contains(acct) && !accountRecord.equals(acct)) {
```
kreedula: Did we override `AccountRecord` equals?

iambmelt: Yup - [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/dto/AccountRecord.java#L426-L477).

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -233,6 +234,24 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
         return result;
     }
 
+    public List<ICacheRecord> saveAndLoadAggregatedAccountData(
+            @NonNull final AccountRecord accountRecord,
+            @NonNull final IdTokenRecord idTokenRecord,
+            @NonNull final AccessTokenRecord accessTokenRecord,
+            @Nullable final String familyId) throws ClientException {
+        // TODO Wire this up _for real_
```
iambmelt: TODO

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -402,6 +486,80 @@ public ICacheRecord load(@NonNull final String clientId,
...
+        if (shouldUseFociCache) {
+            // We do not have a cache for this app or it is not yet known to be a member of the family
+            // use the foci cache....
+
+            // Load a sparse-record (if available) containing only the desired account and a
+            // refresh token if available...
+            resultRecords = new ArrayList<>();
+            resultRecords.add(
+                    mFociCache.loadByFamilyId(
+                            clientId,
```
heidijinxujia: Just qq. ClientId should not be necessary when fetching the FRT. Do you know why we need the clientId as a parameter here to load the foci cache?

heidijinxujia: Is there any future extensibility you are considering here?

iambmelt: This is only for [logging](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java#L78-L81) (so we can see in logs which app is querying broker), the field is `@Nullable`; this is mentioned in the code comments [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java#L88-L91)

PR #481, opened 2019-04-22
Closes #480

### LINKED ISSUE #480: [Cache] Values which cannot be successfully decrypted should be removed with a warning logged
If we load a cache value and try to decrypt it and `null` is returned we log a warning, remove the value, and pretend we didn't see it.

This can _reasonably_ be hit if encryption keys are rolled within an app, Company Portal does this during unenroll.

iambmelt:
See #480

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -27,6 +27,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.text.TextUtils;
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -170,7 +177,14 @@ public final String getSharedPreferencesFileName() {
 
         if (null != mStorageHelper) {
             for (final Map.Entry<String, String> entry : entries.entrySet()) {
-                entry.setValue(decrypt(entry.getValue()));
+                final String decryptedValue = decrypt(entry.getValue());
```
iambmelt: This can return `null` values with non-null keys. Fix it.

iambmelt: Resolved in 2522327248df100e3c728f9ea860d39063442f0d

PR #482, opened 2019-04-23
[FLW] Make future class generic + add a function to .aidl file.

rpdome:


rpdome:
Thanks @iambmelt and @kreedula!

PR #485, opened 2019-04-24
Browser safe list support

heidijinxujia:
**Problem**
To avoid the broken user experience when auth through browsers.

**Proposal**
BrowserDescriptor describes the browser, including the package name, browser's signature hashes, if using the custom tab and the version range that supports the MSAL browser authorization flow.
```json
{
  "browser_package_name": "com.android.chrome",
  "browser_signature_hashes": [
 "7fmduHKTdHHrlMvldlEqAIlSfii1tl35bxj1OXN5Ve8c4lU6URVu4xtSHc3BVZxS6WWJnxMDhIfQN0N0K2NDJg=="
      ],
  "browser_use_customTab" : true,
  "browser_version_lower_bound" : "45"
}
```

The browser safe list is a list of BrowserDescriptor objects and is defined in the msal configuration json file. Before MSAL app trying to launch a browser, it will look into the browser safe list. If there is a match, the app will launch the matching browser, otherwise, MSAL app will try the next installed browser until it finds a match. If no matched browser found, MSAL will fallback into the embedded webview flow. However, for the B2C scenario, as the browser flow is required, if no matching browser found, MSAL will throw the ClientException to the dev.

**Work Items**
- Add the BrowserDescriptor class to represent the browser attributes
- Add the BrowserSafeList property into the BrowserAuthorizationStrategy
- Add the safe list filtering inside the browser selecting logic during the authorization process
- Add unit test

**Discussion**
For the safe list object, is it better to keep it inside the BrowserAuthorizationStrategy or make it a single instance class?

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/BrowserSelectorTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/BrowserSelectorTest.java
@@ -1,0 +1,221 @@
+package com.microsoft.identity.common;
+
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.pm.ActivityInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
+import android.content.pm.ResolveInfo;
+import android.content.pm.Signature;
+import android.net.Uri;
+import android.support.test.runner.AndroidJUnit4;
+
+import com.microsoft.identity.common.internal.ui.browser.Browser;
+import com.microsoft.identity.common.internal.ui.browser.BrowserSelector;
+
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+@RunWith(AndroidJUnit4.class)
+public class BrowserSelectorTest {
+    private static final String SCHEME_HTTP = "http";
+    private static final String SCHEME_HTTPS = "https";
+
+    static final Intent BROWSER_INTENT = new Intent(
+            Intent.ACTION_VIEW,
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserDescriptor.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * The function to compare the two versions.
+     *
+     * @param thisVersion
+     * @param thatVersion
+     * @return int -1 if thisVersion is smaller than thatVersion,
+     *         1 if thisVersion is larger than thatVersion,
+     *         0 if thisVersion is equal to thatVersion.
+     */
+    private int compareSemanticVersion(
```
iambmelt: Didn't we just add this function somewhere else? For broker `hello()`?

iambmelt: `BrokerUtils` -- this should be refactored into `common`, not copy/pasted

heidijinxujia: Thank you, Brian. Addressed.

common/src/androidTest/java/com/microsoft/identity/common/BrowserSelectorTest.java
```diff
@@ -0,0 +1,221 @@
+package com.microsoft.identity.common;
```
iambmelt: License info

iambmelt:
There is a failing unit test

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
```diff
@@ -48,21 +48,31 @@
...
-            return allBrowsers.get(0);
-        } else {
-            Logger.error(TAG, "No available browser installed on the device.", null);
-            throw new ClientException(ErrorStrings.NO_AVAILABLE_BROWSER_FOUND, "No available browser installed on the device.");
+
+        for (Browser browser : allBrowsers) {
+            for (BrowserDescriptor browserDescriptor : browserSafeList) {
+                if (browserDescriptor.matches(browser)) {
+                    Logger.verbose(TAG, "Select the browser to launch.");
+                    Logger.verbosePII(
```
iambmelt: I think we can probably log this as non-PII without issue - is there something I'm not considering?

PR #486, opened 2019-04-24
Default Prompt to None and changing tenant to organizations if it's common

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -71,12 +71,12 @@ public static OpenIdConnectPromptParameter _fromPromptBehavior(final String prom
             case "Auto":
             case "REFRESH_SESSION":
             case "Always":
-                return OpenIdConnectPromptParameter.SELECT_ACCOUNT;
+                return OpenIdConnectPromptParameter.NONE;
```
iambmelt: Can we drop a comment in here to explain what appears to be otherwise counterintuitive behavior?

PR #489, opened 2019-04-29
Add getAccountMode() and getCurrentAccount() to MSAuthService

rpdome:
Introducing 2 more functions to MicrosoftAuthService.
- getAccountMode()
- getCurrentAccount()

Please see [the spec](https://identitydivision.visualstudio.com/IDDP/_git/general-docs/pullrequest/744?_a=files&path=%2FSharedDevice%2FSharedDeviceAndroid.md) for reference.

### DIFF
--- a/common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
+++ b/common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
@@ -32,7 +32,11 @@ interface IMicrosoftAuthService {
 
     Bundle hello(in Bundle bundle);
 
+    Bundle getAccountMode();
+
     Bundle getAccounts(in Bundle bundle);
+
+    Bundle getCurrentAccount();
 
     Bundle acquireTokenSilently(in Bundle requestBundle);
 
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1022,20 +1022,20 @@ public final class AuthenticationConstants {
          * String to return the broker account mode.
          * This is used to determine what PublicClientApplication MSAL will return to its caller.
          */
-        public static final String BROKER_ACCOUNT_MODE = "broker.account.mode";
+        public static final String BROKER_ACCOUNT_MODE = "broker_account_mode";
 
         /**
          * Represents MSAL's single account mode.
          * In this mode, there will be only one signed-in account in this device (meaning that other accounts will not be able to acquire token.)
          * See SingleAccountPublicClientApplication for more details.
          */
-        public static final String BROKER_ACCOUNT_MODE_SINGLE_ACCOUNT = "broker.account.mode.single.account";
+        public static final String BROKER_ACCOUNT_MODE_SINGLE_ACCOUNT = "broker_account_mode_single_account";
 
         /**
          * Represents MSAL's multiple account mode.
          * See MultipleAccountPublicClientApplication for more details.
          */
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1018,6 +1018,25 @@ private AuthenticationConstants() {
          */
         public static final String BROKER_RESULT_V2 = "broker.result.v2";
 
+        /**
+         * String to return the broker account mode.
+         * This is used to determine what PublicClientApplication MSAL will return to its caller.
+         */
+        public static final String BROKER_ACCOUNT_MODE = "broker.account.mode";
```
kreedula: We recently switched to make these broker constant values snake case (changes in release branch). Can we do the same here for all new constants like `broker.account.mode` to `broker_account_mode

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -417,4 +419,62 @@ public Bundle bundleFromAccountRecordList(@NonNull final List<AccountRecord> rec
...
+        resultBundle.putString(BROKER_ACCOUNT_MODE, accountMode);
+        return resultBundle;
+    }
+
+    /**
+     * Get Account mode from bundle.
+     * @param bundle Bundle
+     * @return Account mode.
+     */
+    public static String getAccountModeFromBundle(@NonNull final Bundle bundle) {
```
kreedula: Rename to `accountModeFromBundle` for method naming consistency in this class

rpdome:
force pushed as I accidentally add the changes of another PR into this one.

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
```diff
@@ -32,8 +32,12 @@ interface IMicrosoftAuthService {
 
     Bundle hello(in Bundle bundle);
 
+    Bundle getAccountMode();
```
heidijinxujia: Question, who is setting the account mode? Broker or MSAL application?

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
```diff
@@ -32,8 +32,12 @@ interface IMicrosoftAuthService {
...
+
     Bundle getAccounts(in Bundle bundle);
 
+    Bundle getCurrentAccount();
```
heidijinxujia: 1. Should the getCurrentAccount() only be called when the account mode is "single"?
2.  If not, what should it return when multiple accounts exist? 
3. What kind of accounts are returning here? I assume WPJ account only, right?

rpdome: 1. Yes
2. An error bundle.
3. an account that holds BRT. In the shared device mode, there will be only 0 or 1 account that holds BRT, and that account MUST not be the WPJ account. (The WPJ account is used by the admin to bootstrap the WPJ/shared device state only)

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
rpdome: Broker. 

Here's roughly what happens in FLW scenario
1. Admin bootstrap a device with their 'admin' WPJ account through Authenticator app.
2. WPJ flows sends the flag to the server side, the server validates WPJ account's role.
3. If the role matches, the WPJ operation succeeds and the device is marked as shared - therefore a 'mode' is set.
4. MSAL queries this 'mode' to determine which set of Broker functionality it should expose to MSAL application.

PR #490, opened 2019-04-29
Allow a request MSAL Redirect URI to pass broker validation.

rpdome:
Allow a request MSAL Redirect URI to pass broker validation.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
@@ -200,14 +200,14 @@ public class BrokerValidator {
         return selfSignedCert;
     }
 
-
     public static boolean isValidBrokerRedirect(@Nullable final String redirectUri,
                                                 @NonNull final Context context,
-                                                @NonNull final String packageName){
+                                                @NonNull final String packageName,
+                                                @NonNull final String clientId){
         return !TextUtils.isEmpty(redirectUri) &&
-                redirectUri.equalsIgnoreCase(getBrokerRedirectUri(context, packageName));
+                    (redirectUri.equalsIgnoreCase(getBrokerRedirectUri(context, packageName)) ||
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -222,4 +222,14 @@ public static String getBrokerRedirectUri(final Context context, final String pa
...
                 signatureDigest);
     }
+
+    /**
+     * Helper method to get MSAL Redirect Uri
+     *
+     * @param clientId
+     * @return String : MSAL Redirect Uri
+     */
+    public static String getMsalRedirectUri(@NonNull final String clientId){
```
kreedula: I was under the impression that MSAL with V2 broker's redirect URI should always be of the form msauth://<package_name>/<sign_hash>. Did the requirement change ?

rpdome: I'm not certain about this, but this redirect URL pattern is available as an option in the portal. We're also using it in the test app (which is why I was hitting this error).

Might be worth bringing this up during the standup.

iambmelt: The scheme suggested in the _new portal_ should now match the broker redirect uri

```
msauth://<package_name>/<signature_hash_decoded>
```

iambmelt: @danieldobalian can link to new docs

rpdome: Does this mean we're not supporting "msal" + clientId + "://auth"; anymore?

If yes, then I'll drop this PR.

kreedula: Yes that redirect is only supported for just standalone MSAL, not with broker.

PR #493, opened 2019-05-02
Changes to cancel an existing auth request if a new auth request is invoked by the app

kreedula:
- Removed previous timeout change in controllers as it's no longer needed.
- Introduced new error and error strings as needed.
- Removed a redundant constant class `UIResponse` in `AuthorizationStrategy` and replaced with `AuthenticationConstants.UIResponse`

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
@@ -31,6 +31,7 @@ import com.microsoft.identity.common.exception.BaseException;
 import com.microsoft.identity.common.exception.UserCancelException;
 import com.microsoft.identity.common.internal.logging.DiagnosticContext;
 import com.microsoft.identity.common.internal.logging.Logger;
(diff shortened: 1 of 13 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -56,6 +57,14 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 "Beginning interactive request"
         );
         synchronized (sLock) {
+
+            if(!sInteractiveExecutor.isTerminated()){
```
iambmelt: Are we calling `shutdown()` or `shutdownNow()` anywhere?

kreedula: Actually this if loop is not needed, was part of my older change, missed to remove, we just have to send a broadcast and if there is an active `AuthorizationActivity` to will finish itself.

Updated it. Thanks Brian

kreedula:
Thanks @iambmelt

PR #499, opened 2019-05-08
Adds shared HttpResponseCache implementation to common

iambmelt:
Slated for `0.0.9-hf1` release.

Summary of changes:
- Introduces class `HttpCache` to handle bootstrapping/flushing of [`HttpResponseCache`](https://developer.android.com/reference/android/net/http/HttpResponseCache).
- Added in common to support eventual inclusion in MSAL/broker

Up next:
- ADAL integration in `1.16.2-hf1` release

Proposed usage in ADAL:
- Create cache in [`AuthenticationContext#initialize`](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/release/1.16.2-hf1/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L144-L161)
- [Flush cache](https://developer.android.com/reference/android/net/http/HttpResponseCache#flush()) after returning ATS result to caller?
    * Ideally, in the `finalize()` method of `AuthenticationContext` but the general consensus seems to be that it's a faux pas to override `finalize()`

Additional considerations:
- ADAL caches Instance Discovery Metadata in-memory already after loading. The optimization benefit in ADAL really comes through across app restarts. If the app is killed/restarted before the cache's `max-age` constraint is satisfied, then the Instance Discovery Metadata request can be serviced by the cache without a network round-trip. If restarts occur less frequently and `max-age` is frequently exceeded prior to initiating the Instance Discovery Metadata request, then I would expect no performance benefit.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -1,0 +1,124 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.net.cache;
+
+import android.net.http.HttpResponseCache;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import com.microsoft.identity.common.internal.logging.Logger;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Utility class for caching HTTP and HTTPS responses.
+ *
+ * @see HttpResponseCache
+ */
+public class HttpCache {
+
+    private static final String TAG = HttpCache.class.getSimpleName();
+
+    /**
+     * The default name of the HTTP/S cache on disk.
+     */
+    public static final String DEFAULT_HTTP_CACHE_NAME = "com.microsoft.identitiy.http-cache";
+
+    /**
+     * The default size of the HTTP/S cache on disk.
+     */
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param cacheDirectory The parent-directory in which the cache should reside.
+     * @param cacheFileName  The directory name for the HTTP cache.
+     * @param maxSizeBytes   The maximum allowed size of the cache. Once capacity is hit,
+     *                       entries are removed according to an LRU strategy.
+     * @return True if the cache was successfully installed. False otherwise.
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
+                                     @NonNull final String cacheFileName,
+                                     final long maxSizeBytes) {
+        final String methondName = ":initialize (3 arg)";
```
Om83: Do we need to be more descriptive here ? So Instead of 3 args ,something like Directory,FileName and Size?

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
+        } catch (IOException e) {
```
heidijinxujia: final

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
+                                     @NonNull final String cacheFileName,
+                                     final long maxSizeBytes) {
+        final String methondName = ":initialize (3 arg)";
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
```
heidijinxujia: It could be better if we set the success to true here.

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
shoatman: typo "methodName"

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Initializes a new {@link HttpResponseCache} based on the supplied parameters.
+     *
+     * @param cacheDirectory The parent-directory in which the cache should reside.
+     * @param cacheFileName  The directory name for the HTTP cache.
+     * @param maxSizeBytes   The maximum allowed size of the cache. Once capacity is hit,
+     *                       entries are removed according to an LRU strategy.
+     * @return True if the cache was successfully installed. False otherwise.
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
```
shoatman: Should cacheDirectory path be externally settable in this case?  I'm assuming this location is relative to the sandbox within which the app is running.  I'm actually wondering if all of these parameters should just be constants in the file.

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+        final String methondName = ":initialize (3 arg)";
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
+        } catch (IOException e) {
+            success = false;
+            Logger.error(
+                    TAG + methondName,
```
shoatman: typo

iambmelt:
Changing base branch to `release-candidate-0.0.10` under the assumption we want to ship this in v2 broker release + MSAL.

There will need to be additional integration PRs in MSAL and broker once this is merged. If we'd like to hold this change until  _after_ v2 goes live please chime in @shoatman @heidijinxujia @Om83

iambmelt:
I am going to close this PR and reopen with a different base branch. CR feedback will be resolved there and linked to the feedback provided in this PR.

PR #503, opened 2019-05-15
Adds shared HttpResponseCache implementation to common

iambmelt:
Summary of changes:
- Introduces class `HttpCache` to handle bootstrapping/flushing of [`HttpResponseCache`](https://developer.android.com/reference/android/net/http/HttpResponseCache).
- Added in common to support eventual inclusion in MSAL/broker

Proposed usage in ADAL:
- Create cache in [`AuthenticationContext#initialize`](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/release/1.16.2-hf1/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L144-L161)
- [Flush cache](https://developer.android.com/reference/android/net/http/HttpResponseCache#flush()) after returning ATS result to caller?
    * Ideally, in the `finalize()` method of `AuthenticationContext` but the general consensus seems to be that it's a faux pas to override `finalize()`

Proposed usage in MSAL:
- TBD

Additional considerations:
- ADAL caches Instance Discovery Metadata in-memory already after loading. The optimization benefit in ADAL really comes through across app restarts. If the app is killed/restarted before the cache's `max-age` constraint is satisfied, then the Instance Discovery Metadata request can be serviced by the cache without a network round-trip. If restarts occur less frequently and `max-age` is frequently exceeded prior to initiating the Instance Discovery Metadata request, then I would expect no performance benefit.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -63,13 +63,13 @@ public class HttpCache {
                                      @NonNull final String cacheFileName,
                                      final long maxSizeBytes) {
         final String methodName = ":initialize (3 arg)";
-        boolean success = true;
+        boolean success = false;
 
         try {
             final File httpCacheDir = new File(cacheDirectory, cacheFileName);
             HttpResponseCache.install(httpCacheDir, maxSizeBytes);
+            success = true;
         } catch (IOException e) {
-            success = false;
             Logger.error(
                     TAG + methodName,
                     "HTTP Response cache installation failed.",
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -62,7 +62,7 @@ public class HttpCache {
     public static boolean initialize(@NonNull final File cacheDirectory,
                                      @NonNull final String cacheFileName,
                                      final long maxSizeBytes) {
-        final String methondName = ":initialize (3 arg)";
+        final String methodName = ":initialize (3 arg)";
         boolean success = true;
 
         try {
@@ -71,7 +71,7 @@ public class HttpCache {
         } catch (IOException e) {
             success = false;
             Logger.error(
-                    TAG + methondName,
+                    TAG + methodName,
                     "HTTP Response cache installation failed.",
                     e
             );
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -62,7 +62,7 @@ public class HttpCache {
     public static boolean initialize(@NonNull final File cacheDirectory,
                                      @NonNull final String cacheFileName,
                                      final long maxSizeBytes) {
-        final String methodName = ":initialize (3 arg)";
+        final String methodName = ":initialize (File, Filename, Capacity)";
         boolean success = false;
 
         try {
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -43,7 +43,7 @@ public class HttpCache {
     /**
      * The default name of the HTTP/S cache on disk.
      */
-    public static final String DEFAULT_HTTP_CACHE_NAME = "com.microsoft.identitiy.http-cache";
+    public static final String DEFAULT_HTTP_CACHE_NAME = "com.microsoft.identity.http-cache";
 
     /**
      * The default size of the HTTP/S cache on disk.
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
@@ -1,0 +1,124 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param cacheDirectory The parent-directory in which the cache should reside.
+     * @param cacheFileName  The directory name for the HTTP cache.
+     * @param maxSizeBytes   The maximum allowed size of the cache. Once capacity is hit,
+     *                       entries are removed according to an LRU strategy.
+     * @return True if the cache was successfully installed. False otherwise.
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
+                                     @NonNull final String cacheFileName,
+                                     final long maxSizeBytes) {
+        final String methondName = ":initialize (3 arg)";
```
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r282318843

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
+        } catch (IOException e) {
```
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r282605114

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
+                                     @NonNull final String cacheFileName,
+                                     final long maxSizeBytes) {
+        final String methondName = ":initialize (3 arg)";
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
```
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r282605483

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r284410899

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Initializes a new {@link HttpResponseCache} based on the supplied parameters.
+     *
+     * @param cacheDirectory The parent-directory in which the cache should reside.
+     * @param cacheFileName  The directory name for the HTTP cache.
+     * @param maxSizeBytes   The maximum allowed size of the cache. Once capacity is hit,
+     *                       entries are removed according to an LRU strategy.
+     * @return True if the cache was successfully installed. False otherwise.
+     */
+    public static boolean initialize(@NonNull final File cacheDirectory,
```
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r284411342

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
```diff
@@ -0,0 +1,124 @@
...
+        final String methondName = ":initialize (3 arg)";
+        boolean success = true;
+
+        try {
+            final File httpCacheDir = new File(cacheDirectory, cacheFileName);
+            HttpResponseCache.install(httpCacheDir, maxSizeBytes);
+        } catch (IOException e) {
+            success = false;
+            Logger.error(
+                    TAG + methondName,
```
iambmelt: See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/499#discussion_r284411585

iambmelt: Resolved in 6f00685036fdcf44f45e4ca7128018af692084f5

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: Resolved in 6f00685036fdcf44f45e4ca7128018af692084f5

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: Sure

iambmelt: Resolved in f124f5b1d98ffb555720e2b65e71766d2cc8c9a6

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: Resolved in bb1ed11592fd511b4a65f53818833aedc42dd56e

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: Resolved in fd2effd6de09fbee23697428930e459adb5abcc0

common/src/main/java/com/microsoft/identity/common/internal/net/cache/HttpCache.java
iambmelt: I'm thinking this should stay as-is... I plan to invoke this new API from `AuthenticationContext` and `PublicClientApplication` in the following way...

```java
HttpCache.initialize(context.getCacheDir());
```

Since a `Context` is required to get access to the `cacheDir`, I'd like for us to not _have_ to pass in an entire context when only the target directory is required.

iambmelt:
Ready for another round of review - feedback resolved/replied to accordingly.

Please note the base/target branch has changed to point to `release-candidate-0.0.10`

@Om83 @shoatman @heidijinxujia

iambmelt:
I should mention, `/token` responses are caching-safe, as they are transmitted with: 
`Cache-Control: no-cache, no-store`

PR #504, opened 2019-05-16
Pass claims in non-joined acquireTokenSilent call

rpdome:
Found this issue in the Authenticator app bug bash.

Here's the scenario.
1. Authenticator app NGC registration requires deviceid + ngcmfa claims.
2. NGC registration was successfully performed.
3. WPJ was unregistered.
4. Tried NGC registration again. Since WPJ is gone, it would fall back to the non-joined flow.
5. The non-joined flow will try to use existing RT. Since it's not passing the claim this time, Broker will return an AT that AuthApp can't use. Therefore the flow will fail.


rpdome:
Thanks!

PR #513, opened 2019-06-03
Fix an issue where V1 Id token needs to be returned

kreedula:


common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
```diff
@@ -323,6 +323,41 @@ public void saveTokensV1Compat() throws ClientException {
         assertEquals(defaultTestBundleV1.mGeneratedIdToken, ids.get(0));
     }
 
+    @Test
+    public void saveTokensOverloadMethodV1Compat() {
+        // This test asserts that if an IdToken is returned in the v1 format (broker cases),
+        // it is saved property.
```
iambmelt: super nit `properly`

kreedula:
Issue for unit test : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/514

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -25,6 +25,7 @@
 import android.content.Context;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.annotation.VisibleForTesting;
```
iambmelt: We can remove this until a test is added

PR #515, opened 2019-06-04
Add the adapter between cache record list and bundle

heidijinxujia:


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -392,38 +393,40 @@ private ServiceException getServiceException(@NonNull final BrokerResult brokerR
...
-     * @return List of AccountRecord
+     * @return List of CacheRecord
      */
-    public static List<AccountRecord> getAccountRecordListFromBundle(@NonNull final Bundle bundle) {
+    public static List<ICacheRecord> getCacheRecordListFromBundle(@NonNull final Bundle bundle) {
         final ArrayList<String> accountsList = bundle.getStringArrayList(BROKER_ACCOUNTS);
-        final List<AccountRecord> result = new ArrayList<>();
-        if (accountsList == null) {
+        final List<ICacheRecord> result = new ArrayList<>();
+        if (accountsList == null || accountsList.isEmpty()) {
```
iambmelt: `null` here seems distinctly different than empty....

Receiving `null` in this case would signal a bug, correct? If yes, let's `throw` some meaningful Exception.

Empty to me says: no accounts were found

iambmelt: We could change this check to just be:

```java
if (accountsList == null) {
    // ...
}
```

Since the `foreach` loop will basically skip over an empty `List`

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -322,36 +323,39 @@ private ServiceException getServiceException(@NonNull final BrokerResult brokerR
...
+        ArrayList<String> cacheRecordString = new ArrayList<>();
+        if (records != null && !records.isEmpty()) {
+            for (ICacheRecord record : records) {
+                final String recordInGson = new Gson().toJson(record, CacheRecord.class);
+                cacheRecordString.add(recordInGson);
+            }
         }
 
-        resultBundle.putStringArrayList(BROKER_ACCOUNTS, accountRecordString);
+        resultBundle.putStringArrayList(BROKER_ACCOUNTS, cacheRecordString);
```
iambmelt: Can we make this implementation match: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/520/files#diff-44f2260eafcc843207898d147d7285beR388 ?

PR #518, opened 2019-06-07
Adding negotiate challenge support to WebViews

dcale:
There is a hidden feature in Android WebViews that can enable them to handle Negotiate challenges (see https://chromium.googlesource.com/chromium/src/+/master/docs/webview_policies.md). 

This pull request aims at enabling exactly that. The negotiate challenge support is especially helpful for federated logins leveraging spnego authentication. This would enable intranet SSO for all MS Apps using the library (hopefull the entire Native Office Android Apps). By default the features will remain disabled (thus not changing current behaviour), only if an Administrator populates the added managed configurations, the feature is enabled. 

These configurations are also available in the native MS Edge application.

### DIFF
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -4,7 +4,12 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+
     <application>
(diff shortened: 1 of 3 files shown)


iambmelt:
Adding @shoatman as a reviewer

@dcale Because this change would expose dependents of `common` to [remote/managed configurations](https://developer.android.com/work/managed-configurations) I suspect this is going to require extra review relative to threat modeling and functionality across apps integrated with ADAL already

PR #519, opened 2019-06-11
Telemetry

heidijinxujia:
[In progress] updating the description and proposal now... 
common_telemetry_flow_example (1)

- Telemetry class
- TelemetryConfiguration class -> move the initialization into into PCAconfiguration.
- TelemetryContext class
- Properties class 
- Telemetry Dispatcher
- helpers methods into StringUtils class

Events:
1. BaseEvent
2.ApiStartEvent
3. ApiEndEvent
4. CacheStartEvent
5. CacheEndEvent
6. HttpStartEvent
7. HttpEndEvent
8. UiStartEvent
9. UiEndEvent

Observers
1. interface
2. aggregated observer (for MATS)
3. default observer

Rules
1. TelemetryPiiOiiRules
2.  TelemetryAggregatredRules

TODO:
- code review
- unit tests
- integration tests
- plug in the events into the calls.


common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+
+        public Builder() {
+        }
+
+        public Builder withContext(final Context context) {
+            if (context == null) {
+                throw new IllegalArgumentException("Context must not be null.");
+            }
+            if (!hasPermission(context, Manifest.permission.INTERNET)) {
+                throw new IllegalArgumentException("INTERNET permission is required.");
```
shoatman: Why is this permission required for Telemetry?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+            if (mContext == null) {
+                throw new IllegalArgumentException("Application context must not be null.");
+            }
+
+            mTelemetryContext = TelemetryContext.create(mContext);
+
+            try {
+                String packageName = context.getPackageName();
+                int flags = context.getPackageManager().getApplicationInfo(packageName, 0).flags;
+                mIsDebugging = (flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;
```
shoatman: Now I understand how this intended to work.  Thanks!

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
```diff
@@ -0,0 +1,204 @@
...
+        this.put(TELEMETRY_KEY_APPLICATION_VERSION, parameters.getApplicationVersion());
+        this.put(TELEMETRY_KEY_SDK_NAME, parameters.getSdkType().name());
+        this.put(TELEMETRY_KEY_SDK_VERSION, parameters.getSdkVersion());
+        this.put(TELEMETRY_KEY_CLAIM_REQUEST, StringUtil.isEmpty(
+                parameters.getClaimsRequestJson())? TELEMETRY_VALUE_NO : TELEMETRY_VALUE_YES
+        );
+        this.put(TELEMETRY_KEY_REDIRECT_URI, parameters.getRedirectUri());
+        this.put(TELEMETRY_KEY_CLIENT_ID, parameters.getClientId());
+        this.put(TELEMETRY_KEY_SCOPE_SIZE, String.valueOf(parameters.getScopes().size()));
+        this.put(TELEMETRY_KEY_SCOPE, parameters.getScopes().toString());
```
shoatman: Do these properties align with what iOS is planning to emit?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheStartEvent.java
```diff
@@ -0,0 +1,57 @@
...
+        put(TELEMETRY_KEY_IS_FRT, String.valueOf(isMrrt));
+        return this;
+    }
+    public CacheStartEvent isRt(final boolean isRt) {
+        put(TELEMETRY_KEY_IS_FRT, String.valueOf(isRt));
+        return this;
+    }
+
+    public CacheStartEvent putWipeApp(final boolean appWiped) {
+        put(TELEMETRY_KEY_WIPE_APP, String.valueOf(appWiped));
```
shoatman: What is this event?  Is this clear entire cache?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/HttpStartEvent.java
```diff
@@ -0,0 +1,79 @@
...
+        this.put(TELEMETRY_KEY_HTTP_ERROR_DOMAIN, errorDomain);
+        return this;
+    }
+
+    public HttpStartEvent isNetworkConnected(boolean isConnected) {
+        this.put(TELEMETRY_KEY_NETWORK_CONNECTION, String.valueOf(isConnected));
+        return this;
+    }
+
+    public HttpStartEvent isPowerOptimizationOn(boolean isDozed) {
```
shoatman: Nice.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
```diff
@@ -0,0 +1,105 @@
...
+import com.microsoft.identity.common.internal.telemetry.TelemetryEventStrings;
+import com.microsoft.identity.common.internal.util.StringUtil;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.microsoft.identity.common.internal.telemetry.TelemetryEventStrings.*;
+
+final public class TelemetryPiiOiiRules {
```
rpdome: Would you mind adding a comment here describing what PII and OII stand for?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
```diff
@@ -0,0 +1,105 @@
...
+            TELEMETRY_KEY_APPLICATION_VERSION
+    };
+
+    private TelemetryPiiOiiRules() {
+        piiPropertiesSet = new HashSet<>(Arrays.asList(piiArray));
+        oiiPropertiesSet = new HashSet<>(Arrays.asList(oiiArray));
+    }
+
+    public static TelemetryPiiOiiRules getInstance() {
+        if (sInstance == null) {
```
rpdome: This might need to be made synchronized.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryAggregationRules.java
```diff
@@ -0,0 +1,47 @@
...
+    private Set<String> aggregatedPropertiesSet;
+
+    final private String aggregatedArray[] = { };
+
+    private TelemetryAggregationRules() {
+        aggregatedPropertiesSet = new HashSet<>(Arrays.asList(aggregatedArray));
+    }
+
+    public TelemetryAggregationRules getInstance() {
+        if (sInstance == null) {
```
rpdome: Should this be synchronized?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+        if (null == mObservers) {
+            mObservers = new LinkedList<>();
+        }
+
+        mObservers.add(observer);
+    }
+
+    public synchronized void removeObserver(final ITelemetryObserver observer) {
+        if (null == observer || null == mObservers) {
+            return;
```
rpdome: nit: add logging point?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+
+    private Map<String, String> applyPiiOiiRule(final Map<String, String> properties) {
+        if (mDefaultConfiguration.isPiiEnabled()) {
+            Logger.warn(TAG, "Telemetry PII/OII is enabled by the developer.");
+            return properties;
+        }
+
+        final Map<String, String> nonPiiProperties = new HashMap<>();
+        for (Map.Entry<String, String> entry : properties.entrySet()) {
+            if (!TelemetryPiiOiiRules.getInstance().isPiiOrOii(entry.getKey())) {
```
rpdome: Do these 'rules' need to be a singleton? Do we plan to use it somewhere else?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+        return singleton;
+    }
+
+    /**
+     * This is for getting instance of Telemetry
+     **/
+    public static Telemetry getInstance() {
+        return singleton;
```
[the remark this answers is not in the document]
rpdome: I agree with Shane, but I personally couldn't think of a way to resolve potential race condition cleanly either.

Maybe we could have default value of the singleton to be an object with 'mIsTelemetryEnabled = false'?

Meaning that until it's properly initialized (which may takes time), if somehow some flow tries to log telemetry event, it wouldn't be recorded.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+
+            mTelemetryContext = TelemetryContext.create(mContext);
+
+            try {
+                String packageName = context.getPackageName();
+                int flags = context.getPackageManager().getApplicationInfo(packageName, 0).flags;
+                mIsDebugging = (flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;
+            } catch (final PackageManager.NameNotFoundException exception) {
+                Logger.warn(TAG, "The application is not found from PackageManager.");
+                mIsDebugging = true;
```
rpdome: In which scenario would this happen? Can this happen in prod? Why are we setting debugging = true in this case?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Field names used for serialization by Gson.
+     */
+    public static final class SerializedNames {
+        public static final String PII_ENABLED = "pii_enabled";
+        public static final String NOTIFY_ON_FAILTURE_ONLY = "notify_on_failure_only";
+        public static final String DEBUG_ENABLED = "debug_enabled";
+    }
+
+
```
rpdome: nit: extra line.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -0,0 +1,112 @@
...
+    TelemetryContext(ConcurrentHashMap<String, String> delegate) {
+        super(delegate);
+    }
+
+    /**
+     * Create a new {@link TelemetryContext} instance filled in with information from the given {@link
+     * Context}. The {@link Telemetry} client can be called from anywhere, so the returned instances
+     * is thread safe.
+     */
+    static synchronized TelemetryContext create(final Context context) {
```
rpdome: Just curious. Why does this need to be synchronized?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
```diff
@@ -0,0 +1,204 @@
...
+public class ApiStartEvent extends BaseEvent {
+    private static final String TAG = ApiStartEvent.class.getSimpleName();
+
+    public ApiStartEvent() {
+        super();
+        names(TELEMETRY_EVENT_API_EVENT_START);
+    }
+
+    public ApiStartEvent putProperties(OperationParameters parameters) {
+        this.put(TELEMETRY_KEY_AUTHORITY, parameters.getAuthority().toString());
```
rpdome: I'd personally wrap everything with try-cache, and log if there's any exception. (Mainly to cache unexpected nullpointerException or casting exception). My stance is that telemetry error shouldn't crash the app.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/HttpEndEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public HttpEndEvent() {
+        super();
+        names(TELEMETRY_EVENT_HTTP_REQUEST_END);
+    }
+
+    public HttpEndEvent putResponse(final HttpResponse httpResponse) {
+        put(TELEMETRY_KEY_HTTP_RESPONSE_CODE, String.valueOf(httpResponse.getStatusCode()));
+        //TODO discuss what kind of telemetry need to be refined from response body and header.
+        return this;
```
rpdome: FYI, the interrupt flow could contain a PRT as part of the request header.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    void putDevice() {
+        put(DEVICE_MANUFACTURER_KEY, Build.MANUFACTURER);
+        put(DEVICE_MODEL_KEY, Build.MODEL);
+        put(DEVICE_NAME_KEY, Build.DEVICE);
+    }
+
+    void putLibrary() {
+        put(LIBRARY_NAME_KEY, "common-android");
```
kreedula: Can we declare a constant for this "common-android"

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -0,0 +1,112 @@
...
+        super(delegate);
+    }
+
+    /**
+     * Create a new {@link TelemetryContext} instance filled in with information from the given {@link
+     * Context}. The {@link Telemetry} client can be called from anywhere, so the returned instances
+     * is thread safe.
+     */
+    static synchronized TelemetryContext create(final Context context) {
+        TelemetryContext telemetryContext = new TelemetryContext(new ConcurrentHashMap<String, String>());
```
kreedula: Nit : final

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final String DEVICE_ID_KEY = "device_id";
+    // Library
+    private static final String LIBRARY_NAME_KEY = "library_name";
+    private static final String LIBRARY_VERSION_KEY = "library_version";
+    // OS
+    private static final String OS_NAME_KEY = "os_name";
+    private static final String OS_VERSION_KEY = "os_version";
+    private static final String TIMEZONE_KEY = "timezone";
+
+    TelemetryContext(ConcurrentHashMap<String, String> delegate) {
```
kreedula: Nit : final, annotations

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -0,0 +1,169 @@
...
+    public static final String TELEMETRY_EVENT_API_EVENT_START = "api_start_event";
+    public static final String TELEMETRY_EVENT_API_EVENT_END = "api_end_event";
+
+    public static final String TELEMETRY_EVENT_CACHE_EVENT_START = "cache_start_event";
+    public static final String TELEMETRY_EVENT_CACHE_EVENT_END = "cache_end_event";
+
+    public static final String TELEMETRY_EVENT_UI_EVENT_START = "ui_start_event";
+    public static final String TELEMETRY_EVENT_UI_EVENT_END = "ui_end_event";
+
+    public static final String TELEMETRY_EVENT_HTTP_REQUEST_START = "http_start_event";
```
kreedula: can these be changed to match the constants name for consistency ?  like "http_start_event" - > "http_request_start_event"

Feedback applies for the constants below

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiEndEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class ApiEndEvent extends BaseEvent {
+    private static final String TAG = ApiStartEvent.class.getSimpleName();
+
+    public ApiEndEvent() {
+        super();
+        names(TELEMETRY_EVENT_API_EVENT_END);
+    }
+
+    public ApiEndEvent putResult(@NonNull final AcquireTokenResult result) {
```
kreedula: Shouldn't this be `public ApiEndEvent putResult(@NonNull final ILocalAuthenticationResult result) {`

Beacuse if `result.getSucceeded()` we parse Auth and Token response in `ApiDispatcher` and create a `BaseException` . From there we can call `putException` method.

Otherwise here we just add ` put(TELEMETRY_KEY_IS_SUCCESSFUL, result.getSucceeded().toString());` and nothing else if `getLocalAuthenticationResult` is null

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
```diff
@@ -0,0 +1,204 @@
...
+        if (parameters instanceof AcquireTokenOperationParameters) {
+            this.put(
+                    TELEMETRY_KEY_USER_AGENT,
+                    ((AcquireTokenOperationParameters) parameters).getAuthorizationAgent().name()
+            );
+            this.put(TELEMETRY_KEY_LOGIN_HINT,
+                    ((AcquireTokenOperationParameters) parameters).getLoginHint()
+            );
+            this.put(TELEMETRY_KEY_REQUEST_QUERY_PARAMS,
+                    String.valueOf(((AcquireTokenOperationParameters) parameters).getExtraQueryStringParameters().size())
```
kreedula: This can throw a null pointer if extraQP are not set right?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/BaseEvent.java
```diff
@@ -0,0 +1,79 @@
...
+package com.microsoft.identity.common.internal.telemetry.events;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import com.microsoft.identity.common.internal.logging.DiagnosticContext;
+import com.microsoft.identity.common.internal.telemetry.Properties;
+
+public class BaseEvent extends Properties {
+    public static final String OCCUR_TIME = "start_time";
```
kreedula: rename constant to `START_TIME`

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/BaseEvent.java
```diff
@@ -0,0 +1,79 @@
...
+        return this;
+    }
+
+    /**
+     * Put the event occurring time into the properties map.
+     *
+     * @param eventStartTime Long of the event start time. If null, then put the current time as the start time.
+     * @return the event object
+     */
+    public BaseEvent occurs(@Nullable Long eventStartTime) {
```
kreedula: Can we rename to `startEvent` may be ?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryAggregationRules.java
```diff
@@ -0,0 +1,47 @@
...
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.telemetry.rules;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+//TODO Place holder
+public class TelemetryAggregationRules {
+    private TelemetryAggregationRules sInstance;
```
kreedula: Is this supposed to be singleton ? if yes , this needs to `static` and instantiated in `synchronized` block below

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return sInstance;
+    }
+
+    /**
+     * @param propertyName String of propertyName {@link TelemetryEventStrings}
+     * @return true if the property belongs to Personally identifiable information. False otherwise.
+     */
+    public boolean isPii (final String propertyName) {
```
Om83: nit: Oii and Pii should use capital I because they are abbreviation. So it should be PII and OII

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryAggregationRules.java
```diff
@@ -0,0 +1,47 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.telemetry.rules;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+//TODO Place holder
```
Om83: This is placeholder for what?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheStartEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    public CacheStartEvent isMrrt(final boolean isMrrt) {
+        put(TELEMETRY_KEY_IS_FRT, String.valueOf(isMrrt));
+        return this;
+    }
+    public CacheStartEvent isRt(final boolean isRt) {
+        put(TELEMETRY_KEY_IS_FRT, String.valueOf(isRt));
+        return this;
+    }
+
+    public CacheStartEvent putWipeApp(final boolean appWiped) {
```
Om83: Name of the method is not very clear, please rename to reflect what is happening.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheEndEvent.java
```diff
@@ -0,0 +1,59 @@
...
+        put(TELEMETRY_KEY_MRRT_STATUS, mrrtStatus);
+        return this;
+    }
+
+    public CacheEndEvent putFrtStatus(final String frtStatus) {
+        put(TELEMETRY_KEY_FRT_STATUS, frtStatus);
+        return this;
+    }
+
+    public CacheEndEvent putSpeInfo(final String speInfo) {
```
Om83: Sorry what is SPE? Add comments?

Om83:
@heidijinxujia , I see you have Unit Tests and Integrations tests in the todo list. Given that it is a big PR we need to make sure that we test the changes before merging in the dev. What's your plan?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
heidijinxujia: This is for the future need of the MATS upload inside common... I will remove it for now.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
iambmelt: The Android style guide says otherwise - [acronyms should be treated as words](https://source.android.com/setup/contribute/code-style#treat-acronyms-as-words)

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
heidijinxujia: It contains what iOS already has and I also added other info into the Api events based our previous icm experience, including
TELEMETRY_KEY_AUTHORITY
TELEMETRY_KEY_AUTHORITY_TYPE
TELEMETRY_KEY_CLAIM_REQUEST
TELEMETRY_KEY_REDIRECT_URI
TELEMETRY_KEY_SCOPE_SIZE
TELEMETRY_KEY_SCOPE
TELEMETRY_KEY_REQUEST_QUERY_PARAMS
TELEMETRY_KEY_IS_FORCE_REFRESH
TELEMETRY_KEY_BROKER_PROTOCOL_VERSION

Please let me know which you want to remove or add.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheStartEvent.java
heidijinxujia: I think this is for the share device scenario, which removing the entire cache when signing out. p.s. Here is trying to align with iOS properties.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+
+        return nonPiiProperties;
+    }
+
+    /**
+     * API for creating {@link Telemetry} instances.
+     */
+    public static class Builder {
```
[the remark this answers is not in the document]
heidijinxujia: Because the Telemetry instance needs to be initialized with Context and Configuration as parameters. And I prefer it use Builder instead of an initialize(context, telemetryConfiguration) and having something like
```java
Telemetry telemetry = Telemetry.getInstance();
telemetry.initialize(context, configuration);
```

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    void addDeviceInfo() {
+        put(TELEMETRY_KEY_DEVICE_MANUFACTURER, Build.MANUFACTURER);
+        put(TELEMETRY_KEY_DEVICE_MODEL, Build.MODEL);
+        put(TELEMETRY_KEY_DEVICE_NAME, Build.DEVICE);
+    }
+
+    void addOsInfo() {
+        put(TELEMETRY_KEY_OS_NAME, TELEMETRY_VALUE_OS_NAME);
```
heidijinxujia: Add Build.version.securityPatch

common/src/main/java/com/microsoft/identity/common/internal/telemetry/observers/ITelemetryDefaultObserver.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The default telemetry observer interface which upload raw telemetry data.
+ */
+public interface ITelemetryDefaultObserver extends ITelemetryObserver<List<Map<String, String>>> {
+    @Override
+    void upload(List<Map<String, String>> telemetryData);
```
heidijinxujia: onReceive()

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,295 @@
...
+    public synchronized void addObserver(final ITelemetryObserver observer) {
+        if (null == observer) {
+            throw new IllegalArgumentException("Telemetry Observer instance cannot be null");
+        }
+
+        // check to make sure we're not already dispatching elsewhere
+        if (null == mObservers) {
+            mObservers = new LinkedList<>();
+        }
+
```
heidijinxujia: extra lines

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,283 @@
...
+    public synchronized void removeObserver(final ITelemetryObserver observer) {
+        if (null == observer || null == mObservers) {
+            return;
+        }
+
+        mObservers.remove(observer);
+    }
+
+    public List<ITelemetryObserver> getObservers() {
```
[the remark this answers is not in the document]
heidijinxujia: Add identifier for the telemetry observers.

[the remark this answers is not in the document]
heidijinxujia: memory leak check for activity and fragement checks.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,295 @@
...
+        return mObservers;
+    }
+
+    /**
+     * Emit the event into the telemetry raw data map.
+     *
+     * @param event BaseEvent object
+     * @return the event reference for future properties modification.
+     */
+    public static BaseEvent emit(final BaseEvent event) {
```
heidijinxujia: change return to void.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,295 @@
...
+            mObservers = new LinkedList<>();
+        }
+    }
+
+    /**
+     * Prepares instance using builder.
+     **/
+    private static Telemetry prepareInstance(Builder builder) {
+        if (sTelemetryInstance == null) {
+            synchronized (Telemetry.class) {
```
iambmelt: Either the null check should move into the `synchronized` block or the whole method should be synchronized. If two threads are competing for this method, synchronizing _after_ the null check is too late

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -0,0 +1,295 @@
...
+        return sTelemetryInstance;
+    }
+
+    /**
+     * This is for getting instance of Telemetry
+     **/
+    public static Telemetry getInstance() {
+        // If sTelemetryInstance is not initialized, telemetry will be disabled.
+        if (sTelemetryInstance == null) {
+            synchronized (Telemetry.class) {
```
iambmelt: Same here - synchronizing _after_ the null check is too late

iambmelt: I would declare the whole method synchronized - intent is clearer

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Properties.java
```diff
@@ -0,0 +1,78 @@
...
+        if (mProperties == null) {
+            mProperties = appendProperties.getProperties();
+        } else {
+            mProperties.putAll(appendProperties.getProperties());
+        }
+
+        return this;
+    }
+
+    public ConcurrentHashMap<String, String> getProperties() {
```
iambmelt: Two points here:
1. Do we need to expose the concrete class `ConcurrentHashMap`? Or would `Map` suffice?
2. Should this be immutable?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -0,0 +1,189 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.telemetry;
+
+public final class TelemetryEventStrings {
```
iambmelt: One suggestion to keep this class organized - using static inner classes might force ppl to keep this organized when added new constants.

Something like....

```java
public final class TelemetryConstants {

    public static final class App {
        public static final String BUILD = "app_build";
    }

    public static final class Device {
         public static final String MANUFACTURER = "device_manufacturer";
         public static final String MODEL = "device_model";
         public static final String NAME = "device_name";
         public static final String TIMEZONE = "time_zone";
    }

    public static final class Os {
        public static final String NAME = "os_name";
    }

    public static final class Events {
        public static final String API_EVENT = "api_event";
        public static final String API_EVENT_START = API_EVENT + "_start";
        public static final String API_EVENT_END = API_EVENT + "_end";
    }

    public static final class Keys {
        public static final String EVENT_NAME = "event_name";
        public static final String EVENT_TYPE = "event_type";
    }
}
```

Everything nice and sorted, avoids the hungarian notation tarpit, works with intellisense, less typing, easy to keep from becoming a giant dumping ground of strings

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public TelemetryAggregationAdapter(@NonNull final ITelemetryAggregatedObserver observer) {
+        mObserver = observer;
+    }
+
+    public ITelemetryAggregatedObserver getObserver() {
+        return mObserver;
+    }
+
+    public void process(List<Map<String, String>> rawData) {
```
iambmelt: `@NonNull`

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+            aggregatedData.putAll(applyAggregationRule(event));
+        }
+
+        aggregatedData.put(TELEMETRY_KEY_START_TIME, String.valueOf(apiStartTime));
+        aggregatedData.put(TELEMETRY_KEY_END_TIME, String.valueOf(apiEndTime));
+
+        mObserver.upload(aggregatedData);
+    }
+
+    private Map<String, String> applyAggregationRule(final Map<String, String> properties) {
```
iambmelt: `@NonNull`

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryDefaultAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public TelemetryDefaultAdapter(@NonNull final ITelemetryDefaultObserver observer) {
+        mObserver = observer;
+    }
+
+    public ITelemetryDefaultObserver getObserver() {
+        return mObserver;
+    }
+
+    public void process(List<Map<String, String>> rawData) {
```
iambmelt: Presumably `@NonNull` but can't tell

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
iambmelt: I don't like plastering over bugs like this. I'd rather see unit tests and just have confidence that everything works. If we shipped an NPE that caused clients not to send telemetry how would we know?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
```diff
@@ -0,0 +1,198 @@
...
+
+        if (parameters instanceof BrokerAcquireTokenSilentOperationParameters) {
+            //TODO when integrate the telemetry with broker.
+        }
+
+        return this;
+    }
+
+    public ApiStartEvent authority(@NonNull final String authority) {
+        super.put(TELEMETRY_KEY_AUTHORITY, sanitizeUrlForTelemetry(authority));
```
iambmelt: This class seems to arbitrarily switch between calling `super.put` and `this.put` - why not

 `return put(TELEMETRY_KEY_AUTHORITY, sanitizeUrlForTelemetry(authority));`

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheEndEvent.java
iambmelt: `Secure Productive Enterprise Ring` which I don't think _really_ clarifies... 😅 

This is the data coming from `x-ms-clitelem` header, (corpnet) spec is [here](https://microsoft.sharepoint.com/:w:/r/teams/office_rdx/_layouts/15/Doc.aspx?sourcedoc=%7BF604EC14-7A7D-4490-A3EA-617E1E0FE7D3%7D&file=Microsoft-wide%20SPE%20%E2%80%9Cring%E2%80%9D%20alignment.docx&action=default&mobileredirect=true&DefaultItemOpen=1)

common/src/main/java/com/microsoft/identity/common/internal/telemetry/observers/ITelemetryDefaultObserver.java
iambmelt: I personally like `onEventsReceived` but that's just me

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -405,12 +413,18 @@ public void run() {
                     handler.post(new Runnable() {
                         @Override
                         public void run() {
+                            // end api event
+                            Telemetry.emit(
+                                    new ApiEndEvent()
```
rpdome: What's the reason we're putting  ApiEndEvent() here?

Asking because it seems very 'out of place', as in I'm not seeing any other ApiEndEvent() and ApiStartEvent() in other functions. Are they being handled in other layers? If that's the case, shall we move this event up there (or bring all of them down here)?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -405,12 +413,18 @@ public void run() {
...
+                                            .putException(finalException)
+                                            .putApiId("acquire_token_silent")
```
rpdome: nit: ApiId should be a constant.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
rpdome: Alternatively, we could just use this as a class variable instead of having it as singleton.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -0,0 +1,189 @@
...
+    public static final String TELEMETRY_VALUE_SUCCEEDED = "succeeded";
+    public static final String TELEMETRY_VALUE_FAILED = "failed";
+    public static final String TELEMETRY_VALUE_CANCELLED = "cancelled";
+    public static final String TELEMETRY_VALUE_UNKNOWN = "unknown";
+    public static final String TELEMETRY_VALUE_AUTHORITY_AAD = "aad";
+    public static final String TELEMETRY_VALUE_AUTHORITY_ADFS = "adfs";
+    public static final String TELEMETRY_VALUE_AUTHORITY_B2C = "b2c";
+
+    //Telemetry API ID
+    public static final String API_BROKER_ACQUIRE_TOKEN_INTERACTIVE = "201";
```
rpdome: What was the reason of sending number instead of actual strings? (as this would require a mapping on the other side)

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -0,0 +1,115 @@
...
+
+            final long eventOccurTime = Long.parseLong(event.get(TELEMETRY_KEY_OCCUR_TIME));
+            if (eventName.equalsIgnoreCase(TELEMETRY_EVENT_API_START_EVENT)
+                    && (apiStartTime == -1
+                    || eventOccurTime < apiStartTime)) {
+                apiStartTime = eventOccurTime;
+            }
+
+            if (eventName.equalsIgnoreCase(TELEMETRY_EVENT_API_END_EVENT)
+                    && (apiEndTime == -1
```
rpdome: we probably won't need this -1 check (as eventOccurTime should always be bigger than -1).

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
heidijinxujia: Start and End events do not need to be in the same layer. This is the flexibility of the design here. It emits a start event before sending a request and emits an end event when either getting a result or catching an exception, like here. As the exception could be thrown anywhere in the nested calls and I do not want to add try-catch and emit it everywhere. This is the right place to emit the api end event when the api call ends with any exception throwing in the middle.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
heidijinxujia: I've talked with Shane before that one aim here is to make Telemetry calls as "quiet" as possible and not break any request because of the telemetry calls. We can follow-up tmr morning if this is still a concern.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
heidijinxujia: The Pii/Oii rules here is only used by the Telemetry here.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
heidijinxujia: Shorten the blob we sent. This is the same with ADAL telemetry as well as the iOS telemetry.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiEndEvent.java
heidijinxujia: I saw the return of the AT and ATS result is in type AcquireTokenResult

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiStartEvent.java
heidijinxujia: @rpdome  try-cache will stop the running of the following code. And we want to collect useful data as much as possible.

heidijinxujia: @iambmelt Agree.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/BaseEvent.java
heidijinxujia: The event here is a one-time and atomic concept. There is no time window like starting time and ending time for the event. Thus, "occurs" would be a better naming here.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
kreedula: Agree the Telemetry design gives you the flexibility . But why is there only one emit event here on exception ? Ideally I would see API end events for success case as well here in APIDispatcher (both acquire token and acquire token silent). Any reason why we are not doing it?

kreedula: We can delegate this discussion to later as well considering we want to merge this soon, since this is about where to emit the events.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/CacheStartEvent.java
heidijinxujia: Here is trying to align with iOS properties. I think this is for the share device scenario, which removing the entire cache when signing out.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
heidijinxujia: Working on this.

heidijinxujia: Done.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/ApiEndEvent.java
kreedula: yeah let's sort this out the after PR is in , primary feedback is acquiretoken result doesn't have the exception info to pass the telemetry

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
heidijinxujia: Let me add it.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
heidijinxujia: Maybe some confusion here. The plug-in here is not done yet. This was for testing purpose before.

Issue #524, opened 2019-06-14
MicrosoftTokenRequest needs to differentiate between scopes/extra scopes


shoatman:
When a developer is asking for consent to multiple resources the token endpoint needs to receive a list of scopes that only apply to a single resource.  If the authorization code represents consent for multiple resources (scopes associated within more than one API) then the client must specify via the scope parameter a list of scopes belonging to a single resource.

shoatman:
Fixed.

PR #527, opened 2019-06-17
Updates to fix non ms graph token requests and caching of result

shoatman:


shoatman:
fix #524

common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -106,6 +114,21 @@ public AccessTokenRecord createAccessToken(
...
+    /**
+     * Returns the correct target based on whether the default scopes were returned or not
+     * @param responseScope
+     * @return
+     */
+    private String getTarget(String responseScope){
+        if(responseScope.contains(AuthenticationConstants.OAuth2Scopes.OPEN_ID_SCOPE)){
+            return responseScope;
+        }else{
+            return responseScope + " " + AuthenticationConstants.OAuth2Scopes.OPEN_ID_SCOPE
```
heidijinxujia: I have one dumb question. The scopes openid, profile and offline_access are returned when talking to Graph. Is it required to have these scopes when requesting for other resources? If so, is it a server bug that these scopes are not returned? Just wondering why we need to make additional modification for the server response.

shoatman: Not dumb at all.  I find the server behavior confusing to say the least.  My understanding is that offline_access scope is what ensures that a refresh token is returned;  openid scope is what ensures that an id_token is returned.... and profile ensures that the id_token contains the claims that you are expecting.  

I believe these should be returned in this scenario; however I doubt the service can make this change now... as it would be a breaking change.

PR #528, opened 2019-06-17
Updating Interactive redirect that detects the Company Portal enrollment

titorodi:
UI to check specifically for Teams Phone app. This is due to Company
Portal and Microsoft Intune COBO apps being deployed together by default
for 1906, while there are still 4.4 IPPhone devices without a browser
but could still have MDM CA Policy.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -217,6 +217,9 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
                 && packageHelper.isPackageInstalledAndEnabled(AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME)
                 && AuthenticationConstants.Broker.IPPHONE_APP_SIGNATURE.equals(
                 packageHelper.getCurrentSignatureForPackage(AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME))) {
+            // TODO: This flow should really check if the Microsoft Intune or the Company Portal apps are installed,
+            // which is the correct client app to launch depending on the enrollment, and launch that app, to permanently skip the browser.
+            // Also it must handle 3rd party MDMs as appropriate, which will depend on the browser flow determining the authority.
             Logger.verbose(TAG + methodName, "It is a device CA request on IPPhone. Company Portal is installed.");
             try {
                 Logger.verbose(TAG + methodName, "Sending intent to launch the CompanyPortal.");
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -213,8 +213,10 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
         final String methodName = "#processWebsiteRequest";
         final PackageHelper packageHelper = new PackageHelper(getActivity().getApplicationContext());
         if (url.startsWith(AuthenticationConstants.Broker.BROWSER_DEVICE_CA_URL)
-                && packageHelper.isPackageInstalledAndEnabled(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)) {
-            Logger.verbose(TAG + methodName, "It is a device CA request. Company Portal is installed.");
+                && packageHelper.isPackageInstalledAndEnabled(AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME)
```
kreedula: This would break the existing scenario to check if Company Portal is installed or not right ? i.e. in mobile devices?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -213,8 +213,10 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
...
+                && AuthenticationConstants.Broker.IPPHONE_APP_SIGNATURE.equals(
```
heidijinxujia: Is Teams IP Phones a separate app? According to the javadoc here, it seems when the "com.microsoft.skype.teams.ipphone" installed, you should also check the company portal installation.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
heidijinxujia: +1

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
titorodi: Hm. Fair. The only supported scenario is their app on their phone with the CP. But I should check for both. Thank you! I'll add that back in.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
titorodi: Well, this was added to support their phone in general, and was an optimization for this link for the CP if their phone was not installed.

We don't have a better way of detecting that we're on their phone than by their app is installed on the device.

As long as there is a browser, this MDM CA link works by going to Intune Web Portal, and there is a link to the Google Play store there. So we'll 'regress' back to an extra click for Company Portal flows, we'll fix Microsoft Intune COBO flows from going to the wrong app, and we'll continue to work on Sakurai phones.

PR #529, opened 2019-06-17
TSL Integration

iambmelt:
See https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/656


common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,101 @@
...
+public class TokenShareUtility implements ITokenShareInternal {
+
+    private MsalOAuth2TokenCache mTokenCache;
+
+    public TokenShareUtility(@NonNull final MsalOAuth2TokenCache cache) {
+        mTokenCache = cache;
+    }
+
+    /*
+    Design questions:
```
iambmelt: @shoatman Some design questions I thought you and I should sync on -- I posted my thoughts in-line, lmk what you think.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,105 @@
...
+        1. When saving these tokens, do I persist them for my own clientId it so that this app can
+           use it? Or should I try to parse the clientId out of the input and save it under that?
+           Should I add a FoCI lookup that is agnostic of clientIds completely?
+
+           Brian's answer:
+               Add parsing to determine the _other_ app's client id. Parse it and save under that.
```
iambmelt: ✔️

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,105 @@
...
+               the RT of another client if the current one is FOCI.
+
+        2. If I already have an IdToken/RT for the current user and TSL tries to save a new one,
+           do I delete the one I have so that I maintain the 1 RT per user rule?
+
+           Brian's answer:
+               I think 'yes', maintain 1 RT per user. I don't want to potentially delete a good token for a bad one though...
+               What about...
+               If you already have tokens for this user and they're not expired, do nothing.
+               If you already have tokens for this user and they're expired, save the new tokens (see next question).
```
iambmelt: This section basically ✔️ 

> If you already have tokens for this user and they're expired, save the new tokens.

Might be better rephrased as....

> If you already have tokens for this user and they're expired, delete them and save the new tokens (REPLACE action).

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,105 @@
...
+               If you don't have tokens for this user, save the incoming (see next question).
+
+        3. When saving, what do I do about information I don't have such as home_account_id? (This is part of the
+           cache-key).
+
+           Brian's answer:
+               ...? (I would say, "make a request" but that's a lot to go wrong, slow, etc.)
```
iambmelt: Make a request

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,105 @@
...
+               If I already have records in the cache for this OID, I could potentially 'peek' into
+               them to try and get home_account_id info but that seems ill-advised.
+
+        4. For back-compat, it would seem that MSAL needs to both receive v1 idtokens AND supply them.
+           Is that true? What about when ADAL goes away? Will we add new API surface or version the
+           payloads so we stop trying to convert everything between v1/v2 and back again?
+
+           Brian's answer:
+               I think we should beef-up this implementation to return v1 and v2 artifacts BUT the
```
iambmelt: For MVP, new v2-ish payload not _required_ (but still good to have).

Points to consider:
If we're going to implement a version, make sure it behaves a future-looking way: ie don't just include a version without a plan for it.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,225 @@
...
+import static com.microsoft.identity.common.internal.migration.TokenCacheItemMigrationAdapter.renewToken;
+import static com.microsoft.identity.common.internal.migration.TokenCacheItemMigrationAdapter.sBackgroundExecutor;
+
+public class TokenShareUtility implements ITokenShareInternal {
+
+    private static final String TAG = TokenShareUtility.class.getSimpleName();
+    private static final Map<String, String> mClaimRemapper = new HashMap<>();
+
+    /*
+     * Populate v1/v2 IdToken claims mapping here.
```
iambmelt: @shoatman Here's where I'm doing v2 -> v1 claim mapping. To add more mappings, we add more `put()` calls here...

So far, I have just one special case:
- `preferred_username` becomes `upn`

I suspect there's a handful of these... something related to guest/idp/iss? Anything else?

iambmelt: Not clear on how `sub`, `email`, and `upn` align exactly...

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,225 @@
...
+
+    @SuppressWarnings("PMD.UnusedPrivateMethod")
+    @NonNull
+    private static TokenCacheItem adapt(@NonNull final IdTokenRecord idTokenRecord,
+                                        @NonNull final RefreshTokenRecord refreshTokenRecord) throws ServiceException {
+        final TokenCacheItem tokenCacheItem = new TokenCacheItem();
+        tokenCacheItem.setClientId(refreshTokenRecord.getClientId());
+        tokenCacheItem.setAuthority(idTokenRecord.getAuthority());
+        tokenCacheItem.setRefreshToken(refreshTokenRecord.getSecret());
+        tokenCacheItem.setResource(null); // TODO Does this need to be present? You can't turn scopes into a resource AFAIK...
```
iambmelt: This is my other TODO -- here I am transforming v2 credentials into a v1 equivalent.

Not clear to me what I can do about populating the resource property if everything I have in the cache is scopes based...

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenCacheItem.java
```diff
@@ -0,0 +1,140 @@
...
+ * This class is a shallow copy of the ADAL TokenCacheItem.  The ADAL TokenCacheItem is currently what
+ * the token sharing library expects to receive for "ORGID" accounts.  (AAD V1)
+ * <p>
+ * The following are required fields for deserialization in ADAL...
+ * jsonObj.add(OAuth2.AUTHORITY, new JsonPrimitive(tokenCacheItem.getAuthority()));
+ * jsonObj.add(OAuth2.REFRESH_TOKEN, new JsonPrimitive(tokenCacheItem.getRefreshToken()));
+ * jsonObj.add(OAuth2.ID_TOKEN, new JsonPrimitive(tokenCacheItem.getRawIdToken()));
+ * jsonObj.add(OAuth2.ADAL_CLIENT_FAMILY_ID, new JsonPrimitive(tokenCacheItem.getFamilyClientId()));
+ */
+public class TokenCacheItem implements ITokenCacheItem, Serializable {
```
iambmelt: I may pull this class out and use `ADALTokenCacheItem` directly instead - from a serialization perspective it looks to make no material difference.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
shoatman: I think for FOCI tokens this is null, but we can check the ADAL implementation to confirm.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenCacheItem.java
shoatman: That's fine with me.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
shoatman: sub is a pairwise identifier... an unique identifier for the app that only a particular client id will receive.  I doubt that EVO sends a pairwise identifier for the Microsoft family... and generally the FOCI apps are not using sub that I"m aware of.  Email and UPN can be different, but often are the same.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/ITokenShareInternal.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * For the supplied user unique identifier (OID), return the corresponding refresh token for
+     * that account if the current application is a member of FoCI (family of clientIds).
+     *
+     * @param oid The OID of the sought user's FRT.
+     * @return The {@link com.microsoft.identity.common.adal.internal.tokensharing.TokenCacheItem}
+     * serialized to JSON.
+     */
+    String getFamilyRefreshToken(String oid) throws Exception;
```
shoatman: if this is going to support object id and username.... likely should just be identifier.  We may want to support both.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenCacheItem.java
shoatman: That's fine with me.... this just leaves out all the fields that are not used by common/MSAL.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
shoatman: ADAL only falls back to subject if OID is not present.  I believe that OID is present for all accounts outside of MSA passthrough.  So likely don't need to populate sub.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -0,0 +1,225 @@
...
+        tokenCacheItem.setRawIdToken(mintV1IdToken(idTokenRecord.getSecret()));
+        tokenCacheItem.setFamilyClientId(refreshTokenRecord.getFamilyId());
+
+        return tokenCacheItem;
+    }
+
+    @NonNull
+    private static String mintV1IdToken(@NonNull final String rawV2IdToken) throws ServiceException {
```
shoatman: We should specify in the name of the method that we're creating/minting a v1 token from a v2 id_token

iambmelt: Resolved in 042e9639d273e518ded148bd55cb2848c155c12a

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenCacheItem.java
iambmelt: Partially resolved in 5c05e41f2ac1d140e990cbfd4fe1350bf97d0538

iambmelt: Actually, I'm going to call this fully resolved. I was considering trying to share the SSOStateSerializer as well but this looks to be lots of work for little payoff. Moving on...

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
iambmelt: Confirming; in the examples I captured from ADAL the resource value is `null`

iambmelt: Resolved in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/529/commits/00902758e3df779e300f1a5cdbb7990ab345fa2e

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
iambmelt: OK - sounds like this stays as-is... that is: we remap `preferred_username` to `upn` and that's it.

ACK?

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/ITokenShareInternal.java
iambmelt: ACK - previous implementation supports both `oid` and `upn` -- will update 👍

iambmelt: Resolved in:
946147769d0c7c1ba75487243c150e1162ddb4f8
862f13e2166b490ee85a8f089ec73c95596af171

iambmelt:
@shoatman Made updates per your review. Ready for another round. Also have a PR [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/657) for public MSAL.

PR #534, opened 2019-06-21
[WIP] Adding logic for broker retry (StorageHelper keyswap fallback)

iambmelt:
See title.

PR #535, opened 2019-06-22
Adding logic for broker retry (StorageHelper keyswap fallback)

iambmelt:
Iterates on #534

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -356,9 +356,9 @@ public class StorageHelper implements IStorageHelper {
         if (!encryptedBlob.substring(1, 1 + encodeVersionLength).equals(ENCODE_VERSION)) {
             throw new IllegalArgumentException(
                     String.format(
-                            "Encode version received was: '%s', Encode version supported is: '%s'",
-                            encryptedBlob,
-                            ENCODE_VERSION)
+                            "Unsupported encode version received. Encode version supported is: '%s'",
+                            ENCODE_VERSION
+                    )
             );
         }
     }
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -228,7 +228,6 @@ public class StorageHelper implements IStorageHelper {
         }
 
         int encodeVersionLength = encryptedBlob.charAt(0) - 'a';
-
         validateEncodeVersion(encryptedBlob, encodeVersionLength);
 
         final byte[] bytes = Base64.decode(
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -220,42 +231,77 @@ public String encrypt(final String clearText)
...
+            Logger.error(
+                    TAG + methodName,
+                    "Failed to decrypt with package name: "
+                            + packageName,
+                    e
+            );
+
+            if (!BROKER_PACKAGE_NAMES.contains(packageName)) {
+                // If we are not the broker, do not retry, immediately throw.
+                throw e;
```
sdabbiru: This check can be moved to right after line 250 (final String packageName = mContext.getPackageName();) - we should not even attempting to read if we are not one of these 2 apps.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -220,42 +231,77 @@ public String encrypt(final String clearText)
...
+            }
+
+            // We are the broker... try with the other broker's package name
+            final String nextPackageToUse =
+                    packageName.equals(AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
```
sdabbiru: >equals [](start = 32, length = 6)

Ignore case?

sdabbiru: Or does it matter?

---
In reply to: [296825941](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/535#discussion_r296825941) [](ancestors = 296825941)

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -276,16 +322,47 @@ public String decrypt(final String encryptedBlob)
...
+                            encodeVersionLength
+                    )
+            );
+        }
+
+        if (!encryptedBlob.substring(1, 1 + encodeVersionLength).equals(ENCODE_VERSION)) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Encode version received was: '%s', Encode version supported is: '%s'",
+                            encryptedBlob,
```
sdabbiru: >encryptedBlob [](start = 28, length = 13)

Should we be dumping this in the exception message?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
iambmelt: Not true; this is shared code with MSAL

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
iambmelt: Package name [cannot change](https://android-developers.googleblog.com/2011/06/things-that-cannot-change.html) post-publish. All of our package names are lowercase, so makes no difference

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
iambmelt: We should not, will remove. Thanks

iambmelt: Resolved in 05bfd4cc9398988713ed8f8b55e232aa6dc83ad9

iambmelt:
@sdabbiru Edits made as requested/relevant

PR #538, opened 2019-07-01
Enhance MicrosoftAuthClient to support binding between brokers

rpdome:
Check into dev for now. 

Will cherry-pick (and update .aidl order) if we're going to release this fix sooner.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
@@ -30,6 +30,7 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
@@ -174,14 +175,15 @@ public class MicrosoftAuthClient {
      * @param context
      * @return package name of the installed inactive broker app, if there's any.
      */
+    @Nullable
     private String getInactiveBrokerPackageName(@NonNull final Context context) {
         final String methodName = ":getInactiveBrokerPackageName";
         String currentBroker = getCurrentActiveBrokerPackageName(context);
 
         final String inactiveBrokerPackageName;
-        if (currentBroker.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
+        if (AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(currentBroker)) {
             inactiveBrokerPackageName = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
-        } else if (currentBroker.equalsIgnoreCase(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)) {
+        } else if (AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(currentBroker)) {
             inactiveBrokerPackageName = AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
         } else {
             return null;
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
@@ -172,7 +172,7 @@ public class MicrosoftAuthClient {
      * - If authenticator is the broker, this will return CP.
      *
      * @param context
-     * @return package name of the installed inactive authenticator, if there's any.
+     * @return package name of the installed inactive broker app, if there's any.
      */
     private String getInactiveBrokerPackageName(@NonNull final Context context) {
         final String methodName = ":getInactiveBrokerPackageName";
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+
+    /**
+     * Returns the inactive installed authenticator.
+     * - If CP is the broker, this will return authenticator.
+     * - If authenticator is the broker, this will return CP.
+     *
+     * @param context
+     * @return package name of the installed inactive broker app, if there's any.
+     */
+    private String getInactiveBrokerPackageName(@NonNull final Context context) {
```
iambmelt: Minor nit: can declare method `@Nullable`

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+        final String methodName = ":getInactiveBrokerPackageName";
+        String currentBroker = getCurrentActiveBrokerPackageName(context);
+
+        final String inactiveBrokerPackageName;
+        if (currentBroker.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
```
iambmelt: This looks to be a possible NPE... `getCurrentActiveBrokerPackageName()` will return `null` if there is no active broker -- flipping this check around a comparing against the constant would avoid this....

so

```java
if (AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(currentBroker)) {
    // ...
}
```

The same would apply to the `else if` clauses

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +166,40 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+        } else {
+            return null;
+        }
+
+        // Verify the signature to make sure that we're not binding to malicious apps.
+        final BrokerValidator validator = new BrokerValidator(context);
+        try {
+            if (validator.verifySignature(inactiveBrokerPackageName)) {
+                return inactiveBrokerPackageName;
+            }
```
kreedula: May be we should log here if verification fails

sdabbiru: Yes, actually, we should send telemetry if the signature check fails.

---
In reply to: [299733100](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/538#discussion_r299733100) [](ancestors = 299733100)

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+        } else if (currentBroker.equalsIgnoreCase(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)) {
+            inactiveBrokerPackageName = AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+        } else {
+            return null;
+        }
+
+        // Verify the signature to make sure that we're not binding to malicious apps.
+        final BrokerValidator validator = new BrokerValidator(context);
+        try {
+            if (validator.verifySignature(inactiveBrokerPackageName)) {
```
sdabbiru: >inactiveBrokerPackageName [](start = 42, length = 25)

Will "inactiveBrokerPackage" be even present on the device at all times?

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+                return inactiveBrokerPackageName;
+            }
+        } catch (Exception e) {
+            // Log. It could be that inactiveBrokerPackageName is not installed on the app, or something's wrong.
```
sdabbiru: >app [](start = 87, length = 3)

device?

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+            Logger.error(TAG + methodName, "Failed to get inactive Broker package name.", e);
```
sdabbiru: >e [](start = 90, length = 1)

Will the exception e have sufficient information to distinguish between the 'inactive' broker package not installed on the device vs the package signature validation failure?

rpdome:
Close as this should also support V1.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -140,4 +165,39 @@ private String getCurrentActiveBrokerPackageName(@NonNull final Context context)
...
+            return null;
+        }
+
+        // Verify the signature to make sure that we're not binding to malicious apps.
+        final BrokerValidator validator = new BrokerValidator(context);
+        try {
+            if (validator.verifySignature(inactiveBrokerPackageName)) {
+                return inactiveBrokerPackageName;
+            }
+        } catch (Exception e) {
```
sdabbiru: >Exception [](start = 17, length = 9)

I do not like this type of exception handling. I would like us to handle specific exceptions that are expected and then fallback on the 'Exception'.

rpdome: This one is closed. Please review https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/547 instead. Thanks!

PR #543, opened 2019-07-03
Adding support for MSA to TSL

iambmelt:
Adds support for MSA FRTs to TSL.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.adal.internal.tokensharing;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Pair;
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -180,6 +207,57 @@ public void saveFamilyRefreshToken(@NonNull final String tokenCacheItemJson) thr
         }
     }
 
+    @Override
+    public String getMsaFamilyRefreshToken(@NonNull final String identifier) throws Exception {
+        final ICacheRecord cacheRecord = getCacheRecordForIdentifier(identifier);
+
+        throwIfCacheRecordIncomplete(identifier, cacheRecord);
+
+        return cacheRecord.getRefreshToken().getSecret();
```
heidijinxujia: Is cacheRecord.getRefreshToken() always nonnull?

iambmelt: Yep - `throwIfCacheRecordIncomplete` will check the contents of the cache record and throw an Exception if it's incomplete

PR #546, opened 2019-07-04
[Part 1] StorageHelper fix

rpdome:
Doesn't include IPC call.

The code is verified by test cases only. Please let me know if there's any other case you'd like me to cover.

Given that this is a sensitive layer, I'll complete the PR once I have at least 2 sign-offs.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -253,7 +253,7 @@ public class StorageHelper implements IStorageHelper {
 
     @Override
     public String decrypt(final String encryptedBlob) throws GeneralSecurityException, IOException {
-        final String methodName = "decrypt";
+        final String methodName = ":decrypt";
         Logger.verbose(TAG + methodName, "Starting decryption");
 
         if (StringExtensions.isNullOrBlank(encryptedBlob)) {
@@ -266,10 +266,10 @@ public class StorageHelper implements IStorageHelper {
         }
 
         final String packageName = getPackageName();
-        final List<KeyType> potentialDecryptionKeyList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
+        final List<KeyType> potentialDecryptionKeyTypeList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
 
         final byte[] bytes = getByteArrayFromEncryptedBlob(encryptedBlob);
-        for (final KeyType keyType : potentialDecryptionKeyList) {
+        for (final KeyType keyType : potentialDecryptionKeyTypeList) {
             try {
                 final SecretKey secretKey = loadSecretKey(keyType);
                 if (secretKey == null) {
@@ -301,7 +301,7 @@ public class StorageHelper implements IStorageHelper {
      * NOTE :If it cannot verify the keyVersion, it will assume that this data is not encrypted.
      * */
     public EncryptionType getEncryptionType(@NonNull final String data) throws IOException {
-        final String methodName = "isEncryptedWithUserDefinedKey";
+        final String methodName = ":getEncryptionType";
 
         final byte[] bytes;
         try {
@@ -345,9 +345,10 @@ public class StorageHelper implements IStorageHelper {
     /**
      * Get all the key type that could be potential candidates for decryption.
      **/
+    @NonNull
     private List<KeyType> initializeDecryptionKeyTypeList(@NonNull final String encryptedBlob,
                                                           @NonNull final String packageName) throws IOException {
-        final String methodName = "initializeDecryptionKeyTypeList";
+        final String methodName = ":initializeDecryptionKeyTypeList";
         List<KeyType> keyTypeList = new ArrayList<>();
 
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
@@ -374,7 +375,7 @@ public class StorageHelper implements IStorageHelper {
     private String decryptWithSecretKey(@NonNull final byte[] bytes,
                                         @NonNull final SecretKey secretKey)
             throws GeneralSecurityException, IOException {
-        final String methodName = "decryptWithSecretKey";
+        final String methodName = ":decryptWithSecretKey";
 
         final SecretKey hmacKey = getHMacKey(secretKey);
 
@@ -507,7 +508,7 @@ public class StorageHelper implements IStorageHelper {
      */
     @Nullable
     protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOException, GeneralSecurityException {
-        final String methodName = "loadSecretKey";
+        final String methodName = ":loadSecretKey";
 
         switch (keyType) {
             case LEGACY_AUTHENTICATOR_APP_KEY:
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -346,22 +346,22 @@ public class StorageHelper implements IStorageHelper {
      * Get all the key type that could be potential candidates for decryption.
      **/
     @NonNull
-    private List<KeyType> getKeysForDecryptionType(@NonNull final String encryptedBlob,
-                                                   @NonNull final String packageName) throws IOException {
+    public List<KeyType> getKeysForDecryptionType(@NonNull final String encryptedBlob,
+                                                  @NonNull final String packageName) throws IOException {
         final String methodName = ":initializeDecryptionKeyTypeList";
         List<KeyType> keyTypeList = new ArrayList<>();
 
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
 
         if (encryptionType == EncryptionType.USER_DEFINED) {
-            if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
+            if (AuthenticationSettings.INSTANCE.getSecretKeyData() != null) {
+                keyTypeList.add(KeyType.ADAL_USER_DEFINED_KEY);
+            } else if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
                 keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
                 keyTypeList.add(KeyType.LEGACY_AUTHENTICATOR_APP_KEY);
             } else if (AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
                 keyTypeList.add(KeyType.LEGACY_AUTHENTICATOR_APP_KEY);
                 keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
-            } else {
-                keyTypeList.add(KeyType.ADAL_USER_DEFINED_KEY);
             }
         } else if (encryptionType == EncryptionType.ANDROID_KEY_STORE) {
             keyTypeList.add(KeyType.KEYSTORE_ENCRYPTED_KEY);
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -181,13 +181,13 @@ public class StorageHelper implements IStorageHelper {
     @Override
     public String encrypt(final String clearText)
             throws GeneralSecurityException, IOException {
-        final String methodName = ":readKeyData";
-
-        Logger.verbose(TAG + methodName, "Starting encryption");
+        final String methodName = ":encrypt";
 
         if (StringExtensions.isNullOrBlank(clearText)) {
             throw new IllegalArgumentException("Input is empty or null");
         }
+
+        Logger.verbose(TAG + methodName, "Starting encryption");
 
         // load key for encryption if not loaded
         mEncryptionKey = loadSecretKeyForEncryption();
@@ -256,7 +256,9 @@ public class StorageHelper implements IStorageHelper {
                     continue;
                 }
 
-                return decryptWithSecretKey(bytes, secretKey);
+                String result = decryptWithSecretKey(bytes, secretKey);
+                Logger.verbose(TAG + methodName, "Finished decryption.");
+                return result;
             } catch (GeneralSecurityException | IOException e) {
                 Logger.error(
                         TAG + methodName,
@@ -278,8 +280,6 @@ public class StorageHelper implements IStorageHelper {
 
         final byte[] bytes = getByteArrayFromEncryptedBlob(encryptedBlob);
 
-        // get key version used for this data. If user upgraded to different
-        // API level, data needs to be updated
         final String keyVersion = new String(
                 bytes,
                 0,
@@ -415,9 +415,8 @@ public class StorageHelper implements IStorageHelper {
     /**
      * Load SecretKey for Broker for encryption.
      * This will try getting Keystore-encrypted symmetric key in the following order.
-     * 1. Existing key.
-     * 2. the key owned by inactive broker.
-     * 3. a newly-generated key.
+     * 1. Keystore-encrypted symmetric key.
+     * 2. a newly-generated key.
      */
     private SecretKey loadSecretKeyForBrokerEncryption() throws IOException, GeneralSecurityException {
         setBlobVersion(VERSION_ANDROID_KEY_STORE);
@@ -732,7 +731,6 @@ public class StorageHelper implements IStorageHelper {
     private void deleteKeyFile() {
         final String methodName = ":deleteKeyFile";
 
-        // Store secret key in a file after wrapping
         final File keyFile = new File(mContext.getDir(mContext.getPackageName(),
                 Context.MODE_PRIVATE), ADALKS);
         if (keyFile.exists()) {
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -28,6 +28,7 @@ import android.content.Context;
 import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Base64;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
@@ -514,7 +515,7 @@ public class StorageHelper implements IStorageHelper {
     /**
      * Given the key type, load a secret key.
      */
-    @NonNull
+    @Nullable
     protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOException, GeneralSecurityException {
         final String methodName = "loadSecretKey";
 
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -261,15 +261,15 @@ public class StorageHelper implements IStorageHelper {
         }
 
         if (getEncryptionType(encryptedBlob) == EncryptionType.UNENCRYPTED) {
-            Logger.verbose(TAG + methodName, "This string is not encrypted. Finished decryption.");
+            Logger.warn(TAG + methodName, "This string is not encrypted. Finished decryption.");
             return encryptedBlob;
         }
 
         final String packageName = getPackageName();
-        final List<KeyType> potentialDecryptionKeyTypeList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
+        final List<KeyType> keysForEncryptionType = getKeysForEncryptionType(encryptedBlob, packageName);
 
         final byte[] bytes = getByteArrayFromEncryptedBlob(encryptedBlob);
-        for (final KeyType keyType : potentialDecryptionKeyTypeList) {
+        for (final KeyType keyType : keysForEncryptionType) {
             try {
                 final SecretKey secretKey = loadSecretKey(keyType);
                 if (secretKey == null) {
@@ -346,8 +346,8 @@ public class StorageHelper implements IStorageHelper {
      * Get all the key type that could be potential candidates for decryption.
      **/
     @NonNull
-    private List<KeyType> initializeDecryptionKeyTypeList(@NonNull final String encryptedBlob,
-                                                          @NonNull final String packageName) throws IOException {
+    private List<KeyType> getKeysForEncryptionType(@NonNull final String encryptedBlob,
+                                                   @NonNull final String packageName) throws IOException {
         final String methodName = ":initializeDecryptionKeyTypeList";
         List<KeyType> keyTypeList = new ArrayList<>();
 
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -28,9 +28,7 @@ import android.content.Context;
 import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
 import android.support.annotation.NonNull;
-import android.support.annotation.Nullable;
 import android.util.Base64;
-import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
@@ -45,6 +43,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.security.DigestException;
 import java.security.GeneralSecurityException;
@@ -57,11 +56,11 @@ import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.spec.AlgorithmParameterSpec;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 
 import javax.crypto.Cipher;
 import javax.crypto.KeyGenerator;
@@ -145,10 +144,15 @@ public class StorageHelper implements IStorageHelper {
 
     private static final String ANDROID_KEY_STORE = "AndroidKeyStore";
 
-    private static final Set<String> BROKER_PACKAGE_NAMES = new HashSet<String>() {{
-        add(AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
-        add(COMPANY_PORTAL_APP_PACKAGE_NAME);
-    }};
+    /**
+     * Type of Secret key to be used.
+     */
+    public enum KeyType {
+        LEGACY_AUTHENTICATOR_APP_KEY,
+        LEGACY_COMPANY_PORTAL_KEY,
+        ADAL_USER_DEFINED_KEY,
+        KEYSTORE_ENCRYPTED_KEY
+    }
 
     private final Context mContext;
     private final SecureRandom mRandom;
@@ -158,8 +162,8 @@ public class StorageHelper implements IStorageHelper {
      */
     private KeyPair mKeyPair;
     private String mBlobVersion;
-    private SecretKey mKey = null;
-    private SecretKey mHMACKey = null;
+    private SecretKey mEncryptionKey = null;
+    private SecretKey mEncryptionHMACKey = null;
     private SecretKey mSecretKeyFromAndroidKeyStore = null;
 
     /**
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -177,17 +181,19 @@ public StorageHelper(@NonNull final Context context) {
     @Override
     public String encrypt(final String clearText)
             throws GeneralSecurityException, IOException {
-        Log.v(TAG, "Starting encryption");
+        final String methodName = ":readKeyData";
```
sdabbiru: >readKeyData [](start = 36, length = 11)

Please update.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -177,17 +181,19 @@ public StorageHelper(@NonNull final Context context) {
...
+
+        Logger.verbose(TAG + methodName, "Starting encryption");
```
sdabbiru: >Logger.verbose(TAG + methodName, "Starting encryption"); [](start = 8, length = 56)

Move this below the clearText check.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,64 +231,53 @@ public String encrypt(final String clearText)
...
-
-        final byte[] bytes = Base64.decode(
-                encryptedBlob.substring(1 + encodeVersionLength),
-                Base64.DEFAULT
-        );
-
         final String packageName = mContext.getPackageName();
+        final List<KeyType> potentialDecryptionKeyList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
+
+        final byte[] bytes = getByteArrayFromEncryptedBlob(encryptedBlob);
```
sdabbiru: >getByteArrayFromEncryptedBlob [](start = 29, length = 29)

You can move this into the below loop, but below continue.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,64 +231,53 @@ public String encrypt(final String clearText)
...
 
-    private String decryptWithPackageName(final byte[] bytes,
-                                          @NonNull final String packageName)
-            throws GeneralSecurityException, IOException {
+    public boolean isEncryptedWithUserDefinedKey(@NonNull final String encryptedBlob) throws UnsupportedEncodingException {
+
+        final byte[] bytes = getByteArrayFromEncryptedBlob(encryptedBlob);
+
         // get key version used for this data. If user upgraded to different
         // API level, data needs to be updated
```
sdabbiru: Does this comment still makes sense to keep?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,64 +231,53 @@ public String encrypt(final String clearText)
...
-                    TAG + methodName,
-                    "Failed to decrypt with package name: "
-                            + packageName,
-                    e
-            );
-
-            if (!BROKER_PACKAGE_NAMES.contains(packageName)) {
-                // If we are not the broker, do not retry, immediately throw.
-                throw e;
+                return decryptWithSecretKey(bytes, secretKey);
```
sdabbiru: >decryptWithSecretKey [](start = 23, length = 20)

I would like to see the "closing" log statements to be in the same method - in this case, log message "starting decryption" is in this method, while "Finished decryption" is elsewhere.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +397,113 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
-    @Nullable
-    private byte[] getSecretKeyData(@Nullable final String pkgName) {
-        byte[] secretKeyData;
+    /**
+     * Load SecretKey for Broker for encryption.
+     * This will try getting Keystore-encrypted symmetric key in the following order.
+     * 1. Existing key.
+     * 2. the key owned by inactive broker.
+     * 3. a newly-generated key.
+     */
```
sdabbiru: Is this accurate?

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -335,4 +335,8 @@ private ErrorStrings() {
      */
     public static final String UNSUPPORTED_BROKER_VERSION = "unsupported_broker_version";
 
+    /**
+     * Decryption failed.
+     */
+    public static final String DECRYPTION_FAILED = "decryption_failed";
```
sdabbiru: >decryption_failed [](start = 52, length = 17)

Is this going to be consumed by client apps?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -628,36 +707,38 @@ private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
             throw ex;
         }
 
         return unwrappedSecretKey;
     }
 
     private void deleteKeyFile() {
+        final String methodName = ":deleteKeyFile";
+
         // Store secret key in a file after wrapping
```
sdabbiru: >// Store secret key in a file after wrapping [](start = 8, length = 44)

This comment does not apply here. Please remove.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -628,36 +707,38 @@ private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
         final File keyFile = new File(mContext.getDir(mContext.getPackageName(),
                 Context.MODE_PRIVATE), ADALKS);
         if (keyFile.exists()) {
-            Log.v(TAG, "Delete KeyFile");
+            Logger.verbose(TAG + methodName, "Delete KeyFile");
             if (!keyFile.delete()) {
-                Log.v(TAG, "Delete KeyFile failed");
+                Logger.verbose(TAG + methodName, "Delete KeyFile failed");
```
sdabbiru: >failed [](start = 65, length = 6)

Does it make sense to log the reason why it failed?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: not anymore. Will update.

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
rpdome: It should be handled by broker.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: The delete() method won't tell us why it fails.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +236,117 @@ public String encrypt(final String clearText)
...
+        return Base64.decode(
+                encryptedBlob.substring(1 + encodeVersionLength),
+                Base64.DEFAULT
         );
+    }
+
+    /**
+     * Get all the key type that could be potential candidates for decryption.
+     **/
+    private List<KeyType> initializeDecryptionKeyTypeList(@NonNull final String encryptedBlob,
```
alanlaims: nit: a `@NonNull` annotation could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +236,117 @@ public String encrypt(final String clearText)
...
+                                                          @NonNull final String packageName) {
+        List<KeyType> keyTypeList = new ArrayList<>();
+
+        if (isEncryptedWithUserDefinedKey(encryptedBlob)) {
```
alanlaims: Is it useful to include logging for each of these conditions?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +236,117 @@ public String encrypt(final String clearText)
...
+            }
+        } else {
+            keyTypeList.add(KeyType.KEYSTORE_ENCRYPTED_KEY);
+        }
 
-        Log.v(TAG, "Encrypt version:" + keyVersion);
+        return keyTypeList;
+    }
+
+    private String decryptWithSecretKey(final byte[] bytes,
```
alanlaims: nit: a `@NonNull` could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +418,112 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
-    @Nullable
-    private byte[] getSecretKeyData(@Nullable final String pkgName) {
-        byte[] secretKeyData;
+    /**
+     * Load SecretKey for Broker for encryption.
+     * This will try getting Keystore-encrypted symmetric key in the following order.
+     * 1. Keystore-encrypted symmetric key.
+     * 2. a newly-generated key.
+     */
+    private SecretKey loadSecretKeyForBrokerEncryption() throws IOException, GeneralSecurityException {
```
alanlaims: nit: `@NonNull` annotation could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +418,112 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
-        // restart the device anyway. It will load the correct key for new API.
-        if (mKey != null && mHMACKey != null) {
-            return mKey;
+    /**
+     * Load SecretKey for ADAL/MSAL for encryption.
+     * This will try getting key in the following order.
+     * 1. User-defined key.
+     * 2. Keystore-encrypted symmetric key.
+     */
+    private SecretKey loadSecretKeyForADALMSALEncryption() throws IOException, GeneralSecurityException {
```
alanlaims: nit: `@NonNull` annotation could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +418,112 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
+     * Exposed for test cases.
+     */
+    protected void setBlobVersion(@NonNull String blobVersion) {
+        mBlobVersion = blobVersion;
+    }
+
+    /**
+     * Given the key type, load a secret key.
+     */
+    protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOException, GeneralSecurityException {
```
alanlaims: nit: `@NonNull` annotation could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +236,117 @@ public String encrypt(final String clearText)
...
+                keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
+            } else {
+                keyTypeList.add(KeyType.ADAL_USER_DEFINED_KEY);
+            }
+        } else {
+            keyTypeList.add(KeyType.KEYSTORE_ENCRYPTED_KEY);
+        }
 
-        Log.v(TAG, "Encrypt version:" + keyVersion);
+        return keyTypeList;
```
heidijinxujia: It might be useful if we log the length of the key list.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +418,112 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
+                return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
+
+            case KEYSTORE_ENCRYPTED_KEY:
+                return getKey();
         }
 
-        mBlobVersion = defaultBlobVersion;
-        return getKeyOrCreate(mBlobVersion);
+        Logger.verbose(TAG + methodName, "Unknown KeyType. This code should never be reached.");
+        throw new GeneralSecurityException(ErrorStrings.UNKNOWN_ERROR);
```
heidijinxujia: Could you specify the error code here? Like unable to load the secret key or something else.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: This is already logged inside decrypt() for loop, when the decryption fails. I think that should be enough.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +245,133 @@ public String encrypt(final String clearText)
...
+        Logger.verbose(TAG + methodName, "Starting decryption");
 
         if (StringExtensions.isNullOrBlank(encryptedBlob)) {
             throw new IllegalArgumentException("Input is empty or null");
         }
 
-        int encodeVersionLength = encryptedBlob.charAt(0) - 'a';
-        validateEncodeVersion(encryptedBlob, encodeVersionLength);
+        if (getEncryptionType(encryptedBlob) == EncryptionType.UNENCRYPTED) {
+            Logger.verbose(TAG + methodName, "This string is not encrypted. Finished decryption.");
```
rpdome: This behavior change is introduced because we'll be treating PRT, BRT, etc as encrypted value after this change is shipped.

These values are currently unencrypted, and if the migration fails, then these values will remain unencrypted. 

Technically, we could keep the same behavior, and it would throw an exception (since it fails to verify what type of 'encryption' was used on the blob. We're proposing that if such things happen, then we should fail the flow and wipe the account. 

I think this is a better way to gracefully support such case. However, if you think it's an edge case and not worth making the change, I could revert back to the original logic.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +245,133 @@ public String encrypt(final String clearText)
...
-        Log.v(TAG, "Starting decryption");
+        Logger.verbose(TAG + methodName, "Starting decryption");
 
         if (StringExtensions.isNullOrBlank(encryptedBlob)) {
             throw new IllegalArgumentException("Input is empty or null");
         }
 
-        int encodeVersionLength = encryptedBlob.charAt(0) - 'a';
-        validateEncodeVersion(encryptedBlob, encodeVersionLength);
+        if (getEncryptionType(encryptedBlob) == EncryptionType.UNENCRYPTED) {
```
rpdome: This behavior change is introduced because we'll be treating BRT, etc as encrypted value after this change is shipped.

These values are currently unencrypted, and if the migration fails, then these values will remain unencrypted. 

This change will gracefully return the very same string back to the caller.

Technically, we could keep the same behavior, and it would throw an exception (since it fails to verify what type of 'encryption' was used on the blob. We're proposing that if such things happen, then we should fail the flow and wipe the account. 

I think this is a better way to support such case. However, if you think it's an edge case and not worth making the change, I could revert back to the original logic.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: We could, but given that this code should never be reached, if it happens, our logic is wrong and there is no way to recover from such state. Therefore I don't think introducing another error code would benefit here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: This could be nullable.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,136 @@ public String encrypt(final String clearText)
...
+            Logger.verbose(TAG + methodName, "This string is not encrypted. Finished decryption.");
+            return encryptedBlob;
+        }
 
-        final byte[] bytes = Base64.decode(
-                encryptedBlob.substring(1 + encodeVersionLength),
-                Base64.DEFAULT
-        );
+        final String packageName = getPackageName();
+        final List<KeyType> potentialDecryptionKeyList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
```
shahzaibj: This can be renamed to `potentialDecryptionKeyTypeList `

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,136 @@ public String encrypt(final String clearText)
...
+
+        throw new GeneralSecurityException(ErrorStrings.DECRYPTION_FAILED);
+    }
+
+    /**
+     * Determine type of encryption performed on the given data blob.
+     * NOTE :If it cannot verify the keyVersion, it will assume that this data is not encrypted.
+     * */
+    public EncryptionType getEncryptionType(@NonNull final String data) throws IOException {
+        final String methodName = "isEncryptedWithUserDefinedKey";
```
shahzaibj: The method name string has colon as prefix in some methods whereas it is not present in other places such as this one. I think we should adopt one pattern and be consistent. For example, we can change this to `final String methodName = ":isEncryptedWithUserDefinedKey";` and do the same in all other places as well.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -145,10 +145,24 @@
...
-        add(COMPANY_PORTAL_APP_PACKAGE_NAME);
-    }};
+    /**
+     * Type of Secret key to be used.
+     */
+    public enum KeyType {
+        LEGACY_AUTHENTICATOR_APP_KEY,
+        LEGACY_COMPANY_PORTAL_KEY,
+        ADAL_USER_DEFINED_KEY,
+        KEYSTORE_ENCRYPTED_KEY
```
Om83: We are planning to store the new key to keystore but just in case it changes we should name something LATEST_LOCALGEN_KEY

rpdome: If we're going to change, it's likely that we'll have to introduce a new one (i.e. as we have to 'decrypt' with the 'old type' and 'encrypt' them with the new type.

rpdome: If we're just going to change how we store the very key, in that case I'd still prefer change the name at that point. I personally found LATEST_LOCALGEN_KEY too generic (i.e. if we introduce hardware-backed encryption, then it's still technically 'locally generated').

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
shoatman: I'd consider logging a warning here rather than verbose.  Only since this is unexpected.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+            Logger.verbose(TAG + methodName, "This string is not encrypted. Finished decryption.");
+            return encryptedBlob;
+        }
 
-        final byte[] bytes = Base64.decode(
-                encryptedBlob.substring(1 + encodeVersionLength),
-                Base64.DEFAULT
-        );
+        final String packageName = getPackageName();
+        final List<KeyType> potentialDecryptionKeyTypeList = initializeDecryptionKeyTypeList(encryptedBlob, packageName);
```
shoatman: getKeysForEncryptionType?

shoatman: Why not just return the keys here... rather than the list of key types?

rpdome: After scoping down the work, this is no longer needed. (This is to prevent making unnecessary IPC calls in an edge case - we only invoke them right before we're decrypting - not when we're initializing the list).

I'll just return the list of keys as suggested, as that's more straightforward.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+    private List<KeyType> getKeysForEncryptionType(@NonNull final String encryptedBlob,
+                                                   @NonNull final String packageName) throws IOException {
+        final String methodName = ":initializeDecryptionKeyTypeList";
+        List<KeyType> keyTypeList = new ArrayList<>();
+
+        EncryptionType encryptionType = getEncryptionType(encryptedBlob);
+
+        if (encryptionType == EncryptionType.USER_DEFINED) {
+            if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
+                keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
```
titorodi: Should these add try keystore too?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+                if (secretKey == null) {
+                    continue;
+                }
+
+                String result = decryptWithSecretKey(bytes, secretKey);
+                Logger.verbose(TAG + methodName, "Finished decryption.");
+                return result;
+            } catch (GeneralSecurityException | IOException e) {
+                Logger.error(
+                        TAG + methodName,
```
titorodi: Will we get an error for every value if we have the keysForEncryptionType order wrong as we use StorageHelper? I'm not sure what to recommend.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +448,105 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
-        final byte[] secretKeyData = getSecretKeyData(pkgName);
-        return loadSecretKeyForEncryption(secretKeyData == null ? VERSION_ANDROID_KEY_STORE : VERSION_USER_DEFINED);
+        // Loading key only once for performance. If API is upgraded, it will
+        // restart the device anyway. It will load the correct key for new API.
+        if (mEncryptionKey != null && mEncryptionHMACKey != null) {
+            return mEncryptionKey;
+        }
+
+        final String packageName = getPackageName();
+        if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName) ||
```
titorodi: Does this mean the Authenticator and Company Portal will be using these keys too for their local auth?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
shahzaibj: @rpdome has this been updated?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: We will see it in the logs, yes.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+                Base64.DEFAULT
+        );
+    }
 
-            return decryptWithPackageName(bytes, nextPackageToUse);
+    /**
+     * Get all the key type that could be potential candidates for decryption.
+     **/
+    @NonNull
+    private List<KeyType> getKeysForEncryptionType(@NonNull final String encryptedBlob,
```
rpdome: nit: wrong name

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: I think we're good. In this case we know that the key is not encrypted by the Keystore-encrypted symmetric key.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: Good catch. I'll simplify the logic to do just the following 

1. look for ADAL secret Key -> if exist, returns.
2. Use KeyStore secret key.

No need to separate cases anymore as we're no longer re-encrypt inline.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+                Logger.error(
+                        TAG + methodName,
+                        "Failed to decrypt with KeyType: "
+                                + keyType.toString(),
+                        e
+                );
+            }
+        }
+
+        Logger.verbose(
```
sdabbiru: >verbose [](start = 15, length = 7)

Info.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+                TAG + methodName,
+                "Tried all decryption keys and decryption still fails. Throw an exception.");
+
+        throw new GeneralSecurityException(ErrorStrings.DECRYPTION_FAILED);
```
sdabbiru: >DECRYPTION_FAILED [](start = 56, length = 17)

I know this is a late ask... I think it is worthwhile to have telemetry on all failed decryption attempts - these instances should go down and if they are not going down, then we know we have a bug.

rpdome: already bubbling this up in part 2.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
-            return decryptWithPackageName(bytes, packageName);
-        } catch (GeneralSecurityException | IOException e) {
-            Logger.error(
-                    TAG + methodName,
-                    "Failed to decrypt with package name: "
-                            + packageName,
-                    e
+            bytes = getByteArrayFromEncryptedBlob(data);
+        } catch (IllegalArgumentException e) {
+            Logger.error(TAG + methodName, "This data is not an encrypted blob.", e);
```
kreedula: Let's not log this as error instead have a verbose log as we use this method to determine the type , and for unencrypted type it shows an error which will be a false alarm

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -225,76 +246,137 @@ public String encrypt(final String clearText)
...
+    }
 
-            return decryptWithPackageName(bytes, nextPackageToUse);
+    /**
+     * Get all the key type that could be potential candidates for decryption.
+     **/
+    @NonNull
+    public List<KeyType> getKeysForDecryptionType(@NonNull final String encryptedBlob,
+                                                  @NonNull final String packageName) throws IOException {
+        final String methodName = ":initializeDecryptionKeyTypeList";
```
shahzaibj: methodName string does not match the actual methodName

rpdome: forgot to change this after renaming -_-''' Thanks for the catch!

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -366,59 +448,76 @@ private void validateEncodeVersion(String encryptedBlob, int encodeVersionLength
...
+            case LEGACY_COMPANY_PORTAL_KEY:
+                Logger.verbose(TAG + methodName, "Loading legacy companyPortal key.");
+                return getSecretKey(AuthenticationSettings.INSTANCE.getBrokerSecretKeys().get(COMPANY_PORTAL_APP_PACKAGE_NAME));
+
+            case ADAL_USER_DEFINED_KEY:
+                Logger.verbose(TAG + methodName, "Loading ADAL userDefined key.");
+                return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
+
+            case KEYSTORE_ENCRYPTED_KEY:
+                return getKey();
```
Om83: May be add logging here too .

rpdome:
Thanks for the overwhelming support 😃

PR #547, opened 2019-07-04
[Part 2] Add Intra-broker IPC support

rpdome:
Verified with the test case with the following configuration

1. Broker app installed, and it contains the updated .aidl
2. Broker app installed, but it doesn't have the new function added to .aidl.
3. Broker app not installed.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -28,6 +28,7 @@ import android.content.Context;
 import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Base64;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
@@ -514,26 +515,36 @@ public class StorageHelper implements IStorageHelper {
      *
      * @return true if the migration happened and succeeded. false otherwise.
      */
-    public boolean migrateEncryptionKeyIfNeeded(@NonNull final String callingPackageName) throws GeneralSecurityException, IOException {
-        if (!(COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(callingPackageName) ||
-                AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(callingPackageName))) {
-            // Caller is not broker. No need to migrate.
+    public boolean migrateEncryptionKeyIfNeeded(@NonNull final String activeBrokerPackageName) throws GeneralSecurityException, IOException {
+        final String methodName = ":migrateEncryptionKeyIfNeeded";
+
+        if (!(COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(activeBrokerPackageName) ||
+                AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(activeBrokerPackageName))) {
+            Logger.verbose(TAG + methodName, "Caller is not a broker. Migration is not needed.");
             return false;
         }
 
         mEncryptionKey = getKey();
         if (mEncryptionKey != null){
+            Logger.verbose(TAG + methodName, "Key already exists. Migration is not needed.");
             return false;
         }
 
-        mEncryptionKey = getKeyFromInactiveBroker(callingPackageName);
-        return mEncryptionKey != null;
+        mEncryptionKey = getKeyFromInactiveBroker(activeBrokerPackageName);
+        if (mEncryptionKey != null){
+            Logger.verbose(TAG + methodName, "Key is successfully retrieved from inactive broker. Saving key...");
+            saveSecretKey(mEncryptionKey);
+            return true;
+        }
+
+        Logger.verbose(TAG + methodName, "Key migration failed.");
+        return false;
     }
 
-
+    @Nullable
     protected SecretKey getKeyFromInactiveBroker(@NonNull final String callingPackageName) {
         final String methodName = ":getKeyFromInactiveBroker";
-        String serializedKey = InactiveBrokerClient.getSerializedKeyFromInactiveBroker(mContext, callingPackageName);
+        final String serializedKey = InactiveBrokerClient.getSerializedSymmetricKeyFromInactiveBroker(mContext, callingPackageName);
 
         if (serializedKey == null || serializedKey.length() == 0) {
             Logger.verbose(TAG + methodName, "The returned bundle doesn't contain any key.");
@@ -541,6 +552,16 @@ public class StorageHelper implements IStorageHelper {
         }
 
         return deserializeSecretKey(serializedKey);
+    }
+
+    private void saveSecretKey(@NonNull SecretKey key) throws GeneralSecurityException, IOException {
+        final String methodName = ":saveSecretKey";
+
+        Logger.verbose(TAG + methodName, "Saving secret key to storage.");
+
+        mKeyPair = generateKeyPairFromAndroidKeyStore();
+        final byte[] keyWrapped = wrap(key);
+        writeKeyData(keyWrapped);
     }
 
     /**
@@ -563,12 +584,8 @@ public class StorageHelper implements IStorageHelper {
 
         if (mSecretKeyFromAndroidKeyStore == null) {
             // If encountering exception for reading keys, try to generate new keys
-            mKeyPair = generateKeyPairFromAndroidKeyStore();
-
-            // Also generate new secretkey
             mSecretKeyFromAndroidKeyStore = generateSecretKey();
-            final byte[] keyWrapped = wrap(mSecretKeyFromAndroidKeyStore);
-            writeKeyData(keyWrapped);
+            saveSecretKey(mSecretKeyFromAndroidKeyStore);
         }
 
         return mSecretKeyFromAndroidKeyStore;
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -354,6 +354,7 @@ public class StorageHelper implements IStorageHelper {
      **/
     private List<KeyType> initializeDecryptionKeyTypeList(@NonNull final String encryptedBlob,
                                                           @NonNull final String packageName) throws IOException {
+        final String methodName = "initializeDecryptionKeyTypeList";
         List<KeyType> keyTypeList = new ArrayList<>();
 
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
@@ -372,10 +373,12 @@ public class StorageHelper implements IStorageHelper {
             keyTypeList.add(KeyType.KEYSTORE_ENCRYPTED_KEY);
         }
 
+        Logger.verbose(TAG + methodName, "Decryption key list's size = " + keyTypeList.size());
         return keyTypeList;
     }
 
-    private String decryptWithSecretKey(final byte[] bytes,
+    @NonNull
+    private String decryptWithSecretKey(@NonNull final byte[] bytes,
                                         @NonNull final SecretKey secretKey)
             throws GeneralSecurityException, IOException {
         final String methodName = "decryptWithSecretKey";
@@ -472,6 +475,7 @@ public class StorageHelper implements IStorageHelper {
      * 1. Keystore-encrypted symmetric key.
      * 2. a newly-generated key.
      */
+    @NonNull
     private SecretKey loadSecretKeyForBrokerEncryption() throws IOException, GeneralSecurityException {
         setBlobVersion(VERSION_ANDROID_KEY_STORE);
         SecretKey key = loadSecretKey(KeyType.KEYSTORE_ENCRYPTED_KEY);
@@ -489,6 +493,7 @@ public class StorageHelper implements IStorageHelper {
      * 1. User-defined key.
      * 2. Keystore-encrypted symmetric key.
      */
+    @NonNull
     private SecretKey loadSecretKeyForADALMSALEncryption() throws IOException, GeneralSecurityException {
         if (AuthenticationSettings.INSTANCE.getSecretKeyData() != null) {
             setBlobVersion(VERSION_USER_DEFINED);
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -465,7 +465,7 @@ public class StorageHelper implements IStorageHelper {
     /**
      * Given the key type, load a secret key.
      */
-    protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOException, GeneralSecurityException {
+    public SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOException, GeneralSecurityException {
         final String methodName = "loadSecretKey";
 
         switch (keyType) {
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -303,7 +303,7 @@ public class StorageHelper implements IStorageHelper {
             }
         }
 
-        Logger.verbose(
+        Logger.info(
                 TAG + methodName,
                 "Tried all decryption keys and decryption still fails. Throw an exception.");
 
@@ -320,8 +320,8 @@ public class StorageHelper implements IStorageHelper {
         final byte[] bytes;
         try {
             bytes = getByteArrayFromEncryptedBlob(data);
-        } catch (RuntimeException e) {
-            Logger.error(TAG + methodName, "This data is not an encrypted blob.", e);
+        } catch (IllegalArgumentException e) {
+            Logger.verbose(TAG + methodName, "This data is not an encrypted blob.");
             return EncryptionType.UNENCRYPTED;
         }
 
@@ -362,7 +362,7 @@ public class StorageHelper implements IStorageHelper {
     @NonNull
     public List<KeyType> getKeysForDecryptionType(@NonNull final String encryptedBlob,
                                                   @NonNull final String packageName) throws IOException {
-        final String methodName = ":initializeDecryptionKeyTypeList";
+        final String methodName = ":getKeysForDecryptionType";
         List<KeyType> keyTypeList = new ArrayList<>();
 
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
@@ -511,6 +511,7 @@ public class StorageHelper implements IStorageHelper {
                 return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
 
             case KEYSTORE_ENCRYPTED_KEY:
+                Logger.verbose(TAG + methodName, "Loading KeyStore encrypted key.");
                 return loadKey();
         }
 
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -28,9 +28,7 @@ import android.content.Context;
 import android.os.Build;
 import android.security.KeyPairGeneratorSpec;
 import android.support.annotation.NonNull;
-import android.support.annotation.Nullable;
 import android.util.Base64;
-import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
@@ -45,6 +43,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.security.DigestException;
 import java.security.GeneralSecurityException;
@@ -57,11 +56,11 @@ import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.spec.AlgorithmParameterSpec;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 
 import javax.crypto.Cipher;
 import javax.crypto.KeyGenerator;
@@ -145,10 +144,15 @@ public class StorageHelper implements IStorageHelper {
 
     private static final String ANDROID_KEY_STORE = "AndroidKeyStore";
 
-    private static final Set<String> BROKER_PACKAGE_NAMES = new HashSet<String>() {{
-        add(AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
-        add(COMPANY_PORTAL_APP_PACKAGE_NAME);
-    }};
+    /**
+     * Type of Secret key to be used.
+     */
+    public enum KeyType {
+        LEGACY_AUTHENTICATOR_APP_KEY,
+        LEGACY_COMPANY_PORTAL_KEY,
+        ADAL_USER_DEFINED_KEY,
+        KEYSTORE_ENCRYPTED_KEY
+    }
 
     private final Context mContext;
     private final SecureRandom mRandom;
@@ -158,8 +162,8 @@ public class StorageHelper implements IStorageHelper {
      */
     private KeyPair mKeyPair;
     private String mBlobVersion;
-    private SecretKey mKey = null;
-    private SecretKey mHMACKey = null;
+    private SecretKey mEncryptionKey = null;
+    private SecretKey mEncryptionHMACKey = null;
     private SecretKey mSecretKeyFromAndroidKeyStore = null;
 
     /**
@@ -177,17 +181,19 @@ public class StorageHelper implements IStorageHelper {
     @Override
     public String encrypt(final String clearText)
             throws GeneralSecurityException, IOException {
-        Log.v(TAG, "Starting encryption");
+        final String methodName = ":readKeyData";
+
+        Logger.verbose(TAG + methodName, "Starting encryption");
 
         if (StringExtensions.isNullOrBlank(clearText)) {
             throw new IllegalArgumentException("Input is empty or null");
         }
 
         // load key for encryption if not loaded
-        mKey = loadSecretKeyForEncryption();
-        mHMACKey = getHMacKey(mKey);
+        mEncryptionKey = loadSecretKeyForEncryption();
+        mEncryptionHMACKey = getHMacKey(mEncryptionKey);
 
-        Log.v(TAG, "Encrypt version:" + mBlobVersion);
+        Logger.verbose(TAG + methodName, "Encrypt version:" + mBlobVersion);
         final byte[] blobVersion = mBlobVersion.getBytes(AuthenticationConstants.ENCODING_UTF8);
         final byte[] bytes = clearText.getBytes(AuthenticationConstants.ENCODING_UTF8);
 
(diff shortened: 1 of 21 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -488,6 +489,40 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+            // Caller is not broker. No need to migrate.
+            return false;
+        }
+
+        mEncryptionKey = getKey();
+        if (mEncryptionKey != null){
+            return false;
+        }
+
+        mEncryptionKey = getKeyFromInactiveBroker(callingPackageName);
```
sdabbiru: >mEncryptionKey [](start = 8, length = 14)

What happens to mEncryptionKey? Is this persisted somewhere? or do we fetch each time?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+
+        if (mBrokerAccountServiceIntent == null) {
+            throw new ClientException("mBrokerAccountServiceIntent is null. BrokerAccountService.");
+        }
+
+        BrokerAccountServiceFuture future = new BrokerAccountServiceFuture();
+        mBrokerAccountServiceConnection = new BrokerAccountServiceConnection(future);
+
+        mBound = mContext.bindService(mBrokerAccountServiceIntent, mBrokerAccountServiceConnection, Context.BIND_AUTO_CREATE);
+        Logger.verbose(TAG + "connect", "The status for MicrosoftAuthService bindService call is: " + Boolean.valueOf(mBound));
```
sdabbiru: >MicrosoftAuthService [](start = 56, length = 20)

BrokerAccountService?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final String BROKER_ACCOUNT_SERVICE_CLASS_NAME = "com.microsoft.aad.adal.BrokerAccountService";
+
+    private Context mContext;
+    private String mCallingPackageName;
+    private BrokerAccountServiceConnection mBrokerAccountServiceConnection;
+    private Intent mBrokerAccountServiceIntent;
+    private Boolean mBound = false;
+
+    // Must be called from BG thread.
+    public static String getSerializedKeyFromInactiveBroker(@NonNull Context context,
```
kreedula: Should we rename `getSymmetricKeyFromInactiveBroker` ? and add a comment about decoding/serializing where we convert to `SecretKey`

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+                                                            @NonNull final String callingPackageName) {
+        final String methodName = ":getKeyFromInactiveBroker";
+
+        final InactiveBrokerClient client = new InactiveBrokerClient(context, callingPackageName);
+        try {
+            final BrokerAccountServiceFuture brokerAccountServiceFuture = client.connect();
+            IBrokerAccountService service = brokerAccountServiceFuture.get();
+
+            Bundle resultBundle = service.getInactiveBrokerKey();
```
kreedula: Nit : final

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+
+            // This means that the inactive broker doesn't support getInactiveBrokerKey().
+            if (resultBundle == null){
+                Logger.verbose(TAG + methodName, "resultBundle is null");
```
kreedula: should be log the package name (which broker i mean)  and also version here  ?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+                return null;
+            }
+
+            return resultBundle.getString(AuthenticationConstants.Broker.BROKER_KEYSTORE_SYMMETRIC_KEY);
+
+        } catch (final BaseException | InterruptedException | ExecutionException | RemoteException e) {
+            Logger.error(
+                    TAG + methodName,
```
kreedula: Same here, log broker name and version ?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
kreedula: typo, rename `MicrosoftAuthService`

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @return MicrosoftAuthServiceFuture
+     */
+    private BrokerAccountServiceFuture connect() throws ClientException {
+
+        if (mBrokerAccountServiceIntent == null) {
+            throw new ClientException("mBrokerAccountServiceIntent is null. BrokerAccountService.");
+        }
+
+        BrokerAccountServiceFuture future = new BrokerAccountServiceFuture();
```
kreedula: nit : final

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+     * Gets the intent that points to the bound service on the device... if available
+     * You shouldn't get this far if it's not available
+     *
+     * @param context
+     * @return Intent
+     */
+    private Intent getIntentForBrokerAccountService(final Context context) {
+        final String inactiveBrokerPackageName = getInactiveBrokerPackageName(context);
+
+        if (inactiveBrokerPackageName == null || inactiveBrokerPackageName.length() == 0) {
```
kreedula: can be simplified to `TextUtils.isEmpty(inactiveBrokerPackageName)`

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+    private String getInactiveBrokerPackageName(@NonNull final Context context) {
+        final String methodName = ":getInactiveBrokerPackageName";
+
+        final String inactiveBrokerPackageName;
+        if (AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(mCallingPackageName)) {
+            inactiveBrokerPackageName = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
+        } else if (AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(mCallingPackageName)) {
+            inactiveBrokerPackageName = AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+        } else {
+            return null;
```
kreedula: We should never get into this state, but lets log the unknown package name here

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        // Verify the signature to make sure that we're not binding to malicious apps.
+        final BrokerValidator validator = new BrokerValidator(context);
+        if (validator.verifySignature(inactiveBrokerPackageName)) {
+            Logger.verbose(TAG + methodName, "inactive broker package found:" + inactiveBrokerPackageName);
+            return inactiveBrokerPackageName;
+        }
+
+        Logger.verbose(TAG + methodName, "Inactive broker package not found.");
```
kreedula: same here let's log the package name and signature here and also change level to `error`

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerAccountServiceFuture.java
```diff
@@ -0,0 +1,75 @@
...
+    public IBrokerAccountService get(long l, TimeUnit timeUnit) throws InterruptedException, ExecutionException, TimeoutException {
+        if (mCountDownLatch.await(l, timeUnit)) {
+            return mBrokerAccountService;
+        } else {
+            throw new TimeoutException();
+        }
+
+    }
+
+    public void setBrokerAccountService(IBrokerAccountService result) {
```
kreedula: nit : rename variable to `brokerAccountService`

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: My bad. The key storing part got lost during code refactoring. Will fix.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -488,6 +489,40 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+        return mEncryptionKey != null;
+    }
+
+
+    protected SecretKey getKeyFromInactiveBroker(@NonNull final String callingPackageName) {
```
kreedula: what do you think of renaming `callingPackageName` to `activeBrokerPackageName` ?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private Context mContext;
+    private String mCallingPackageName;
+    private BrokerAccountServiceConnection mBrokerAccountServiceConnection;
+    private Intent mBrokerAccountServiceIntent;
+    private Boolean mBound = false;
+
+    // Must be called from BG thread.
+    public static String getSerializedKeyFromInactiveBroker(@NonNull Context context,
+                                                            @NonNull final String callingPackageName) {
```
kreedula: what do you think of renaming `callingPackageName` to `activeBrokerPackageName` ?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+        final String methodName = ":getKeyFromInactiveBroker";
```
kreedula: method name here `getKeyFromInactiveBroker` needs to be fixed

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
rpdome: adding package name logging at the top of the function instead.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
alanlaims: Nit: a `@Nullable` annotation could be added here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -488,6 +489,40 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+        final String methodName = ":getKeyFromInactiveBroker";
+        String serializedKey = InactiveBrokerClient.getSerializedKeyFromInactiveBroker(mContext, callingPackageName);
```
alanlaims: Nit: a `final` could be added here.

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+                                 @NonNull final String callingPackageName) {
+        mContext = context;
+        mCallingPackageName = callingPackageName;
+        mBrokerAccountServiceIntent = getIntentForBrokerAccountService(mContext);
+    }
+
+    /**
+     * Binds to the service and returns a future that provides the proxy for the calling the BrokerAccountService
+     *
+     * @return MicrosoftAuthServiceFuture
```
alanlaims: nit: rename to `BrokerAccountServiceFuture`

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    private BrokerAccountServiceFuture connect() throws ClientException {
```
alanlaims: nit: Could also add a `@NonNull` annotation.

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Gets the intent that points to the bound service on the device... if available
+     * You shouldn't get this far if it's not available
+     *
+     * @param context
+     * @return Intent
+     */
+    private Intent getIntentForBrokerAccountService(final Context context) {
```
alanlaims: nit: can add a `@Nullable` annotation here.

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
rpdome: Renamed to getSerializedSymmetricKeyFromInactiveBroker

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
rpdome: If you're referring to the inactive broker's name, it's already logged in getInactiveBrokerPackageName().

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -530,6 +540,65 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+    public void migrateEncryptionKeyIfNeeded() throws GeneralSecurityException, IOException {
+        final String methodName = ":migrateEncryptionKeyIfNeeded";
+
+        final String activeBrokerPackageName = getPackageName();
+
+        if (!(COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(activeBrokerPackageName) ||
+                AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(activeBrokerPackageName))) {
+            final String errorMessage = "Caller is not a broker. Migration is not needed";
+            Logger.error(TAG + methodName, errorMessage, null);
+            throw new InvalidParameterException(errorMessage);
```
shahzaibj: This method does not accept any parameters yet we throw an `InvalidParameterException`. May be we can throw some other exception that will be better suited here. I'll see if I can find any existing java exception that will fit our needs.

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Constructor for the BrokerAccountServiceClient
+     *
+     * @param context
+     * @param callingPackageName name of the calling package.
+     */
+    private InactiveBrokerClient(@NonNull final Context context,
+                                 @NonNull final String callingPackageName) {
```
shahzaibj: `callingPackageName` has been renamed to `activeBrokerPackageName` in other places. May be we can also make that change here.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
iambmelt: Good alternatives here might be `IllegalStateException` or `UnsupportedOperationException`

kreedula: yes `IllegalStateException` might be good here

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -49,6 +50,19 @@
...
+    /**
+     * Constructs an instance of SharedPreferencesFileManager with the default shared preference.
+     *
+     * @param context Interface to global information about an application environment.
+     */
+    public SharedPreferencesFileManager(
+            final Context context) {
+        Logger.verbose(TAG, "Init: " + TAG);
+        mSharedPreferencesFileName = "default";
+        mSharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
```
rpdome: @kreedula

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -531,6 +541,65 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+
+        return deserializeSecretKey(serializedKey);
+    }
+
+    private void saveSecretKey(@NonNull SecretKey key) throws GeneralSecurityException, IOException {
+        final String methodName = ":saveSecretKey";
+
+        Logger.verbose(TAG + methodName, "Saving secret key to storage.");
+
+        mKeyPair = generateKeyPairFromAndroidKeyStore();
```
sdabbiru: >mKeyPair [](start = 8, length = 8)

Shouldn't we check if mKeyPair is not null before generating a new keypair in Android key store?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -550,25 +619,30 @@ protected synchronized SecretKey loadOrCreateKey()
...
         }
 
         return mSecretKeyFromAndroidKeyStore;
     }
 
+    /**
+     * Generate a new key and save to storage.
+     */
+    private synchronized SecretKey createKey() throws GeneralSecurityException, IOException {
+        mSecretKeyFromAndroidKeyStore = generateSecretKey();
```
sdabbiru: >mSecretKeyFromAndroidKeyStore [](start = 8, length = 29)

mSecretKeyFromAndroidKeyStore is somewhat misleading - it is not coming from the Android key store... Consider renaming it.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -620,22 +698,29 @@ private synchronized KeyPair generateKeyPairFromAndroidKeyStore()
...
-                    KEY_STORE_CERT_ALIAS, null);
-        } catch (final RuntimeException e) {
+            final KeyStore keyStore = KeyStore.getInstance(ANDROID_KEY_STORE);
+            keyStore.load(null);
+
+            final Certificate cert = keyStore.getCertificate(KEY_STORE_CERT_ALIAS);
+            final Key privateKey = keyStore.getKey(KEY_STORE_CERT_ALIAS, null);
+
+            if (cert == null || privateKey == null) {
+                Logger.verbose(TAG + methodName, "Key entry doesn't exist.");
```
sdabbiru: >Logger [](start = 16, length = 6)

I want this to be in our telemetry stream.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -644,34 +729,6 @@ private synchronized KeyPair readKeyPair() throws GeneralSecurityException, IOEx
             // handle it as regular KeyStoreException
             throw new KeyStoreException(e);
```
sdabbiru: >KeyStoreException [](start = 22, length = 17)

This too in telemetry please.

sdabbiru:
>             Logger.verbose(TAG + methodName, "Removed previous key pair info.");

Please put this in the telemetry.

---
Refers to: common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java:831 in 9323db7. [](commit_id = 9323db79a3a7f47aba7520881a05afdfe50b0806, deletion_comment = False)

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: Technically it's not needed (given that this function will be called only when there is no existing secretKey (after it gets a key from IPC, or created a new one).

I'll add comments to clarify this.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -531,6 +541,65 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+            return;
+        }
+
+        Logger.verbose(TAG + methodName, "Key migration failed. Create a new key.");
+        mEncryptionKey = createKey();
+        return;
+    }
+
+    @Nullable
+    protected SecretKey getKeyFromInactiveBroker(@NonNull final String activeBrokerPackageName) {
```
sdabbiru: >getKeyFromInactiveBroker [](start = 24, length = 24)

This naming is throwing my mind off - Are we supposed to get the key from the current inactive broker? If so, why is activeBrokerPackageName has getPackageName() output value (Line 553)?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
sdabbiru: I want to keep the code that initializes member variables together. Otherwise, it is too hard to read.

---
In reply to: [301375521](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/547#discussion_r301375521) [](ancestors = 301375521)

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: It takes activeBrokerPackageName to help determine what could possibly be the inactive broker.
i.e. if activeBrokerPackageName = CP, then inactive broker = authApp

Would adding more comment, or renaming help? Do you have any suggestion?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
shahzaibj: @rpdome unresolving this as this is still not fixed after the `Resolve comments` commit

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
rpdome: same comment as above.

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
shahzaibj: @rpdome unresolving this as this is still not fixed after the `Resolve comments` commit

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
shahzaibj: @rpdome unresolving this as this is still not fixed after the `Resolve comments` commit

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
shahzaibj: @rpdome unresolving this as this is still not fixed after the `Resolve comments` commit

rpdome: This might not be an error though, as the target broker app might not exist.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: I decided to put the check as originally suggested to minimize the change - as we want to push this out asap.

It's hard to group initializer together because they are being generated in different places. In this case, the secretKey is obtained **from the other broker**, while this keypair is generated locally.

I was (in the process of) refactoring this layer to make it easier to read (i.e. extracting KeyStore and SecretKey wrapper with internal cache, meaning all this layer would have to do is to call wrapper.get())

However, that would require more change, and therefore might not be worth doing now.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: removed

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: Telemetry is not wired into common yet. We could add this after Heidi has finished wiring her new telemetry work.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: Telemetry is not wired into common yet. We could add this after Heidi has finished wiring her new telemetry work.

rpdome: Best thing I can do here is to bubble up the exception, and have the upper layer (AccountManagerStorageHelper) log the telemetry. Would that work?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: bubbled up the exception

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -524,13 +540,73 @@ protected SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExcep
...
+    /**
+     * If needed, get the key from inactive broker (if there is one).
+     * If it fails to get a new key. It will create a new one.
+     *
+     * This function can only be invoked by Broker apps.
+     */
+    public void migrateEncryptionKeyIfNeeded() throws GeneralSecurityException, IOException {
+        final String methodName = ":migrateEncryptionKeyIfNeeded";
+
+        if (!(COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(getPackageName()) ||
```
titorodi: These are only called from broker auth paths right? As at least Company Portal currently does local auth, it wants to do broker auth ...

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerAccountServiceConnection.java
```diff
@@ -0,0 +1,69 @@
...
+        Logger.verbose(TAG, "BrokerAccountService is connected.");
+        mBrokerAccountService = IBrokerAccountService.Stub.asInterface(service);
+        mBrokerAccountServiceFuture.setBrokerAccountService(mBrokerAccountService);
+    }
+
+    /*
+
+    //Needed for API level 26 and above
+    @Override
+    public void onBindingDied(ComponentName name){
```
titorodi: ? Not needed though?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -0,0 +1,179 @@
...
+    private BrokerAccountServiceConnection mBrokerAccountServiceConnection;
+    private Intent mBrokerAccountServiceIntent;
+    private Boolean mBound = false;
+
+    // Must be called from BG thread.
+    public static String getSerializedSymmetricKeyFromInactiveBroker(@NonNull Context context,
+                                                                     @NonNull final String activeBrokerPackageName) {
+        final String methodName = ":getSerializedSymmetricKeyFromInactiveBroker";
+
+        final InactiveBrokerClient client = new InactiveBrokerClient(context, activeBrokerPackageName);
```
titorodi: Is the ActiveBrokerPackageName ever not the current package name? I think this is based on the design supports push or pull models, so we don't care? Or if we're deciding on pull, we should at least log or throw in the unexpected case of them not matching?

rpdome: It is the current package name. I just expose this (instead of directly pulling it from context), for mock testing).

I could rename it if this is confusing.

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: Migration should only be called by Broker. yes.

PR #549, opened 2019-07-09
Integrate the protocol verification with MSAL | part1/3

heidijinxujia:


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -623,7 +623,7 @@ private AuthenticationConstants() {
         /**
          * The maximum broker protocol version that common supports.
          */
-        public static final String BROKER_PROTOCOL_VERSION_CODE = "2.0";
+        public static final String BROKER_PROTOCOL_VERSION_CODE = "2.0"; //TODO need to update the broker propocol version here.
```
shoatman: I have a PR to do this... but let's abandon mine... and you just do this as part of this PR.  Thanks Heidi.

heidijinxujia: K. I was thinking to wait for your change to merge and then rebase.

PR #550, opened 2019-07-09
Fix auth url containing mTokenScope

shahzaibj:
Link to GitHub Issue: https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/674

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
@@ -133,11 +133,7 @@ public abstract class BaseController {
                         .setMultipleCloudAware(requestAuthority.mMultipleCloudsSupported)
                         .setSlice(requestAuthority.mSlice);
             }
-
-            if(builder instanceof MicrosoftStsAuthorizationRequest.Builder){
-                ((MicrosoftStsAuthorizationRequest.Builder)builder).setTokenScope(TextUtils.join(" ", parameters.getScopes()));
-            }
-
+            
             if (acquireTokenOperationParameters.getExtraScopesToConsent() != null) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -133,11 +133,7 @@ public abstract boolean removeAccount(final OperationParameters parameters)
                         .setMultipleCloudAware(requestAuthority.mMultipleCloudsSupported)
                         .setSlice(requestAuthority.mSlice);
             }
-
-            if(builder instanceof MicrosoftStsAuthorizationRequest.Builder){
-                ((MicrosoftStsAuthorizationRequest.Builder)builder).setTokenScope(TextUtils.join(" ", parameters.getScopes()));
-            }
-
+
```
iambmelt: Not quite this easy, I'm afraid 😞 

Resolving this issue by removing this setter call reintroduces the bug reported in #524. I suggest taking a look at #527 to see why this was introduced.

In essence, what we need to do is not include this value _in the authorization URL_ but still have it available on this object. I suspect this means marking `mTokenScope` as `transient` so that it is not serialized into the request, rather than removing the setter call to it.

shahzaibj: Pushed the fix. Thanks

iambmelt: Thanks @shahzaibj - Can I get you to run the Android Studio code formatter and remove the stray whitespace in this file?

(It's in the top-menu under `Code` > `Reformat Code`)

I don't see any import issues, but you could run `Optimize Imports` as well, good habit for code cleanliness in general

shahzaibj: Pushed

PR #557, opened 2019-07-12
Remove the duplicate of the auth query parameters

heidijinxujia:
Fix for issue https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/688

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
@@ -141,8 +141,21 @@ public class MicrosoftStsAuthorizationRequest extends MicrosoftAuthorizationRequ
 
     @Override
     public Uri getAuthorizationRequestAsHttpRequest() {
-        Map<String, Object> qpMap = new HashMap<>();
+        final Map<String, Object> qpMap = new HashMap<>();
         qpMap.putAll(ObjectMapper.serializeObjectHashMap(this));
+
+        for (Map.Entry<String, String> entry : mFlightParameters.entrySet()) {
+            qpMap.put(entry.getKey(), entry.getValue());
+        }
+
+        if (mSlice != null) {
+            if (!TextUtils.isEmpty(mSlice.getSlice())) {
+                qpMap.put(AzureActiveDirectorySlice.SLICE_PARAMETER, mSlice.getSlice());
+            }
+            if (!TextUtils.isEmpty(mSlice.getDC())) {
+                qpMap.put(AzureActiveDirectorySlice.DC_PARAMETER, mSlice.getDC());
+            }
+        }
 
         // Add extra qp, if present...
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -153,4 +154,25 @@ public static int compareSemanticVersion(final String thisVersion, final String
...
+     * @param str
+     * @param subString
+     * @return int
+     */
+    public static int countMatches(@NonNull String str, @Nullable String subString) {
+        int count = 0;
+        if (StringUtil.isEmpty(str) || StringUtil.isEmpty(subString)) {
+            return count;
+        }
+        for(int i = 0; i < (str.length() - subString.length()); i++) {
```
shahzaibj: There is an off by one error here. Need `less than or equal to comparison <=` to get accurate count

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -139,32 +140,43 @@ public String getPrompt() {
...
 
     @Override
-    public Uri getAuthorizationRequestAsHttpRequest() throws UnsupportedEncodingException {
-        Uri.Builder uriBuilder = Uri.parse(getAuthorizationEndpoint()).buildUpon();
-        for (Map.Entry<String, Object> entry : ObjectMapper.serializeObjectHashMap(this).entrySet()) {
-            uriBuilder.appendQueryParameter(entry.getKey(), entry.getValue().toString());
-        }
+    public Uri getAuthorizationRequestAsHttpRequest() {
+        Map<String, Object> qpMap = new HashMap<>();
+        qpMap.putAll(ObjectMapper.serializeObjectHashMap(this));
```
Om83: I am assuming that this is the behavior for iOS, where if the qmap already has the item it is prioritized over the similar  query parameter item.

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -153,4 +154,25 @@ public static int compareSemanticVersion(final String thisVersion, final String
...
+            if(String.copyValueOf(str.toCharArray(), i, subString.length()).equalsIgnoreCase(subString)) {
```
shahzaibj: Why are we converting string to char array and then using String.copyValueOf? The string `substring` method can achieve the same result with less work I believe. Is there a reason you chose to do it this way?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
heidijinxujia: No, actually, iOS does not do any thing on the MSAL calls. iOS only whitelisted all the extra query parameters when doing the broker calls, including authorization request and token request.

rpdome:
Once you're done, can you bump the common version to -RC12? Thanks!

PR #561, opened 2019-07-22
Migrate Key only + Inject telemetry

rpdome:
- Inject telemetry.
- Revert the code to migrate key only.
- Move some of the function to ad-accounts.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -83,7 +83,7 @@ public class StorageHelper implements IStorageHelper {
     /**
      * A flag to turn on/off keystore encryption on Broker apps.
      */
-    public static final boolean SHOULD_ENCRYPT_BROKER_VALUES_WITH_KEYSTORE_ENCRYPTED_KEY = false;
+    public static final boolean sShouldEncryptWithKeyStoreKey = false;
 
     /**
      * HMac key hashing algorithm.
@@ -364,7 +364,6 @@ public class StorageHelper implements IStorageHelper {
     @NonNull
     public List<KeyType> getKeysForDecryptionType(@NonNull final String encryptedBlob,
                                                   @NonNull final String packageName) throws IOException {
-        final String methodName = ":getKeysForDecryptionType";
         List<KeyType> keyTypeList = new ArrayList<>();
 
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
@@ -383,7 +382,6 @@ public class StorageHelper implements IStorageHelper {
             keyTypeList.add(KeyType.KEYSTORE_ENCRYPTED_KEY);
         }
 
-        Logger.verbose(TAG + methodName, "Decryption key list's size = " + keyTypeList.size());
         return keyTypeList;
     }
 
@@ -469,14 +467,14 @@ public class StorageHelper implements IStorageHelper {
         }
 
         // The current app runtime is the broker; load its secret key.
-        if (!SHOULD_ENCRYPT_BROKER_VALUES_WITH_KEYSTORE_ENCRYPTED_KEY &&
+        if (!sShouldEncryptWithKeyStoreKey &&
                 AuthenticationSettings.INSTANCE.getBrokerSecretKeys().containsKey(getPackageName())) {
 
             // Try to read keystore key - so that we get telemetry data on its reliability.
             // If anything happens, do not crash the app.
             try {
                 loadSecretKey(KeyType.KEYSTORE_ENCRYPTED_KEY);
-            } catch (Exception e){
+            } catch (Exception e) {
                 // Best effort.
             }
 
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -78,13 +78,12 @@ public BrokerValidator(final Context context) {
...
-     * If signature hash verification fails, this will throw a Client exception, with its error message being a list of hashes.
+     * If signature hash verification fails, this will throw a Client exception containing the cause of error, which could contain a list of mismatch hashes.
      *
      * @param brokerPackageName The broker package to inspect.
-     * @return True if the certificate hash is known. False otherwise.
+     * @return SignatureHash of brokerPackageName, if the verification succeeds..
      */
-    public boolean verifySignatureAndThrowIfHashVerificationFailed(final String brokerPackageName) throws ClientException {
-        final String methodName = ":verifySignatureAndThrowIfHashVerificationFailed";
+    public String verifySignatureAndThrow(final String brokerPackageName) throws ClientException {
```
sdabbiru: >verifySignatureAndThrow [](start = 18, length = 23)

I wish we log telemetry in this method instead of anywhere , if the signature validation fails...

rpdome: Once telemetry is wired into common, that'll be doable. As of now we'll have to pass it to ad-accounts :(

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -497,147 +526,108 @@ protected void setBlobVersion(@NonNull String blobVersion) {
...
-            Logger.verbose(TAG + methodName, "The returned bundle doesn't contain any key.");
-            return null;
-        }
-
-        return deserializeSecretKey(serializedKey);
-    }
-
-    private void saveSecretKey(@NonNull SecretKey key) throws GeneralSecurityException, IOException {
+    public void saveKeyStoreEncryptedKey(@NonNull SecretKey unencryptedKey) throws GeneralSecurityException, IOException {
         final String methodName = ":saveSecretKey";
```
alanlaims: Nit: Could rename this value to `saveKeyStoreEncryptedKey`

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -82,10 +77,14 @@
...
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
 
-
 public class StorageHelper implements IStorageHelper {
     private static final String TAG = "StorageHelper";
 
+    /**
+     * A flag to turn on/off keystore encryption on Broker apps.
+     */
+    public static final boolean SHOULD_ENCRYPT_BROKER_VALUES_WITH_KEYSTORE_ENCRYPTED_KEY = false;
```
kreedula: nit : not a constant so uppercase not required, use variable name like : `sShouldEncryptWithKeyStoreKey`

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -497,147 +526,108 @@ protected void setBlobVersion(@NonNull String blobVersion) {
...
-        return deserializeSecretKey(serializedKey);
-    }
-
-    private void saveSecretKey(@NonNull SecretKey key) throws GeneralSecurityException, IOException {
+    public void saveKeyStoreEncryptedKey(@NonNull SecretKey unencryptedKey) throws GeneralSecurityException, IOException {
         final String methodName = ":saveSecretKey";
 
-        Logger.verbose(TAG + methodName, "Saving secret key to storage.");
+        try {
+            logFlowStart(methodName, AuthenticationConstants.TelemetryEvents.KEYCHAIN_WRITE);
```
kreedula: Why is this a `KEYCHAIN_WRITE` event here? , shouldn't this be when create a asymetric key pair in `generateKeyPairFromAndroidKeyStore`  , this incorrectly could mark any exceptions here in methods `wrap` and `writeKeyData` as key store related failures right ?

common/src/main/java/com/microsoft/identity/common/internal/broker/InactiveBrokerClient.java
```diff
@@ -2,23 +2,13 @@
...
 import android.content.Intent;
-import android.os.Bundle;
-import android.os.Looper;
-import android.os.RemoteException;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
-import android.text.TextUtils;
 
-import com.microsoft.aad.adal.IBrokerAccountService;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
```
alanlaims: Nit: this import is unused.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -0,0 +1,55 @@
...
+    public final String packageName;
+    public final String signatureHash;
+
+    private BrokerData(String packageName, String hash) {
+        this.packageName = packageName;
+        this.signatureHash = hash;
+    }
+
+    /**
+     * Given a broker package name, verify its signature and returns a BrokerData object.
```
Om83: nit:return

PR #564, opened 2019-07-26
Add shouldResolveInterrupt field

rpdome:
related commit: https://github.com/AzureAD/ad-accounts-for-android/pull/1073


common/src/main/java/com/microsoft/identity/common/internal/request/BrokerAcquireTokenOperationParameters.java
```diff
@@ -99,6 +102,14 @@ public void setRequestType(RequestType requestType) {
         this.mRequestType = requestType;
     }
 
+    public Boolean getShouldResolveInterrupt() {
```
kreedula: can we use primitive `boolean` here ? to avoid nulls

PR #565, opened 2019-07-27
Integrate the telemetry events

heidijinxujia:
Update the api events and add constants for telemetry string.

Related PR for MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/692


heidijinxujia:
@iambmelt Thank you, Brian. Fixed.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -129,9 +142,21 @@ public static HttpResponse sendPost(final URL requestUrl, final Map<String, Stri
      */
     public static HttpResponse sendGet(final URL requestUrl, final Map<String, String> requestHeaders)
             throws IOException {
+        Telemetry.emit(
+                new HttpStartEvent()
+                        .putMethod(REQUEST_METHOD_GET)
+                        .putPath(requestUrl)
+                        .putRequestIdHeader(requestHeaders.get("client-request-id"))
```
iambmelt: Is there a constant we can use here?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/HttpStartEvent.java
```diff
@@ -35,23 +37,28 @@ public HttpStartEvent putMethod(String method) {
...
 
-    public HttpStartEvent putRequestIdHeader(String requestIdHeader) {
-        put(Key.HTTP_REQUEST_ID_HEADER, requestIdHeader);
-        return this;
-    }
+        // we do not want to send tenant information
+        // index 0 is blank
+        // index 1 is tenant
+        final String[] splitArray = path.getPath().split("/");
+        for (int i = 2; i < splitArray.length; i++) {
```
iambmelt: This looks like a potential hazard for b2c urls where the format is pretty undefined (now that vanity urls are supported) - we may need to take a // TODO for unit testing here.

iambmelt: We could also cross-reference w/ iOS to see what they're doing - we could just flag the whole UJRL as PII/OII and not do any magic on it at all

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
heidijinxujia: Sure.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
```diff
@@ -53,9 +53,9 @@
...
             Key.AUTHORITY,
-            Key.IDP,
-            Key.APPLICATION_NAME,
-            Key.APPLICATION_VERSION
+            Key.IDP_NAME,
+            //App.NAME,
+            //App.VERSION
     };
 
     private TelemetryPiiOiiRules() {
```
iambmelt: Minor complaint about `isPiiOrOii()` impl: rather than duplicate the logic out of `isPii` and `isOii` I'd prefer to see:

```java
    public boolean isPiiOrOii (final String propertyName) {
        return isPii(propertyName) || isOii(propertyName);
    }
```

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -114,9 +117,19 @@ private HttpRequest(final URL requestUrl, final Map<String, String> requestHeade
     public static HttpResponse sendPost(final URL requestUrl, final Map<String, String> requestHeaders,
                                         final byte[] requestContent, final String requestContentType)
             throws IOException {
+        Telemetry.emit(
+                new HttpStartEvent()
+                        .putMethod(REQUEST_METHOD_POST)
+                        .putPath(requestUrl)
```
iambmelt: Question: is this on the pii/oii switch?

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -114,9 +117,19 @@ private HttpRequest(final URL requestUrl, final Map<String, String> requestHeade
...
+                        .putRequestIdHeader(requestHeaders.get("client-request-id"))
```
iambmelt: Ah - same thing here about use of a constant

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
heidijinxujia: Addressed.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/rules/TelemetryPiiOiiRules.java
heidijinxujia: Addressed

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
heidijinxujia: Addressed

iambmelt:
@heidijinxujia Correct me if I'm wrong; looks like there's a handful of scenarios not fully captured here.

#### Caching
- SSO state sync
- ADAL to MSAL upgrade
- FoCI lookups
- `save()` actions

Am I correct in assuming these are TODO?

heidijinxujia:
@iambmelt  Added the issue here. https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/627

iambmelt:
> @iambmelt Added the issue here. #627

Great, thanks for capturing this

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
heidijinxujia: Yes, it is. Addressed.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/events/HttpStartEvent.java
heidijinxujia: iOS is using the absoluteString  of the url. https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/blob/68aa3f4116c818fb4486a56d067641858ee575d5/IdentityCore/src/telemetry/MSIDTelemetryAuthorityValidationEvent.m#L40

iambmelt: Ah - so looks like they're doing nothing fancy at all. That's probably what we should do.

heidijinxujia: The reason is the same with [ADAL telemetry](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/01a117bf3f544d527d4f41a99c5dcb79bc745163/adal/src/telemetry/java/com/microsoft/aad/adal/HttpEvent.java#L64) where we want to filter out the query parameters as well as the tenant info from the path.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -106,4 +102,45 @@ public void process(@NonNull final List<Map<String, String>> rawData) {
...
+                    null == event.get(Key.OCCUR_TIME) ?
+                            null
+                            : event.get(Key.OCCUR_TIME)
+            );
+        }
+    }
+
+    private void calculateEventResponseTime(@NonNull final Map<String, String> responseTimeMap,
+                                            @NonNull final Map<String, String> aggregatedData) {
+        for (Map.Entry<String,String> entry : responseTimeMap.entrySet()) {
```
shahzaibj: **Suggestion:**
Multiple calls are made to entry.getKey()
Instead we can make the call once and store in a variable and use the variable everywhere. Also since there are multiple keys used in this method, one for responseTimeMap and one for aggregatedData, it might also help to store both in a variable with different names to distinguish them

`String responseTimeKey= entry.getKey()
final String aggregatedDataKey = responseTimeKey.replace(START, "response"); `

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -106,4 +102,45 @@ public void process(@NonNull final List<Map<String, String>> rawData) {
...
+            responseTimeMap.put(
+                    eventEndTime,
+                    null == event.get(Key.OCCUR_TIME) ?
+                            null
+                            : event.get(Key.OCCUR_TIME)
+            );
+        }
+    }
+
+    private void calculateEventResponseTime(@NonNull final Map<String, String> responseTimeMap,
```
iambmelt: Finding this function very hard to understand - maybe some javadoc?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -106,4 +102,45 @@ public void process(@NonNull final List<Map<String, String>> rawData) {
...
+                                            @NonNull final Map<String, String> aggregatedData) {
+        for (Map.Entry<String,String> entry : responseTimeMap.entrySet()) {
+            if (entry.getKey().contains(START)
+                    && responseTimeMap.containsKey(entry.getKey().replace(START, END))
```
iambmelt: This `if` is the really challenging part to read here - isn't this check equivalent to....

```java
if (entry.getKey().contains(START) &&
    responseTimeMap.containsKey(END) &&
    responseTimeMap.get(END) != null) {
        // ...
}
```

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -106,4 +102,45 @@ public void process(@NonNull final List<Map<String, String>> rawData) {
...
+                    && responseTimeMap.get(entry.getKey().replace(START, END)) != null) {
+                final String key = entry.getKey().replace(START, "response");
+                final long startTime = Long.parseLong(entry.getValue());
+                final long endTime = Long.parseLong(responseTimeMap.get(entry.getKey().replace(START, END)));
```
iambmelt: This might be easier to read if the calls were less nested

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -23,24 +23,27 @@
 package com.microsoft.identity.common.internal.telemetry;
 
 public final class TelemetryEventStrings {
+    private static final String EVENT_PREFIX = "Microsoft.MSAL.";
```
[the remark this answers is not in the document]
heidijinxujia: You can and in the future plan, it should use the same TelemetryEventStrings for both broker telemetry and msal telemetry as they use common. But Broker telemetry is covered in this PR.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -74,15 +79,18 @@ public boolean isNetworkDisabledFromOptimizations() {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             final UsageStatsManagerWrapper usageStatsManagerWrapper = UsageStatsManagerWrapper.getInstance();
             if (usageStatsManagerWrapper.isAppInactive(mConnectionContext)) {
+                Telemetry.emit((BaseEvent) new BaseEvent().put(TelemetryEventStrings.Key.POWER_OPTIMIZATION, String.valueOf(true)));
```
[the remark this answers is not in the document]
heidijinxujia: The aim of this function `isNetworkDisabledFromOptimizations()` is to check if the network is not functional because of power optimization.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
heidijinxujia: Addressed. Thanks!

common/src/main/java/com/microsoft/identity/common/internal/telemetry/adapter/TelemetryAggregationAdapter.java
heidijinxujia: Addressed, thanks!

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -321,6 +329,7 @@ private void cancelAuthorization() {
         final Intent resultIntent = new Intent();
         resultIntent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
         sendResult(AuthenticationConstants.UIResponse.BROWSER_CODE_CANCEL, resultIntent);
+        Telemetry.emit(new UiEndEvent().isUserCancelled());
```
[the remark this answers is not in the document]
heidijinxujia: Good point! Addressed. Thanks!

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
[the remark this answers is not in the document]
rpdome: Is that the same for telemetry event?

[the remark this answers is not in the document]
heidijinxujia: The telemetry here is to track the power optimization is on. The telemetry for the network unavailable is in the calling method.

heidijinxujia:
@iambmelt Thanks a lot for the reviews, could you have another round of look?

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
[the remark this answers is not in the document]
heidijinxujia: Please let me know if you prefer another naming of the property.

PR #568, opened 2019-07-30
[WIP] Fix the IntuneAppProtectionPolicyRequiredException for token response

heidijinxujia:
In this case, the browser cookie persistent after Company Portal installed if broker enabled the browser access. Thus, the authorization request got succeeded. And the token endpoint returned the AADSTS53005 error. I checked the MSAL code. We did not catch the IntuneAppProtectionPolicyRequiredException exception from the token result. Instead, it returns a general MSALServiceException.


common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -235,6 +240,32 @@ public void run() {
                                         command.getCallback().onCancel();
                                     }
                                 });
+                            } else if (finalException instanceof IntuneAppProtectionPolicyRequiredException) {
```
rpdome: Should this go into ExceptionAdapter?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -105,7 +111,9 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
             );
         }
 
-        return exceptionFromTokenResult(result.getTokenResult());
+        final ServiceException exception = exceptionFromTokenResult(result.getTokenResult());
```
rpdome: nit: revert the change.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -128,6 +132,7 @@ protected MicrosoftAuthorizationRequest(final Builder builder) {
         mCodeChallengeMethod = mPkceChallenge.getCodeChallengeMethod();
         mCodeChallenge = mPkceChallenge.getCodeChallenge();
         mState = generateEncodedState();
+        mClientInfoEnabled = "1";
```
rpdome: can you add some comment specifying what '1' stands for (and what are the possible values)?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -143,7 +143,7 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         headers.putAll(Device.getPlatformIdParameters());
 
         return HttpRequest.sendPost(
-                new URL(mTokenEndpoint),
+                new URL(mTokenEndpoint + "?dc=PROD-WST-TEST1"), //TODO why token request does not take query parameters as authorization request's?
```
rpdome: Is this intentional?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -211,6 +221,43 @@ public static void applyCliTelemInfo(@Nullable final CliTelemInfo cliTelemInfo,
         }
     }
 
+    private static IntuneAppProtectionPolicyRequiredException getIntuneAppProtectionPolicyRequiredException(@NonNull final TokenErrorResponse errorResponse) {
```
rpdome: I noticed there's a class called MsalBrokerResultAdapter.java, which might essentially be doing the same thing as this class in its exception handling function? (Seems like this Intune exception is being handled there)

Should we centralize the code?

rpdome: There is an identical-ish piece of code in PRTController. If you're using this one, can we make that one in PRTController invoke this new function of yours?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
```diff
@@ -235,6 +240,32 @@ public void run() {
...
+                               if (result.getAuthorizationResult() != null
```
heidijinxujia: This is not implemented in ExceptionAdapter, because the properties needed by MAM sdk cannot be retrieved by the authentication result alone.
1. uid and utid from authrization response
2. upn from token response
3. authority url from request parameter

Please let me know if there is a better place to put this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
heidijinxujia: The input parameters for PrtController is different from the msal's.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
heidijinxujia: @rpdome ^

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
heidijinxujia: Let me try to get more details to answer this one better.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ApiDispatcher.java
rpdome: apart from authority, it seems like everything else is read from the token result object.

I brought this up because I noticed that the code is being duplicated in two places, and it might be better to consolidate it if possible. (Personally, I'd prefer passing the request param as an optional parameter of .exceptionFromAcquireTokenResult() if that helps consolidating the code, because that means we need to fix it, we can fix it in one place)

kreedula:
This was specifically removed from generic `ExceptionAdapter` as this only needs to thrown in `Joined account cases. 
See bug : https://github.com/AzureAD/ad-accounts-for-android/issues/1000
PR's to fix the bug : https://github.com/AzureAD/ad-accounts-for-android/pull/1019
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/521

This change will regress the behavior. I don't follow the use case here, why would MSAL make a request if broker is installed and has a joined account in this case ?

rpdome:
> This was specifically removed from generic `ExceptionAdapter` as this only needs to thrown in `Joined account cases.
> See bug : [AzureAD/ad-accounts-for-android#1000](https://github.com/AzureAD/ad-accounts-for-android/issues/1000)
> PR's to fix the bug : [AzureAD/ad-accounts-for-android#1019](https://github.com/AzureAD/ad-accounts-for-android/pull/1019)
> #521
> 
> This change will regress the behavior. I don't follow the use case here, why would MSAL make a request if broker is installed and has a joined account in this case ?

I'm trying to understand the reason we ran into the issue in #521 in the first place. 
- Why are we treating the same error code differently based on the WPJ state? If we are not expecting them in non-joined flow, why are we getting back the error in non-joined flow? 
- Did the MAMSDK issue happened because we didn't have enough information to fill the error object? (Specifically, UPN). If we expand ExceptionAdapter to take all of those necessary values, would that fix the issue?

kreedula:
> > This was specifically removed from generic `ExceptionAdapter` as this only needs to thrown in `Joined account cases.
> > See bug : [AzureAD/ad-accounts-for-android#1000](https://github.com/AzureAD/ad-accounts-for-android/issues/1000)
> > PR's to fix the bug : [AzureAD/ad-accounts-for-android#1019](https://github.com/AzureAD/ad-accounts-for-android/pull/1019)
> > #521
> > This change will regress the behavior. I don't follow the use case here, why would MSAL make a request if broker is installed and has a joined account in this case ?
> 
> I'm trying to understand the reason we ran into the issue in #521 in the first place.
> 
> * Why are we treating the same error code differently based on the WPJ state? If we are not expecting them in non-joined flow, why are we getting back the error in non-joined flow?
> * Did the MAMSDK issue happened because we didn't have enough information to fill the error object? (Specifically, UPN). If we expand ExceptionAdapter to take all of those necessary values, would that fix the issue?

1) This is because the service thinks the device is still registered (Device registration is not tied to user) and throws this `IntunePolicyRequiredException` for non-joined case. This is the same issue Heidi is trying to fix (where service thinks broker is installed and registered due to a browser cookie, even though it's not)

2) We might still get the upn from login hint, but we  would never have the `uid` in the case which is necessary for MAM to resolve this. Unless it's registered we can't get the uid.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
heidijinxujia: Here, the IntuneAppProtectionPolicyRequiredException is constructed by the client info returned from Server's token response. However, in the PRT flow, the IntuneAppProtectionPolicyRequiredException is constructed by the silent request parameters.

heidijinxujia:
@kreedula @rpdome 
The scenario of this issue. 
1. The true mam user joined the device with the broker installed and enable the browser access to the certificate inside the broker settings. 
2. The user uninstalled the broker app. (The certificate is not removed when the broker uninstalled. Because on most OEM Android devices, we do not have a way to clear the WPJ state in the system keychain when CP app was uninstalled. In this case, **System browser can theoretically continue to use the WPJ state when it should not**.)
3. The true mam user opens an MSAL app and trying to login.
4. As the broker is not installed, msal will start the login with the browser.
5. After entering the credentials, as the WPJ state is still true, it continues to the token request.
6. At this point, the server returns the "protection_policy_required" suberror from the token endpoint to the native MSAL app.
7. It is expected to throw the IntunePolicyRequiredException at this point to the MSAL app.
8. The MSAL app calls the MAM sdk and make a silent call to MSAL to get the token.

For this issue, the broker is not installed on the device. And the IntunePolicyRequiredException will only be thrown when getting the "protection_policy_required" suberror.

As to the required fields, like uid and tid, they are included in the ClientInfo from the authorization response returned in **step 5**. To get the client info, the authorization request needs to pass the clientinfo=1.

heidijinxujia:
This issue needs more discussion with server dev. Two ways  to fix the issue.

Fix from the server side. Send the user-agent in the token request where the token endpoint will redirect the uri to the broker installation page when getting the IntuneAppProtectionPolicyRequiredException sub error.

Fix from the client side. When getting the IntuneAppProtectionPolicyRequiredException from token endpoint and a browser interactive flow is happening, the client side need to return the token error response saying the broker is not installed and restart with an embedded web view interactive flow which will trigger the broker installation page at the :auth endpoint.

ukanth:
I'm having same problem on Android where IntuneAppProtectionPolicyRequiredException is been thrown after the call to acquireToken. Is there any way to resolve it ? I'm not clear on "embedded web view interactive flow"

Thanks.

PR #572, opened 2019-08-01
Adding flexibility around pwd_exp return type, string, int, long

iambmelt:
Optional claim may be returned as:
- `"pwd_exp" : 1337`
- `"pwd_exp" : "1337"`

Current code will throw a `ClassCastException`, the PR tries to add some flexibility here.

Sample test here: https://repl.it/repls/ExhaustedExtrasmallNewsaggregator

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
@@ -106,9 +106,9 @@ public abstract class MicrosoftAccount extends BaseAccount {
         mUtid = clientInfo.getUtid();
 
         long mPasswordExpiration = 0;
+        final Object expiry = claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION);
 
-        if (null != claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION)) {
-            final Object expiry = claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION);
+        if (null != expiry) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAccount.java
```diff
@@ -107,8 +107,9 @@ public MicrosoftAccount(@NonNull final IDToken idToken,
 
         long mPasswordExpiration = 0;
 
-        if (!StringExtensions.isNullOrBlank((String) claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION))) {
-            mPasswordExpiration = Long.parseLong((String) claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION));
+        if (null != claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION)) {
```
shahzaibj: It might be better doing something like this, so that we perform claims.get() only once

final Object expiry = claims.get(AzureActiveDirectoryIdToken.PASSWORD_EXPIRATION);
if (expiry != null) {
    mPasswordExpiration = Long.valueOf(expiry.toString());
}

iambmelt: Like this feedback and incorporated into another part of the same method.

iambmelt: Resolved in 83e6e51e3694a8d0a80bf3b1f039f16a32770899

PR #574, opened 2019-08-03
Read user info id from the request bundle on silent request

kreedula:
Icm : https://icm.ad.msft.net/imp/v3/incidents/details/136543718/home

### DIFF
--- a/common/versioning/version.properties
+++ b/common/versioning/version.properties
@@ -1,3 +1,3 @@
 #Thu Aug 02 12:03:16 PDT 2018
-versionName=0.0.12
+versionName=0.0.14-RC1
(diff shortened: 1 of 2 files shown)


common/versioning/version.properties
```diff
@@ -1,3 +1,3 @@
 #Thu Aug 02 12:03:16 PDT 2018
-versionName=0.0.12
+versionName=0.0.14-RC1
```
rpdome: 13?

kreedula: We already used 13 for teams/msal release from `dev` branch. I will use 14 for this and 1.0.0 for msal GA release when cut a release branch

PR #579, opened 2019-08-07
Adding null safety check to avoid crash on EmbeddedWebViewStrategy if broadcast is sent multiple times

kreedula:
- This could potentially happen if the `Activity` is in a stale state in call stack  where `onDestroy` is not called and `broadcastrececiver` is not unregistered.

crash details : https://rink.hockeyapp.net/manage/apps/231071/app_versions/170/crash_reasons/280769560

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
@@ -88,13 +88,18 @@ public class EmbeddedWebViewAuthorizationStrategy<GenericOAuth2Strategy extends 
     public void completeAuthorization(int requestCode, int resultCode, Intent data) {
         if (requestCode == BROWSER_FLOW) {
             if (mOAuth2Strategy != null && mAuthorizationResultFuture != null) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -88,8 +87,21 @@ public EmbeddedWebViewAuthorizationStrategy(@NonNull Activity activity) {
...
+                final AuthorizationResult result = mOAuth2Strategy
+                        .getAuthorizationResultFactory()
+                        .createAuthorizationResult(
+                                resultCode,
+                                data,
+                                mAuthorizationRequest
+                        );
+                mAuthorizationResultFuture.setResult(result);
+            } else {
+                Logger.warn(TAG, "SDK Cancel triggering before request is sent out. " +
```
rpdome: If we get a log from customer, would this line be enough for investigation (as in to know what exactly happened)?

kreedula: Good point, added if `oAuth2Strategy` and `mAuthorizationResultFuture ` are null, i.e. both should be null and there is nothing we need to do as there is no future waiting to be ended, in any case only one of them is null then we have another problem which are not aware of.

PR #581, opened 2019-08-07
Update Travis-CI build SDK version 27 -> 28, Fix AndroidX Compat

iambmelt:
See title + diff


gradle/versions.gradle
```diff
@@ -16,13 +16,13 @@ ext {
     appCompatVersion = "1.0.2"
     browserVersion = "1.0.0"
     dexmakerMockitoVersion = "1.4"
-    espressoCoreVersion = "3.1.0"
```
rpdome: Can you also push this change to android_complete once this PR goes in?

common/build.gradle
```diff
@@ -93,6 +93,7 @@ dependencies {
     testImplementation "org.powermock:powermock-classloading-xstream:$rootProject.ext.powerMockVersion"
 
     // Instrumented dependencies
+    androidTestImplementation 'androidx.test.ext:junit:1.1.1'
```
rpdome: Can we add this to the version.gradle file? (Most likely we'll need the same thing for ADAL)

iambmelt: Yep - once this is fully working I'll tune it up for cleanliness

iambmelt: @rpdome I am going to merge this as-is for now and will take you suggestion as a follow-up PR into `android-complete`

PR #582, opened 2019-08-07
Rolling back mockito-core version for stability

iambmelt:


iambmelt:
This appears to be more than just a dependency version issue - espresso is also not working. Suspect these tests need some refactoring.

PR #593, opened 2019-08-17
Fix for foci lookup issue relative to migration

iambmelt:
Fixes an issue where foci lookups could fail when no rt is available with the current broker binding app's process uid


iambmelt:
superseded by #594

PR #598, opened 2019-08-21
Add PCA initialization error strings

rpdome:
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/709

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
+++ b/common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
@@ -353,31 +353,31 @@ public final class ErrorStrings {
     /**
      * AccountMode in configuration is set to multiple. However, the device is marked as shared (which requires single account mode).
      */
-    public static final String CANNOT_INITIALIZE_MULTIPLE_ACCOUNT_PCA_ON_SHARED_DEVICE_ERROR_CODE = "cannot_initialize_multiple_account_pca_on_shared_device";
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -350,4 +350,34 @@ private ErrorStrings() {
      */
     public static final String KEY_NOT_FOUND = "key_not_found";
 
+    /**
+     * AccountMode in configuration is set to multiple. However, the device is marked as shared (which requires single account mode).
+     */
+    public static final String CANNOT_INITIALIZE_MULTIPLE_ACCOUNT_PCA_ON_SHARED_DEVICE_ERRORCODE = "cannot_initialize_multiple_account_pca_on_shared_device";
```
shahzaibj: ERRORCODE can be separated with an underscore as ERROR_CODE

same for all below

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -350,4 +350,34 @@ private ErrorStrings() {
...
+    public static final String CANNOT_INITIALIZE_MULTIPLE_ACCOUNT_PCA_ON_SHARED_DEVICE_ERRORMESSAGE = "AccountMode in configuration is set to multiple. However, the device is marked as shared (which requires single account mode).";
```
shahzaibj: ERRORMESSAGE can be separated with an underscore as ERROR_MESSAGE

same for all below

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
kreedula: Error constant name seems too long, consider shortening the constant name like `MULTIPLE_ACCOUNT_INIT_FAIL_ERROR_CODE` for all of them below

PR #606, opened 2019-08-23
Fix the back-compat of hello()

heidijinxujia:
Remove the unused property in BrokerResult
Add the helper function getHelloResultFromBundle


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -176,6 +177,29 @@ public static BrokerResult brokerResultFromBundle(@NonNull final Bundle resultBu
...
+            final String negotiatedBrokerProtocolVersion = bundle.getString(AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY);
+            Logger.verbose(TAG + methodName,
+                    "Able to establish the connect, " +
+                            "the broker protocol version in common is ["
+                            + negotiatedBrokerProtocolVersion + "]");
+            return true;
+        } else if (bundle.get(AuthenticationConstants.Broker.BROKER_RESULT_V2) != null
+                && bundle.get(AuthenticationConstants.Broker.BROKER_RESULT_V2) instanceof BrokerResult) {
+            //no common broker version found and create the client exception from the result bundle.
+            //final BrokerResult brokerResult = MsalBrokerResultAdapter.brokerResultFromBundle(resultBundle);
```
rpdome: Shouldn't we use this?

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -176,6 +177,29 @@ public static BrokerResult brokerResultFromBundle(@NonNull final Bundle resultBu
...
+        } else if (!StringUtil.isEmpty(bundle.getString(AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY))) {
+            final String negotiatedBrokerProtocolVersion = bundle.getString(AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY);
+            Logger.verbose(TAG + methodName,
+                    "Able to establish the connect, " +
+                            "the broker protocol version in common is ["
+                            + negotiatedBrokerProtocolVersion + "]");
+            return true;
+        } else if (bundle.get(AuthenticationConstants.Broker.BROKER_RESULT_V2) != null
+                && bundle.get(AuthenticationConstants.Broker.BROKER_RESULT_V2) instanceof BrokerResult) {
+            //no common broker version found and create the client exception from the result bundle.
```
kreedula: Add comments to say we are using `serializable` for back compat

kreedula: As discussed, do not use `BrokerResult` to send result here, instead use `errorcode` and `errormessage` as is and for back compat, have this fall back logic

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -176,6 +177,33 @@ public static BrokerResult brokerResultFromBundle(@NonNull final Bundle resultBu
                 );
     }
 
+    public static boolean getHelloResultFromBundle(final Bundle bundle) throws ClientException {
+        final String methodName = ":getHelloResultFromBundle";
+        if (bundle == null) {
+            Logger.warn(TAG + methodName, "The hello result bundle is null.");
+            return false;
+        } else if (!StringUtil.isEmpty(bundle.getString(AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY))) {
```
kreedula: No need of `else if` here as we are returning right away. Just using `if` will improve readability

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerResult.java
```diff
@@ -85,6 +84,8 @@
         static final String CLI_TELEM_SUB_ERROR_CODE = "cli_telem_suberror_code";
     }
 
+    private static final long serialVersionUID = 8606631820514878489L;
```
rpdome: What is this for?

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
rpdome: Personally, I always do early return wherever possible. (Also, if I can't do early return, it could be a sign that the function is doing 'too much').

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerResult.java
heidijinxujia: Fixed serialVersionUID is needed for back-compat. Otherwise, the serialVersionUID will change every time when any code changing in the serializable class.

Issue #604, opened 2019-08-22
MicrosoftFamilyOAuth2TokenCache#loadByFamilyId should include AT's in its lookup


iambmelt:
See [`MicrosoftFamilyOAuth2TokenCache#loadByFamilyId`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java#L67-L141)

Currently AT's are not loaded for this lookup type - this increases COGs by forcing use of the RT, when there _is a valid AT in the cache_.

Including AT's in the query needs to include a `target` - this requires an API change to `MicrosoftFamilyOAuth2TokenCache`

iambmelt:
Wrinkles to consider:
- The lookup for the AT needs to be bound for the `clientId` but the lookup for the RT does not
- The lookup for the AT needs to check that the scopes are a complete/superset of the request scopes
- The IdToken just needs to match the `clientId`

shahzaibj:
@iambmelt I believe this can be closed as it seems like it has been resolved by #608 ?

iambmelt:
Yep -- thanks @shahzaibj

PR #608, opened 2019-08-23
Closes #607, #604

### LINKED ISSUE #607: FoCI cache tests need updating
Tests were broken by fixes submitted in #599 and #594 - this item is to update the tests

iambmelt:
Re-introduces AT lookups into family queries- API has been updated to include a `target` parameter.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java
@@ -39,6 +39,8 @@ import com.microsoft.identity.common.internal.providers.oauth2.AuthorizationRequ
 import com.microsoft.identity.common.internal.providers.oauth2.OAuth2Strategy;
 import com.microsoft.identity.common.internal.providers.oauth2.TokenResponse;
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftFamilyOAuth2TokenCache.java
```diff
@@ -125,9 +130,27 @@ public ICacheRecord loadByFamilyId(@Nullable final String clientId,
             }
         }
 
+        if (null != target) {
+            for (final Credential credential : getAccountCredentialCache().getCredentials()) {
+                if (credential instanceof AccessTokenRecord) {
+                    final AccessTokenRecord atRecord = (AccessTokenRecord) credential;
+
+                    if (null != clientId && clientId.equals(atRecord.getClientId())
```
shahzaibj: This entire boolean condition could be refactored into a separate boolean method

iambmelt: I'll elect not to take this change based on a few points:

- Increases method count
- You still need to read the method to determine how the check works
- Little potential for reuse

PR #624, opened 2019-09-03
Ends activity when it's killed (i.e. by user)

rpdome:
Reported by M&S.

**Issue**
When the activity (i.e. webView page) is killed by user (i.e. by swiping it away), we currently do not invoke any callback, and therefore sInteractiveExecutor is blocked - as it is single-threaded.


rpdome:
Already fixed in master.

PR #625, opened 2019-09-03
Merge release to dev.

rpdome:


common/versioning/version.properties
```diff
@@ -1,3 +1,3 @@
 #Thu Aug 02 12:03:16 PDT 2018
-versionName=0.0.17
+versionName=1.0.0-RC3
```
iambmelt: Does this need to match `0.0.18` from the release notes?

rpdome: I think we're fine. Once we're actually shipping the next change, we'll have to add 1.0.0 to the release note.

PR #634, opened 2019-09-05
PPE support for Broker

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -128,6 +133,12 @@ public BrokerAcquireTokenOperationParameters brokerInteractiveParametersFromActi
                 )
         );
 
+        if (!TextUtils.isEmpty(brokerRequest.getEnvironment())) {
```
shahzaibj: Can this piece of code be moved somewhere else? Since the "environment' is not a field in BrokerAcquireTokenOperationParameters, could we move this piece of code somewhere else?

shahzaibj: Same question for similar code in method brokerSilentParametersFromBundle

kreedula: We only unpack `BrokerRequest` here , so yeah  we would have to unpack it again if it's moved and I want to avoid that. I have added a comment to be explicit

shahzaibj: I see, thanks for adding the comment

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -152,6 +157,13 @@ public BrokerAcquireTokenOperationParameters brokerInteractiveParametersFromActi
 
         parameters.setAuthorizationAgent(AuthorizationAgent.WEBVIEW);
 
+        // Set Global environment variable for instance discovery if present
```
rpdome: Do we have to add this in getAcquireTokenParametersForBrokerRTRequest()? (or somehow refactor the code in a  way that we don't have to dupe this logic, if that's the case).

kreedula: We would have set the global parameter by then, i.e. when the client first calls acquire token we set the environment given by the client

PR #641, opened 2019-09-11
Adding error string for MSAL declined scope exception

kreedula:


common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -380,4 +380,11 @@ private ErrorStrings() {
     public static final String SINGLE_ACCOUNT_PCA_INIT_FAIL_UNKNOWN_REASON_ERROR_CODE = "single_account_pca_init_fail_unknown_reason";
     public static final String SINGLE_ACCOUNT_PCA_INIT_FAIL_UNKNOWN_REASON_ERROR_MESSAGE = "A single account public client application could not be created for unknown reasons.";
 
+    /**
+     * Some or all requested scopes where declined by the server. Developer should decide whether to continue
+     * authentication with the granted scopes or end the authentication process.
+     */
+    public static final String DECLINED_SCOPE_ERROR_CODE = "Declined scope error";
```
shahzaibj: Can we change this string to the same underscored format as the rest for consistency?

PR #643, opened 2019-09-13
Always compute the redirectUri when ADAL calls broker - ignore the parameterized value

iambmelt:
This is a change to match the former [Unity implementation](https://github.com/AzureAD/azure-activedirectory-library-for-android-unity-fork/blob/master/adal-unity-fork/src/main/java/com/microsoft/aad/adal/unity/BrokerClient.java#L670) where the broker redirect_uri was _always_ computed when broker makes requests on behalf of an ADAL client.

Things to note:
- The `redirect_uri` parameter `ACCOUNT_REDIRECT` is now _ignored_ except when performing validation
- This value is computed for both the silent and the interactive calls

Alternative approaches that could have been taken:
- Encode the URL _only_ when embedded in PRT requests
    * I chose not to do this because this would mean the same the URL with various encodings is passed around at different parts of the code. This felt confusing and not the right way to address the bug

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
@@ -28,10 +28,11 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import android.text.TextUtils;
 import android.util.Pair;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.internal.authorities.Authority;
@@ -105,7 +106,7 @@ public class AdalBrokerRequestAdapter implements IBrokerRequestAdapter {
         // V1 endpoint always add an organizational account if the tenant id is common.
         // We need to explicitly add tenant id as organizations if we want similar behavior from V2 endpoint
 
-        if(authority.getAudience().getTenantId().equalsIgnoreCase(AzureActiveDirectoryAudience.ALL)){
+        if (authority.getAudience().getTenantId().equalsIgnoreCase(AzureActiveDirectoryAudience.ALL)) {
             authority.getAudience().setTenantId(AzureActiveDirectoryAudience.ORGANIZATIONS);
         }
         parameters.setAuthority(authority);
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
```diff
@@ -28,17 +28,19 @@
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import android.text.TextUtils;
 import android.util.Pair;
 
+import androidx.annotation.NonNull;
```
shoatman: I think in the release branch they haven't added support for Androidx as a result of the Intune MAM team issue with AndroidX

common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
```diff
@@ -121,8 +125,18 @@ public BrokerAcquireTokenOperationParameters brokerInteractiveParametersFromActi
                 AuthenticationConstants.Broker.ACCOUNT_CLIENTID_KEY)
         );
 
+        // V1 Broker would compute the redirect_uri for the calling package, rather than
+        // 'trust' the provided value -- this had the unfortunate consequence of allowing
+        // callers to pass non-URL-encoded signature hashes into the library despite the documentation
+        // prescribing otherwise. The ADAL.NET implementation unfortunately RELIES on this behavior,
+        // forcing customers to use non-encoded values in order to pass validation check inside of
+        // ADAL.NET. In order to not regress this experience, the redirect URI must now be computed
+        // meaning that the ACCOUNT_REDIRECT parameter is basically ignored.
```
rpdome: might be worth considering getting rid of ACCOUNT_REDIRECT entirely (from MSAL/Broker/Common)

common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
rpdome: We have another branch for that. (release/1.0.0-NoAndroidX)

common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
iambmelt: Thanks @rpdome -- I think this is the right call. I'll exclude those changes from this PR for now and will include them in a follow up

PR #650, opened 2019-09-17
Use /consumers when imported RT is MSA (TSL)

iambmelt:
Use `/common` as our authority tenant when MSA RT are saved via TSL for refresh.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.adal.internal.tokensharing;
 
+import android.net.Uri;
 import android.util.Pair;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -232,7 +234,25 @@ public void saveMsaFamilyRefreshToken(@NonNull final String refreshToken) throws
                 sBackgroundExecutor.submit(new Callable<Pair<MicrosoftAccount, MicrosoftRefreshToken>>() {
                     @Override
                     public Pair<MicrosoftAccount, MicrosoftRefreshToken> call() throws ClientException {
-                        final ADALTokenCacheItem cacheItemToRenew = createTokenCacheItem(refreshToken);
+                        // Use the /consumers endpoint relative to the current cloud
+                        final Uri defaultAuthorityUri = Uri.parse(mDefaultAuthority);
```
iambmelt: Just calling out that this logic assumes the default authority takes the form of:
```
https://<host>.com/<tenant_or_tenant_alias>
```

PR #652, opened 2019-09-19
Fix null pointer issue when constructing Auth Request

kreedula:
Icm : https://icm.ad.msft.net/imp/v3/incidents/details/147446572/home

Power apps has a flow where user doesn't have to enter `login_hint` which seems to be issue here as the Auth Request builder could call `toString` on a null value.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -292,7 +292,12 @@ public Uri getAuthorizationRequestAsHttpRequest() {
         final Uri.Builder uriBuilder = Uri.parse(getAuthorizationEndpoint()).buildUpon();
 
         for (Map.Entry<String, Object> entry : qpMap.entrySet()) {
-            uriBuilder.appendQueryParameter(entry.getKey(), entry.getValue().toString());
+            if (entry.getKey() != null && entry.getValue() != null) {
```
rpdome: we probably don't need null check for key here.

kreedula: yes agree, just added as it's no harm

PR #670, opened 2019-10-03
Fixes Telemetry thread issues and NullPointerExceptions

shahzaibj:
This PR fixes the following MSAL Issues:

- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/785
- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/786

I've changed the data types for `mTelemetryRawDataMap` and the `mObservers` variables and now we are using a [ConcurrentLinkedQueue ](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html)which is thread-safe. The emit and flush methods are no longer synchronized as it is not needed with a ConcurrentLinkedQueue.

Good Stack Overflow reads related to our issue:

- https://stackoverflow.com/questions/20503627/concurrentmodificationexception-on-synchronized-method
- https://stackoverflow.com/questions/25113987/why-is-there-a-concurrentmodificationexception-even-when-list-is-synchronized
- https://stackoverflow.com/questions/8203864/choosing-the-best-concurrency-list-in-java


common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -131,13 +131,18 @@ public static HttpResponse sendPost(final URL requestUrl, final Map<String, Stri
...
+                        .putRequestIdHeader(requestHeaders == null ? null : requestHeaders.get(CLIENT_REQUEST_ID))
         );
 
         final HttpRequest httpRequest = new HttpRequest(requestUrl, requestHeaders, REQUEST_METHOD_POST,
                 requestContent, requestContentType);
         final HttpResponse response = httpRequest.send();
-        Telemetry.emit(new HttpEndEvent().putStatusCode(response.getStatusCode()));
+
+        if (response != null) {
+            Telemetry.emit(new HttpEndEvent().putStatusCode(response.getStatusCode()));
```
kreedula: nit: Refactor to below for better readability. if/else seems redundant here. Feedback applies below also
```
final HttpEndEvent httpEvent = new HttpEndEvent();
if(response != null){
httpEvent.putStatusCode(response.getStatusCode());
}
Telemetry.emit(httpEvent);
```

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
```diff
@@ -176,7 +173,8 @@ public synchronized void removeObserver(final ITelemetryObserver observer) {
      * @return List of ITelemetryObserver object.
      */
     public List<ITelemetryObserver> getObservers() {
-        return Collections.unmodifiableList(mObservers);
+        List observersList = new CopyOnWriteArrayList<>(mObservers);
+        return Collections.unmodifiableList(observersList);
```
kreedula: why a `CopyOnWriteArrayList` here , we are passing an unmodifiable list anyways right i.e. we should be thread safe already isn't it?

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -156,16 +161,21 @@ public static HttpResponse sendGet(final URL requestUrl, final Map<String, Strin
...
+        final HttpResponse response = httpRequest.send();
+
+        if (response != null) {
+            Telemetry.emit(
+                    new HttpEndEvent()
+                            .putStatusCode(response.getStatusCode())
+            );
+        } else {
+            Telemetry.emit(new HttpEndEvent());
+        }
```
kreedula: nit : same feedback as above

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
shahzaibj: This method returns a `List` and `Collections.unmodifiableList` accepts a list as the parameter, however `mObservers` is a queue instead. So first converting the `mObservers` queue to a list by using `CopyOnWriteArrayList` and then I create an Unmodifiable List by that.

Actually, I'm going to refactor this method to instead return a `queue` which is the actual data type for `mObservers` and I think that makes more sense, and I won't return an unmodifable version anymore as the queue will be thread-safe anyway.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/telemetry/Telemetry.java
shahzaibj: @kreedula I've pushed the change to return a Queue instead as discussed above. Let me know how it looks.

shahzaibj: @kreedula As discussed `mObservers` has been changed back to a `List`

PR #671, opened 2019-10-04
Implement Server Side Telemetry V1

shahzaibj:
In this PR, I've implemented server side telemetry for MSAL. We will be sending some telemetry to the eSTS as part of each request by attaching it to the headers. There will be two headers, one for telemetry associated to the current request and one for last request. The telemetry contained within these headers MUST follow the schema defined for the server side telemetry. 

Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/791
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1155


common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/IRequestTelemetryCache.java
```diff
@@ -0,0 +1,26 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License info needed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/PublicApiId.java
```diff
@@ -0,0 +1,19 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License info needed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
```diff
@@ -0,0 +1,134 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License info needed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
```diff
@@ -0,0 +1,147 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: Same here

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetry.java
```diff
@@ -0,0 +1,242 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: Same here

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/SharedPreferencesLastRequestTelemetryCache.java
```diff
@@ -0,0 +1,55 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/SharedPreferencesRequestTelemetryCache.java
```diff
@@ -0,0 +1,58 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/TelemetryUtils.java
```diff
@@ -0,0 +1,39 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License

common/src/test/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetryTest.java
```diff
@@ -0,0 +1,115 @@
+package com.microsoft.identity.common.internal.servertelemetry;
```
iambmelt: License

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/IRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/PublicApiId.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetry.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/SharedPreferencesLastRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/SharedPreferencesRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/TelemetryUtils.java
shahzaibj: Addressed

common/src/test/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetryTest.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -134,6 +135,9 @@ public static HttpResponse sendPost(final URL requestUrl, final Map<String, Stri
                         .putRequestIdHeader(requestHeaders == null ? null : requestHeaders.get(CLIENT_REQUEST_ID))
         );
 
+        final Map<String, String> telemetryHeaders = ServerTelemetry.getTelemetryHeaders();
```
iambmelt: The network layer should not know about the telemetry layer, instead these headers should be passed in by the caller above

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -164,6 +168,9 @@ public static HttpResponse sendGet(final URL requestUrl, final Map<String, Strin
...
```
iambmelt: Same here - this is a violation of the separation of concerns: the network layer should not know about the telemetry layer, instead these headers should be passed in by the caller above

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
```diff
@@ -0,0 +1,156 @@
...
+
+    Map<String, String> getCommonTelemetry() {
+        return mCommonTelemetry;
+    }
+
+    Map<String, String> getPlatformTelemetry() {
+        return mPlatformTelemetry;
+    }
+
+    String getCompleteTelemetryHeaderString() {
```
iambmelt: Declare return type `@Nullable`

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
```diff
@@ -0,0 +1,169 @@
...
+        public static final String ACCOUNT_STATUS = TelemetryEventStrings.Key.ACCOUNT_STATUS;
+        public static final String ID_TOKEN_STATUS = TelemetryEventStrings.Key.ID_TOKEN_STATUS;
+        public static final String AT_STATUS = TelemetryEventStrings.Key.AT_STATUS;
+        public static final String RT_STATUS = TelemetryEventStrings.Key.RT_STATUS;
+        public static final String FRT_STATUS = TelemetryEventStrings.Key.FRT_STATUS;
+        public static final String MRRT_STATUS = TelemetryEventStrings.Key.MRRT_STATUS;
+    }
+
+    public static final class Value {
+        public static final String SCHEMA_VERSION = "1";
```
iambmelt: Two variables named `SCHEMA_VERSION` in this file is confusing -- I suggest making this harder to harder to accidentally use the wrong value by renaming these to something like:
`SCHEMA_VERSION` and `VERSION_1`

Consider what happens when the version changes to `2` in the future, now you'll be changing the value of this constant, rather than adding a new one -- if you decide you want to clean up the v1 cache when you rev to v2 you'll end up hardcoding `1` again under a different variable name so might as well plan to have a few schema versions

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
```diff
@@ -0,0 +1,156 @@
...
+     * It is important to ensure that the fields array passed to this method is in the correct order,
+     * as determined in {@link Schema}.
+     * Failure to do so will return a malformed header string.
+     *
+     * @param fields    The fields that need to be included in the header string
+     * @param telemetry A HashMap of telemetry data that maps keys (fields) to their values
+     * @return a telemetry header string composed from provided telemetry fields and values
+     */
+    private String getTelemetryHeaderStringFromFields(String[] fields, Map<String, String> telemetry) {
+        if (fields == null || telemetry == null) {
```
iambmelt: If you _really_ want to support `null` and `null` as params, then the parameters should be annotated as `@NonNull` (and `final`)

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetry.java
```diff
@@ -0,0 +1,264 @@
...
+import com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager;
+import com.microsoft.identity.common.internal.logging.DiagnosticContext;
+import com.microsoft.identity.common.internal.logging.Logger;
+import com.microsoft.identity.common.internal.result.AcquireTokenResult;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ServerTelemetry {
```
iambmelt: I would rename this class to something _more specific_ such as `EstsTelemetry`

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private static String[] getLastRequestCommonFields() {
+        return lastRequestCommonFields;
+    }
+
+    private static String[] getLastRequestPlatformFields() {
+        return lastRequestPlatformFields;
+    }
+
+    static String[] getCommonFields(final boolean isCurrent) {
```
iambmelt: I suggest javadoc'ing `getCommonFields` and `getPlatformFields` to call out what the difference is.

(For those who don't have the spec committed to memory 😉 )

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * The name of the SharedPreferences file on disk for the last request.
+     */
+    private static final String LAST_REQUEST_TELEMETRY_SHARED_PREFERENCES =
+            "com.microsoft.identity.client.last_request_telemetry";
+
+    private static IRequestTelemetryCache sLastRequestTelemetryCache;
+    private static Map<String, RequestTelemetry> sTelemetryMap;
+
+    public static void initializeServerTelemetry(@NonNull final Context context) {
```
iambmelt: So, this class isn't quite a singleton but it also isn't quite fully static -- would this ever need to be initialized >1x?

If not, I suggest making the `static` methods _instance_ methods and removing `initializeServerTelemetry` (which should be renamed) and replacing it with a `synchronized getInstance()` method that initializes the object if not instance already exists or otherwise returns the existing instance

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
```diff
@@ -0,0 +1,48 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.eststelemetry;
+
+public interface IRequestTelemetryCache {
```
iambmelt: I would javadoc that this is really a "cache of _one_"

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/SharedPreferencesLastRequestTelemetryCache.java
```diff
@@ -0,0 +1,77 @@
...
+package com.microsoft.identity.common.internal.eststelemetry;
+
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.common.internal.cache.ISharedPreferencesFileManager;
+import com.microsoft.identity.common.internal.logging.Logger;
+
+import java.util.Map;
+
+public class SharedPreferencesLastRequestTelemetryCache extends SharedPreferencesRequestTelemetryCache {
```
iambmelt: Unless `save()` and `get()` are in the same class, what is the purpose of a subclass? Isn't it important to know how `save()` works to write the `get()` function?

How would a second subclass implement new functionality without modifying the parent class? Wouldn't they have to `@Override` the `save()` method? If yes, why not declare both `save()` and `get()` as `abstract` in the `super` class?

shahzaibj: @iambmelt You're right, this is messed up.

I think I'll just remove the `SharedPreferencesRequestTelemetryCache` class, and instead just implement the methods directly in the `SharedPreferencesLastRequestTelemetryCache` class. I don't think we are ever going to be tracking any request beyond the last request so I think just having one class will be simpler.

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
shahzaibj: Addressed as discussed offline.

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/RequestTelemetry.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/ServerTelemetry.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/servertelemetry/Schema.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/SharedPreferencesLastRequestTelemetryCache.java
shahzaibj: Addressed as stated above.

shahzaibj:
> Thanks @shahzaibj Posted some comments -- let me know when you're ready for more review and we can go for another round 👍

@iambmelt PR is available for another review!

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,348 @@
...
+    private final static String TAG = EstsTelemetry.class.getSimpleName();
+
+    /**
+     * The name of the SharedPreferences file on disk for the last request.
+     */
+    private static final String LAST_REQUEST_TELEMETRY_SHARED_PREFERENCES =
+            "com.microsoft.identity.client.last_request_telemetry";
+
+    private static volatile EstsTelemetry sEstsTelemetryInstance = null;
+    private static IRequestTelemetryCache sLastRequestTelemetryCache;
```
iambmelt: Let's make `sEstsTelemetryInstance` and `sLastRequestTelemetryCache` _member_ variables of `EstsTelemetry` and keep a reference to a single instance -- rather than provide the _veneer_ of OO over a `static` implementation, my preference would be to have a _real_ OO implementation with an API that programmatically enforces that the entire app has a single instance

iambmelt: Similar to many `Singleton` examples: https://www.baeldung.com/java-singleton

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,348 @@
...
+        flush(correlationId, baseException == null ? null : baseException.getErrorCode());
+    }
+
+    public void flush(final String correlationId, final AcquireTokenResult acquireTokenResult) {
+        final String errorCode = TelemetryUtils.errorFromAcquireTokenResult(acquireTokenResult);
+        flush(correlationId, errorCode);
+    }
+
+    /**
+     * Flush the telemetry data associated to the correlation id.
```
iambmelt: "Flush to disk/sharedprefs/whatever"

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,376 @@
...
+
+        return sEstsTelemetryInstance;
+    }
+
+    /**
+     * Initializes the Ests Telemetry by creating a new instance of {@link EstsTelemetry}
+     *
+     * @param context the application context
+     */
+    public static void initializeEstsTelemetry(@NonNull final Context context) {
```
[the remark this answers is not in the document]
shahzaibj: Yeah that's a problem. Let me fix it and push another commit. Thanks @kreedula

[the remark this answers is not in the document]

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -74,20 +69,26 @@ public static synchronized EstsTelemetry getInstance() {
...
-     * Initializes the Ests Telemetry by creating a new instance of {@link EstsTelemetry}
+     * Initializes the Ests Telemetry Cache by creating a new instance of {@link EstsTelemetry} if
+     * an instance already didn't exist, and then creates and sets the last request telemetry cache
+     * on that instance.
      *
      * @param context the application context
      */
-    public static void initializeEstsTelemetry(@NonNull final Context context) {
-        final String methodName = ":initializeEstsTelemetry";
+    public static synchronized void initializeEstsTelemetryCache(@NonNull final Context context) {
```
kreedula: why `static`, can be an instance method right?

shahzaibj: This needs to be called from anywhere. I should probably rename to `initializeEstsTelemetry` as it also creates a new instance of `EstsTelemetry` if one wasn't available. But the idea is that this is the first method that calls get called before any other method from this class is called, and it will create an instance of `EstsTelemetry` and also setup the cache on that instance.

So basically this will be called in the `initializeApplication` method in `PublicClientApplication`.

We can certainly delete this method and have an instance method that takes a `context` and creates and sets the cache on that instance and actually that might be better.

shahzaibj: @kreedula Addressed. This method has been removed and the instance method to set the cache on the instance has been provided.

PR #674, opened 2019-10-07
Command refactoring

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -53,7 +53,7 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Acqu
                                 : null
                 ).prompt(parameters.getOpenIdConnectPromptParameter().name())
                 .claims(parameters.getClaimsRequestJson())
-                .forceRefresh(!TextUtils.isEmpty(parameters.getClaimsRequestJson()))
+                .forceRefresh(parameters.getForceRefresh())
```
kreedula: needs a merge from `dev` I guess , I believe these `forceRefresh` changes already went to dev

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -147,13 +101,48 @@ public void run() {
...
+                                    @Override
+                                    public void run() {
+                                        command.getCallback().onError(finalException);
+                                    }
+                                });
+                            }
                         }
-                    });
+                    }else{
+                        final Object returnResult = result;
```
kreedula: should we null check for `returnResult` here ?, Also may be worth adding a comment that we get into the else loop here for `LoadAccountCommand` and `RemoveAccountCommand`

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -147,13 +101,48 @@ public void run() {
                         }
                     });
                 } else {
-                    final boolean finalResult = result;
-                    handler.post(new Runnable() {
-                        @Override
-                        public void run() {
-                            command.getCallback().onTaskCompleted(finalResult);
+                    if(result != null && result instanceof AcquireTokenResult){
```
kreedula: nit: better to add a helper method to post the result for readability

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
shoatman: resolve.d

PR #675, opened 2019-10-09
First commit adding command caching

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
```diff
@@ -0,0 +1,45 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.controllers;
+
+public class CommandResult {
```
iambmelt: For either current or future use, it would be useful to be able to specify a type on this class so that the caller doesn't necessarily have to cast the result

Something like
```java
public class CommandResult<T> {

    private ResultStatus mStatus;
    private T mResult;

    public Command Result(ResultStatus status, T result) {
        // ..
    }

    T getResult() {
        return mResult;
    }
}
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResultCache.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public CommandResultCache(){
+        mCache = new LruCache<>(DEFAULT_ITEM_COUNT);
+    }
+
+    public CommandResultCache(int maxItemCount){
+        mCache = new LruCache<>(maxItemCount);
+    }
+
+    public CommandResult get(BaseCommand key){
```
iambmelt: I suspect for future debugging, it would be useful to log in the method whether or not this command's result was serviced from the cache or not

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -68,101 +68,156 @@ public static void submitSilent(@NonNull final BaseCommand command){
...
+                commandResult = new CommandResult(CommandResult.ResultStatus.COMPLETED, result);
+            }
+        }
+
+        return commandResult;
+
+    }
+
+    /**
+     * Return the result of the command to the caller via teh callback associated with the command
```
iambmelt: nit: `via the callback`

iambmelt:
Discussed in standup and just reposting here....

We may want to think about what actions (if any) would _force_ the cache to invalidate - possible scenarios that could cause a stale/incorrect result to return from the cache....
- A user initiates an interactive auth and adds a new `TenantProfile` (such as from a guest tenant)
- `removeAccount()` is called followed by an ATS call

Come to think of it, the `removeAccount()` scenario might be the most likely to cause confusion

iambmelt:
More food for thought:
- Perhaps caching can be toggled and tweaked from `PublicClientApplicationConfiguration`?
    * Duration of cache validity, maybe _max_ 1 minute?
    * Enable/disable (defaults to enabled)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResultCache.java
shoatman: Added.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
shoatman: Let's review this together... not sure this helps me in terms of creating the objects.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Fixed.  thanks!

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -65,104 +70,182 @@ public static void submitSilent(@NonNull final BaseCommand command){
...
-                    }
-                }
+                sExecutingCommands.remove(command);
 
                 Telemetry.getInstance().flush(correlationId);
             }
         });
     }
 
+    public static void clearCommandCache(){
```
shahzaibj: Why is this method public? Anyone can call this method from anywhere, and if they are able do execute this as part of their loop, then I believe we won't throttle and we won't provide any protection against excessive requests.

I think this method can package-private or even private as it is not being used anywhere at the moment, other than the [command caching tests in MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/801/files#diff-721e4828ff2ace5339c66f849f90da31R289-R344).

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -65,104 +70,182 @@ public static void submitSilent(@NonNull final BaseCommand command){
...
+        sCommandResultCache.clear();
+    }
+
     /**
-     * We need to inspect the AcquireTokenResult type to determine whether the request was successful, cancelled or encountered an exception
+     * Execute the command provide to the command dispatcher
```
shahzaibj: nit: provided

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -54,6 +57,8 @@
     private static final ExecutorService sSilentExecutor = Executors.newCachedThreadPool();
     private static final Object sLock = new Object();
     private static InteractiveTokenCommand sCommand = null;
+    private static final CommandResultCache sCommandResultCache = new CommandResultCache();
+    private static final Set sExecutingCommands = Collections.synchronizedSet(new HashSet<BaseCommand>());
```
shahzaibj: Is this going to be thread-safe? From my experience, creating a synchronized collection is not enough since different threads can still modify the collection at the same time.

See these stack overflow posts:
- https://stackoverflow.com/questions/20503627/concurrentmodificationexception-on-synchronized-method
- https://stackoverflow.com/questions/25113987/why-is-there-a-concurrentmodificationexception-even-when-list-is-synchronized

shahzaibj: We may need a concurrent data structure.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Let's discuss.  I want to be able to re-use your investments in the existing tests.  If package private or private this call will be inaccessible to the tests.  What's your suggestion on how to achieve the same thing?  I don't thing our goal in this case is prevent malicious apps from misusing the library... if anyone wants to attack the service they can by simply opening an http connection and making many requests.  Our  goal here is to protect the system against coding bugs in clients.  Unintentional misuse.  Note that many other things are in "internal" packages that we don't intend for clients to call directly.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Fixed.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: I think this code is going to be ok.  Synchronized set will ensure that the threads don't invoke the same methods at the same time.  In this particular scenario I couldn't think of any way in which ordering of execution mattered.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Ok.  I took this feedback and worked around accessibility by creating a helper class in the MSAL unit tests.

testutils/src/main/java/com/microsoft/identity/internal/testutils/strategies/MockDelayedResponseStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+import com.microsoft.identity.common.internal.providers.oauth2.TokenResponse;
+import com.microsoft.identity.common.internal.providers.oauth2.TokenResult;
+import com.microsoft.identity.internal.testutils.MockTokenResponse;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+public class MockDelayedResponseStrategy extends ResourceOwnerPasswordCredentialsTestStrategy {
```
shahzaibj: Can `extend` from `MockTestStrategy` instead, which would lead you to overriding less methods (only one actually)

testutils/src/main/java/com/microsoft/identity/internal/testutils/strategies/MockTestStrategy.java
```diff
@@ -62,6 +62,13 @@ public TokenResult getTokenResult() {
     protected HttpResponse performTokenRequest(final MicrosoftStsTokenRequest tokenRequest) {
         final TokenResult tokenResult = getTokenResult();
         final TokenResponse tokenResponse = tokenResult.getTokenResponse();
+        /*
```
shahzaibj: Seems like this is not needed and can be removed

testutils/src/main/java/com/microsoft/identity/internal/testutils/authorities/MockDelayedResponseAuthority.java
```diff
@@ -0,0 +1,44 @@
...
+package com.microsoft.identity.internal.testutils.authorities;
+
+import com.microsoft.identity.common.internal.authorities.AnyOrganizationalAccount;
+import com.microsoft.identity.common.internal.authorities.AzureActiveDirectoryAuthority;
+import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Configuration;
+import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Strategy;
+import com.microsoft.identity.common.internal.providers.oauth2.OAuth2Strategy;
+import com.microsoft.identity.internal.testutils.strategies.MockTestStrategy;
+
+public class MockDelayedResponseAuthority extends AzureActiveDirectoryAuthority {
```
shahzaibj: Can `extend` from `MockAuthority` instead.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -65,104 +70,182 @@ public static void submitSilent(@NonNull final BaseCommand command){
...
+    }
+
+    /**
+     * Cache the result of the command (if eligible to do so) in order to protect the service from clients
+     * making the requests in a tight loop
      * @param command
+     * @param commandResult
+     */
+    private static void cacheCommandResult(BaseCommand command, CommandResult commandResult){
+        if(eligibleToCache(commandResult)){
```
kreedula: nit : logging would be helpful here , whether we cached or not

PR #685, opened 2019-10-10
Fix No Scope Test for Password Grant + Simplify Test Code

shahzaibj:


common/src/test/java/com/microsoft/identity/common/ropc/PasswordGrantTest.java
```diff
@@ -142,21 +141,21 @@ public void testRopcFailureManagedUserNoPassword() throws IOException {
...
-    @Ignore // this has started to fail for some reason, will need investigation
+    // test that ROPC flow fails if scope is not provided
     public void testRopcFailureManagedUserNoScope() throws IOException {
         final Credential credential = getUserCredentialsForManagedUser();
 
         try {
             final TokenResult tokenResult = performRopcTokenRequest(null, credential.userName, credential.password);
 
-            assertEquals(true, tokenResult.getSuccess());
+            assertFalse(tokenResult.getSuccess());
```
iambmelt: Does the code even get this far? Isn't the `IllegalArgumentException` thrown before this?

shahzaibj: Yes, the exception is thrown before that point and code should not reach that far. I wasn't sure if I should put anything there at all. 

I think the right thing to do might be to replace those `assertFalse` statements with an explicit `fail()` so that if we ever get to that line due to coding errors then the test will fail we can catch those bug. What do you say @iambmelt ?

iambmelt: @shahzaibj +1

iambmelt: I would apply the same approach to the other blocks

shahzaibj: Addressed

PR #686, opened 2019-10-11
Refactor code to support FLW command migration

rpdome:
- Introduce new FLW commands.
- accountsFromBundle() now throws exception derived from the given error bundle.
- Add fields to OperationParameter to support FLW operations.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenOperation.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenOperation.java
@@ -31,7 +31,7 @@ import java.io.IOException;
 import java.util.concurrent.ExecutionException;
 
 public interface TokenOperation {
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenOperation.java
```diff
@@ -31,7 +31,7 @@
 import java.util.concurrent.ExecutionException;
 
 public interface TokenOperation {
-    AcquireTokenResult execute() throws InterruptedException, ExecutionException, IOException, BaseException;
+    AcquireTokenResult execute() throws InterruptedException, ExecutionException, IOException, BaseException, Exception;
```
iambmelt: I'm not sure how useful it is to throw `Exception` _and_ various subclasses thereof --  are these used in telemetry/logging? Could be we get away with just `Exception`?

rpdome:
Thanks @iambmelt!

PR #687, opened 2019-10-11
Add public api ids for overloads to capture actual caller in Server-Side Telemetry

shahzaibj:
Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/803
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1165

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
@@ -24,18 +24,83 @@ package com.microsoft.identity.common.internal.eststelemetry;
 
 public final class PublicApiId {
 
-    // Silent Apis
-    public static final String BROKER_ACQUIRE_TOKEN_SILENT = "21";
-    public static final String LOCAL_ACQUIRE_TOKEN_SILENT = "22";
+    //region Silent APIs
 
-    // Interactive APIs
-    public static final String BROKER_ACQUIRE_TOKEN_INTERACTIVE = "121";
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
```diff
@@ -24,18 +24,83 @@
...
+    public static final String MULTIPLE_ACCOUNT_PCA_GET_ACCOUNT_WITH_IDENTIFIER = "925";
+    public static final String MULTIPLE_ACCOUNT_PCA_GET_ACCOUNT_WITH_IDENTIFIER_CALLBACK = "926";
+
+    //endregion
+
+    // region REMOVE account
+
+    // SingleAccountPublicClientApplication
+    //==============================================================================================
+    public static final String SINGLE_ACCOUNT_PCA_SIGN_OUT = "927"; //REMOVE
```
iambmelt: > `//REMOVE`

Is this an unresolved TODO?

shahzaibj: No that is just a comment indicating that this actually does a Remove account, since the term sign out was not explicit about what it does. But since I created regions for each get/remove accounts, I think I can remove that comment.

iambmelt: I would either remove the comment or provide more context around it -- in the current state it looks like something you want to remove from the code

shahzaibj: Addressed

PR #688, opened 2019-10-14
Make cache key unique between each command type

rpdome:
Issue:

1. run GetDeviceModeCommand.
2. run GetCurrentAccountCommand.

app crashes since the 2) hits the cache and return the value of 1). Since 1) returns boolean, it crashes when it was trying to cast to a list.


common/src/main/java/com/microsoft/identity/common/internal/controllers/GetCurrentAccountCommand.java
```diff
@@ -68,4 +68,9 @@ public GetCurrentAccountCommand(@NonNull final OperationParameters parameters,
 
         return result;
     }
+
+    @Override
+    public int getCommandCode() {
```
shahzaibj: nit: Rename to `getCommandNameHashCode`

PR #690, opened 2019-10-15
Add eligible to cache to BaseCommand

shoatman:


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseCommand.java
```diff
@@ -93,6 +93,10 @@ public BaseController getDefaultController() {
 
     public abstract int getCommandNameHashCode();
 
+    public boolean isEligibleForCaching(){
+        return false;
+    }
+
```
rpdome: Not really related to this PR, but should we use hashCode in the equal function below? (or do a command type check before comparing parameters).

PR #692, opened 2019-10-16
Fix WebView SDK28 issue

rpdome:
See: https://developer.android.com/about/versions/pie/android-9.0-changes-28#web-data-dirs

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
@@ -2,13 +2,11 @@ package com.microsoft.identity.common.internal.providers.oauth2;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
-import android.app.ActivityManager;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.net.Uri;
-import android.os.Build;
 import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -178,11 +180,32 @@ private void extractState(final Bundle state) {
...
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         final String methodName = "#onCreate";
         super.onCreate(savedInstanceState);
 
+        // Must be invoked before WebvView in common_activity_authentication is inflated.
+        // See https://developer.android.com/about/versions/pie/android-9.0-changes-28#web-data-dirs for more info.
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            try {
+                WebView.setDataDirectorySuffix(getProcessName(getApplicationContext()));
```
kreedula: should we check if process name is empty and only set if it's `auth` , i.e. for MSAL local auth, what would be the process name?

rpdome: fixed.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/WebViewUtil.java
```diff
@@ -23,26 +23,75 @@
...
+     * */
+    @SuppressLint("NewApi")
+    public static void setDataDirectorySuffix(@NonNull final Context context){
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            try {
+                if (ProcessUtil.isRunningOnAuthService(context)) {
+                    WebView.setDataDirectorySuffix(AuthServiceProcess);
+                }
+            } catch (final IllegalStateException e) {
+                // If webView is already initialized, this will be thrown.
```
kreedula: may be add a `warn` log here

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/WebViewUtil.java
```diff
@@ -23,26 +23,75 @@
...
             final CookieSyncManager syncManager = CookieSyncManager.createInstance(context);
-            syncManager.startSync();
             cookieManager.removeAllCookie();
-            syncManager.stopSync();
             syncManager.sync();
         }
     }
+
+    /**
+     * Clear session cookies from embedded webview.
```
iambmelt: We should probably javadoc that this is blocking call and so should not be called on UI thread

PR #695, opened 2019-10-18
Revert the ArgumentException removal change

kreedula:
This [change](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/679) broke ADAL with broker scenerio.


shahzaibj:
Thanks for fixing this @kreedula ! Can you also provide a bit of context around what scenario broke exactly or what the bug/error code was so that we can be more careful about these changes in the future?

PR #704, opened 2019-10-23
Refactor code for enabling MSAL-Broker communication via AccountManager without GET_ACCOUNT permission

rpdome:
Also moving MSAL code that can be shared to common.


common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -257,6 +265,44 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
...
+            requestBundle.putString(ENVIRONMENT, parameters.getAccount().getEnvironment());
+            requestBundle.putString(ACCOUNT_HOME_ACCOUNT_ID, parameters.getAccount().getHomeAccountId());
+        }
+
+        return requestBundle;
+    }
+
+    public static Bundle getRequestBundleForGetAccounts(@NonNull final OperationParameters parameters) {
+        final Bundle requestBundle = new Bundle();
+        requestBundle.putString(ACCOUNT_CLIENTID_KEY, parameters.getClientId());
```
shahzaibj: Can we rename to `ACCOUNT_CLIENTID_KEY` to `ACCOUNT_CLIENT_ID_KEY`?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -257,6 +265,44 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
...
+        requestBundle.putString(ACCOUNT_REDIRECT, parameters.getRedirectUri());
+        //Disable the environment and tenantID. Just return all accounts belong to this clientID.
```
shahzaibj: What does disable mean in this context?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1187,6 +1193,29 @@ private AuthenticationConstants() {
         public static final String AZA_SCOPE = "aza";
     }
 
+    /**
+     * Represents Broker operations that should be invoked by Authenticator.java (MSAL-Broker AccountManager flow).
```
shahzaibj: Why is this class called `MsalAccountManagerOperation` when the javadoc says that it represents Broker Operations?
May be rename to `BrokerAccountManagerOperation`?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
rpdome: I'll keep it as is right now to keep the amount of change as minimal as possible - MSAL and Broker change is already big enough IMHO.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
rpdome: as synced with @iambmelt  offline, we're not passing them to Broker right now.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -257,8 +266,48 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
         return new HashSet<>(Arrays.asList(scopes));
     }
 
+    public static Bundle getSilentBrokerRequestBundle(final AcquireTokenSilentOperationParameters parameters) {
```
kreedula: nit: lets try to follow the adapter conventions here, any reason why it's static?
also rename to `requestBundleFromSilentParameters` to be consistent and remove `static`

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -257,8 +266,48 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
...
+        final Bundle requestBundle = new Bundle();
+        final BrokerRequest brokerRequest = msalBrokerRequestAdapter.
+                brokerRequestFromSilentOperationParameters(parameters);
+
+        requestBundle.putString(
+                AuthenticationConstants.Broker.BROKER_REQUEST_V2,
+                new Gson().toJson(brokerRequest, BrokerRequest.class)
+        );
+
+        requestBundle.putInt(
```
kreedula: not part of this change, but we should revisit why `CALLER_INFO_UID` is not part of `BrokerRequest`, could be a miss on my part when i was implementing this

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -257,8 +266,48 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
...
+                AuthenticationConstants.Broker.CALLER_INFO_UID,
+                parameters.getAppContext().getApplicationInfo().uid
+        );
+
+        return requestBundle;
+    }
+
+    public static Bundle getRequestBundleForRemoveAccount(@NonNull final OperationParameters parameters) {
```
kreedula: same here , rename to something like `removeAccountRequestBundleFromParameters`

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -275,6 +324,20 @@ public static Bundle getBrokerHelloBundle(@NonNull final OperationParameters par
         return requestBundle;
     }
 
+    public static Bundle getRequestBundleForRemoveAccountFromSharedDevice(@NonNull final OperationParameters parameters) {
```
kreedula: same feedback with `name` and `static` here

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -422,4 +430,18 @@ public static void verifyRemoveAccountResultFromBundle(@NonNull final Bundle bun
         }.getType();
         return builder.create().fromJson(accountJson, listOfCacheRecords);
     }
+
+    public static AcquireTokenResult getAcquireTokenResult(@NonNull final Bundle resultBundle) throws BaseException {
```
kreedula: same feedback with `name` and `static` here

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
rpdome: After taking another look, it makes more sense to put these functions in BrokerBaseStrategy instead.

I'll move them there, and I'll remove static and rename the method for consistency also.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
rpdome: Do you think it makes sense to set CALLER_INFO_UID to every request bundle? 

(Note: this value is already obtainable through Binder.getCallerUid and AccountManager's bundle.)

kreedula: yeah I think this is not needed as we are explicity adding anyway her e: https://github.com/AzureAD/ad-accounts-for-android/blob/dev/AADAuthenticator/src/msal/java/com/microsoft/identity/client/MicrosoftAuthService.java#L293

rpdome: Keeping it for now because we have a validation logic that check this in Broker.

PR #708, opened 2019-10-29
Add Swagger Client for new LAB API

shahzaibj:
Most of the code here is auto generated via swagger


iambmelt:
> Do we really want to version this?
> 
> I would propose that the following is a preferable approach:
> 
> * Publish Swagger metadata to exposed endpoint (lab team)
> * Add a Gradle task to query this metadata and run codegen
> * Add a `.gitignore` exemption for the generated files

Synced offline -- we'll version this for now until Swagger files stabilize

PR #710, opened 2019-11-01
Fix issue #709

### LINKED ISSUE #709: Environment not set correctly on Credentials (to be cached)
When creating cache records (AccessTokenRecord, RefreshTokenRecord, IdTokenRecord) to save to the cache, we are not setting the environment correctly on them. As a result, we get a cache miss when making a silent request.

shahzaibj:
Fixes #709


shahzaibj:
Closing this as moving forward with a different fix in #714

PR #714, opened 2019-11-01
Fix issue #709, #718

### LINKED ISSUE #709: Environment not set correctly on Credentials (to be cached)

### LINKED ISSUE #718: Need to match authority with account for silent request
Need to match authority with account for silent request before making refresh token request. If the account is a sovereign cloud, however, the authority passed is worldwide, we should not make the request to the server and just throw an exception if account and authority don't match.

We are currently making the request to the server and receive an error response from the server that says Invalid National Cloud is specified.

shahzaibj:
Fixes #709 
Fixes #718 

In this PR, I've added the following fix for issue #709 :
Instead of setting the environment from the authorization request, we would now set it from the token endpoint as that will be the correct one.

Also set the token endpoint by querying openid provider config metadata

To fix issue #718, we will know throw an exception if the cloud associated to the account passed by the user and the authority passed by the user as part of silent parameters do not match. We achieve this by doing a reverse lookup from the `mEnvironment` field on the account object and get the cloud associated to that preferredCacheHostName. From the cloud, then we get the `preferredNetworkHostName` and we see if it matches with the authority passed by the user. If not, we throw an exception.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
@@ -84,7 +84,7 @@ public class MicrosoftStsAccountCredentialAdapter
             if (!StringUtil.isEmpty(response.getAuthority())) {
                 accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromAuthority(new URL(response.getAuthority())));
             } else {
-                accessToken.setEnvironment(strategy.getIssuerCacheIdentifier(request));
+                accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
             }
 
             accessToken.setClientId(request.getClientId());
@@ -153,7 +153,7 @@ public class MicrosoftStsAccountCredentialAdapter
             if (null != response.getAuthority()) {
                 refreshToken.setEnvironment(strategy.getIssuerCacheIdentifierFromAuthority(new URL(response.getAuthority())));
             } else {
-                refreshToken.setEnvironment(strategy.getIssuerCacheIdentifier(request));
+                refreshToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
             }
 
             refreshToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
@@ -193,7 +193,7 @@ public class MicrosoftStsAccountCredentialAdapter
                         )
                 );
             } else {
-                idToken.setEnvironment(strategy.getIssuerCacheIdentifier(request));
+                idToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
             }
 
             idToken.setRealm(getRealm(strategy, response));
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
@@ -42,8 +42,6 @@ import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.M
 import com.microsoft.identity.common.internal.providers.oauth2.IDToken;
 import com.microsoft.identity.common.internal.util.StringUtil;
 
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.concurrent.TimeUnit;
 
 public class MicrosoftStsAccountCredentialAdapter
@@ -81,11 +79,7 @@ public class MicrosoftStsAccountCredentialAdapter
             accessToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
             accessToken.setRealm(getRealm(strategy, response));
 
-            if (!StringUtil.isEmpty(response.getAuthority())) {
-                accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromAuthority(new URL(response.getAuthority())));
-            } else {
-                accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
-            }
+            accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
 
             accessToken.setClientId(request.getClientId());
             /*
@@ -112,7 +106,7 @@ public class MicrosoftStsAccountCredentialAdapter
             accessToken.setAccessTokenType(response.getTokenType());
 
             return accessToken;
-        } catch (ServiceException | MalformedURLException e) {
+        } catch (ServiceException e) {
             // TODO handle this properly
             throw new RuntimeException(e);
         }
@@ -150,11 +144,7 @@ public class MicrosoftStsAccountCredentialAdapter
             final RefreshTokenRecord refreshToken = new RefreshTokenRecord();
             // Required
             refreshToken.setCredentialType(CredentialType.RefreshToken.name());
-            if (null != response.getAuthority()) {
-                refreshToken.setEnvironment(strategy.getIssuerCacheIdentifierFromAuthority(new URL(response.getAuthority())));
-            } else {
-                refreshToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
-            }
+            refreshToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
 
             refreshToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
             refreshToken.setClientId(request.getClientId());
@@ -168,7 +158,7 @@ public class MicrosoftStsAccountCredentialAdapter
             refreshToken.setCachedAt(String.valueOf(cachedAt)); // generated @ client side
 
             return refreshToken;
-        } catch (ServiceException | MalformedURLException e) {
+        } catch (ServiceException e) {
             // TODO handle this properly
             throw new RuntimeException(e);
         }
@@ -186,15 +176,8 @@ public class MicrosoftStsAccountCredentialAdapter
             // Required fields
             idToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
 
-            if (null != response.getAuthority()) {
-                idToken.setEnvironment(
-                        strategy.getIssuerCacheIdentifierFromAuthority(
-                                new URL(response.getAuthority())
-                        )
-                );
-            } else {
-                idToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
-            }
+            idToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
+
 
             idToken.setRealm(getRealm(strategy, response));
             idToken.setCredentialType(
@@ -213,7 +196,7 @@ public class MicrosoftStsAccountCredentialAdapter
             }
 
             return idToken;
-        } catch (ServiceException | MalformedURLException e) {
+        } catch (ServiceException e) {
             // TODO handle this properly
             throw new RuntimeException(e);
         }
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
@@ -111,6 +111,14 @@ public class AzureActiveDirectory
     }
 
     /**
+     * @param preferredCacheHostName String
+     * @return AzureActiveDirectoryCloud
+     */
+    public static AzureActiveDirectoryCloud getAzureActiveDirectoryCloud(final String preferredCacheHostName) {
+        return sAadClouds.get(preferredCacheHostName.toLowerCase(Locale.US));
+    }
+
+    /**
      * @param host  String
      * @param cloud AzureActiveDirectoryCloud
      */
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
@@ -114,7 +114,7 @@ public class AzureActiveDirectory
      * @param preferredCacheHostName String
      * @return AzureActiveDirectoryCloud
      */
-    public static AzureActiveDirectoryCloud getAzureActiveDirectoryCloud(final String preferredCacheHostName) {
+    public static AzureActiveDirectoryCloud getAzureActiveDirectoryCloudFromHostName(final String preferredCacheHostName) {
         return sAadClouds.get(preferredCacheHostName.toLowerCase(Locale.US));
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
@@ -24,6 +24,9 @@ package com.microsoft.identity.common.internal.providers.microsoft.microsoftsts;
 
 import android.net.Uri;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import com.microsoft.identity.common.exception.ServiceException;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectoryOAuth2Configuration;
@@ -41,6 +44,12 @@ public class MicrosoftStsOAuth2Configuration extends AzureActiveDirectoryOAuth2C
     private static final String FALLBACK_AUTHORIZE_ENDPOINT_SUFFIX = FALLBACK_ENDPOINT_SUFFIX + "/authorize";
     private static final String FALLBACK_TOKEN_ENDPOINT_SUFFIX = FALLBACK_ENDPOINT_SUFFIX + "/token";
 
+    /**
+     * Get the authorization endpoint to be used for making a authorization request.
+     * This must NOT be called from the main thread.
+     *
+     * @return URL the authorization endpoint
+     */
     public URL getAuthorizationEndpoint() {
         final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
         if (openIdConfig != null) {
@@ -49,6 +58,12 @@ public class MicrosoftStsOAuth2Configuration extends AzureActiveDirectoryOAuth2C
         return getEndpoint(getAuthorityUrl(), FALLBACK_AUTHORIZE_ENDPOINT_SUFFIX);
     }
 
+    /**
+     * Get the token endpoint to be used for making a token request.
+     * This must NOT be called from the main thread.
+     *
+     * @return URL the token endpoint
+     */
     public URL getTokenEndpoint() {
         final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
         if (openIdConfig != null && openIdConfig.getTokenEndpoint() != null) {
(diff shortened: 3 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -212,23 +233,7 @@ public MicrosoftStsAccount createAccount(
...
-            Logger.verbose(
-                    TAG + methodName,
-                    "Creating account from TokenResponse failed due to malformed URL (mTokenEndpoint)..."
-            );
-        }
-
-        if (authority != null) {
-            account.setEnvironment(getIssuerCacheIdentifierFromAuthority(authority));
-        }
+        account.setEnvironment(getIssuerCacheIdentifierFromTokenEndpoint());
```
rpdome: should we use this as a single source of truth?

shahzaibj: I've fixed this in MSAL, we will now get cache identifier from token endpoint in MSALOAuth2TokenCache. However, I am not going to remove the `getIssuerCacheIdentifier(AuthorizationRequest)` method as that is an abstract method and is being overridden by a few different classes.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdProviderConfigurationClient.java
```diff
@@ -54,16 +56,30 @@
     private static final ExecutorService sBackgroundExecutor = Executors.newCachedThreadPool();
     private static final Map<URL, OpenIdProviderConfiguration> sConfigCache = new HashMap<>();
 
+    private static final String NA = "";
```
iambmelt: I don't see a constant being necessary for one usage -- I'm fine if you just inline this

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -311,11 +320,23 @@ public MicrosoftStsTokenRequest createTokenRequest(@NonNull final MicrosoftStsAu
                 "Creating TokenRequest..."
         );
 
-        if (mConfig.getMultipleCloudsSupported() || request.getMultipleCloudAware()) {
-            Logger.verbose(TAG, "get cloud specific authority based on authorization response.");
-            setTokenEndpoint(getCloudSpecificTenantEndpoint(response));
+        OpenIdProviderConfigurationClient configurationClient = new OpenIdProviderConfigurationClient(request, response, ENDPOINT_VERSION);
```
iambmelt: nit `final`

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdProviderConfigurationClient.java
```diff
@@ -54,16 +56,30 @@
...
     public OpenIdProviderConfigurationClient(@NonNull final String issuer) {
-        mIssuer = sanitize(issuer);
+        mIssuer = Uri.parse(sanitize(issuer)).toString();
+    }
+
+    public OpenIdProviderConfigurationClient(@NonNull final String authority, @NonNull final String path) {
+        this(authority, path, NA);
+    }
+
+    public OpenIdProviderConfigurationClient(@NonNull final String authority, @NonNull final String path, @NonNull final String endpointVersion) {
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdProviderConfigurationClient.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -81,11 +79,7 @@ public AccessTokenRecord createAccessToken(
             accessToken.setHomeAccountId(SchemaUtil.getHomeAccountId(clientInfo));
             accessToken.setRealm(getRealm(strategy, response));
 
-            if (!StringUtil.isEmpty(response.getAuthority())) {
-                accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromAuthority(new URL(response.getAuthority())));
-            } else {
-                accessToken.setEnvironment(strategy.getIssuerCacheIdentifier(request));
-            }
+            accessToken.setEnvironment(strategy.getIssuerCacheIdentifierFromTokenEndpoint());
```
iambmelt: 💪

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -24,18 +24,48 @@
...
+        }
+        return getEndpoint(getAuthorityUrl(), FALLBACK_TOKEN_ENDPOINT_SUFFIX);
+    }
+
+
+    private URL getEndpoint(String endpoint) {
+        try {
+            return new URL(endpoint);
+        } catch (Exception e) {
+            e.printStackTrace();
```
iambmelt: Let's log two errors here:
- 1 reporting the error using `Logger.error`
- 2 reporting the stacktrace + failing uri with `Logger.errorPII` (since this may contain OII)

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -24,18 +24,48 @@
...
-        return getEndpoint(getAuthorityUrl(), "/oAuth2/v2.0/token");
+        final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
+        if (openIdConfig != null && openIdConfig.getTokenEndpoint() != null) {
+            return getEndpoint(openIdConfig.getTokenEndpoint());
+        }
+        return getEndpoint(getAuthorityUrl(), FALLBACK_TOKEN_ENDPOINT_SUFFIX);
+    }
+
+
+    private URL getEndpoint(String endpoint) {
```
iambmelt: Let's also declare this return type `@Nullable`

iambmelt: And param `@NonNull final`

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -24,18 +24,48 @@
...
+    private static final String TAG = MicrosoftStsOAuth2Configuration.class.getSimpleName();
+
+    private static final String ENDPOINT_VERSION = "v2.0";
+    private static final String FALLBACK_ENDPOINT_SUFFIX = "/oAuth2/v2.0";
+    private static final String FALLBACK_AUTHORIZE_ENDPOINT_SUFFIX = FALLBACK_ENDPOINT_SUFFIX + "/authorize";
+    private static final String FALLBACK_TOKEN_ENDPOINT_SUFFIX = FALLBACK_ENDPOINT_SUFFIX + "/token";
+
     public URL getAuthorizationEndpoint() {
-        return getEndpoint(getAuthorityUrl(), "/oAuth2/v2.0/authorize");
+        final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
```
iambmelt: Do we need to be mindful of what thread we're on when we call this method? If yes, can we javadoc?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -53,4 +83,29 @@ private URL getEndpoint(URL root, String endpoint) {
 
     }
 
+    private OpenIdProviderConfiguration getOpenIdWellKnownConfigForAuthority() {
+        final URL authority = getAuthorityUrl();
+        return getOpenIdWellKnownConfig(authority.getHost(), authority.getPath());
+    }
+
+    OpenIdProviderConfiguration getOpenIdWellKnownConfig(final String host, final String audience) {
```
iambmelt: This looks like something that can't be call from the main thread -- if true, we should probably javadoc

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -53,4 +83,29 @@ private URL getEndpoint(URL root, String endpoint) {
 
     }
 
+    private OpenIdProviderConfiguration getOpenIdWellKnownConfigForAuthority() {
```
iambmelt: Suggest we declare the return type `@Nullable` and javadoc that this method can return `null` if errors are encountered

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
iambmelt: Let's also annotate the params as `@Nullable` or `@NonNull` as appropriate

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
iambmelt: and for completeness, that callers should check the result before using it

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -155,6 +160,27 @@ public String getIssuerCacheIdentifierFromAuthority(final URL authority) {
...
 
+    public String getIssuerCacheIdentifierFromTokenEndpoint() {
+        final String methodName = ":getIssuerCacheIdentifierFromTokenEndpoint";
+        URL authority = null;
+        String cacheIdentifier = null;
+
+        try {
+            authority = new URL(mTokenEndpoint);
+        } catch (MalformedURLException e) {
+            Logger.verbose(
```
iambmelt: Suggest we crank this up higher than `verbose`.... `warning` or `error` feels appropriate

shahzaibj:
@iambmelt Yeah the pmd failed. Just pushed a fix for that. Also, I believe some end-to-end tests will also be broken and will need refactoring within MSAL. I'll create a PR to fix those prior to merging this.

iambmelt:
> @iambmelt Yeah the pmd failed. Just pushed a fix for that. Also, I believe some end-to-end tests will also be broken and will need refactoring within MSAL. I'll create a PR to fix those prior to merging this.

Sounds great, thanks @shahzaibj

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed

shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed. Yes, this is synchronous so it should not be called from the main thread. I've added the javadoc.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed. Yes, this is synchronous so it should not be called from the main thread. I've added the javadoc.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/request/AcquireTokenSilentOperationParameters.java
```diff
@@ -61,8 +63,32 @@ public void validate() throws ArgumentException {
...
+                performCloudDiscovery();
+            }
+            final AzureActiveDirectoryCloud cloud = AzureActiveDirectory.getAzureActiveDirectoryCloud(mAccount.getEnvironment());
+            return cloud != null && cloud.getPreferredNetworkHostName().equals(getAuthority().getAuthorityURL().getAuthority());
+        } catch (IOException e) {
+            Logger.error(
+                    TAG + methodName,
+                    "Unable to perform cloud discovery",
+                    e);
+            return false;
```
iambmelt: What happens if this fails and we return false? Does the caller currently handle this in a reasonable way?

shahzaibj: We are just throw an ArgumentException in the case.
Since we are not sure if account matches the authority (as we don't have cloud information), we just say that the account does not match the authority and throw the exception.

shahzaibj: This is where this is called:

`    @Override
    public void validate() throws ArgumentException {
        super.validate();

        if (mAccount == null) {
            Logger.warn(TAG, "The account set on silent operation parameters is NULL.");
        } else if (!authorityMatchesAccountEnvironment()) {
            throw new ArgumentException(
                    ArgumentException.ACQUIRE_TOKEN_SILENT_OPERATION_NAME,
                    ArgumentException.AUTHORITY_ARGUMENT_NAME,
                    "Authority passed to silent parameters does not match with the cloud associated to the account."
            );
        }
    }`

common/src/main/java/com/microsoft/identity/common/internal/request/AcquireTokenSilentOperationParameters.java
```diff
@@ -61,8 +63,32 @@ public void validate() throws ArgumentException {
...
+        }
+    }
+
+    private static void performCloudDiscovery() throws IOException {
+        final Object sLock = new Object();
```
iambmelt: I don't think this offering any protection as an instance variable....

iambmelt: each call would construct a new instance of the lock so it would always be free

shahzaibj: Addressed

shahzaibj:
Thanks @iambmelt for review!

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -488,7 +497,7 @@ protected TokenResult getTokenResultFromHttpResponse(@NonNull final HttpResponse
         return result;
     }
 
-    private String getCloudSpecificTenantEndpoint(
+    private String buildCloudSpecificTokenEndpoint(
             @NonNull final MicrosoftStsAuthorizationResponse response) {
         if (!StringUtil.isEmpty(response.getCloudGraphHostName())) {
```
shahzaibj: @iambmelt I believe this should have been `if (!StringUtil.isEmpty(response.getCloudInstanceHostName()))`? as that is the one we use below. I'm going to make the change but wanted to double check if this was on purpose?

iambmelt: Looking....

iambmelt: This change was actually introduced in:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/371212f51df69f7ad4bfafc3980b37c50ed2067b

It appears to be a typo

shahzaibj: Thanks, I'll make the change in this PR.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -110,6 +110,14 @@ public static AzureActiveDirectoryCloud getAzureActiveDirectoryCloud(final URL a
         return sAadClouds.get(authorityUrl.getHost().toLowerCase(Locale.US));
     }
 
+    /**
+     * @param preferredCacheHostName String
+     * @return AzureActiveDirectoryCloud
+     */
+    public static AzureActiveDirectoryCloud getAzureActiveDirectoryCloud(final String preferredCacheHostName) {
```
kreedula: nit : rename `getAzureActiveDirectoryCloudFromHostName` to distinguish from the method above

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -24,18 +24,73 @@
...
-        return getEndpoint(getAuthorityUrl(), "/oAuth2/v2.0/token");
+        final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
+        if (openIdConfig != null && openIdConfig.getTokenEndpoint() != null) {
+            return getEndpoint(openIdConfig.getTokenEndpoint());
+        }
+        return getEndpoint(getAuthorityUrl(), FALLBACK_TOKEN_ENDPOINT_SUFFIX);
+    }
+
+    @Nullable
+    private URL getEndpoint(@NonNull final String endpoint) {
```
kreedula: nit : rename method for clarity, also there is a method with a same name (from before) in this class , which lacks clarity too, these can get confusing

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
shahzaibj: Addressed

PR #719, opened 2019-11-07
Record non-https URL in log

rpdome:
So that we know what url causes it.


rpdome:
Force-pushed to merge this commit into dev instead.

PR #726, opened 2019-11-12
Expose key deletion

rpdome:
For testing the new migration logic.


shahzaibj:
Where are the tests? or where are they going to be written?

PR #728, opened 2019-11-13
Support System Browser in Broker

kreedula:
corresponding broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/1202
corresponding MSAL PR : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/839

- Changes to send `AuthorizationAgent` as Browser for COBO if the parameter is set
- Validation logic to make sure we only enable this for COBO package in Broker (to avoid mis use)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
@@ -52,7 +52,7 @@ public class BrokerRequest implements Serializable {
         final static String CLIENT_VERSION = "client_version";
         final static String ENVIRONMENT = "environment";
         final static String MULTIPLE_CLOUDS_SUPPORTED = "multiple_clouds_supported";
-        final static String AUTHOTIZATION_AGENT = "authorization_agent";
+        final static String AUTHORIZATION_AGENT = "authorization_agent";
     }
 
     /**
@@ -178,7 +178,7 @@ public class BrokerRequest implements Serializable {
     private boolean mMultipleCloudsSupported;
 
     @NonNull
-    @SerializedName(SerializedNames.AUTHOTIZATION_AGENT)
+    @SerializedName(SerializedNames.AUTHORIZATION_AGENT)
     private String mAuthorizationAgent;
 
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
@@ -52,6 +52,7 @@ public class BrokerRequest implements Serializable {
         final static String CLIENT_VERSION = "client_version";
         final static String ENVIRONMENT = "environment";
         final static String MULTIPLE_CLOUDS_SUPPORTED = "multiple_clouds_supported";
+        final static String AUTHOTIZATION_AGENT = "authorization_agent";
     }
 
     /**
@@ -176,6 +177,10 @@ public class BrokerRequest implements Serializable {
     @SerializedName(SerializedNames.MULTIPLE_CLOUDS_SUPPORTED)
     private boolean mMultipleCloudsSupported;
 
+    @NonNull
+    @SerializedName(SerializedNames.AUTHOTIZATION_AGENT)
+    private String mAuthorizationAgent;
+
 
     private BrokerRequest(BrokerRequest.Builder builder) {
         mAuthority = builder.mAuthority;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -176,6 +177,10 @@
     @SerializedName(SerializedNames.MULTIPLE_CLOUDS_SUPPORTED)
     private boolean mMultipleCloudsSupported;
 
+    @NonNull
+    @SerializedName(SerializedNames.AUTHORIZATION_AGENT)
+    private String mAuthorizationAgent;
```
rpdome: If this scenario is COBO-specific, can we use extra query parameter instead? (so that we don't expose this).

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -327,9 +335,21 @@ private void completeAuthorization() {
         Logger.info(TAG, null, "Received redirect from customTab/browser.");
         final String url = getIntent().getExtras().getString(AuthorizationStrategy.CUSTOM_TAB_REDIRECT);
         final Intent resultIntent = createResultIntent(url);
-        final Map<String, String> parameters = StringExtensions.getUrlParameters(url);
-        if (parameters.containsKey(INSTALL_URL_KEY)) {
-            final String appLink = parameters.get(INSTALL_URL_KEY);
+        final Map<String, String> urlQueryParameters = StringExtensions.getUrlParameters(url);
+        final String userName = urlQueryParameters.get(AuthenticationConstants.Broker.INSTALL_UPN_KEY);
```
rpdome: nit: move this inside if(isDeviceRegisterRedirect())

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -364,4 +374,32 @@ private boolean getMultipleCloudsSupported(@NonNull final OperationParameters pa
...
     }
+
+    /**
+     * List of System Browsers which can be used from broker, currently only Chrome is supported.
+     * @return
+     */
+    public static List<BrowserDescriptor> getBrowserSafeListForBroker(){
+        List<BrowserDescriptor>  browserDescriptors = new ArrayList<>();
+        final HashSet<String> signatureHashes = new HashSet();
+        signatureHashes.add("7fmduHKTdHHrlMvldlEqAIlSfii1tl35bxj1OXN5Ve8c4lU6URVu4xtSHc3BVZxS6WWJnxMDhIfQN0N0K2NDJg==");
```
rpdome: nit: add comment on where this comes from.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
iambmelt: Dissenting opinion: I'd like to avoid using query params to smuggle arguments into the broker unless those arguments' audience is the STS (ESTS/B2C) -- separation of concerns

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -424,6 +444,17 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.setWebViewClient(webViewClient);
     }
 
+    private boolean isDeviceRegisterRedirect(@NonNull final String redirectUrl){
+        try {
+            URI uri = new URI(redirectUrl);
+            return uri.getScheme().equalsIgnoreCase(REDIRECT_PREFIX) &&
+                    uri.getHost().equalsIgnoreCase(DEVICE_REGISTRATION_REDIRECT_URI_HOSTNAME);
+        } catch (URISyntaxException e) {
+            return false;
```
iambmelt: `Logger.error` ?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -366,4 +381,45 @@ private boolean getMultipleCloudsSupported(@NonNull final OperationParameters pa
             return false;
         }
     }
+
+    /**
+     * List of System Browsers which can be used from broker, currently only Chrome is supported.
+     * This information here is populated from the default browser safelist in MSAL.
+     * @return
+     */
+    public static List<BrowserDescriptor> getBrowserSafeListForBroker(){
```
iambmelt: For a future release we might consider how/what we might refactor here to not duplicate constants across common and msal. Maybe this just means moving the trusted hashes to `common`.....

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
kreedula: yup, we are using `AuthorizationAgent` as intended to send `Browser` vs `Webview`

PR #729, opened 2019-11-13
Add telemetry event in StorageHelper

rpdome:
to see how often this is invoked before https://github.com/AzureAD/ad-accounts-for-android/pull/1196 is done. (Ideally, none).

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -295,7 +295,7 @@ public class StorageHelper implements IStorageHelper {
                 if (key == null) {
                     mTelemetryCallback.logEvent(mContext, methodName, false, "KEY_DECRYPTION_KEYSTORE_KEY_NOT_INITIALIZED");
                 }
-            } catch (Exception e) {
+            } catch (final Exception e) {
                 // Best effort.
                 mTelemetryCallback.logEvent(mContext, methodName, false, "KEY_DECRYPTION_KEYSTORE_KEY_FAILED_TO_LOAD");
             }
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -287,6 +287,20 @@ public String decrypt(final String encryptedBlob) throws GeneralSecurityExceptio
...
 
+        // Try to read keystore key - to verify how often this is invoked before the migration is done.
+        // TODO: remove this whole try-catch clause once the experiment is done.
+        if (mTelemetryCallback != null) {
+            try {
+                final SecretKey key = loadSecretKey(KeyType.KEYSTORE_ENCRYPTED_KEY);
+                if (key == null) {
+                    mTelemetryCallback.logEvent(mContext, methodName, false, "KEY_DECRYPTION_KEYSTORE_KEY_NOT_INITIALIZED");
+                }
+            } catch (Exception e) {
```
iambmelt: nit: `final`

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -287,6 +287,20 @@ public String decrypt(final String encryptedBlob) throws GeneralSecurityExceptio
...
+                // Best effort.
```
iambmelt: In addition to telemetry, should we log here?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -501,12 +515,18 @@ public synchronized SecretKey loadSecretKeyForEncryption() throws IOException,
...
-                // Best effort.
+            // Try to read keystore key - to verify how often this is invoked before the migration is done.
+            // TODO: remove this whole try-catch clause once the experiment is done.
+            if (mTelemetryCallback != null) {
+                try {
+                    final SecretKey key = loadSecretKey(KeyType.KEYSTORE_ENCRYPTED_KEY);
+                    if (key == null) {
+                        mTelemetryCallback.logEvent(mContext, methodName, false, "KEY_ENCRYPTION_KEYSTORE_KEY_NOT_INITIALIZED");
+                    }
+                } catch (Exception e) {
```
iambmelt: Same as above, nit on `final` + optional log message

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
rpdome: I think we don't need to, as this won't make much sense in log file.

rpdome:
Thanks @iambmelt @shahzaibj !

PR #732, opened 2019-11-14
Fix ests telemetry memory leak and improve telemetry caching logic

shahzaibj:
We have a memory leak in the ests telemetry hashmap where telemetry objects for all silent calls are not being removed from the hashmap. This is happening by calling a flush method that does not accept a correlation id and that results in using the correlation id from the Diagnostic Context which does not match with anything in the map and hence the object is not removed.

This bug got introduced when we refactored Command Dispatcher to implement the command cache. 

To fix this, I've removed flush overloads that do not accept a correlation id, and refactored the `returnCommandResult` method to pass the correct correlation id to the flush method.

I've also improved telemetry caching logic in this PR, by adding a check to not cache telemetry associated to requests where we don't set the public api id so that those are not cached and are not sent to ests as the last request.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
@@ -301,6 +301,9 @@ public class EstsTelemetry {
         }
     }
 
+    // if we don't have api id then we won't save telemetry to cache
+    // this can happen for commands like the GetDeviceModeCommand
+    // that are generated via a method for which we don't want telemetry
     private boolean eligibleToCache(RequestTelemetry lastTelemetry) {
         return !TextUtils.isEmpty(lastTelemetry.getSchemaVersion()) &&
                 !TextUtils.isEmpty(lastTelemetry.getCommonTelemetry().get(Schema.Key.API_ID));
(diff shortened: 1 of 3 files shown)


rpdome:
Do you know why they're not the same correlationID?

- initializeDiagnosticContext creates a requestContext based on the returned correlationID.
- The function you removed uses the correlationID from DiagnosticContext.getRequestContext()

Do we have a bug anywhere that causes the mismatch?

shahzaibj:
> Do you know why they're not the same correlationID?
> 
> * initializeDiagnosticContext creates a requestContext based on the returned correlationID.
> * The function you removed uses the correlationID from DiagnosticContext.getRequestContext()
> 
> Do we have a bug anywhere that causes the mismatch? (i.e. change of state).

@rpdome This is happening because we start a new thread inside return command result and hence that correlation id

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -298,21 +287,25 @@ public void flush(final String correlationId, final String errorCode) {
...
             );
+        } else if (eligibleToCache(lastTelemetry)) {
+            // remove old last request telemetry data from cache
+            mLastRequestTelemetryCache.clearAll();
+            // save new last request telemetry data to cache
+            mLastRequestTelemetryCache.saveRequestTelemetryToCache(lastTelemetry);
         }
     }
 
+    private boolean eligibleToCache(RequestTelemetry lastTelemetry) {
```
rpdome: nit: add some comment describing the reason we have to introduce this.

PR #736, opened 2019-11-21
Add support to silently acquire token for a another tenant for an already authorized user

kreedula:
- Also added a new method `getTenantIdForAlias` in `AzureActiveDirectoryAudience.java` to get tenant GUID.

MSAL PR :  https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/841
Broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/1207

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
@@ -22,9 +22,21 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.internal.authorities;
 
+import android.net.Uri;
+
 import com.google.gson.annotations.SerializedName;
+import com.microsoft.identity.common.exception.ClientException;
+import com.microsoft.identity.common.exception.ServiceException;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectory;
+import com.microsoft.identity.common.internal.providers.oauth2.OpenIdProviderConfiguration;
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -55,6 +67,63 @@ public String getTenantId() {
         return mTenantId;
     }
 
+
+    /**
+     *
+     * Must me called on a worker thread.
```
iambmelt: minor typo in javadoc - 'be'

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -55,6 +67,63 @@ public String getTenantId() {
...
+            com.microsoft.identity.common.internal.logging.Logger.error(
+                    TAG,
+                    errMsg,
+                    null
+            );
+
+            throw new ClientException(errMsg);
+        }
+
+        return paths.get(0);
```
rpdome: nit: to make this future-proof, should we also invoke StringUtil.isUuid again on this value?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -468,6 +470,18 @@ protected AccountRecord getCachedAccountRecord(
         return targetAccount;
     }
 
+    protected boolean isRequestAuthorityRealmSameAsATRealm(@NonNull final Authority requestAuthority,
```
rpdome: nit: comment.

PR #738, opened 2019-11-25
Set correct authority on Access Token Record, ID Token Record

shahzaibj:
The authority set on the AccessTokenRecord is wrong. I've fixed this by setting it from the token endpoint which should always be correct as we just went there.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/IDToken.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/IDToken.java
@@ -187,11 +187,6 @@ public class IDToken {
      */
     public static final String UPDATED_AT = "updated_at";
 
-    /**
-     * The authority that issued the token
-     */
-    public static final String ISSUER = "iss";
-
 
     private Map<String, ?> mTokenClaims = null;
     private final String mRawIdToken;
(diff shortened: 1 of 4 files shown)


shahzaibj:
Tests are failing inside `MicrosoftStsAccountCredentialAdapterTest.class`. Possible issues with mocked data.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/IDToken.java
```diff
@@ -187,6 +187,11 @@
      */
     public static final String UPDATED_AT = "updated_at";
 
+    /**
+     * The authority that issued the token
+     */
+    public static final String ISSUER = "iss";
```
kreedula: where is this used ?

shahzaibj: I was initially using this in MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/843

But removed it from there, and hard coded the `iss` string to remove that PR's dependency on this common PR as the tests failed in this PR.

shahzaibj: Since that MSAL PR no longer depends on this PR, I'll remove this constant for now.

shahzaibj:
@kreedula I've addressed your comments

PR #746, opened 2019-12-05
StorageHelper Refactor.

rpdome:
**Why making this change?**
The current StorageHelper relies on the keys (loaded by Broker/ADAL/MSAL) to decide which encryption key it should be using. This was fine until we started shipping BrokerV2 + ADAL with common in CP, as AuthenticationSettings is now shared by both BrokerV2 and ADAL. This leads into a situation where both keys might accidentally be loaded. (https://github.com/AzureAD/ad-accounts-for-android/issues/1109). If such thing happens, wrong encryption key might be used. 

As of now we should be in a good shape, but I'd like to fix this for good, as god knows we might forgot about this and regress it in the future.

**Proposed change**
- Rename StorageHelper to BaseEncryptionManager, and extract its logic into 3 subclasses, 
     1. BrokerEncryptionManager - decides how keys are picked in Broker flows
     2. MsalEncryptionManager - decides how keys are picked in MSAL flows
     3. KeystoreEncryptedKeyManager - to deal with Keystore-related operation.
- Make sure that Broker flows only use BrokerEncryptionManager.
- Make sure that ADAL/MSAL flows only use MsalEncryptionManager. 
- Add the following test classes
     1. BrokerEncryptionManager with Legacy keys.
     2. BrokerEncryptionManager with keystore keys.
     3. MsalEncryptionManager with user defined keys.
     4. MsalEncryptionManager with keystore keys.
     5. KeystoreEncryptedKeyManager

Changes are also needed to be made in ADAL/MSAL/Broker as well.


common/src/main/java/com/microsoft/identity/common/internal/encryption/BaseEncryptionManager.java
```diff
@@ -0,0 +1,508 @@
...
+        return getEncodeVersionLengthPrefix() + ENCODE_VERSION + encryptedText;
+    }
+
+    @Override
+    public synchronized String decrypt(final String encryptedBlob) throws GeneralSecurityException, IOException {
+        final String methodName = ":decrypt";
+        Logger.verbose(TAG + methodName, "Starting decryption");
+
+        if (StringExtensions.isNullOrBlank(encryptedBlob)) {
+            throw new IllegalArgumentException("Input is empty or null");
```
kreedula: Should we just return the `empty` or `null` string back here instead of throwing an exception ?

Om83:
Like the detailed description for the change!! All changes should include some form of this.

common/src/main/java/com/microsoft/identity/common/internal/encryption/BrokerEncryptionManager.java
```diff
@@ -0,0 +1,187 @@
...
+        if (sShouldEncryptWithKeyStoreKey || mforceEnableKeyStoreKey) {
+            // Try loading existing keystore-encrypted key.
+            // If it doesn't exist nor it is not loadable, fall back to legacy keys.
+            try {
+                final SecretKey key = loadSecretKey(KeyType.KEYSTORE_ENCRYPTED_KEY);
+                if (key != null) {
+                    return new EncryptionKey(KeyType.KEYSTORE_ENCRYPTED_KEY, key);
+                }
+            } catch (final IOException | GeneralSecurityException e) {
+                // If we fail to load key, proceed and generate a new one.
```
Om83: We are falling back to legacy key, instead of generating a new one. Comment should reflect the same

iambmelt:
> Like the detailed description for the change!! All changes should include some form of this.

+1!

PR #748, opened 2019-12-07
Refactor for Broker E2E Tests

shahzaibj:


testutils/src/main/java/com/microsoft/identity/internal/testutils/TestUtils.java
```diff
@@ -0,0 +1,137 @@
...
+
+
+    /**
+     * Inspects the supplied cache key to determine the target CredentialType.
+     *
+     * @param cacheKey The cache key to inspect.
+     * @return The CredentialType or null if a proper type cannot be resolved.
+     */
+    @Nullable
+    private static CredentialType getCredentialTypeForCredentialCacheKey(@NonNull final String cacheKey) {
```
iambmelt: If you want to refactor the original implementation so it doesn't get duplicated here, that seems preferable

shahzaibj: Okay, I'll change the original implementation to be `public static` so we can call it in our tests. Based on the nature of the method, I think that is fine to do as it does not seems necessary for this method to be an instance method or be private either.

shahzaibj: Addressed

PR #749, opened 2019-12-09
Lab Api Refactor (Fixes Authentication in Lab Api, & Use KeyVault to get lab app id and app secret)

shahzaibj:
In this PR, I have refactored the lab api a little bit. We will no longer go directly to the lab api. Instead, we will do as follows:

1. Go to ESTS, and get a token to access KeyVault using a client_assertion
2. Go to KeyVault and get App ID for the Lab Api
3. Go to KeyVault and get App Secret for Lab Api
4. Take the App Id (client_id) and App Secret (client_secret), and go to ESTS to get a token to access LAB API
5. Now use the access token to make requests against the lab api

Also change the scope and the lab endpoint as it has been changed by the Lab Team


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/ConfidentialClientHelper.java
```diff
@@ -0,0 +1,75 @@
...
+            e.printStackTrace();
+        }
+
+        return accessToken;
+    }
+
+    void setupApiClientWithAccessToken() {
+        try {
+            setupApiClientWithAccessToken(this.getAccessToken());
+        } catch (CertificateException e) {
```
rpdome: It might be better if we throw an exception here (so that we know when test case fails that this labAPI auth part is at fault - and the Broker/MSAL/common is not necessarily broken).

shahzaibj: Addressed

labapi/src/main/java/com/microsoft/identity/internal/test/labapi/api/AppApi.java
```diff
@@ -100,7 +100,7 @@ public void setApiClient(ApiClient apiClient) {
             });
         }
 
-        String[] localVarAuthNames = new String[] {  };
+        String[] localVarAuthNames = new String[] { "Access Token" };
```
rpdome: Make this a constant (maybe in ApiClient?)

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/ConfidentialClientHelper.java
```diff
@@ -0,0 +1,63 @@
...
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateException;
+
+abstract class ConfidentialClientHelper {
+
+    private final static String GRANT_TYPE = "client_credentials";
+    private final static String TENANT_ID = "72f988bf-86f1-41af-91ab-2d7cd011db47";
```
rpdome: nit: add a comment specifying whose tenantID this is.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/KeyVaultAuthHelper.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static ConfidentialClientHelper getInstance() {
+        if (sKeyVaultAuthHelper == null) {
+            sKeyVaultAuthHelper = new KeyVaultAuthHelper();
+        }
+        return sKeyVaultAuthHelper;
+    }
+
+    @Override
+    public String getAccessToken() throws UnrecoverableKeyException, NoSuchAlgorithmException, KeyStoreException, NoSuchProviderException, IOException, InterruptedException, CertificateException {
```
rpdome: should we move this function and setupApiClientWithAccessToken to base class?

shahzaibj: we need to have an implementation for both the LAB API and the Key Vault so I had to separate it out.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabAuthenticationHelper.java
```diff
@@ -42,86 +36,84 @@
...
+    private String getAppSecret() {
+        if (mAppSecret == null) {
+            mAppSecret = getLabSecret(SECRET_NAME_LAB_APP_SECRET);
+        }
+
+        return mAppSecret;
     }
 
-    static void setupApiClientWithAccessToken() {
+    private String getLabSecret(final String secretName) {
```
rpdome: nit: getSecretFromKeyVault()?

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/KeyVaultAuthHelper.java
shahzaibj: if we move `getAccessToken` to the base class, then we will need two member variables for access token in that base class. One to hold access token for key vault and other to hold access token for lab

labapi/src/main/java/com/microsoft/identity/internal/test/labapi/api/AppApi.java
shahzaibj: Addressed

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/ConfidentialClientHelper.java
shahzaibj: Addressed

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/KeyVaultAuthHelper.java
shahzaibj: Addressed

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabAuthenticationHelper.java
shahzaibj: Addressed

labapi/src/main/java/com/microsoft/identity/internal/test/labapi/ApiClient.java
```diff
@@ -945,6 +947,7 @@ public void onResponse(Response response) throws IOException {
      * @throws ApiException If fail to serialize the request body object
      */
     public Call buildCall(String path, String method, List<Pair> queryParams, List<Pair> collectionQueryParams, Object body, Map<String, String> headerParams, Map<String, Object> formParams, String[] authNames, ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
+        authNames[0] = AUTH_TYPE;
```
rpdome: I think we should either,

1. remove authNames from buildCall and construct it here.
or
2. pass authNames, but define its value outside.

shahzaibj: Addressed with first approach

PR #754, opened 2019-12-11
Remove `mUseCustomTabs` from BrowserDescriptor to fix browser safe list

shahzaibj:
This change is needed to fix [MSAL Issue 859](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/859)

Note that the field `mUseCustomTab` was irrelevant, and was only being used to determine if a particular browser version is safe to use with MSAL. Whether custom tab is supported or not is based on the `Browser` object and the field `mIsCustomTabsServiceSupported` flag on the Browser object. For details, see [Browser.java](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/e2f5c3db43763541e7651100c6b4a1f6f0c41257/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/Browser.java)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserDescriptor.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserDescriptor.java
@@ -39,9 +39,6 @@ public class BrowserDescriptor implements Serializable {
     @SerializedName("browser_signature_hashes")
     private Set<String> mSignatureHashes;
 
-    @SerializedName("browser_use_customTab")
-    private boolean mUseCustomTab;
-
     @SerializedName("browser_version_lower_bound")
     private String mVersionLowerBound;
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserDescriptor.java
```diff
@@ -83,10 +76,6 @@ public boolean matches(@NonNull Browser browser) {
             return false;
         }
 
-        if (mUseCustomTab != browser.isCustomTabsServiceSupported()) {
```
rpdome: should it be this instead?

if (mUseCustomTab){
    if (! browser.isCustomTabsServiceSupported()){
        return false;
    }
}

shahzaibj: I don't think we need to be specifying whether to use custom tabs in a browser or not. If the browser supports it, they will be used. If not, then they won't be used. I have therefore removed `mUseCustomTab` from the BrowserDescriptor class. Do you think we need to keep it there?

PR #758, opened 2019-12-18
Create in memory cache for lab config

shahzaibj:
`BeforeClass` & `AfterClass` test methods does not work with Robolectric, so we use the `before` and `after` methods instead to perform common tasks that need to be performed prior to running a test. These include hitting the lab api to get a user for the automated test. Initially we had only few tests but as we write more tests, this will get expensive for both us, as well as the lab api. To solve this problem, I've created an in-memory cache (modeled by HashMap) for the lab config. So now if we try to get a user/config from lab and we recently made a request with the same query, then we can just return the result from cache and don't need to hit the lab api.


rpdome:
Please add description on the changes you made.

shahzaibj:
> Please add a brief description on the changes you made (why+how)

@rpdome See description

PR #763, opened 2020-01-01
Introduce AuthorizationFragment

rpdome:
Requirement: 
- There's an ask to display MSAL AuthorizationActivity in a certain area on a screen. 

Proposed change:
- Extracts AuthorizationFragment from AuthorizationActivity
- Expose an API on MSAL to take in a fragment.
- If a fragment is provided, replace that fragment with AuthorizationFragment (instead of popping up an activity).

Limitation:
- This will only work with MSAL-only (no broker), WebView flow. (Which satisfies the ask).

Related PR in MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/881


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -1,473 +1,35 @@
...
-        setDiagnosticContextForNewThread(state.getString(DiagnosticContext.CORRELATION_ID));
-        mAuthIntent = state.getParcelable(KEY_AUTH_INTENT);
-        mBrowserFlowStarted = state.getBoolean(KEY_BROWSER_FLOW_STARTED, false);
-        mPkeyAuthStatus = state.getBoolean(KEY_PKEYAUTH_STATUS, false);
-        mAuthorizationRequestUrl = state.getString(KEY_AUTH_REQUEST_URL);
-        mRedirectUri = state.getString(KEY_AUTH_REDIRECT_URI);
-        mRequestHeaders = getRequestHeaders(state);
-        mAuthorizationAgent = (AuthorizationAgent) state.getSerializable(KEY_AUTH_AUTHORIZATION_AGENT);
-    }
+public final class AuthorizationActivity extends FragmentActivity {
```
kreedula: `FragmentActivity` seems to be added in API 22, however the documentation says it's in AndroidX jetpack, did we test this on lower level devices (SDK 21 or below)?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationStrategy.java
```diff
@@ -50,6 +58,46 @@
...
+
+    protected Context getApplicationContext() {
+        return mReferencedApplicationContext.get();
+    }
+
+    /**
+     * If fragment is provided, add AuthorizationFragment on top of that fragment.
+     * Otherwise, launch AuthorizationActivity.
+     */
+    protected void launchIntent(@NonNull Intent intent) {
```
kreedula: This method is used both by `EmbeddedWebViewStrategy` and `BrowserStrategy`. Shouldn't this be only for `Webview` ? i.e. in case of browser, what does adding a fragment do ?

rpdome: AuthorizationFragment will be launched, which in turn would launch CustomTabsActivity - inside onResume().

CustomTabsActivity will be popped up. It won't be displayed in-line in the specified fragment area.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
rpdome: Verified with emulator SDK 16/21.

PR #764, opened 2020-01-02
Fix msal issue 882

shahzaibj:
Fix [MSAL Issue 882](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/882)

If the authority is B2C, then we don't need to check if the Environment on the account that is passed matches the host or not as B2C only exists in one cloud.


common/src/main/java/com/microsoft/identity/common/internal/request/AcquireTokenSilentOperationParameters.java
```diff
@@ -54,7 +55,7 @@ public void validate() throws ArgumentException {
 
         if (mAccount == null) {
             Logger.warn(TAG, "The account set on silent operation parameters is NULL.");
-        } else if (!authorityMatchesAccountEnvironment()) {
+        } else if (!isAuthorityB2C() && !authorityMatchesAccountEnvironment()) {
```
rpdome: nit: I'd personally do something like

if (mAccount == null){
 return.
}

if (isAuthorityB2C()){
// Add a comment why we're not verifying.
return
}

if (!authorityMatchesAccountEnvironment()){
throw
}

Nasicus:
One question:
Why do you not need to check the configuration at all now? Why can you not compare the passed in authority to the one configured in the passed in config? Or has this nothing to do with it, since this check only concerns the `account` object?

shahzaibj:
> One question:
> Why do you not need to check the configuration at all now? Why can you not compare the passed in authority to the one configured in the passed in config? Or has this nothing to do with it, since this check only concerns the `account` object?

@Nasicus This check indeed concerns the account object only. We are not comparing the passed in authority with the one in the configuration here, instead we are comparing the passed in authority with the authority (or the environment) associated to the account that was passed for the silent call.

shahzaibj:
Closing this as the issue has been fixed in release with #767

PR #767, opened 2020-01-06
Fix msal issue 882

shahzaibj:
Fix [MSAL Issue 882](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/882)

If the authority is B2C, then we don't need to check if the Environment on the account that is passed matches the host or not as B2C only exists in one cloud.

cherry picked commit from pr #764


iambmelt:
Approving - I assume there will be a follow-up PR to rev version numbers and update the changelog?

shahzaibj:
> Approving - I assume there will be a follow-up PR to rev version numbers and update the changelog?

Yes

shahzaibj:
> I have the same comment as the other one. Either that or add comments describing why we add this B2C check (for readability). Otherwise LGTM

Added a comment @rpdome

Issue #770, opened 2020-01-08
Null pointer on AcquireTokenResult.getSucceeded 


kreedula:
COBO app reported a null pointer on a failed silent flow where`AcquireTokenResult` is null. 

StackTrace
```
com.microsoft.intune.common.presentationcomponent.abstraction.UiModelErrorState$Authentication@6b4a87a
2020-01-08T00:02:51.4700000	INFO	com.microsoft.intune.authentication.authcomponent.abstraction.AadTokenRepo	 9274	00002	Calling AAD auth library for Intune(ignored=true) token.
2020-01-08T00:02:51.6980000	WARNING	com.microsoft.intune.authentication.authcomponent.implementation.MsalAuthWrapper	 9274	00002	Silent MSAL auth failed.
	com.microsoft.identity.client.exception.MsalClientException: Attempt to invoke virtual method 'java.lang.Boolean com.microsoft.identity.common.internal.result.AcquireTokenResult.getSucceeded()' on a null object reference
		com.microsoft.identity.client.internal.controllers.MsalExceptionAdapter.msalExceptionFromBaseException(MsalExceptionAdapter.java:51)
		com.microsoft.identity.client.PublicClientApplication$9.onError(PublicClientApplication.java:1668)
		com.microsoft.identity.client.PublicClientApplication$9.onError(PublicClientApplication.java:1658)
		com.microsoft.identity.common.internal.controllers.CommandDispatcher$2.run(CommandDispatcher.java:176)
		android.os.Handler.handleCallback(Handler.java:883)
		android.os.Handler.dispatchMessage(Handler.java:100)
		android.os.Looper.loop(Looper.java:214)
		android.app.ActivityThread.main(ActivityThread.java:7356)
		java.lang.reflect.Method.invoke(Native Method)
		com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
		com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
	com.microsoft.identity.common.exception.ClientException: Attempt to invoke virtual method 'java.lang.Boolean com.microsoft.identity.common.internal.result.AcquireTokenResult.getSucceeded()' on a null object reference
		com.microsoft.identity.common.internal.controllers.ExceptionAdapter.baseExceptionFromException(ExceptionAdapter.java:252)
[...]

PR #771, opened 2020-01-08
Fix null pointer

kreedula:
Resolves https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/770

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
@@ -24,10 +24,7 @@ package com.microsoft.identity.common.internal.controllers;
 
 import android.content.Intent;
 
-import androidx.annotation.NonNull;
-
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
-import com.microsoft.identity.common.exception.BaseException;
 import com.microsoft.identity.common.exception.ClientException;
 import com.microsoft.identity.common.exception.ErrorStrings;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/TokenCommand.java
```diff
@@ -74,7 +71,7 @@ public AcquireTokenResult execute() throws Exception {
                         (AcquireTokenSilentOperationParameters) getParameters()
                 );
 
-                if (result.getSucceeded()) {
+                if (result != null && result.getSucceeded()) {
```
iambmelt: How can it be that we returned `null` but _didn't_ throw an Exception? That would seem to indicate some bad behavior at a lower level, no?

kreedula: Agreed. Investigating further , my primary suspicion could of null here : https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java#L187

Per logs Broker returned the right exception and adapter did transform to the right exception. However we have some logic here https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java#L206 where we could miss to throw an Exception after catching and return null result. I will sync with @rpdome to understand this better

kreedula:
Closing this as @rpdome is making some changes which should avoid this. Also this change could mask underlying issues , so better to close this.

PR #774, opened 2020-01-11
Added new public api to save credentials directly

kreedula:
- Changed was required to support MSAL CPP adapters

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
@@ -74,7 +74,12 @@ public class MsalCppOAuth2TokenCache
                 GenericTokenResponse extends TokenResponse,
                 GenericAccount extends BaseAccount,
                 GenericRefreshToken extends com.microsoft.identity.common.internal.providers.oauth2.RefreshToken>
-        extends MsalOAuth2TokenCache<GenericOAuth2Strategy, GenericAuthorizationRequest, GenericTokenResponse, GenericAccount, GenericRefreshToken> {
+        extends MsalOAuth2TokenCache<
+        GenericOAuth2Strategy,
+        GenericAuthorizationRequest,
+        GenericTokenResponse,
+        GenericAccount,
+        GenericRefreshToken> {
 
 
     /**
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
@@ -1,0 +1,129 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -151,6 +152,52 @@ public MsalOAuth2TokenCache(final Context context,
...
 
+    /**
+     * @param accountRecord : AccountRecord associated with the input credentials.
+     * @param credentials   : list of Credential which can include AccessTokenRecord, IdTokenRecord and RefreshTokenRecord.
+     *                      Note : Both IdTokenRecord and RefreshTokenRecord need to be non null. AccessTokenRecord can be optional.
+     * @throws ClientException : If the supplied Account or Credential are null or schema invalid.
+     */
+    public void saveCredentials(@NonNull final AccountRecord accountRecord,
+                                @NonNull final Credential... credentials) throws ClientException {
+        if (credentials == null) {
```
iambmelt: Optionally you could test if this array is `isEmpty()`

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -151,6 +152,52 @@ public MsalOAuth2TokenCache(final Context context,
         );
     }
 
+    /**
+     * @param accountRecord : AccountRecord associated with the input credentials.
+     * @param credentials   : list of Credential which can include AccessTokenRecord, IdTokenRecord and RefreshTokenRecord.
+     *                      Note : Both IdTokenRecord and RefreshTokenRecord need to be non null. AccessTokenRecord can be optional.
+     * @throws ClientException : If the supplied Account or Credential are null or schema invalid.
+     */
+    public void saveCredentials(@NonNull final AccountRecord accountRecord,
```
rpdome: I'm thinking if we should override this class and add this method only in the inherited classs (i.e. MsalCppOAuth2TokenCache).

This way, we would be able to minimize exposing this to MSAL devs.

We can mark the original saveCredentials as protected.

iambmelt: +1 Like this idea

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -151,6 +152,52 @@ public MsalOAuth2TokenCache(final Context context,
...
+                                @NonNull final Credential... credentials) throws ClientException {
+        if (credentials == null) {
+            throw new ClientException("Credential array passed in is null");
+        }
+
+        AccessTokenRecord accessTokenRecord = null;
```
rpdome: Should we explicitly list these three in the params?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
kreedula: Thanks for the suggestion Dome. Addressed

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -0,0 +1,151 @@
...
+        );
+    }
+
+    /**
+     * @param accountRecord : AccountRecord associated with the input credentials.
+     * @param credentials   : list of Credential which can include AccessTokenRecord, IdTokenRecord and RefreshTokenRecord.
+     *                      Note : Both IdTokenRecord and RefreshTokenRecord need to be non null. AccessTokenRecord can be optional.
+     * @throws ClientException : If the supplied Account or Credential are null or schema invalid.
+     */
+    public void saveCredentials(@NonNull final AccountRecord accountRecord,
```
iambmelt: @kreedula It occurs to me now that we ought to add some thread safety around this API -- if a few different threads start calling `removeRefreshTokenIfNeeded()` simultaneously we're going to have race conditions writing/reading from disk (saving new RTs while writing deleting old ones)

PR #775, opened 2020-01-14
Fix version task

### LINKED ISSUE #776: Snapshot not versioned correctly
The task `versionSnapshot` does not take hotfixes into account when versioning snapshot. As a result, the following two versions have the same snapshot:

1.0.13        -> 1.0.13-SNAPSHOT
1.0.13-hf1 -> 1.0.13-SNAPSHOT

shahzaibj:
Fixes #776


iambmelt:
Reviewing now...

PR #781, opened 2020-01-16
Fix msal issue 862

shahzaibj:
Fix [MSAL Issue 862](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/862)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
@@ -186,6 +186,11 @@ public abstract class BaseController {
         HttpWebRequest.throwIfNetworkNotAvailable(parameters.getAppContext());
 
         TokenRequest tokenRequest = strategy.createTokenRequest(request, response);
+
+        if (tokenRequest instanceof MicrosoftTokenRequest) {
+            ((MicrosoftTokenRequest) tokenRequest).setClaims(parameters.getClaimsRequestJson());
+        }
+
         logExposedFieldsOfObject(TAG + methodName, tokenRequest);
 
         TokenResult tokenResult = strategy.requestToken(tokenRequest);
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -186,6 +186,11 @@ protected TokenResult performTokenRequest(@NonNull final OAuth2Strategy strategy
         HttpWebRequest.throwIfNetworkNotAvailable(parameters.getAppContext());
 
         TokenRequest tokenRequest = strategy.createTokenRequest(request, response);
+
+        if (tokenRequest instanceof MicrosoftTokenRequest) {
```
rpdome: should we do this in getAuthorizationRequest? (adding claims as part of the AuthorizationRequest object, then use those values to construct TokenRequest in strategy.createTokenRequest()).

Lengthier solution, but I think it fits with the pattern we have here better.

rpdome: Actually... I just noticed that we already set the claims over there.

`

            // Add additional fields to the AuthorizationRequest.Builder to support interactive
            builder.setLoginHint(
                    acquireTokenOperationParameters.getLoginHint()
            ).setExtraQueryParams(
                    acquireTokenOperationParameters.getExtraQueryStringParameters()
            ).setPrompt(
                    acquireTokenOperationParameters.getOpenIdConnectPromptParameter().toString()
            ).setClaims(
                    parameters.getClaimsRequestJson()
            ).setRequestHeaders(
                    acquireTokenOperationParameters.getRequestHeaders()
            );

`

shahzaibj: Addressed

PR #784, opened 2020-01-17
Extract Browser/WebView specific codes from lifecycle code

rpdome:
the goal is to make the base class (AuthorizationFragment) to be able to focus on basic operations, like receiving and returning values from/to its caller or any related class.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final String TAG = WebViewAuthorizationFragment.class.getSimpleName();
+
+    @VisibleForTesting
+    private static final String PKEYAUTH_STATUS = "pkeyAuthStatus";
+
+    private WebView mWebView;
+
+    private Intent mAuthIntent;
+
+    private boolean mPkeyAuthStatus = false; //NOPMD //TODO Will finish the implementation in Phase 1 (broker is ready).
```
kreedula: nit : this comment can be removed i guess

PR #785, opened 2020-01-18
Add Exception Type to Broker Result to transform to correct exception

kreedula:
- Change to transform exceptions in msal based on type sent by broker, rather than relying on error code parsing.

- Also fixes https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/898, where we look for invalid broker RT.


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -182,12 +183,112 @@ public BaseException getBaseExceptionFromBundle(@NonNull final Bundle resultBund
...
+     * @return BaseException
+     */
+    private BaseException getBaseExceptionFromErrorCodes(@NonNull final BrokerResult brokerResult){
         final String errorCode = brokerResult.getErrorCode();
+        final BaseException baseException;
 
         if (AuthenticationConstants.OAuth2ErrorCode.INTERACTION_REQUIRED.equalsIgnoreCase(errorCode) ||
-                AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(errorCode)) {
+                AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(errorCode) ||
+                ErrorStrings.INVALID_BROKER_REFRESH_TOKEN.equalsIgnoreCase(errorCode)) {
```
shahzaibj: `ErrorStrings.NO_TOKENS_FOUND` is another case for UiRequiredException

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -182,12 +183,112 @@ public BaseException getBaseExceptionFromBundle(@NonNull final Bundle resultBund
...
+            Logger.warn(TAG, "Received a UIRequired exception from Broker : "
+                    + brokerResult.getErrorCode()
+            );
+            baseException = new UiRequiredException(
+                    brokerResult.getErrorCode(),
+                    brokerResult.getErrorMessage()
+            );
+        }
+
+        if(exceptionType.equalsIgnoreCase(ServiceException.sName)){
```
shahzaibj: might be better to use `else if` here and below

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -182,12 +183,112 @@ public BaseException getBaseExceptionFromBundle(@NonNull final Bundle resultBund
...
+            Logger.warn(TAG, "Received a Service exception from Broker : "
+                    + brokerResult.getErrorCode()
+            );
+            baseException = getServiceException(brokerResult);
+        }
+
+        if(exceptionType.equalsIgnoreCase(IntuneAppProtectionPolicyRequiredException.sName)){
+            Logger.warn(TAG,
```
shahzaibj: May be we can move log statement outside the if blocks, and call it only once as follows:

`Logger.warn(TAG, "Received a " + exceptionType +  " from Broker : "
                    + brokerResult.getErrorCode()
            );`

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -182,12 +183,112 @@ public BaseException getBaseExceptionFromBundle(@NonNull final Bundle resultBund
...
+                    brokerResult.getErrorMessage()
+            );
+        }
+
+        if(exceptionType.equalsIgnoreCase(ArgumentException.sName)){
+            Logger.warn(TAG, "Received a Argument exception from Broker : "
+                    + brokerResult.getErrorCode()
+            );
+            baseException = new ArgumentException(
+                    ArgumentException.ACQUIRE_TOKEN_OPERATION_NAME,
```
shahzaibj: How do we know that it is `ACQUIRE_TOKEN_OPERATION_NAME`?

common/src/main/java/com/microsoft/identity/common/exception/ArgumentException.java
```diff
@@ -60,4 +62,9 @@ public String getArgumentName() {
         return mArgumentName;
     }
 
+    @Override
+    public String getExceptionName(){
+        return sName;
```
rpdome: return ArgumentException.class.getName();?

Here and everywhere.

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
kreedula: Done

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
kreedula: Thanks for pointing out.

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
kreedula: Good point, I introduced a new operation to indicate that it'a broker request

common/src/main/java/com/microsoft/identity/common/exception/ArgumentException.java
kreedula: I'm using `sName` to compare , so that in future if the value of `sName` is changed , the logic would work as expected

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
shahzaibj: What about silent?

shahzaibj: `ACQUIRE_TOKEN_SILENT_OPERATION_NAME` ?

kreedula: I have introduced a new operation name `BROKER_TOKEN_REQUEST_OPERATION_NAME` which is passed here, since the basic MSAL argument validation is successful and is sent to broker, here if an argument exception is thrown it would be of broker. For now I added a generic operation name here, if need specific things in the future it can be added as needed

PR #788, opened 2020-01-22
Add throwable to BrokerCommunicationException's parameter list

rpdome:


rpdome:
Will remove this from common instead.

PR #794, opened 2020-01-29
Resolves #793 - Fix for Token Caching

### LINKED ISSUE #793: [Cache] Access Token Intersection / Lookup Changes
Related:
https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/904

Cache lookup and token persistence strategy (adapted from iOS logic)
- ~Look up `access_tokens` by scopes provided by developers _only_~ (89fefb9222e410aba77afe85d574a9119025e135)
- ~Do not append default scopes (`openid`, `profile`, `offline_access`) when performing lookups~ (89fefb9222e410aba77afe85d574a9119025e135)
- ~Write tokens to cache as returned by STS with no post-processing~ (d8da04b02b153d7ff1e78d1839a49d4ffe91a197)
- ~Before writing a token to cache, remove other tokens with intersecting scopes since these are cumulative~ (89fefb9222e410aba77afe85d574a9119025e135)
- ~Do not allow developers to pass `openid`, `profile`, or `offline_access` in their own request.~ (https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/386)
- **If no scopes are returned in response at all, save tokens with _requested_ scopes****

** Pending clarification, as this looks to be an indistinguishable scenarios from _declined scopes_

iambmelt:
See #793

Related:
https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/904

This makes the following changes:
- Write tokens to cache using IdP responses (do not inject default scopes)
- Exclude default scopes from intersect criteria

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
@@ -39,6 +39,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import static com.microsoft.identity.common.internal.controllers.BaseController.DEFAULT_SCOPES;
+
 public abstract class AbstractAccountCredentialCache implements IAccountCredentialCache {
 
     private static final String TAG = AbstractAccountCredentialCache.class.getSimpleName();
@@ -185,10 +187,10 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
             if (mustMatchOnTarget) {
                 if (credential instanceof AccessTokenRecord) {
                     final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                    matches = matches && targetsIntersect(target, accessToken.getTarget());
+                    matches = matches && targetsIntersect(target, accessToken.getTarget(), true);
                 } else if (credential instanceof RefreshTokenRecord) {
                     final RefreshTokenRecord refreshToken = (RefreshTokenRecord) credential;
-                    matches = matches && targetsIntersect(target, refreshToken.getTarget());
+                    matches = matches && targetsIntersect(target, refreshToken.getTarget(), true);
                 } else {
                     Logger.warn(TAG, "Query specified target-match, but no target to match.");
                 }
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
```diff
@@ -210,13 +212,16 @@
...
+     * @param credentialTarget  The target against which our sought value will be compared.
+     * @param omitDefaultScopes True if MSAL's default scopes should be considered in this lookup.
+     *                          False otherwise.
      * @return True, if the credentialTarget contains all of the targets (scopes) declared by
      * targetToMatch. False otherwise.
      */
     static boolean targetsIntersect(@NonNull final String targetToMatch,
-                                    @NonNull final String credentialTarget) {
+                                    @NonNull final String credentialTarget,
+                                    boolean omitDefaultScopes) {
```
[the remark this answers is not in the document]
shahzaibj: I think it would be good to have in case if a child implementation of `AbstractAccountCredentialCache` wishes to override it in the future

common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -129,18 +130,27 @@ private String getCredentialType(@NonNull final String tokenType) {
...
+
+        if (TextUtils.isEmpty(responseScope)) {
+            // The response scopes were empty -- per https://tools.ietf.org/html/rfc6749#section-3.3
+            // we are going to fall back to a the request scopes minus any default scopes....
+            final String[] requestScopes = requestScope.split("\\s+");
+            final Set<String> requestScopeSet = new HashSet<>(Arrays.asList(requestScopes));
+            requestScopeSet.removeAll(DEFAULT_SCOPES);
+
+            for (final String scope : requestScopeSet) {
+                scopesToCache += scope + " ";
```
shahzaibj: I think it would be better to use a `StringBuilder` here as the size of the scope set isn't known and could be big

iambmelt: The perf of `+` and `StringBuilder` is fairly comparable due to compiler optimizations. Given that `StringBuilder` hurts readability IMO, I'm going to leave as is.

(Citation: [RedFin Engineering Blog](https://redfin.engineering/java-string-concatenation-which-way-is-best-8f590a7d22a8))
Capture

PR #795, opened 2020-01-29
Add spinner on first page load

rpdome:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -80,10 +84,18 @@ void extractState(final Bundle state){
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
         final View view = inflater.inflate(R.layout.common_activity_authentication, container, false);
+        mProgressBar = view.findViewById(R.id.common_auth_webview_progressbar);
 
         AzureActiveDirectoryWebViewClient webViewClient = new AzureActiveDirectoryWebViewClient(
                 getActivity(),
-                new AuthorizationCompletionCallback(), mRedirectUri);
+                new AuthorizationCompletionCallback(),
+                new OnPageLoadedCallback() {
```
iambmelt: As a thought, what if this callback _also_ listened for `onPageStarted()` so that the `WebView` determines both when to show _and_ dismiss the spinner? Currently there's a bit of action-at-a-distance where `onStart()` shows and the callback hides.

rpdome: I took that out because OAuth2WebViewClient.onPageStarted() is invoked every time a new page is loaded, and we don't want that because the user would see 2 progress bar at the same time (One rendered by the server, and the other by us).

iambmelt: Oh yuck. OK, thanks Dome

PR #797, opened 2020-01-31
Add js injection for test cases

rpdome:
MSAL CPP Team would like to inject JS as part of their test suites.
(I intentionally don't expose this in createStartIntent())

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1231,6 +1231,8 @@ public final class AuthenticationConstants {
 
         public static final String REQUEST_HEADERS = "com.microsoft.identity.request.headers";
 
+        public static final String JAVASCRIPT_INJECTION_STRING = "com.microsoft.identity.javascript.injection.string";
+
         public static final String AUTHORIZATION_AGENT = "com.microsoft.identity.client.authorization.agent";
 
         public static final String REQUEST_ID = "com.microsoft.identity.request.id";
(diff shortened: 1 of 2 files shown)


iambmelt:
This looks like something we need to think about / threat model before merging since this looks like something a bad actor could conceivably leverage to display phishing content, screen scrape or otherwise. 

I'm also a little apprehensive about the variable/field naming, as this looks to me generally a way to override the navigation URL rather than an [explicit API for injecting javascript into WebView](https://github.com/iambmelt/WebViewInjector) (shameless plug, sorry 😅)

iambmelt:
> This looks like something we need to think about / threat model before merging since this looks like something a bad actor could conceivably leverage to display phishing content, screen scrape or otherwise.
> 
> I'm also a little apprehensive about the variable/field naming, as this looks to me generally a way to override the navigation URL rather than an explicit API for injecting javascript into a WebView

@rpdome Can you support this _only_ if this is a `DEBUG` build?

rpdome:
> > This looks like something we need to think about / threat model before merging since this looks like something a bad actor could conceivably leverage to display phishing content, screen scrape or otherwise.
> > I'm also a little apprehensive about the variable/field naming, as this looks to me generally a way to override the navigation URL rather than an explicit API for injecting javascript into a WebView
> 
> @rpdome Can you support this _only_ if this is a `DEBUG` build?

I think that should be fine.

iambmelt:
> > > This looks like something we need to think about / threat model before merging since this looks like something a bad actor could conceivably leverage to display phishing content, screen scrape or otherwise.
> > > I'm also a little apprehensive about the variable/field naming, as this looks to me generally a way to override the navigation URL rather than an explicit API for injecting javascript into a WebView
> > 
> > 
> > @rpdome Can you support this _only_ if this is a `DEBUG` build?
> 
> EDIT: Just realized that this might not work for them as they'll be consuming the release build of common.

Synced offline. Since this uses the same mechanism as our standing `loadUrl` function (which has already been sec reviewed / modeled) these changes are likely OK. Will still review in standup

PR #806, opened 2020-02-05
Make Bearer the default scheme

rpdome:


iambmelt:
Leaving a comment for future reference: this change is so that callers do not have to explicitly set the `Bearer` scheme if one is not provided (say, as in broker).

iambmelt:
> Leaving a comment for future reference: this change is so that callers do not have to explicitly set the `Bearer` scheme if one is not provided (say, as in broker).

Just posting our offline thoughts here /cc @rpdome -- we don't want to inadvertantly downgrade the security posture of a request by falling on `Bearer` when we don't mean to. This is the API tradeoff between resiliency (if we forget to specify a scheme) and explicitness (validation throws if not provided).

We can discuss in scrum to determine which approach team feels is best

rpdome:
We decided to make this scheme explicit. Any crashes/errors should be caught in bug bash.

PR #809, opened 2020-02-05
Have WebView load once per fragment lifecycle, instead of repeatedly.

arislomovic:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -109,15 +109,7 @@ public void onPageLoaded() {
                 },
                 mRedirectUri);
         setUpWebView(view, webViewClient);
-
-        return view;
-    }
-
-    @Override
-    public void onStart() {
         final String methodName = "#onCreateView";
```
rpdome: nit: would you mind moving this to the top of this method?

arislomovic: Sure Will do.

PR #818, opened 2020-02-06
803 - WebView Zoom Toggle

arislomovic:


wmiranda:
@shoatman plz let us know if we can have it approved in time for the next release you were planning to wrap up today. Thanks!

rpdome:
@wmiranda This change would also requires a change in MSAL. We've pushed our code complete date to **tomorrow** for the MSAL PR.

PR #828, opened 2020-02-20
Fix msal crash on cancel interactive broadcast

shahzaibj:
Fix [MSAL Issue 940](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/940)
Fix [MSAL Issue 935](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/935)
Fix [MSAL Issue 920](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/920)
Fix [MSAL Issue 924](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/924)

MSAL crashes intermittently due to the CANCEL_INTERACTIVE_BROADCAST. This usually happens and is easy to reproduce when the developer starts another interactive call from the onError callback of the first one. This can result in a loop where each new request cancel the previous one. This is because cancelling the request by the SDK also results in an error and hits the onError again (_May be this should also hit `onCancel` instead of onError?_) and starts a chain reaction where the SDK keeps starting a new interactive request and keeps cancelling it.  After some time MSAL runs into a some weird threading state where it calls the `completeAuthorization` method on a null authorization strategy object. See [this file](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/d806fa789f29c77e9d1e190f4149c91af8923c53/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java#L217). I'm not too sure why and how it happens as it is not possible to reproduce it while debugging and seems to be a timing/threading issue to me.

I also discussed with @shoatman and he suggested that this entire broadcast wasn't even meant for MSAL but rather only for the Broker to avoid multiple apps making interactive calls against the broker at the same time which would result in the current one being blocked until the previous one finishes. I think this makes sense and hence I've removed the broadcast from the local MSAL case and made it specific to just the broker.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -43,6 +43,7 @@ import com.microsoft.identity.common.internal.logging.DiagnosticContext;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.request.AcquireTokenOperationParameters;
 import com.microsoft.identity.common.internal.request.AcquireTokenSilentOperationParameters;
+import com.microsoft.identity.common.internal.request.BrokerAcquireTokenOperationParameters;
 import com.microsoft.identity.common.internal.result.AcquireTokenResult;
 import com.microsoft.identity.common.internal.result.ILocalAuthenticationResult;
 import com.microsoft.identity.common.internal.telemetry.Telemetry;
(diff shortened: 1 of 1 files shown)


kreedula:
Thanks @shahzaibj for the fix. Can we add a description of the root cause and how the bug is effecting only local MSAL for reference ?

iambmelt:
> Thanks @shahzaibj for the fix. Can we add a description of the root cause and how the bug is effecting only local MSAL for reference ?

+1 Not obvious what's going wrong

shahzaibj:
@iambmelt @kreedula Description updated

wmiranda:
Hi @shahzaibj !
When can we expect to have at lease a release branch with this fix?
We would like to test it asap as we have tomorrow as the cut date for our project.
Thank you!

shahzaibj:
@wmiranda The latest release of MSAL with this fix is available on Maven Central. The release version is `1.4.0`

PR #838, opened 2020-03-06
Fix fragment state issue

shahzaibj:
If no state available we need to finish and return. This is what we used to do, but it was regressed with Fragment support.

Repro (There could be other ways to repro this too, but the fix will take care of all repros):

- Start an interactive request in MSAL
- Quickly lock the device and unlock again (before the browser even opens)

This results in a NPE when extractState is called. I wasn't able to be as quick as required in repro but slowed down code execution in debug mode by putting breakpoints and hence reproduced it.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
@@ -94,6 +94,12 @@ public class BrowserAuthorizationFragment extends AuthorizationFragment {
 
     @Override
     void extractState(final Bundle state){
+        if (state == null) {
+            Logger.warn(TAG, "No stored state. Unable to handle response");
+            finish();
+            return;
+        }
+
         super.extractState(state);
         mAuthIntent = state.getParcelable(AUTH_INTENT);
         mBrowserFlowStarted = state.getBoolean(BROWSER_FLOW_STARTED, false);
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
```diff
@@ -94,6 +94,12 @@ public void onSaveInstanceState(@NonNull Bundle outState) {
 
     @Override
     void extractState(final Bundle state){
+        if (state == null) {
```
rpdome: Should we do this in the base class?

shahzaibj: It is already done there but is not sufficient as that method returns back here

shahzaibj: super is called first, which does this logic and returns. Then we are back here, try to use the state param, it is NULL and we get NPE

rpdome: ahh okay.

In the base class, can we move the logic to onCreate instead? (before extractState is called).

Thanks for catching the bug btw.

shahzaibj: Why do you want to finish in the onCreate? The state param in onCreate could be NULL and is currently handled by the `mInstanceState` instance variable. See this: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/744fb75a6d443aef8ee11c88a22886128cd4a9f9/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L84-L101

rpdome: We could make extractState() takes NonNull parameter only, and then put the check inside onCreate(). We're already doing half the check there (if savedInstanceState==null) anyway.

I just prefer having the check in one place - it's easier to maintain.

shahzaibj: Addressed

PR #839, opened 2020-03-06
Fix crash on redirect if no activity available

shahzaibj:
Pls see related MSAL PR:  https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/966

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
@@ -34,7 +34,7 @@ import static com.microsoft.identity.common.internal.providers.oauth2.Authorizat
 
 /**
  * Authorization fragment with customTabs or browsers.
- * */
+ */
 public class BrowserAuthorizationFragment extends AuthorizationFragment {
 
     private static final String TAG = BrowserAuthorizationFragment.class.getSimpleName();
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
```diff
@@ -69,8 +69,15 @@
      * @param context     the package context for the app.
      * @param responseUri the response URI, which carries the parameters describing the response.
      */
+    @Nullable
     public static Intent createCustomTabResponseIntent(final Context context,
                                                        final String responseUri) {
+        if (sCallingActivityClass == null) {
+            // can't create intent for response if no activity available
+            Logger.warn(TAG, "Calling activity class is NULL. Unable to create intent for response.");
```
shoatman: Shahazib - Would you mind adding a comment here about the result of the authorization request will be.  I'm assuming the only this can happen is if the user closes the app or if the OS closes the app due to memory pressure.  I guess I want to know returning null here results in.  I'm assuming no callback at all... since we no longer have something to call back to.

shahzaibj: Yeah this would happen if the app got closed either by user or OS. This method is called in BrowserTabActivity. See this PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/966/files

I catch the null, log a warning, show a toast, and finish

There won't be any result, no tokens etc. as there is no callback and no interactive request in memory

shahzaibj: I've added the comment

PR #843, opened 2020-03-11
Fix state issue - port to dev of #838

shahzaibj:
Port #838 to dev


shahzaibj:
Going to create another intermediate branch for this

PR #846, opened 2020-03-12
Use default scopes for foci request

shahzaibj:


common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -156,7 +156,7 @@ public static boolean tryFociTokenWithGivenClientId(@NonNull final BrokerOAuth2T
         final MicrosoftStsOAuth2Strategy strategy = new MicrosoftStsOAuth2Strategy(config, strategyParameters);
 
         final String refreshToken = cacheRecord.getRefreshToken().getSecret();
-        final String scopes = cacheRecord.getRefreshToken().getTarget();
+        final String scopes = DEFAULT_SCOPES;
```
iambmelt: There's a minor refactor/code-sharing opportunity we should take here.... Looks like we have defined two constants for this, would be great not to trip over this in the future because there are two fields.....

The first is here (which you've used).

The second is in [`BaseController`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java#L88-L94) -- can we make sure all the same constants are used everywhere in the code so if we change which scopes comprise the defaults we only have to do it in one place?

I would probably add a `static` method to the `BaseController` to return the contents of the list with a `" "` delimiter.

Seem OK?

shahzaibj: Yeah, sure

shahzaibj: @iambmelt Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/846/commits/df82cb7678e8f1113b6bd014dfc1f8c14080a113

iambmelt:
LGTM, thanks @shahzaibj!

:shipit:

PR #853, opened 2020-03-16
Implement server telemetry v2

shahzaibj:
Please refer to server side telemetry v2 schema doc for detailed guidance on the schema changes. This document was shared with the team via email.

Changes done in this PR: 

- Implement V2 schema (refer to the doc or see summary below)
- Use GSON to get cache value for telemetry objects saved to cache
- Add a field called as `servicedFromCache` to `LocalAuthenticationResult`, this indicates whether token was returned from cache lookup. The field was also added to BrokerResult
- Refactor the beginInteractive in CommandDisptacher to use the CommandResult object to return the result back to caller (submitSilent was already using this pattern)
- Add a method called `isEligibleForEstsTelemetry` to the command interface that indicates whether we would capture telemetry for this command.
- Remove the unit tests for telemetry in favor of E2E test that was added in the MSAL pr

Here's a quick summary for the changes in v2 schema:

Schema:

x-client-current-telemetry = `schema_version | api_id, force_refresh | platform_fields`
x-client-last-telemetry = `schema_version | silent_succesful_count | failed_requests | errors | platform_fields`

Failed request and error fields are arrays.

Example: `schema_version | silent_succesful_count | api_id_1, correlation_id_1, api_id_2, correlation_id_2, api_id_3, correlation_id_3 | error_1, error_2, error_3 | platform_fields`

Some other important changes:

- Telemetry headers will only be sent to /token endpoint
- Telemetry data won't be removed if not logged by sts even if sent to token endpoint by client (This includes status codes such 429 and/or status codes greater than or equal to 500)
- There is a 8KB limit for payload in request headers sent to sts, we will be capping at 4KB. We will be performing validation prior to sending header to ensure we stay within those bounds. If we have more data than that, then we keep it in the cache and pick it up during the next request to token endpoint.


common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -321,17 +363,63 @@ String getCurrentTelemetryHeaderString() {
...
+        // the array index that marks how many elements were included in the header sent in the request
+        int indexOfFailedRequestsAndErrors = 0;
+
+        for (int i = 0; i < originalFailedRequests.size(); i++) {
+            // there is a limit of 8KB for the payload sent in request headers
+            // we will be maxing out at 4KB to avoid HTTP 413 errors
+            // check if we have enough space in the String to store another failed request/error element
+            // if yes, then add it to the failed request array (for the copy)
+            if (lastRequestTelemetryCopy.getCompleteHeaderString().length() < 3800) {
+                lastRequestTelemetryCopy.appendFailedRequestWithError(
```
kreedula: nit: constant for the data limit value

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -321,17 +363,63 @@ String getCurrentTelemetryHeaderString() {
...
+                        originalFailedRequests.get(indexOfFailedRequestsAndErrors),
+                        originalErrors.get(indexOfFailedRequestsAndErrors++)
+                );
+            } else {
+                // if there is no room for more data, then break out of this loop
+                break;
```
kreedula: Should we indicate the server somehow  that we are not sending the data due to size restrictions. i.e. just for us to be aware that this data could be skewed?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -321,17 +363,63 @@ String getCurrentTelemetryHeaderString() {
...
-        RequestTelemetry lastTelemetry = loadLastRequestTelemetryFromCache();
-
-        if (lastTelemetry == null) {
+    private String getLastTelemetryHeaderString() {
+        if (mLastRequestTelemetryCache == null) {
             return null;
         }
 
-        return lastTelemetry.getCompleteTelemetryHeaderString();
+        LastRequestTelemetry lastRequestTelemetry = (LastRequestTelemetry) mLastRequestTelemetryCache.getRequestTelemetryFromCache();
```
kreedula: nit : `final` here and everywhere else in the method for local variables

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -81,6 +86,27 @@ public void setupLastRequestTelemetryCache(@NonNull final Context context) {
...
+    public void initTelemetryForCommand(@NonNull final BaseCommand command) {
+        setupLastRequestTelemetryCache(command.getParameters().getAppContext());
+        final String correlationId = command.getParameters().getCorrelationId();
+        if (command.isEligibleForEstsTelemetry()) {
+            final CurrentRequestTelemetry currentRequestTelemetry = new CurrentRequestTelemetry();
+            mTelemetryMap.put(correlationId, currentRequestTelemetry);
+        }
+    }
+
+    private boolean isDisabled() {
```
kreedula: When would we disable this telemetry? If this method is only to check if we have telemetry or not, we might to have rename this method more appropriately

shahzaibj: It is disabled in the sense disabled for a command such as the GetDeviceModeCommand that returns false for `isEligibleForEstsTelemetry`, but I agree this can be renamed to something better. Perhaps `isCurrentTelemetryAvailable` or `isTelemetryCaptured`?

rpdome:
suggestion: list the changes that are done in this PR. It doesn't have to be super detailed, but one level below the headline ("implement server telemetry") would really help.

i.e. I didn't know what 'servicedFromCache' was added as part of this PR in the first glance.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
kreedula: Yup, `isCurrentTelemetryAvailable` seems better to me

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: Such a field is not currently part of the schema. I can add something to the platform telemetry for Android. Something like a 0 or 1 value indicating whether all the data is sent or not. So when we view our own telemetry then we know whether we have sent all the data or not. Or we could actually add a number indicating how many errors were either sent or kept on hold. I think the former is better i.e. it is probably sufficient to know whether all was sent or not and tell us to look at additional rows in the PerRequestTable. I don't see much value from knowing exactly how much was sent or not. What do you think @kreedula ?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
rpdome: I'd suggest using 'eligible' instead of 'avaiable' - so that we know it came from isEligibleForEstsTelemetry

shahzaibj:
@rpdome I've updated the description

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -106,33 +132,28 @@ public void emit(@Nullable final Map<String, String> telemetry) {
...
     public void emit(final String key, final String value) {
         final String correlationId = DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID);
-        emit(correlationId, key, value);
+        final String compliantValueString = TelemetryUtils.getSchemaCompliantString(value);
+        emit(correlationId, key, compliantValueString);
     }
 
     private void emit(final String correlationId, final String key, final String value) {
-        RequestTelemetry currentTelemetryInstance = getCurrentTelemetryInstance(correlationId);
+        final CurrentRequestTelemetry currentTelemetryInstance = getCurrentTelemetryInstance(correlationId);
```
rpdome: nit: is this private function being used elsewhere?

shahzaibj: No it is not being elsewhere

shahzaibj: I can remove it and move the logic to the two param method

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: That may not be necessarily true. If we refactor the code in the future and introduce a bug then we may not have telemetry due to a bug and not simply because a command wasn't eligible.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +215,143 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
-        lastTelemetry.putTelemetry(Schema.Key.CORRELATION_ID, correlationId);
-        lastTelemetry.putTelemetry(Schema.Key.ERROR_CODE, errorCode);
+        // load the last request object from cache
+        LastRequestTelemetry lastRequestTelemetry = loadLastRequestTelemetryFromCache();
+
+        // We did not have a last request object in cache, let's create a new one and derive
+        // fields from current request where applicable
+        if (lastRequestTelemetry == null) {
+            lastRequestTelemetry = new LastRequestTelemetry(currentTelemetry.getSchemaVersion());
+            lastRequestTelemetry = (LastRequestTelemetry) lastRequestTelemetry.derive(currentTelemetry);
```
rpdome: Why are we duplicating the data?

shahzaibj: duplicating?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +215,143 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
-     * Removes the telemetry associated to the correlation id from the telemetry map,
-     * and saves it to the cache (SharedPreferences) as the last request telemetry.
-     *
-     * @param correlationId correlation id of the request
-     * @param baseException exception that may have occurred during the request
-     */
-    public void flush(final String correlationId, final BaseException baseException) {
-        flush(correlationId, baseException == null ? null : baseException.getErrorCode());
-    }
+    public void flush(@NonNull final BaseCommand command, @NonNull final CommandResult commandResult) {
```
rpdome: is this thread safe? (Since we support parallel request)

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: The data there needs to be on both current request and last request telemetry objects. `derive` is just a helper method to grab all the common platform fields easily without explicitly calling emit for each one.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: Yes, we are using concurrent data structures.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
kreedula: I think sending a boolean (0 or 1) to indicate if the data is partial or complete would be helpful .My primary concern raising this is , potentially in future we might end up sending partial data for every request and keep on accumulating old data , thus latest data could be lagging. Atleast having this information will help us to investigate and we probably can increase the payload if needed.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
rpdome: What I meant was... is it technically possible for any two concurrent requests to send the same set of 'failed_requests' or 'errors'. If so, would the server side be able to deal with it.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
rpdome: It was sort of confusing to me to have the same telemetry object in both fields (current and last).

I would expect us to only send 'previous failing requests' in this 'Last request' fields. Isn't that the case?

(If this particular request is not sent and it failed, only then we added it to the cache, and the next request would pick it up and send it to the server).

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: Yes, it is possible, and the server won't do anything. It means that we would have duplicated data there.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +215,143 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
+            // telemetry headers have been sent to token endpoint and logger by sts
+            // this is the time to reset local telemetry state
+
+            // reset silent successful count as we just went to token endpoint
+            lastRequestTelemetry.resetSilentSuccessCount();
+
+            // get the index for the last element of the failed request/error array so we can wipe
+            // all array elements up to this point (as that data is already in PerRequestTable)
+            // and doesn't need to be held onto or sent again
+            final int loggedIndex = getLoggedTelemetryIndex();
```
rpdome: I think it might be risky using index as we're supporting concurrent requests.

What if we have the command store a list of last request correlation ID, and if sent, then we remove the entry with the associated correlation ID from the cache?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: Yes, that's right, we only send previous failing requests/errors. However, we also have to abide by the schema and send all fields/delimiters regardless of whether they have value or not. So it there are no last request telemetry available in the cache, then we need to send something like this:

`2|0|||`

i.e. We still send the schema version all delimiters

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: It is basically separation of concerns. I don't fully understand what you're suggesting but right now the data is stored in the SharedPreferences and persists. Are you suggesting to store in the Command? Commands are temporary so once those are over we would lose information.

rpdome: Let's say if we have more than 1 command running at the same time, and then they get the same value from getLoggedTelemetryIndex() and then try to wipe at the same time, would index out of bound occur?

I was suggesting storing the 'key' of failed request/error cache entry inside command, and then use those keys to wipe the values out. This would make it thread-safe. (the first wipe will win, the later wipes will not do anything).

rpdome: or there might be another way to address this. My concern is primarily the OutOfBound exception.

shahzaibj: OutOfBound should not occur as `wipe` is an instance method so even though these are concurrent requests, they each should have a different instance of the Last Request Telemetry object and hence when wiped is called, then it is called on the instance that is being used during the execution of that flush method. Inside wipe, we have a check like this for index:

`        if (index < 0 || index > failedRequests.size() || index > errors.size()) {
            return;
        }`

I think this should ensure that we're safe. What do you think?

rpdome: ahh.. ok. so you wipe from each instance then proceed to save those instances to the cache, and in this case the last one wins.

Ok in this case we should be safe from OOB. I'm still in the middle of PR and thought that you saved those request/error entries separately.

rpdome: The potential error that could happen in this case is a missing telemetry (i.e. There are 5 concurrent requests, the 3rd request failed, but the 5th overwrites the lastRequestTelemetry object).

rpdome: Having each command keeping track of what 'last request/error' it sent might be more straightforward IMHO. In this case, we can be certain we only wipe the ones that are already sent.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +215,143 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
+
+    private boolean isTelemetryLoggedByServer(@NonNull final BaseCommand command, @NonNull final CommandResult commandResult) {
+        // This was a local operation - we didn't reach token endpoint and hence telemetry wasn't sent
+        if (!(command instanceof TokenCommand)) {
+            return false;
+        }
+
+        if (commandResult.getStatus() == CommandResult.ResultStatus.ERROR) {
+            BaseException baseException = (BaseException) commandResult.getResult();
+            if (!(baseException instanceof ServiceException)) {
```
rpdome: How do we know if this is a service exception from token endpoint (not auth endpoint)?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +215,143 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
+        }
+
+        // if we get here that means we went to token endpoint and headers were logged by sts
+        return true;
     }
 
-    String getCurrentTelemetryHeaderString() {
+    private int getLoggedTelemetryIndex() {
+        return ((SharedPreferencesLastRequestTelemetryCache) mLastRequestTelemetryCache)
+                .getLastTelemetryIndexSentInHeaderFromCache();
```
rpdome: Shouldn't this value be part of the lastRequestTelemetry object (like the associated failedRequest/error arrays).

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -321,17 +363,63 @@ String getCurrentTelemetryHeaderString() {
...
+                );
+            } else {
+                // if there is no room for more data, then break out of this loop
+                break;
+            }
+        }
+
+        // tell cache where we stopped adding data to header so we can pick this up in the next
+        // network request
+        ((SharedPreferencesLastRequestTelemetryCache) mLastRequestTelemetryCache)
```
rpdome: I suggest keeping a separate copy of this inside the last request object (as same as the arrays).

Every telemetry values should be modified in only one place - flush().

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -321,17 +363,63 @@ String getCurrentTelemetryHeaderString() {
...
+        LastRequestTelemetry lastRequestTelemetry = (LastRequestTelemetry) mLastRequestTelemetryCache.getRequestTelemetryFromCache();
+
+        if (lastRequestTelemetry == null) {
+            // we did not have anything in the telemetry cache for the last request
+            // let's create a new object based on the data available from the current request object
+            // and return the header string formed via that object
+            final String correlationId = DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID);
+            final CurrentRequestTelemetry currentRequestTelemetry = mTelemetryMap.get(correlationId);
+            lastRequestTelemetry = new LastRequestTelemetry(currentRequestTelemetry.getSchemaVersion());
+            lastRequestTelemetry = (LastRequestTelemetry) lastRequestTelemetry.derive(currentRequestTelemetry);
```
rpdome: nit: the term 'derive', along with the comment above, gave me an impression that we're actually copying everything from CurrentRequestTelemetry. 

Should we rename this to something along the line of '_copySharedValues_'?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
```diff
@@ -36,17 +36,31 @@
...
      * @return a {@link RequestTelemetry} object
      */
     RequestTelemetry getRequestTelemetryFromCache();
 
+    /**
+     * Get the schema version of this telemetry object from cache
+     *
+     * @return schema version for telemetry object
+     */
+    String getTelemetrySchemaVersionFromCache();
```
rpdome: Is this being used?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
```diff
@@ -36,17 +36,31 @@
...
+
+    /**
+     * Get the header string for telemetry object from cache
+     *
+     * @return header string for telemetry
+     */
+    String getTelemetryHeaderStringFromCache();
```
rpdome: Is this being used?

common/src/main/java/com/microsoft/identity/common/internal/result/LocalAuthenticationResult.java
```diff
@@ -200,4 +203,14 @@ public void setRefreshTokenAge(final String refreshTokenAge) {
     public AccessTokenRecord getAccessTokenRecord() {
         return mAccessTokenRecord;
     }
+
+
+    public void setServicedFromCache(Boolean servicedFromCache) {
```
rpdome: nit: add comment, here and every new fields introduced in this PR.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
shahzaibj: No, I added them initially because I planned to use them but didn't. We may need these when we implement future schema versions, but I can remove it for now.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
shahzaibj: No, I added them initially because I planned to use them but didn't. We may need these when we implement future schema versions, but I can remove it for now.

common/src/main/java/com/microsoft/identity/common/internal/result/LocalAuthenticationResult.java
shahzaibj: Seems like this method is actually not being used - so I'll just remove it

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: No, this value isn't actually part of telemetry i.e. it isn't actually a field in the schema and hence I didn't add it LastRequest Telemetry class.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
```diff
@@ -0,0 +1,118 @@
...
+        return failedRequests;
+    }
+
+    List<String> getErrors() {
+        return errors;
+    }
+
+    @Override
+    public String getHeaderStringForFields() {
+        return silentSuccessfulCount + "|" + getHeaderStringForFields(failedRequests) + "|" + getHeaderStringForFields(errors);
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/RequestTelemetry.java
```diff
@@ -112,34 +82,34 @@ String getCompleteTelemetryHeaderString() {
...
-        final String schemaVersionString = Schema.getSchemaCompliantString(mSchemaVersion);
-        final String commonSchemaString = getCommonTelemetryHeaderString();
-        final String platformSchemaString = getPlatformTelemetryHeaderString();
-        return schemaVersionString + "|" + commonSchemaString + "|" + platformSchemaString;
-    }
-
-    private String getCommonTelemetryHeaderString() {
-        final String[] commonFields = Schema.getCommonFields(mIsCurrentRequest);
-        return getTelemetryHeaderStringFromFields(commonFields, mCommonTelemetry);
+        return mSchemaVersion + "|" + this.getHeaderStringForFields() + "|" + getPlatformTelemetryHeaderString();
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/IRequestTelemetryCache.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/result/LocalAuthenticationResult.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: @rpdome I've taken your suggestion and refactored the code. I didn't do at the command though as I believe telemetry should be kept separate from commands as much as possible due to separation of concerns. Also the currently executing command isn't available from everywhere. We would have to put the entire command object on the `DiagnosticContext` if we are to achieve that. I want to avoid doing that. So to solve this problem I've create a new `Set` at the telemetry layer that keeps track of all the failed requests sent to the server for the currently executing request and then later when we flush, we just remove all of these from the cache as these are sent. Two `FailedRequest` objects are considered the same if they have the same `correlationId` so I'm using that as you suggested to keep track of this. The index is no longer used and I've removed that. Thanks for the suggestion!

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
shahzaibj: This field has been removed now per your suggestion.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/EstsTelemetry.java
```diff
@@ -194,122 +221,139 @@ private IRequestTelemetryCache createLastRequestTelemetryCache(@NonNull final Co
...
+
+        // we're done processing telemetry for this command, let's remove it from the map
         mTelemetryMap.remove(correlationId);
+        mSentFailedRequests.remove(correlationId);
 
-        if (mLastRequestTelemetryCache == null) {
+        if (mLastRequestTelemetryCache != null) {
+            // save the (updated) telemetry object back to telemetry cache
+            mLastRequestTelemetryCache.saveRequestTelemetryToCache(lastRequestTelemetry);
+        } else {
```
rpdome: How would this work for parallel commands?

Each of them will have their own copies,  how are we going to resolve the conflict? As of now I believe the last one wins (which would result in missing telemetries, if any of the commands failed).

shahzaibj: If it happens at exactly the same time then yes some data will get overriden and some telemetry may go missing. Do you think I should make this method `synchronized`?

rpdome: If we want to make it static, then we have to make flush() static - one has to save before the other loads.

alternatively, we could save the 'changes' instead of the whole request telemetry object. (i.e. delete A, add B).

shahzaibj: Why static?

shahzaibj: I was thinking about making both `flush` and `getTelemetryHeaders` synchorized. What do you think about that?

rpdome: That would be simpler to implement, but probably be a bottleneck if there are multiple requests.... so.. tradeoffs here.

I personally prefer #2, but if you think it's too complicated and not worth the performance gain, then I'm also fine with #1 as well. I'll let you make the call.

shahzaibj: Yeah I think 2 would be complicated to maintain and not worth the effort/performance gain. 

I went ahead and using the 1st approach and addressed it here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/853/commits/c05a8102c3c4218981eab9abc9ec93d8b4131e79

PR #854, opened 2020-03-17
Add dual screen support in common

rpdome:
Common's AuthorizationActivity (used by MSAL and Broker) is now able to detect dual screen and rearrange its UI accordingly.

DualScreenUtil will also be reused in ADAL, so I figured I should get this in first.

Let me know if you have any question. I can also send screenshots via Teams.

NOTE: we're still using public preview version for com.microsoft.device.display.displaymask.jar.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -75,6 +102,8 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
                 .setTransitionStyle(FragmentTransaction.TRANSIT_FRAGMENT_FADE)
                 .replace(R.id.authorization_activity_content, mFragment)
                 .commit();
+
+        adjustLayout();
```
iambmelt: This seems to early to detect rotation. What if the `Activity` is created, user leaves, rotates device, and returns?

Do we need to move this to `onResume()`?

iambmelt: Oh, I see there's also a call in `onConfigurationChanged()`

I guess this is fine?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -83,4 +112,49 @@ public void onBackPressed() {
...
         }
     }
+
+    @Override
+    public void onConfigurationChanged(Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+        adjustLayout();
+    }
+
+    private void adjustLayout(){
```
iambmelt: Can these constraints be defined declaratively in the XML?

Something like:
```java
if (appIsSpannned) {
    setContentView(R.layout.spanned_authorization_activity);
} else {
    setContentView(R.layout.authorization_activity);
}
```

rpdome: Yes, but we'll need three files for each dual screen activities.
1. Not spanned.
2. Spanned - horizontal.
3. Spanned - vertical.

NOTE: after getting this feedback, I realized that the current solution can be made more reusable. Therefore, I made a change to do the following.

1. Created a 'dual_screen_layout.xml'. 
   - Internally, this contains the 'content' area, and the 'blank' area.

2. Moved adjustLayout() to DualScreenUtil.
  - This function now takes an activity that uses 'dual_screen_layout.xml'.
  - Based on the screen state, it will adjust the 'content' and the 'blank' area appropriately.

3. Any page that needs to support duo will have to do the following
 - call setContentView(R.layout.dual_screen_layout);
 - attach its 'content to be displayed' to the 'content' area of dual_screen_layout.xml.
 - call DualScreenUtil.adjustLayoutForDualScreenActivity().

This means I'll be able to reuse this with AuthorizationActivity (Broker/MSAL),  AuthenticationActivity (ADAL), and any other Broker pages that might have to support Duo in the future.

iambmelt: @rpdome Overkill to define an `Activity` subclass that overrides `setContentView()` for this to support dual-screen?

Just riffing ideas

rpdome: That might make it cleaner. Let me try that.

rpdome: @iambmelt  done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
```diff
@@ -0,0 +1,177 @@
...
+            // The windowRect doesn't intersect hinge
+            return hinge.intersect(windowRect);
+        }
+
+        return false;
+    }
+
+    public int getRotation(Activity activity) {
+        WindowManager wm = (WindowManager) activity.getSystemService(Context.WINDOW_SERVICE);
+        int rotation = 0;
```
iambmelt: I don't see the `int` result defined here. What does `0` mean? Can we javadoc?

common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenUtil.java
```diff
@@ -0,0 +1,145 @@
...
+        final ConstraintSet constraintSet = new ConstraintSet();
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.LEFT, R.id.dual_screen_layout, ConstraintSet.LEFT, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.RIGHT, R.id.dual_screen_layout, ConstraintSet.RIGHT, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.TOP, R.id.dual_screen_layout, ConstraintSet.TOP, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.BOTTOM, R.id.dual_screen_layout, ConstraintSet.BOTTOM, 0);
+
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.LEFT, R.id.dual_screen_layout, ConstraintSet.LEFT, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.RIGHT, R.id.dual_screen_layout, ConstraintSet.RIGHT, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.TOP, R.id.dual_screen_layout, ConstraintSet.TOP, 0);
+        constraintSet.connect(R.id.dual_screen_content, ConstraintSet.BOTTOM, R.id.dual_screen_layout, ConstraintSet.BOTTOM, 0);
```
kreedula: these looks exactly the same as the above 4 statements, is this intended?

rpdome: nice catch. 😆

common/build.gradle
```diff
@@ -77,10 +77,13 @@ android {
 }
 
 dependencies {
+    implementation files('libs/com.microsoft.device.display.displaymask.jar')
```
shahzaibj: What is this?

PR #865, opened 2020-03-20
Detect MDM link

rpdome:
Server will attach a query string (ismdmurl=1) along with the browser link to tell us that this link belongs to Intune (MDM) flow.

WebView will detect that query string and return AuthenticationConstants.UIResponse.BROWSER_CODE_MDM (instead of BROWSER_CODE_CANCEL)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -176,6 +176,11 @@ public final class AuthenticationConstants {
          * launched a new interactive auth request
          */
         public static final int BROWSER_CODE_SDK_CANCEL = 2008;
+
+        /**
+         * MDM Flow is triggered.
+         */
+        public static final int BROWSER_CODE_MDM = 2009;
     }
 
     /**
@@ -1016,14 +1021,14 @@ public final class AuthenticationConstants {
         public static final String BROWSER_EXT_INSTALL_PREFIX = "msauth://";
 
         /**
-         * Prefix in the redirect to open external browser to finish the CA auth.
+         * A query param indicating that this is an intune device CA link.
          */
-        public static final String BROWSER_DEVICE_CA_URL = "browser://go.microsoft.com/fwlink/?LinkId=396941";
+        public static final String BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "&ismdmurl=1";
 
         /**
          * Activity name to launch company portal.
          */
-        public static final String COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME = "com.microsoft.windowsintune.companyportal.views.SplashActivity";
+        public static final String COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME + ".views.SplashActivity";
 
         /**
          * Redirect URI parameter key to get link to install broker
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
@@ -98,15 +98,15 @@ public class AzureActiveDirectoryAuthorizationResultFactory extends Authorizatio
                 Logger.verbose(TAG, "Device needs to have broker installed, we expect the apps to call us"
                         + "back when the broker is installed");
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
-                        MicrosoftAuthorizationErrorResponse.AUTHORIZATION_FAILED,
-                        MicrosoftAuthorizationErrorResponse.BROKER_NEEDS_TO_BE_INSTALLED);
+                        MicrosoftAuthorizationErrorResponse.BROKER_NEEDS_TO_BE_INSTALLED,
+                        MicrosoftAuthorizationErrorResponse.BROKER_NEEDS_TO_BE_INSTALLED_ERROR_DESCRIPTION);
                 break;
 
             case AuthenticationConstants.UIResponse.BROWSER_CODE_DEVICE_REGISTER:
                 Logger.info(TAG, "Device Registration needed, need to start WPJ");
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
                         MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED,
-                        MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED);
+                        MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED_ERROR_DESCRIPTION);
                 // Set username returned from the service
                 result.getAuthorizationErrorResponse().setUserName(data.getStringExtra(
                         AuthenticationConstants.Broker.INSTALL_UPN_KEY)
@@ -117,7 +117,7 @@ public class AzureActiveDirectoryAuthorizationResultFactory extends Authorizatio
                 Logger.info(TAG, "MDM required. Launching Intune MDM link on browser.");
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
                         MicrosoftAuthorizationErrorResponse.DEVICE_NEEDS_TO_BE_MANAGED,
-                        MicrosoftAuthorizationErrorResponse.DEVICE_NEEDS_TO_BE_MANAGED);
+                        MicrosoftAuthorizationErrorResponse.DEVICE_NEEDS_TO_BE_MANAGED_ERROR_DESCRIPTION);
                 break;
 
             default:
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
@@ -103,11 +103,21 @@ public class AzureActiveDirectoryAuthorizationResultFactory extends Authorizatio
                 break;
 
             case AuthenticationConstants.UIResponse.BROWSER_CODE_DEVICE_REGISTER:
-                Logger.verbose(TAG, "Device needs to have broker installed, we expect the apps to call us"
-                        + "back when the broker is installed");
+                Logger.info(TAG, "Device Registration needed, need to start WPJ");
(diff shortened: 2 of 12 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1016,14 +1021,14 @@ private AuthenticationConstants() {
         public static final String BROWSER_EXT_INSTALL_PREFIX = "msauth://";
 
         /**
-         * Prefix in the redirect to open external browser to finish the CA auth.
+         * A query param indicating that this is an intune device CA link.
          */
-        public static final String BROWSER_DEVICE_CA_URL = "browser://go.microsoft.com/fwlink/?LinkId=396941";
+        public static final String BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "&ismdmurl=1";
```
Om83: Looks we already had the constant defined using the forward link instead of the query string and were using it to launch CP. So server change was not a must but good to have.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
```diff
@@ -103,11 +103,21 @@ public AzureActiveDirectoryAuthorizationResult createAuthorizationResult(final i
                 break;
 
             case AuthenticationConstants.UIResponse.BROWSER_CODE_DEVICE_REGISTER:
-                Logger.verbose(TAG, "Device needs to have broker installed, we expect the apps to call us"
-                        + "back when the broker is installed");
+                Logger.info(TAG, "Device Registration needed, need to start WPJ");
                 result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
-                        MicrosoftAuthorizationErrorResponse.AUTHORIZATION_FAILED,
-                        MicrosoftAuthorizationErrorResponse.BROKER_NEEDS_TO_BE_INSTALLED);
+                        MicrosoftAuthorizationErrorResponse.DEVICE_REGISTRATION_NEEDED,
```
Om83: Why are we passing the same parameter twice?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
```diff
@@ -103,11 +103,21 @@ public AzureActiveDirectoryAuthorizationResult createAuthorizationResult(final i
...
+                // Set username returned from the service
+                result.getAuthorizationErrorResponse().setUserName(data.getStringExtra(
+                        AuthenticationConstants.Broker.INSTALL_UPN_KEY)
+                );
+                break;
+
+            case AuthenticationConstants.UIResponse.BROWSER_CODE_MDM:
+                Logger.info(TAG, "MDM required. Launching Intune MDM link on browser.");
+                result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
+                        MicrosoftAuthorizationErrorResponse.DEVICE_NEEDS_TO_BE_MANAGED,
```
Om83: Same here

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResultFactory.java
```diff
@@ -123,7 +123,12 @@ public MicrosoftStsAuthorizationResult createAuthorizationResult(final int resul
                         );
                 break;
 
-
+            case AuthenticationConstants.UIResponse.BROWSER_CODE_MDM:
+                Logger.info(TAG, "MDM required. Launching Intune MDM link on browser.");
+                result = createAuthorizationResultWithErrorResponse(AuthorizationStatus.FAIL,
+                        MicrosoftAuthorizationErrorResponse.DEVICE_NEEDS_TO_BE_MANAGED,
```
Om83: And here

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -214,35 +214,40 @@ protected boolean processRedirectUrl(@NonNull final WebView view, @NonNull final
...
+                    return true;
+                } catch (final Exception ex) {
+                    Logger.warn(TAG + methodName, "Failed to launch Company Portal, falling back to browser.");
+                }
             }
-        } else {
-            openLinkInBrowser(url);
+
+            // Otherwise, strip out BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER and launch in Browser.
+            openLinkInBrowser(url.replace(AuthenticationConstants.Broker.BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER, ""));
```
Om83: Why do we need this. Shouldn't the server already be rejecting it?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
rpdome: I actually prefer not hardcoding this - as the link defined by Intune could be changed on the server side at any point in time.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
rpdome: The pattern I'm seeing is that the error code itself (MicrosoftAuthorizationErrorResponse.XXXX) can also be used as a error message.

We could also have separate values for this one.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
rpdome: This is what I and Carlos originally agreed on. Although I agree with you that if we make the query string parameter unique enough, we might not have to do this.

I've forwarded him this PR and will let him chime in.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -214,35 +214,40 @@ protected boolean processRedirectUrl(@NonNull final WebView view, @NonNull final
 
     private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull final String url) {
         final String methodName = "#processWebsiteRequest";
-        final PackageHelper packageHelper = new PackageHelper(getActivity().getPackageManager());
-        final Context applicationContext = getActivity().getApplicationContext();
-        if (url.startsWith(AuthenticationConstants.Broker.BROWSER_DEVICE_CA_URL)
-                && packageHelper.isPackageInstalledAndEnabled(applicationContext, AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)
-                && packageHelper.isPackageInstalledAndEnabled(applicationContext, AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME)
-                && AuthenticationConstants.Broker.IPPHONE_APP_SIGNATURE.equals(
-                packageHelper.getCurrentSignatureForPackage(AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME))) {
```
kreedula: Why is this check removed ? I see a TODO, but did we test this ?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -214,35 +214,40 @@ protected boolean processRedirectUrl(@NonNull final WebView view, @NonNull final
...
-                Logger.verbose(TAG + methodName, "Sending intent to launch the CompanyPortal.");
-                final Intent intent = new Intent();
-                intent.setComponent(new ComponentName(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME,
-                        AuthenticationConstants.Broker.COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME));
-                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
-                getActivity().startActivity(intent);
-            } catch (final SecurityException ex) {
-                Logger.warn(TAG + methodName, "Failed to launch Company Portal, falling back to browser.");
-                openLinkInBrowser(url);
+
```
kreedula: nit : most of the lines in this method are very long, way beyond 100 chars, see if you can format to improve readability

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
rpdome: Once we get a clear signal, we can remove this app-specific workaround.

I've tested this by adding ismdmurl=1 at the end of the url and set COMPANY_PORTAL_APP_PACKAGE_NAME to CP package in debug mode, CP is launched.

rpdome: (if we use brokertestapp, then an exception will be thrown as it wouldn't be able to find SplashActivity. In that case we'll proceed to the next step - launching browser and returning AuthenticationConstants.UIResponse.BROWSER_CODE_MDM.

kreedula: okay makes sense, probably this check is the reason, we always end up in Playstore even if CP is installed

rpdome: I guess this check was probably added because playstore is not available on ipphone devices. 

Otherwise the user can redirect to the app through playstore. (one more click -but it's not blocking).

kreedula: Yup, never made sense to me,  as why not directly go to CP in regular case also, i.e. why a specific app check. anyways this should fix

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
Om83: Yes I would be curious to understand why server can't reject it .

rpdome: Intune confirmed they'll reject it. I'll remove this logic.

PR #867, opened 2020-03-20
Fix msal issue 963

shahzaibj:
Fix https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/963

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
@@ -168,6 +168,7 @@ public class BrowserAuthorizationFragment extends AuthorizationFragment {
             } else {
                 cancelAuthorization(true);
             }
+            sCustomTabResponseUri = null;
         }
     }
 


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/BrowserAuthorizationFragment.java
```diff
@@ -168,6 +168,7 @@ public void onResume() {
             } else {
                 cancelAuthorization(true);
             }
+            sCustomTabResponseUri = null;
```
kreedula: This is the problem having static global variables here, changing this would be quite a bit of refactor . But for the fix , i think we should make it null in the `sendResult`  as it's the last exit point. What do you think?

iambmelt: Was going to add a similar comment; since there is static state at-play here, it would be great to have clear entry/exit points where variable state is set/reset

shahzaibj: `sendResult` is on the parent class and this field is not exposed to the parent. I think it is better to keep it here anyway as it is at the end of this block and this means that all the logic within it has already executed and we can safely set it to NULL

shahzaibj: Yeah I ideally we won't have static variables and we've had a number of issues in this fragment, but I believe they are needed in our case. @rpdome Do you have any comments on this as I believe this change was introduced as part of Fragment changes and I see comments from you about using static variables. Can we move away from static?

rpdome: It has been there before my refactoring. From my understanding, this is the only way we can get the result back from Browser.

shahzaibj: Thanks @rpdome for confirming.

shahzaibj: I agree, this would have to be static as it needs to be called from the BrowserTabActivity

shahzaibj: @iambmelt @kreedula What are you guys suggesting? The way I see it right now, the entry point (set) happens in the `createCustomTabResponseIntent` whereas the exit / reset will be the `onResume` method. I can move it to `sendResult` but that means exposing the field to the parent. It is already a static, I don't want to make it public as that would just cause more problems. Can't put the field on the parent class either as should not belong there as it is specific to the Browser case.

kreedula: Okay, lets take the fix in then

iambmelt: k sounds like this is as good as we're gonna do in the near term

PR #870, opened 2020-03-24
Shoatman/rawtypes warnings

shoatman:
Creating the PR, but haven't done any testing yet.  Looking forward to your feedback.  Would also be good to get assistance on coordinating a test pass using this updated version.


shahzaibj:
Why is `IBrokerAccountService` `aidl` changed to a `txt` file?

shoatman:
In order to avoid it being to regen the IBrokerAccountService.java file.  I've added annotations to suppress warning relative to the generated code.  When we want to make a change we can change it back... and repeat the process.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseCommand.java
```diff
@@ -24,20 +24,21 @@
...
 import androidx.annotation.NonNull;
 
+import com.microsoft.identity.common.exception.BaseException;
 import com.microsoft.identity.common.internal.request.OperationParameters;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public abstract class BaseCommand<T> implements Command<T> {
+public abstract class BaseCommand<T extends Object> implements Command<T> {
```
iambmelt: I don't understand this bit -- my understanding is that `BaseCommand<T>` and `BaseCommand<T extends Object>` are synonymous -- what else could T possibly extend?

common/src/main/java/com/microsoft/aad/adal/IBrokerAccountService.java
```diff
@@ -0,0 +1,268 @@
+/*
```
kreedula: Without `.aidl` file , will the broker communication work ? I see the interface is generated here but want to make sure 
Are we seeing warnings due to this regeneration?
Also we do another aidl for MSAL `IMicrosoftAuthService` , isn't that throwing any warnings/errors?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -71,11 +94,10 @@ public EmbeddedWebViewAuthorizationStrategy(@NonNull Context applicationContext,
      * The activity result is set in Authorization.setResult() and passed to the onActivityResult() of the calling activity.
      */
     @Override
-    public Future<AuthorizationResult> requestAuthorization(GenericAuthorizationRequest authorizationRequest,
-                                                            GenericOAuth2Strategy oAuth2Strategy) throws UnsupportedEncodingException {
-        mAuthorizationResultFuture = new ResultFuture<>();
-        mOAuth2Strategy = oAuth2Strategy;
-        mAuthorizationRequest = authorizationRequest;
+    @SuppressWarnings("unchecked")
```
kreedula: Aren't we suppressing warnings `unchecked` here, do we still need to fill in all the generics here ?

PR #873, opened 2020-03-27
Update DisplayMask dependency

rpdome:
DisplayMask is now pulled from the public feed (which is set as an upstream feed in AndroidADAL).


shahzaibj:
@rpdome The build is failing

iambmelt:
@rpdome I just noticed that this PR added `.DS_STORE` file to the repo... we should make sure our `.gitignore` is set to ignore these files

PR #879, opened 2020-04-01
2.0.4 changelog

rpdome:


changelog.txt
```diff
@@ -1,8 +1,20 @@
+Version 2.0.4
```
shahzaibj, line 1: @rpdome I see this got merged but I believe this should be `2.0.5` right?

rpdome: Yup. I fixed that separately.

PR #880, opened 2020-04-02
Refactoring with lombok + Throttling

shahzaibj:
Refactor Operation Parameters into immutable Command Parameters classes using Lombok

Use toBuilder to create a clone when necessary

I've also added back throttling that we used to have in our code before we got the memory leak bug.


common/src/main/java/com/microsoft/identity/common/internal/commands/CommandCallback.java
```diff
@@ -20,7 +20,9 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
-package com.microsoft.identity.common.internal.controllers;
+package com.microsoft.identity.common.internal.commands;
+
+import com.microsoft.identity.common.internal.controllers.TaskCompletedCallbackWithError;
```
shahzaibj, line 25: Is this import needed?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResultCache.java
```diff
@@ -24,6 +24,8 @@
 
 import android.util.LruCache;
 
+import com.microsoft.identity.common.internal.commands.BaseCommand;
```
shahzaibj, line 27: Is this needed?

common/build.gradle
```diff
@@ -84,6 +84,8 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
     implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayoutVersion"
+    compileOnly 'org.projectlombok:lombok:1.18.12'
```
shahzaibj, line 87: Need to move these versions to version files

shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/commands/CommandCallback.java
shahzaibj: Yes, it is needed ofcourse

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResultCache.java
shahzaibj: Yes

common/src/main/java/com/microsoft/identity/common/internal/authscheme/AbstractAuthenticationScheme.java
```diff
@@ -26,9 +26,12 @@
...
 
+import lombok.EqualsAndHashCode;
+
 /**
  * Abstract base class for AuthenticationSchemes.
  */
+@EqualsAndHashCode
 public abstract class AbstractAuthenticationScheme implements INameable {
 
     private static final long serialVersionUID = -2437270903389813253L;
```
rpdome, line 37: exclude?

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
```diff
@@ -0,0 +1,85 @@
...
+    private transient HashMap<String, String> requestHeaders;
+
+    private boolean brokerBrowserSupportEnabled;
+
+    private String loginHint;
+
+    @Expose()
+    private OpenIdConnectPromptParameter prompt;
+
+    @Expose()
```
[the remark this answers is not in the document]
shahzaibj: `Expose` is GSON annotation to indicate whether a field should be serialized/deserialized or not. https://www.javadoc.io/doc/com.google.code.gson/gson/2.6.2/com/google/gson/annotations/Expose.html

It doesn't have anything to do with Lombok and this annotation was already present in our code.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/RemoveAccountCommandParameters.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@SuperBuilder(toBuilder = true)
+@EqualsAndHashCode(callSuper = true)
+public class RemoveAccountCommandParameters extends CommandParameters {
+
+    private IAccountRecord account;
+    private List<BrowserDescriptor> browserSafeList;
```
shahzaibj, line 40: Need immutable getter for this collection too

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -79,6 +86,17 @@ public static void submitSilent(@NonNull final BaseCommand command) {
                 "Beginning execution of silent command."
         );
 
+        if (sExecutingCommands.contains(command)) {
```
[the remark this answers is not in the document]
shahzaibj: Yes, we are. I've updated the title.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -99,9 +106,9 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Acqu
                         parameters.getExtraQueryStringParameters() != null ?
                                 QueryParamsAdapter._toJson(parameters.getExtraQueryStringParameters())
                                 : null
-                ).prompt(parameters.getOpenIdConnectPromptParameter().name())
+                ).prompt(parameters.getPrompt().name())
                 .claims(parameters.getClaimsRequestJson())
-                .forceRefresh(parameters.getForceRefresh())
+                .forceRefresh(false)
```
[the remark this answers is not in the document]
shahzaibj: Interactive doesn't have anything like `forceRefresh`

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -304,40 +280,35 @@ public BrokerAcquireTokenSilentOperationParameters brokerSilentParametersFromBun
...
+                .redirectUri(brokerRequest.getRedirect())
+                .clientId(brokerRequest.getClientId())
+                .forceRefresh(brokerRequest.getForceRefresh())
+                .claimsRequestJson(brokerRequest.getClaims())
+                .loginHint(brokerRequest.getUserName())
+                .homeAccountId(brokerRequest.getHomeAccountId())
+                .localAccountId(brokerRequest.getLocalAccountId())
+                .build();
 
-        if (!TextUtils.isEmpty(brokerRequest.getExtraQueryStringParameter())) {
```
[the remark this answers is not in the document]
shahzaibj: Extra Query parameters are not used for silent token commands and have been removed from silent parameters therefore.

common/src/main/java/com/microsoft/identity/common/internal/authscheme/AbstractAuthenticationScheme.java
```diff
@@ -26,9 +26,12 @@
 
 import com.google.gson.annotations.SerializedName;
 
+import lombok.EqualsAndHashCode;
+
 /**
  * Abstract base class for AuthenticationSchemes.
  */
+@EqualsAndHashCode
```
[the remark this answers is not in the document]
shahzaibj: We don't need to - as part of this change I am not removing existing getters and setters. This equals was only added as it was needed and wasn't there.

common/src/main/java/com/microsoft/identity/common/internal/commands/Command.java
```diff
@@ -20,10 +20,15 @@
...
-package com.microsoft.identity.common.internal.controllers;
+package com.microsoft.identity.common.internal.commands;
+
+import com.microsoft.identity.common.internal.controllers.BaseController;
 
 public interface Command<T> {
     T execute() throws Exception;
 
     boolean isEligibleForEstsTelemetry();
+    BaseController getDefaultController();
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/commands/GetDeviceModeCommand.java
```diff
@@ -20,33 +20,35 @@
...
  */
+@EqualsAndHashCode(callSuper = true)
 public class GetDeviceModeCommand extends BaseCommand<Boolean> {
-    private static final String TAG = GetDeviceModeCommand.class.getSimpleName();
 
-    public GetDeviceModeCommand(@NonNull final OperationParameters parameters,
-                                @NonNull final BaseController controller,
-                                @NonNull final CommandCallback callback) {
-        super(parameters, controller, callback);
+    public GetDeviceModeCommand(@NonNull CommandParameters parameters, @NonNull BaseController controller, @NonNull CommandCallback callback, @NonNull String publicApiId) {
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/commands/SilentTokenCommand.java
```diff
@@ -20,7 +20,7 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
-package com.microsoft.identity.common.internal.controllers;
+package com.microsoft.identity.common.internal.commands;
```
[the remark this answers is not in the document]
shahzaibj: Sure

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -210,42 +200,47 @@ public BrokerAcquireTokenOperationParameters brokerInteractiveParametersFromActi
...
         if (brokerRequest.getAuthorizationAgent() != null
                 && brokerRequest.getAuthorizationAgent().equalsIgnoreCase(AuthorizationAgent.BROWSER.name())
-                && isCallingPackageIntune(parameters.getCallerPackageName())) { // TODO : Remove this whenever we enable System Browser support in Broker for apps.
+                && isCallingPackageIntune(brokerRequest.getApplicationName())) { // TODO : Remove this whenever we enable System Browser support in Broker for apps.
             Logger.info(TAG, "Setting Authorization Agent to Browser for Intune app");
-            parameters.setAuthorizationAgent(AuthorizationAgent.BROWSER);
-            parameters.setBrokerBrowserSupportEnabled(true);
-            parameters.setBrowserSafeList(getBrowserSafeListForBroker());
+            commandParametersBuilder.authorizationAgent(AuthorizationAgent.BROWSER);
+            commandParametersBuilder.brokerBrowserSupportEnabled(true);
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/RemoveAccountCommandParameters.java
shahzaibj: Addressed

shahzaibj:
@rpdome All of your comments are addressed

common/src/main/java/com/microsoft/identity/common/internal/commands/TokenCommand.java
```diff
@@ -0,0 +1,55 @@
...
+import com.microsoft.identity.common.internal.commands.parameters.TokenCommandParameters;
+import com.microsoft.identity.common.internal.controllers.BaseController;
+import com.microsoft.identity.common.internal.result.AcquireTokenResult;
+
+import java.util.List;
+
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode(callSuper = true)
+public abstract class TokenCommand extends BaseCommand<AcquireTokenResult> {
```
kreedula, line 38: Why this class , if we don't have any additional properties associated with it?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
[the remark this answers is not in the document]
kreedula: This would break `claimsRequest` flow. Before this change , `forceRefresh` was on `OperationParameter` [here ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/request/OperationParameters.java#L85) and we set `forceRefresh` to true for interactive calls when `claimsRequest` is present [here ](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java#L147). This is done because, in broker we would attempt to silently get a token for interactive calls as well if we a matching account using PRT, FRT, and if `claimsRequest` is present we would need to skip cache and refresh the token.

So i think you should move back `forceRefresh` to base class and add the logic back

kreedula, line 111: This would break `claimsRequest` flow. Before this change , `forceRefresh` was on `OperationParameter` [here ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/request/OperationParameters.java#L85) and we set `forceRefresh` to true for interactive calls when `claimsRequest` is present [here ](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java#L147). This is done because, in broker we would attempt to silently get a token for interactive calls as well if we a matching account using PRT, FRT, and if `claimsRequest` is present we would need to skip cache and refresh the token.

So i think you should move back `forceRefresh` to base class and add the logic back

shahzaibj: @kreedula `forceRefresh` was indeed set on base class operation parameters but I believe it was never used in any interactive flow and hence I removed it and put it only on silent. Can you share the code snippet where this flow is actually being used during an interactive call?

common/src/main/java/com/microsoft/identity/common/internal/commands/TokenCommand.java
shahzaibj: It has an `abstract` method, and primarily meant as a base `abstract` command for any type of Token Command. Other token commands can just extend this.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
shahzaibj: Okay, I see so it seems it was never used on the `OperationParameters` but we need it on the `BrokerRequest` object to decide if we need to skip cache or not.

shahzaibj: @kreedula I don't think I need to put them on base command parameter class in that case. I think I can just do [this ](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java#L147)logic here at the adapter level and set the value of `forceRefresh` on `BrokerRequest` based on that.

kreedula: It should work with the current implementation, as we unpack the BrokerRequest twice , once [here  ](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/AADAuthenticator/src/msal/java/com/microsoft/identity/client/BrokerClientApplication.java#L349) and the other one to convert to `BrokerAcquireTokenParameters`. I would like to make some changes to avoid unpacking `BrokerRequest` twice especially with compressed payloads, have a work item for it. So it should be on `TokenCommandParameters` so that when we transform `BrokerRequest` to `BrokerAcquireTokenCommandParameters` , we would have the value to set and get

shahzaibj: @kreedula That doesn't make sense to me. All you need is for `forceRefresh` to be available on `BrokerRequest` regardless of how many times we unpack it. Why do you need it on `TokenCommandParameters`? When we skip cache in broker, then we create silent token command parameters and not interactive so you shouldn't need it to be on either interactive parameters or the base class. Would there be a case where we would need `forceRefresh` to live on interactive parameters? Am i missing something here?

kreedula: let's chat offline, so we can resolve this faster

kreedula: Resolved offline, consensus is to `forceRefresh` to `BrokerRequest` based on `claimsRequest`  and leave it out of `TokenCommandParameters`

shahzaibj: Per second discussion decided to go with @kreedula suggestion to put `force refresh` on base token command parameters as we would need to send it to broker request based on the value of Claims Request passed in by the developer and not the merged claims request as that could contain client capabilities.

Addressed it in common here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/880/commits/d5674111b5f49ff0701ffdc437b0baed90848cb5
MSAL here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/993/commits/69984b2d5e9f76c4b686d4e0da42b60bf03186f0
Added back tests here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/993/commits/13a7dae3b6490d88cfc3d6b22917dcb7100f5a09

PR #881, opened 2020-04-07
Add new API ID for SA-PCA overloads

iambmelt:
Adds new API IDs for overloads introduced in:
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/992

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
@@ -59,6 +59,8 @@ public final class PublicApiId {
     // SingleAccountPublicClientApplication;
     //==============================================================================================
     public static final String SINGLE_ACCOUNT_PCA_SIGN_IN = "123";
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
```diff
@@ -59,6 +59,8 @@
     // SingleAccountPublicClientApplication;
     //==============================================================================================
     public static final String SINGLE_ACCOUNT_PCA_SIGN_IN = "123";
+    public static final String SINGLE_ACCOUNT_PCA_SIGN_IN_WITH_PROMPT = "130";
```
iambmelt, line 62: @shahzaibj Once this PR is approved/merged, we should update the telemetry spreadsheet with this info

shahzaibj: Yeah, I'll update that. Thanks @iambmelt

PR #890, opened 2020-04-15
Send cancel command result if base exception is user cancel exception

shahzaibj:
When broker sends broker result to MSAL, then in case of error MSAL adapts it into some type of exception and then the MSAL Broker controller will throw that exception. In case of a User Cancel, that will be the UserCancel exception. In this case, the Command Result returned by MSAL  should be the CANCEL result. Currently, in broker scenarios, we were just returning it an error result even though it should have been a CANCEL result. This PR fixes that scenario.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -153,8 +153,12 @@ public class CommandDispatcher {
         }
 
         if (baseException != null) {
-            //Post On Error
-            commandResult = new CommandResult(CommandResult.ResultStatus.ERROR, baseException);
+            if (baseException instanceof UserCancelException) {
+                commandResult = new CommandResult(CommandResult.ResultStatus.CANCEL, null);
+            } else {
+                //Post On Error
+                commandResult = new CommandResult(CommandResult.ResultStatus.ERROR, baseException);
+            }
         } else {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -153,8 +153,12 @@ private static CommandResult executeCommand(BaseCommand command) {
         }
 
         if (baseException != null) {
-            //Post On Error
-            commandResult = new CommandResult(CommandResult.ResultStatus.ERROR, baseException);
+            if (baseException instanceof UserCancelException) {
+                commandResult = new CommandResult(CommandResult.ResultStatus.CANCEL, null);
```
kreedula, line 157: Shouldn't this be `new CommandResult(CommandResult.ResultStatus.CANCEL, baseException)` ?

shahzaibj: @kreedula It wouldn't make a difference since in the case of cancel, the result object isn't used because onlike the onSuccess and onError, cancel callback does not accept any parameter such as a result value or exception as is the case with onSuccess and onError. See this: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandCallback.java#L26

PR #893, opened 2020-04-22
Shared cache RT pickup, dual client stack support

iambmelt:
### Tracking (corpnet required):
- https://identitydivision.visualstudio.com/Engineering/DevEx%20-%20Android/_queries/query/785e80fc-880c-4271-bfdf-4e15ac87f5d2/
- https://domoreexp.visualstudio.com/DefaultCollection/MSTeams/_workitems/edit/881449

### Changes in this PR:
- When looking up RTs in the local cache, if we don't find any that match our exact client id, use a FoCI RT if one is available

### What does this change enable?
- If another app has stashed tokens in our local cache, we'll pick them up even if the `client_id` doesn't match if one of those apps is FoCI enabled
- Silent auth between two FoCI apps without broker if they dual-client stack

### Don't we risk picking up the wrong RT?
We shouldn't because of few things.
- Non-broker apps are sandboxed, so apps only a share a cache if they are FoCI or if they are "dual-client" (two client ids in one binary)
- As a rule, we should say that FoCI apps can only dual-client stack with another FoCI app. Thou shalt not blend 1P & 3P registrations in a single binary
- Since 3P apps aren't FoCI, this new fallback/lookup logic will never yield a token in those cases
- This logic can be hit inside the broker, but broker sandboxes app caches in a way that this shouldn't make a difference. The broker cache sandboxing works like this:
    * All FoCI apps together
    * All non-FoCI apps get their own cache

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
@@ -553,7 +553,7 @@ public class MsalOAuth2TokenCache
                 null // not applicable
         );
 
-        {
+        if (refreshTokens.isEmpty()) {
             // If we didn't find an RT in the cache, this could be a "TSL-seed" or "dual-client stack"
             // scenario
             //
@@ -569,51 +569,10 @@ public class MsalOAuth2TokenCache
             // Unlike the broker, where we check if an app is FoCI prior to making a network call
             // with an arbitrary FoCI RT we find in the cache, if we're in standalone mode and find
             // a FoCI RT in the cache, the current app must also be FoCI (!!!)
+            final Credential fallbackFrt = getFamilyRefreshTokenForAccount(account);
 
-            if (refreshTokens.isEmpty()) {
-                // Look for an arbitrary RT matching the current user.
-                // If we find one, check that it is FoCI, if it is, assume it works.
-                final List<Credential> fallbackRts = mAccountCredentialCache.getCredentialsFilteredBy(
-                        account.getHomeAccountId(),
-                        account.getEnvironment(),
-                        CredentialType.RefreshToken,
-                        null, // wildcard (*)
-                        isMultiResourceCapable
-                                ? null // wildcard (*)
-                                : account.getRealm(),
-                        isMultiResourceCapable
-                                ? null // wildcard (*)
-                                : target,
-                        null // not applicable
-                );
-
-                if (!fallbackRts.isEmpty()) {
-                    Logger.verbose(
-                            TAG + methodName,
-                            "Inspecting fallback RTs for a FoCI match."
-                    );
-
-                    // Any arbitrary RT should be OK -- if multiple clients are stacked,
-                    // they're either "all FoCI" or none are.
-                    final Credential rt = fallbackRts.get(0);
-
-                    if (rt instanceof RefreshTokenRecord) {
-                        final RefreshTokenRecord refreshTokenRecord = (RefreshTokenRecord) rt;
-
-                        final boolean isFamilyRefreshToken = !StringExtensions.isNullOrBlank(
-                                refreshTokenRecord.getFamilyId()
-                        );
-
-                        if (isFamilyRefreshToken) {
-                            Logger.verbose(
-                                    TAG + methodName,
-                                    "Fallback RT found."
-                            );
-
-                            refreshTokens.add(rt);
-                        }
-                    }
-                }
+            if (null != fallbackFrt) {
+                refreshTokens.add(fallbackFrt);
             }
         }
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -552,6 +553,76 @@ public ICacheRecord load(@NonNull final String clientId,
...
+            // Dual-Client stack: two FoCI-enabled app registrations are sharing a single binary
+            //     and accordingly, can share RTs.
+            //     Examples for this might be TFL/TFW - which uses multiple client ids to enable
+            //     different scenarios depending on enterprise vs. consumer usage
+
+            // Unlike the broker, where we check if an app is FoCI prior to making a network call
+            // with an arbitrary FoCI RT we find in the cache, if we're in standalone mode and find
+            // a FoCI RT in the cache, the current app must also be FoCI (!!!)
+            //
+            // Making the assumption that the current client id can use any FoCI RT we find in the
```
iambmelt, line 573: Remove this comment, may add more confusion than clarity

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -552,6 +553,76 @@ public ICacheRecord load(@NonNull final String clientId,
...
+                        null // not applicable
+                );
+
+                if (!fallbackRts.isEmpty()) {
+                    Logger.verbose(
+                            TAG + methodName,
+                            "Inspecting fallback RTs for a FoCI match."
+                    );
+
+                    // Any arbitrary RT should be OK -- if multiple clients are stacked, they're either
```
shahzaibj, line 602: Are there any (or could there be any) first party apps that are not part of FoCI for some reason?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
shahzaibj: I would prefer to keep the comment, I think it is quite helpful.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -552,6 +553,76 @@ public ICacheRecord load(@NonNull final String clientId,
...
+
+            // Unlike the broker, where we check if an app is FoCI prior to making a network call
+            // with an arbitrary FoCI RT we find in the cache, if we're in standalone mode and find
+            // a FoCI RT in the cache, the current app must also be FoCI (!!!)
+            //
+            // Making the assumption that the current client id can use any FoCI RT we find in the
+            // cache is strictly contingent that app developers NOT mix 1P & 3P registrations into
+            // the same binary. If you do this, Bad Things will happen and you'll get confusing
+            // errors that the RT used doesn't match the client app registration. Also, this
+            // assumption means we don't need to implement "FoCI probing" and/or track FoCI app meta
```
shahzaibj: Left similar comment below, basically I want to know if some app can have multiple client id stacked, let's say three, and all of them are 1P registrations. Could there be case that even though all three are 1P, but for some reason one of them is not part of FoCI, whereas the other two are. Is this case possible? And should we implement probing for it?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Yes, we do have 1P apps that are not FoCI. They won't benefit from these changes

Had an offline sync w/ @shoatman -- the output of our discussion was that any 1P app that plans to dual-client stack _must_ have both be FoCI, or neither. We won't support a 'blended' model where one is and the other is not

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -552,6 +553,70 @@ public ICacheRecord load(@NonNull final String clientId,
                 null // not applicable
         );
 
+        {
```
AdamBJohnsonx, line 556: nit: Since finding refresh tokens is now complicated, should it be in its own (potentially set of) private method(s)?

iambmelt: Yeah, I'll take that feedback. This method becoming long

iambmelt: Resolved via
- a22f456f20136fe51c96c5c2cf5d786c39b841ba

Tweaked javadoc in:
- 3ab2e13c109012a373bc56cee8ad0f81382d2644

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: See comment [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/893#discussion_r413221883) -- we won't support this by-design. Not supporting it let's us dodge having to bring probe logic out of broker

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Ok, will re-add with some slightly modified language based on feedback I received offline

iambmelt:
PMD is angry. Will resolve

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: Readded in 63183e37f5a41ef6f676c57f23377a26d3f584bb

iambmelt:
> PMD is angry. Will resolve

Fixed in 2ecfac346edd4b2a6077f8b127faa28915cd2df0

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -552,6 +552,35 @@ public ICacheRecord load(@NonNull final String clientId,
...
+            //
+            // Making the assumption that the current client id can use any FoCI RT we find in the
+            // cache is strictly contingent that app developers NOT mix FoCI/non-FoCI registrations
+            // into same binary. If you do this, you'll get confusing errors that the RT used doesn't
+            // match the client app registration. This assumption means we don't need to implement
+            // "FoCI probing" and/or track FoCI app meta
+            final Credential fallbackFrt = getFamilyRefreshTokenForAccount(account);
+
+            if (null != fallbackFrt) {
+                refreshTokens.add(fallbackFrt);
```
AdamBJohnsonx, line 580: Is this a safe copy that we're adding to, with guaranteed mutability, for all implementations of the cache interface?  One would think, but the interface documentation is vague on what we're getting back here other than that it implements List.  Should we take a copy before we mutate it or relax the final modifier on the refreshTokens reference so we can just reassign it since we only want the first element anyway?

iambmelt: The `List` is mutable, but as you mention, the interface is vague on this point.

I'll take the suggestion to remove the `final` modifier and will reassign the `List` -- API vagueries should ideally be ironed out as well so I'll spot check classes and make updates to the javadoc where appropriate as part of this PR

There's probably >1 interface floating around here that is ambiguous on the mutability/immutability of the resulting Collection. As you see them, feel free to file an issue (or plural issues) if you see more places like that this that need clarity and those can be addressed either piecemeal or as part of a larger effort

iambmelt: Java/mutability clarifed in:
- 7df4abf2dc56e75d737ab1c7f6c94c99fe8e4c81

List reinit + add (even tho we know List is mutable) in:
- a1b4f1b113bf53c57c349ab7bb114dcd7e83a965

Strictly speaking we could take the attitude that the changes in 7df4abf2dc56e75d737ab1c7f6c94c99fe8e4c81 make a1b4f1b113bf53c57c349ab7bb114dcd7e83a965 unnecessary, but I'm fine to incorporate that original suggestion and leave as-is

PR #894, opened 2020-04-23
Add a new boolean to enable/disable power optimization check for network requests

kreedula:
- Also removed `usageStatsManagerWrapper.isAppInactive` check, as it doesn't seem(no documentation) the right away for us to determine if the app as network or not.


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -476,6 +488,11 @@ public AbstractAuthenticationScheme getAuthenticationScheme() {
             return this;
         }
 
+        public BrokerRequest.Builder powerOptCheckEnabled(@NonNull final boolean powerOptCheckEnabled) {
```
shahzaibj, line 491: nonnull isn't needed

common/src/main/java/com/microsoft/identity/common/adal/internal/net/HttpWebRequest.java
```diff
@@ -188,21 +188,20 @@ public HttpWebResponse send() throws IOException {
      * @param context Context
      * @throws ClientException throw network exception
      */
-    public static void throwIfNetworkNotAvailable(final Context context) throws ClientException {
+    public static void throwIfNetworkNotAvailable(final Context context, final boolean powerOptCheckEnabled) throws ClientException {
```
iambmelt, line 191: Added param needs javadoc

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -64,7 +64,7 @@ public boolean isConnectionAvailable() {
                 .getSystemService(Context.CONNECTIVITY_SERVICE);
         NetworkInfo activeNetwork = connectivityManager.getActiveNetworkInfo();
         @SuppressWarnings("deprecation")
-        final boolean isConnectionAvailable = activeNetwork != null && activeNetwork.isConnectedOrConnecting() && !isNetworkDisabledFromOptimizations();
+        final boolean isConnectionAvailable = activeNetwork != null && activeNetwork.isConnectedOrConnecting();
```
iambmelt, line 67: Just leaving a link to this issue here in case we ever go back and wonder "why didn't we just query `getActiveNetworkInfo()`?"

https://code.google.com/archive/p/android-developer-preview/issues/3164

iambmelt: (Not an action item)

iambmelt:
Related:
https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/998

common/src/main/java/com/microsoft/identity/common/adal/internal/net/HttpWebRequest.java
iambmelt: I might prefer a more descriptive name here. Something along the lines of `skipPowerOptimizationCheck`

PR #898, opened 2020-04-27
Move canUseAccountManagerOperation() to common

rpdome:
As this'll be used in both MSAL and Broker. I...
- moved the logic here from WPJ - see https://github.com/AzureAD/ad-accounts-for-android/pull/1286
- migrated the MANAGE_ACCOUNT permission check from MSAL. (We currently don't run into any issue in Broker because all Broker consumers are >= M)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AccountManagerUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AccountManagerUtil.java
@@ -34,10 +34,12 @@ import androidx.annotation.NonNull;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.internal.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/util/AccountManagerUtil.java
```diff
@@ -0,0 +1,90 @@
...
+import android.os.Build;
+import android.os.UserManager;
+
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.internal.logging.Logger;
+
+public class AccountManagerUtil {
+    private final static String TAG = AccountManagerUtil.class.getSimpleName();
```
AdamBJohnsonx, line 38: nit: modifier ordering should be consistent

PR #902, opened 2020-05-04
ADAL adapter fix

rpdome:
We accidentally passed a hardcoded key into the clientId field, instead of the actual value from the bundle.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
@@ -165,7 +165,7 @@ public class AdalBrokerRequestAdapter implements IBrokerRequestAdapter {
                 .extraQueryStringParameters(extraQP)
                 .authority(authority)
                 .scopes(scopes)
-                .clientId(AuthenticationConstants.Broker.ACCOUNT_CLIENTID_KEY)
+                .clientId(intent.getStringExtra(AuthenticationConstants.Broker.ACCOUNT_CLIENTID_KEY))
                 .redirectUri(redirectUri)
                 .loginHint(intent.getStringExtra(AuthenticationConstants.Broker.ACCOUNT_NAME))
                 .correlationId(correlationIdString)


common/src/main/java/com/microsoft/identity/common/internal/request/AdalBrokerRequestAdapter.java
```diff
@@ -165,7 +165,7 @@ public BrokerInteractiveTokenCommandParameters brokerInteractiveParametersFromAc
                 .extraQueryStringParameters(extraQP)
                 .authority(authority)
                 .scopes(scopes)
-                .clientId(AuthenticationConstants.Broker.ACCOUNT_CLIENTID_KEY)
+                .clientId(intent.getStringExtra(AuthenticationConstants.Broker.ACCOUNT_CLIENTID_KEY))
```
AdamBJohnsonx, line 168: Is this related to an issue, or should there be more in the commit message stating why this alteration needs to be made?

kreedula, line 168: How are the requests working without a client id or how did they work :roll_eyes:  ?
Can you add a desc of the issue?

kreedula: Oh okay this is a regression due to `lombok` related changes, so looks like we haven't tested ADAL with broker for these changes.

shahzaibj: Yeah we haven't tested ADAL with Broker since my Lombok changes. I believe we don't have a version of ADAL with the latest common so it should be working. I guess, Teams will run into issues as they dual stack

rpdome: Yes, we accidentally passed a hardcoded key into the clientId field, instead of the actual value.

rpdome: @kreedula  added.

rpdome: @kreedula On top of what @shahzaibj said, prod ADAL uses 2.0.5, so they won't run into this issue.

shahzaibj: @rpdome I believe Teams can still run into the issue, right? I believe they use the `common` pulled by `MSAL` and they could run into this with the latest MSAL I believe. Have they reported anything?

rpdome: Oh, by 'they' I meant prod ADAL. Yes, Teams would still run into the issue (which is why I said "on top of what you said". Grammar mistake :D

rpdome: It's actually super easy to repro. My guess is that they haven't tried ADAL flow :D

PR #919, opened 2020-05-28
Allow for different Http invocation styles.

AdamBJohnsonx:
- Create a new interface, HttpClient, and an implementation to hold properties of the connection.
- Rebase the HttpRequest methods onto this new client object, and deprecate them.
- Add RetryPolicy objects to handle retries and failures, an interface and 2 impls.
- Add tests to insure that the behavior of the new client and old one.


common/src/main/java/com/microsoft/identity/common/internal/net/HttpClient.java
```diff
@@ -0,0 +1,106 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public interface HttpClient {
```
AdamBJohnsonx, line 35: Javadoc.

AdamBJohnsonx:
> Reviewed offline (and here); overall looks good so approving.
> 
> Leaving a question + a comment:
> 
> * Q: A new `gradlew` (wrapper) is being committed with this change; just wanted to clarify if this was an update to the wrapper or if it was included in this PR accidentally

I set u+x on the wrapper so I could run it from the command line.  The question is - does that hurt anything?  (100644 → 100755, which is odd, because it looks like a+x was set).

> * Comment: We don't have any enforcement on it, but I try to keep my lines roughly in the ballpark of 100 chars wide or less so it looks OK if I view or diff in a terminal. This PR sticks pretty closely to that, but thought I'd mention it.

I try to keep my lines short-ish.  I'm used to working with a hard limit of 150 characters/line and a mental mode of "I don't want to scroll right".  We should probably enforce line length. Is there a particular place that I should trim back?

iambmelt:
> > Reviewed offline (and here); overall looks good so approving.
> > Leaving a question + a comment:
> > 
> > * Q: A new `gradlew` (wrapper) is being committed with this change; just wanted to clarify if this was an update to the wrapper or if it was included in this PR accidentally
> 
> I set u+x on the wrapper so I could run it from the command line. The question is - does that hurt anything? (100644 → 100755, which is odd, because it looks like a+x was set).
> 
> > * Comment: We don't have any enforcement on it, but I try to keep my lines roughly in the ballpark of 100 chars wide or less so it looks OK if I view or diff in a terminal. This PR sticks pretty closely to that, but thought I'd mention it.
> 
> I try to keep my lines short-ish. I'm used to working with a hard limit of 150 characters/line and a mental mode of "I don't want to scroll right". We should probably enforce line length. Is there a particular place that I should trim back?

Oh ok, just a perms change. No worries on the `gradlew`. Shouldn't be a problem.

As I looked thru the PR only 1 place jumped out; `L168` of `UrlConnectionHttpClient`. Rest LGTM 👍

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -87,19 +98,30 @@
         HTTP_METHODS.add(REQUEST_METHOD_PATCH);
     }
 
-    /**
-     * Value of read timeout in milliseconds.
-     */
-    public static int READ_TIMEOUT = 30000;
```
shahzaibj, line 93: @AdamBJohnsonx Why were these removed? It's breaking MSAL. Do you have the corresponding MSAL pr that goes with this common PR?

PR #920, opened 2020-06-01
Changes to look for foci RT and use it for FOCI clients for local msal silent requests

kreedula:
work item : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1008396


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -672,6 +672,27 @@ private Credential getFamilyRefreshTokenForAccount(@NonNull final AccountRecord
         return result;
     }
 
+    /**
+     * Load  FRTs from the cache for an account matching the homeAccountId
+     * @param homeAccountId : homeAccountId for which FRT is sought
+     * @return an FRT if available else null.
+     */
+    @Nullable
+    public RefreshTokenRecord getFamilyRefreshTokenForHomeAccountId(@NonNull final String homeAccountId) {
```
iambmelt, line 681: Can I get you to add two tests to the `MsalOAuth2TokenCacheTest` for this? 1 to test successful retrieval and another to test `null` when no matches

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -127,25 +128,26 @@
...
      * @return true if the given client id can use the cached foci token. False, otherwise.
      * @throws ClientException
      * @throws IOException
      */
-    public static boolean tryFociTokenWithGivenClientId(@NonNull final BrokerOAuth2TokenCache brokerOAuth2TokenCache,
+    public static boolean tryFociTokenWithGivenClientId(@NonNull final OAuth2TokenCache brokerOAuth2TokenCache,
                                                         @NonNull final String clientId,
                                                         @NonNull final String redirectUri,
-                                                        @NonNull final ICacheRecord cacheRecord)
+                                                        @NonNull final RefreshTokenRecord refreshTokenRecord,
```
iambmelt, line 140: API breaking change FYI

Would it be better to make an additive change here?

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -617,14 +619,14 @@ public static MicrosoftStsTokenRequest createTokenRequest(@NonNull final String
         return tokenRequest;
     }
 
-    public static MicrosoftStsAuthorizationRequest createAuthRequest(@NonNull final MicrosoftStsOAuth2Strategy strategy,
-                                                                     @NonNull final ICacheRecord cacheRecord,
+    private static MicrosoftStsAuthorizationRequest createAuthRequest(@NonNull final MicrosoftStsOAuth2Strategy strategy,
```
iambmelt, line 622: Same here it looks like.... another API breaking change I think?

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
kreedula: yup, `broker` uses it too. I made changes on broker , shouldn't have an impact,  but I think it is better to add as an additive to be safe

kreedula: added as an additive change

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
kreedula: added tests as suggested

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -672,6 +672,27 @@ private Credential getFamilyRefreshTokenForAccount(@NonNull final AccountRecord
...
+        final List<AccountRecord> accountRecords = mAccountCredentialCache.getAccounts();
+
+        for (AccountRecord accountRecord : accountRecords) {
+            if (accountRecord.getHomeAccountId().equals(homeAccountId)) {
+                final Credential rt = getFamilyRefreshTokenForAccount(accountRecord);
+
+                if (rt instanceof RefreshTokenRecord) {
+                    return (RefreshTokenRecord) rt;
+                }
```
AdamBJohnsonx, line 690: When does this return something that's not a RefreshTokenRecord?  If I go look at the code, it only sets credential being returned if it is instanceof RefreshTokenRecord, I think.  It's a private method - should we change the signature?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -60,6 +57,9 @@
 import java.util.List;
 import java.util.Set;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
```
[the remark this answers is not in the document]
kreedula: Android studio's  Auto rearrange moved it, so i think was not formatted as AndroidStudio's import format

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -672,6 +672,27 @@ private Credential getFamilyRefreshTokenForAccount(@NonNull final AccountRecord
         return result;
     }
 
+    /**
+     * Load  FRTs from the cache for an account matching the homeAccountId
```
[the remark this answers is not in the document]
kreedula: Addressed via https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -127,25 +128,26 @@
      *
      * @param clientId    String of the given client id.
      * @param redirectUri redirect url string of the given client id.
-     * @param cacheRecord Foci cache record.
+     * @param accountRecord account record of request
+     * @param refreshTokenRecord : refresh token record of FOCI account
```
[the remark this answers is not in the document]
kreedula: typo, Addressed via https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -127,25 +128,26 @@
...
      */
-    public static boolean tryFociTokenWithGivenClientId(@NonNull final BrokerOAuth2TokenCache brokerOAuth2TokenCache,
+    public static boolean tryFociTokenWithGivenClientId(@NonNull final OAuth2TokenCache brokerOAuth2TokenCache,
                                                         @NonNull final String clientId,
                                                         @NonNull final String redirectUri,
-                                                        @NonNull final ICacheRecord cacheRecord)
+                                                        @NonNull final RefreshTokenRecord refreshTokenRecord,
+                                                        @NonNull final IAccountRecord accountRecord)
             throws ClientException, IOException {
         final String methodName = ":tryFociTokenWithGivenClientId";
```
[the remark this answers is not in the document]
kreedula: Currently I believe our `logger` doesn't have functionality to get method names atleast for logs without stack trace. If we get to that functionality it would be great. To answer your question , it's been a pattern we historically followed, I am not a big fan of it personally too, but we haven't invested any cycles to change this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -525,6 +527,41 @@ protected AccountRecord getCachedAccountRecord(
...
+                    Logger.warn(TAG,
+                            "Error while attempting to validate client: "
+                                    + clientId + " is part of family " + e.getMessage()
+                    );
+                }
+            } else {
+                Logger.info(TAG, "No Foci tokens found for homeAccountId " + homeAccountId);
+            }
+        }
+
```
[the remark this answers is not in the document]
kreedula: I'd buy that argument. Addressed via https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -525,6 +527,41 @@ protected AccountRecord getCachedAccountRecord(
                     );
         }
 
+        if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
+            // check for FOCI tokens, if available make a request to service to see if the client id is FOCI and save the tokens
+            final RefreshTokenRecord refreshTokenRecord = ((MsalOAuth2TokenCache) parameters
+                    .getOAuth2TokenCache())
```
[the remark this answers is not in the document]
kreedula: fixed it , now taking it has an `arg`. See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -525,6 +527,41 @@ protected AccountRecord getCachedAccountRecord(
...
+                    .getFamilyRefreshTokenForHomeAccountId(homeAccountId);
+
+            if (refreshTokenRecord != null) {
+                try {
+                    TokenCacheItemMigrationAdapter.tryFociTokenWithGivenClientId(
+                            parameters.getOAuth2TokenCache(),
+                            parameters.getClientId(),
```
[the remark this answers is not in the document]
kreedula: Good catch, fixed it here : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
kreedula: Good point, agree. Addressed via https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/fff3f7e02530a7052cbd302dc428616fff91c435

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -525,6 +528,13 @@ protected AccountRecord getCachedAccountRecord(
                     );
         }
 
+        if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
+            targetAccount = getAccountWithFRTIfAvailable(
+                    parameters,
+                    (MsalOAuth2TokenCache)parameters.getOAuth2TokenCache()
```
AdamBJohnsonx, line 534: nit: space after cast.

kreedula:
@iambmelt @AdamBJohnsonx : Thanks for the reviews, addressed your comments

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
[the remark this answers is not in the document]
AdamBJohnsonx: Cool.  I actually tried writing a few things to change the logging stack, but it gets to be a really big change quickly.  I'm curious why we don't use one of the more standard logging frameworks like, say, slf4j, maybe it's binary size?

Anyway, why I'm concerned a little, from https://developer.android.com/reference/android/util/Log.html :

that when you're building the string to pass into Log.d, the compiler uses a StringBuilder and at least three allocations occur: the StringBuilder itself, the buffer, and the String object. Realistically, there is also another buffer allocation and copy, and even more pressure on the gc. That means that if your log message is filtered out, you might be doing significant work and incurring significant overhead.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -672,6 +672,23 @@ private Credential getFamilyRefreshTokenForAccount(@NonNull final AccountRecord
...
     }
 
+    /**
+     * Load FRTs from the cache for an account matching the homeAccountId
+     * @param homeAccountId : homeAccountId for which FRT is sought
+     * @return an FRT if available else null.
+     */
+    @Nullable
+    public RefreshTokenRecord getFamilyRefreshTokenForHomeAccountId(@NonNull final String homeAccountId) {
+        final List<AccountRecord> accountRecords = mAccountCredentialCache.getAccounts();
```
AdamBJohnsonx, line 682: nit: inline?

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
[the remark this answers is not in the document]
iambmelt: IIRC, we looked at doing this awhile back and saw a lot of slowness around using `Thread.getStackTrace` to inspect the caller function. Now that threading has moved to a background thread this may be less of concern, depending on the internal mechanics of the `getStackTrace` implementation and runtime (ART vs Dalvik).

common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -135,17 +139,40 @@
...
         final MicrosoftStsOAuth2Configuration config = new MicrosoftStsOAuth2Configuration();
 
         //Get authority url
         final Uri.Builder requestUrlBuilder = new Uri.Builder();
-        final String tenantId = cacheRecord.getAccount().getRealm();
         requestUrlBuilder.scheme("https")
-                .authority(cacheRecord.getRefreshToken().getEnvironment())
-                .appendPath(StringUtil.isEmpty(tenantId) ? ALL_ACCOUNTS_TENANT_ID : tenantId);
+                .authority(refreshTokenRecord.getEnvironment())
+                .appendPath(StringUtil.isEmpty(accountRecord.getRealm()) ? ALL_ACCOUNTS_TENANT_ID : accountRecord.getRealm());
```
AdamBJohnsonx, line 175: Random comment: I kinda want to write StringUtil.defaultIfEmpty(String a, String b) -> isEmpty(a) ? b : a;

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -552,6 +562,48 @@ protected AccountRecord getCachedAccountRecord(
...
+        if (refreshTokenRecord != null) {
+            try {
+                // foci token is available, make a request to service to see if the client id is FOCI and save the tokens
+                TokenCacheItemMigrationAdapter.tryFociTokenWithGivenClientId(
+                        parameters.getOAuth2TokenCache(),
+                        clientId,
+                        parameters.getRedirectUri(),
+                        refreshTokenRecord,
+                        parameters.getAccount()
+                );
```
AdamBJohnsonx, line 585: OK, we have the token then we make this call.  We stuff it into the datastore as a side effect, then go back to the datastore to pull it out again.  There's probably a reason we don't just return it... but anyway, this method returns a boolean value that we're not checking.  Should we check it?  Should we return the token instead?

kreedula: The use case here is:

We have a `Family Refresh Token` for an Microsoft first party client id(Teams for Life here) for this user after he signs in

Now we get a silent request for this `IAccount` with another Microsoft first party client id(Teams for Work) - at this point we don't have any account associated with this client id. 

What this change enables is, if the user has Family Refresh Token (applicable only for MSFT ), we try to make a `dummy` silent request in `tryFociTokenWithGivenClientId` with the FRT, default scopes and the new client id  and if the request succeeds we save the account and tokens for this client id in this cache. The access token here won't be used and will be replaced as it has just the default scopes.

Now the caller when it looks for the `IAccount` for this client, it gets one and makes the actual request for an access token for the required scopes passed in using the new RT which was saved. 

The boolean can be ignored in our case, as we have sufficient logging in `tryFociTokenWithGivenClientId` to understand if the request succeeds. 

Hope this helps

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
kreedula: Addressed via : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/b8f4d0ed5b84c37dd34def53aeada19dce3418c0

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
kreedula: Addressed via : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/920/commits/b8f4d0ed5b84c37dd34def53aeada19dce3418c0

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
AdamBJohnsonx: Oh, no, thanks for bearing with me. 

OK, so the interesting side effect of tryFociWithGivenClientId is to bind the FRT to the new client id and store these refresh tokens by client id in the cache.  Is it possible to return a non-null AccountRecord from the call to getAccountByLocalAccountId even if the attempt to execute the family token refresh failed with the new client id? It seems like it shouldn't, right now, but that's largely because I'm thinking that the sole purpose of this method is to return an account iff this is a client id that part of the family, and if that was the case the refresh would have succeeded.

kreedula: yes, not just the side, i could say the primary purpose of this `tryFociWithGivenClientId` is to insert an `AccountRecord` for the given client, it's part of FOCI (we determine this by sending a request with FRT and if it succeeds).

And yes, if the token request with FRT failed , we should not get an `AccountRecord` when `getAccountByLocalAccountId` is invoked with that client id in this scenerio

PR #924, opened 2020-06-09
Bring back ipphone check

rpdome:
Ideally, we would want to launch CP when it's installed (instead of going to the browser). This was introduced as an ask from ipphone.

However, I missed the case where CP could be preloaded on the device and the customer is actually using 3rd party MDM, and I **prematurely** removed the ipphone check (and also introduce a stupid bug :/)

(Eventually, we'll bring this behavior back, but Intune will have to work with eSTS to forward us a signal).

There is no issue/icm being filed yet.

related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/865/files (AzureActiveDirectoryWebViewClient.java, line 220)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -223,7 +223,15 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             final Context applicationContext = getActivity().getApplicationContext();
 
             // If CP is installed, redirect to CP.
-            if (!packageHelper.isPackageInstalledAndEnabled(
+            // TODO: Until we get a signal from eSTS that CP is the MDM app, we cannot assume that.
+            //       CP is currently working on this.
+            //       Until that comes, we'll only handle this in ipphone.
+            if (packageHelper.isPackageInstalledAndEnabled(
+                    applicationContext,
+                    AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME) &&
+                AuthenticationConstants.Broker.IPPHONE_APP_SIGNATURE.equals(
+                        packageHelper.getCurrentSignatureForPackage(AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME)) &&
+                packageHelper.isPackageInstalledAndEnabled(
                     applicationContext,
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -223,7 +223,15 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
             final Context applicationContext = getActivity().getApplicationContext();
 
             // If CP is installed, redirect to CP.
-            if (!packageHelper.isPackageInstalledAndEnabled(
+            // TODO: Until we get a signal from eSTS that CP is the MDM app, we cannot assume that.
+            //       CP is currently working on this.
+            //       Until that comes, we'll only handle this in ipphone.
+            if (packageHelper.isPackageInstalledAndEnabled(
+                    applicationContext,
+                    AuthenticationConstants.Broker.IPPHONE_APP_PACKAGE_NAME) &&
```
AdamBJohnsonx, line 231: nit: if this became local ipPhonePackage, and ipPhoneSignature, say, this statement would be easier to read.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -223,14 +226,17 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
...
             // If CP is installed, redirect to CP.
-            if (!packageHelper.isPackageInstalledAndEnabled(
-                    applicationContext,
-                    AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)) {
+            // TODO: Until we get a signal from eSTS that CP is the MDM app, we cannot assume that.
+            //       CP is currently working on this.
+            //       Until that comes, we'll only handle this in ipphone.
+            if (packageHelper.isPackageInstalledAndEnabled(applicationContext, IPPHONE_APP_PACKAGE_NAME) &&
+                IPPHONE_APP_SIGNATURE.equals(packageHelper.getCurrentSignatureForPackage(IPPHONE_APP_PACKAGE_NAME)) &&
+                packageHelper.isPackageInstalledAndEnabled(applicationContext, COMPANY_PORTAL_APP_PACKAGE_NAME)) {
```
AdamBJohnsonx, line 234: As I read this, we're changing from "If company portal IS NOT installed and enabled, try to launch company portal" to "If ipphone is installed and enabled and has the correct signature and company portal IS installed, try to launch company portal"

PR #925, opened 2020-06-09
Add correlation id to local authentication result

shahzaibj:
See MSAL pr: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1063 for context


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -341,6 +350,14 @@ public void onReceive(Context context, Intent intent) {
...
                     localBroadcastManager.unregisterReceiver(resultReceiver);
 
+                    // set correlation id on Local Authentication Result
+                    if (commandResult.getResult() != null &&
+                            commandResult.getResult() instanceof LocalAuthenticationResult) {
+                        final LocalAuthenticationResult localAuthenticationResult =
+                                (LocalAuthenticationResult) commandResult.getResult();
+                        localAuthenticationResult.setCorrelationId(correlationId);
+                    }
+
```
AdamBJohnsonx, line 360: should this become a method?

common/src/main/java/com/microsoft/identity/common/internal/result/LocalAuthenticationResult.java
```diff
@@ -209,4 +210,14 @@ public AccessTokenRecord getAccessTokenRecord() {
     public boolean isServicedFromCache() {
         return mServicedFromCache;
     }
+
+    public void setCorrelationId(@NonNull final String correlationId) {
```
AdamBJohnsonx, line 214: nit: @Override?

shahzaibj: The setter is not declared on the interface and instead declared directly here. So override isn't needed

AdamBJohnsonx: Oops.  You're right.  It says set, not get.  You could @Setter that then if you wanted to.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shahzaibj: @AdamBJohnsonx Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/925/commits/21e14b3679de09f10658a2f08142f1ca41182627

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -495,4 +502,15 @@ public static int getCachedResultCount() {
         return sCommandResultCache.getSize();
     }
 
+    private static void setCorrelationIdOnResult(@NonNull final CommandResult commandResult,
+                                                 @NonNull final String correlationId) {
+        // set correlation id on Local Authentication Result
+        if (commandResult.getResult() != null &&
```
shoatman, line 508: I think it would be a good idea to comment on why we have the check for localAuthenticationResult.  I'm assuming for the broker scenarios, but adding the comment will help future readers.

shahzaibj: Sure, I'll add that - the check is there because `CommandResult` is a generic object and can contain the result of other commands such as GetAccount command or GetDeviceCommand and those results are not instances of `LocalAuthenticationResult` and don't contain a property for correlation id.

shoatman: I think it's the local prefix that confuses me here.  So perhaps the comment isn't required...

PR #927, opened 2020-06-09
Removed AccountRecord validation while saving credentials

kreedula:
- Change for OneAuth
- MSALCPP expects `AccountRecord` and `Credentials` to be independant of each other, so removed hard validation of `AccountRecord` `non-null` while writing credentials.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
@@ -24,16 +24,11 @@ package com.microsoft.identity.common.internal.cache;
 
 import android.content.Context;
 
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-
 import com.microsoft.identity.common.BaseAccount;
 import com.microsoft.identity.common.exception.ClientException;
-import com.microsoft.identity.common.internal.dto.AccessTokenRecord;
 import com.microsoft.identity.common.internal.dto.AccountRecord;
 import com.microsoft.identity.common.internal.dto.Credential;
 import com.microsoft.identity.common.internal.dto.CredentialType;
-import com.microsoft.identity.common.internal.dto.IdTokenRecord;
 import com.microsoft.identity.common.internal.dto.RefreshTokenRecord;
 import com.microsoft.identity.common.internal.logging.Logger;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -93,37 +91,26 @@ public static MsalCppOAuth2TokenCache create(@NonNull final Context context) {
...
 
-        AccessTokenRecord accessTokenRecord = null;
-        IdTokenRecord idTokenRecord = null;
         RefreshTokenRecord refreshTokenRecord = null;
 
         for (final Credential credential : credentials) {
-            if (credential instanceof AccessTokenRecord) {
-                accessTokenRecord = (AccessTokenRecord) credential;
-            } else if (credential instanceof IdTokenRecord) {
-                idTokenRecord = (IdTokenRecord) credential;
```
jbzdarkid, line 115: I'm not sure I understand this removal -- we still will make a single WriteCredentials call, which may contain multiple credentials

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -93,37 +91,26 @@ public static MsalCppOAuth2TokenCache create(@NonNull final Context context) {
...
-
-        validateNonNull(accountRecord, "AccountRecord");
-        validateNonNull(refreshTokenRecord, "RefreshTokenRecord");
-        validateNonNull(idTokenRecord, "IdTokenRecord");
-        validateCacheArtifacts(accountRecord, accessTokenRecord, refreshTokenRecord, idTokenRecord);
-
-        removeRefreshTokenIfNeeded(accountRecord, refreshTokenRecord);
+        if (accountRecord != null && refreshTokenRecord != null) {
+            removeRefreshTokenIfNeeded(accountRecord, refreshTokenRecord);
+        }
```
jbzdarkid, line 113: Just for my own edification, can you add a comment here?
```suggestion
        if (accountRecord != null && refreshTokenRecord != null) {
            // This will not be true when we are called from MSAL C++, as the account will always be null.
            removeRefreshTokenIfNeeded(accountRecord, refreshTokenRecord);
        }
```
Feel free to reword.

PR #928, opened 2020-06-11
Add consumer rules for common; dontwarn for lombok

shahzaibj:
A recent change in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/919 seems to break CP builds throwing some proguard warning associated to Lombok.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -52,6 +52,7 @@ android {
             debuggable false
             testCoverageEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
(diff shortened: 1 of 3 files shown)


kreedula:
the file should be referenced [here ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/build.gradle#L54)as `consumerProguardFiles 'consumer-rules.pro'` :

shahzaibj:
@kreedula Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/928/commits/d39e4e657140c58965b3368d69425b443af3596c

PR #929, opened 2020-06-12
Changes to allow applications to set OIDC scopes

kreedula:


shahzaibj:
So looks like this is targeted for the current broker release?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -469,8 +469,12 @@ protected boolean idTokenIsNull(@NonNull final ICacheRecord cacheRecord,
         return null == idTokenRecord;
     }
 
-    protected Set<String> addDefaultScopes(@NonNull final TokenCommandParameters commandParameters) {
+    protected Set<String> addDefaultAndOidcScopes(@NonNull final TokenCommandParameters commandParameters) {
         final Set<String> requestScopes = commandParameters.getScopes();
+
+        if (commandParameters.getOidcScopes() != null && !commandParameters.getOidcScopes().isEmpty()) {
```
AdamBJohnsonx, line 475: TokenCommandParameters::getScopes() can pretty explicitly return null.  Should we handle that?

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/TokenCommandParameters.java
```diff
@@ -61,6 +61,9 @@
     @Expose()
     private boolean forceRefresh;
 
+    @Expose()
+    private Set<String> oidcScopes;
+
     public Set<String> getScopes() {
```
AdamBJohnsonx, line 67: Maybe add @Nullable to getScopes(), or is that not an actual possibility?

iambmelt:
Can we add a description to PR covering what's being added, how/why?

kreedula:
Closing this as this is no longer required for Teams

PR #933, opened 2020-06-14
Adds new tests for MsalCppOAuth2TokenCache

iambmelt:
Adds new test, API surface to caching components for CPP scenarios

## Summary of changes
- Adds `MsalCppOAuth2TokenCacheTest`
- Adds new `removeAccount` overload to `OAuth2TokenCache`
    * API accepts a varargs array of `CredentialType` to delete upon removal of the `AccountRecord`
- Makes behavior of `MsalCppOAuth2TokenCache#remove` match the `Credential` deletion behavior specified [here](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/blob/develop/source/zgenerated/java/com/microsoft/identity/internal/StorageManager.java#L68-L77).
    - Deletes: access tokens, id tokens, account
    - Preserves: the refresh token

## Points to clarify before merge
- Android MSAL has previously defined the presence of an Account (for `getAccounts()`'s purposes) as: an `AccountRecord` with a matching `RefreshTokenRecord` for a given `client_id`
    * The CPP APIs are more free form: `getAllAccounts()` returns whatever `AccountRecords` are in the cache, irrespective of whether we have an RT for this client or not
    * `removeAccount()` has different behavior: you can only delete an `AccountRecord` for which we have an RT.... **should this be the case?**
        - This _could_ indiscriminately delete the account and its associated AT, ID tokens
        - It's possible, in the current implementation for an `AccountRecord` to get "stuck" -- if you save an `AccountRecord` but don't supply an RT for it, then it's as if we don't have it. `getAccount()` will still return it though....

>Brian's thoughts....
> I am thinking that `removeAccount()` should also call `forceRemove()` to _forcibly_ expel the `AccountRecord` from the cache, if it exists. This is probably the closest behavior to what CPP expects.... 

## Summary of offline decision
For now, the implementation will assume a single stack. Force delete any account that was requested for removal even if no RTs exist for it; under normal single-stack conditions this is fine.

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
@@ -1,0 +1,249 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common;
+
+import android.content.Context;
+
+import androidx.test.InstrumentationRegistry;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import com.microsoft.identity.common.adal.internal.AndroidSecretKeyEnabledHelper;
+import com.microsoft.identity.common.exception.ClientException;
+import com.microsoft.identity.common.internal.authscheme.BearerAuthenticationSchemeInternal;
+import com.microsoft.identity.common.internal.cache.AccountDeletionRecord;
+import com.microsoft.identity.common.internal.cache.ICacheRecord;
+import com.microsoft.identity.common.internal.cache.MsalCppOAuth2TokenCache;
+import com.microsoft.identity.common.internal.dto.AccountRecord;
+import com.microsoft.identity.common.internal.dto.Credential;
+import com.microsoft.identity.common.internal.dto.CredentialType;
+import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftAccount;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.List;
+
+import static com.microsoft.identity.common.MicrosoftStsAccountCredentialAdapterTest.MOCK_ID_TOKEN_WITH_CLAIMS;
+import static com.microsoft.identity.common.MsalOAuth2TokenCacheTest.AccountCredentialTestBundle;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.CACHED_AT;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.CLIENT_ID;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.ENVIRONMENT;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.EXPIRES_ON;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.HOME_ACCOUNT_ID;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.LOCAL_ACCOUNT_ID;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.REALM;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.SECRET;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.TARGET;
+import static com.microsoft.identity.common.SharedPreferencesAccountCredentialCacheTest.USERNAME;
+
+@RunWith(AndroidJUnit4.class)
+public class MsalCppOAuth2TokenCacheTest extends AndroidSecretKeyEnabledHelper {
+
+    private MsalCppOAuth2TokenCache mCppCache;
+
+    // Test Accounts/Credentials
+    private AccountCredentialTestBundle mTestBundle;
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+
+        // Context and related init
+        final Context context = InstrumentationRegistry.getTargetContext();
+        mCppCache = MsalCppOAuth2TokenCache.create(context);
+
+        // Credentials for testing
+        mTestBundle = new AccountCredentialTestBundle(
+                MicrosoftAccount.AUTHORITY_TYPE_V1_V2,
+                LOCAL_ACCOUNT_ID,
+                USERNAME,
+                HOME_ACCOUNT_ID,
+                ENVIRONMENT,
+                REALM,
+                TARGET,
+                CACHED_AT,
+                EXPIRES_ON,
+                SECRET,
+                CLIENT_ID,
+                SECRET,
+                MOCK_ID_TOKEN_WITH_CLAIMS,
+                null,
+                CredentialType.V1IdToken
+        );
+    }
+
+    @After
+    public void tearDown() {
+        mCppCache.clearCache();
+    }
+
+    @Test
+    public void saveAndGetAccountTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+
+        // Save the the cache
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        // Restore it
+        final AccountRecord restoredAccount = mCppCache.getAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+
+        Assert.assertNotNull(restoredAccount);
+        Assert.assertEquals(generatedAccount, restoredAccount);
+    }
+
+    @Test
+    public void getAccountNullTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        final AccountRecord restoredAccount = mCppCache.getAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertNull(restoredAccount);
+    }
+
+    @Test
+    public void getAllAccountsTest() {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        final List<AccountRecord> accounts = mCppCache.getAllAccounts();
+        final AccountRecord restoredAccount = accounts.get(0);
+        Assert.assertEquals(generatedAccount, restoredAccount);
+
+        // TODO this API makes no statement about the immutability of the List returned
+        // TODO results should not be mutable
+    }
+
+    @Test
+    public void getAllAccountsEmptyTest() {
+        final List<AccountRecord> accounts = mCppCache.getAllAccounts();
+        Assert.assertTrue(accounts.isEmpty());
+    }
+
+    @Ignore // Ignore this test until API behavior is decided
+    @Test
+    public void removeAccountTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertEquals(generatedAccount, deletionRecord.get(0));
+
+        // TODO There's confusing behavior here...
+        // In MSAL we have defined an account as "existing" if we have an RT for it
+        // This API doesn't make this constraint obvious. Should we delete any matching account?
+        // In a double-client stack situation, this could delete an account from another app (TFL/TFW)
+    }
+
+    @Test
+    public void removeNonexistentAccountTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertTrue(deletionRecord.isEmpty());
+    }
+
+    @Test
+    public void saveAccountRecordIncompleteTest() {
+
+    }
+
+    @Test
+    public void saveCredentialsWithAccountTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        mCppCache.saveCredentials(
+                generatedAccount,
+                mTestBundle.mGeneratedAccessToken,
+                mTestBundle.mGeneratedIdToken,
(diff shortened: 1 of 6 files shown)


common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,249 @@
...
+    @Test
+    public void getAllAccountsTest() {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        final List<AccountRecord> accounts = mCppCache.getAllAccounts();
+        final AccountRecord restoredAccount = accounts.get(0);
+        Assert.assertEquals(generatedAccount, restoredAccount);
+
+        // TODO this API makes no statement about the immutability of the List returned
```
iambmelt, line 142: 👀

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,249 @@
...
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertEquals(generatedAccount, deletionRecord.get(0));
+
+        // TODO There's confusing behavior here...
```
iambmelt, line 164: 👀

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
AdamBJohnsonx: What happens if you write to it?  Should we have standard tests against our interface packages that all implementations must pass?

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,249 @@
...
+
+        // Restore it
+        final AccountRecord restoredAccount = mCppCache.getAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+
+        Assert.assertNotNull(restoredAccount);
+        Assert.assertEquals(generatedAccount, restoredAccount);
```
AdamBJohnsonx, line 119: Should this be the same object reference?

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
AdamBJohnsonx: What's the use case for removeAccount?  It seems like we either need to remove it from *all* clients (because, say, this is called when the user is being removed from their home tenant) or we need to be able to specify the client we're removing the account from (because, say, the user is being removed from an application).  Either way, the use case should tell you which way to go.

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,249 @@
...
+    public void removeAccountTest() throws ClientException {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertEquals(generatedAccount, deletionRecord.get(0));
+
```
AdamBJohnsonx, line 163: Should we go on and check that we don't get it from the cache?

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,249 @@
...
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                generatedAccount.getRealm()
+        );
+        Assert.assertTrue(deletionRecord.isEmpty());
+    }
+
+    @Test
+    public void saveAccountRecordIncompleteTest() {
+
```
AdamBJohnsonx, line 183: Do we need this test?

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
AdamBJohnsonx: It looks like it delegates here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/c2dedc4786862ea48f67bf00f85c0c1e3d9ec127/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java#L1039 which is an unmodifiable wrapper.

The other implementation is SharedPreferencesAccountCredentialCache, which returns a safe-copy.  https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/c2dedc4786862ea48f67bf00f85c0c1e3d9ec127/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java#L223 We should pick one strategy.  And document it. 

And, in relation to my other comment, do such things *need* to return a list?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -132,10 +134,24 @@ public synchronized void saveAccountRecord(@NonNull AccountRecord accountRecord)
...
     }
 
+    /**
+     * API to inspect cache contents.
+     * Note: This method is intended to be only used for testing purposes.
+     *
+     * @return A immutable List of Credentials contained in this cache.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    public synchronized List<Credential> getCredentials() {
```
AdamBJohnsonx, line 149: Since we're adding this API, do we *need* to return a list?  Would it be sufficient just to return an Iterator<Credential>, since the contains method on List is going to have to search it anyway?

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
AdamBJohnsonx: OK, yeah, this is still WIP.  Nevermind.

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
AdamBJohnsonx: Looks like no, because this will write to the storage layer and something else could also write there and we should see it.

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
iambmelt: I am leaning towards a standardized approach to the APIs whereby we designate the results are read-only and enforced via some unmodifiable wrapper is my thinking. There's a bit of blend of these in the code, as you're noticing.

IMO we should document on the API (via javadoc) and encourage our developers/contributors to do the same; we should be diligent about it in our code reviews. W.r.t to where else we could/should document such style decisions, the contributors guide/wiki seem like good candidates.

AdamBJohnsonx: Yes.  I'm a huge proponent of the ImmutableObjects paradigm, if it hasn't become apparent.  It simplifies a lot of thread-safety concerns, though it does imply behaviour changes on the part of clients.  We probably can't get there from here without a MV release, and it might be moot via OneAuth and MSAL-cpp.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
iambmelt: This certainly _could_ return an Iterator, though the immediate benefit of this change isn't obvious to me since basically the only thing I want to do with this collection is call `contains()` on it in a test

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
iambmelt: Will add

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
AdamBJohnsonx: Point.  This is only being added for testing.

iambmelt: I think the implicit/broader point though is "do we need all these `List` functions in a world of immutable state?" which is a good point to consider esp. given the potential perks an `Iterator` can provide (lazy loading, etc).

So I'll decline this particular change, but ACK the sentiment and will ponder it henceforth if we add additional cache APIs

iambmelt:
@AdamBJohnsonx 
>What's the use case for removeAccount? It seems like we either need to remove it from all clients (because, say, this is called when the user is being removed from their home tenant) or we need to be able to specify the client we're removing the account from (because, say, the user is being removed from an application). Either way, the use case should tell you which way to go.

The CPP [javadocs](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/blob/develop/source/zgenerated/java/com/microsoft/identity/internal/StorageManager.java#L68-L77) describe the desired behavior as:
```java
    /**
     * Deletes an account and all associated credentials.
     * Specifically, it removes all associated access tokens and id tokens.
     * It does not remove any refresh tokens or family refresh tokens, because those are associated with multiple accounts.
     * correlationId: optional
     * homeAccountId: required
     * environment: required
     * realm: required
     */
    public abstract OperationStatus deleteAccount(String correlationId, String homeAccountId, String environment, String realm);
```

It's a little questionable that the docs here say: "removes all associated credentials" and then backtrack to say "well, actually, it's only idtokens and access tokens" 🤔 

So the behavior here is different than Android MSAL, where refresh tokens are _also_ removed. This is a bug, which I'll fix as part of this PR /cc @kreedula FYI

iambmelt:
> I'm not sure this is done, because there is a blank method in the test file. But it looks like there are a lot of questions still in this review.

Yup, work-in-prog

iambmelt:
> > What's the use case for removeAccount? It seems like we either need to remove it from all clients (because, say, this is called when the user is being removed from their home tenant) or we need to be able to specify the client we're removing the account from (because, say, the user is being removed from an application). Either way, the use case should tell you which way to go.
> 
> The CPP [javadocs](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/blob/develop/source/zgenerated/java/com/microsoft/identity/internal/StorageManager.java#L68-L77) describe the desired behavior as:
> 
> ```java
>     /**
>      * Deletes an account and all associated credentials.
>      * Specifically, it removes all associated access tokens and id tokens.
>      * It does not remove any refresh tokens or family refresh tokens, because those are associated with multiple accounts.
>      * correlationId: optional
>      * homeAccountId: required
>      * environment: required
>      * realm: required
>      */
>     public abstract OperationStatus deleteAccount(String correlationId, String homeAccountId, String environment, String realm);
> ```
> 
> It's a little questionable that the docs here say: "removes all associated credentials" and then backtrack to say "well, actually, it's only idtokens and access tokens" 🤔
> 
> So the behavior here is different than Android MSAL, where refresh tokens are _also_ removed. This is a bug, which I'll fix as part of this PR /cc @kreedula FYI

Added new API surface to support this in: 3f554281d073e1ef36db626afb77dc122caf7958

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    @Test
+    public void getAllAccountsTest() {
+        final AccountRecord generatedAccount = mTestBundle.mGeneratedAccount;
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        final List<AccountRecord> accounts = mCppCache.getAllAccounts();
+        final AccountRecord restoredAccount = accounts.get(0);
+        Assert.assertEquals(generatedAccount, restoredAccount);
```
shahzaibj, line 139: Let's also add an assert statement for the size of the list

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -147,8 +205,20 @@ public synchronized void clearCache() {
...
+        // Normally on Android, apps are 'sandboxed' such that each app has their own cache
+        // and we don't have to worry about 1 app stomping on another's cache
+        //
+        // TFW/TFL however, "double stacked" their app registrations into a single binary
+        // Such that calling removeAccount() will potentially remove the Account being used by
+        // another app.
+        //
+        // This API assumes the *general* case where an app is single stacked. If special
+        // accommodations need to come later for Teams then we can reevaluate the logic here.
+
```
shahzaibj, line 218: This cache `MsalCppOauth2TokenCache` is local cache, right? Like this is when local auth is happening and not broker?

iambmelt: Correct, this is always local cache -- not a broker

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
iambmelt: Will do

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
iambmelt: Resolving this; feel free to reactivate if you have any more questions

shahzaibj: In that case, this shouldn't be problematic, right? TFW and TFL are two different packages right?

iambmelt: Nope, they are a single binary. 2 app registrations shipped inside of a single app from the app store. 

If a user switches between them, they use the same cache on disk because they are for Android's purposes "the same app"

iambmelt: It's configured somewhat similarly to our testapp; Teams chooses which app.json to load and which client_id to use based on if the user is enterprise vs. consumer

common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
iambmelt: 12e089b35a9012b7423ac1190782eccb037bbcb4

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
shahzaibj: Got it, Thanks!

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -123,19 +127,73 @@ public synchronized void saveCredentials(@Nullable final AccountRecord accountRe
...
+                                                                 @Nullable final String environment,
+                                                                 @NonNull final String realm) throws ClientException {
+        validateNonNull(homeAccountId, "homeAccountId");
+        validateNonNull(realm, "realm");
+
+        final boolean mustMatchOnEnvironment = !StringExtensions.isNullOrBlank(environment);
+
+        final List<AccountRecord> removedAccounts = new ArrayList<>();
+
+        for (final AccountRecord accountRecord : getAllAccounts()) {
```
shahzaibj, line 176: Don't quite remember how JAVA executes this....would the `getAllAccounts` method be executed on each iteration of the loop? Would it be better to call this method before the loop and store the result in a variable and pass that to the loop?

iambmelt: This syntax is shorthand for using an iterator; `getAllAccounts()` will be executed once, then an iterator from it will be returned.  Example below:
image

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -1175,6 +1189,42 @@ public AccountDeletionRecord removeAccount(@Nullable final String environment,
...
+     * @param realm         The tenant id of the targeted Account (if applicable).
+     * @param typesToRemove The CredentialTypes to delete for the targeted Account.
+     * @return An {@link AccountDeletionRecord}, containing the deleted {@link AccountDeletionRecord}s.
+     */
+    @Override
+    public AccountDeletionRecord removeAccount(@Nullable final String environment,
+                                               @Nullable final String clientId,
+                                               @Nullable final String homeAccountId,
+                                               @Nullable final String realm,
+                                               @Nullable final CredentialType... typesToRemove) {
```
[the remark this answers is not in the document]
iambmelt: If that happens, we should proceed to remove the account and preserve any credentials. Added a warning log message if null or empty array is provided (cb2ff2d0a902b2626ba7c645cdf511418aabe984) 👍

PR #938, opened 2020-06-20
Saml Assertion associated constant

paivybhav:
Made changes to the AuthenticationConstants file by adding constants needed for the saml assertion feature. These constants are needed for building the request message.


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -321,6 +321,11 @@
          */
         public static final String SCOPE = "scope";
 
+        /**
+         * Scope value used in acquiring tokens with saml assertion.
+         */
+        public static final String MSID_OAUTH2_SCOPE_OPENID_VALUE = "openid";
```
iambmelt, line 327: Oops. Looks like we already have a constant for this
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/6dda4f95ca792b6c6f77a69347d56693dc8e75e1/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java#L1233

paivybhav: > Oops. Looks like we already have a constant for this

I was following the constants added in iOS for supporting the assertion feature should have checked if this one was present. I have removed this constant and made changes in the ADAL SDK too

paivybhav: Hey @iambmelt, I have made the requested changes and added a new commit. Hoping everything else is fine and resolving the conversation

shahzaibj:
Could these constants be moved to ADAL? That would avoid doing a common release. (We just did a common release 5 days ago and the only delta are these constants)

shahzaibj:
Looks like declare constants in ADAL that reference these constants [here](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/625ad73311fcb34197b1aa05a65ffc4bb9db6adf/adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java#L342-L348) We could probably just declare these there to avoid doing a common release just for the constants.

paivybhav:
> Could these constants be moved to ADAL? That would avoid doing a common release. (We just did a common release 5 days ago and the only delta are these constants)

I dont mind keeping the constants in ADAL, since all the other constant values were here I added them here.

PR #940, opened 2020-06-24
Changes to delete RT's if we get an invalid_grant with bad token while renewing tokens

kreedula:


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -300,6 +302,16 @@ protected void renewAccessToken(@NonNull final SilentTokenCommandParameters para
...
             acquireTokenSilentResult.setLocalAuthenticationResult(authenticationResult);
+        }else {
+            if (tokenResult.getErrorResponse() != null) {
+                final String errorCode = tokenResult.getErrorResponse().getError();
+                final String subErrorCode = tokenResult.getErrorResponse().getSubError();
+
+                if (errorCode.equals(INVALID_GRANT) && subErrorCode.equals(BAD_TOKEN)) {
+                    Logger.info(TAG, "Refresh token is invalid, deleting from cache");
+                    tokenCache.removeCredential(cacheRecord.getRefreshToken());
+                }
```
iambmelt, line 313: Can we log the contents of the `sub_error`?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -300,6 +302,16 @@ protected void renewAccessToken(@NonNull final SilentTokenCommandParameters para
 
             // Set the AuthenticationResult on the final result object
             acquireTokenSilentResult.setLocalAuthenticationResult(authenticationResult);
+        }else {
```
iambmelt, line 305: Nit formatter

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
kreedula: logging both error and suberror now

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -300,6 +302,17 @@ protected void renewAccessToken(@NonNull final SilentTokenCommandParameters para
...
+            if (tokenResult.getErrorResponse() != null) {
+                final String errorCode = tokenResult.getErrorResponse().getError();
+                final String subErrorCode = tokenResult.getErrorResponse().getSubError();
+                Logger.info(TAG, "Error: " + errorCode + " Suberror: " + subErrorCode);
+
+                if (errorCode.equals(INVALID_GRANT) && subErrorCode.equals(BAD_TOKEN)) {
+                    Logger.info(TAG, "Refresh token is invalid, deleting from cache");
+                    tokenCache.removeCredential(cacheRecord.getRefreshToken());
+                }
+            }
```
iambmelt, line 315: Can we add one more `else` case here where if the request isn't successful, but _also_ doesn't contain an error response, we log a warning message?

iambmelt: (Obviously, this condition _shouldn't_ ever happen)

kreedula: yep, should never happen, but no harm adding

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -300,6 +302,17 @@ protected void renewAccessToken(@NonNull final SilentTokenCommandParameters para
...
             acquireTokenSilentResult.setLocalAuthenticationResult(authenticationResult);
+        } else {
+            if (tokenResult.getErrorResponse() != null) {
+                final String errorCode = tokenResult.getErrorResponse().getError();
+                final String subErrorCode = tokenResult.getErrorResponse().getSubError();
+                Logger.info(TAG, "Error: " + errorCode + " Suberror: " + subErrorCode);
+
+                if (errorCode.equals(INVALID_GRANT) && subErrorCode.equals(BAD_TOKEN)) {
+                    Logger.info(TAG, "Refresh token is invalid, deleting from cache");
+                    tokenCache.removeCredential(cacheRecord.getRefreshToken());
```
iambmelt, line 313: Let's also check the return value here, should be ` boolean` if this removal was successful

PR #952, opened 2020-07-02
Additional utilities for working with test accounts.

AdamBJohnsonx:
There are two of these.  

One to get the AppInfo object, and since the temp accounts are only used in the "cloud" space I wasn't sure how much flexibility they needed.

One to get the tempUser for a given user type and allow a consumer to act on it.

### DIFF
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
@@ -226,12 +226,12 @@ public class LabUserHelper {
         return tempUser.getUpn();
     }
 
-    public static TempUser loadTempUserForTest(final String userType) {
+    public static TempUser loadTempUserForTest(final UserType userType) {
         LabAuthenticationHelper.getInstance().setupApiClientWithAccessToken();
         CreateTempUserApi createTempUserApi = new CreateTempUserApi();
 
         try {
-            return createTempUserApi.post(userType);
+            return createTempUserApi.post(userType.getValue());
         } catch (ApiException e) {
             throw new RuntimeException("Error retrieving lab user", e);
         }
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
@@ -226,12 +226,12 @@ public class LabUserHelper {
         return tempUser.getUpn();
     }
 
-    public static TempUser loadTempUserForTest(final UserType userType) {
+    public static TempUser loadTempUserForTest(final String userType) {
         LabAuthenticationHelper.getInstance().setupApiClientWithAccessToken();
         CreateTempUserApi createTempUserApi = new CreateTempUserApi();
 
         try {
-            return createTempUserApi.post(userType.getValue());
+            return createTempUserApi.post(userType);
         } catch (ApiException e) {
             throw new RuntimeException("Error retrieving lab user", e);
         }
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
@@ -71,6 +71,82 @@ public class LabUserHelper {
         return configInfos;
     }
 
+    public enum UserType {
+        CLOUD(LabConstants.UserType.CLOUD),
+        B2C(LabConstants.UserType.B2C),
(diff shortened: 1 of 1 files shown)


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -175,6 +192,16 @@ public static Credential getCredentials(LabUserQuery query) {
         return credential;
     }
 
+    public static AppInfo getAppInfo() {
+        LabAuthenticationHelper.getInstance().setupApiClientWithAccessToken();
+        AppApi api = new AppApi();
+        try {
+            return api.getAppByParam("cloud", "azurecloud", "azureadmultipleorgs", "yes", "yes").get(0);
```
shahzaibj, line 199: Would probably be better to accept each of these as method parameters instead of hard coding here

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -71,6 +71,82 @@
...
+        ;
+        String constant;
+        PublicClient(String constant) {
+            this.constant = constant;
+        }
+        String getValue() {
+            return constant;
+        }
+    }
+
```
shahzaibj, line 149: Why duplicate all of these as enums? Are the String constants in `LabConstants.java` not sufficient?

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -193,10 +269,17 @@ public static Credential getCredentials(LabUserQuery query) {
     }
 
     public static AppInfo getAppInfo() {
+            return getAppInfo(UserType.CLOUD, AzureEnvironment.AZURE_CLOUD, SignInAudience.AZURE_AD_MULTIPLE_ORGS,
+                    IsAdminConsented.YES, PublicClient.YES);
+   }
```
shahzaibj: Do we really need this method? Is the one below not sufficient?

shahzaibj: Do we really need this method? Is the one below not sufficient?

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
AdamBJohnsonx: I'm going to say no, they're not.  Because nothing stops you from putting the wrong string, or nay, any string into the method arguments nothing saves you from transposing the arguments or provides a path from the method to what the potential values are unless you happen to know that there's a class called LabConstants.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
AdamBJohnsonx: This is just convenience, since these are supposed to be the "default" values according to the swagger document.  I could theoretically just send a to Api/App and get these, and that's what I should probably change this to do.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
shahzaibj: Makes sense

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
shahzaibj: Okay I would suggest renaming the method then to something like `getDefaultAppInfo`

AdamBJohnsonx: Yeah, it's interesting the mismatch between what the service does and what the AppApi class allows.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -146,6 +226,19 @@ public static String loadTempUser(final String userType) {
         return tempUser.getUpn();
     }
 
+    public static TempUser loadTempUserForTest(final String userType, Consumer<TempUser> consumer) {
```
shahzaibj, line 229: Synced offline...let's make sure if this `Consumer` is needed...if not, then let's remove it

AdamBJohnsonx: If I'm returning the reference, I don't need it.

PR #961, opened 2020-07-14
Fix MSAL issue 1096

shahzaibj:
Partial Fix for https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1096

The other half of this fix is located here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1097

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/RemoveCurrentAccountCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/RemoveCurrentAccountCommand.java
@@ -58,8 +58,7 @@ public class RemoveCurrentAccountCommand extends BaseCommand<Boolean> {
     public Boolean execute() throws Exception {
         final String methodName = ":execute";
 
-        for (int ii = 0; ii < getControllers().size(); ii++) {
-            final BaseController controller = getControllers().get(ii);
+        for (final BaseController controller : getControllers()) {
             com.microsoft.identity.common.internal.logging.Logger.verbose(
                     TAG + methodName,
                     "Executing with controller: "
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/commands/RemoveCurrentAccountCommand.java
```diff
@@ -54,7 +56,24 @@ public RemoveCurrentAccountCommand(@NonNull RemoveAccountCommandParameters param
 
     @Override
     public Boolean execute() throws Exception {
-        return getDefaultController().removeCurrentAccount((RemoveAccountCommandParameters) getParameters());
+        final String methodName = ":execute";
+
+        for (int ii = 0; ii < getControllers().size(); ii++) {
```
iambmelt, line 61: Minor code golf: could use a foreach if you don't need the index's value

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/961/commits/a683e653c33ce9a961675e44b4177a8e130ec589

common/src/main/java/com/microsoft/identity/common/internal/commands/RemoveCurrentAccountCommand.java
```diff
@@ -54,7 +56,23 @@ public RemoveCurrentAccountCommand(@NonNull RemoveAccountCommandParameters param
...
+            com.microsoft.identity.common.internal.logging.Logger.verbose(
+                    TAG + methodName,
+                    "Executing with controller: "
+                            + controller.getClass().getSimpleName()
+            );
+
+            final boolean removed = controller.removeCurrentAccount((RemoveAccountCommandParameters) getParameters());
+
+            if (removed) {
+                return true;
```
AdamBJohnsonx, line 71: Do we not want to remove it from the rest of them, if it happens to exist there?

PR #963, opened 2020-07-16
Detect Broker Process during encryption/decryption

rpdome:
**Why making this fix?**
Teams Desk phone found an issue where it was unable to sign in when they're trying to change CP account. (logs [here](https://powerlift.acompli.net/#/incidents/9dfa7f87-60bb-447e-bf88-289456700597))
The root cause is that 
1. our StorageHelper would keep using Broker's secret key after it's loaded. (technical debt that I was trying to fix in #746 , which I later had to drop).
2. In this case, CP invokes WPJ prior to its sign in
3. its local ADAL process unable to decrypt the value. 
4. RT was wiped
```
2020-07-15T11:56:28.1690000	INFO	MSALCommonLoggerCallback	 7335	00230	MSALCommon: Tag: MsalOAuth2TokenCachesetSingleSignOnState, Message:  [2020-07-15 11:56:28 - {"thread_id":"253","correlation_id":"ad65f8c8-1612-4415-b9ea-19d34e57d9ab"}] Refresh tokens removed: [0] Android 25
```
5. (I suspect) that AT is returned, but RT is not, so CP went ahead and perform WPJ without RT.
6. Teams Phone can't silently sign in as BRT does not exist.

**Proposed fix**

This fix is much simpler than #746. We use the key whenever we're in the broker auth process.
Not sure why I couldn't think of this at the time :/

(getProcessName() is a courtesy from CP team).

**How I verified**
1. Put a debugger on both :auth and non-auth process
2. Perform Join() from BrokerHost. Signs in, isBrokerProcess function returns false. (Local ADAL is used to acquire token).
3. Launch MSAL and trigger join flow. isBrokerProcess function returns true (Broker is used).

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -415,7 +415,7 @@ public class StorageHelper implements IStorageHelper {
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
 
         if (encryptionType == EncryptionType.USER_DEFINED) {
-            if (ProcessUtil.isAuthProcess(mContext)){
+            if (ProcessUtil.isBrokerProcess(mContext)){
                 if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
                     keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
                     keyTypeList.add(KeyType.LEGACY_AUTHENTICATOR_APP_KEY);
@@ -515,7 +515,7 @@ public class StorageHelper implements IStorageHelper {
         }
 
         // The current app runtime is the broker; load its secret key.
-        if (!sShouldEncryptWithKeyStoreKey && ProcessUtil.isAuthProcess(mContext)) {
+        if (!sShouldEncryptWithKeyStoreKey && ProcessUtil.isBrokerProcess(mContext)) {
             // Try to read keystore key - to verify how often this is invoked before the migration is done.
             // TODO: remove this whole try-catch clause once the experiment is done.
             if (mTelemetryCallback != null) {
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -422,6 +422,8 @@ public class StorageHelper implements IStorageHelper {
                 } else if (AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
                     keyTypeList.add(KeyType.LEGACY_AUTHENTICATOR_APP_KEY);
                     keyTypeList.add(KeyType.LEGACY_COMPANY_PORTAL_KEY);
+                } else {
+                    throw new IllegalStateException("Unexpected Broker package name.");
                 }
             } else {
                 keyTypeList.add(KeyType.ADAL_USER_DEFINED_KEY);
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -38,6 +38,7 @@ import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.exception.ErrorStrings;
 import com.microsoft.identity.common.internal.logging.Logger;
+import com.microsoft.identity.common.internal.util.ProcessUtil;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+import androidx.annotation.NonNull;
+
+import java.util.List;
+
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
+
+public class ProcessUtil {
+
```
iambmelt, line 37: Since this is a new utility class, you could put in a private constructor so it can't be instantiated

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,65 @@
...
+        final String processName = getProcessName(context);
+
+        final String authAppAuthProcess = AZURE_AUTHENTICATOR_APP_PACKAGE_NAME + ":auth";
+        final String cpAuthProcess = COMPANY_PORTAL_APP_PACKAGE_NAME + ":auth";
+
+        return authAppAuthProcess.equalsIgnoreCase(processName) ||
+                cpAuthProcess.equalsIgnoreCase(processName);
+    }
+
+    public static String getProcessName(final Context context) {
```
iambmelt, line 51: Javadoc? Nullability annots?

iambmelt: I see this can return `null`

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,65 @@
...
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+import java.util.List;
+
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
+
+public class ProcessUtil {
```
iambmelt, line 36: Low hanging fruit: Class description for javadoc

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
AdamBJohnsonx: public?

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Returns true if the calling app is the auth process.
+     */
+    public static boolean isBrokerProcess(@NonNull final Context context) {
+        final String processName = getProcessName(context);
+
+        final String authAppAuthProcess = AZURE_AUTHENTICATOR_APP_PACKAGE_NAME + ":auth";
+        final String cpAuthProcess = COMPANY_PORTAL_APP_PACKAGE_NAME + ":auth";
```
shahzaibj, line 50: I think we have a method called as `getCurrentActiveBrokerPackageName`, we could use that....it is also fine the way it is too. I'll leave it up to you

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,73 @@
...
+
+        return authAppAuthProcess.equalsIgnoreCase(processName) ||
+                cpAuthProcess.equalsIgnoreCase(processName);
+    }
+
+    /**
+     * Returns the running process name.
+     */
```
iambmelt, line 58: Can/should be declared as

```java
@Nullable
public static String getProcessName(@NonNull final Context context) {
    // ...
}
```

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -414,14 +415,16 @@ public EncryptionType getEncryptionType(@NonNull final String data) throws Unsup
         EncryptionType encryptionType = getEncryptionType(encryptedBlob);
 
         if (encryptionType == EncryptionType.USER_DEFINED) {
-            if (AuthenticationSettings.INSTANCE.getSecretKeyData() != null) {
+            if (ProcessUtil.isBrokerProcess(mContext)){
+                if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
```
AdamBJohnsonx, line 419: There's some odd coupling in functionality here.  You probably want an else branch here to throw an IllegalStateException or panic or something for the case where the process name doesn't match the package name.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: I think we have that kind of function all over the place 😆 . Eventually we should consolidate, but let's keep this PR minimum for now.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+import android.app.ActivityManager;
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+import java.util.List;
+
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
```
AdamBJohnsonx, line 35: Should we take this as an argument instead of looking for both of them?

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
AdamBJohnsonx: Yeah, you know the package name.  If you pass that into your process util you won't need to look for both of them and have the confusing case where you found one in the package name and the other in the process name.

AdamBJohnsonx: Hm..  Either - or.  I see you're using it elsewhere, and I missed that on the first time through.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: Was thinking of that, but decided against it as other app might create their own :auth process and invokes ADAL in it.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: We might have to use this in MSAL/ADAL/Broker in the future, so I intentionally leave it public.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
AdamBJohnsonx: It's moot if your only use was in line 418/419, but I see another one.  You can consider this moot.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
AdamBJohnsonx: You can always make it public later.  You can never make it private later.

rpdome: just tried. Seems like they're in different packages, so I can't make it package-private.

rpdome: nvm, looking at the wrong function. just made it private.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: Even if it's only used in line 418/419, the customer could run ADAL in like... com.myapp.test:auth 
In that case it's totally valid, and we want them to use KeyType.ADAL_USER_DEFINED_KEY.

PR #964, opened 2020-07-16
Adding "junk data" tests to the MSAL Cache

iambmelt:
These are tests I wrote to verify the cache skips/preserves unexpected values; this is a setup for a future conversation around potentially storing data from OneAuth in this cache.

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
@@ -123,6 +123,8 @@ public class MsalOAuth2TokenCacheTest extends AndroidSecretKeyEnabledHelper {
     AccountCredentialTestBundle defaultTestBundleV2;
 
     IAccountCredentialCache accountCredentialCache;
+    public static final String JUNK_KEY = "9ac15f53-27ad-471a-ad57-033cdacb0ee9";
+    public static final String JUNK_VALUE = "OWFjMTVmNTMtMjdhZC00NzFhLWFkNTctMDMzY2RhY2IwZWU5";
 
     public static class AccountCredentialTestBundle {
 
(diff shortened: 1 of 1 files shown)


common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
```diff
@@ -344,6 +389,28 @@ public void saveTokensWithAggregationSingleEntry() throws ClientException {
...
+        );
+
+        assertEquals(1, result.size());
+
+        final ICacheRecord entry = result.get(0);
+
+        assertNotNull(entry.getAccount());
+        assertNotNull(entry.getIdToken());
+        assertNotNull(entry.getAccessToken());
+        assertNotNull(entry.getRefreshToken());
```
AdamBJohnsonx, line 408: Should we check that these have some particular value?

common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
```diff
@@ -294,6 +297,48 @@ public void saveTokens() throws Exception {
...
+
+        for (final Credential credential : credentials) {
+            if (credential.getCredentialType().equalsIgnoreCase(CredentialType.AccessToken.name())) {
+                ats.add(credential);
+            } else if (credential.getCredentialType().equalsIgnoreCase(CredentialType.RefreshToken.name())) {
+                rts.add(credential);
+            } else if (credential.getCredentialType().equalsIgnoreCase(CredentialType.IdToken.name())) {
+                ids.add(credential);
+            } else {
+                fail();
```
AdamBJohnsonx, line 330: nit: if CredentialType is an enum, then you could replace with a switch.  Also, I think you can provide a string in the fail() method saying why it is failing and assist the poor person viewing the test failure to fix it.

common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
iambmelt: Refined in 895c3d2c36748b8a7c40d6664bd472f11d22bf52

common/src/androidTest/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
iambmelt: Applied changes in 8d37efeb340eb3cfe438317e799fa5b3d4320cdb, c6044e92dab13397356bce8bdd4a64ccd673704a

iambmelt:
> Q: Should we also be able to write data there after it is populated and not affect the values we have?

Can add an additional test for this

iambmelt:
> > Q: Should we also be able to write data there after it is populated and not affect the values we have?
> 
> Can add an additional test for this

Added [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/964/commits/ee39d4c1419cae8ad70142799bc215e4aeec91b7#diff-1c70fbcb6294d45c130a60470b33b716R327)

PR #966, opened 2020-07-20
DCF API-Side PR

t-fadura:
First pull request in Common for DCF, consists of api-side code and some testing/telemetry constants.

Added `DeviceCodeFlowCommandCallback` to Common
Added `DeviceCodeFlowCommand` to Common
Added `DeviceCodeFlowCommandParameters` to Common
Added `DEVICE_CODE_FLOW_CALLBACK` to PublicApiId in Common for Telemetry
Added config file path to `TestConstants` in Common

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
@@ -1,0 +1,29 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.commands;
+
+import androidx.annotation.NonNull;
+
+public interface DCFCommandCallback<T, U> extends CommandCallback<T, U> {
+    void getUserCode(@NonNull String vUri, @NonNull String user_code, @NonNull String message);
+}
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
@@ -35,11 +35,15 @@ public class DeviceCodeFlowCommand extends TokenCommand {
      * Relevant error codes and messages go here
      */
 
+    public static DCFCommandCallback dcfCallback;
+
     public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
                                  @NonNull BaseController controller,
                                  @NonNull CommandCallback callback,
                                  @NonNull String publicApiId) {
         super(parameters, controller, callback, publicApiId);
+
+        dcfCallback = (DCFCommandCallback) callback;
     }
 
     @Override
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
@@ -30,12 +30,19 @@ import com.microsoft.identity.common.internal.commands.parameters.DeviceCodeFlow
 import com.microsoft.identity.common.internal.controllers.BaseController;
 import com.microsoft.identity.common.internal.result.AcquireTokenResult;
 
+/**
+ * This command is used to execute the device code flow protocol.
+ * Takes in a parameters object containing the  desired access scopes along and returns
+ * a token result.
+ * Class also includes some pre-defined error codes and messages to be used in
+ * exception handling.
+ */
 public class DeviceCodeFlowCommand extends TokenCommand {
     /**
      * Relevant error codes and messages go here
      */
 
-    public static DCFCommandCallback dcfCallback;
+    public static DeviceCodeFlowCommandCallback sDeviceCodeFlowCallback;
 
     public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
                                  @NonNull BaseController controller,
@@ -43,7 +50,8 @@ public class DeviceCodeFlowCommand extends TokenCommand {
                                  @NonNull String publicApiId) {
         super(parameters, controller, callback, publicApiId);
 
-        dcfCallback = (DCFCommandCallback) callback;
+        // Fetch callback object
+        sDeviceCodeFlowCallback = (DeviceCodeFlowCommandCallback) callback;
     }
 
     @Override
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
@@ -38,10 +38,6 @@ import com.microsoft.identity.common.internal.result.AcquireTokenResult;
  * exception handling.
  */
 public class DeviceCodeFlowCommand extends TokenCommand {
-    /**
-     * Relevant error codes and messages go here
-     */
-
     public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
                                  @NonNull BaseController controller,
                                  @NonNull DeviceCodeFlowCommandCallback callback,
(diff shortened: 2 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/DeviceCodeFlowCommandParameters.java
```diff
@@ -0,0 +1,33 @@
...
+package com.microsoft.identity.common.internal.commands.parameters;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder(toBuilder = true)
+public class DeviceCodeFlowCommandParameters extends TokenCommandParameters {
```
shahzaibj, line 32: Do we need this class? Doesn't seem to have anything

t-fadura: I decided to do it this way for two reasons:

1. `getAuthorizationRequest()` in `BaseController` takes in a `TokenCommandParameters` object as parameter, so I can't create a command parameters class for DCF that only extends `CommandParameters` and add the needed fields.

2. I looked through the code and it seemed like there weren't any commands that received a `TokenCommandParameters` object from PublicClientApplication, so I figured it would make more sense to create a new class that specifically maps to DCF. It also allows future developers to add DCF-specific parameters here if they want to implement any extra functionality.

common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
```diff
@@ -0,0 +1,29 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.commands;
+
+import androidx.annotation.NonNull;
+
+public interface DCFCommandCallback<T, U> extends CommandCallback<T, U> {
```
shoatman, line 27: I'd probably just spell out the name here for clarity.  If not, casing for acronyms is the same as words.  https://source.android.com/setup/contribute/code-style.  I'd also add some javadoc here which also helps with clarity.  You spell it out below ;-)

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+import android.content.Intent;
+
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.common.internal.commands.parameters.DeviceCodeFlowCommandParameters;
+import com.microsoft.identity.common.internal.controllers.BaseController;
+import com.microsoft.identity.common.internal.result.AcquireTokenResult;
+
+public class DeviceCodeFlowCommand extends TokenCommand {
```
shoatman, line 33: nit: javdoc

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Relevant error codes and messages go here
+     */
+
+    public static DCFCommandCallback dcfCallback;
```
shoatman, line 38: take a look at field naming conventions: https://source.android.com/setup/contribute/code-style#follow-field-naming-conventions.  I'd also consider spelling the variable name out as well.

common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
t-fadura: Fair point, I missed that

common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
```diff
@@ -0,0 +1,29 @@
...
+    void getUserCode(@NonNull String vUri, @NonNull String user_code, @NonNull String message);
```
shahzaibj, line 28: nit: rename to `userCode`

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
shahzaibj, line 38: Why is this static? More importantly we probably don't even need this field. Why is it here?

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
+                                 @NonNull BaseController controller,
+                                 @NonNull CommandCallback callback,
```
shahzaibj, line 42: This should be `DcfCommandCallback`

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,64 @@
...
+                                 @NonNull String publicApiId) {
+        super(parameters, controller, callback, publicApiId);
+
+        dcfCallback = (DCFCommandCallback) callback;
```
shahzaibj, line 46: This should be taken care of by the parent

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
```diff
@@ -30,6 +30,7 @@
     //==============================================================================================
     public static final String PCA_ACQUIRE_TOKEN_SILENT_WITH_PARAMETERS = "21";
     public static final String PCA_ACQUIRE_TOKEN_SILENT_ASYNC_WITH_PARAMETERS = "22";
+    public static final String DEVICE_CODE_FLOW_CALLBACK = "30";
```
shahzaibj, line 33: We'll need to double check if we can use this number.....I'll check that and let you know

shahzaibj: We can't use this....it is already used by iOS and this range is actually for silent flows. The range for Device Code Flow public API is 600-699. The following are already in use:


Java | AcquireToken(DeviceCodeParameters) | 620
-- | -- | --
Python | acquire_token_by_device_flow | 622
.net | AcquireTokenByDeviceCode | 1008
Node | msal-node/acquireTokenByDeviceCode | 671

Not quite sure why .net is using a number outside the range, but I'm thinking we can pick anything between 600-699. May be 650? I'll let you decide what you want 😁

We have a doc (Excel sheet) with all of these public API ids across platforms. If you don't have that and/or would like to see it, DM me and I will send that over to you via Teams.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: I added this field for testing with a shadow class. Making it public made it possible for me to fetch the callback object outside of the command class. I essentially wanted to be able to call the callback from within my shadow class.

common/src/main/java/com/microsoft/identity/common/internal/commands/DCFCommandCallback.java
t-fadura: I've taken into account the suggestions you made in the design doc PR (including this one). These will be reflected in a commit i'll be pushing soon.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
t-fadura: Ok! I'll change it to 650. I'll ping you on teams, too.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: I'm not sure what the better way to do this is. If I change the type of the parameter to DeviceCodeFlowCommandCallback, would that be enough?

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: Ended up deleting the static field.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,67 @@
...
+ * This command is used to execute the device code flow protocol.
+ * Takes in a parameters object containing the  desired access scopes along and returns
+ * a token result.
+ * Class also includes some pre-defined error codes and messages to be used in
+ * exception handling.
+ */
+public class DeviceCodeFlowCommand extends TokenCommand {
+    /**
+     * Relevant error codes and messages go here
+     */
```
shahzaibj, line 43: I believe this is not needed

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -0,0 +1,67 @@
...
+    public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
+                                 @NonNull BaseController controller,
+                                 @NonNull DeviceCodeFlowCommandCallback callback,
+                                 @NonNull String publicApiId) {
+        super(parameters, controller, callback, publicApiId);
+    }
+
+    @Override
+    public AcquireTokenResult execute() throws Exception {
+        // TODO
```
shahzaibj, line 54: can add more details as to when this TODO will resolve

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/PublicApiId.java
```diff
@@ -105,4 +105,12 @@
     public static final String MULTIPLE_ACCOUNT_PCA_REMOVE_ACCOUNT_WITH_ACCOUNT_CALLBACK = "930";
 
     //endregion
+
+    // region Device Code Flow
+
+    // PublicClientApplication
+    //==============================================================================================
+    public static final String DEVICE_CODE_FLOW_CALLBACK = "650";
```
shahzaibj, line 113: nit: rename to `DEVICE_CODE_FLOW_WITH_CALLBACK` to follow the same pattern as other consants

PR #974, opened 2020-07-23
Adding Locale.Root to toLowerCase calls

shoatman:
Fix issue with Turkish language: https://developer.android.com/reference/java/lang/String#toLowerCase(java.util.Locale)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
@@ -37,6 +37,7 @@ import com.microsoft.identity.common.internal.logging.Logger;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 
 import static com.microsoft.identity.common.internal.controllers.BaseController.DEFAULT_SCOPES;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
```diff
@@ -235,11 +236,11 @@ static boolean targetsIntersect(@NonNull final String targetToMatch,
 
         // Add the array values to these sets, lowercasing them
         for (final String target : targetToMatchArray) {
-            soughtTargetSet.add(target.toLowerCase());
+            soughtTargetSet.add(target.toLowerCase(Locale.ROOT));
```
shoatman, line 239: @iambmelt - Am a bit concerned about using ROOT here instead of US which appears to be what we did for the rest of caching.  What are your thoughts?

shoatman:
Assuming that it's not possible in Azure AD to define scopes using the Turkish language.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -59,7 +61,7 @@ public String toString() {
             return null;
         }
 
-        return this.name().toLowerCase();
+        return this.name().toLowerCase(Locale.ROOT);
```
shoatman, line 64: This was the actual reported bug.

common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
AdamBJohnsonx: https://cs.android.com/android/platform/superproject/+/master:libcore/libart/src/main/java/java/lang/CaseMapper.java;drc=master;l=45 it looks like there are a few cases where the Locale matters - Turkish, Lithuanian, and Azerbaijani, etc.  Other than those and a few special cases it shouldn't matter.

PR #975, opened 2020-07-23
DCF PR: Adding Command/Controller code

t-fadura:
In this PR, moved on from API-side implementation and implemented code on the command and controller ends.

Implemented code in `DeviceCodeFlowCommand` in Common
Added expected errors in `DeviceCodeFlowCommand` in Common
Added abstract 'deviceCodeFlowAuthRequest()` and `acquireDeviceCodeFlowToken()` to `BaseController` in Common
Added DCF fields to `MicrosoftAuthorizationResponse' in Common

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
@@ -28,6 +28,8 @@ import androidx.annotation.NonNull;
 
 import com.microsoft.identity.common.internal.commands.parameters.DeviceCodeFlowCommandParameters;
 import com.microsoft.identity.common.internal.controllers.BaseController;
+import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsAuthorizationResponse;
+import com.microsoft.identity.common.internal.providers.oauth2.AuthorizationResult;
 import com.microsoft.identity.common.internal.result.AcquireTokenResult;
 
 /**
@@ -38,6 +40,16 @@ import com.microsoft.identity.common.internal.result.AcquireTokenResult;
  * exception handling.
  */
 public class DeviceCodeFlowCommand extends TokenCommand {
+
+    public final static String AUTH_DECLINED_CODE = "authorization_declined";
+    public final static String AUTH_DECLINED_MSG = "The end user has denied the authorization request. Re-run the Device Code Flow Protocol with another user.";
+    public final static String EXPIRED_TOKEN_CODE = "expired_token";
+    public final static String EXPIRED_TOKEN_MSG = "The token has expired, therefore authentication is no longer possible with this flow attempt. Re-run the Device Code Flow Protocol to try again.";
+    // Add more errors
+
+    // Use this message for when DCF fails with an error code that doesn't match any of the codes above
+    public final static String DEFAULT_ERROR_MSG = "Device Code Flow has failed with an unexpected error. The error code shown was received from the authorization result.";
+
     public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters,
                                  @NonNull BaseController controller,
                                  @NonNull DeviceCodeFlowCommandCallback callback,
@@ -47,8 +59,30 @@ public class DeviceCodeFlowCommand extends TokenCommand {
 
     @Override
     public AcquireTokenResult execute() throws Exception {
-        // TODO: Implement execute body in next PR, when DCF methods have been declared/defined in LocalMsalController
-        return null;
+
+        // Get the controller used to execute the command
+        final BaseController controller = getDefaultController();
+
+        // Fetch the parameters
+        final DeviceCodeFlowCommandParameters commandParameters = (DeviceCodeFlowCommandParameters) getParameters();
+
+        // Call deviceCodeFlowAuthRequest to get authorization result (Part 1 of DCF)
+        final AuthorizationResult authorizationResult = controller.deviceCodeFlowAuthRequest(commandParameters);
+
+        // Fetch the authorization response
+        final MicrosoftStsAuthorizationResponse authorizationResponse =
+                (MicrosoftStsAuthorizationResponse) authorizationResult.getAuthorizationResponse();
+
+        // Communicate with user app and provide authentication information
+        final DeviceCodeFlowCommandCallback deviceCodeFlowCommandCallback = (DeviceCodeFlowCommandCallback) getCallback();
+        deviceCodeFlowCommandCallback.onUserCodeReceived(
+                authorizationResponse.getVerificationUri(),
+                authorizationResponse.getUserCode(),
+                authorizationResponse.getMessage()
+        );
+
+        // Call acquireDeviceCodeFlowToken to get token result (Part 2 of DCF)
+        return controller.acquireDeviceCodeFlowToken(authorizationResult, commandParameters);
     }
 
     @Override
@@ -58,6 +92,6 @@ public class DeviceCodeFlowCommand extends TokenCommand {
 
     @Override
     void notify(int requestCode, int resultCode, Intent data) {
-        // TODO: Same as above
+        getDefaultController().completeAcquireToken(requestCode, resultCode, data);
     }
 }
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -38,6 +40,16 @@
  * exception handling.
  */
 public class DeviceCodeFlowCommand extends TokenCommand {
+
+    public final static String AUTH_DECLINED_CODE = "authorization_declined";
+    public final static String AUTH_DECLINED_MSG = "The end user has denied the authorization request. Re-run the Device Code Flow Protocol with another user.";
```
rpdome, line 45: Double check these error strings with @hamiltonha

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -143,6 +144,11 @@ public abstract boolean getDeviceMode(final CommandParameters parameters)
     public abstract boolean removeCurrentAccount(final RemoveAccountCommandParameters parameters)
             throws Exception;
 
+    public abstract AuthorizationResult deviceCodeFlowAuthRequest(final DeviceCodeFlowCommandParameters parameters)
+            throws Exception;
+    public abstract AcquireTokenResult acquireDeviceCodeFlowToken(final AuthorizationResult authorizationResult, DeviceCodeFlowCommandParameters commandParameters)
```
rpdome, line 149: nit: add an extra line

rpdome:
Test case is failing.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -38,6 +40,16 @@
...
+    public final static String EXPIRED_TOKEN_CODE = "expired_token";
+    public final static String EXPIRED_TOKEN_MSG = "The token has expired, therefore authentication is no longer possible with this flow attempt. Re-run the Device Code Flow Protocol to try again.";
+    // Add more errors
+
+    // Use this message for when DCF fails with an error code that doesn't match any of the codes above
+    public final static String DEFAULT_ERROR_MSG = "Device Code Flow has failed with an unexpected error. The error code shown was received from the authorization result.";
```
shahzaibj, line 51: I don't think these should live here. They should instead be in `ErrorStrings.java`

shahzaibj: Also following the same pattern in that class, let's use complete variable names

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -47,8 +59,30 @@ public DeviceCodeFlowCommand(@NonNull DeviceCodeFlowCommandParameters parameters
 
     @Override
     public AcquireTokenResult execute() throws Exception {
-        // TODO: Implement execute body in next PR, when DCF methods have been declared/defined in LocalMsalController
-        return null;
+
+        // Get the controller used to execute the command
+        final BaseController controller = getDefaultController();
```
shahzaibj, line 64: This is wrong as stated the CR on an earlier PR. This would use the `BrokerMsalController` if a broker is present on the device. 

This should instead be `getControllers()` and loop over them to execute for each controller. For an example, see `SilentTokenCommand.java`

rpdome: This is not supported in broker. so either we find a way to inject LocalMsalController as the default controller elsewhere, or we simply just hard code it here and add some comment.

shahzaibj: Then we should use `getControllers().get(0)` - The command object has at least one controller to begin with. In this case, that is the `LocalMsalController` as that is being supplied to the command during creation via the method in PCA

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
iambmelt: +1 on this feedback, let's organize these constants into `ErrorStrings`

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationResponse.java
```diff
@@ -85,6 +105,15 @@ public void setCorrelationId(final String correlationId) {
 
     public String getSessionState() { return mSessionState;}
 
+    public String getDeviceCode() { return mDeviceCode;}
```
iambmelt, line 108: Most of these properties are going to be unfamiliar to us, as we're just now adding support for DCF. Despite the other getters not having javadoc in this class, can I request we add it for these new methods? I think it's is particularly important to call out that these getters and their values will return `null` in non-DCF cases, and that these values may be applicable only in DC flows

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: The error codes or the error messages?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftAuthorizationResponse.java
t-fadura: done!

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: @shahzaibj `getDefaultController()` here actually points to a method in `BaseController` that simply calls `controllers.get(0)` (`controllers` being the object that would be returned by `getControllers()`), so it's the same code either way I think. And yes, since the command object was created with a `LocalMsalController` in `PublicClientApplication`, `LocalMsalController` should be the "default" controller.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: Done!

t-fadura:
New version up for review, fixed suggestions and moved errors. Also, checks now passing.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
shahzaibj: @t-fadura Okay, I see. We have multiple methods called `getDefaultController` and each of them do different things. Thanks for pointing out.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResponse.java
```diff
@@ -24,6 +24,7 @@
 
 import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftAuthorizationResponse;
 
+import java.net.URL;
```
AdamBJohnsonx, line 27: nit: What requires this import?

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
rpdome: error messages.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResponse.java
t-fadura: It was left here from previous code by mistake, will remove

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: Ok, I'll be sure to sync up with @hamiltonha as I add more error messages in the next PR.

PR #978, opened 2020-07-24
Some common utilities to avoid duplication.

AdamBJohnsonx:
- Since we can't count on Objects.equals, reimplemenit it for now.
- We do a lot of Thread.sleep() - remove some boilerplate.


common/src/main/java/com/microsoft/identity/common/Utils.java
```diff
@@ -0,0 +1,49 @@
+package com.microsoft.identity.common;
```
iambmelt, line 1: Missing license info in this file

AdamBJohnsonx: Yup.

AdamBJohnsonx:
> We'll need some license info, but LGTM

Actually, there is a far better place for this, I'm moving it there.

common/src/main/java/com/microsoft/identity/common/internal/util/ObjectUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * This is a local implementation of Objects.equals.  It is a null-safe equals execution.
+     * This should be removed if we get to an API version that has Objects.equals.
+     *
+     * @param o1 the first object
+     * @param o2 the second object
+     * @return true if the objects are both null or if they are both non-null and o1.equals(o2).
+     */
+    public static boolean equals(@Nullable final Object o1, @Nullable final Object o2) {
+        return (o1 == null ^ o2 == null) || (o1 != null && o1.equals(o2));
```
AdamBJohnsonx, line 48: This is wrong...

AdamBJohnsonx: I've fixed it.  Lesson, check yourself.  I originally wrote not-equals, got feedback to extract it, agreed and went ahead.  But it wasn't equals I was getting.

common/src/main/java/com/microsoft/identity/common/internal/util/ObjectUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * This is a local implementation of Objects.equals.  It is a null-safe equals execution.
+     * This should be removed if we get to an API version that has Objects.equals.
+     *
+     * @param o1 the first object
+     * @param o2 the second object
+     * @return true if the objects are both null or if they are both non-null and o1.equals(o2).
+     */
+    public static boolean equals(@Nullable final Object o1, @Nullable final Object o2) {
+        return (o1 == o2) || (o1 != null && o1.equals(o2));
```
AdamBJohnsonx, line 48: I feel pretty safe about this now because Android Studio offers to replace this with Objects.equals().

PR #983, opened 2020-07-29
Device Code Flow Final PR

t-fadura:
Final major Common PR for the Device Code Flow implementation.

Added `invalid_grant` for DCF to `ErrorStrings`
Added some logging to `DeviceCodeFlowCommand`
Adjusted `createTokenRequest()` in `MicrosoftStsOAuth2Strategy` for DCF
Added 'mDeviceCode' to `MicrosoftTokenRequest'
Implemented `getDeviceCode()` in `OAuth2Strategy`
Added `DEVICE_CODE` to `GrantTypes` in `TokenRequest`
Added Telemtry codes in `TelemetryEventStrings`

`LocalMsalController` was recently moved to Common, so included following changes:
Implemented `deviceCodeFlowAuthRequest` and `acquireDeviceCodeFlowToken` in `LocalMsalController`
Added Exception handling in `LocalMsalController`

Should be merged before MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1112


common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -422,9 +422,16 @@ private ErrorStrings() {
     public final static String DEVICE_CODE_FLOW_EXPIRED_TOKEN_CODE = "expired_token";
     public final static String DEVICE_CODE_FLOW_EXPIRED_TOKEN_MESSAGE = "The token has expired, therefore authentication is no longer possible with this flow attempt. Re-run the Device Code Flow Protocol to try again.";
 
+    /**
+     * Device Code Flow only.
+     * The token was polled again after it was already received.
+     */
+    public final static String DEVICE_CODE_FLOW_INVALID_GRANT_CODE = "invalid_grant";
```
iambmelt, line 429: We _should_ have a constant for this somewhere already, let me track it down

iambmelt: Let's use this value here:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java#L1414

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -422,9 +422,16 @@ private ErrorStrings() {
...
+    public final static String DEVICE_CODE_FLOW_INVALID_GRANT_MESSAGE = "The token for this device code has already been redeemed. To receive another access token, please re-run the Device Code Flow protocol.";
```
iambmelt, line 430: Will this be shown to users? If yes, we'll need to have this localized

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftTokenRequest.java
```diff
@@ -180,4 +185,12 @@ public String getMicrosoftEnrollmentId() {
     public void setMicrosoftEnrollmentId(String microsoftEnrollmentId) {
         this.mMicrosoftEnrollmentId = microsoftEnrollmentId;
     }
+
+    public String getDeviceCode() {
+        return mDeviceCode;
+    }
+
+    public void setDeviceCode(String deviceCode) {
```
iambmelt, line 193: Minor nit: input variable could be `final`

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftTokenRequest.java
```diff
@@ -180,4 +185,12 @@ public String getMicrosoftEnrollmentId() {
     public void setMicrosoftEnrollmentId(String microsoftEnrollmentId) {
         this.mMicrosoftEnrollmentId = microsoftEnrollmentId;
     }
+
+    public String getDeviceCode() {
```
iambmelt, line 189: Let's declare this value `@Nullable` (as looks to be unassigned in auth code flow)

```java
@Nullable
public String getDeviceCode() {
    // ...
}
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
         mAuthorizationEndpoint = authorizationEndpoint;
     }
 
+    public AuthorizationResult getDeviceCode(MicrosoftStsAuthorizationRequest authorizationRequest, @Nullable String authorityUrl) throws IOException {
```
iambmelt, line 248:
```suggestion
    public AuthorizationResult getDeviceCode(@NonNull final MicrosoftStsAuthorizationRequest authorizationRequest, @Nullable final String authorityUrl) throws IOException {
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+            // Set up Device Code Flow authorization url
+            urlBody = "https://login.microsoftonline.com/" + tenant + "/oauth2/v2.0/devicecode";
+        }
+        else {
+            urlBody = authorityUrl;
+        }
+
+        URL url = new URL(urlBody);
+
+        final HttpURLConnection con = (HttpsURLConnection) url.openConnection();
```
iambmelt, line 267: Networking code shouldn't be at this level of abstraction; instead delegate to existing implementation in `HttpRequest.java`

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+
+        // Set up connection
+        String urlBody;
+
+        if (authorityUrl == null) {
+            // Fetch the tenant from the authority
+            String tenant = authorizationRequest.getAuthority().getPath().split("/")[1];
+
+            // Set up Device Code Flow authorization url
+            urlBody = "https://login.microsoftonline.com/" + tenant + "/oauth2/v2.0/devicecode";
```
iambmelt, line 259: This is Microsoft specific, this should move _down_ the object inheritance hierarchy and into `MicrosoftStsOAuth2Strategy`

iambmelt: This URL should be sourced from the `MicrosoftStsOAuth2Configuration` -- something like `getDeviceCodeEndpoint()`

iambmelt: I would expect that the URL for this endpoint should be available at:
`https://login.microsoftonline.com/common/.well-known/openid-configuration`
via `device_authorization_endpoint` but it's not....

We should reach out to server team to determine why this is missing; I can help facilitate that communication

/cc @shoatman

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
 
+    public AuthorizationResult getDeviceCode(MicrosoftStsAuthorizationRequest authorizationRequest, @Nullable String authorityUrl) throws IOException {
+        final String methodName = ":getDeviceCode";
+
+        // Set up connection
+        String urlBody;
+
+        if (authorityUrl == null) {
+            // Fetch the tenant from the authority
+            String tenant = authorizationRequest.getAuthority().getPath().split("/")[1];
```
iambmelt, line 256: We need to pull this logic into the `MicrosoftStsOAuth2Configuration` -- I don't think we want to extract the tenant info like this from the URL, but instead we should inspect the `PublicClientApplicationConfiguration` to get the audience/tenant.

Hazard from parsing it from the URL is that this basically assumes the URL structure will never change. For AAD today, that's a safe assumption -- if we add support for direct-ADFS auth later though, this will break

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
t-fadura: will do!

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
iambmelt: Here is the relevant OID docs:
https://tools.ietf.org/id/draft-ietf-oauth-device-flow-08.html#Discovery

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: Parsing is definitely not ideal. I wanted to get a working PR out, I couldn't find the tenant as it's own field in the request's authority. Hopefully this will be fixed with the inclusion of `getDeviceAuthorizationEndpoint()` in `MicrosoftStsOAuth2Configuration`.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: Will convert to use `HttpRequest`

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
rpdome: On a side note, we should be able to just append /oauth2/v2.0/devicecode to the authority URL (getAuthority() inside MicrosoftStsOAuth2Configuration - see getTokenEndpoint() and getEndpointUrlFromRootAndSuffix() for more info)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+        if (authorizationCode < HttpsURLConnection.HTTP_MULT_CHOICE){
+            // Get and parse response body
+            streamReader = new BufferedReader(new InputStreamReader(con.getInputStream()));
+            final String responseBody = streamReader.readLine();
+            final HashMap<String, String> parsedResponseBody = new Gson().fromJson(responseBody, new TypeToken<HashMap<String, String>>() {}.getType());
+
+            // Create response and result objects
+            // Not sure where "code" is stored. Not found in the input stream...
+            final MicrosoftStsAuthorizationResponse authorizationResponse =
+                    new MicrosoftStsAuthorizationResponse(null, authorizationRequest.getState(), parsedResponseBody);
```
t-fadura, line 312: Any ideas where this would be found?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+        con.setConnectTimeout(5000);
+        con.setReadTimeout(5000);
+
+        // Set up request body
+        con.setDoOutput(true);
+        final DataOutputStream out = new DataOutputStream(con.getOutputStream());
+        StringBuilder encodedResult = new StringBuilder();
+
+        // Encode client_id
+        encodedResult.append(URLEncoder.encode("client_id", "UTF-8"));
```
rpdome, line 279: utilize serializeObjectToFormUrlEncoded()

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
iambmelt: I mean to say, this shouldn't even be something developers are _tempted_ to show users if it's not localized.

`devicecode_token_already_redeemed` for example, nobody would try and show that to a user. Whereas a full sentence worht of error may imply _maybe_ this is intended to be consumed by the user

iambmelt: On further consideration, I think this is probably OK as is; developers who are concerned with localization should be aware _already_ that MSAL isn't localizing messages

iambmelt: I'll self-resolve. Anyone feel free to reactivate if you thing that rationale is flawed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: the Authority is the same in the `authorizationRequest` and the `mConfig`. The issue is that `path` (which contains tenant) ends with `/common`, so it's structured like `/{tenant}/common`. For DCF, I need to remove the `/common` at the end of it before adding `/oauth2/v2.0/devicecode` or else the request will return 404 "not found".

t-fadura: Same thing occurs in the token polling request, but I can just change `mTokenEndpoint` there and the whole process is a lot less messy.

rpdome: Would you mind sharing your config file? I'd expect either /{tenantID} or /common, /organization, etc.

t-fadura: Git doesn't like Json files? Here's a screenshot:
Annotation 2020-07-29 123118

t-fadura: At some point, /common is being added to the tenant in `authority_url`

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
AdamBJohnsonx: > Will convert to use `HttpRequest`

Better, make an HttpClient from the same package and use it.  We can replace your member field with an injection point in the future.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftTokenRequest.java
AdamBJohnsonx: or get prepared for type annotations and place it on the String type declaration.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: @AdamBJohnsonx I've already pushed a version using HttpRequest. I can go back and update it to use HttpClient once a few of my other tasks are complete.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
rpdome: That sounds like a bug to me....

t-fadura: iirc that value would be the tenant and /common at the end of it

t-fadura: i'm assuming you're referring to the `mConfig` object in `OAuth2Strategy`?

rpdome: Yes, but I'm referring to the class itself. @iambmelt's original suggestion is to put this url logic in that class and get the endpoint from there.

t-fadura:
Parsing only used in getDeviceCode() if authority is not provided. In LocalMsalController in MSAL (Code bellow), authorityUri provided from audience object.
Annotation 2020-07-30 143521

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
iambmelt: This looks to be referring to the `auth_code` normally returned by the Authorization Code Flow. I think you can leave null if this is unused in your scenario and if the field is allowed to be nullable

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -192,6 +192,8 @@
         public static final String LOCAL_ACQUIRE_TOKEN_SILENT= "103";
         public static final String LOCAL_GET_ACCOUNTS= "106";
         public static final String LOCAL_REMOVE_ACCOUNT= "107";
+        public static final String LOCAL_DEVICE_CODE_FLOW_ACQUIRE_URL_AND_CODE = "108";
```
iambmelt, line 195: These constants look to be out of range for the DCF IDS (600-699)

For this, we'll need to update the value to be in this range, and update the spec to account for which value that you use:
- [Spec -- Msft corpnet reqd](https://microsoft-my.sharepoint-df.com/:x:/p/sagonzal/EXSrr4vM1utAqQQfD6bMln4BYKwjrqh3cagiNJWPVNjLzw?e=DH7Kmd)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
iambmelt: Heard back from the service team, this should be available around the first week of September. For now, we will hardcode and make the necessary change to resolve this value dynamically later

iambmelt: Created a tracking work item here:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1096018

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
```diff
@@ -192,6 +192,8 @@
         public static final String LOCAL_ACQUIRE_TOKEN_SILENT= "103";
```
t-fadura, line 192: Is `LOCAL_COMPLETE_ACQUIRE_TOKEN_INTERACTIVE` supposed to be "102"?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
t-fadura: I think that excel sheet might only apply to the codes in `PublicApiId`? I think so, because the codes that are already in `TelemetryEventStrings` don't seem to match up with what's in the excel sheet.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: It's nullable, but the JavaDoc for the field states that it's required. This may just be a remnant of the old code back when only auth_code was supported.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
iambmelt: You are right; I am confused by who/what is consuming _these values_. Let's get some input from @shahzaibj (and maybe @sangonzal or @logdog16 if they know who or what is supposed to consume these values: these numbers don't match what's documented in the API ID Excel Sheet)

iambmelt: These constants were added as part of this PR:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/519

Haven't located a spec capturing the numbers (yet)

t-fadura: Gotcha, let me know if I need to make any changes

iambmelt: Let's get input from @shahzaibj on Monday; I can't locate a spec that captures what these values are or should be

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
iambmelt: Let's hold for @shahzaibj -- can't locate this information

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
iambmelt: Since you won't have this value, I'll suggest we add a new constructor to use for DCF flows. If that ends up messy (or complicates the code), feel free to update the docs and pass `null`

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryEventStrings.java
shahzaibj: @iambmelt Thanks for catching this. The values declared in the excel sheet you shared are the Api Ids for the Public APIs in MSAL. These are declared in `PublicApiId.java` in our code base, 

However, the values declared in this file i.e `TelemetryEventStrings.java` actually seem to capture an API ID for the internal controller methods in MSAL/common/broker. So there is an API Id for the `acquireTokenSilent` method in `LocalMsalController` vs one for the similar method in `BrokerMsalController` -- so these are different from the Public Api. 

With that said, even I'm not aware if there was ever a spec for these and/or if these were documented anywhere. Perhaps since these are internal, we could use anything for them? TBH, I'm not so sure. I'll see if I can track down any document that captures these, but for the time being @t-fadura let's make sure that the values we put here are different from the ones we put in `PublicApiId.java` as these fundamentally represent a different thing. I'm not too worried about the actual values you put here, let's just put anything but just make sure it is not same as public api id.

rpdome:
test is failing.

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
```diff
@@ -408,6 +408,12 @@ private ErrorStrings() {
                     "Please make sure to use your organizational account. " +
                     "If that doesn’t help, please return the device to your administrator.";
 
+    /**
+     * Device Code Flow only.
+     * Authorization has not been completed yet.
+     */
+    public final static String DEVICE_CODE_FLOW_AUTHORIZATION_PENDING_CODE = "authorization_pending";
```
shahzaibj, line 415:
```suggestion
    public final static String DEVICE_CODE_FLOW_AUTHORIZATION_PENDING_ERROR_CODE = "authorization_pending";
```

shahzaibj: Looks like we've followed the pattern `....ERROR_CODE` and `....ERROR_MESSAGE` but I don't see that being used in these new constants. Let's try to be consistent.

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
```diff
@@ -72,7 +80,14 @@ public AcquireTokenResult execute() throws Exception {
         );
 
         // Call acquireDeviceCodeFlowToken to get token result (Part 2 of DCF)
-        return controller.acquireDeviceCodeFlowToken(authorizationResult, commandParameters);
+        final AcquireTokenResult tokenResult = controller.acquireDeviceCodeFlowToken(authorizationResult, commandParameters);
+
+        Logger.verbose(
+                TAG + methodName,
+                "Device Code Flow command exiting with token..."
```
shahzaibj, line 87: > with token

Are we certain about this?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
shahzaibj: @t-fadura AFAIK, DCF is only supported for AAD and not B2C. So you should be able to get the `tenantId` as follows:

`((AzureActiveDirectoryAuthority) mConfig.getDefaultAuthority()).getAudience().getTenantId();`

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -343,9 +343,23 @@ public MicrosoftStsTokenRequest createTokenRequest(@NonNull final MicrosoftStsAu
...
         tokenRequest.setClaims(request.getClaims());
-        tokenRequest.setGrantType(TokenRequest.GrantTypes.AUTHORIZATION_CODE);
         setTokenRequestCorrelationId(tokenRequest);
 
+        // Existence of a device code inside of the response object implies Device Code Flow is being used
+        if (response.getDeviceCode() != null) {
+            tokenRequest.setGrantType(TokenRequest.GrantTypes.DEVICE_CODE);
+            tokenRequest.setDeviceCode(response.getDeviceCode());
+
+            // mTenantId from mAudience (being "common") is being added to the path in AzureActiveDirectoryAuthority
```
shahzaibj, line 353: Why was common the audience to begin with?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -343,9 +343,23 @@ public MicrosoftStsTokenRequest createTokenRequest(@NonNull final MicrosoftStsAu
...
+            // Needed to omit "\common" from the endpoint so as to not break DCF
+            // mTokenEndPoint be automatically reset in future commands
+            String newTokenEndpoint = this.mTokenEndpoint.replace("/common/", "/");
```
shahzaibj, line 356: We shouldn't be doing this hack -- we need to find the exact issue here

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -77,6 +87,7 @@
     private static final String TAG = OAuth2Strategy.class.getSimpleName();
 
     protected static final String TOKEN_REQUEST_CONTENT_TYPE = "application/x-www-form-urlencoded";
+    protected static final String DEVICE_CODE_CONTENT_TYPE = "application/x-www-form-urlencoded";
```
shahzaibj, line 90:
```suggestion
    protected static final String DEVICE_CODE_CONTENT_TYPE = TOKEN_REQUEST_CONTENT_TYPE;
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +241,80 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+            final MicrosoftStsAuthorizationResponse authorizationResponse =
+                    new MicrosoftStsAuthorizationResponse(null, authorizationRequest.getState(), parsedResponseBody);
+            authorizationResult = new MicrosoftStsAuthorizationResult(AuthorizationStatus.SUCCESS, authorizationResponse);
+
+            Logger.verbose(
+                    TAG + methodName,
+                    "Device Code Flow authorization successful..."
+            );
+        }
+        else { // Request failed
```
rpdome, line 296: nit: new line for comment.

common/src/main/java/com/microsoft/identity/common/exception/ErrorStrings.java
t-fadura: Will do!

common/src/main/java/com/microsoft/identity/common/internal/commands/DeviceCodeFlowCommand.java
t-fadura: I think this is a safe assumption. If the token request had failed, a service exception would've been thrown and this log wouldn't have been created.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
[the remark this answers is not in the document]
t-fadura: I get a 404 error, also happens with the authorization step if /common is placedafter the tenant

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -459,14 +466,227 @@ public boolean removeCurrentAccount(RemoveAccountCommandParameters parameters) t
...
+                ThreadUtils.sleepSafely(interval, TAG,
+                        "Attempting to sleep thread during Device Code Flow token polling...");
+            }
+        }
+
+        // Assign token result
+        acquireTokenResult.setTokenResult(tokenResult);
+
+        // Validate request success, may throw MsalServiceException
+        validateServiceResult(tokenResult);
```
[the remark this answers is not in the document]
t-fadura: Yes, good point

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -459,14 +466,227 @@ public boolean removeCurrentAccount(RemoveAccountCommandParameters parameters) t
...
+                parameters.getAuthenticationScheme()
+        );
+
+        // Fetch wait interval
+        final int interval = Integer.parseInt(authorizationResponse.getInterval()) * 1000;
+
+        String errorCode = ErrorStrings.DEVICE_CODE_FLOW_AUTHORIZATION_PENDING_CODE;
+
+        // Loop to send multiple requests checking for token
+        while (authorizationPending(errorCode)) {
```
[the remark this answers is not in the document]
t-fadura: @shahzaibj suggested this for better readability

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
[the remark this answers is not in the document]
rpdome: Device Code Flow supports /common though.

Parameter | Condition | Description
tenant | Required | Can be /common, /consumers, or /organizations. It can also be the directory tenant that you want to request permission from in GUID or friendly name format.

t-fadura:
Tests passing again

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -459,14 +466,227 @@ public boolean removeCurrentAccount(RemoveAccountCommandParameters parameters) t
...
+
+            // Fetch error if the request failed
+            if (tokenResult.getErrorResponse() != null) {
+                errorCode = tokenResult.getErrorResponse().getError();
+            }
+
+            // Check if authorization is pending
+            if (authorizationPending(errorCode)) {
+                // Wait between polls
+                ThreadUtils.sleepSafely(interval, TAG,
```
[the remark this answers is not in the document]
t-fadura: good idea

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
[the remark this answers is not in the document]
t-fadura: Tested in authorization step, supplying "common" in place of the tenant ID return "invalid_request". Even if it worked, wouldn't we run into the same issue where we would have to parse out the tenant id in order to just get /common in the endpoint?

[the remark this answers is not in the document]
rpdome: `Tested in authorization step, supplying "common" in place of the tenant ID return "invalid_request".`
>> What are the exact error message that you're getting? Does it have something to do with single tenant/multiple tenant applications? /common is not supported if your device is marked as a single tenant application.

` Even if it worked, wouldn't we run into the same issue where we would have to parse out the tenant id in order to just get /common in the endpoint?`
>> This is what I've been saying that this is a bug. It doesn't make sense to be to have /common append after tenantID in the first place, and we should be fixing the problem at the root cause (why /common was appended, and if it's justified.)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: I have an `AzureActiveDirectoryAuthority` object in `LocalMsalController` being passed through command parameters. `mTenantId` in it is "common", where as `mCloudUrl` in it is "https://login.microsoftonline.com/{MyTenant}". Maybe this is the base issue?

t-fadura: @rpdome @shahzaib I found one issue, my methods were missing this line
`Authority.KnownAuthorityResult authorityResult = Authority.getKnownAuthorityResult(parametersWithScopes.getAuthority());`

This line would implicitly call `performCloudDiscovery()` in `AzureActiveDirectory`, which plays a big part in how the authorities are structured since `mAzureActiveDirectoryCloud` is no longer null in `AzureActiveDirectoryAuthority`. (Sorry for the confusing explanation)

Currently tying to figure out why "common" is in the mTenantId field. Once this is fixed there shouldn't been any need to mess with the token endpoint

t-fadura: Authority issue fixed, I was not passing the correct authority object to the command parameters object. Resolving after fix is commited.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +245,101 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+
+        final HttpURLConnection con = (HttpsURLConnection) url.openConnection();
+        con.setRequestMethod("POST");
+        con.setRequestProperty("Content-Type", "application/x-www-form-urlencoded");
+        con.setConnectTimeout(5000);
+        con.setReadTimeout(5000);
+
+        // Set up request body
+        con.setDoOutput(true);
+        final DataOutputStream out = new DataOutputStream(con.getOutputStream());
```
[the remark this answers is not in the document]
t-fadura: Luckily this ended up being code I didn't use 😄

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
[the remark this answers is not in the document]
t-fadura: Authority issue has been fixed. The error was essentially saying that there was no tenant information included.

t-fadura: Fixed

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
t-fadura: The default authority in PCA defaults to having common as the tenant, I had to create a custom authority.

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -459,14 +465,236 @@ public boolean removeCurrentAccount(RemoveAccountCommandParameters parameters) t
...
+
+        // Create token request outside of loop so it isn't re-created after every loop
+        final MicrosoftStsTokenRequest tokenRequest = (MicrosoftStsTokenRequest) oAuth2Strategy.createTokenRequest(
+                mAuthorizationRequest,
+                authorizationResponse,
+                parameters.getAuthenticationScheme()
+        );
+
+        // Fetch wait interval
+        final int interval = Integer.parseInt(authorizationResponse.getInterval()) * 1000;
```
rpdome, line 585: nit: intervalInMilliseconds?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +240,70 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
         mAuthorizationEndpoint = authorizationEndpoint;
     }
 
+    public AuthorizationResult getDeviceCode(@NonNull final MicrosoftStsAuthorizationRequest authorizationRequest) throws IOException {
+        final String methodName = ":getDeviceCode";
+
+        final String urlBody = ((MicrosoftStsOAuth2Configuration) mConfig).getAuthorityUrl().toString() + "/oauth2/v2.0/devicecode";
```
rpdome, line 246: Can we extract this into MicrosoftStsOAuth2Configuration.getDeviceCodeEndpoint()? 

No need to fetch the endpoint from .well-known/openid-configuration yet. (If you could add a TODO there that we're pending server change, that would be really great.)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +240,70 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+        AuthorizationResult authorizationResult;
+
+        // Check if the request was successful
+        // Any code below 300 (HTTP_MULT_CHOICE) is considered a success
+        if (response.getStatusCode() < HttpsURLConnection.HTTP_MULT_CHOICE){
+            // Get and parse response body
+            final HashMap<String, String> parsedResponseBody = new Gson().fromJson(response.getBody(), new TypeToken<HashMap<String, String>>() {}.getType());
+
+            // Create response and result objects
+            // Not sure where "code" is stored. Not found in the input stream...
```
rpdome, line 273: nit: Update this comment to reflect your conversation with @iambmelt  below?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -230,6 +240,70 @@ protected final void setAuthorizationEndpoint(final String authorizationEndpoint
...
+            final MicrosoftStsAuthorizationResponse authorizationResponse =
+                    new MicrosoftStsAuthorizationResponse(null, authorizationRequest.getState(), parsedResponseBody);
+            authorizationResult = new MicrosoftStsAuthorizationResult(AuthorizationStatus.SUCCESS, authorizationResponse);
+
+            Logger.verbose(
+                    TAG + methodName,
+                    "Device Code Flow authorization successful..."
+            );
+        }
```
rpdome, line 282: nit: early return? (so there's no need to create a AuthorizationResult pointer in line 264).

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
t-fadura: Ok, will do!

PR #985, opened 2020-07-29
replacing commit() with apply() to for async access

tanmaymanolkar1:
Replacing commit() with apply() for async access to the persistent storage to avoid android.os.strictmode.DiskWriteViolation.
Commit() synchronously perform disk read/writes on the main UI thread which possibly leads to violation.

Link to the bug:
https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1041

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -139,7 +139,7 @@ public class SharedPreferencesFileManager implements ISharedPreferencesFileManag
             editor.putString(key, encryptedValue);
         }
 
-        editor.commit();
+        editor.apply();
     }
 
     @Override
@@ -223,7 +223,7 @@ public class SharedPreferencesFileManager implements ISharedPreferencesFileManag
     public final void clear() {
         final SharedPreferences.Editor editor = mSharedPreferences.edit();
         editor.clear();
-        editor.commit();
+        editor.apply();
     }
 
     @SuppressLint("ApplySharedPref")
@@ -236,7 +236,7 @@ public class SharedPreferencesFileManager implements ISharedPreferencesFileManag
 
         final SharedPreferences.Editor editor = mSharedPreferences.edit();
         editor.remove(key);
-        editor.commit();
+        editor.apply();
 
         Logger.infoPII(
                 TAG,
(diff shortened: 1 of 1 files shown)


kreedula:
Just looking at documentation of [apply](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()). The line below  indicates that , we can use `apply` if we don't care for a `boolean` return value from `commit`.  Here even though, we are not explicitly checking the `boolean` from the `commit`, it seems safer to use `commit` given the ambiguity around this.

*As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.*

iambmelt:
> Just looking at documentation of [apply](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()). The line below indicates that , we can use `apply` if we don't care for a `boolean` return value from `commit`. Here even though, we are not explicitly checking the `boolean` from the `commit`, it seems safer to use `commit` given the ambiguity around this.
> 
> _As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value._

An API change to include the return value would be welcome. I can't fathom why this wasn't done this way in the first place.... an oversight on my part 🤦

tanmaymanolkar1:
The reason why I ended up using the apply() was precisely because the return values of the commit() weren't being used. It would be better to use the values to get the status of the operations on sharedpreferences.
Is the fix for async access around `persistCurrentAccount()` intended to be made by us? @iambmelt

iambmelt:
> The reason why I ended up using the apply() was precisely because the return values of the commit() weren't being used. It would be better to use the values to get the status of the operations on sharedpreferences.
> Is the fix for async access around `persistCurrentAccount()` intended to be made by us? @iambmelt

@tanmaymanolkar1 yeah, though that PR would be for MSAL, not for Common. If you wanted to open a PR to wire up the return values from the `commit()` I'd happily review/take that change

tanmaymanolkar1:
Oh this PR is for tackling the issue of strict mode disk reads and writes and async is the way to go about it, but rather than playing with strictmode and giving the main UI thread the access for reads and writes, I changed the commit() to apply() since apply is an async job ,which would not be done using the main UI thread but using a new child thread, and then we wouldn't have the need to explicitly change the permissions of strict mode.

iambmelt:
Synced offline. Sharing notes:

- API to remain as-is (`void`)
- Switch `commit()` to `apply()`
- Remove suppressions and relevant commentary from affected source file

/cc @tanmaymanolkar1

tanmaymanolkar1:
Sounds good @iambmelt

tanmaymanolkar1:
All the test cases passed.

PR #988, opened 2020-08-04
Changes to resurrect telemetry test + add another constant

shahzaibj:


testutils/src/main/java/com/microsoft/identity/internal/testutils/mocks/MockServerResponse.java
```diff
@@ -52,7 +52,7 @@ public static HttpResponse getMockTokenFailureInvalidGrantResponse() {
                 "\t\"trace_id\": \"8497799a-e9f9-402f-a951-7060b5014600\",\n" +
                 "\t\"correlation_id\": \"390d7507-c607-4f05-bb8a-51a2a7a6282b\",\n" +
                 "\t\"error_uri\": \"https://login.microsoftonline.com/error?code=70000\",\n" +
-                "\t\"suberror\": \"bad_token\"\n" +
+                "\t\"suberror\": \"\"\n" +
```
rpdome, line 55: So.. we're not expecting bad_token anymore?

rpdome, line 55: Why are we removing this?

shahzaibj: Back when the test was written, we weren't deleting tokens when a `bad_token` suberror was thrown by ESTS. However, we changed that behavior [here ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/940/files) so now I've refactored this mock response to not throw the `bad_token` suberror so that MSAL doesn't remove tokens.....because the telemetry test was written around the old behavior. There were two ways to fix:

- Keep this mocked response and refactor the test to account for the fact that a `bad_token` suberror could wipe RT
- Or just refactor the Mock Response to not throw the `bad_token` suberror

So I went with the second approach

rpdome: given that the server is now returning bad_token, shouldn't we keep the response valid and modify the test instead?

shahzaibj: I don't think the server will always return `bad_token` when `invalid_grant` occurs, so all we have done here is that we are mocking a different error scenario from the server. Essentially we are saying that the server is throwing an `invalid_grant` error here without any sub error.

Do we know if the server would always return `bad_token` with an `invalid_grant`? (I don't think so). May be @kreedula can provide some insight?

PR #989, opened 2020-08-05
Extract the cancel broadcast method from CommandDispatcher.

AdamBJohnsonx:
If we want to cancel requests from OneAuth, we'll want a method to let us do so.
We'll probably also want to synchronize on the broadcast and execution methods so that
we don't suffer from bad interleaving of requests to cancel and execute.  This is
still a living race condition, but at least it's a correct race condition.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -300,10 +300,7 @@ public class CommandDispatcher {
 
             // only send broadcast to cancel if within broker
             if (command.getParameters() instanceof BrokerInteractiveTokenCommandParameters) {
-                // Send a broadcast to cancel if any active auth request is present.
-                localBroadcastManager.sendBroadcast(
-                        new Intent(CANCEL_INTERACTIVE_REQUEST)
-                );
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -356,6 +353,17 @@ public void onReceive(Context context, Intent intent) {
         }
     }
 
+    /**
+     * Cancel any current interactive requests.
+     * @param broadcastManager the broadcast manager to use to cancel requests.
+     */
+    public static void cancelInteractiveRequest(LocalBroadcastManager broadcastManager) {
```
shahzaibj, line 360: nit: `NonNull` and `final`

AdamBJohnsonx: OK.  In case we need it.  Otherwise we shouldn't ship this.  I'm going to fix this and make it a draft.

AdamBJohnsonx, line 360:
```suggestion
    public static void cancelInteractiveRequest(@NonNull final LocalBroadcastManager broadcastManager) {
```

PR #990, opened 2020-08-05
Add additional suberrors which ESTS could return to map to MSAL CPP sub error constants

kreedula:
- Also made `PopAuthenticationSchemeInternal` constructor public to use in MSAL CPP

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1447,6 +1447,47 @@ public final class AuthenticationConstants {
          * Broker should make a request to DRS to get the current device status and act accordingly.
          */
         public static final String DEVICE_AUTHENTICATION_FAILED = "device_authentication_failed";
+
+        /**
+         * SubError code for cases when client not in the Microsoft first party family group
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1447,6 +1447,47 @@
...
+
+        /**
+         * OpenId connect suberror code, when user consent is required.
+         */
+        public static final String CONSENT_REQUIRED = "consent_required";
+
+        /**
+         * Custom sub error that notifies the user that their password has expired.
+         */
+        public static final String USER_PASSWORD_EXPIRED = "user_password_expired";
```
shahzaibj, line 1490: The javadoc for each of these doesn't seem to follow a pattern.

Some places it starts with `Oauth2 suberror code`
Some places it starts with `SubError code`
Some places it starts with `Conditional access suberror code`
Some places it starts with `OpenId connect suberror code`
Some places it starts with `Custom sub error`

And this entire class itself has the following javadoc: `Represents the oauth2 sub error code.`

Would be good if we can make these consistent somehow.

kreedula: Thanks Shahzaib, made it more consistent

PR #991, opened 2020-08-06
Handle full authority URL properly

rpdome:
fix https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1121

When the following authority pattern is used in the config file
```
"authorities" : [
    {
      "type": "AAD",
      "authority_url": "https://login.microsoftonline.com/rapong.onmicrosoft.com"
    }
  ]
```

AuthorityDeserializer would just take the whole "authority_url" and save it as cloudUrl.

When we're actually trying to get an authority, we would construct that by appending cloudUrl with tenantId.

In this case, we'll end up getting https://login.microsoftonline.com/rapong.onmicrosoft.com/common.


common/src/main/java/com/microsoft/identity/common/internal/authorities/AuthorityDeserializer.java
```diff
@@ -48,9 +51,14 @@ public Authority deserialize(JsonElement json, Type typeOfT, JsonDeserialization
...
-                    if (aadAuthority != null && aadAuthority.mAudience != null) {
-                        aadAuthority.mAudience.setCloudUrl(aadAuthority.mAuthorityUrl);
+                    final AzureActiveDirectoryAuthority aadAuthority = context.deserialize(authorityObject, AzureActiveDirectoryAuthority.class);
+                    if (aadAuthority != null && aadAuthority.mAudience != null && aadAuthority.mAuthorityUrl != null) {
+                        final Uri uri = Uri.parse(aadAuthority.mAuthorityUrl);
+                        aadAuthority.mAudience.setCloudUrl(uri.getScheme() + "://" + uri.getHost());
+                        // Try to extract tenant ID from the authority link.
+                        if (!TextUtils.isEmpty(uri.getLastPathSegment())) {
+                            aadAuthority.mAudience.setTenantId(uri.getLastPathSegment());
+                        }
```
shahzaibj, line 61: Looks like we are just setting the tenant id on the audience -- whatever audience was created; which may not be correct in the first place i.e. not having any audience in the config defaults to `AllAccounts`. I think we need to make sure the the audience on the Authority is of the correct type.

rpdome: In this case, audience won't be provided. 

i.e. 
```
"authorities" : [
    {
      "type": "AAD",
      "authority_url": "https://login.microsoftonline.com/rapong.onmicrosoft.com"
    }
  ]
```

We wouldn't be able to identify the audience type in the first place.

rpdome: (both tenant name and tenant id work interchangeably)

shahzaibj: We should be able to by splitting the URL in parts and checking the last segment and create the following mapping:

- /common - AllAccounts
- /organizations - AnyOrganizationalAccount
- /consumers - AnyPersonalAccount
- /tenantId (or tenantName) - AccountsInOneOrganization

and then we can initialize the authority object with the correct audience

rpdome: ah ok. makes sense. will do.

rpdome:
> I find it odd that this didn't require any tests to change. Should we exercise this code in a test? It might be a bit of a pain, because you'll need to create a shadow for AzureActiveDirectoryAudience.

I don't think we have any automated test that cover this scenarios.... otherwise we would've caught this error long time back.

I definitely could add a test case for this.

rpdome:
Will create a separate PR for adding the missing test coverage.

PR #992, opened 2020-08-07
Add app-name and app-ver headers to /auth and /token requests

shahzaibj:
Address this API consistency item: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/865193

Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1352


common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -188,6 +189,18 @@ public abstract AcquireTokenResult acquireDeviceCodeFlowToken(final Authorizatio
                 scopes.addAll(interactiveTokenCommandParameters.getExtraScopesToConsent());
             }
 
+            final HashMap<String, String> completeRequestHeaders = new HashMap<>();
+
+            if (interactiveTokenCommandParameters.getRequestHeaders() != null) {
+                completeRequestHeaders.putAll(interactiveTokenCommandParameters.getRequestHeaders());
```
AdamBJohnsonx, line 195: This is super-trivial, but we can save some allocations on the map resize if we size it to start with or use the Map/Map constructor to handle this trivially.

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -469,6 +470,12 @@ private HttpResponse performPKeyAuthRequest(
         headers.put("client-request-id", DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
         headers.putAll(Device.getPlatformIdParameters());
 
+        final String appName = request.getClientAppName();
+        final String appVer = request.getClientAppVersion();
```
AdamBJohnsonx, line 474: nit: Do we need these variables?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -185,6 +186,14 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         headers.putAll(Device.getPlatformIdParameters());
         headers.putAll(EstsTelemetry.getInstance().getTelemetryHeaders());
 
+        if (request instanceof MicrosoftTokenRequest) {
+            final String appName = ((MicrosoftTokenRequest) request).getClientAppName();
+            final String appVer = ((MicrosoftTokenRequest) request).getClientAppVersion();
```
AdamBJohnsonx, line 191: nit: do we need these variables?

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
shahzaibj: Yeah, I'll keep as is for now to avoid branched logic for creating hashmap. because this could be null

PR #996, opened 2020-08-10
Expose RT to broker apps

rpdome:
Allowing Broker apps to extract BRT to perform WPJ.

Related PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1126


common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRtAccessor.java
```diff
@@ -0,0 +1,85 @@
...
+import com.microsoft.identity.common.internal.broker.BrokerValidator;
+import com.microsoft.identity.common.internal.cache.ICacheRecord;
+import com.microsoft.identity.common.internal.cache.MsalOAuth2TokenCache;
+import com.microsoft.identity.common.internal.dto.AccountRecord;
+
+import java.util.List;
+
+import static com.microsoft.identity.common.exception.ClientException.TOKEN_CACHE_ITEM_NOT_FOUND;
+
+public class BrokerRtAccessor {
```
shahzaibj, line 40: This should be `final` I think

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRtAccessor.java
```diff
@@ -0,0 +1,83 @@
...
+                accountObjectId
+        );
+
+        // Check it's not null
+        if (null == localAccountRecord) {
+            // Unrecognized identifier, cannot supply a token.
+            throw new ClientException(TOKEN_CACHE_ITEM_NOT_FOUND);
+        }
+
+        return mTokenCache.load(
```
kreedula, line 76: Given today's meeting, we might to to rethink this logic, as this would save the BRT in the MSAL unified cache

rpdome: The first one is clean enough to do, the second one I think it'll break the pattern we have.

Eventually, I think if we want to restrict BRT lifetime, we should be enforcing this on the server side.

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRtAccessor.java
```diff
@@ -0,0 +1,83 @@
...
+import com.microsoft.identity.common.exception.ClientException;
+import com.microsoft.identity.common.internal.authscheme.BearerAuthenticationSchemeInternal;
+import com.microsoft.identity.common.internal.broker.BrokerValidator;
+import com.microsoft.identity.common.internal.cache.ICacheRecord;
+import com.microsoft.identity.common.internal.cache.MsalOAuth2TokenCache;
+import com.microsoft.identity.common.internal.dto.AccountRecord;
+
+import static com.microsoft.identity.common.exception.ClientException.TOKEN_CACHE_ITEM_NOT_FOUND;
+
+public final class BrokerRtAccessor {
```
iambmelt, line 38: Class javadoc?

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRefreshTokenAccessor.java
```diff
@@ -0,0 +1,103 @@
...
+    private final MsalOAuth2TokenCache mTokenCache;
+    private final Context mContext;
+
+    public BrokerRefreshTokenAccessor(@NonNull final MsalOAuth2TokenCache cache,
+                                      @NonNull final Context context) {
+        mTokenCache = cache;
+        mContext = context;
+    }
+
+    public @Nullable String getBrokerRefreshToken(@NonNull final String accountObjectId) throws ClientException {
```
iambmelt, line 56: Method javadoc? No description of format for `accountObjectId` could be confusing

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRefreshTokenAccessor.java
```diff
@@ -0,0 +1,103 @@
...
+ * For Broker apps to obtain Broker RT (for WPJ scenario).
+ * */
+public final class BrokerRefreshTokenAccessor {
+
+    private static final String TAG = BrokerRefreshTokenAccessor.class.getSimpleName();
+
+    private final MsalOAuth2TokenCache mTokenCache;
+    private final Context mContext;
+
+    public BrokerRefreshTokenAccessor(@NonNull final MsalOAuth2TokenCache cache,
```
iambmelt, line 50: Call me crazy, but would prefer to see `Context` as the first argument -- I can't really provide a justification tho 😅

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRefreshTokenAccessor.java
```diff
@@ -0,0 +1,103 @@
...
+                                      @NonNull final Context context) {
+        mTokenCache = cache;
+        mContext = context;
+    }
+
+    public @Nullable String getBrokerRefreshToken(@NonNull final String accountObjectId) throws ClientException {
+        final String methodName = "getBrokerRefreshToken";
+
+        if (!AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equals(mContext.getPackageName()) &&
```
iambmelt, line 59: Maybe `static import` these constants? Pretty clear what they are, could be less verbose

common/src/main/java/com/microsoft/identity/common/adal/internal/BrokerRefreshTokenAccessor.java
```diff
@@ -0,0 +1,112 @@
...
+                                      @NonNull final MsalOAuth2TokenCache cache) {
+        mContext = context;
+        mTokenCache = cache;
+    }
+
+    /**
+     * Retrieves Broker Refresh Token ONLY IF the caller is the broker apps and has previously made a request with Broker Client ID.
+     * This will also wipe the RT from MSAL local cache.
+     *
+     * @param accountObjectId object id of the account.
```
iambmelt, line 62: We may want to say _specifically_ that this is the `local_account_id`

PR #1007, opened 2020-08-20
Update lab classes to support migration testing

shahzaibj:
- Add lab constants for migrated tenants
- Update lab secret helper to get password secret name from credentialVaultKeyName (lab name doesn't work for migrated tenants 😢)

Related: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1133


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -22,8 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.identity.internal.testutils.labutils;
 
-import androidx.core.util.Consumer;
-
```
AdamBJohnsonx, line 26: separate commits for formatter runs would be nice.

Issue #905, opened 2020-05-06
Duplicate command errors


slowcar:
When executing multiple workers concurrently i get duplicate command errors from time to time.

Responsible is the code added here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/abdcee2c6aec85b43cf5d53ce763a4151efbb494

I would like to be able to retrieve the current token from multiple threads/workers at once.

iambmelt:
Fixed in [2.0.1](https://github.com/AzureAD/microsoft-authentication-library-for-android/releases/tag/v2.0.1)

PR #1010, opened 2020-08-22
(Experimental) An approach to triggering fewer duplicate command errors

iambmelt:
Related:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/905

Add observers to Commands, once those commands finish notify the observers.


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -88,12 +88,9 @@ public static void submitSilent(@NonNull final BaseCommand command) {
         );
 
         if (sExecutingCommands.contains(command)) {
```
AdamBJohnsonx, line 90: OK, what would make this a little more readable, and flexible, is to cache the Future<Result> of this command.  We might want to look into importing streamsupport-cfuture or doing our own back port of CompletableFuture using androidx's SettableFuture as the basis if the GPLv2\Classpath license isn't acceptable.  This will give us a class that we can use for this kind of behavior in multiple places.

AdamBJohnsonx: The main thing that would change would be sExecutingCommands would change from a Set to a Map<BaseCommand, CompletableFuture<CommandResult>>.  Essentially what this becomes is a Cache that can store incomplete values.

iambmelt: Going to close this PR, will prep another with discussed changes

PR #1011, opened 2020-08-25
Get device authorization endpoint from well-known config

rpdome:
.well-known is now returning device_authorization_endpoint, so we'll use that.

[Here's](https://login.microsoftonline.com/msidlab4.onmicrosoft.com/v2.0/.well-known/openid-configuration) an example of how the whole thing looks like.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
@@ -65,7 +65,7 @@ public class MicrosoftStsOAuth2Configuration extends AzureActiveDirectoryOAuth2C
      */
     public URL getDeviceAuthorizationEndpoint() {
         final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
-        if (openIdConfig != null && openIdConfig.getTokenEndpoint() != null) {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -59,6 +59,18 @@ public URL getAuthorizationEndpoint() {
         return getEndpointUrlFromRootAndSuffix(getAuthorityUrl(), FALLBACK_AUTHORIZE_ENDPOINT_SUFFIX);
     }
 
+    /**
+     * Return device authorization endpoint to bo used in the authorization step of Device Code Flow.
+     * @return a URL object for the /devicecode endpoint
+     */
+    public URL getDeviceAuthorizationEndpoint() {
+        final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
+        if (openIdConfig != null && openIdConfig.getTokenEndpoint() != null) {
```
shahzaibj, line 68:
```suggestion
        if (openIdConfig != null && openIdConfig.getDeviceAuthorizationEndpoint() != null) {
```

rpdome: oh shoot. thanks for catching this.

PR #1012, opened 2020-08-25
An approach to triggering fewer duplicate command errors

iambmelt:
Redux of #1010 with suggestions from @AdamBJohnsonx 

Related:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/905

Add observers to Commands, once those commands finish notify the observers.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -94,21 +94,21 @@ public class CommandDispatcher {
         final Handler handler = new Handler(Looper.getMainLooper());
         ResultFuture<CommandResult> future = sExecutingCommandMap.get(command);
 
-        if (null != future) {
-            future.whenComplete(getCommandResultConsumer(command, handler));
-            return;
-        } else {
-            final ResultFuture<CommandResult> resultFuture = new ResultFuture<>();
-            final ResultFuture<CommandResult> putValue = sExecutingCommandMap.putIfAbsent(command, resultFuture);
+        if (null == future) {
+            future = new ResultFuture<>();
+            final ResultFuture<CommandResult> putValue = sExecutingCommandMap.putIfAbsent(command, future);
 
             if (null == putValue) {
                 // our value was inserted.
-                resultFuture.whenComplete(getCommandResultConsumer(command, handler));
+                future.whenComplete(getCommandResultConsumer(command, handler));
             } else {
                 // Our value was not inserted, grab the one that was and hang a new listener off it
                 putValue.whenComplete(getCommandResultConsumer(command, handler));
                 return;
             }
+        } else {
+            future.whenComplete(getCommandResultConsumer(command, handler));
+            return;
         }
 
         sSilentExecutor.execute(new Runnable() {
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -105,7 +105,7 @@ public class CommandDispatcher {
                 // our value was inserted.
                 resultFuture.whenComplete(getCommandResultConsumer(command, handler));
             } else {
-                // Our value was not inserted, grab the one that was and hand a new listener off it
+                // Our value was not inserted, grab the one that was and hang a new listener off it
                 putValue.whenComplete(getCommandResultConsumer(command, handler));
                 return;
             }
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -49,12 +49,16 @@ import com.microsoft.identity.common.internal.logging.DiagnosticContext;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.result.AcquireTokenResult;
 import com.microsoft.identity.common.internal.result.LocalAuthenticationResult;
+import com.microsoft.identity.common.internal.result.ResultFuture;
 import com.microsoft.identity.common.internal.telemetry.Telemetry;
+import com.microsoft.identity.common.internal.util.BiConsumer;
 
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -73,7 +77,8 @@ public class CommandDispatcher {
     private static final Object sLock = new Object();
     private static InteractiveTokenCommand sCommand = null;
     private static final CommandResultCache sCommandResultCache = new CommandResultCache();
-    private static final Set<BaseCommand> sExecutingCommands = Collections.synchronizedSet(new HashSet<BaseCommand>());
+    //private static final Set<BaseCommand> sExecutingCommands = Collections.synchronizedSet(new HashSet<BaseCommand>());
+    private static final ConcurrentMap<BaseCommand, ResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
 
     /**
      * submitSilent - Run a command using the silent thread pool
@@ -87,15 +92,16 @@ public class CommandDispatcher {
                 "Beginning execution of silent command."
         );
 
-        if (sExecutingCommands.contains(command)) {
-            new Handler(Looper.getMainLooper()).post(new Runnable() {
-                @Override
-                public void run() {
-                    command.getCallback().onError(new ClientException(ClientException.DUPLICATE_COMMAND, "The same command was already received and is being processed."));
-                }
-            });
-        } else if (command.isEligibleForCaching()) {
-            sExecutingCommands.add(command);
+        final Handler handler = new Handler(Looper.getMainLooper());
+        ResultFuture<CommandResult> future = sExecutingCommandMap.get(command);
+
+        if (null != future) {
+            future.whenComplete(getCommandResultConsumer(command, handler));
+            return;
+        } else {
+            final ResultFuture<CommandResult> resultFuture = new ResultFuture<>();
+            resultFuture.whenComplete(getCommandResultConsumer(command, handler));
+            sExecutingCommandMap.put(command, resultFuture);
         }
 
         sSilentExecutor.execute(new Runnable() {
@@ -111,7 +117,6 @@ public class CommandDispatcher {
                 EstsTelemetry.getInstance().emitApiId(command.getPublicApiId());
 
                 CommandResult commandResult = null;
-                Handler handler = new Handler(Looper.getMainLooper());
 
                 //Log operation parameters
                 if (command.getParameters() instanceof SilentTokenCommandParameters) {
@@ -139,12 +144,22 @@ public class CommandDispatcher {
                 Telemetry.getInstance().flush(correlationId);
                 EstsTelemetry.getInstance().flush(command, commandResult);
 
-                sExecutingCommands.remove(command);
-
                 //Return the result via the callback
-                returnCommandResult(command, commandResult, handler);
+                sExecutingCommandMap.remove(command).setResult(commandResult);
             }
         });
+    }
+
+    private static BiConsumer<CommandResult, Throwable> getCommandResultConsumer(
+            @NonNull final BaseCommand command,
+            @NonNull final Handler handler) {
+        return new BiConsumer<CommandResult, Throwable>() {
+            @Override
+            public void accept(CommandResult result, Throwable throwable) {
+                // We can ignore the second (exception) param, not used.
+                returnCommandResult(command, result, handler);
+            }
+        };
     }
 
     static void clearCommandCache() {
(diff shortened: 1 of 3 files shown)


iambmelt:
@AdamBJohnsonx Let's see how I did 😆

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -28,21 +59,55 @@ public boolean isDone() {
...
+            consumer.accept(result, mException);
+        }
+    }
+
+    public synchronized void whenComplete(@NonNull final BiConsumer<T, Throwable> consumerToAdd) {
+        if (isDone()) {
+            consumerToAdd.accept(mResult, mException);
+        }
+
+        // TODO Should we still add the consumer even if the ResultFuture isDone() == true?
```
AdamBJohnsonx, line 110: nah.  You can return.

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -28,21 +59,55 @@ public boolean isDone() {
...
 
-    public void setResult(T result) {
+    public synchronized void setResult(@Nullable final T result) {
         mResult = result;
         mCountDownLatch.countDown();
+
+        for (final BiConsumer<T, Throwable> consumer : mConsumers) {
+            consumer.accept(result, mException);
+        }
+    }
```
AdamBJohnsonx, line 113: At this point, you could drop all references in this list.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -73,7 +77,8 @@
     private static final Object sLock = new Object();
     private static InteractiveTokenCommand sCommand = null;
     private static final CommandResultCache sCommandResultCache = new CommandResultCache();
-    private static final Set<BaseCommand> sExecutingCommands = Collections.synchronizedSet(new HashSet<BaseCommand>());
+    //private static final Set<BaseCommand> sExecutingCommands = Collections.synchronizedSet(new HashSet<BaseCommand>());
```
iambmelt, line 80: Remove commented out code

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: Great --  I'll yank that out

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -28,21 +59,55 @@ public boolean isDone() {
...
+
+    public synchronized void whenComplete(@NonNull final BiConsumer<T, Throwable> consumerToAdd) {
```
iambmelt, line 105: Add javadoc

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -28,21 +59,55 @@ public boolean isDone() {
...
-            throw new TimeoutException();
+            throw new TimeoutException(
+                    "Timed out waiting for: "
+                            + l // duration
+                            + timeUnit.name() // units
+            );
         }
+    }
+
+    public synchronized void setException(@NonNull final Exception exception) {
```
iambmelt, line 87: Add javadoc

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -28,21 +59,55 @@ public boolean isDone() {
...
+        if (null != mException) {
+            throw new RuntimeException(mException);
+        }
+
         return mResult;
     }
 
     @Override
-    public T get(long l, TimeUnit timeUnit) throws InterruptedException, TimeoutException {
+    public T get(final long l, @NonNull final TimeUnit timeUnit) throws InterruptedException, TimeoutException {
```
AdamBJohnsonx, line 71: I was looking at this, I thought Future threw ExecutionExcepttion as well.  If it does, that's a better throw than runtime below.

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: 40c7fbca336602ea74b7fa06af766c94fb39fc4b

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: 6fdb1526d5b9562fe6e332faf13e7d41a1813ace

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -87,15 +92,16 @@ public static void submitSilent(@NonNull final BaseCommand command) {
...
+        final Handler handler = new Handler(Looper.getMainLooper());
+        ResultFuture<CommandResult> future = sExecutingCommandMap.get(command);
+
+        if (null != future) {
+            future.whenComplete(getCommandResultConsumer(command, handler));
+            return;
+        } else {
+            final ResultFuture<CommandResult> resultFuture = new ResultFuture<>();
+            resultFuture.whenComplete(getCommandResultConsumer(command, handler));
+            sExecutingCommandMap.put(command, resultFuture);
```
AdamBJohnsonx, line 103: OK, this is your tricky synchronization part using concurrentMap.  You don't want to risk overwriting a concurrent writer without synchronizing this entire method.  So you make the result future and it's empty.  You then putIfAbsent into the sExecutingCommandMap.  If it returns null, then it did not populate the map and you add your listener to the future that you created since it was added to the map.  If it returns non-null, then you lost the race and you need add a listener to that future and return, since the command is already running.

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: 1d0cfe608fbadb0ae79bde44e49e467f0dc2c371

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: 1d0cfe608fbadb0ae79bde44e49e467f0dc2c371

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -139,14 +144,24 @@ public void run() {
...
         });
     }
 
+    private static BiConsumer<CommandResult, Throwable> getCommandResultConsumer(
+            @NonNull final BaseCommand command,
+            @NonNull final Handler handler) {
+        return new BiConsumer<CommandResult, Throwable>() {
+            @Override
+            public void accept(CommandResult result, Throwable throwable) {
+                // We can ignore the second (exception) param, not used.
```
AdamBJohnsonx, line 158: Well, do we?  Long term, it's a better way of communicating error states.  The contract here is that either you have a result or you have a throwable.  Mostly, the code for these looks like if (t == null) { error handling; return; } result handling;

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -87,15 +92,16 @@ public static void submitSilent(@NonNull final BaseCommand command) {
...
+        if (null != future) {
+            future.whenComplete(getCommandResultConsumer(command, handler));
+            return;
+        } else {
+            final ResultFuture<CommandResult> resultFuture = new ResultFuture<>();
+            resultFuture.whenComplete(getCommandResultConsumer(command, handler));
+            sExecutingCommandMap.put(command, resultFuture);
         }
 
         sSilentExecutor.execute(new Runnable() {
```
AdamBJohnsonx, line 106: Hm.  OK, so we *might* want to pass in this executor to whenComplete.  As you may notice, in the event of a complete request, we're going to run the call back on whatever the current thread is running on.  If that's OK, it's more straightforward.  But if we want our callback to run on the silentExecutor thread in the event that the result is already completed by the time we attempt to run it, we probably want to pass this executor into whenComplete to use in the event we hit that case.

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: 47d8fa9856e1f784755abe10c2a61311ed78712d

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
iambmelt: fae08476ddf46ae797c5adfd62f1632a4190014a

Adding this checked Exception will require another PR to add `try`/`catch` blocks to MSAL so keep an eye out for that once this merges

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: >If it returns null, then it did not populate the map and you add your listener to the future that you created since it was added to the map. If it returns non-null, then you lost the race and you need add a listener to that future and return, since the command is already running.

This seems backwards -- wouldn't returning `null` indicate that I _did_ win? Looking at:
https://developer.android.com/reference/java/util/concurrent/ConcurrentHashMap?hl=en#putIfAbsent(K,%20V)

AdamBJohnsonx: You're right.  And my statement was internally inconsistent, even.  If it returns null, it *did* populate the map and you add the callback to the future that you created.  If it returns non-null, add and return.

iambmelt: Think we're saying the same thing; just read too quickly

iambmelt: Ah gotcha

AdamBJohnsonx: > Think we're saying the same thing?

Yes.  I'm trying to agree. ;-)

iambmelt: @AdamBJohnsonx Pushed changes in 98528c6c95c4eefe9d806b5f465bc8965b7f4fcb -- can you spot check me on this commit?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: Correct me if I'm wrong here.... I think we want to leave as-is, as calling back on the `silentExecutor` would be a threading behavioral change

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -87,15 +91,26 @@ public static void submitSilent(@NonNull final BaseCommand command) {
...
+            if (null == putValue) {
+                // our value was inserted.
+                resultFuture.whenComplete(getCommandResultConsumer(command, handler));
+            } else {
+                // Our value was not inserted, grab the one that was and hang a new listener off it
+                putValue.whenComplete(getCommandResultConsumer(command, handler));
+                return;
+            }
+
+            sExecutingCommandMap.put(command, resultFuture);
```
AdamBJohnsonx, line 113: Don't think you need this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
AdamBJohnsonx: returnCommandResult is called by the runnable that was submitted previously, I thought.  Unless it has a way to transfer back to the calling thread, which I don't see, we currently call back from that executor.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: Already removed, just didn't push the change -- it's in b908342ba087ef5bccca5e7eb486ea49093da911

iambmelt:
> Looks right, though you don't need to put it in the executing command map at the end (line 113). You can compact it if you like by inverting your first condition and reusing the future reference.

7495fa9c7dd03ba71a2244340a7b3b9d82617a35

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: It posts back on the `Handler` created at ~L94 see:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/command-sub-fix-905-redux/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java#L230

AdamBJohnsonx: Wild.  OK, then it doesn't matter. Good - it's simpler like this.

iambmelt: gg

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
iambmelt: Applied in 91ed0a71841cdf4095b7fac398dda2330df16da7

PR #1013, opened 2020-08-27
Adding the OpenIdConnectPromptParameter field UNSET

tanmaymanolkar1:
Fix for the github issue https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1122

Added the field NOT_SENT and replaced the NONE, for the reasons of naming conventions

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdConnectPromptParameter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdConnectPromptParameter.java
@@ -35,7 +35,7 @@ public enum OpenIdConnectPromptParameter {
     /**
      * No prompt parameter will be injected into the request.
      */
-    NOT_SENT,
+    UNSET,
 
     /**
      * this value is not used currently, kept as a placeholder if required for usage in the future.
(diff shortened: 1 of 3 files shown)


shahzaibj:
Can you elaborate on the naming conventions?

tanmaymanolkar1:
OpenIdConnectPromptParameter .NOT_SENT is used specifically to indicate that no prompt parameter would be injected into the Request. The OpenIdConnectPromptParameter.NONE naming was a bit ambiguous and this condition wasn't specifically being used initially, since there is a need to use this OIDC prompt parameter now, it would be better to correct the naming to resolve the ambiguity.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -35,6 +35,11 @@
     /**
      * No prompt parameter will be injected into the request.
      */
+    NOT_SENT,
```
kreedula, line 38: `NOT_SENT` seems a bit strange, how about `NOT_SET` , `NULL` . Shane suggested `UNSET` when we were discussing this as well

tanmaymanolkar1: @kreedula sure UNSET would be good!

iambmelt:
There is a document we need to update associated with these changes:
https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/docs/prompt_behavior.md

PR #1015, opened 2020-08-27
Revert getting token endpoint from well known config

shahzaibj:
This to fix the issue caused by Blackforest migrating to Public cloud as the well known config still returns the DE token endpoint for Blackforest cloud even after migration.

Also the other reason to revert this is to avoid potential simulated instance aware in the case of a tenanted authority.

The other reason to revert is because we don't know what would be expected behavior once cross cloud is implemented.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 Version 2.1.1
 ----------
 - Introduces result sharing to minimize duplicate_command errors
-- Device Code authorization endpoint is now sourced from .Well-Known config
+- No longer query well known config to obtain token endpoint - build it manually instead
 
 Version 2.1.0
 ----------
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
@@ -40,10 +40,10 @@ public class MicrosoftStsOAuth2Configuration extends AzureActiveDirectoryOAuth2C
     private static final String TAG = MicrosoftStsOAuth2Configuration.class.getSimpleName();
 
     private static final String ENDPOINT_VERSION = "v2.0";
-    private static final String FALLBACK_ENDPOINT_SUFFIX = "/oAuth2/v2.0";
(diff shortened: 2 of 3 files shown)


changelog.txt
```diff
@@ -1,7 +1,7 @@
 Version 2.1.1
 ----------
 - Introduces result sharing to minimize duplicate_command errors
-- Device Code authorization endpoint is now sourced from .Well-Known config
+- No longer query well known config to obtain token endpoint - build it manually instead
```
AdamBJohnsonx, line 4: I mean, if this is a revert, then nothing changes, right?

shahzaibj: Nothing changes for Device Code flow, hence I removed that line. But I added a new line about reverting using well known config for Token Endpoint as that change was made back in November 2019 here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/714

shahzaibj:
> PMD failed
> 
> :(

Fixed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1015/commits/5568d59f42e987c5f7a191f68b9c300bd5752c10

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Configuration.java
```diff
@@ -52,22 +52,15 @@
      * @return URL the authorization endpoint
      */
     public URL getAuthorizationEndpoint() {
-        final OpenIdProviderConfiguration openIdConfig = getOpenIdWellKnownConfigForAuthority();
-        if (openIdConfig != null) {
-            return getEndpointUrlFromAuthority(openIdConfig.getAuthorizationEndpoint());
-        }
         return getEndpointUrlFromRootAndSuffix(getAuthorityUrl(), FALLBACK_AUTHORIZE_ENDPOINT_SUFFIX);
```
kreedula, line 55: Rename to `AUTHORIZE_ENDPOINT_SUFFIX` ? and similar below

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1015/commits/161957ecc585b2fb06601d6ff19c96df7f3d494c

PR #1018, opened 2020-08-28
Place cap on the number failed request data in Last Request Telemetry at any given time

shahzaibj:
Address: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1081276

I've put a cap of 100 -- we could go much higher but I've seen that we are generally able to send data relevant to about 50-60 failed requests in a given header (due to header length constraints). So let's say we accumulate 100 items, then we send about 60 of those in the next network request, and we send the remaining 40 in the following one. So this way in the cases where we do accumulate too much telemetry, then we are able to send the most recent and most relevant telemetry in 2 network requests.

IMO, if we keep more than 100, let's say 200, then all of that may take 4 network requests to reach the server. At that point, the telemetry is probably not that relevant because it is old, and also most likely a developer doing debugging won't look 4 requests back to get all the telemetry. Also generally speaking, most devices that run into accumulating too much telemetry without actually sending it are the devices that keep getting the same error over and over again such as a device network not available error. So there isn't usually a need to know what the remaining errors in the past may have been.

So I think a cap of 100 seems most reasonable.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
@@ -50,7 +50,8 @@ public class LastRequestTelemetry extends RequestTelemetry {
     }
 
     /**
-     * Get a list of Failed Request objects.
+     * Get a list of Failed Request objects. The list returned here is unmodifiable. To add new
+     * elements to this list use the {@link LastRequestTelemetry#appendFailedRequest} method.
      *
      * @return an unmodifiable list of {@link FailedRequest} objects
      */
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
@@ -49,7 +49,12 @@ public class LastRequestTelemetry extends RequestTelemetry {
         failedRequests = new ArrayList<>();
     }
 
-    List<FailedRequest> getFailedRequests() {
+    /**
+     * Get a list of Failed Request objects.
+     *
+     * @return an unmodifiable list of {@link FailedRequest} objects
+     */
+    /* package */ List<FailedRequest> getFailedRequests() {
         return Collections.unmodifiableList(failedRequests);
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
@@ -30,9 +30,12 @@ import com.google.gson.annotations.SerializedName;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 public class LastRequestTelemetry extends RequestTelemetry {
+
+    final static int FAILED_REQUEST_CAP = 100;
 
     @SerializedName("silent_successful_count")
     private int silentSuccessfulCount;
@@ -47,7 +50,7 @@ public class LastRequestTelemetry extends RequestTelemetry {
     }
 
     List<FailedRequest> getFailedRequests() {
-        return failedRequests;
+        return Collections.unmodifiableList(failedRequests);
     }
 
     @Override
@@ -76,10 +79,22 @@ public class LastRequestTelemetry extends RequestTelemetry {
 
 
     void appendFailedRequest(final String apiId, final String correlationId, final String error) {
-        failedRequests.add(new FailedRequest(apiId, correlationId, error));
+        appendFailedRequest(new FailedRequest(apiId, correlationId, error));
     }
 
     void appendFailedRequest(final FailedRequest failedRequest) {
+        // this should usually not be greater than - at most should be equal to the cap
+        // because the only we to add to this list is via this append method.
+        // The only time this could be greater than the cap is for some existing devices that may
+        // have caught themselves in a bad state after accumulating too much telemetry in
+        // Shared Preferences. (of course prior to the cap being put in place).
+        // So will just take the last (most recent) 100 items here to get those out of the bad state,
+        // and also to avoid having too much telemetry in the cache going forward.
+        if (failedRequests.size() >= FAILED_REQUEST_CAP) {
+            final int beginIndex = failedRequests.size() - FAILED_REQUEST_CAP + 1;
+            final int endIndex = failedRequests.size();
+            failedRequests = failedRequests.subList(beginIndex, endIndex);
+        }
         failedRequests.add(failedRequest);
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/SharedPreferencesLastRequestTelemetryCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/SharedPreferencesLastRequestTelemetryCache.java
@@ -66,17 +66,17 @@ public class SharedPreferencesLastRequestTelemetryCache implements IRequestTelem
     public synchronized RequestTelemetry getRequestTelemetryFromCache() {
         final String methodName = ":getRequestTelemetryFromCache";
 
-        final String cacheValue = mSharedPreferencesFileManager.getString(LAST_TELEMETRY_OBJECT_CACHE_KEY);
-
-        if (cacheValue == null) {
-            Logger.info(TAG + methodName, "There is no last request telemetry saved in " +
-                    "the cache. Returning NULL");
-
-            return null;
-        }
-
         try {
-            LastRequestTelemetry lastRequestTelemetry = mGson.fromJson(cacheValue, LastRequestTelemetry.class);
+            final String cacheValue = mSharedPreferencesFileManager.getString(LAST_TELEMETRY_OBJECT_CACHE_KEY);
+
+            if (cacheValue == null) {
+                Logger.info(TAG + methodName, "There is no last request telemetry saved in " +
+                        "the cache. Returning NULL");
+
+                return null;
+            }
+
+            final LastRequestTelemetry lastRequestTelemetry = mGson.fromJson(cacheValue, LastRequestTelemetry.class);
 
             if (lastRequestTelemetry == null) {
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
```diff
@@ -47,7 +50,7 @@
     }
 
     List<FailedRequest> getFailedRequests() {
```
AdamBJohnsonx, line 52: Package-private, right?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
```diff
@@ -47,7 +50,7 @@
...
-        return failedRequests;
+        return Collections.unmodifiableList(failedRequests);
```
AdamBJohnsonx, line 53: This doesn't take a safe-copy, right?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
```diff
@@ -76,10 +79,22 @@ void resetSilentSuccessCount() {
...
+        // because the only we to add to this list is via this append method.
+        // The only time this could be greater than the cap is for some existing devices that may
+        // have caught themselves in a bad state after accumulating too much telemetry in
+        // Shared Preferences. (of course prior to the cap being put in place).
+        // So will just take the last (most recent) 100 items here to get those out of the bad state,
+        // and also to avoid having too much telemetry in the cache going forward.
+        if (failedRequests.size() >= FAILED_REQUEST_CAP) {
+            final int beginIndex = failedRequests.size() - FAILED_REQUEST_CAP + 1;
+            final int endIndex = failedRequests.size();
+            failedRequests = failedRequests.subList(beginIndex, endIndex);
```
AdamBJohnsonx, line 96: I'm not sure at all that this is what we want to do.  subList returns a view on the original list.  So, as I read it, over time you'll end up with an ever expanding chain of references to new sublists (see SubList.subList).

https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/util/ArrayList.java;drc=6bd07db6e4f0806b658fc44bfee5dbef2c409540;l=1007

If we're only doing 100 of these, it might be better to just declare your own array and use it like a circular buffer, even though it means you'll have more complex code when somebody requests a view.

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
shahzaibj: It is package private

shahzaibj: Are you looking for some comment indicating that or what?

AdamBJohnsonx: Because the authors left it out of the language, I've always taken it upon myself to say so using comments.

/* package */ void aPackagePrivateFunction()

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
shahzaibj: Not sure what you mean here, can you tell more?

AdamBJohnsonx: This is more about being sure that whoever is using this is OK with the list changing underneath them.

shahzaibj: Yes, they are okay with that. Added javadoc here as well about being unmodifiable: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1018/commits/d71f93b10e05f31e6d3768f07117004b30837621

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1018/commits/89e243e0f75a7d427e781d36a7687773162dd1da

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
shahzaibj: Yeah, I would say let's keep the code simple for now. What do you say?

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/SharedPreferencesLastRequestTelemetryCache.java
```diff
@@ -66,28 +66,35 @@ ISharedPreferencesFileManager getSharedPreferencesFileManager() {
...
             Logger.error(TAG + methodName,
                     "Last Request Telemetry deserialization failed", e);
             return null;
+        } catch (final OutOfMemoryError e) {
+            Logger.error(TAG + methodName,
+                    "Encountered OOM while restoring last request telemetry from " +
+                            "Shared Preferences. Clearing telemetry cache to recover and returning " +
+                            "null.", e);
+            mSharedPreferencesFileManager.clear();
+            return null;
```
AdamBJohnsonx, line 97: There's not really any recovery from OOM.

shahzaibj: Yeah, if we can't load the String in memory, then can't do much from there. So just clearing that telemetry from cache to allow the device/user to avoid the crash.

AdamBJohnsonx: But in the mean time, other operations have failed in potentially other threads.  You probably can't even call logger, and can't execute TAG+methodName.  We've already bounded the side of the returned telemetry - I don't know that we need to try this.

shahzaibj: Why can't we call logger and TAG? This is in catch block so it should execute right?

Even though we've bounded the telemetry, that means we're good for the future, but we still need to account for a scenario where the telemetry String in Shared Pref (from earlier before this fix gets out), may be too huge that Android isn't able to fit the String in memory. So in that case, we just clear the cache so it doesn't happen again and return null here as we don't have any telemetry that we can safely return from cache.

AdamBJohnsonx: You're OOM.  If logger needs to allocate, it might not be able to.  The new string that gets created by TAG+methodName might also cause an OOM.

iambmelt: Wouldn't we fail to allocate the large string thus keeping the heap still alive and able to keep allocating?

Thinking about the notes in this SO answer
https://stackoverflow.com/a/2680784

shahzaibj: But the `TAG+methodName` String is going to be much smaller so it should get memory allocated for it. The OOM we ran into here for telemetry was that JAVA was trying to allocated about 9MB 😁so we're trying to protect against that.

AdamBJohnsonx: The stack overflow people seem to also think that they only thing to do at that point is shut down the application.  But we're not the application, we're a library.  We could be handing back control to a JVM that's utterly destroyed and we really have no way of knowing what's gone wrong in the mean time.

iambmelt: In that case, should we wipe the SharedPrefs and rethrow the error?

iambmelt: That would seem to be the most-ideal from a state-perspective right? Clear out what we failed to allocate and don't continue in an undefined state?

AdamBJohnsonx: If we can.  I'm looking to shared preferences now, to see if there's a way to get the object size before we try to read it.

AdamBJohnsonx: If we can find the size of the string in advance, we can avoid this by putting a reasonable bound on it.

AdamBJohnsonx: Thing is, it's just an assumption that this allocation caused the OOM - we could be wrong here.

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1018/commits/57d235941d4ed4ffd451cb3e40197f10f30fa652

common/src/main/java/com/microsoft/identity/common/internal/eststelemetry/LastRequestTelemetry.java
shahzaibj: Synced offline - @AdamBJohnsonx will write up a circular buffer and we can pick it up in the future.

shahzaibj:
@AdamBJohnsonx All comments have been addressed!

AdamBJohnsonx (approved):

PR #1021, opened 2020-09-01
Consider controllers while throttling

shahzaibj:
Controllers were not being considered in the Equals comparison of commands object. In this PR, I've updated the logic to consider them so that if we get a request with the same parameters, however, different controllers then we do NOT throttle.

Consider this scenario:

- Request with parameters A comes in (uses BrokerLocalController as device is NOT joined) - Fails as device registration required for resource
- User registers device
- In less than 30 seconds since first request, Another Request with same parameters A comes in (uses BrokerJoinedAccountController as device is now joined) - Fails due to throttling as the parameters were matched.

So now with this change, we wouldn't throttle in the second request with same parameters as now we would consider the controller as well so the request thumbprint is different i.e. the equals comparison for the two commands object fails and hence we do NOT throttle in such a case.

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1367

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/BaseCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/BaseCommand.java
@@ -42,7 +42,6 @@ public abstract class BaseCommand<T> implements Command<T> {
     @EqualsAndHashCode.Exclude
     private CommandCallback callback;
 
-    @EqualsAndHashCode.Exclude
     private List<BaseController> controllers;
 
     @EqualsAndHashCode.Exclude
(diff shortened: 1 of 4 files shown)


AdamBJohnsonx:
Since what's going on here is really caching this is one way to try to skin this cat, but there is another, potentially bigger hammer that we may want to use.  What looks to be going on is that there is a state change on the device that's causing previous requests to have different results.  We should probably invalidate our caches on these state changes, since there may be other errors lurking here.

common/src/main/java/com/microsoft/identity/common/internal/commands/BaseCommand.java
```diff
@@ -42,7 +42,6 @@
     @EqualsAndHashCode.Exclude
     private CommandCallback callback;
 
-    @EqualsAndHashCode.Exclude
     private List<BaseController> controllers;
```
AdamBJohnsonx, line 45: How many controllers can we have?

shahzaibj: AFAIK, we only have 2 max at any given time (with the current design). i.e one for local flow and the other for broker flow.

PR #1029, opened 2020-09-11
Removing compile time warnings - Iteration 3

neeravagarwalmicrosoft:
Task: Removing compile time warnings

Action taken: 
1. Suppressed most of the warnings. 
2. Made generic type changes for few of the warnings.

Testing:

Common repository is compiling successfully with the changes
Tested these changes in OneAuth code by running the MSAL flow in Android test app.
Outcome:
Before this change: Common warnings: 195
After this change: Common warnings: 80


common/src/main/java/com/microsoft/identity/common/exception/BrokerCommunicationException.java
```diff
@@ -27,6 +27,8 @@
  * An exception that represents an error where MSAL cannot reach Broker (i.e. through Bind Service or AccountManager).
  */
 public class BrokerCommunicationException extends BaseException {
+    private static final long serialVersionUID = 4959278068787428329L;
+
```
saugup-ms, line 31: Where it is required? And from where this value comes.

AdamBJohnsonx: The compiler generates a warning because this class implements Serializable without an identifier.  I would guess that it's autogenerated by Android Studio based on the class structure.

If you're curious: https://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html

neeravagarwalmicrosoft: Serial version UID in a class is used to ensure that an object of the class is of the particular version of the class.
Yes, this is autogenerated by Android Studio based on the class structure using the plugin GenerateSerialVersionUID

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -302,10 +302,15 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
         if (isFoci) {
             targetCache = mFociCache;
         } else {
+
+            // Suppressing unchecked warning as the generic type was not provided for oAuth2Strategy
+            @SuppressWarnings(WarningType.unchecked_warning)
+            final String environment = oAuth2Strategy.getIssuerCacheIdentifier(request);
```
shoatman, line 308: Is this authority or environment?  Just not sure what this method returns hence the question.  No change to the code though... so not worried about this.

shoatman (approved):
Thanks Neerav.

saugup-ms (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
neeravagarwalmicrosoft: Hi Shane, this is because of the request parameter of type GenericAuthorizationRequest and not due to the environment. I'll mention this in the comment in the next PR.

Thanks

PR #1030, opened 2020-09-12
Add utilities to interact with broker host during UI Test

shahzaibj:
See title

Related: https://github.com/AzureAD/ad-accounts-for-android/pull/1377


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/RemoveBrokersBeforeTestRule.java
```diff
@@ -41,22 +42,19 @@ public Statement apply(final Statement base, final Description description) {
...
-                // CP may still be installed if device admin
-                if (companyPortal.isInstalled()) {
-                    TestContext.getTestContext().getTestDevice().getSettings().disableAdmin(DeviceAdmin.COMPANY_PORTAL);
-                    companyPortal.uninstall();
+                    if (broker.isInstalled()) {
+                        // The broker app will still be installed on the device if it is enabled
+                        // as a device admin. In this case, we need to disable the admin and then
+                        // uninstall.
+                        TestContext.getTestContext().getTestDevice()
+                                .getSettings().disableAdmin(broker.getAdminName());
```
AdamBJohnsonx, line 55: nit: Hm.  Would it break if we just disabled it whether or not if needs it?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    @Override
+    public void performDeviceRegistration(@NonNull final String username,
+                                          @NonNull final String password) {
+        performDeviceRegistrationHelper(username);
+
+        // Click the join btn
+        final UiObject joinBtn = UiAutomatorUtils.obtainUiObjectWithResourceIdAndEnabledFlag(
+                "com.microsoft.identity.testuserapp:id/buttonJoin", true
```
AdamBJohnsonx, line 62: nit: Should you build this from the package name as an instance var in the abstract class?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/RemoveBrokersBeforeTestRule.java
shahzaibj: Yeah, you actually can't disable it if it is NOT ad admin. This is disabling admin via UI....and the button for **disabling** would be _disabled_ if not an admin. It will break the tests.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -0,0 +1,274 @@
...
+                "Microsoft Authenticator - cert installer dialog appears.",
+                certInstaller.exists()
+        );
+
+        UiAutomatorUtils.handleButtonClick("android:id/button1");
+    }
+
+    @Override
+    public void createPowerLiftIncident() {
+        throw new UnsupportedOperationException("Not supported on Broker Host App :(");
```
AdamBJohnsonx, line 227: Nit:  Could we fix that?  I mean, it's our test app, so I suppose it can do whatever we want including buffering all the logs internally and displaying them, so it might not be worthwhile.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -0,0 +1,274 @@
...
+        );
+
+        Assert.assertTrue(showUpnDialog.exists());
+
+        try {
+            // Obtain the text on the UPN dialog box
+            final String[] upnDialogTextParts = showUpnDialog.getText().split(":");
+
+            // get the UPN if it is there, else return null (in case of error)
+            if (upnDialogTextParts[0].equalsIgnoreCase("UPN")) {
```
AdamBJohnsonx, line 262: "UPN".equals...

AdamBJohnsonx (approved):
A munch of minor comments.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: We could build it from the package name.....but the logic to build will still need to be in the class I think....because the logic here i.e. which buttons to click and the resource ids are completely different across brokers (Auth app, CP, broker host) so I think we can't have shared logic for this method. 

We do have shared logic for `join via join activity` and that one is living in the abstract class.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: We could may be consume the powerlift sdk in the broker host app....we should talk about this in the next team meeting.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
AdamBJohnsonx: I just see the repeated strings everywhere and have to wonder whether separating them would make it a little more self-documenting.  I was mainly thinking about a buildLocator method that took the suffix and added the package name.

shahzaibj: Yeah I agree, let me change it then. I could use this method I wrote for similar purpose some time ago: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/CommonUtils.java#L87

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1030/commits/6012894adad9fc883be444247ae7f9d6db733871

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1030/commits/19da90c74fbf695d71b4cc92a15bf10fc3097c92

PR #1032, opened 2020-09-12
Perform browser sign out on MSAL side

rpdome:
In the FLW sign out scenario, we've been trying to let the broker do all the work, and this includes launching default browser to Sign off the device.

However, [this change in API 29](https://developer.android.com/guide/components/activities/background-starts) seems to break this. Having said that, it's not totally broken - it still works, but unreliable. (e.g. if you sign in, then sign out right away, it seems to work fine. However, if you sign in, close the app (or launch another app), and then sign out, it doesn't seem to work).

This fix is to move that operation (that was being run in the background (auth) process) to foreground process.

I'll keep the change on the broker side for now since Android is smart enough to try to launch chrome twice - which means we do not need to bump protocol version.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -24,11 +24,17 @@ package com.microsoft.identity.common.internal.controllers;
 
 import android.accounts.AuthenticatorException;
 import android.accounts.OperationCanceledException;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ProviderInfo;
+import android.net.Uri;
 import android.os.Bundle;
 import android.os.RemoteException;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.WorkerThread;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.exception.BaseException;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -567,6 +572,39 @@ public void putValueInSuccessEvent(ApiEndEvent event, Boolean result) {
...
+            final Intent intent = new Intent(Intent.ACTION_VIEW);
+            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+            intent.setData(Uri.parse(AuthenticationConstants.Browser.LOGOUT_ENDPOINT_V2));
+            if (browserPackageName != null) {
+                intent.setPackage(browserPackageName);
+            }
+            context.startActivity(intent);
+
+        } catch (final ActivityNotFoundException e) {
+            Logger.info(TAG + methodName,
```
AdamBJohnsonx, line 603: Is info really loud enough for this message?  I think it's at least a warning.

rpdome: This means no browser is installed on the device (which is a valid case for some 'work devices')- nothing we can really do here.

AdamBJohnsonx: In that case, though, we wouldn't find a package name by examining the installed packages in BrowserSelector, right?  And we'll already log an error in those cases on line 590.  But if we didn't log that error, I think this would be a really strange occurrence.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -567,6 +572,39 @@ public void putValueInSuccessEvent(ApiEndEvent event, Boolean result) {
                 });
     }
 
+    /**
+     * Invoke the logout endpoint on the specified browser.
+     * If there are more than 1 session in the browser, an account picker will be displayed.
+     * (Alternatively, we could pass the optional sessionID as one of the query string parameter, but we're not storing that at the moment).
+     */
+    private void logOutFromBrowser(@NonNull final Context context,
```
AdamBJohnsonx, line 580: If you're going to add javadoc, add the method @param -eters.

AdamBJohnsonx (approved):
There's no "approve with suggestions" here, my only comments are about log level and making sure that we're using the appropriate volume when strange things happen with a minor javadoc quibble.  But I think there's a feature-ask we should capture to save the session id - is it worth a PBI so we don't lose it?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
rpdome: fair. I'll change this to warning.

rpdome:
I think you're suggesting adding a telemetry event? I added a comment as part of this [work item](https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1051454).

PR #1033, opened 2020-09-14
Add metadata to TSL result payload

iambmelt:
Adds additional metadata to _internal_ TSL-API -- request from @shoatman


AdamBJohnsonx:
Why do we want to do this?

iambmelt:
> Why do we want to do use a string instead of an enum?

See [this pr in MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1152/files#diff-dc8f3a585fd30bbede56d1d9713377d5R38) -- want to expose these values without making an `enum` in `common` public _in MSAL_. Since `enums` can't be extended, I opted to pass a string thru, and then redeclare that same string in the public package so there's a private (internal) declaration and public one which refers to it.

shoatman (approved):
lgtm  Thanks Brian

PR #1034, opened 2020-09-14
Adding Enum of MSALCPP in SdkType

tanmaymanolkar1:
Adding Enum of MSALCPP in SdkType, to help defferentiate between MSAL and MSALCPP.

Another aspect is to remove the dependency on Device.java, for the purpose of using dynamic x-client-SKU and x-client-Ver Values, which are to be passed around from CommandParameter and BrokerRequest.

These values are then fed into the command Dispatcher, since the values are per thread (i.e request).
It makes accessing these values easier form obscure locations (Device.java absence), and saves on a lot of code changes.

Associated changes AzureAD/ad-accounts-for-android#1380

Follow up Action Item :
- Migrating the Application Name and Version to DiagnosticContext for uniformity.


tanmaymanolkar1:
Associated changes https://github.com/AzureAD/ad-accounts-for-android/pull/1380

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -27,5 +27,19 @@
  */
 public enum SdkType {
     ADAL,
-    MSAL
+    MSAL,
+    MSALCPP;
+
+    private final String PRODUCT_NAME_MSAL = "MSAL.Android";
+
+    private final String PRODUCT_NAME_MSAL_CPP = "MSAL.xplat.Android";
```
kreedula, line 35: Move this to Authentication Constants

tanmaymanolkar1: [https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1034/commits/45fba567fbf3c42c2a50bb4229b00b969db134b4](url)

shahzaibj:
Is this still supposed to be draft?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -53,6 +54,7 @@
         final static String CLIENT_APP_NAME = "client_app_name";
         final static String CLIENT_APP_VERSION = "client_app_version";
         final static String CLIENT_VERSION = "client_version";
+        final static String CLIENT_SDKTYPE = "client_sdktype";
```
shahzaibj, line 57:
```suggestion
        final static String CLIENT_SDK_TYPE = "client_sdk_type";
```

shahzaibj, line 57: Do we actually need to send this to the Broker? IIRC, MSAL / ADAL we don't need to send it based on the current design around adapters. Perhaps it is need for CPP?

common/src/main/java/com/microsoft/identity/common/internal/logging/DiagnosticContext.java
```diff
@@ -85,4 +94,30 @@ private static boolean hasThreadId() {
     public static void clear() {
         REQUEST_CONTEXT_THREAD_LOCAL.remove();
     }
+
+    /**
+     * Function previously resided in Device.java, contains Constants
+     * Attaching the constants to Diagnostic Context, for the purpose of having them in one place.
```
shahzaibj, line 100: What's being attached to DiagnosticContext here? As far as I can see nothing.....and I think this method could easily stay inside `Device.java`. Is there a specific reason for moving it here?

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,24 @@
...
+
+    public String getProductName() {
+        if ((SdkType.ADAL == this) || (SdkType.MSAL == this)) {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        }
+
+    }
 }
```
shahzaibj, line 45: Add new line at end of file

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
tanmaymanolkar1: Yes it is need for CPP.

tanmaymanolkar1: It wasn't sent initially because Android.MSAL value was being hardcoded on the receiving end.

shahzaibj: Can you share CPP model for sending/unpacking broker request? Are they not using (custom) adapters?

common/src/main/java/com/microsoft/identity/common/internal/logging/DiagnosticContext.java
tanmaymanolkar1: There could be some more code optimizations done, and having all the constants in a single place would help. It would be then easier to have these constants accessible from a single place. Hence setting it up for the later use. @kreedula anything I missed?

kreedula: Well I would like to keep all of it in Device.java and use it in DiagonisticContext to put all the parameters inside this. And later , you get the map from `DiagnosticContext`.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -182,11 +182,13 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
...
                 !TextUtils.isEmpty(((MicrosoftTokenRequest) request).getBrokerVersion())) {
             headers.put(
-                    Device.PlatformIdParameters.BROKER_VERSION,
+                    AuthenticationConstants.PlatformIdParameters.BROKER_VERSION,
                     ((MicrosoftTokenRequest) request).getBrokerVersion()
             );
         }
-        headers.putAll(Device.getPlatformIdParameters());
+        headers.putAll(DiagnosticContext.getPlatformIdParameters());
+        headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
```
kreedula, line 201: `headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));`

`headers.put(AuthenticationConstants.SdkPlatformFields.VERSION,DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION));`

these can be part of `DiagnosticContext.getPlatformIdParameters()` as well right , why adding it seperately

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -468,7 +464,9 @@ private HttpResponse performPKeyAuthRequest(
         final String requestBody = ObjectMapper.serializeObjectToFormUrlEncoded(request);
         final Map<String, String> headers = new TreeMap<>();
         headers.put("client-request-id", DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
-        headers.putAll(Device.getPlatformIdParameters());
+        headers.putAll(DiagnosticContext.getPlatformIdParameters());
+        headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
+        headers.put(AuthenticationConstants.SdkPlatformFields.VERSION,DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION));
```
kreedula, line 469: same comment as below

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
kreedula: @shahzaibj  CPP will need to set SdkType as MSALCPP [here ]( https://github.com/AzureAD/microsoft-authentication-library-for-cpp/blob/develop/source/android/msal/app/src/main/java/com/microsoft/identity/internal/broker/BrokerRequestConverter.java#L79)and version too. Tanmay should make the changes there after this.

broker request will send this to broker (it will be MSAL or MSALCPP now )

shahzaibj: @kreedula I'm looking for the code where Broker will unpack the BrokerRequest that was sent to it via MSALCPP.

In broker, we currently have the `IBrokerRequestAdapter`, and we have implementations of it for ADAL/MSAL, but I don't see one for MSALCPP. Which adapter does MSALCPP use?

kreedula: it's the same format , i.e. common sends the request to broker . MSAL CPP just uses `BrokerMSALController" to call `acquireToken` or `acquireTokenSilent`

tanmaymanolkar1: 798ca9f7d8e3a969707905150533c290ff12eb97

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 798ca9f7d8e3a969707905150533c290ff12eb97

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -606,6 +609,8 @@ public static String initializeDiagnosticContext(@Nullable final String requestC
         final com.microsoft.identity.common.internal.logging.RequestContext rc =
                 new com.microsoft.identity.common.internal.logging.RequestContext();
         rc.put(DiagnosticContext.CORRELATION_ID, correlationId);
+        rc.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, sdkType);
+        rc.put(AuthenticationConstants.SdkPlatformFields.VERSION, sdkVersion);
```
AdamBJohnsonx, line 613: Can we clear the diagnostic context at the end of execution state?

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -89,7 +89,7 @@ public void validate() throws ArgumentException {
                     "mCallerPackageName", "Caller package name is not set"
             );
         }
-        if (SdkType.MSAL == getSdkType() &&
+        if ((SdkType.MSAL == getSdkType() || SdkType.MSALCPP == getSdkType()) &&
```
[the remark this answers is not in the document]
tanmaymanolkar1: The sdktype is hardcoded in the logic when creating the requests, so technically we should only encounter one of these values.

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,24 @@
...
+
+
+    public String getProductName() {
+        if ((SdkType.ADAL == this) || (SdkType.MSAL == this)) {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        }
+
+    }
```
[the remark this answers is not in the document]
tanmaymanolkar1: Same comment mentioned applies here as well, since we are hardcoding the values in every request, there isnt a way where we could encounter any funny values, and hence only the values used in the enum would be considered for getProductName.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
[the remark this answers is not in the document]
AdamBJohnsonx: Sure.  Now.  But that might not be true forever.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: 26cfa7eb0a274713270351aede7d18c256f05eef

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -198,6 +198,8 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
             );
         }
         headers.putAll(Device.getPlatformIdParameters());
+        headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
+        headers.put(AuthenticationConstants.SdkPlatformFields.VERSION,DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION));
```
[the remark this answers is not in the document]
tanmaymanolkar1: 26cfa7eb0a274713270351aede7d18c256f05eef

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -469,6 +465,8 @@ private HttpResponse performPKeyAuthRequest(
         final Map<String, String> headers = new TreeMap<>();
         headers.put("client-request-id", DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
         headers.putAll(Device.getPlatformIdParameters());
+        headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
+        headers.put(AuthenticationConstants.SdkPlatformFields.VERSION,DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION));
```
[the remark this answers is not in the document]
tanmaymanolkar1: 26cfa7eb0a274713270351aede7d18c256f05eef

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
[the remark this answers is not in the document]
shahzaibj: For now, this is an `enum` so I don't expect it to be anything other than the values currently declared in the `SdkType.java`. @AdamBJohnsonx not clear to me how this could be an unknow value....are you anticipating something specific?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1567,4 +1567,24 @@
...
+         */
+        public static final String PRODUCT = "x-client-SKU";
+
+        /**
+         * The String representing the sdk version.
+         */
+        public static final String VERSION = "x-client-Ver";
+
+        /**
+         * The Strings representing SdkType
```
shahzaibj, line 1583:
```suggestion
         * The String representing SdkType
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1567,4 +1567,24 @@
...
+         */
+        public static final String PRODUCT_NAME_MSAL = "MSAL.Android";
+
+        public static final String PRODUCT_NAME_MSAL_CPP = "MSAL.xplat.Android";
```
shahzaibj, line 1587: needs javadoc

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
shahzaibj, line 1583:
```suggestion
         * The String representing the MSAL SdkType.
```

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -168,6 +170,13 @@
     @SerializedName(SerializedNames.CLIENT_VERSION)
     private String mMsalVersion;
 
+    /**
+     * Sdk Type
```
shahzaibj, line 174:
```suggestion
     * Sdk Type.
```

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -89,7 +89,11 @@ public void validate() throws ArgumentException {
                     "mCallerPackageName", "Caller package name is not set"
             );
         }
-        if (SdkType.MSAL == getSdkType() &&
+        /**
+         * The check for SdkType.MSALCPP is added to ensure
+         * the uniformity of logic with SdkType.MSAL
+         */
```
shahzaibj, line 95: I believe this javadoc needs to be updated......and actually probably needs to be a comment instead of javadoc

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,30 @@
...
+
+    public boolean isCapableOfMSA() {
+        if ((this == SdkType.MSAL) || (this == SdkType.MSALCPP)) {
+            return true;
+        }
+        return false;
```
shahzaibj, line 49:
```suggestion
        return (this == SdkType.MSAL) || (this == SdkType.MSALCPP);
```

shahzaibj (changes requested):
Left some comments

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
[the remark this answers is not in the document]
AdamBJohnsonx: This enumeration looks like it exists on the interface between broker and MSAL - is that correct?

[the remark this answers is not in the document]
AdamBJohnsonx: This means that at some point this exists as a series of bytes in IPC.  You do not know what is going to land on the other side, since it can be newer or older than what you expect.  We're going to decouple this layer in the broker to allow for us to pivot the serialization mechanisms more smoothly, but this means that there will be cases where you will not be able to put the SDKType into a bucket with which you are familiar. What will you do in that case?

tanmaymanolkar1:
> Left some comments

addressed

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -468,6 +484,16 @@ public boolean isPowerOptCheckEnabled(){
             return this;
         }
 
+        /**
+         * Sdk Type
```
shahzaibj, line 488:
```suggestion
         * Sdk Type.
```

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -89,7 +89,9 @@ public void validate() throws ArgumentException {
                     "mCallerPackageName", "Caller package name is not set"
             );
         }
-        if (SdkType.MSAL == getSdkType() &&
+
+        // isCapableOfMSA() to ensure there is uniformity of logic with SdkType.MSALCPP and SdkType.MSAL
```
shahzaibj, line 93:
```suggestion
        // Check if SDK is capable of MSA to ensure there is uniformity of logic with SdkType.MSALCPP and SdkType.MSAL
```

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,31 @@
...
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        }
+
+    }
+
+    /**
+     * The check for SdkType.MSALCPP is added to ensure
+     * the uniformity of logic with SdkType.MSAL
```
shahzaibj, line 47: Needs a better javadoc

shahzaibj: Something along the lines of `Determines if the Sdk supports Microsoft Personal Accounts.`

tanmaymanolkar1: any specific suggestions for the change?
How about something on lines of needed check to ensure the request is for a MSA account.

tanmaymanolkar1: addressed

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -426,7 +427,9 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 @Override
                 public void run() {
                     final String correlationId = initializeDiagnosticContext(
-                            command.getParameters().getCorrelationId()
+                            command.getParameters().getCorrelationId(),
```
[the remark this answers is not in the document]
tanmaymanolkar1: Is it alright if I resolve this @AdamBJohnsonx , since we had a discussion about the thread local context?

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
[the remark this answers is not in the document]
tanmaymanolkar1: @AdamBJohnsonx In this case, this wont be a blocker since, for the mentioned cases : 

1. **new MSAL - old broker** : the old broker wont have the sdktype and even if the new MSAL sends the new sdktype (msal.cpp) the broker will neglect it and use the hardcoded value of sdktype (msal)

2. **old msal - new broker** : since old msal wont be sending the sdktype, the new broker has a default value of selecting msal as sdktype.

3. **old msal - old broker** : working fine

4. **new msal - new broker** : working fine

I am going to be testing these e2e scenarios just to make sure, we see the flow we are expecting to see, I would post the results here.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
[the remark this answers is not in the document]
AdamBJohnsonx: I don't see that they're related?

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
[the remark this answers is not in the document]
AdamBJohnsonx: Why is it OK if the old broker cannot accurately report the sdk type?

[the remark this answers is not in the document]
tanmaymanolkar1: Since the old broker is already deployed and is in production(the current implementation) is using the common which uses the hardcoded value? I mean to say, this added new functionality wont break the functionality of the broker versions already in the market today not having this change.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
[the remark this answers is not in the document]
tanmaymanolkar1: Oh my bad @AdamBJohnsonx ! I misconstrued the comment, I will factor out the command.getParameters()

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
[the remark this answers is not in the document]
AdamBJohnsonx: Let's just make sure that having the divergence won't crash the broker in the event of an either-way-mismatch, and you're good to go after that.  We'll take on making this forwards-compatible for the next SDK addition later, and revisit any parameter that's similarly dangerous.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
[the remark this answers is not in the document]
tanmaymanolkar1: 71971f7aba6ef749710983fa8508b9ebb864d48a

tanmaymanolkar1:
Testing is done, covered all cases, working as expected

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/CommandParameters.java
```diff
@@ -60,9 +60,8 @@
     @Expose()
     private String requiredBrokerProtocolVersion;
 
-    @Builder.Default
     @Expose()
-    private SdkType sdkType = SdkType.MSAL;
```
AdamBJohnsonx, line 65: Do you need to get rid of this parameter?

tanmaymanolkar1: yes because this was preventing picking up the value when I checked using the delombok'd code.

shahzaibj: Do we know why it wasn't being picked up? because you said earlier that when `null` was passed then the default was working as expected and setting it to `MSAL` -- so in what cases does it not work? and why?

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,30 @@
...
+        if ((SdkType.ADAL == this) || (SdkType.MSAL == this)) {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        }
+
+    }
+
+    /**
+     * Determines if the Sdk supports Microsoft Personal Accounts
```
shahzaibj, line 46:
```suggestion
     * Determines if the Sdk supports Microsoft Personal Accounts.
```

shahzaibj: Also need javadoc for return type here

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,30 @@
...
 public enum SdkType {
     ADAL,
-    MSAL
-}
\ No newline at end of file
+    MSAL,
+    MSALCPP;
+
+
+    public String getProductName() {
```
shahzaibj, line 36: Javadoc

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1567,4 +1567,27 @@
 
         public static final String USER_SIGNED_OUT_FROM_SHARED_DEVICE = "user_signed_out_from_shared_device";
     }
+
+    public static final class SdkPlatformFields {
```
shahzaibj, line 1571: Some javadoc for this public class would be good

tanmaymanolkar1: 170586b

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 170586b

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 9b05ccd

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -165,7 +165,7 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
                 public void run() {
                     try {
                         final CommandParameters commandParameters = command.getParameters();
-                        final String correlationId = initializeDiagnosticContext(commandParameters.getCorrelationId(), commandParameters.getSdkType().getProductName(), commandParameters.getSdkVersion());
+                        final String correlationId = initializeDiagnosticContext(commandParameters.getCorrelationId(), commandParameters.getSdkType() == null ? "" : commandParameters.getSdkType().getProductName(), commandParameters.getSdkVersion());
```
shahzaibj, line 168: Are there any other cases where this could happen (besides the test)? and should we pick a better default value for such scenarios?

AdamBJohnsonx: Realistically, we should accept any String that a client sends us and choose to sanitize it accordingly.  If that means accepting anything of a given pattern and converting everything else to UNKNOWN and pushing the transmitted value over telemetry as unknownClientSpecifier = <blah>, that's perfect.

tanmaymanolkar1: No, there wouldn't be any such scenarios, since these values are coded when the requests are being spawned. These exist only in the test case. and with this implementation the test cases are also taken care of and are executed successfully.

tanmaymanolkar1: There is also an additional null check for this field on the broker side, when rebuilding the CommandParameters from teh BrokerRequest.

AdamBJohnsonx: I've made a bug to remove type-sharing from this interface, since it's just a problem waiting to happen.

tanmaymanolkar1: @AdamBJohnsonx yes I agree, since this field is not exposed to user input, hence it is safeguarded by that. And like you suggested there is a refactor task needed with certain fields involved in interaction over wire and once things settle down we can club in this change as well

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/CommandParameters.java
tanmaymanolkar1: No, I meant that if the field is not sent over the wire, we have nothing to compare against null on the receiving end. The annotation as well wasn't a good addition, it was a part of an experimentation which I was trying out, and missed out on actually servicing it.

AdamBJohnsonx (approved):
I don't have a problem with this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
AdamBJohnsonx: @tanmaymanolkar1 - don't be so trusting. You could get anything over that interface and it shouldn't crash you.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/CommandParameters.java
shahzaibj: But if there is nothing, isn't there the initial value of `null` for all java objects? I mean the default should be kicking in whether you pass `null` to the builder or pass nothing

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shahzaibj: @tanmaymanolkar1 I think you can create a constant for `unknown` and use that here....I don't personally like the empty string

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -216,6 +219,7 @@ public void run() {
                             }
                             finalFuture.setCleanedUp();
                         }
+                        //DiagnosticContext.clear();
```
shahzaibj, line 222: Did we mean to do this or not? If not, then let's remove the comment.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -426,48 +430,55 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
-                            command.getParameters().getCorrelationId()
-                    );
-
-                    // set correlation id on parameters as it may not already be set
-                    command.getParameters().setCorrelationId(correlationId);
+                    try {
+                        final CommandParameters commandParameters = command.getParameters();
+                        final String correlationId = initializeDiagnosticContext(
+                                commandParameters.getCorrelationId(),
+                                commandParameters.getSdkType() == null ? "" : commandParameters.getSdkType().getProductName(),
```
shahzaibj, line 437: Same comment as above about empty string

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,34 @@
...
 /**
  * Enum to indicate if the request came from an app hosting Adal or Msal Sdk
  */
 public enum SdkType {
     ADAL,
-    MSAL
-}
\ No newline at end of file
+    MSAL,
+    MSALCPP;
```
shahzaibj, line 33:
```suggestion
    MSAL_CPP;
```
I think that may be more readable

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,34 @@
...
+
+    /**
+     * Function for mapping the SdkType to appropriate String
```
shahzaibj, line 36: Functions in java are called as `methods`

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,34 @@
...
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        }
+
+    }
+
+    /**
+     * Determines if the Sdk supports Microsoft Personal Accounts.
+     * @return True if Sdk supports Microsoft Personal Accounts, false otherwise.
```
shahzaibj, line 50: needs formatting

shahzaibj (commented):
Let's run the formatter on all these classes, and also can we write a Robolectric for this feature (that grabs the headers and ensures that correct values are being sent over the wire)?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: 38b8d704207fb078d883d644a73e0835fa8ded39

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: 38b8d704207fb078d883d644a73e0835fa8ded39

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 38b8d704207fb078d883d644a73e0835fa8ded39

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 38b8d704207fb078d883d644a73e0835fa8ded39

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: 38b8d704207fb078d883d644a73e0835fa8ded39

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/CommandParameters.java
AdamBJohnsonx: Is this coming through Parcelable?

tanmaymanolkar1: The Builder.Default is broken according to sum, and seems like people are recommending against using it. This might explain the builder not picking up the value passed.
https://github.com/rzwitserloot/lombok/issues/1347

tanmaymanolkar1: @AdamBJohnsonx  I dont think its coming through parcelable.

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,39 @@
...
+     * for the purpose of sending it to the telemetry.
+     */
+    public String getProductName() {
+        if ((SdkType.ADAL == this) || (SdkType.MSAL == this)) {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL;
+        } else if ((SdkType.MSAL_CPP == this)) {
+            return AuthenticationConstants.SdkPlatformFields.PRODUCT_NAME_MSAL_CPP;
+        } else {
+            // value is intended for test-cases, eg. CommandDispatcherTest.java.
+            return "";
```
shahzaibj, line 47: Aren't we using default of `unknown` in this case?

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
```diff
@@ -22,10 +22,39 @@
...
-}
\ No newline at end of file
+    MSAL,
+    MSAL_CPP,
+    UNKNOWN;
+
+    /**
+     * Method for mapping the SdkType to appropriate String
+     * for the purpose of sending it to the telemetry.
+     */
```
shahzaibj, line 39: return not documented

shahzaibj (approved):
Left some comments, else LGTM!

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: yes, we are using it in CommandDispatcher.java for submitSilent and beginInteractive methods. this field is primarily used to adhere to already written test-cases.

common/src/main/java/com/microsoft/identity/common/internal/request/SdkType.java
tanmaymanolkar1: c361c635cdbd853343a0ed18cc0b426d9c508037

PR #1040, opened 2020-09-17
Fix misc http issues

kkrishnareddy:
1. remove null key headers from response.
2. no post request if the content length is zero.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
@@ -50,7 +50,7 @@ public final class HttpResponse {
                         final Map<String, List<String>> responseHeaders) {
         mStatusCode = statusCode;
         mResponseBody = responseBody;
-        mResponseHeaders = responseHeaders;
+        mResponseHeaders = filterNullKeyHeaders(responseHeaders);
     }
 
     public HttpResponse(@Nullable final Date date,
@@ -60,7 +60,7 @@ public final class HttpResponse {
         mDate = date;
         mStatusCode = statusCode;
         mResponseBody = responseBody;
-        mResponseHeaders = headerFields;
+        mResponseHeaders = filterNullKeyHeaders(headerFields);
     }
 
     public Date getDate() {
@@ -102,4 +102,17 @@ public final class HttpResponse {
                 '}';
     }
     //CHECKSTYLE:ON
+}
+
+private Map<String, List<String>> filterNullKeyHeaders(@NonNull final Map<String, List<String>> headerFields)    
+{        
+    Map<String, List<String>> filteredHeaders = new HashMap<String, List<String>>();        
+    for(String key: headerFields.keySet())        
+    {           
+        if(key!= null)            
+        {                
+            filteredHeaders.put(key,headerFields.get(key));            
+        }
+    }
+    return filteredHeaders;
 }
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -102,4 +102,17 @@ public String toString() {
...
+{        
+    Map<String, List<String>> filteredHeaders = new HashMap<String, List<String>>();        
+    for(String key: headerFields.keySet())        
+    {           
+        if(key!= null)            
+        {                
+            filteredHeaders.put(key,headerFields.get(key));            
+        }
+    }
+    return filteredHeaders;
```
AdamBJohnsonx, line 117: Why are we doing this in MSAL common?  This seems to be a problem specific to the C++ applications, and should be fixed in the adapter layer there.

dabbcomputers (changes requested):
> 1. remove null key headers from response.
> 2. no post request if the content length is zero.

Don't complete the CR before getting approval from Shane. He wants to have  discussion with Adam first.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
AdamBJohnsonx: Thing is, this is actually fixed in MSAL-Cpp now.  So I'm not sure we need to do this here.  If we were going to take on moving this to a new field, I'd create one in HttpResponse called mStatusLine and put it there.  The question is - do we need to do this at all?

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -50,7 +50,7 @@ public HttpResponse(final int statusCode, final String responseBody,
                         final Map<String, List<String>> responseHeaders) {
         mStatusCode = statusCode;
         mResponseBody = responseBody;
-        mResponseHeaders = responseHeaders;
+        mResponseHeaders = filterNullKeyHeaders(responseHeaders);
```
AdamBJohnsonx, line 53:
```suggestion
        mResponseHeaders = responseHeaders;
```

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -60,7 +60,7 @@ public HttpResponse(@Nullable final Date date,
         mDate = date;
         mStatusCode = statusCode;
         mResponseBody = responseBody;
-        mResponseHeaders = headerFields;
+        mResponseHeaders = filterNullKeyHeaders(headerFields);
```
AdamBJohnsonx, line 63:
```suggestion
        mResponseHeaders = headerFields;
```

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
AdamBJohnsonx, line 117:
```suggestion
```

iambmelt (changes requested):
I don't think we want this change; as @AdamBJohnsonx states this looks be
1. The wrong repo for this code
2. Already being done in MSAL CPP

As a small aside, the code style isn't consistent with this repo.

Our braces are
```java
void braces() {
    // like this
}
```

and

```java
void braces() 
{
    // Not like this
}
```

And we use `final` wherever possible

iambmelt:
Closing due to inactivity

PR #1041, opened 2020-09-18
Added a check for UNSET OpenIdConnectPromptParameter value

tanmaymanolkar1:
If the new Msal versions send UNSET Value to old Brokers, the UNSET value would be converted to null and then converted back to the UNSET OpenIdConnectPromptParameter value.

Older versions of MSALs would fare well with new brokers as no deleteions were made from the OpenIdConnectPromptParameter values in the newer Brokers

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
@@ -108,7 +108,7 @@ public class MsalBrokerRequestAdapter implements IBrokerRequestAdapter {
                         parameters.getExtraQueryStringParameters() != null ?
                                 QueryParamsAdapter._toJson(parameters.getExtraQueryStringParameters())
                                 : null
-                ).prompt(parameters.getPrompt().name())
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -107,8 +107,8 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Inte
                 .extraQueryStringParameter(
                         parameters.getExtraQueryStringParameters() != null ?
                                 QueryParamsAdapter._toJson(parameters.getExtraQueryStringParameters())
-                                : null
-                ).prompt(parameters.getPrompt().name())
+                                : null)
+                .prompt(("UNSET".equals(parameters.getPrompt().name())) ? null : parameters.getPrompt().name())
```
rpdome, line 111: please use the enum.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
shahzaibj: +1

PR #1046, opened 2020-09-22
Cut down logs

rpdome:
This is to declutter broker logs. 

As we're moving to a model where we would always try to decrypt first, logging full error will generate a lot of noises.

### DIFF
--- a/common/versioning/version.properties
+++ b/common/versioning/version.properties
@@ -1,3 +1,3 @@
 #Thu Aug 02 12:03:16 PDT 2018
-versionName=3.0.2
+versionName=3.0.3-RC1
 versionCode=1
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -374,7 +374,7 @@ public class StorageHelper implements IStorageHelper {
         try {
             bytes = getByteArrayFromEncryptedBlob(data);
         } catch (Exception e) {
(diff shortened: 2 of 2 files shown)


AdamBJohnsonx:
Should we log the line at all in that case?

> This is to declutter broker logs.
> 
> As we're moving to a model where we would always try to decrypt first, logging full error will generate a lot of noises.

Should we just remove the log line in that case, limit its rate, or improve it?  As it is, I don't know what it's telling me about, even, without the stack trace other than something unspecified isn't encrypted.

rpdome:
> Should we log the line at all in that case?
> 
> > This is to declutter broker logs.
> > As we're moving to a model where we would always try to decrypt first, logging full error will generate a lot of noises.
> 
> Should we just remove the log line in that case, limit its rate, or improve it? As it is, I don't know what it's telling me about, even, without the stack trace other than something unspecified isn't encrypted.

We could. Let me do that.

AdamBJohnsonx (approved):

iambmelt (approved):

common/versioning/version.properties
```diff
@@ -1,3 +1,3 @@
 #Thu Aug 02 12:03:16 PDT 2018
-versionName=3.0.2
+versionName=3.0.3-RC1
```
shahzaibj, line 2: We are doing another release right now, just for this?

shahzaibj (approved):

tanmaymanolkar1 (approved):

common/versioning/version.properties
rpdome: Yes, otherwise we'll lose a lot of logs in Broker due to this noise.
I'm not planning to publish this to maven repo though.

shahzaibj: Okay

PR #1049, opened 2020-09-23
Merge 3.0.2 + 3.0.3 to Master

rpdome:


changelog.txt
```diff
@@ -1,3 +1,11 @@
+Version 3.0.3
+----------
+- Cut down logs (#1046)
+
+Version 3.0.2
```
tanmaymanolkar1, line 5: I have updated the changelog updates on 3.0.2 release branch.

tanmaymanolkar1: tag : [v3.0.2](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/releases/tag/v3.0.2)

shahzaibj:
Do we need to update Device,java? I guess we're not doing MSAL release (only Broker) so probably not.

shahzaibj (approved):

rpdome:
> Do we need to update Device,java? I guess we're not doing MSAL release (only Broker) so probably not.

Seems like We missed it in 2.0.1 also. 😢  It's still using 2.0.0 (having said that 2.0.1 is a hotfix that was shipped right after 2.0.0 - so I think it should be okay, no need to cut a new build specifically for this)

I'm wondering if we could turn this into a module that allows the caller (MSAL, Broker) to 'plug' its own IDevice class during initialization.

shahzaibj:
> > Do we need to update Device,java? I guess we're not doing MSAL release (only Broker) so probably not.
> 
> Seems like We missed it in 2.0.1 also. 😢 It's still using 2.0.0 (having said that 2.0.1 is a hotfix that was shipped right after 2.0.0 - so I think it should be okay, no need to cut a new build specifically for this)
> 
> I'm wondering if we could turn this into a module that allows the caller (MSAL, Broker) to 'plug' its own IDevice class during initialization (and we can pull the version number directly from BuildConfig.VERSION_NAME in MSAL/Broker).

I believe @tanmaymanolkar1 was recently doing refactoring around this, and the version is no longer actually pulled from Device.java.....not sure what is the status of that PR. @tanmaymanolkar1 did that pr get merged?

tanmaymanolkar1:
> > > Do we need to update Device,java? I guess we're not doing MSAL release (only Broker) so probably not.
> > 
> > 
> > Seems like We missed it in 2.0.1 also. 😢 It's still using 2.0.0 (having said that 2.0.1 is a hotfix that was shipped right after 2.0.0 - so I think it should be okay, no need to cut a new build specifically for this)
> > I'm wondering if we could turn this into a module that allows the caller (MSAL, Broker) to 'plug' its own IDevice class during initialization (and we can pull the version number directly from BuildConfig.VERSION_NAME in MSAL/Broker).
> 
> I believe @tanmaymanolkar1 was recently doing refactoring around this, and the version is no longer actually pulled from Device.java.....not sure what is the status of that PR. @tanmaymanolkar1 did that pr get merged?

No the PR is not yet merged, we can ship it in the next release, as it is not an urgent issue, the issue also needs some addition in the msal side so I will need to co-ordinate the release for both. For now we can go ahead without adding it in this release.

PR #1050, opened 2020-09-23
Create main.yml

shoatman:
Adding github action to run the snapshot build when a push to dev branch occurs.  Pushes to dev branch should only happen when a PR completes or during release process.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,28 @@
+# This is a basic workflow to help you get started with Actions
+
+name: dev_updated
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the dev branch
+on:
+  push:
+    branches: [ dev ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - name: Azure Pipelines Action
+        uses: Azure/pipelines@v1
+        with:
+          # Fullyqualified URL to the Azure DevOps organization along with project name(eg, https://dev.azure.com/organization/project-name or https://server.example.com:8080/tfs/DefaultCollection/project-name)
+          azure-devops-project-url: 'https://dev.azure.com/IdentityDivision/IDDP'
+          # Name of the Azure Pipline to be triggered
+          azure-pipeline-name: 'Nightly Common VSTS Release'
+          # Paste personal access token of the user as value of secret variable:AZURE_DEVOPS_TOKEN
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
+# This is a basic workflow to help you get started with Actions
```
AdamBJohnsonx, line 1: Maybe this comment should be updated. ;-)

shoatman: It's a template.  Not too worried :-)

shahzaibj:
@shoatman When a PR completes and merges into dev, then the PR could contain multiple commits within it. Will the build run for each commit so after merge there are N builds executed simultaneously? Or will there just be one build for the merge commit?

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+name: dev_updated
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the dev branch
+on:
+  push:
+    branches: [ dev ]
```
shahzaibj, line 9: Is there a way we can restrict it to only run on merge commits for PRs? Otherwise how many builds will be kicked off if a PR with n commits merges into dev? Will it be `n + 1` builds kicked off at once?

shoatman:
@shahzaibj - From what I understand from the documentation a pull request merge will result in a push to dev.  I think i need to test to be sure though.

shoatman:
I'll update this once I see it in action, so this isn't something that we're committing to forever.  Just need to merge in order to be able to test.

AdamBJohnsonx (commented):
I don't see any reason to not try this.

shahzaibj (approved):
Just one comment, else LGTM

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Azure Pipelines Action
+        uses: Azure/pipelines@v1
+        with:
+          # Fullyqualified URL to the Azure DevOps organization along with project name(eg, https://dev.azure.com/organization/project-name or https://server.example.com:8080/tfs/DefaultCollection/project-name)
+          azure-devops-project-url: 'https://dev.azure.com/IdentityDivision/IDDP'
+          # Name of the Azure Pipline to be triggered
+          azure-pipeline-name: 'Nightly Common VSTS Release'
+          # Paste personal access token of the user as value of secret variable:AZURE_DEVOPS_TOKEN
+          azure-devops-token: '${{ secrets.IDDP_PIPELINE }}'
```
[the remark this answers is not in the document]
shoatman: PR's for common won't merge to dev until they pass CI.  This workflow runs after the merge to dev has happened.  Not as part of validating the code as part of CI.  At least in theory.  If this pipeline fails, which it shouldn't then we'll see that Azure DevOps.  We can notify on failure within the pipeline if we want to.

PR #1055, opened 2020-09-25
Fix a potential issue with command dispatcher.

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -110,48 +115,68 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
...
+                    Telemetry.getInstance().flush(correlationId);
+                    EstsTelemetry.getInstance().flush(command, commandResult);
+                    finalFuture.setResult(commandResult);
+                    //Return the result via the callback
+                } catch (Throwable t) {
+                    finalFuture.setException(new ExecutionException(t));
+                } finally {
+                    final ResultFuture mapFuture = sExecutingCommandMap.remove(command);
+                    if (mapFuture == null) {
+                        // If this has happened, the command that we started with has mutated.  We will
```
rpdome, line 167: May I ask what you meant by 'mutated' here? In what (hypothetical) scenario this might happen?

AdamBJohnsonx: If the object used as the key in the map is changed in such a way to cause it no longer to have the same value via equals/hashCode, then it will not be found in the map when we go to remove it.

rpdome: (also makes me wonder if the command object itself it a good key here)

AdamBJohnsonx: We could generate a key from the command object and remove that worry, too.

shahzaibj: Did we find anything (so far) within the command that could change during execution?

rpdome: That might be better, but I think that'll be big and for now this fix should prevent NPE in the meantime. Let's put a todo here and create a tracking work item maybe?

AdamBJohnsonx: > Did we find anything (so far) within the command that could change during execution?

Nothing obvious.  If you look at the other CR, I put a bunch of safe-copy in place where I found it, but the scope is huge.  The thing with this is that it should be safe even if we write bugs into it, and that's really best.

shahzaibj: Yup, agreed!

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
AdamBJohnsonx: I'm actually going to wait on this until we have a unit test.  Brian and I are both trying to write some so there are more eyes on it.

AdamBJohnsonx: And if anybody else wants to play this game, more different viewpoints are better.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -74,12 +78,31 @@
...
-    private static final ConcurrentMap<BaseCommand, ResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
+    private static ConcurrentMap<BaseCommand, ResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
+
+    /**
+     * MUST be executed under the mapAccessLock;
+     * @param command
+     */
+    // Suppressing rawtype warnings due to the generic type BaseCommand
+    @SuppressWarnings(WarningType.rawtype_warning)
+    private static void cleanMap(BaseCommand command) {
```
iambmelt, line 94: Can we javadoc why this function is required? Otherwise not apparent to those who didn't debug this 😄

iambmelt: 6d73792efa129d69a0aaa08c94322f1dd09e047f

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
AdamBJohnsonx: Brian and I have added some tests.  I've put the todo item in place.

iambmelt (approved):

iambmelt:
Oh no! There looks to be some non-determinism happening here

```
com.microsoft.identity.common.CommandDispatcherTest > testSubmitSilentWithParamMutation[test(AVD) - 5.0.2] FAILED 
	java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
com.microsoft.identity.common.HttpUrlConnectionFactoryTest > null[test(AVD) - 5.0.2] SKIPPED 
> Task :common:connectedDebugAndroidTest FAILED
```

AdamBJohnsonx:
I see it.

common/src/main/java/com/microsoft/identity/common/internal/result/ResultFuture.java
```diff
@@ -129,4 +129,5 @@ public synchronized void whenComplete(@NonNull final BiConsumer<T, Throwable> co
 
         mConsumers.add(consumerToAdd);
     }
+
```
AdamBJohnsonx, line 132:
```suggestion
```

PR #1056, opened 2020-09-25
An initial ruleset for adal using common.

AdamBJohnsonx:


shahzaibj:
Shouldn't this live in the adal repo?

AdamBJohnsonx:
> Shouldn't this live in the adal repo?

Moved to that package.

PR #1060, opened 2020-09-28
Apply Fixes from #1055

iambmelt:
Applies the changes in #1055 to a changes from `release/3.0.3` and merges to `3.0.4`

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -57,26 +57,6 @@ public class CommandDispatcherTest {
 
     private static final String TEST_RESULT_STR = "test_result_str";
 
-    /*
-     * Tests:
-     * - sumbit command latched command.
-     * - submit duplicate latched commadn
-     * - Alter command 1 so that it will not longer be .equals command2
-     * - submit another duplicate.
-     * - release the latch
-     * - all the commands should reive results
-     * - nothing should be left in the map
-     *
-     * submit command
-     * command throws an Error during execution
-     * the callback should still occur, with an excecutionexception wrapping the error
-     *
-     * submit null
-     * NPE
-     *
-     *
-     */
-
     @Test
     public void testCanSubmitSilently() throws InterruptedException {
         final CountDownLatch testLatch = new CountDownLatch(1);
@@ -149,10 +129,7 @@ public class CommandDispatcherTest {
         CommandDispatcher.submitSilent(getTestCommand(testLatch2));
         testLatch2.await();
 
-        Field f = CommandDispatcher.class.getDeclaredField("sExecutingCommandMap");
-        f.setAccessible(true);
-        Map map = (Map) f.get(null);
-        Assert.assertEquals(0, map.size());
+        Assert.assertEquals(0, CommandDispatcher.outstandingCommands());
     }
 
     @Test
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -81,6 +81,35 @@ public class CommandDispatcherTest {
         testLatch.await();
     }
 
+    @Test
+    public void testSubmitSilentWithParamMutation() {
+        // TODO
+    }
+
+    @Test
+    public void testSubmitSilentWithException() {
+        // TODO Use the ExceptionCommand
+    }
+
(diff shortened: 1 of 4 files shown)


common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -0,0 +1,298 @@
...
+                    }
+                }, 0, submitLatch, submitLatch1);
+        FinalizableResultFuture<CommandResult> f = CommandDispatcher.submitSilentReturningFuture(testCommand);
+        submitLatch1.await();
+        testCommand.value = 2;
+        submitLatch.countDown();
+        testLatch.await();
+        Assert.assertTrue(f.isDone());
+        Assert.assertEquals(TEST_RESULT_STR, f.get().getResult());
+        f.isCleanedUp();
```
rpdome, line 134: This looks like something we can use RoboTestUtils.flushScheduler() for?

(no need to make changes in this PR)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -73,84 +77,145 @@
...
+    private static ConcurrentMap<BaseCommand, FinalizableResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
+
+    /**
+     * Remove all keys that are the command reference from the executing command map.  Since if they key has
+     * been changed, remove will not work, construct a new map and add all keys that are not identically
+     * that key into the new map.
+     *
+     * @param command the command whose identity to use to cleanse the map.
+     */
+    private static void cleanMap(BaseCommand command) {
```
rpdome, line 92: synchronized?

AdamBJohnsonx: Maybe document that it must be used under the mapAccessLock?

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
AdamBJohnsonx: Hah.  Yeah, I could have just gotten a reference on the executor and shut it down, I suppose.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
rpdome: Can we wrap them here? synchronize block is reentrant right?

rpdome: I'm fine either way as soon as we can make sure the future dev are aware of this or can do the right thing)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -73,84 +77,145 @@
...
-    private static final ConcurrentMap<BaseCommand, ResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
+
+    private static final Object mapAccessLock = new Object();
+    @GuardedBy("mapAccessLock")
+    private static ConcurrentMap<BaseCommand, FinalizableResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
+
+    /**
+     * Remove all keys that are the command reference from the executing command map.  Since if they key has
+     * been changed, remove will not work, construct a new map and add all keys that are not identically
+     * that key into the new map.
```
AdamBJohnsonx, line 88:
```suggestion
     * that key into the new map.  <strong>MUST</strong> only be used under the mapAccessLock.
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
AdamBJohnsonx: Yeah we could.  I'l like to dispose of this completely once we can stabilize the key space for this map either by generating stable keys or making the objects immutable.

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
rpdome: could we add a TODO to migrate to that one (and get rid of the new future class) at some point in the future? (e.g. when you're working on the root cause fix)

rpdome (approved):

AdamBJohnsonx (approved):

PR #1068, opened 2020-10-04
Do not allow for missing secret key data to cause an NPE

AdamBJohnsonx:
Check for null data and do not decrypt it.
Properly synchronize and safe-copy the key array.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -22,9 +22,13 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.adal.internal;
 
+import androidx.annotation.GuardedBy;
+import androidx.collection.ArrayMap;
+
 import com.microsoft.identity.common.WarningType;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -88,8 +93,8 @@
      *
      * @return byte[] secret data
      */
-    public byte[] getSecretKeyData() {
-        return mSecretKeyData.get();
```
shahzaibj, line 92: So are we no longer using getters and setters? Do we still need them?

AdamBJohnsonx: we're using them.  Just adding synchronization/safe-copy.

shahzaibj: Ohh okay, so `Arrays.copyOf()` acts as a copy + get.....hadn't used this one before

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -590,7 +590,13 @@ public SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExceptio
...
 
             case ADAL_USER_DEFINED_KEY:
-                return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
+                final byte[] secretKeyData = AuthenticationSettings.INSTANCE.getSecretKeyData();
+                if (secretKeyData == null) {
+                    Logger.error("StorageHelper:loadSecretKey", "Went looking for a key that wasn't there.", null);
+                    return null;
+                }
+                SecretKey key = getSecretKey(secretKeyData);
+                return key;
```
shahzaibj, line 599:
```suggestion
                return getSecretKey(secretKeyData);
```

shahzaibj (approved):
Just one suggestion.....LGTM

andrewchoi5 (approved):

PR #1072, opened 2020-10-07
Introduce custom count down latch for UI Automation Tests

shahzaibj:
This count down latch has a custom `await` method that accepts an enum type of wait duration (instead of specifying time and unit directly). Also this `await` method will fail the test if the `await` actually times out.


rpdome (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/TokenRequestLatch.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.junit.Assert;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A custom {@link CountDownLatch} to be used during UI Automation Token Requests.
+ */
+public class TokenRequestLatch extends CountDownLatch {
```
AdamBJohnsonx, line 35: Can we contain rather than extend here?

shahzaibj: As in add `CountDownLatch` as a member field to this class?

AdamBJohnsonx: And, for real, you should really look at moving this whole thing to a Future, where you don't need to worry about constructing whichever version of a latch that you want, or if we later want to change our synchronization on this we don't have to change all the code that uses it.

shahzaibj: Okay, let me take a look into it

AdamBJohnsonx: But extending the java base classes isn't always safe.  We got in trouble doing this this involving HashMap once, and I'm leery of it forevermore.

shahzaibj: I'll be looking into the synchronization again as part of this work item: https://dev.azure.com/IdentityDivision/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1184801

(So I only have to do it once for both ADAL / MSAL).

shahzaibj: See comment above, going to merge this in for now to do some work on running tests on pipeline. Will come back to it.

PR #1081, opened 2020-10-09
add Coverage Report tasks for Common

ayushmeghwani:
This Pull Request is related to ITEM - https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1152996

This PR adds the tasks which generates tasks for all buildFlavour combination to generate Coverage Report.

To generate the Coverage report:

Run the task
Then run -> ./gradlew debugCommonBothTestCoverageReport (where debug(or release) is Flavour, and Both(or Android or unit) is Test Type)
Reports will be available in report folder


common/build.gradle
```diff
@@ -156,6 +167,54 @@ task sourcesJar(type: Jar) {
...
+    def buildTypes = android.buildTypes.collect { type ->
+        type.name
+    }
+    def productFlavors = android.productFlavors.collect { flavor ->
+        flavor.name
+    }
+    // When no product flavors defined, use empty string
+    if (!productFlavors) productFlavors.add('')
+    productFlavors.each { productFlavorName ->
+        buildTypes.each { buildTypeName ->
```
shahzaibj, line 183: Instead of having a nested loop here for build types and flavors, it would be better to just loop over the `build variants` -- that way you don't have nested loop and also don't need a placeholder empty string for flavors.

common/build.gradle
```diff
@@ -156,6 +167,54 @@ task sourcesJar(type: Jar) {
...
+            def sourceName
+            if (!productFlavorName) {
+                sourceName = "${buildTypeName}"
+            } else {
+                sourceName = "${productFlavorName}${buildTypeName.capitalize()}"
+            }
```
shahzaibj, line 189: see my comment, `sourceName` can then just be the variant name

common/build.gradle
ayushmeghwani: added buildVariants to loop over, have a look.

shahzaibj (approved):

PR #1085, opened 2020-10-13
Fix querying broker packages on SDK 30

shahzaibj:
Address: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1177201 (CORPNET REQUIRED)


common/src/main/AndroidManifest.xml
```diff
@@ -17,6 +17,9 @@
     <!-- Required for API Level 30 to make sure we can detect browsers that support custom tabs -->
```
rpdome, line 17: we might want to move this down to <intent>

common/src/main/AndroidManifest.xml
```diff
@@ -17,6 +17,9 @@
...
     <!-- https://developers.google.com/web/updates/2020/07/custom-tabs-android-11#detecting_browsers_that_support_custom_tabs -->
     <queries>
+        <package android:name="com.azure.authenticator" />
```
rpdome, line 20: can we add comment specifying why we're doing this?

rpdome (approved):
Thanks for picking this up Shahzaib!

PR #1087, opened 2020-10-16
Do not silently swallow exception in requestAuthorization(), Launch AuthorizationActivity with different flags for broker and non-broker auth

rpdome:
1. Do not silently swallow exception in requestAuthorization(),
- I accidentally hit this error when I disabled chrome and re enable it (making the 'emulator browser' the default browser). MSAL throws an NPE on the future object. 
- After debugging, I found out that we only pull 'default browser that ~~supports custom tabs~~ that is in the browser safe list'. 
- In this case, the emulator browser ~~does not support custom tabs~~ is not in the safe list, and chrome is no longer the default browser, so a ClientException with NO_AVAILABLE_BROWSER_FOUND is thrown.
- However, that NO_AVAILABLE_BROWSER_FOUND is silently swallowed. the future object is never set and therefore we hit an NPE down the line.

2. Launch AuthorizationActivity with different flags for broker and non-broker auth
- This is a regression from https://github.com/AzureAD/ad-accounts-for-android/pull/1399, will only happen if MSAL is used inside the broker host app. (ipphone CP is in the progress to consume MSAL).
- AuthorizationActivity has to be launched with Intent.FLAG_ACTIVITY_CLEAR_TASK too, otherwise Authorization Activity on finish can land on the broker host app's activity (if it's in the background).

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
@@ -26,6 +26,9 @@ import android.content.Context;
 import android.content.Intent;
 import android.text.TextUtils;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.WorkerThread;
+
 import com.microsoft.identity.common.WarningType;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.exception.ArgumentException;
@@ -69,9 +72,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
-
-import androidx.annotation.NonNull;
-import androidx.annotation.WorkerThread;
 
 import lombok.EqualsAndHashCode;
 
@@ -409,8 +409,7 @@ public class LocalMSALController extends BaseController {
                         .putApiId(TelemetryEventStrings.Api.LOCAL_GET_ACCOUNTS)
         );
 
-        @SuppressWarnings(WarningType.unchecked_warning)
-        final List<ICacheRecord> accountsInCache =
+        @SuppressWarnings(WarningType.unchecked_warning) final List<ICacheRecord> accountsInCache =
                 parameters
                         .getOAuth2TokenCache()
                         .getAccountsWithAggregatedAccountData(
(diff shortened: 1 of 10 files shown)


shahzaibj:
What is emulator browser?

Also yes we only pull default browser, but I don't think it matters whether it supports custom tabs or not. As long as the default browser is in the browser safe list it should work. I suspect the browser you used wasn't in the safe list and hence it didn't work. Do you know which browser you were using?

rpdome:
> What is emulator browser?
> 
> Also yes we only pull default browser, but I don't think it matters whether it supports custom tabs or not. As long as the default browser is in the browser safe list it should work. I suspect the browser you used wasn't in the safe list and hence it didn't work. Do you know which browser you were using?

Screen Shot 2020-10-15 at 5 16 32 PM

ah good point on the safe list. I don't think this one is on it.

rpdome:
@shahzaibj  updated description.

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -126,21 +126,14 @@ public OAuth2Strategy(GenericOAuth2Configuration config,
      */
     public Future<AuthorizationResult> requestAuthorization(
```
AdamBJohnsonx, line 127: Is this return now @NonNull?

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -213,7 +214,7 @@ public AcquireTokenResult acquireToken(
     private AuthorizationResult performAuthorizationRequest(@NonNull final OAuth2Strategy strategy,
                                                             @NonNull final Context context,
                                                             @NonNull final InteractiveTokenCommandParameters parameters)
-            throws ExecutionException, InterruptedException, ClientException {
+            throws ExecutionException, InterruptedException, ClientException, UnsupportedEncodingException {
```
AdamBJohnsonx, line 217: How are we ending up with UnsupportedEncoding here?

AdamBJohnsonx: It looks like it's declared as thrown from requestAuthorization in both implementations, but is not actually thrown there.

AdamBJohnsonx (approved):

PR #1088, opened 2020-10-21
[IPC Part 1] Separate communication logic from business logic

rpdome:
This is part 1 of my Broker API changes.
Full spec here: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2098

**Changes**
1. Extracted `BrokerOperationExecutor` from `BrokerMsalController`
2. Converted `MicrosoftAuthServiceStrategy` (which was exclusive for MSAL-Broker communication) into a generic `BoundServiceStrategy`
3. Extracted base class `BoundServiceClient` from `MicrosoftAuthClient` (so that it can be used with the generic `BoundServiceStrategy` - BoundServiceStrategy takes this object as its parameter)
4. Added `BrokerOperationBundle` to act as a 'bridge' between business logic and communication logic.
5. Added annotations in ResultAdapter classes (and made some changes to eliminate warnings)

To scope down the work, I'm disabling AccountManager and ContentProvider strategies in BrokerMsalController for now.

Verified by a smoke test (acquire token interactive + silent + remove account) in both shared and non-shared mode. joined flow.


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -585,4 +627,8 @@ public boolean getDeviceModeFromResultBundle(@NonNull final Bundle bundle) throw
 
         return bundle.getBoolean(BROKER_DEVICE_MODE);
     }
+
+    public ClientException getExceptionForEmptyResultBundle() {
```
rpdome, line 631: will add NonNull.

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static com.microsoft.identity.common.exception.ClientException.BOUND_SERVICE_UNAVAILABLE_OR_NOT_SUPPORTED;
+
+/**
+ * Base class for a Bound Service client.
+ * A separate client is required for each AIDL interface (android.os.IInterface)
+ */
+public abstract class BoundServiceClient {
```
AdamBJohnsonx, line 52: Pretty sure that this should implement an interface.

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceFuture.java
```diff
@@ -23,18 +23,24 @@
...
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-public class MicrosoftAuthServiceFuture implements Future<IMicrosoftAuthService> {
+/**
+ * A Future object for bound service.
+ * Will block until a bound service connection is established.
+ */
+public class BoundServiceFuture implements Future<IInterface> {
```
AdamBJohnsonx, line 40: Can we just use a SettableFuture here?

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
rpdome: This is different from the strategy. Given the nature of bound service (we're not passing a 'flag' inside the bundle, but we'll be picking a method to call), we can't truly make the ipc strategy generic.

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceFuture.java
rpdome: by importing com.google.guava:guava:28.1-android, yes.

rpdome: (although I don't think it's worth it for one-time use - since this will also inflate the customer's apk size).

AdamBJohnsonx: I was more thinking androidx.

rpdome: replaced BoundServiceFuture with common's ResultFuture<T>

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -0,0 +1,78 @@
...
+        return this.operation.name();
+    }
+
+//    public String getAccountManagerOperationKey(){
+//        // TODO
+//    }
+//
+//    public String getContentProviderUriPath(){
+//        // TODO
+//    }
```
AdamBJohnsonx, line 77: Should we have these comments here?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,187 @@
...
+                lastCaughtException = exception;
+            }
+        }
+
+        // This means that we've tried every strategies...
+        if (result == null) {
+            final ClientException exception = new ClientException(
+                    ErrorStrings.BROKER_BIND_SERVICE_FAILED,
+                    "Unable to connect to the broker",
+                    lastCaughtException);
```
AdamBJohnsonx, line 164: So here, we'd dump all of the exception information.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IIpcStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.microsoft.identity.common.exception.BaseException;
+
+/**
+ * an interface for inter-process communication strategies.
+ */
+public interface IIpcStrategy {
```
AdamBJohnsonx, line 13: Can we give them a getName()?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,187 @@
...
+                    new ApiStartEvent()
+                            .putProperties(parameters)
+                            .putApiId(operation.getTelemetryApiId())
+            );
+        }
+
+        U result = null;
+        Exception lastCaughtException = null;
+        for (int ii = 0; ii < mStrategies.size(); ii++) {
+            final IIpcStrategy strategy = mStrategies.get(ii);
```
AdamBJohnsonx, line 123: Here we can time the strategy execution from outside, by using the getName() method which, for bonus points, is by contract the telemetry name.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.List;
+
+/**
+ * Classes for executing Broker operations.
+ * It takes in a list of IIpcStrategy, and will try to connect to the broker with each strategy. one by one.
+ * - If the current strategy succeeds, it will return the result right away.
+ * - If the current strategy fails due to an internal Broker error, it will return the result right away.
+ * - If the current strategy fails to connect to the broker, it will try the next one until the list is exhausted.
+ */
+public class BrokerOperationExecutor {
```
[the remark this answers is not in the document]
AdamBJohnsonx: OK, so I know what's been bothering me about this. Typically these would be Handlers (BoundServiceHandler, etc.) and then this would be a BrokerOperationHandlerChain where we'd try the handlers in sequence until one worked.  Because we're not selecting a strategy, we're trying a bunch of them until one works.  It's not a big difference, really, so it's not a hill worth dying on.

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
AdamBJohnsonx: I don't think it's likely important.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
rpdome: These will be added in the next PRs.... so ... more like a reminder for myself (That the work in common is not done until these are implemented).

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
rpdome: I'll make this as part of the ests telemetry work.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IIpcStrategy.java
rpdome: I'll make this as part of the ests telemetry work.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,187 @@
...
+                                    .putException(exception)
+                                    .putApiId(operation.getTelemetryApiId())
+                    );
+                }
+                throw exception;
+            } catch (final Exception exception) {
+                // We know for a fact that in some OEM, bind service might throw a runtime exception.
+                // Given that the type of exception here could be unpredictable,
+                // it is better to catch 'every' exception so that we could try the next strategy - which could work.
+                lastCaughtException = exception;
```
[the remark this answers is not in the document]
rpdome: I'll put them in a list. if we fail all strategies, then i'l log all BrokerCommunicationException and return a ClientException without any internal exception object. I'll specify that if you want to see why it fails, please refer to MSAL logs.

rpdome:
- Address comments.
- Make BoundServiceClient and BrokerOperationBundle more generic. (so that we can reuse this for the broker key migration).
     - InactiveBrokerClient -> BrokerAccountServiceClient.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,203 @@
...
+                emitOperationFailureEvent((BrokerOperation<U>) operation, exception);
+                throw exception;
+            }
+        }
+
+        // This means that we've tried every strategies... log everything...
+        for (final BrokerCommunicationException e : communicationExceptionStack) {
+            Logger.error(TAG + methodName, e.getMessage(), e);
+        }
+        final ClientException exception = new ClientException(
```
AdamBJohnsonx, line 144: We have API 19+, right?  We can add all the exceptions that you caught in this processed as suppressedExceptions to the exception that you throw by calling addSuppressed(e) on it.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -0,0 +1,79 @@
...
+    @Getter
+    @NonNull final private Operation operation;
+
+    @Getter
+    @NonNull final private String targetBrokerAppPackageName;
+
+    @Getter
+    @Nullable final private Bundle bundle;
+
+    public BrokerOperationBundle(@NonNull final Operation operation,
```
AdamBJohnsonx, line 64: nit: @AllArgsConstructor?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
rpdome: min is 16. I could add this 

if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
                exception.addSuppressed(e);
            }

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -0,0 +1,203 @@
...
+         */
+        void putValueInSuccessEvent(ApiEndEvent event, T result);
+    }
+
+    private final List<IIpcStrategy> mStrategies;
+
+    /**
+     * @param strategies list of IIpcStrategy to be invoked.
+     */
+    public BrokerOperationExecutor(@NonNull final List<IIpcStrategy> strategies) {
```
AdamBJohnsonx, line 103: Thought experiment: should this be specified in a configuration?

AdamBJohnsonx (approved):
This is on the right path.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
rpdome: IMO, this kind of stuff is too low-level and we should decide for them - with telemetry.
(e.g. once bind service is 'on the way out' in MSAL, there's no reason to let the devs pick it. we should just remove it).

rpdome:
force-pushed to clean commit history.

PR #1090, opened 2020-10-23
[IPC part 2] Refactor content provider strategy

rpdome:
part 1: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1088

- Add ContentProviderStrategy (mostly extracted from BrokerContentProviderStrategy, which is now gone).
- Make some changes on how the  interactiveRequest response is handled.
   - In Bound Service, Broker would return an Intent, containing a bundle inside it.
   - In ContentProvider, Broker would return the bundle only.
   - To standardize the logic (without making changes to the protocol itself), I'll be extracting the bundle from the intent in the bound service case, after that, we'll treat all strategy the same - constructing an intent from the bundle.


shahzaibj:
Perhaps, rename PR to `Refactor Content Provider strategy` (because it was already added)

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+            final String message = "Failed to get result from Broker Content Provider, cursor is null";
+            Logger.error(TAG + methodName, message, null);
+            throw new BrokerCommunicationException(message, null);
+        }
+    }
+
+    /**
+     * Returns a content provider URI for the given path.
+     */
+    private static Uri getContentProviderURI(@NonNull final String activeBrokerPackageName,
```
AdamBJohnsonx, line 102: Do these need to be static?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+                                             @NonNull final String path) {
+        final String authority = getContentProviderAuthority(activeBrokerPackageName);
+        return Uri.parse(CONTENT_SCHEME + authority + path);
+    }
+
+    /**
+     * Returns content provider authority.
+     */
+    private static String getContentProviderAuthority(@NonNull final String activeBrokerPackageName) {
```
AdamBJohnsonx, line 111: Is it important that these are static?

shahzaibj:
Can you add a link to the part 1 PR in the description of this PR? (so one can go back and relate)

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -496,29 +495,12 @@
...
-        final String className = resultBundle.getString(CLASS_NAME);
-        final Bundle bundle = resultBundle.getBundle(BUNDLE);
+        final String packageName = resultBundle.getString(BROKER_PACKAGE_NAME);
+        final String className = resultBundle.getString(BROKER_ACTIVITY_NAME);
 
         if (StringUtil.isEmpty(packageName) ||
-                StringUtil.isEmpty(className) ||
-                bundle == null) {
+                StringUtil.isEmpty(className)) {
             throw new ClientException("Unexpected error. Content of Authorization Intent bundle should not be null.");
```
AdamBJohnsonx, line 504: Is this message still accurate?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,132 @@
...
+        return activeBrokerPackageName + "." + AUTHORITY;
+    }
+
+    /**
+     * Returns true of the target package name supports this content provider strategy.
+     */
+    public static boolean isBrokerContentProviderAvailable(@NonNull final Context applicationContext,
```
AdamBJohnsonx, line 118: It would be convenient to make these non-static, and just configure with a list of them.

AdamBJohnsonx (approved):
I think there's an inaccurate message.  If we want to use static references, that's OK, but it would be more flexible and easier to test if we worked with a configured list of objects.  Also, we'd probably like to control them for testing/performance purposes.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BoundServiceStrategy.java
```diff
@@ -16,7 +16,7 @@
 import java.util.concurrent.TimeoutException;
 
 /**
- * A strategy for communicating with the active broker (:auth process) via Bound Service.
+ * A strategy for communicating with the targeted broker via Bound Service.
```
shahzaibj, line 19: How is the broker going to be specified?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -63,11 +69,42 @@
...
+                return BrokerContentProvider.REMOVE_ACCOUNTS_PATH;
+
+            case MSAL_GET_DEVICE_MODE:
+                return BrokerContentProvider.GET_DEVICE_MODE_PATH;
+
+            case MSAL_GET_CURRENT_ACCOUNT_IN_SHARED_DEVICE:
+                return BrokerContentProvider.GET_CURRENT_ACCOUNT_SHARED_DEVICE_PATH;
+
+            case MSAL_SIGN_OUT_FROM_SHARED_DEVICE:
+                return BrokerContentProvider.SIGN_OUT_FROM_SHARED_DEVICE_PATH;
```
shahzaibj, line 102: Could these paths be derived out of the operation enum? (so we don't need to have this switch)

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,131 @@
...
+            final String message = "Failed to get result from Broker Content Provider, cursor is null";
+            Logger.error(TAG + methodName, message, null);
+            throw new BrokerCommunicationException(message, null);
+        }
+    }
+
+    /**
+     * Returns a content provider URI for the given path.
+     */
+    private Uri getContentProviderURI(@NonNull final String activeBrokerPackageName,
```
shahzaibj, line 102:
```suggestion
                uri,
                null,
                marshalledRequestString,
                null,
                null
        );

        if (cursor != null) {
            final Bundle resultBundle = cursor.getExtras();
            cursor.close();
            Logger.info(TAG + methodName, "Received successful result from Broker Content Provider.");
            return resultBundle;
        } else {
            final String message = "Failed to get result from Broker Content Provider, cursor is null";
            Logger.error(TAG + methodName, message, null);
            throw new BrokerCommunicationException(message, null);
        }
    }

    /**
     * Returns a content provider URI for the given path.
     */
    private Uri getContentProviderURI(@NonNull final String targetedBrokerPackageName,
```

shahzaibj, line 102: targetedBrokerPackageName

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,131 @@
...
+                                      @NonNull final String path) {
+        final String authority = getContentProviderAuthority(activeBrokerPackageName);
+        return Uri.parse(CONTENT_SCHEME + authority + path);
+    }
+
+    /**
+     * Returns content provider authority.
+     */
+    private String getContentProviderAuthority(@NonNull final String activeBrokerPackageName) {
```
shahzaibj, line 111:
```suggestion
    private String getContentProviderAuthority(@NonNull final String targetedBrokerPackageName) {
```

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BoundServiceStrategy.java
rpdome: BrokerOperationBundle.targetBrokerAppPackageName.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -0,0 +1,131 @@
...
+        return activeBrokerPackageName + "." + AUTHORITY;
+    }
+
+    /**
+     * Returns true of the target package name supports this content provider strategy.
```
shahzaibj, line 116:
```suggestion
     * Returns true if the target package name supports this content provider strategy.
```

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
rpdome: not really. the operation could be converted to a content provider path, or account manager key, or something else, depending on the communication mechanism. so it's not a 1:1 relationship.

shahzaibj (approved):
Left some questions / comments, but LGTM

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
shahzaibj: Yeah what I meant is that should we store each of those mappings inside the enum? (rather than using constants from a different file)

shahzaibj: So the operation contains mapping for each communication mechanism

rpdome: I'd say no, but I don't feel strong on this.

The reason is that the same constants are also used on the server (broker) side, and that side should not be aware of these operation object (or it should?)

PR #1091, opened 2020-10-23
Add abstract class for HttpClient

AdamBJohnsonx:
Implementations of HttpClient need to implament a lot of functionality.
We really only care about one thing, so handle that.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
@@ -1,0 +1,59 @@
+package com.microsoft.identity.common.internal.net;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Map;
+
+public abstract class AbstractHttpClient implements HttpClient {
+
+    @Override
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.util.Map;
+
+public abstract class AbstractHttpClient implements HttpClient {
+
+    @Override
+    public HttpResponse method(@NonNull String httpMethod, @NonNull URL requestUrl, @NonNull Map<String, String> requestHeaders, @Nullable byte[] requestContent) throws IOException {
+        return method(HttpClient.HttpMethod.validateAndNormalizeMethod(httpMethod), requestUrl, requestHeaders, requestContent);
+    }
```
shahzaibj, line 15: Is this method being used anywhere (after your changes)?

AdamBJohnsonx: Probably.  I just moved it from the non-abstract class to the abstract class.  HttpRequest.sendWithMethod delegates here.

shahzaibj (approved):

tanmaymanolkar1 (approved):

PR #1092, opened 2020-10-25
[IPC part 3] Refactor AccountManager Strategy.

rpdome:
part 1: #1088 
part 2: #1090 

- Add AccountManagerAddAccountStrategy (mostly extracted from BrokerAccountManagerStrategy, which is now gone).


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -495,13 +498,15 @@
...
 
+    public @NonNull Intent getIntentForInteractiveRequestFromResultBundle( @NonNull final Bundle resultBundle) throws ClientException {
+        final Bundle interactiveRequestBundle = extractInteractiveRequestBundleFromResultBundle(resultBundle);
+
+        final String packageName = interactiveRequestBundle.getString(BROKER_PACKAGE_NAME);
+        final String className = interactiveRequestBundle.getString(BROKER_ACTIVITY_NAME);
         if (StringUtil.isEmpty(packageName) ||
                 StringUtil.isEmpty(className)) {
-            throw new ClientException("Unexpected error. Content of Authorization Intent's package and class name should not be null.");
+            throw new ClientException(ClientException.UNKNOWN_ERROR, "Content of Authorization Intent's package and class name should not be null.");
```
AdamBJohnsonx, line 509: Should we define ClientException.INVALID_BROKER_INTENT?

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -57,8 +60,7 @@ public static boolean isBrokerProcess(@NonNull final Context context) {
     /**
      * Returns the running process name.
      */
-    @Nullable
-    private static String getProcessName(@NonNull final Context context) {
+    private static @Nullable String getProcessName(@NonNull final Context context) {
```
AdamBJohnsonx, line 63: Under what circumstances will this return null?

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -72,4 +74,15 @@ private static String getProcessName(@NonNull final Context context) {
 
         return null;
     }
+
+    /**
+     * Returns a handler based on the current looper.
+     */
+    public static @NonNull Handler getPreferredHandler() {
+        if (null != Looper.myLooper() && Looper.getMainLooper() != Looper.myLooper()) {
```
AdamBJohnsonx, line 82: I'm not sure what the difference would be if we just said if (null != Looper.myLooper()) ?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/AccountManagerAddAccountStrategy.java
```diff
@@ -0,0 +1,103 @@
...
+        Bundle requestBundle = brokerOperationBundle.getBundle();
+        if (requestBundle == null) {
+            requestBundle = new Bundle();
+        }
+
+        requestBundle.putString(
+                AuthenticationConstants.Broker.BROKER_ACCOUNT_MANAGER_OPERATION_KEY,
+                brokerOperationBundle.getAccountManagerAddAccountOperationKey());
+
+        return requestBundle;
```
AdamBJohnsonx, line 101: Should we put this into a method on BrokerOperationBundle instead of putting it here?  Maybe putRequestString(key, value) that could contain all of this logic (check if the request bundle is created, create if necessary, add appropriate string under the correct specified key)?

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -510,10 +515,22 @@
...
         return intent;
     }
 
+    /**
+     * AccountManager strategy is currently returning result in a different format compared with the other strategies.
+     * We should make sure this does NOT happen going forward.
+     */
+    public Bundle extractInteractiveRequestBundleFromResultBundle(@NonNull final Bundle resultBundle) {
+        final Intent interactiveRequestIntent = resultBundle.getParcelable(AccountManager.KEY_INTENT);
+        if (interactiveRequestIntent != null) {
```
AdamBJohnsonx, line 528: Is there a way to be explicit about when we want to have this behavior (ONLY when talking over accountManager) rather than implicit, so that it doesn't become an unintended feature?

AdamBJohnsonx (approved):
Mostly minor, but can we prevent any other return method from mis-constructing the return bundle and having it work?

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: when we can't find a matching process id, which should never happen.

I'm hesitant to throw a RuntimeException here (IllegalStateException?). Maybe that's the right thing to do.

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
rpdome: I actually moved this back from AccountManager Strategy, and the reason is that the ipc strategy itself should not have to care about the content of the payload. It's a bug on our side that sends a different payload for each communication channel.

this layer won't/shouldn't be aware of the ipc strategy either (unless we modify the BrokerOperation interface, which I'm not keen of) - so... the answer is no, I can't think of any clean way to handle this.

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
rpdome: Keeping it the same for now.

PR #1094, opened 2020-10-26
Changes for UI Automation Pipeline

shahzaibj:
This PR achieves the following:

- Add rule to copy installed APK to /data/local/tmp folder
-- This is needed because Firebase will install the APK on the device even before a test begins....and our test infrastructure will go and delete it right away....because we do clean up before a test begins i.e. we uninstall Broker apps before each test and re-install if the test requires it. This then poses a problem with Firebase because we need the ability to uninstall and re-install these APKs at specific point in execution of the test. To achieve this, we first copy the installed APK to /data/local/tmp folder....and then we proceed with our regular flow...which is to uninstall and re-install...but now since we have copied to a different location, we can actually proceed with uninstall/re-install.
- Add support to accept Google Play terms of service when installing APK from Google Play
- Add a RulesHelper to create a RuleChain and apply that directly when possible (instead of using the rule order notation)...this is much cleaner
- Add some logs to rules execution
- Enable logcat logs for MSAL

We will be adding more logs to automation code over time

Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1231


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/AbstractTestBroker.java
```diff
@@ -53,6 +55,25 @@
...
+            // The broker app will still be installed on the device if it is enabled
+            // as a device admin. In this case, we need to disable the admin and then
+            // uninstall.
+            Log.w(TAG, "Unable to uninstall broker " + getAppName() + " from device..." +
+                    "the broker is potentially enabled as an active device admin.");
+            Log.i(TAG, "Disabling admin for " + getAppName());
+            TestContext.getTestContext().getTestDevice().getSettings().disableAdmin(getAdminName());
+            Log.i(TAG, "Reattempting uninstall for " + getAppName());
+            super.uninstall();
+        }
```
shahzaibj, line 74: This is just refactor....just relocating this code here from another class where it was previously located.

iambmelt (approved):

PR #1084, opened 2020-10-12
Android/update versions gradle

shoatman:

PR #1082, opened 2020-10-09
Update gradle plugin version, buildToolsVersion, targetSdkVersion, compileSdkVersion, explicitly set useAndroidX=true, add SpotBugs tasks

iambmelt:
Draft PR for updating Gradle, adding SpotBugs.

Work in progress.

### Available SpotBugs Tasks:
```
common:spotbugsDebug
common:spotbugsRelease
```

### Execute SpotBugs Tasks Locally:
```
> ./gradlew spotBugsDebug // Or whichever variant you require
```

Similar:
- #1002

PR #1095, opened 2020-10-26
Android 11 Fixes (All fixes combined now in one simple PR)

shahzaibj:
Address the following Android 11 issues:

- Unable to query package manager for and launch browsers that support custom tabs
- Unable to query package manager for and launch regular browsers that don’t support custom tabs
- Unable to query package manager for Broker apps installed on the device (We could still see the Authenticators on the device but when trying to query them via Package Manager it would not be found)

Read More about why these changes are needed: https://developer.android.com/training/basics/intents/package-visibility-use-cases

Note: The new changes introduced in the AndroidManifest as part of these fixes also depends on a more recent version of gradle (old versions of gradle do not recognize those new attributes in the manifest file) and hence my builds are currently broken on the pipeline. (They did work in android-complete as we’ve already upgraded the gradle version in that project.)


AdamBJohnsonx (commented):
If it works, that's fine.  We need to get our CI to work, however.  I know that there are several different attempts to do this out there in the wild:

https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1002
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1084
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1082

Any of these working would mean that your changes would work as well.

tanmaymanolkar1 (commented):
looks good as per android dev doc, CIs are a blocker

AdamBJohnsonx (approved):
Once CI passes then you can ship it.

PR #1101, opened 2020-10-28
[IPC part 5.5] make changes to support the Broker API work.

rpdome:


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -645,9 +645,14 @@
...
-         * The maximum broker protocol version that common supports.
+         * The newest Msal-To-Broker protocol version.
          */
-        public static final String BROKER_PROTOCOL_VERSION_CODE = "5.0";
+        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "5.0";
+
+        /**
+         * The newest BrokerAPI-To-Broker protocol version.
+         */
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = "1.0";
```
AdamBJohnsonx, line 655: Can we describe this via a schema rather than a number?

rpdome: We definitely can, but what would be the benefit over using a number?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -155,7 +155,7 @@ public BrokerOperationExecutor(@NonNull final List<IIpcStrategy> strategies) {
         throw exception;
     }
 
-    private <T extends CommandParameters, U> void emitOperationStartEvent(@NonNull final T parameters,
+    private <T extends CommandParameters, U> void emitOperationStartEvent(@Nullable final T parameters,
```
AdamBJohnsonx, line 158: Do we want to allow parameters to be null here?

AdamBJohnsonx (approved):
Comments are minor, you can take them or not, though I *highly* recommend we turn this from a version number scheme into a message-format scheme.

iambmelt (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
rpdome: The bundle itself is technically a kind of schema. I think we can add metadata to the bundle directly.

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -159,6 +159,22 @@ public static int compareSemanticVersion(@NonNull final String thisVersion,
         return 0;
     }
 
+    /**
+     * Returns true if the first semantic version is smaller or equal to the second version.
```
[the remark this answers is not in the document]
rpdome: This is thought provoking.

IMO, it makes sense to me for MSAL hello actually 'negotiate' the protocol version, and the reason is that, unlike services in general, it's possible that user may never update the Broker app, and ended up having a newer protocol version on the other side.

So it's not just "could you understand a message I send", but more like, "let's find a middle ground that both of us can communicate."

Having said that, we have never really leveraged the negotiated version outside the hello() function, and we should. 

I was originally planning to start documenting each version of IPC protocols, and this comment made me realized that we could go one step forward.

We could actually use the negotiated protocol version to "pick the protocol" that both side uses to communicate.

Instead of having just one copy of the protocol/adapter on each sides - we can have one copy for EACH version of the protocol. Once you get the negotiated version, you plug that into a factory class ON BOTH SIDES, and it would spew out the operation code itself (i.e. how broker would do addAccount, removeAccount ,etc.) and how the bundle should be packed and unpacked (adapter).

This way, the code will document itself, and there's less chance that we'll mess up the protocol logic. (i.e. with one adapter like what we have now, we might accidentally remove a field in the latest protocol version.)

(I actually spent the whole afternoon trying this out.. and I'll send out a (draft) PR soon, but this will have a lower priority than another BRT injection PR I'll send out. I think this can wait. It shouldn't be hard to 'fix forward' this in both BrokerAPI and MSAL).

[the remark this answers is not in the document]
AdamBJohnsonx: Yes.  That's precisely it, I think.  Think HTTP.  I send you a POST or a PUT with a MIME type attached.  You use that MIME type to understand the message that I sent OR you reject it with a list of content types that you do understand.   For each of these content types, I typically have a content type handler, that constructs an object that I can act upon from the message body.  This instruction appears in the envelope surrounding the message body.  Similar for encoding, both Content-Encoding and Transfer-Encoding.

Some of the content that is exchanged is a self-describing structure (XML is an example), in that it is contains information that the recipient could use to construct a way to unpackage the message.  Then the question becomes about the the message - is that something that I can understand - similar enough and asserted to be the same that I can act on it.

To respond, the caller told me what kind of message they expect in his initial correspondence.  I can select any one of the mechanisms that I have available to produce responses by consulting that list.

Now the problem comes in when you don't actually have a common way to pass messages.  If we can assert that the message that's being transmitted is substantially similar for a given path, then we can encode into the envelope away of discovering the ability to understand the structure.

I think that the HTTP designers also have a very similar mechanism (OPTIONS) to our HELLO, but I have always avoided this like the plague, since it's an additional API endpoint to handle as well as an obligate extra round-trip, always opting for the reject-and-specify model.

AdamBJohnsonx (approved):
Still ship it.

rpdome:
force push to rebase and clean the commit history.

PR #1108, opened 2020-11-01
Hello() cache

rpdome:
**Changes**
Introduces a cache for cutting down the amount of redundant hello().

It can be deduced that a negotiated protocol version ***might*** change when one (or more) the following criteria changes.
1. The client (app) bumps minimum version support in config file 
2. The client (MSAL) bumps maximum/curent protocol version.
3. When the targeted broker app changes 
   (i.e. Authenticator app is uninstalled and CP became the new broker)
4. When the targeted app is updated.
   - We do not have an access to the actual protocol version the broker supports - it might or might not be the same, so we make sure nothing breaks by re-negotiating.

This cache will use all the following criterias to create a key - for storing negotiated protocol version in SharedPref.

This should cut the amount of hello() down to lower single digit ***per month***.

I also integrate the cache with BrokerMsalController in this PR.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ vNext
     - [Part 4] test cases for IPC strategies (#1093)
     - [Part 5.5] make changes to support the Broker API work (#1101)
 - Make change to support Broker API's updateBrokerRT() functionality (#1107).
+- Introduce a cache for Hello() protocol (#1108)
 
 Version 3.0.4
 ----------
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -645,9 +645,19 @@ public final class AuthenticationConstants {
         public static final String BROKER_ACTIVITY_NAME = "broker.activity.name";
 
         /**
+         * The Msal-To-Broker protocol name.
+         */
+        public static final String MSAL_TO_BROKER_PROTOCOL_NAME = "msal.to.broker";
+
+        /**
          * The newest Msal-To-Broker protocol version.
          */
         public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "5.0";
+
+        /**
+         * The BrokerAPI-To-Broker protocol name.
+         */
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_NAME = "broker.api.to.broker";
 
         /**
          * The newest BrokerAPI-To-Broker protocol version.
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
@@ -1,0 +1,168 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.identity.common.internal.cache;
+
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.os.Build;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+
+import com.microsoft.identity.common.internal.logging.Logger;
+
+/**
+ * Persisted cache for the IPC hello() protocol.
+ * Use client's protocol version and the targeted app's package name and app version as a key
+ * to cache the negotiated protocol version.
+ * <p>
+ * This means a new hello() call will ONLY be triggered only when.
+ * 1. IPC operation is invoked for the very first time.
+ * 2. Client bumps up protocol version.
+ * 3. The targeted app is updated, uninstalled, reinstalled.
+ */
+public class HelloCache {
+    private static final String TAG = HelloCache.class.getSimpleName();
+
+    private static final String SHARED_PREFERENCE_NAME = "com.microsoft.common.ipc.hello.cache";
+
+    private final SharedPreferencesFileManager mFileManager;
+    private final Context mContext;
+    private final String mProtocolName;
+    private final String mTargetAppPackageName;
+    private static boolean sIsEnabled = true;
+
+    /**
+     * If set to false, Hello cache will be disabled.
+     * When you're developing protocol change, you might not want the cache to be enabled.
+     * <p>
+     * For debugging only.
+     */
+    public static void setIsEnabled(final boolean value) {
+        synchronized (HelloCache.class) {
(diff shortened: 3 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -644,11 +644,21 @@
...
+
         /**
          * The newest Msal-To-Broker protocol version.
          */
         public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "5.0";
 
+        /**
+         * The Msal-To-Broker protocol name.
+         */
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_NAME = "broker.api.to.broker";
```
AdamBJohnsonx, line 660: these will change, so you may want a distinguishing factor in here.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java
```diff
@@ -136,6 +136,10 @@ public boolean isBrokerContentProviderAvailable(@NonNull final String targetedBr
         final List<ProviderInfo> providers = mContext.getPackageManager()
                 .queryContentProviders(null, 0, 0);
 
+        if (providers == null){
+            return false;
```
AdamBJohnsonx, line 140: Log.

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -0,0 +1,152 @@
...
+                + mTargetAppPackageName + "[" + getVersionCode() + "]";
+    }
+
+    /**
+     * To be used while debugging protocol change only.
+     */
+    @VisibleForTesting
+    public void clearCache() {
+        mFileManager.clear();
+    }
```
AdamBJohnsonx, line 141: We might want a way to clean these up so they don't continually build.

AdamBJohnsonx (approved):
I don't have a problem here.  I do think we should have a more thorough negotiation scheme, but that will mainly help us once we drop hello.

changelog.txt
```diff
@@ -14,6 +14,7 @@ vNext
     - [Part 3] Refactor AccountManager Strategy[IPC part 4] test cases for IPC strategies (#1092)
     - [Part 4] test cases for IPC strategies (#1093)
     - [Part 5.5] make changes to support the Broker API work (#1101)
+- Hello() cache (#1108)
```
shahzaibj, line 17: We probably want something more descriptive here (as this changelog is public)

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -644,11 +644,21 @@
...
+         */
+        public static final String MSAL_TO_BROKER_PROTOCOL_NAME = "msal.to.broker";
+
         /**
          * The newest Msal-To-Broker protocol version.
          */
         public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "5.0";
 
+        /**
+         * The Msal-To-Broker protocol name.
```
shahzaibj, line 658:
```suggestion
         * The Broker-To-Broker protocol name.
```

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -0,0 +1,152 @@
...
+ * to cache the negotiated protocol version.
+ * <p>
+ * This means a new hello() call will ONLY be triggered only when.
+ * 1. IPC operation is invoked for the very first time.
+ * 2. Client bumps up protocol version.
+ * 3. The targeted app is updated, uninstalled, reinstalled.
+ * <p>
+ * NOTE: the downside of this is that if you're debugging protocol change on the targeted app side,
+ * you'll either have to keep bumping the test app version or clear the cache.
+ * I've provided clearCache() for this purpose.
```
shahzaibj, line 49: Or you could pass a flag to disable/enable the cache for Hello....I would actually like to see that...doesn't necessarily have to be in this PR.

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -0,0 +1,152 @@
...
+        mFileManager.putString(key, negotiatedProtocolVersion);
+    }
+
+    /**
+     * Generates {@link SharedPreferencesFileManager}'s key.
+     *
+     * @param clientMinimumProtocolVersion minimum version of the protocol that the client supports.
+     * @param clientMaximumProtocolVersion maximum version of the protocol that to be advertised by the client.
+     */
+    private @NonNull String getKey(@Nullable final String clientMinimumProtocolVersion,
```
shahzaibj, line 127:
```suggestion
    private @NonNull String getNegotiatedProtocolVersionCacheKey(@Nullable final String clientMinimumProtocolVersion,
```

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -0,0 +1,152 @@
...
+        } catch (final PackageManager.NameNotFoundException e) {
+            Logger.error(TAG + methodName, "Failed to retrieve key", e);
+            return;
+        }
+
+        mFileManager.putString(key, negotiatedProtocolVersion);
+    }
+
+    /**
+     * Generates {@link SharedPreferencesFileManager}'s key.
```
shahzaibj, line 122:
```suggestion
     * Generates {@link SharedPreferencesFileManager}'s cache key for the negotiated protocol version.
```

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param clientMinimumProtocolVersion minimum version of the protocol that the client supports.
+     * @param clientMaximumProtocolVersion maximum version of the protocol that to be advertised by the client.
+     */
+    private @NonNull String getKey(@Nullable final String clientMinimumProtocolVersion,
+                                   @NonNull final String clientMaximumProtocolVersion)
+            throws PackageManager.NameNotFoundException {
+        return mProtocolName +
+                "[" + clientMinimumProtocolVersion + "," + clientMaximumProtocolVersion + "]:"
+                + mTargetAppPackageName + "[" + getVersionCode() + "]";
```
shahzaibj, line 132: this cache is actually in the calling app itself right? If yes, then do we even need to store the package name?

rpdome: package name of the broker.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -153,16 +170,41 @@ public BrokerMsalController(final Context applicationContext) {
...
+    @VisibleForTesting
+    public @NonNull String hello(@NonNull IIpcStrategy strategy,
+                                 @NonNull final CommandParameters parameters) throws BaseException {
+
+        final Bundle bundle = mRequestAdapter.getRequestBundleForHello(parameters);
+        final String minimumProtocolVersion = bundle.getString(CLIENT_CONFIGURED_MINIMUM_BP_VERSION_KEY);
+        final String maximumProtocolVersion = bundle.getString(CLIENT_ADVERTISED_MAXIMUM_BP_VERSION_KEY);
+
+        // This should be part of the bundle.
+        // If we're hitting this, it means that the hello protocol changed. see getRequestBundleForHello().
```
shahzaibj, line 182: and do we actually want a RuntimeException in that case?

shahzaibj: can we just catch, log, and return anything that is `BaseException`?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -103,15 +108,27 @@
...
         if (TextUtils.isEmpty(mActiveBrokerPackageName)) {
             throw new IllegalStateException("Active Broker not found. This class should not be initialized.");
         }
 
         mBrokerOperationExecutor = new BrokerOperationExecutor(getIpcStrategies(mApplicationContext, mActiveBrokerPackageName));
+        mHelloCache = getHelloCache();
+    }
+
+    @VisibleForTesting
+    public HelloCache getHelloCache() {
```
shahzaibj, line 125: we probably don't need this method....whatever it is doing I think can be done directly in the test

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
shahzaibj: Got it

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
rpdome: Might as well do it otherwise it'll sit in my backlog for eternity :D

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
rpdome: This is for testing if the cache is properly wired.. we need this to inject the mock Token cache into BrokerMsalController.

shahzaibj (approved):

PR #1109, opened 2020-11-02
Android/package test utilities

shoatman:


testutils/build.gradle
```diff
@@ -56,11 +64,86 @@ android {
...
+}
+
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation project(':common')
+
     implementation project(':keyvault')
     implementation project(':labapi')
+
+    implementation(group: 'com.microsoft.identity', name: 'common', version: '0.0.22')
```
shahzaibj, line 123: What version of common is this? Looks very old

common/versioning/version.properties
```diff
@@ -1,3 +1,4 @@
-#Thu Aug 02 12:03:16 PDT 2018
-versionName=3.0.5-RC1
-versionCode=1
+#Mon Nov 02 10:02:30 PST 2020
+versionName=0.0.26
+versionCode=1
+latestPatchVersion=26
```
shahzaibj, line 4: I'm not clear on what is happening here?

shahzaibj:
Can you add some description to the PR about all the things that are happening here?

testutils/build.gradle
shoatman: Good catch.  I'm going to point this to 0.0.+.  Using 0.0.<patch> to ensure that no one tries to ship one of these.

common/versioning/version.properties
shoatman: This is the version file that we've been using to actually version things.  The versionName shouldn't be set to this, but I'll change before merging.  I'll write up some documentation to go along with this PR.

shahzaibj: I know the file....I don't understand what these `^M` symbols are meant to do

shahzaibj: and why the version is down to from `3.0.5` to `0.0.26`

testutils/build.gradle
shahzaibj: but why are we not using the latest? (`3.0.4`)

.github/workflows/main.yml
```diff
@@ -3,18 +3,41 @@ name: dev_updated
...
+
 on:
-  push:
-    branches: [ dev ]
+  watch:
+    types: [started]
+    branches: [android/package-test-utilities]
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+# Jobs run in parrelel by default.  Use "needs:" to indicate dependencies run sequentially
```
AdamBJohnsonx, line 15:
```suggestion
# Jobs run in parallel by default.  Use "needs:" to indicate dependencies run sequentially
```

testutils/build.gradle
shoatman: I thought 0.0 made it clear that this isn't intended to be used... in any production scenario.

common/versioning/version.properties
```diff
@@ -1,3 +1,4 @@
-#Thu Aug 02 12:03:16 PDT 2018
-versionName=3.0.5-RC1
-versionCode=1
+#Mon Nov 02 10:02:30 PST 2020
+versionName=3.0.5-RC1
+versionCode=1
+latestPatchVersion=26
```
AdamBJohnsonx, line 4: Ah, windows newlines versus unix newlines.

shoatman: Yes.

shahzaibj: ohhh this is a newline? I was confused about what this was for too long

AdamBJohnsonx: > ohhh this is a newline? I was confused about what this was for too long

I saw that, so I commented to clarify. ;-)

common/versioning/version_tasks.gradle
```diff
@@ -34,6 +45,18 @@ private void saveChanges(String versionName) {
     versionProps.store(getVersionFile().newWriter(), null)
 }
 
+private void incrementLatestVersion(){
+    def versionProps = getVersionProps()
+    versionProps["latestPatchVersion"] = (getLatestPatchVersion() + 1).toString()
```
AdamBJohnsonx, line 50: Does groovy coerce this to 0?

keyvault/build.gradle
```diff
@@ -1,4 +1,42 @@
-apply plugin: 'java-library'
+plugins {
+    id 'java-library'
+    id 'maven-publish'
+}
+
+apply from: '../common/versioning/version_tasks.gradle'
+
+project.ext.vstsUsername = System.getenv("ENV_VSTS_MVN_ANDROIDCOMMON_USERNAME") != null ? System.getenv("ENV_VSTS_MVN_ANDROIDCOMMON_USERNAME") : project.findProperty("vstsUsername")
+project.ext.vstsPassword = System.getenv("ENV_VSTS_MVN_ANDROIDCOMMON_ACCESSTOKEN") != null ? System.getenv("ENV_VSTS_MVN_ANDROIDCOMMON_ACCESSTOKEN") : project.findProperty("vstsMavenAccessToken")
```
AdamBJohnsonx, line 9: explode if you can't find it, pls?

shoatman: This bit of code is in multiple projects.  If we're going to do this i'd want to do it at once for all projects... probably when i upgrade to latest gradle plugin.  Are you ok with leaving this as is for now?

common/versioning/version_tasks.gradle
shoatman: I assume you're referring to type coercion, but not sure i'm following in this case.  getLatestPatchVersion returns an integer or it throws an exception and would fail the script.

common/versioning/version_tasks.gradle
```diff
@@ -1,5 +1,12 @@
 def getVersionFile() {
-    return file('versioning/version.properties')
+    
+    File file = new File("$rootDir/common/common/versioning/version.properties")
+    if(file.exists()){
+        return file;
+    }else{
+        return new File("$rootDir/common/versioning/version.properties")
+    }
```
AdamBJohnsonx, line 8: This seems to do the same thing in both cases.  Am I missing something?

shoatman: I think so... the path is different.

AdamBJohnsonx: Yup.  So one will cover common as part of android-complete, and the other as part of common.  I wonder that this shouldn't be parameterized to work with all the other packages.

common/versioning/version_tasks.gradle
AdamBJohnsonx: > I assume you're referring to type coercion, but not sure i'm following in this case. getLatestPatchVersion returns an integer or it throws an exception and would fail the script.

I'm just wondering about the case where that file is empty or doesn't exist.  We implicitly accept that earlier.  I'd rather fail there.

keyvault/build.gradle
AdamBJohnsonx: > This bit of code is in multiple projects. If we're going to do this i'd want to do it at once for all projects... probably when i upgrade to latest gradle plugin. Are you ok with leaving this as is for now?

Yes.

common/versioning/version_tasks.gradle
shoatman: I was consider writing a document to cover gradle project composability.  To be honest I think the design of this can be improved.  Again since the same code is in multiple projects i'll take this a feedback for my next update of everything referenced by Android-complete.  Would be interested to see an example of what you're thinking about in terms of prioritization.

shoatman:
I added these 2 issues to android-complete: 
https://github.com/AzureAD/android-complete/issues/55
https://github.com/AzureAD/android-complete/issues/54

common/versioning/version_tasks.gradle
shoatman: I didn't write that code, and i'd like to avoid refactoring something that doesn't have any record of being broken.

common/versioning/version_tasks.gradle
AdamBJohnsonx: So if this becomes a reusable activity that you can just add to your flow, then $rootDir/$versionFile, I would thing, where versionFile is in the workflow instance definition.

AdamBJohnsonx (approved):
No major comments.

common/versioning/version_tasks.gradle
shoatman: Yep.  We can design something better here.  Sounds like a fun hackathon project.

PR #1110, opened 2020-11-03
tamanolk/silent broker request changes

tanmaymanolkar1:
Adding a null check for sdktype in BrokerSilentTokenCommandParameters

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
@@ -324,7 +324,7 @@ public class MsalBrokerRequestAdapter implements IBrokerRequestAdapter {
                 .authenticationScheme(getAuthenticationScheme(context, brokerRequest))
                 .androidApplicationContext(context)
                 .accountManagerAccount(account)
-                .sdkType(brokerRequest.getSdkType())
+                .sdkType(brokerRequest.getSdkType() == null ? SdkType.MSAL : brokerRequest.getSdkType())
                 .sdkVersion(brokerRequest.getMsalVersion())
                 .callerUid(callingAppUid)
                 .applicationName(brokerRequest.getApplicationName())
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):
cc/ @rpdome

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -324,7 +324,7 @@ public BrokerSilentTokenCommandParameters brokerSilentParametersFromBundle(
                 .authenticationScheme(getAuthenticationScheme(context, brokerRequest))
                 .androidApplicationContext(context)
                 .accountManagerAccount(account)
-                .sdkType(brokerRequest.getSdkType())
+                .sdkType(brokerRequest.getSdkType() == null ? SdkType.MSAL : brokerRequest.getSdkType())
```
AdamBJohnsonx, line 327: add a test?

tanmaymanolkar1: this was caught in the test pass

AdamBJohnsonx: Right - we should add a unit test for this behavior of MsalBrokerRequestAdapter so that we don't require a manual test in order to catch it next time.

shahzaibj: Agreed. @tanmaymanolkar1 I suggest you create a work item in your backlog to add a unit test for this. I don't think we necessarily have to do it right now as I don't want to block the release process.

AdamBJohnsonx: That works.  Typically my ideal flow would be - write test exposing break, fix code to close break, break is now proven fixed.

AdamBJohnsonx (approved):
Make that backlog item.

PR #1112, opened 2020-11-05
Revert #1087, Subclass AuthorizationActivity in :auth process

rpdome:
Fix a regression in JoinActivity.

**Background**
- CP will be shipping MSAL with Broker, and they ran into a weird issue.
- What happened was that ad-accounts declares in the manifest that AuthorizationActivity will be run **exclusively** in :auth process. This means the non-broker auth MSAL will never get the response back... since the authentication happens in :auth process, and it was waiting for the result on the main process.
- I tried to fix the issue by making AuthorizationActivity runs on multiple processes (https://github.com/AzureAD/ad-accounts-for-android/pull/1399), side effect of this is that if Authenticator/CP is running in the background, once the AuthorizationActivity is popped from the back stack, those activity of Authenticator/CP is next in line. 
- I mitigated this issue in #1087 (which is what Browser flow in Broker - exclusively to COBO - is doing), by clearing related activity that might be in the Task/stack when launching an activity (Intent.FLAG_ACTIVITY_CLEAR_TASK).
- This introduces a regression in JoinActivity - being in :auth process also, Intent.FLAG_ACTIVITY_CLEAR_TASK also kills JoinActivity.

**Changes**
1. Undo the changes in #1087. Move back the Intent.FLAG_ACTIVITY_CLEAR_TASK logic to Browser flow only.
2. Subclasses AuthorizationActivity (BrokerAuthorizationActivity), so that we can force this process to run on :auth without messing up with MSAL local flow.

related broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1421/files

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
@@ -46,7 +46,7 @@ import static com.microsoft.identity.common.adal.internal.AuthenticationConstant
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_CONTROLS_ENABLED;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_ENABLED;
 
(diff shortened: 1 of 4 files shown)


iambmelt (approved):
No concerns from me

AdamBJohnsonx (approved):
This seems appropriate, given the way android's manifest works.

PR #1113, opened 2020-11-06
Get sdk version from ADAL request

rpdome:
**Problem**
- We recently made a change in Broker to send the actual ADAL/MSAL version to eSTS via query parameter x-client-ver (it used to be hardcoded).
- We forgot to make the change in AdalBrokerRequestAdapter, so sdkVersion is never set.
- We ended up sending a request **without** it, and this triggers a weird behavior in MAM flow (see below - we should be seeing the WPJ register page)

Screen Shot 2020-11-06 at 4 15 02 PM


shahzaibj (approved):
LGTM

PR #1115, opened 2020-11-06
Fix sdk version not being set when ADAL used with Broker

shahzaibj:
Regressed as part of https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1034


tanmaymanolkar1 (approved):

tanmaymanolkar1 (commented):
LGTM! Thanks for doing this :)

PR #1122, opened 2020-11-12
Expose lower level API IAccountCredentialCache for use by MSALCPP

iambmelt:
Need method calls available to lower-level cache functions to support MSAL CPP lookup logic where `Credentials` can be stored without having an `Account`. Just calls `super`, as this parent class' function is `protected`.

Additional context:
- This PR supports the work done in https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/1711
- Description: OneAuth/MSALCPP wants to decouple the saving/read/delete relationship between accounts & credentials. As this connection is expressed in the `MsalCppOAuth2TokenCache`'s APIs, having access to the lower-level `IAccountCredentialCache` object is required in order to read/save/delete credentials in this manner

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 vNext
 ----------
-- Expose IAccountCredentialCache to MSALCPP for accessing lower-level cache functions.
+- Expose IAccountCredentialCache for accessing lower-level cache functions.
 - Adds unit test to verify .trim() behavior of cache keys.
 
 Version 3.0.6
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,7 @@
+vNext
+----------
+- Expose IAccountCredentialCache to MSALCPP for accessing lower-level cache functions.
+
 Version 3.0.6
 ----------
 - Expose expiresIn in MSAL Device Code flow callback (#1064)
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalCppOAuth2TokenCache.java
@@ -106,6 +106,10 @@ public class MsalCppOAuth2TokenCache
         return msalCppOAuth2TokenCache;
     }
 
+    public IAccountCredentialCache getAccountCredentialCache() {
+        return super.getAccountCredentialCache();
+    }
+
     /**
      * @param accountRecord : AccountRecord associated with the input credentials, can be null.
      * @param credentials   : list of Credential which can include AccessTokenRecord, IdTokenRecord and RefreshTokenRecord.


rpdome:
I'm still having a hard time wrapping my head around this change. Why are we now throwing away realm? Has readCredentials not working properly this whole time?

Om83:
@iambmelt Can you please add a more detailed description to the issue here?

shahzaibj (approved):

iambmelt:
> @iambmelt Can you please add a more detailed description to the issue here?

Updated - thanks!

iambmelt:
> I'm still having a hard time wrapping my head around this change. Why are we now throwing away realm? Has readCredentials not working properly this whole time?

@rpdome This is a request from @dabbcomputers to support the tests being written for OneAuth/MSALCPP -- the tests involve writing/reading/deleting credentials which may have no associated `AccountRecord`. 

As for why the `realm` is planned to be ignored in the context of RTs: since RTs can be used x-tenant, OneAuth plans to treat them in a realm-agnostic way

AdamBJohnsonx:
Can we give them their own package, so that this isn't a public method in our repo?  Should this be in testUtils instead of here?  I'm pretty sure I can edit or read these already, for tests, using the work we did in CacheUtils.  This doesn't seem to be about tests at all, but about trying to put a layer on top of these different storage mechanisms.

iambmelt:
> Can we give them their own package, so that this isn't a public method in our repo? Should this be in testUtils instead of here? I'm pretty sure I can edit or read these already, for tests, using the work we did in CacheUtils. This doesn't seem to be about tests at all, but about trying to put a layer on top of these different storage mechanisms.

@AdamBJohnsonx These change are about making the `StorageManager` work in the tests, not for the tests to call directly. See the changes in :
https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/1711

AdamBJohnsonx:
> > Can we give them their own package, so that this isn't a public method in our repo? Should this be in testUtils instead of here? I'm pretty sure I can edit or read these already, for tests, using the work we did in CacheUtils. This doesn't seem to be about tests at all, but about trying to put a layer on top of these different storage mechanisms.
> 
> @AdamBJohnsonx These change are about making the `StorageManager` work in the tests, not for the tests to call directly. See the changes in :
> [AzureAD/microsoft-authentication-library-for-cpp#1711](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/1711)

My only real issue here is that this is showing up as a public method, meaning that other users will find it and try to use it.  We could do something horrible and use package-visibility methods, and drop a class into this package in MSAL-CPP's java code to prevent that.  None of this will protect you from an attacker, mind, it just makes it difficult enough that casual users won't try.

changelog.txt
```diff
@@ -1,3 +1,7 @@
+vNext
+----------
+- Expose IAccountCredentialCache to MSALCPP for accessing lower-level cache functions.
```
AdamBJohnsonx, line 3:
```suggestion
- Expose IAccountCredentialCache for accessing lower-level cache functions.
```

AdamBJohnsonx (approved):

rpdome (approved):

PR #1126, opened 2020-11-13
Add escape sequence test and ensure .trim() lookup rules are applied to in-mem filtering logic

iambmelt:
Adds a test to demonstrate that escape sequences are trimmed from cache keys

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/CacheKeyValueDelegateTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/CacheKeyValueDelegateTest.java
@@ -66,12 +66,17 @@ public class CacheKeyValueDelegateTest {
     private static final String FIRST_NAME = "Jane";
     private static final String FAMILY_NAME = "Doe";
     private static final String AVATAR_URL = "https://fake.cdn.microsoft.com/avatars/1";
+    public static final String ESCAPE_SEQ_CHARS = "\r\f\n\t";
 
     private ICacheKeyValueDelegate mDelegate;
 
     @Before
     public void setUp() {
         mDelegate = new CacheKeyValueDelegate();
+    }
+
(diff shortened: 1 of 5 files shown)


common/src/test/java/com/microsoft/identity/common/CacheKeyValueDelegateTest.java
```diff
@@ -95,6 +95,26 @@ public void accessTokenCreateCacheKeyComplete() {
         assertEquals(expectedKey, mDelegate.generateCacheKey(accessToken));
     }
 
+    @Test
+    public void accessTokenCreateCacheKeyCompleteWithEscapeSequences() {
+        final AccessTokenRecord accessToken = new AccessTokenRecord();
+        accessToken.setHomeAccountId(HOME_ACCOUNT_ID);
+        accessToken.setEnvironment("\r\f\n\t" + ENVIRONMENT + "\r\f\n\t");
```
AdamBJohnsonx, line 102: Is this the only field that has this behavior?

iambmelt: They all should - can broaden this test to make this clearer.

AdamBJohnsonx (approved):
ship it.  To the point, here, have we put a breakpoint in the java code and then managed to hit it with the C++ test to see what we're actually getting sent?

iambmelt:
> ship it. To the point, here, have we put a breakpoint in the java code and then managed to hit it with the C++ test to see what we're actually getting sent?

Yep - it looks like the in-mem filtering is where this went awry, so shipped a simple fix in 699fa0ace59f9ef7ba4fc4cecc78c12b4729315d to make it behave the same as would a key-based lookup.

PR #1130, opened 2020-11-16
(Make changes to) persist refresh token in joined flow.

rpdome:
related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1430

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
@@ -47,6 +47,7 @@ import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftToken
 import com.microsoft.identity.common.internal.providers.oauth2.AuthorizationRequest;
 import com.microsoft.identity.common.internal.providers.oauth2.OAuth2Strategy;
 import com.microsoft.identity.common.internal.providers.oauth2.OAuth2TokenCache;
+import com.microsoft.identity.common.internal.providers.oauth2.RefreshToken;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -177,11 +178,12 @@ public class BrokerOAuth2TokenCache
      * @return The {@link ICacheRecord} result of this save action.
      * @throws ClientException If the supplied Accounts or Credentials are schema invalid.
      */
-    public ICacheRecord save(@NonNull AccountRecord accountRecord,
-                             @NonNull IdTokenRecord idTokenRecord,
-                             @NonNull AccessTokenRecord accessTokenRecord,
-                             @Nullable String familyId) throws ClientException {
-        final String methodName = ":save";
+    @Deprecated
+    public ICacheRecord save(final @NonNull AccountRecord accountRecord,
+                             final @NonNull IdTokenRecord idTokenRecord,
+                             final @NonNull AccessTokenRecord accessTokenRecord,
+                             final @Nullable String familyId) throws ClientException {
+        final String methodName = ":save (4 args)";
 
         final ICacheRecord result;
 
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/migration/TokenCacheItemMigrationAdapter.java
```diff
@@ -184,7 +185,7 @@ public static boolean tryFociTokenWithGivenClientId(@SuppressWarnings(WarningTyp
         final MicrosoftStsOAuth2Strategy strategy = new MicrosoftStsOAuth2Strategy(config, strategyParameters);
 
         final String refreshToken = refreshTokenRecord.getSecret();
-        final String scopes = BaseController.getDelimitedDefaultScopeString();
+        final String scopes = AuthenticationConstants.OAuth2Scopes.OPEN_ID_SCOPE;
```
shahzaibj, line 188: Why changing this?

AdamBJohnsonx: This seems like a mistake.  What should this be?

rpdome: I was trying to scope it down to the minimum - so that the application that does not require MSGraph does not have to request preauthorization for the other 2 - which is optional in the form.

Maybe it's not a good idea.

iambmelt: Can we send the customary `offline_access openid profile` and not specify a resource? Seems preferable, as MS Graph doesn't exist in all clouds

rpdome: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#openid-connect-scopes

The Microsoft identity platform implementation of OpenID Connect has a few well-defined scopes _**that are also hosted on the Microsoft Graph**_: openid, email, profile, and offline_access. The address and phone OpenID Connect scopes are not supported.

This describes why I was hitting the MSgraph preauthorized error with those scopes.

rpdome: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#openid-connect-scopes

The Microsoft identity platform implementation of OpenID Connect has a few well-defined scopes _**that are also hosted on the Microsoft Graph**_: openid, email, profile, and offline_access. The address and phone OpenID Connect scopes are not supported.

This describes why I was hitting the MSgraph preauthorized error with those scopes (AADSTS65002: Consent between first party application '[CLIENT_ID]' and first party resource '00000003-0000-0000-c000-000000000000' must be configured via preauthorization. Visit https://identitydocs.azurewebsites.net/static/aad/preauthorization.html).

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -180,6 +181,7 @@ public BrokerOAuth2TokenCache(@NonNull Context context,
     public ICacheRecord save(@NonNull AccountRecord accountRecord,
                              @NonNull IdTokenRecord idTokenRecord,
                              @NonNull AccessTokenRecord accessTokenRecord,
+                             @NonNull RefreshTokenRecord refreshTokenRecord,
```
iambmelt, line 184: I would prefer to see new API + tests added for this, rather than have to increment the _major_ version of common to change this API

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -243,13 +247,15 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
             @NonNull final AccountRecord accountRecord,
             @NonNull final IdTokenRecord idTokenRecord,
             @NonNull final AccessTokenRecord accessTokenRecord,
+            @NonNull final RefreshTokenRecord refreshTokenRecord,
```
iambmelt, line 250: Same here -- this would be a major version break. Let's add new API + tests

iambmelt (changes requested):
Let's avoid having to do a major version change on common if we can and make your changes to the cache additive

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -219,6 +220,65 @@ ICacheRecord save(@NonNull AccountRecord accountRecord,
...
+     * @param accessTokenRecord  The {@link AccessTokenRecord} to store.
+     * @param refreshTokenRecord The {@link RefreshTokenRecord} to store.
+     * @return The {@link ICacheRecord} result of this save action.
+     * @throws ClientException If the supplied Accounts or Credentials are schema invalid.
+     * @see OAuth2TokenCache#save(AccountRecord, IdTokenRecord)
+     */
+    ICacheRecord save(final @NonNull AccountRecord accountRecord,
+                      final @NonNull IdTokenRecord idTokenRecord,
+                      final @NonNull AccessTokenRecord accessTokenRecord,
+                      final @NonNull RefreshTokenRecord refreshTokenRecord) throws ClientException {
```
iambmelt, line 235: @rpdome Can we make the RT `@Nullable` in this API and have the older API call this one? At a glance, it looks like that would allow for removal of some duplicated code

rpdome: I did that and reverted back - not liking the idea of having refreshToken nullable. It shouldn't be.

iambmelt: OK - thanks @rpdome

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -253,26 +330,51 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
...
+        final String target = cacheRecord.getAccessToken().getTarget();
+        final String environment = cacheRecord.getAccessToken().getEnvironment();
+
+        final MsalOAuth2TokenCache cache = getTokenCacheForClient(
+                clientId,
+                environment,
+                mCallingProcessUid
+        );
+
+        return (List<ICacheRecord>) cache.loadWithAggregatedAccountData(
```
iambmelt, line 370: Not immediately clear to me why this result needs to be casted, doesn't this API always return this type?

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -238,7 +240,82 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
...
+     *
+     * @param accountRecord      The {@link AccountRecord} to store.
+     * @param idTokenRecord      The {@link IdTokenRecord} to store.
+     * @param accessTokenRecord  The {@link AccessTokenRecord} to store.
+     * @param refreshTokenRecord The {@link RefreshTokenRecord} to store.
+     * @param familyId           The family_id or null, if not applicable.
+     * @return The {@link ICacheRecord} result of this save action.
+     * @throws ClientException If the supplied Accounts or Credentials are schema invalid.
+     */
+    public ICacheRecord save(@NonNull AccountRecord accountRecord,
```
iambmelt, line 254: Nit: params final

iambmelt (approved):
Couple of minor comments

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
rpdome: This is automatically generated by the extract function in Android Studio. Will remove 😅

PR #1131, opened 2020-11-16
Add new publishing repository; Modify repository search order

shoatman:
- Pointed common gradle root project to new feed
- Added new feed as a repository to publish to

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -274,6 +274,14 @@ afterEvaluate {
                     password project.vstsPassword
                 }
             }
+            maven {
+                name "vsts-maven-android"
+                url 'https://identitydivision.pkgs.visualstudio.com/IDDP/_packaging/Android/maven/v1'
+                credentials {
+                    username project.vstsUsername
+                    password project.vstsPassword
+                }
+            }
         }
     }
 }
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -286,21 +286,6 @@ afterEvaluate {
     }
 }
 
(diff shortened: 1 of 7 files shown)


common/build.gradle
```diff
@@ -274,6 +274,14 @@ afterEvaluate {
...
                 }
             }
+            maven {
+                name "vsts-maven-android"
+                url 'https://identitydivision.pkgs.visualstudio.com/IDDP/_packaging/Android/maven/v1'
+                credentials {
+                    username project.vstsUsername
+                    password project.vstsPassword
+                }
+            }
```
shahzaibj, line 284: Do we need to do this? Given that we already did in project level build.gradle file

common/versioning/version.properties
```diff
@@ -1,4 +1,5 @@
-#Mon Nov 16 21:58:07 UTC 2020
+#Mon Nov 16 13:39:37 PST 2020
 versionName=3.0.6
 versionCode=1
-latestPatchVersion=50
+latestPatchVersion=51
+
```
shahzaibj, line 5: nit: extra line can be removed

common/build.gradle
shoatman: Nope.  I just removed it... as it was breaking the VSTS pipeline.

shahzaibj: Okay that was for a different feed...resolving

shahzaibj (approved):

iambmelt (approved):

PR #1132, opened 2020-11-17
Add a warning for potentially needless use of setSecretKey

AdamBJohnsonx:


tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -108,11 +112,15 @@
...
      */
-    public void setSecretKey(byte[] rawKey) {
+    public synchronized void setSecretKey(byte[] rawKey) {
         if (rawKey == null || rawKey.length != SECRET_RAW_KEY_LENGTH) {
             throw new IllegalArgumentException("rawKey");
         }
-
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            Logger.warn(":setSecretKey", "You're using setSecretKey in a version of android" +
+                    "that supports keyStore functionality.  Consider not doing this, as it only exists" +
```
AdamBJohnsonx, line 121:
```suggestion
                    "that supports keyStore functionality.  Consider not doing this, as it only exists " +
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -108,11 +112,15 @@
...
      * @param rawKey App related key to use in encrypt/decrypt
      */
-    public void setSecretKey(byte[] rawKey) {
+    public synchronized void setSecretKey(byte[] rawKey) {
         if (rawKey == null || rawKey.length != SECRET_RAW_KEY_LENGTH) {
             throw new IllegalArgumentException("rawKey");
         }
-
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            Logger.warn(":setSecretKey", "You're using setSecretKey in a version of android" +
```
AdamBJohnsonx, line 120:
```suggestion
            Logger.warn(":setSecretKey", "You're using setSecretKey in a version of android " +
```

AdamBJohnsonx:
This is in relation to https://portal.microsofticm.com/imp/v3/incidents/details/210581721

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -108,11 +112,15 @@
      *
      * @param rawKey App related key to use in encrypt/decrypt
      */
-    public void setSecretKey(byte[] rawKey) {
+    public synchronized void setSecretKey(byte[] rawKey) {
```
AdamBJohnsonx, line 115:
```suggestion
    public void setSecretKey(byte[] rawKey) {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -112,7 +116,11 @@ public void setSecretKey(byte[] rawKey) {
         if (rawKey == null || rawKey.length != SECRET_RAW_KEY_LENGTH) {
             throw new IllegalArgumentException("rawKey");
         }
-
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            Logger.warn(":setSecretKey", "You're using setSecretKey in a version of android " +
+                    "that supports keyStore functionality.  Consider not doing this, as it only exists " +
+                    "for devices with an SDK lower than " + Build.VERSION_CODES.JELLY_BEAN_MR2);
```
shahzaibj, line 122: and are we doing this right now?

iambmelt: We support generating our own key, it happens here:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/9a09e2e4a1fe714773a2b679d054b311d2c56f86/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java#L666

From a customer investigation we were working on earlier, we found an app that was calling this method at startup: we suspect it's causing issues with SSO loss due to the host-app [potentially] providing mismatched keys

shahzaibj: Cool.....actually how does user defined key currently work for MSAL? It is only for ADAL right?

AdamBJohnsonx: MSAL reads the value from the AuthenticationSettings.INSTANCE:

        // Try to get user defined key (ADAL/MSAL).
        if (AuthenticationSettings.INSTANCE.getSecretKeyData() != null) {
            setBlobVersion(VERSION_USER_DEFINED);
            return loadSecretKey(KeyType.ADAL_USER_DEFINED_KEY);
        }

Adam.

iambmelt (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
shahzaibj: Yeah but it has to be set on `AuthenticationSettings.INSTANCE` right? ADAL has a public method to do this...but I don't see one in MSAL

shahzaibj: This is ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/cfdb40c5d0fb28e4825df030e909bc493aaaffe9/adal/src/main/java/com/microsoft/aad/adal/AuthenticationSettings.java#L82

Is there something similar in MSAL?

AdamBJohnsonx: PublicClientApplicationConfiguration.java:167

shahzaibj: Cool, Thanks!

PR #1134, opened 2020-11-17
Javadoc special use-case for setSecretKey

iambmelt:
As title suggests, documents a special case for using this API.

Targets PR #1132

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -28,7 +28,6 @@ import com.microsoft.identity.common.WarningType;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.internal.logging.Logger;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -107,8 +106,18 @@
     }
 
     /**
-     * set raw bytes to derive secretKey to use in encrypt/decrypt. KeySpec
+     * Set raw bytes to derive secretKey to use in encrypt/decrypt. KeySpec
      * algorithm is AES.
+     * <p>
+     * Please note: if a device with an existing installation of the ADAL/MSAL host-app is upgraded
```
shahzaibj, line 112:
```suggestion
     * Please note: If a device with an existing installation of the ADAL/MSAL host-app is upgraded
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -107,8 +106,18 @@
...
+     * from API 17 -> API 18+ then the previously-used secret key data must continued to be supplied
```
shahzaibj, line 113:
```suggestion
     * from API 17 -> API 18+ then the previously-used secret key data must continue to be supplied
```

AdamBJohnsonx (approved):

PR #1135, opened 2020-11-18
Get back to avoiding the storageHelper NPE

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -605,13 +605,17 @@ public SecretKey loadSecretKey(@NonNull final KeyType keyType) throws IOExceptio
 
         switch (keyType) {
             case LEGACY_AUTHENTICATOR_APP_KEY:
-                return getSecretKey(AuthenticationSettings.INSTANCE.getBrokerSecretKeys().get(AZURE_AUTHENTICATOR_APP_PACKAGE_NAME));
+                final byte[] legacySecretKeyData = AuthenticationSettings.INSTANCE.getBrokerSecretKeys().get(AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
+                return legacySecretKeyData == null ? null : getSecretKey(legacySecretKeyData);
```
iambmelt, line 609: What do we expect to happen if we return `null` from any one of the `case` statements in this function?

PR #1136, opened 2020-11-18
Shoatman/fix publishing

shoatman:


.gitignore
```diff
@@ -39,17 +39,7 @@ captures/
...
-.idea/gradle.xml
-.idea/assetWizardSettings.xml
-.idea/dictionaries
-.idea/libraries
-.idea/caches
-# Android Studio 3 in .gitignore file.
-.idea/caches/build_file_checksums.ser
-.idea/modules.xml
-.idea/codeStyles/
+.idea
```
iambmelt, line 42: Should this be something like
`.idea/*`
?

shoatman: it seemed to work and this is what's in MSAL so I used that as inspiration

iambmelt (approved):

PR #1137, opened 2020-11-19
Sends CP version to ESTS and handle WebCP uri.

rpdome:
## Current experience
- Whenever you run into Device Enrollment CA during sign in.... Webview is launched -> You have to sign in (again) -> it shows a page that redirects you to play store.
- The experience is bad, and even more so when Work Profile comes into play (because you'll see this if you're trying to log in in your personal profile - even if you already have everything set up in work profile)
- The scope is originally for duo, but eventually we want to improve this experience for everyone (Duo comes with Broker installed).
	
## The proposed fix
- Broker + MSAL (and soon ADAL) detects the existence of company portal in the current profile.
- It sends the installed version of CP as a query parameter along with the auth request to eSTS. (cpVersion=x)
- eSTS passes that information to WebCP. WebCP uses that information to decide how it should proceed with the flow.
	-There are 4 possible states
		1) the webview does not support CP deeplink.
		2) CP is not installed.
		3) CP is installed, but it's an older version
		4) CP is installed, and it contains the latest change with improved UX. (here)
	- If no values is passed to WebCP, then it's either 1,2 or 3 . In this case, we will proceed with the improved browser flow (extra login no longer required)
- For 4), WebCP will tell eSTS (to redirect us) to companyportal://enrollment. To avoid potential phishing, we hardcoded this url in common - saying that if this link appears, launch CP.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1081,7 +1081,7 @@ public final class AuthenticationConstants {
         /**
          * Redirect URL from WebCP that should launch the Intune Company Portal app.
          */
-        public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = "companyportal://enrollment";
+        public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = BROWSER_EXT_WEB_CP + "enrollment";
 
         /**
          * A query param indicating that this is an intune device CA link.
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1074,6 +1074,16 @@ public final class AuthenticationConstants {
         public static final String BROWSER_EXT_INSTALL_PREFIX = "msauth://";
 
         /**
+         * Prefix in the redirect from WebCP.
+         */
+        public static final String BROWSER_EXT_WEB_CP = "companyportal://";
+
+        /**
+         * Redirect URL from WebCP that should launch the Intune Company Portal app.
+         */
+        public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = "companyportal://enrollment";
+
+        /**
          * A query param indicating that this is an intune device CA link.
          */
         public static final String BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "&ismdmurl=1";
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1091,7 +1091,7 @@ public final class AuthenticationConstants {
         /**
          * Activity name to launch company portal.
          */
-        public static final String COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME + ".views.SplashActivity";
+        public static final String COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME = Broker.COMPANY_PORTAL_PROD_APP_PACKAGE_NAME + ".views.SplashActivity";
 
         /**
          * Redirect URI parameter key to get link to install broker
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
@@ -41,7 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
-import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_PROD_APP_PACKAGE_NAME;
 
 /**
  * A class holding the state of the Authorization Request (OAuth 2.0).
@@ -253,7 +253,7 @@ public abstract class AuthorizationRequest<T extends AuthorizationRequest<T>> im
         public Builder<B> setCpInstallationDetail(final @NonNull Context context) {
             try {
                 final PackageInfo packageInfo =
-                        context.getPackageManager().getPackageInfo(COMPANY_PORTAL_APP_PACKAGE_NAME, 0);
+                        context.getPackageManager().getPackageInfo(COMPANY_PORTAL_PROD_APP_PACKAGE_NAME, 0);
                 mCpVersion = packageInfo.versionName;
             } catch (final PackageManager.NameNotFoundException e) {
                 // CP is not installed. No need to do anything.
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
@@ -22,8 +22,13 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.internal.providers.oauth2;
 
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.util.Pair;
+
+import androidx.annotation.NonNull;
 
 import com.google.gson.annotations.Expose;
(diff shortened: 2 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1055,6 +1055,16 @@
...
 
+        /**
+         * Prefix in the redirect from WebCP.
+         */
+        public static final String BROWSER_EXT_WEB_CP = "companyportal://";
+
+        /**
+         * Redirect URL from WebCP that should launch the Intune Company Portal app.
+         */
+        public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = "companyportal://enrollment";
```
AdamBJohnsonx, line 1066: nit
```suggestion
        public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = BROWSER_EXT_WEB_CP + "enrollment";
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -161,6 +161,9 @@ private boolean handleUrl(final WebView view, final String url) {
         } else if (isInstallRequestUrl(formattedURL)) {
             Logger.info(TAG, "It is an install request");
             return processInstallRequest(view, url);
+        } else if (isWebCpUrl(formattedURL)) {
```
AdamBJohnsonx, line 164: Update method docs with additional case.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -161,6 +161,9 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            Logger.info(TAG, "It is a request from WebCP");
+            return processWebCpRequest(view, url);
```
AdamBJohnsonx, line 166: should we write this as a getUriHanlder method that returns a BiFunction that we call to handle the URI and view?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -221,26 +235,39 @@ public B setClaims(String claims) {
...
+
+        public Builder<B> setCpInstallationDetail(final @NonNull Context context) {
+            try {
+                // Do not use COMPANY_PORTAL_APP_PACKAGE_NAME, since we actually want this to be
+                // CP's package name - not brokerHost's - even in localDebug mode.
+                final PackageInfo packageInfo =
+                        context.getPackageManager().getPackageInfo("com.microsoft.windowsintune.companyportal", 0);
+                mCpVersion = packageInfo.versionName;
+            } catch (final PackageManager.NameNotFoundException e) {
+                // CP is not installed. No need to do anything.
```
AdamBJohnsonx, line 261: Should we take charge of logging this here, so we can put in the logs the package name we tried to launch?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
```diff
@@ -221,26 +235,39 @@ public B setClaims(String claims) {
...
+            mWebViewZoomControlsEnabled = webViewZoomControlsEnabled;
+            return self();
+        }
+
+        public Builder<B> setCpInstallationDetail(final @NonNull Context context) {
+            try {
+                // Do not use COMPANY_PORTAL_APP_PACKAGE_NAME, since we actually want this to be
+                // CP's package name - not brokerHost's - even in localDebug mode.
+                final PackageInfo packageInfo =
+                        context.getPackageManager().getPackageInfo("com.microsoft.windowsintune.companyportal", 0);
```
AdamBJohnsonx, line 258: nitpick: make the string a constant.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationRequest.java
rpdome: I intentionally avoid logging here because otherwise this will spam the logs.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1073,6 +1073,16 @@
          */
         public static final String BROWSER_EXT_INSTALL_PREFIX = "msauth://";
 
+        /**
+         * Prefix in the redirect from WebCP.
```
AdamBJohnsonx, line 1077: This scheme we're using, does this provide a deep link into CompanyPortal?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -248,19 +256,9 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
             //       Until that comes, we'll only handle this in ipphone.
             if (packageHelper.isPackageInstalledAndEnabled(applicationContext, IPPHONE_APP_PACKAGE_NAME) &&
                     IPPHONE_APP_SIGNATURE.equals(packageHelper.getCurrentSignatureForPackage(IPPHONE_APP_PACKAGE_NAME)) &&
-                    packageHelper.isPackageInstalledAndEnabled(applicationContext, COMPANY_PORTAL_APP_PACKAGE_NAME)) {
+                    packageHelper.isPackageInstalledAndEnabled(applicationContext, COMPANY_PORTAL_PROD_APP_PACKAGE_NAME)) {
```
AdamBJohnsonx, line 259: I'm assuming that this means that we don't want to link into BrokerHostApp.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -296,6 +310,20 @@ private void openLinkInBrowser(final String url) {
         }
     }
 
+    private boolean processWebCpRequest(@NonNull final WebView view, @NonNull final String url) {
+
+        view.stopLoading();
+        
+        if (url.equalsIgnoreCase(AuthenticationConstants.Broker.WEBCP_LAUNCH_COMPANY_PORTAL_URL)) {
```
AdamBJohnsonx, line 317: do we want to be case-insensitive here?

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx: In other words, can we say what it does in a comment or a link to a doc?

tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
rpdome: That's correct.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
rpdome: I'm inclined toward a 'no'.

While we don't really have an agreement on case-sensitivity (it's more like "I'll give you X and you do Y"), I don't think client side should break if the server side accidentally ship an upper case bug (we've seen that before).

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
rpdome: We steered away from deep linking because any apps can 'declare' the ownership of the link (in such case, Android will let the user choose which app to redirect to - and we don't even want this to happen).

Therefore, what we do here is to 'look for the pattern' here and launch an intent directly with the app package name. we won't be using the URI directly.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
AdamBJohnsonx: Can we say that here?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx: Again, could we add a comment saying what we're doing and why?

PR #1139, opened 2020-11-25
Implementing NavigateTo function in BrowserChrome + Reuse handle aad AccountPicker for sign out as well

kkrishnareddy:
Implementing NavigateTo function in Browser Chrome. if we are on home screen of tabs in chrome then only it works.

Also refactor handleAccountPicker in AadLoginComponentHandler for sign out as well

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
@@ -22,8 +22,6 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.ui.automation.browser;
 
-import android.app.Instrumentation;
-
 import androidx.annotation.NonNull;
 import androidx.test.platform.app.InstrumentationRegistry;
 import androidx.test.uiautomator.UiDevice;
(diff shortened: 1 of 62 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/IBrowser.java
```diff
@@ -34,6 +36,6 @@
      *
      * @param url the url to open
      */
-    void navigateTo(String url);
+    void navigateTo(String url) throws UiObjectNotFoundException;
```
shahzaibj, line 39: I don't think an exception is thrown here....why do we need the throws clause?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
```diff
@@ -46,7 +53,11 @@ public void handleFirstRun() {
     }
 
     @Override
-    public void navigateTo(@NonNull final String url) {
+    public void navigateTo(@NonNull final String url) throws UiObjectNotFoundException {
```
shahzaibj, line 56: I don't think an exception is thrown here....why do we need the throws clause?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
```diff
@@ -46,7 +53,11 @@ public void handleFirstRun() {
...
         //TODO: implement browsing for Chrome
```
shahzaibj, line 57: We can remove the TODO statement now

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/IBrowser.java
kkrishnareddy: yes, initially I used the .click() method that's why it is causing an exception.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
kkrishnareddy: yes, initially I used the .click() method that's why it is causing an exception.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserChrome.java
kkrishnareddy: removed the TODO statement.

shahzaibj (approved):

PR #1146, opened 2020-12-03
Check for eligible for caching before putting in executing command map

shahzaibj:
Only commands eligible for caching should be put in the executing command map....all other commands we don't care if there are multiple duplicate commands in parallel.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
@@ -81,9 +81,4 @@ public class InteractiveTokenCommand extends TokenCommand {
     public boolean isEligibleForEstsTelemetry() {
         return true;
     }
-
-    @Override
-    public boolean isEligibleForCaching() {
-        return true;
-    }
 }
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+vNext
+- Check for eligible for caching when putting command in executing command map
+
 Version 3.0.8
 ----------
 - Hardcode Teams Agent clientID/scope for FoCi call (#1140)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 vNext
+----------
 - Check for eligible for caching when putting command in executing command map
 
 Version 3.0.8
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -140,26 +140,32 @@ public class CommandDispatcher {
 
         final Handler handler = new Handler(Looper.getMainLooper());
         synchronized (mapAccessLock) {
-            FinalizableResultFuture<CommandResult> future = sExecutingCommandMap.get(command);
+            final FinalizableResultFuture<CommandResult> finalFuture;
+            if (command.isEligibleForCaching()) {
+                FinalizableResultFuture<CommandResult> future = sExecutingCommandMap.get(command);
 
-            if (null == future) {
-                future = new FinalizableResultFuture<>();
-                final FinalizableResultFuture<CommandResult> putValue = sExecutingCommandMap.putIfAbsent(command, future);
+                if (null == future) {
+                    future = new FinalizableResultFuture<>();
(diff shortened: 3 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -81,9 +81,4 @@ public void notify(int requestCode, int resultCode, final Intent data) {
     public boolean isEligibleForEstsTelemetry() {
         return true;
     }
-
```
tanmaymanolkar1, line 84: @shahzaibj why this change exactly?

tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
shahzaibj: This was accidently set to true but it should be false....with that said since this interactive there is no impact where this is true or false because interactive requests don't go through the submitSilent method in Command Dispatcher

iambmelt:
@shahzaibj Is this a bugfix? Can you provide a description in the PR?

shahzaibj:
> @shahzaibj Is this a bugfix? Can you provide a description in the PR?

@iambmelt Updated description....kind of a bug...we seemed to have changed the code around this recently here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1012/files

Basically only commands eligible for caching should be put in the executing command map....all other commands we don't care if there are multiple duplicate commands in parallel.

So I am just going back to old behavior because CP expects to get a new token for each DCF and the behavior is confusing

iambmelt (approved):

AdamBJohnsonx:
> > @shahzaibj Is this a bugfix? Can you provide a description in the PR?
> 
> @iambmelt Updated description....kind of a bug...we seemed to have changed the code around this recently here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1012/files
> 
> Basically only commands eligible for caching should be put in the executing command map....all other commands we don't care if there are multiple duplicate commands in parallel.
> 
> So I am just going back to old behavior because CP expects to get a new token for each DCF and the behavior is confusing

I don't know why they care.  They're asking for a token with certain characteristics and we're giving them one.  Whether they get two that behave identically or two copies of the same one should be immaterial, really.

Did we get to ground on why they're not seeing a callback on their second request?  That's what confuses me.

shahzaibj:
@AdamBJohnsonx Yeah, so they were wrong when they said that no callback is invoked....I asked for clarification and the callback is indeed invoked...what is happening is that the DCF token request has a different type of callback where in addition to `cancel, error, onTaskCompleted` there is also a `onUserCodeReceived` which passes the user code and once they redeem that on the different device then they get the token and `onTaskCompleted` is hit -- so what is happening with multiple DeviceCodeFlowCommands in parallel is that we don't execute the second because the first one is running and then when the first one is completed then we hit the `onTaskCompleted` for both of these....so for the second one the user never hits `onUserCodeRecevied` and directly gets the token without actually getting any user code i.e. they get the same token that they got while redeeming the user code from the first one....so CP / customer seems to find this a bit confusing....so the change that I've made here to exempt DCF commands from currently executing command map should be all that is needed here.

AdamBJohnsonx:
> @AdamBJohnsonx Yeah, so they were wrong when they said that no callback is invoked....I asked for clarification and the callback is indeed invoked...what is happening is that the DCF token request has a different type of callback where in addition to `cancel, error, onTaskCompleted` there is also a `onUserCodeReceived` which passes the user code and once they redeem that on the different device then they get the token and `onTaskCompleted` is hit -- so what is happening with multiple DeviceCodeFlowCommands in parallel is that we don't execute the second because the first one is running and then when the first one is completed then we hit the `onTaskCompleted` for both of these....so for the second one the user never hits `onUserCodeRecevied` and directly gets the token without actually getting any user code i.e. they get the same token that they got while redeeming the user code from the first one....so CP / customer seems to find this a bit confusing....so the change that I've made here to exempt DCF commands from currently executing command map should be all that is needed here.

Oh, good.  Yeah, I can see where that might be an issue in the completely general case where you can have any number of processes interacting with the library doing this kind of silent-interactive auth, though we really should be trying to separate the caching for such things.  I am curious how they ended up hitting this case, though, and why.

PR #1148, opened 2020-12-05
Just fix a few lines that generate warnings

AdamBJohnsonx:


testutils/build.gradle
```diff
@@ -10,7 +10,6 @@ project.ext.vstsPassword = System.getenv("ENV_VSTS_MVN_ANDROIDCOMMON_ACCESSTOKEN
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
```
shahzaibj, line 13: Does removing this have any impact? What version is going to be used now?

AdamBJohnsonx: From: https://developer.android.com/studio/releases/gradle-plugin#new_features : You no longer need to specify a version for the build tools (so, you can now remove the android.buildToolsVersion property). By default, the plugin automatically uses the minimum required build tools version for the version of Android plugin you're using.

It actually generates a warning that this line will be ignored, and that it's using 29.0.3 instead.

AdamBJohnsonx: Looks like this: WARNING: The specified Android SDK Build Tools version (28.0.3) is ignored, as it is below the minimum supported version (29.0.2) for Android Gradle Plugin 4.1.1.
Android SDK Build Tools 29.0.2 will be used.

PR #1149, opened 2020-12-05
Remove about:blank initial page load from WebViewAuthorizationFragment

iambmelt:
Impetus:
https://portal.microsofticm.com/imp/v3/incidents/details/212200669/home

Current behavior:
- After creating the `WebView` in `onCreateView()`, `about:blank` is loaded -- it is unclear why this is being done...
- When handling `onBackPressed()` events, this page _should_ be factored in when determining if the hosting `Activity` is being dismissed or not however, this seems to not be totally reliable as I can intermittently 'back' onto this blank screen
- Reference incident ID: `UBT8DZ5P` for incorrect behavior

New behavior:
- Don't load `about:blank` -- this view would have just been created, as `onCreateView()` is at the beginning of the lifecycle (see graphic below)
- Remove this `-2` history check and go `goBack()` if possible or `cancel()` if you can't
- Reference incident ID: `S57C2G76` for correct behavior

image

>Note: I'm not clear on what the original purpose of this `about:blank` page load was, it does not seem to serve a purpose and the accompanying comment is too vague for me to determine what issue it's trying to solve

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -148,8 +148,6 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
         mWebView.post(new Runnable() {
             @Override
             public void run() {
-                // load blank first to avoid error for not loading webView
-                mWebView.loadUrl("about:blank");
                 Logger.info(TAG + methodName, "Launching embedded WebView for acquiring auth code.");
                 Logger.infoPII(TAG + methodName, "The start url is " + mAuthorizationRequestUrl);
                 mWebView.loadUrl(mAuthorizationRequestUrl, mRequestHeaders);
@@ -171,18 +169,14 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
     @Override
     public boolean onBackPressed() {
         Logger.info(TAG, "Back button is pressed");
-        if (null != mWebView && mWebView.canGoBack()) {
-            // User should be able to click back button to cancel. Counting blank page as well.
-            final int BACK_PRESSED_STEPS = -2;
-            if (!mWebView.canGoBackOrForward(BACK_PRESSED_STEPS)) {
-                cancelAuthorization(true);
-            } else {
-                mWebView.goBack();
-            }
-            return true;
+
+        if (mWebView.canGoBack()) {
+            mWebView.goBack();
+        } else {
+            cancelAuthorization(true);
         }
 
-        return false;
+        return true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


iambmelt:
I will plan to produce a test-broker based on these bits for fix verification by the customer before shipping to production

iambmelt:
> I will plan to produce a test-broker based on these bits for fix verification by the customer before shipping to production

Verified correct behavior via logs in Incident ID: 7A4BACES

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -171,18 +169,14 @@ public void run() {
...
-            // User should be able to click back button to cancel. Counting blank page as well.
-            final int BACK_PRESSED_STEPS = -2;
-            if (!mWebView.canGoBackOrForward(BACK_PRESSED_STEPS)) {
-                cancelAuthorization(true);
-            } else {
-                mWebView.goBack();
-            }
-            return true;
+
+        if (mWebView.canGoBack()) {
```
AdamBJohnsonx, line 173: We lost a null check here.  Did we need it?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -171,18 +169,14 @@ public void run() {
...
+            mWebView.goBack();
+        } else {
+            cancelAuthorization(true);
         }
 
-        return false;
+        return true;
```
AdamBJohnsonx, line 179: This used to return true iff we had a WebView and it could go back, false otherwise.  Do we need to?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
iambmelt: Nope, not sure why this was here to begin with. The view will be created as part of the Fragment's start-up lifecycle

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
iambmelt: Nope, unclear to me why this was ever written this way -- `false` should be used by this Fragment to signal "I did not handle this event" but, if it's on-screen, there is _always_ something for this Fragment to do: either `goBack()` or self-cancel. Irrespective of which happens we should `return true`, signaling we handled the `onBackPressed()` event ourselves

AdamBJohnsonx (approved):

PR #1157, opened 2020-12-11
Sanitize error message for csv to avoid spilling to multiple lines

shahzaibj:
We don't fields in csv to span multiple lines due to newline or return characters

### DIFF
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/kusto/CaptureKustoTestResultRule.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/kusto/CaptureKustoTestResultRule.java
@@ -83,6 +83,6 @@ public class CaptureKustoTestResultRule implements TestRule {
     }
 
     private String sanitizeErrorMessage(@NonNull final String msg) {
-        return msg.replaceAll("\n", " ").replaceAll("\r"," ");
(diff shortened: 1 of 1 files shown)


testutils/src/main/java/com/microsoft/identity/internal/testutils/kusto/CaptureKustoTestResultRule.java
```diff
@@ -79,4 +81,8 @@ public void evaluate() throws Throwable {
             }
         };
     }
+
+    private String sanitizeErrorMessage(@NonNull final String msg) {
+        return msg.replaceAll("\n", " ").replaceAll("\r"," ");
```
AdamBJohnsonx, line 86: I think you can use a character class here, and avoid creating two string instances.

AdamBJohnsonx (approved):
minor change.

testutils/src/main/java/com/microsoft/identity/internal/testutils/kusto/CaptureKustoTestResultRule.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1157/commits/32193bb3490e1b1b4ff0d55ca72dd0ef357d6e68

PR #1163, opened 2020-12-17
Temporarily revert jacoco integration to unblock Edge

shahzaibj:
Edge was running into a build issue due to Jacoco...reverting the change for now....we will revisit...investigate and fix the issue and then re-enable jacoco integration. cc/ @ayushmeghwani

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -3,12 +3,12 @@ apply plugin: 'pmd'
 apply plugin: 'checkstyle'
 apply plugin: 'maven-publish'
 apply from: 'versioning/version_tasks.gradle'
-apply plugin: 'jacoco'
+//apply plugin: 'jacoco'
 
 // To include Robolectric tests in the Jacoco report, flag -> "includeNolocationClasses" is set to true
-tasks.withType(Test) {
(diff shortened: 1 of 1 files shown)


common/build.gradle
```diff
@@ -99,10 +99,10 @@ android {
...
 
         // create separate code coverage tasks(for unitTests, androidTests and both Tests) for every build variant
-        def buildVariant = variant.name
-        createCoverageTask(buildVariant, unitTestType)
-        createCoverageTask(buildVariant, androidTestType)
-        createCoverageTask(buildVariant, bothTestType)
+//        def buildVariant = variant.name
+//        createCoverageTask(buildVariant, unitTestType)
+//        createCoverageTask(buildVariant, androidTestType)
+//        createCoverageTask(buildVariant, bothTestType)
```
AdamBJohnsonx, line 105: we should really just not create these tasks for non-debug variants.

AdamBJohnsonx (approved):

common/build.gradle
shahzaibj: Yes I will follow up with Ayush to fix this in a new PR

PR #1169, opened 2020-12-21
Set up CI with Azure Pipelines

shoatman:
Trying out new self-hosted build agent with docker and docker images already installed and pulled.


azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+    echo =============================================
+    docker run --privileged --cpus="3" --memory="12g" -v "$PWD":/home/gradle/ -w /home/gradle/ authclient.azurecr.io/samples/dbi-instrumented-api30 gradle clean
+    echo =============================================
+    echo prune containers to avoid running out of disk space - shutdown containers still exist on disk
+    echo =============================================
+    docker system prune -f
+    echo =============================================
+    echo Dump environment variables for build agent
+    echo =============================================
+    env
```
AdamBJohnsonx, line 35: some reason to do this?  It seems strange.

azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+- script: |
+    docker --version
+    echo =============================================
+    echo Kill all running containers if existing
+    echo =============================================
+    docker container kill $(docker ps -q)
+    echo =============================================
+    echo Run unit and instrumented inside docker container
+    echo =============================================
+    docker run --privileged --cpus="3" --memory="12g" -v "$PWD":/home/gradle/ -w /home/gradle/ authclient.azurecr.io/samples/dbi-instrumented-api30 sh scripts/run-instrumented-tests.sh
```
AdamBJohnsonx, line 21: It'd be pretty cool to cache these containers and script their creation.  So when we move to API31, all we need to do is change this number.  Can the API version be a variable?

azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+  displayName: 'Build and test inside docker container'
+- script: |
+    echo =============================================
+    echo Cleaning up build output that is owned by docker user rather than agent user
+    echo =============================================
+    docker run --privileged --cpus="3" --memory="12g" -v "$PWD":/home/gradle/ -w /home/gradle/ authclient.azurecr.io/samples/dbi-instrumented-api30 gradle clean
```
AdamBJohnsonx, line 27: Yeah, if the API version and/or docker target can become variables so we don't need to keep them in sync manually that would help the next person not mess it up.

azure-pipelines.yml
shoatman: Mostly it's for debugging.  In case something unexpected happens.

azure-pipelines.yml
shoatman: I'm going to check the handlebars templates for the docker files into android-complete, but yes the docker file generation is template and you can change several different variables to produce a different image include: build tools version, api level, type of image (google_apis, default, etc), abi of image, JDK version, etc....

azure-pipelines.yml
shoatman: The images themselves are specific to API versions.  At some point I want to play with a matrix pipeline where the instrumented tests are run against multiple different docker images each with their own api level configuration.

azure-pipelines.yml
shoatman: The container image is cached in an azure container registry.  Hence the long image name which is a prefix of the azure container registry service endpoint.

scripts/run-instrumented-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+echo =============================================
+echo Starting ADB Daemon
+echo =============================================
+adb start-server
+emulator @test -no-window -no-audio -wipe-data &
+sleep 30
```
AdamBJohnsonx, line 7: sleep 30 the best we can do?  Maybe something here: https://stackoverflow.com/questions/3634041/detect-when-android-emulator-is-fully-booted or here https://androidstudio.googleblog.com/2019/05/emulator-2907-canary.html ?

maybe -no-boot-anim ?

shoatman: It's a headless emulator.  adb wait-for-device was hanging so i added the sleep for now while investigating other alternatives.   
 It's not actually necessary... more a reminder to future engineers....The emulator currently starts in 1-1.5 minutes and gradle is currently looking for depedencies for at least 4 minutes.

scripts/run-instrumented-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+gradle -version
```
AdamBJohnsonx, line 8: Can we make this more prominent, like including it in the banner below with backticks?

shoatman: All of the commands or something specific?  I'm fine with making all of this more prominent. :-)

azure-pipelines.yml
AdamBJohnsonx: Main feedback here is "can we put the long name into a variable in the pipeline setup file so we don't have to repeat it exactly".

azure-pipelines.yml
AdamBJohnsonx: I'm fine with the long name. ;-)

scripts/run-instrumented-tests.sh
AdamBJohnsonx: Cool.  I do like scripting that says "Sleeping for <x> <timeunits> because <reason>" so that operators reading the term know what's happening.  I'd rather have it be bulletproof so that we can't break it by fixing the build speed (in some better, future world)... what happens if it outruns the emulator, though?

AdamBJohnsonx (approved):
My comments here re all minor

scripts/run-instrumented-tests.sh
AdamBJohnsonx: It doesn't do a good job of this, but the Gradle version being run.

azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+  clean: all
+steps:
+- script: |
+    docker --version
+    echo =============================================
+    echo Kill all running containers if existing
+    echo =============================================
+    docker container kill $(docker ps -q)
+    echo =============================================
+    echo Run unit and instrumented inside docker container
```
iamgusain, line 19: nit: Run unit and instrumented **tests** inside docker container

iamgusain (approved):

PR #1174, opened 2020-12-22
converting Assert.fail statements into throw new Assertion error

kkrishnareddy:
converting Assert.fail(e.getmessage()) statements into throw new AssertionError(e) whenever possible.


shahzaibj (approved):
LGTM!

PR #1180, opened 2020-12-30
Break load url loop on request completion

dabbcomputers:
Current behavior: even when the call gets completed, web view try for 50 times to search for the component in UI automation tests which slows down the system and cause timeout in tests.
Fix: if the status already returned don't call loadUrl again -- this will speed avoid unnecessary webView load url calls and improve flakyness in tests.


shoatman (approved):
lgtm

PR #1185, opened 2021-01-06
Basic Integration of Code Marker Tool in Common

neeravagarwalmicrosoft:
### Description:
- Most of the information about this work is consolidated on the Perf Testing page on OneNote

### Testing:
Tested on Pipeline: https://dev.azure.com/IdentityDivision/IDDP/_build?definitionId=1254&_a=summary

### Report:
perf result


common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,47 @@
+package com.microsoft.identity.common;
```
shahzaibj, line 1: Add license

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class CodeMarker {
```
shahzaibj, line 3: Add javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,47 @@
...
+    private int marker;
+    private long timeMilliseconds;
```
shahzaibj, line 5:
```suggestion
    private long timeInMilliSeconds;
```

shahzaibj: Can we also add a description of what this time represents?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,47 @@
...
+        this.marker = marker;
+    }
+
+    public long getTimeMilliseconds() {
+        return timeMilliseconds;
+    }
+
+    public void setTimeMilliseconds(long timeMilliseconds) {
+        this.timeMilliseconds = timeMilliseconds;
+    }
```
shahzaibj, line 46: We could potentially use lombok to generate getters and setters

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,75 @@
+package com.microsoft.identity.common;
```
shahzaibj, line 1: Missing license

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+public class CodeMarkerManager {
```
shahzaibj, line 15: Add javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,75 @@
...
+    public static String getFileContent() {
+        String stringToWrite = "TimeStamp,Marker,Time,Thread,CpuUsed,CpuTotal,ResidentSize,VirtualSize,WifiSent,WifiRecv,WwanSent,WwanRecv,AppSent,AppRecv,Battery,SystemDiskRead,SystemDiskWrite";
+
+        for(CodeMarker codeMarker : codeMarkers) {
+            stringToWrite += "\n"
+                            + codeMarker.getTimeStamp()
+                            + "," + codeMarker.getMarker()
+                            + "," + codeMarker.getTimeMilliseconds()
+                            + "," + codeMarker.getThreadId()
+                            + "," + ",NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA";
```
shahzaibj, line 71: Would be better to use a `StringBuilder`

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
```diff
@@ -0,0 +1,40 @@
...
+//  THE SOFTWARE.
+package com.microsoft.identity.client.ui.automation.logging.formatter;
+
+import android.util.Log;
+
+import com.microsoft.identity.client.ui.automation.logging.LogLevel;
+
+/**
+ * A log formatter that formats logs into a single string following the same format as used by the
+ * the Android {@link Log} utility.
```
shahzaibj, line 31: I don't understand this comment. Isn't this the [LogcatLikeFormatter](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/127a730282452d0477df4b01e110282bb8eca41c/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/LogcatLikeFormatter.java#L34-L38)?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class SimpleTextFormatter implements ILogFormatter {
```
shahzaibj, line 33: Why is this class / formatter needed? What is the use case?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -132,6 +133,7 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     public static FinalizableResultFuture<CommandResult> submitSilentReturningFuture(@SuppressWarnings(WarningType.rawtype_warning) @NonNull final BaseCommand command) {
 
+        CodeMarkerManager.markCode(10011);
```
shahzaibj, line 136: Can we put these marker ids in a constants file?

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,75 @@
...
+    public static void setEnableCodeMarker(boolean enableCodeMarker) {
+        CodeMarkerManager.enableCodeMarker = enableCodeMarker;
+    }
+
+    public static void clear(){
+        codeMarkers.clear();
+    }
+
+    public static void writeToFile(String fileName) {
+        String stringToWrite = getFileContent();
```
shahzaibj, line 44: nit: add `final` here and everywhere if applicable

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,75 @@
...
+        if(enableCodeMarker) {
+            if(codeMarkers.size() >= MAX_SIZE_CODE_MARKER) {
+                clear();
+            }
+            long currentMilliSeconds = System.currentTimeMillis();
+            if (codeMarkers.size() == 0) {
+                baseMilliSeconds = currentMilliSeconds;
+            }
+            SimpleDateFormat f = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
+            codeMarkers.add(new CodeMarker(marker, currentMilliSeconds - baseMilliSeconds, f.format(new Date()), Thread.currentThread().getId()));
```
shahzaibj, line 31: When would the markers be popped out of this list? I see it happens if we hit max size - is that the only time?

shahzaibj (changes requested):
Left some comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
neeravagarwalmicrosoft: Yes, this is similar to LogcatLikeFormatter. 
But LogcatLikeFormatter converts the string into its own format.
While I wrote SimpleTextFormatter which returns the string as it is.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
neeravagarwalmicrosoft: LogcatLikeFormatter converts the string into its own format.
While I wrote SimpleTextFormatter which returns the string as it is.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: As per the new iteration I have pushed, the markers will also be cleared on completion of the test cases "TestCasePerf" and "TestCasePerfBrokerHost".

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Many of the setters were not being used and hence removed. 
Currently, there are only 4 getters and no setters in this class. 
Please suggest should we still use lombok for this?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done. Please let me know if the comment is able to describe the variable

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,57 @@
...
+public class CodeMarker {
+
+    private String marker;
+    // timeInMilliseconds represents time in milliseconds where zero(0) means the time when first codemarker was captured.
+    private long timeInMilliseconds;
+    // timeStamp is the system time at the time of the capture of the codemarker.
+    private String timeStamp;
+    private long threadId;
+
+    public CodeMarker(String marker, long timeInMilliseconds, String timeStamp, long id) {
```
shahzaibj, line 34: add final to vars

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Date;
+import java.util.List;
+
+public class CodeMarkerManager {
+
+    private static boolean enableCodeMarker = false;
+    private static int MAX_SIZE_CODE_MARKER = 1000;
+    private static volatile List<CodeMarker> codeMarkers = new ArrayList<CodeMarker>();
+    private static long baseMilliSeconds = 0;
+    private static String scenarioCode = "";
```
shahzaibj, line 43: could this start as null?

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+            thisLine.append("\n").append(codeMarker.getTimeStamp())
+                    .append(",").append(codeMarker.getMarker())
+                    .append(",").append(codeMarker.getTimeInMilliseconds())
+                    .append(",").append(codeMarker.getThreadId())
+                    .append(",").append(",NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA");
+            stringToWrite += thisLine.toString();
+        }
+        return stringToWrite;
+    }
+}
```
shahzaibj, line 109: Can we write this whole class as a Singleton?

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+        CodeMarkerManager.enableCodeMarker = enableCodeMarker;
+    }
+
+    public static void clearMarkers(){
+        codeMarkers.clear();
+    }
+
+    public static void clearAll() {
+        codeMarkers.clear();
+        CodeMarkerManager.scenarioCode = "";
```
shahzaibj, line 73: can we set it to null here?

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+public class CodeMarkerManager {
+
+    private static boolean enableCodeMarker = false;
+    private static int MAX_SIZE_CODE_MARKER = 1000;
+    private static volatile List<CodeMarker> codeMarkers = new ArrayList<CodeMarker>();
+    private static long baseMilliSeconds = 0;
```
shahzaibj, line 42: Please add some comments describing what this base milli seconds represents

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+            fileWriter.write(stringToWrite);
+            fileWriter.close();
+            fileOutputStream.getFD().sync();
+            fileOutputStream.close();
+            System.out.println("Test");
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
```
shahzaibj, line 92: Instead of printing stack trace, you probably want to log an error statement here

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+        codeMarkers.clear();
+    }
+
+    public static void clearAll() {
+        codeMarkers.clear();
+        CodeMarkerManager.scenarioCode = "";
+    }
+
+    public static void writeToFile(final String fileName) {
+        String stringToWrite = getFileContent();
```
shahzaibj, line 77: add final here and everywhere

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static String getFileContent() {
+        String stringToWrite = "TimeStamp,Marker,Time,Thread,CpuUsed,CpuTotal,ResidentSize,VirtualSize,WifiSent,WifiRecv,WwanSent,WwanRecv,AppSent,AppRecv,Battery,SystemDiskRead,SystemDiskWrite";
+
+        for(CodeMarker codeMarker : codeMarkers) {
+            StringBuilder thisLine = new StringBuilder();
```
shahzaibj, line 99: add final here and everywhere

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
```diff
@@ -0,0 +1,11 @@
+package com.microsoft.identity.common;
+
+public class CodeMarkersConstants {
```
shahzaibj, line 3: add javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
```diff
@@ -0,0 +1,11 @@
...
+    public static String BROKER_PROCESS_START = "10111";
+    public static String BROKER_PROCESS_END = "10120";
+    public static String ACQUIRE_TOKEN_SILENT_START = "10011";
+    public static String ACQUIRE_TOKEN_SILENT_EXECUTOR_START = "10012";
+    public static String ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_START = "10013";
+    public static String ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_END = "10014";
+    public static String ACQUIRE_TOKEN_SILENT_FUTURE_OBJECT_CREATION_END = "10020";
```
shahzaibj, line 10: potentially javadoc these too

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
shahzaibj: Yeah so this javadoc is certainly wrong then and very misleading...we need a better javadoc here

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
shahzaibj: So basically we are throwing away all the other values

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,109 @@
...
+            thisLine.append("\n").append(codeMarker.getTimeStamp())
+                    .append(",").append(codeMarker.getMarker())
+                    .append(",").append(codeMarker.getTimeInMilliseconds())
+                    .append(",").append(codeMarker.getThreadId())
+                    .append(",").append(",NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA,NA");
+            stringToWrite += thisLine.toString();
```
shahzaibj, line 105: This seems like it would be easy to mess up what we write to a file....can we validate on the schema prior to dumping these to a file....or even better if the code here is written in such a way that it always spits out a String that conforms to a schema.....you can probably model the file schema a class object

shahzaibj (changes requested):
Left more comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
neeravagarwalmicrosoft: Yes, we are returning message without adding additional parameters of logging.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: I think initializing with null may create a problem for the case someone does not provide any prefix. 

I initialized with an empty string as if someone does not provide any prefix, it works without a prefix.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
neeravagarwalmicrosoft: Please look at the new javadoc. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: I have now initialized with null now and for that, also taken care of the case when reading the null value.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done as part of initialization to null. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: This method is no longer part of this class now as file writing will be done by FileAppender.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Need to discuss this. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
shahzaibj: It is not necessary..resolving.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,61 @@
...
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common;
+
+/**
+ * A Plain Object class containing information of a code marker which is an event in code.
```
shahzaibj, line 26: I'm not sure what you mean by `Plain Object class`....I guess you could just say `Class`?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,61 @@
...
+ * marker is a string which is prefixed by scenario code and is defined in class CodeMarkersConstants.
```
shahzaibj, line 27:
```suggestion
 * Marker is a string which is prefixed by scenario code and is defined in class CodeMarkersConstants.
```

shahzaibj, line 27: Can link to `CodeMarkersConstants` in javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class CodeMarker {
+
+    private String marker;
+    // timeInMilliseconds represents time in milliseconds where zero(0) means the time when first codemarker was captured.
```
shahzaibj, line 32: What does a non-zero value represent?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,61 @@
...
+    private long timeInMilliseconds;
+    // timeStamp is the system time at the time of the capture of the codemarker.
+    private String timeStamp;
+    private long threadId;
+
+    public CodeMarker(final String marker, final long timeInMilliseconds, final String timeStamp, final long id) {
```
shahzaibj, line 38:
```suggestion
    public CodeMarker(final String marker, final long timeInMilliseconds, final String timeStamp, final long threadId) {
```

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * A utility which collects any event's information given by calling markCode on such event(s).
+ * These events can be retrieved in csv format by calling method getFileContent.
```
shahzaibj, line 39: I really don't like how we are calling out method names here in class javadoc....I think class javadoc needs to be simple in a way that someone reading it doesn't need to go and see what methods are in the class to fully understand the javadoc....with that said, I'm not against it either, however, if we do want to call out method names inside the class javadoc then we should link to the method using `@link` notation whenever we choose to do so.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,113 @@
...
+ * The event is recognized by a string marker which is prefixed with scenario code.
+ * MAX_SIZE_CODE_MARKER is the maximum number of markers this utility can have.
```
shahzaibj, line 41: This statement should go on the constant itself

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,113 @@
...
+    private int MAX_SIZE_CODE_MARKER = 1000;
+    private volatile List<CodeMarker> codeMarkers = new ArrayList<CodeMarker>();
+
+    //baseMilliSeconds is the time in milliseconds when first codemarker was captured.
+    private long baseMilliSeconds = 0;
+    private String scenarioCode = null;
+    private static CodeMarkerManager instance = null;
+
+    public static CodeMarkerManager getInstance() {
+        if(CodeMarkerManager.instance == null) {
```
shahzaibj, line 55: I don't think this check is needed twice i.e. you are already doing it inside the `synchronized` block...I don't think you need to do it here

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,113 @@
...
+public class CodeMarkerManager {
+
+    private boolean enableCodeMarker = false;
+    private int MAX_SIZE_CODE_MARKER = 1000;
+    private volatile List<CodeMarker> codeMarkers = new ArrayList<CodeMarker>();
+
+    //baseMilliSeconds is the time in milliseconds when first codemarker was captured.
+    private long baseMilliSeconds = 0;
+    private String scenarioCode = null;
+    private static CodeMarkerManager instance = null;
```
shahzaibj, line 52:
```suggestion
    private static CodeMarkerManager sCodeMarkerManager = null;
```

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * A utility which collects any event's information given by calling markCode on such event(s).
+ * These events can be retrieved in csv format by calling method getFileContent.
+ * The event is recognized by a string marker which is prefixed with scenario code.
+ * MAX_SIZE_CODE_MARKER is the maximum number of markers this utility can have.
+ */
+public class CodeMarkerManager {
+
+    private boolean enableCodeMarker = false;
+    private int MAX_SIZE_CODE_MARKER = 1000;
```
shahzaibj, line 46: This probably should be `static`

shahzaibj: as well as `final`

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.identity.common;
+
+/**
+ * A class having the codemarker value of particular events descriptions.
+ * one or many of these can be used while capturing codemarker event by sending as an artument in method call to markCode of class CodeMarkerManager.
+ */
+public class CodeMarkersConstants {
```
shahzaibj, line 7:
```suggestion
public class CodeMarkerConstants {
```

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.identity.common;
+
+/**
+ * A class having the codemarker value of particular events descriptions.
```
shahzaibj, line 4:
```suggestion
 * A class having the code marker value of particular event's description.
```

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
```diff
@@ -0,0 +1,15 @@
...
+ * one or many of these can be used while capturing codemarker event by sending as an artument in method call to markCode of class CodeMarkerManager.
```
shahzaibj, line 5:
```suggestion
 * One or many of these can be used while capturing codemarker event by sending as an argument in method call to markCode of class CodeMarkerManager.
```

shahzaibj: As a person reading this javadoc...I don't know what `markCode` is and I don't know what is `CodeMarkerManager`....I don't want to go around finding these....can we just look to them in javadoc?

shahzaibj (changes requested):
Left some comments...also build is failing due to PMD errors...so let's fix that too. Thanks!

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: I have used "Double check" here. 
The double checked pattern is used to avoid obtaining the lock every time the code is executed.

See more here: https://www.baeldung.com/java-singleton-double-checked-locking
Let me know your thoughts. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: @shahzaibj, Please suggest why we should use the variable starting with s here.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Hi @shahzaibj,
As we are now using the singleton pattern so I feel, having it non-static will give us advantage that this variable will be initialized only when the class CodeMarkerManager.java is used and object is created otherwise not. 

For your suggestion of using "final", I agree, I'm just pushing in the next iteration.

Please share your thoughts. Thanks.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
neeravagarwalmicrosoft: Hi @shahzaibj ,
Please look at the resolution of this comment. Thanks.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Hi @shahzaibj ,
Please look at the resolution of this comment. Thanks.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Simplified. Thanks for your suggestion.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done.

neeravagarwalmicrosoft: Done

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Added more meaningful and simplified comment.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Simplified. Thanks for your suggestions.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Done

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Hi @shahzaibj, 
I have linked the classes. But markCode method cannot be linked.
Let me know your thoughts please. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Done

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Made this static as well.

common/src/main/java/com/microsoft/identity/common/CodeMarkersConstants.java
neeravagarwalmicrosoft: Done. Thanks for the suggestion @shahzaibj

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,83 @@
...
+    public CodeMarker(final String marker, final long timeInMilliseconds, final String timeStamp, final long threadId) {
+        this.marker = marker;
+        this.timeInMilliseconds = timeInMilliseconds;
+        this.timeStamp = timeStamp;
+        this.threadId = threadId;
+    }
+
+    public String getCSVString() {
+        StringBuilder csvStringBuilder = new StringBuilder();
+        csvStringBuilder.append(this.timeStamp == null ? "NA" : this.timeStamp);
```
shahzaibj, line 62: "NA" can be declared in a constant at the top of file

shahzaibj: same applies for comma separator

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,83 @@
...
+    private String cpuUsed = null;
+    private String cpuTotal = null;
+    private String residentSize = null;
+    private String virtualSize = null;
+    private String  wifiSent = null;
+    private String  wifiRecv = null;
+    private String  wwanSent = null;
+    private String  wwanRecv = null;
+    private String  appSent = null;
+    private String  appRecv = null;
```
shahzaibj, line 48: formatting looks off

common/src/main/java/com/microsoft/identity/common/CodeMarkerConstants.java
```diff
@@ -0,0 +1,20 @@
...
+    public static String BROKER_PROCESS_END = "10120";
+    public static String ACQUIRE_TOKEN_SILENT_START = "10011";
+    public static String ACQUIRE_TOKEN_SILENT_EXECUTOR_START = "10012";
+    public static String ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_START = "10013";
+    public static String ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_END = "10014";
+    public static String ACQUIRE_TOKEN_SILENT_FUTURE_OBJECT_CREATION_END = "10020";
+
+    // Scenario codes:
+    public static String SCENARIO_NON_BROKERED_ACQUIRE_TOKEN_SILENTLY = "100";
+    public static String SCENARIO_BROKERED_ACQUIRE_TOKEN_SILENTLY = "200";
```
shahzaibj, line 19: Perhaps there could be two inner classes here such as `Marker` and `Scenario`

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,97 @@
...
+    public void clearMarkers(){
+        this.codeMarkers.clear();
+    }
+
+    public void clearAll() {
+        this.codeMarkers.clear();
+        this.scenarioCode = null;
+    }
+
+    public String getFileContent() {
```
shahzaibj, line 94: need javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * A utility which collects any event's information and also provides functionality to retrieve the events in csv format.
+ */
+public class CodeMarkerManager {
```
shahzaibj, line 33: need javadoc for all non-private methods

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.identity.common;
+
+import java.util.List;
+
+public class CodeMarkerUtil {
```
shahzaibj, line 5: need javadoc

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.identity.common;
```
shahzaibj, line 1: need license

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.List;
+
+public class CodeMarkerUtil {
+    public static String getCSVContent(List<CodeMarker> codeMarkers) {
+        StringBuilder stringToWrite = new StringBuilder("TimeStamp,Marker,Time,Thread,CpuUsed,CpuTotal,ResidentSize,VirtualSize,WifiSent,WifiRecv,WwanSent,WwanRecv,AppSent,AppRecv,Battery,SystemDiskRead,SystemDiskWrite");
```
shahzaibj, line 7: CodeMarker needs to create this...and be aware of the order...and somehow couple this with generating the csv content

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done thanks

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
neeravagarwalmicrosoft: Done thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerConstants.java
neeravagarwalmicrosoft: Done thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
neeravagarwalmicrosoft: Done. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
neeravagarwalmicrosoft: Done. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
neeravagarwalmicrosoft: Done. Thanks

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
neeravagarwalmicrosoft: Done. Thanks

changelog.txt
```diff
@@ -104,6 +104,7 @@ Version 3.0.8
 Version 3.0.7
 ----------
 - Fix exception casting issue in CommandDispatcher (#1121)
+- Adding codemarkers for usage in the Test cases of brokered and non-brokered Acquire Token silently scenarios.
```
shahzaibj, line 107: We need to remove this line from here and move it to the `vNext` section

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class CodeMarkerUtil {
+    /**
+     * This method converts list of codemarkers to csv content which can be written to a file.
+     *
+     * @param codeMarkers
+     * @return
+     */
+    public static String getCSVContent(List<CodeMarker> codeMarkers) {
+        if (codeMarkers == null) return "";
```
shahzaibj, line 38: not following the coding style of our repos

common/src/main/java/com/microsoft/identity/common/PerfConstants.java
```diff
@@ -0,0 +1,24 @@
+package com.microsoft.identity.common;
```
shahzaibj, line 1: missing license

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -126,7 +130,13 @@ public BrokerOperationExecutor(final @NonNull List<IIpcStrategy> strategies) {
         final List<BrokerCommunicationException> communicationExceptionStack = new ArrayList<>();
         for (final IIpcStrategy strategy : mStrategies) {
             try {
+                Logger.warn(TAG, "marking broker start");
```
shahzaibj, line 133: we probably should only log if perf capture is enabled

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/logging/formatter/SimpleTextFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.ui.automation.logging.formatter;
+
+import android.util.Log;
+
+import com.microsoft.identity.client.ui.automation.logging.LogLevel;
+
+/**
+ * A log formatter that returns a string as it is without adding any additional logging parameters.
```
shahzaibj, line 30:
```suggestion
 * A log formatter that returns a log message as it is without adding any additional logging parameters.
```

shahzaibj (changes requested):
Left some comments

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common;
+
+import android.util.Pair;
```
iambmelt, line 25: Can we eliminate the `android` package dependency and implement our own `Tuple` class instead? This keeps this code more portable to Linux/non-Android platforms

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+        this.marker = marker;
+        this.timeInMilliseconds = timeInMilliseconds;
+        this.timeStamp = timeStamp;
+        this.threadId = threadId;
+    }
+
+    // Returns list of pairs of (key,value) where key is the representation of heading in csv and value is the measurable value.
+    private List<Pair<String,String>> getKeyValuePairsOfCodeMarker() {
+        List<Pair<String,String>> csvKeyValuePairs = new ArrayList<Pair<String,String>>();
+        csvKeyValuePairs.add(new Pair("TimeStamp", this.timeStamp == null ? CodeMarker.csvNoValue : this.timeStamp));
```
iambmelt, line 70: Can we refactor these Strings into constants?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
AdamBJohnsonx: I'm telling you, Map.Entry ships with every jvm.

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public CodeMarker(final String marker, final long timeInMilliseconds, final String timeStamp, final long threadId) {
+        this.marker = marker;
+        this.timeInMilliseconds = timeInMilliseconds;
+        this.timeStamp = timeStamp;
+        this.threadId = threadId;
+    }
+
+    // Returns list of pairs of (key,value) where key is the representation of heading in csv and value is the measurable value.
+    private List<Pair<String,String>> getKeyValuePairsOfCodeMarker() {
```
AdamBJohnsonx, line 68: Would it be *so* bad here to have this be a map?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+    private static char csvSeparator = ',';
+
+    private String marker;
+    /* timeInMilliseconds represents time in milliseconds from the time of creation of first codemarker.
+    If timeInMilliseconds is zero(0) then it means that this is the first codemarker of the scenario.*/
+    private long timeInMilliseconds;
+    // timeStamp is the system time at the time of the capture of the codemarker.
+    private String timeStamp;
+    private long threadId;
+    private String cpuUsed = null;
```
iambmelt, line 46: I don't have strong feelings about this, but as a thought: if we keep member variables prefixed with `m` we can avoid having to disambiguate variables with `this`... your call

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+        csvKeyValuePairs.add(new Pair("WwanRecv", this.wwanRecv == null ? CodeMarker.csvNoValue : this.wwanRecv));
+        csvKeyValuePairs.add(new Pair("AppSent", this.appSent == null ? CodeMarker.csvNoValue : this.appSent));
+        csvKeyValuePairs.add(new Pair("AppRecv", this.appRecv == null ? CodeMarker.csvNoValue : this.appRecv));
+        csvKeyValuePairs.add(new Pair("Battery", this.battery == null ? CodeMarker.csvNoValue : this.battery));
+        csvKeyValuePairs.add(new Pair("SystemDiskRead", this.systemDiskRead == null ? CodeMarker.csvNoValue : this.systemDiskRead));
+        csvKeyValuePairs.add(new Pair("SystemDiskWrite", this.systemDiskWrite == null ? CodeMarker.csvNoValue : this.systemDiskWrite));
+        return csvKeyValuePairs;
+    }
+
+    public String getCSVHeader() {
```
iambmelt, line 90: This file looks like it needs a formatter run on it... some of the spacing is inconsistent (`ctrl` + `alt` +`l`)

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+        StringBuilder csvStringBuilder = new StringBuilder();
```
iambmelt, line 91:
```suggestion
        final StringBuilder csvStringBuilder = new StringBuilder();
```

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+        List<Pair<String,String>> csvKeyValuePairs = getKeyValuePairsOfCodeMarker();
+        for(int i = 0; i < csvKeyValuePairs.size(); i++) {
```
iambmelt, line 93: As a general codestyle (and this is somewhat subjective), I don't favor single letter variables: even for loop indexes. I prefer using `ii` and `jj` as you can `ctrl` + `f` find them in a file, unlike a single letter variable. Just a thought 😉

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
iambmelt, line 90: Re: [codestyle](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) this should be `getCsvHeader()`, same for `getCsvLine()`

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,114 @@
...
+            if(i != 0) {
+                csvStringBuilder.append(CodeMarker.csvSeparator);
+            }
+            csvStringBuilder.append(csvKeyValuePairs.get(i).first);
+        }
+        return csvStringBuilder.toString();
+    }
+
+    public String getCSVLine() {
```
iambmelt, line 102: Javadoc for public methods plz

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
iambmelt, line 90: Javadoc for public methods plz

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,111 @@
...
+    // MAX_SIZE_CODE_MARKER is the maximum number of markers this utility can have.
+    private static final int MAX_SIZE_CODE_MARKER = 1000;
+    private volatile List<CodeMarker> codeMarkers = new ArrayList<>();
+
+    //baseMilliSeconds is the time in milliseconds when first codemarker was captured.
+    private long baseMilliSeconds = 0;
+    private String scenarioCode = null;
+    private static CodeMarkerManager sCodeMarkerManager = new CodeMarkerManager();
+
+    public static CodeMarkerManager getInstance() {
```
iambmelt, line 45: As a minor optimization, we can use a [lazy singleton](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom)

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,111 @@
...
+    private CodeMarkerManager() {
+    }
+
+    /**
+     * This method sets a scenarioCode Defined in {@link PerfConstants.ScenarioConstants}.
+     * This scenario code will be pre-fixed to every code marker.
+     * @param scenarioCode
+     */
+    public void setPrefixScenarioCode(String scenarioCode) {
+        this.scenarioCode = scenarioCode;
```
iambmelt, line 58: Same comment here about var naming, if we use a member prefix, we can avoid sprinkling `this` about the code

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * This method captures a particular marker and records the timestamp on which this has been received.
+     * @param marker : A string code which represents a particular place in code.
+     */
+    public void markCode(String marker) {
+        if(this.enableCodeMarker) {
```
iambmelt, line 66: Same comment here about needing a formatter run, declaring vars `final`, and avoiding `this` if we don't need it

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,111 @@
...
+            if(this.codeMarkers.size() >= CodeMarkerManager.MAX_SIZE_CODE_MARKER) {
+                clearMarkers();
+            }
+            long currentMilliSeconds = System.currentTimeMillis();
+            if (this.codeMarkers.size() == 0) {
+                this.baseMilliSeconds = currentMilliSeconds;
+            }
+            SimpleDateFormat f = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
+            this.codeMarkers.add(new CodeMarker((this.scenarioCode == null ? "" : this.scenarioCode) + marker, currentMilliSeconds - this.baseMilliSeconds, f.format(new Date()), Thread.currentThread().getId()));
```
iambmelt, line 75: Let's break this line down to reduce the length (width) and to just generally have it doing less...

In one line we're:
- Constructing an object
- Subtracting time intervals
- Checking `null`
- Formatting strings
- Constructing date objects
- Getting a thread id

If we could split this up into 3, 4 lines this would be much easier to step through in a debugger

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    /**
+     * This method enables or disables the {@link CodeMarkerManager} as per the argument passed to this.
+     * Only enabled {@link CodeMarkerManager} will be able to capture the codemarkers.
+     * @param enableCodeMarker
+     */
+    public void setEnableCodeMarker(boolean enableCodeMarker) {
```
iambmelt, line 84:
```suggestion
    public void setEnableCodeMarker(final boolean enableCodeMarker) {
```

here and everywhere....

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
iambmelt: Yep, we basically following the interaction of these two

- https://google.github.io/styleguide/javaguide.html
- https://source.android.com/setup/contribute/code-style

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+        StringBuilder stringToWrite = new StringBuilder();
+        if (codeMarkers.size() > 0) {
+            stringToWrite.append(codeMarkers.get(0).getCSVHeader());
+        }
+
+        for (CodeMarker codeMarker : codeMarkers) {
```
iambmelt, line 45:
```suggestion
        for (final CodeMarker codeMarker : codeMarkers) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -189,6 +191,7 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
     public static FinalizableResultFuture<CommandResult> submitSilentReturningFuture(@SuppressWarnings(WarningType.rawtype_warning)
                                                                                          @NonNull final BaseCommand command) {
 
+        CodeMarkerManager.getInstance().markCode(PerfConstants.CodeMarkerConstants.ACQUIRE_TOKEN_SILENT_START);
```
iambmelt, line 194: Rather than calling `getInstance()` at different spots inside the code, would it make sense for this class to have a single reference to it? Or, at least for a method to declare it / access it once?

I would say that if we're in a lexical that allows us to declare it once, that's preferable that calling the getter multiple times

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -256,7 +260,9 @@ public void run() {
                         // commandResult = sCommandResultCache.get(command);
                         //If nothing in cache, execute the command and cache the result
 //                        if (commandResult == null) {
-                            commandResult = executeCommand(command);
+                        CodeMarkerManager.getInstance().markCode(PerfConstants.CodeMarkerConstants.ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_START);
```
iambmelt, line 263: I generally think that `import static` should be used judiciously... this might be a decent place for it though...

```java
// Declare once...
final CodeMarkerManager codeMarker = CodeMarkerManager.getInstance();

codeMarker.markCode(ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_START);
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -256,7 +260,9 @@ public void run() {
...
+                        commandResult = executeCommand(command);
+                        CodeMarkerManager.getInstance().markCode(PerfConstants.CodeMarkerConstants.ACQUIRE_TOKEN_SILENT_COMMAND_EXECUTION_END);
```
iambmelt, line 265: 🐛 What happens if the command throws an exception? You'll never hit this perf marker

iambmelt (changes requested):
Posted some comments/thoughts - thanks!

common/build.gradle
```diff
@@ -140,9 +140,9 @@ dependencies {
     //Java 8 - Desugaring - Enabled/Disabled via plugin
     coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:$rootProject.ext.coreLibraryDesugaringVersion"
 
-    localApi(project(":common4j")) {
-        transitive = false
-    }
+    // localApi(project(":common4j")) {
```
iambmelt, line 143: ?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,129 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common;
+
+import java.util.*;
```
iambmelt, line 25: Please use the import optimizer instead of `*` 

(it's `ctrl` + `alt` + `o` in Android Studio)

iambmelt:
@paulkagiri As a collaboration tip; feel free to 'Resolve' feedback if you have commits that address it

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -69,30 +71,34 @@
...
-        @Nullable String getTelemetryApiId();
+        @Nullable
+        String getTelemetryApiId();
 
         /**
          * A method that will be invoked before the success event is emitted.
          * If the calling operation wants to put any value in the success event, put it here.
          */
-        void putValueInSuccessEvent(final @NonNull ApiEndEvent event,final @NonNull T result);
+        void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonNull T result);
```
iambmelt, line 101:
```suggestion
        void putValueInSuccessEvent(@NonNull final ApiEndEvent event, @NonNull final T result);
```

(annots before reserved words - here and everywhere)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -81,11 +83,11 @@
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.REQUEST_CODE;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.RESULT_CODE;
 import static com.microsoft.identity.common.internal.eststelemetry.EstsTelemetry.createLastRequestTelemetryCacheOnAndroid;
+import static com.microsoft.identity.common.PerfConstants.CodeMarkerConstants.*;
```
iambmelt, line 86: Import optimizer can avoid `*`

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -126,7 +137,21 @@ public BrokerOperationExecutor(final @NonNull List<IIpcStrategy> strategies) {
         final List<BrokerCommunicationException> communicationExceptionStack = new ArrayList<>();
         for (final IIpcStrategy strategy : mStrategies) {
             try {
+                if (codeMarkerIsEnabled) {
```
iambmelt, line 140: Hmmm I don't think having to brace each of these in a bool check is necessarily preferable - could we put a log function on/in the `CodeMarkerManager` itself? /cc @shahzaibj since he had similar feedback

Just thinking out loud:
- Potentially call the logger from the `CodeMarkerManager` itself, when markers are hit
- Conditionally log only if code markers are enabled
- `warn()` seems a bits severe; can it be `info()` or `verbose()`?

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
iambmelt: @paulkagiri You can either implement or decline; if you implement feel free to resolve. If declining, just say so and resolve

iambmelt (changes requested):
Left a bit more feedback and resolved CR items already resolved

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
paulkagiri: I already addressed this @iambmelt -please note there's no usage of `this` in this method - https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/neagarw/perftestcase/common/src/main/java/com/microsoft/identity/common/CodeMarker.java#L69-L88

paulkagiri: this will only be used on the initialization of the fields

common/build.gradle
paulkagiri: This was an erroneous change - was testing something locally

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -69,30 +71,34 @@
...
+        @NonNull
+        BrokerOperationBundle getBundle();
 
         /**
          * Extracts the result object from a bundle returned by an IpcStrategy.
          * If the broker returns an error, this will throw an exception.
          */
-        @NonNull T extractResultBundle(final @Nullable Bundle resultBundle) throws BaseException;
+        @NonNull
+        T extractResultBundle(final @Nullable Bundle resultBundle) throws BaseException;
```
iambmelt, line 82:
```suggestion
        T extractResultBundle(@Nullable final Bundle resultBundle) throws BaseException;
```

iambmelt:
⚠ I am seeing what looks like a db being committed to the repo with these changes - should this be .gitignored? `uiautomationutilities/perf_tool/.vs/slnx.sqlite`

changelog.txt
```diff
@@ -15,6 +15,7 @@ V.Next
 - [PATCH] Rev Nimbus version: 8.2 -> 9.9 (#1346)
 - [PATCH] Disregard pageload errors for the non-primary frame during interactive auth (#1357)
 - [PATCH] Add additional logic to support downlevel devices in KeyStoreAccessor (#1366)
+- [PATCH] Adding codemarkers for usage in the Test cases of brokered and non-brokered Acquire Token silently scenarios. (#1185)
```
iambmelt, line 18:
```suggestion
- [MINOR] Adding codemarkers for usage in the Test cases of brokered and non-brokered Acquire Token silently scenarios. (#1185)
```

Patch? Or Minor? I am thinking this is more like a feature than a fix.

iambmelt (approved):
Approving with a few minor suggestions.

This is good 'first pass' - as we move this forward and build it out, we should be thinking about how we reach for more granularity/visibility (ex: timing of cache events, timing of calls like `getAccounts()`, regularly scheduled pipeline runs, and alerting when changes regress performance)

My feedback in this PR has been resolved /cc @shahzaibj

.gitignore
```diff
@@ -85,3 +85,5 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+.vs/
```
shahzaibj, line 89: new line ending

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,145 @@
...
+        csvKeyValuePairs.put(APP_SENT, appSent == null ? CodeMarker.csvNoValue : appSent);
+        csvKeyValuePairs.put(APP_RECEIVED, appRecv == null ? CodeMarker.csvNoValue : appRecv);
+        csvKeyValuePairs.put(BATTERY, battery == null ? CodeMarker.csvNoValue : battery);
+        csvKeyValuePairs.put(SYSTEM_DISK_READ, systemDiskRead == null ? CodeMarker.csvNoValue : systemDiskRead);
+        csvKeyValuePairs.put(SYSTEM_DISK_WRITE, systemDiskWrite == null ? CodeMarker.csvNoValue : systemDiskWrite);
+        return csvKeyValuePairs;
+    }
+
+    /**
+     * Loop through the headers and have them as a csv separated string
```
shahzaibj, line 108:
```suggestion
     * Loop through the headers and have them as a csv string.
```

common/src/main/java/com/microsoft/identity/common/CodeMarker.java
```diff
@@ -0,0 +1,145 @@
...
+                csvStringBuilder.append(CodeMarker.csvSeparator);
+            }
+            csvStringBuilder.append(key);
+            index++;
+        }
+        return csvStringBuilder.toString();
+    }
+
+    /**
+     * Loop through the values and have them as a csv separated string
```
shahzaibj, line 127:
```suggestion
     * Loop through the values and have them as a csv string.
```

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public static CodeMarkerManager getInstance() {
+        return CodeMarkerHolder.INSTANCE;
+    }
+
+    /**
+     * This method captures a particular marker and records the timestamp on which this has been received.
+     *
+     * @param marker : A string code which represents a particular place in code.
```
shahzaibj, line 60:
```suggestion
     * @param marker A string code which represents a particular place in code.
```

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,134 @@
...
+     * This method sets a scenarioCode Defined in {@link PerfConstants.ScenarioConstants}.
+     * This scenario code will be pre-fixed to every code marker.
+     *
+     * @param scenarioCode
+     */
+    public void setPrefixScenarioCode(final String scenarioCode) {
+        this.scenarioCode = scenarioCode;
+    }
+
+    /**
```
shahzaibj, line 95: Missing description

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public void clearMarkers() {
+        codeMarkers.clear();
+    }
+
+    /**
+     * This method clears all the existing markers as well as the scenario code which might have been set earlier.
+     */
+    public void clearAll() {
+        codeMarkers.clear();
```
shahzaibj, line 124:
```suggestion
        clearMarkers();
```

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
```diff
@@ -0,0 +1,134 @@
...
+        scenarioCode = null;
+    }
+
+    /**
```
shahzaibj, line 128: Missing description

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -255,23 +261,28 @@ public void run() {
...
+                        //                        } else {
+                        //                            Logger.info(
+                        //                                    TAG + methodName,
+                        //                                    "Silent command result returned from cache for correlation id : "
+                        //                                            + correlationId + " having status : " + commandResult.getStatus().getLogStatus()
+                        //                            );
+                        //                            // Added to keep the original correlation id intact, and to not let it mutate with the cascading requests hitting the cache.
+                        //                            commandResult = new CommandResult(commandResult.getStatus(),
+                        //                                    commandResult.getResult(), commandResult.getCorrelationId());
+                        //                        }
```
shahzaibj, line 285: Do we need this commented code?

shahzaibj (approved):
Approving with comments

shoatman:
This change looks like it's needs dogfooding to verify.

common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java
paulkagiri: It's part of the @return

paulkagiri:
> This change looks like it's needs dogfooding to verify.

Do you mean like running the pipeline and getting the perf email reports or what do you have in mind @shoatman?

PR #1187, opened 2021-01-07
Add requested claims to access tokens

logdog16:
Implementation of this PR: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1685

I specifically am only making my modification to expose requested_claims as a parameter the underlying AbstractAccountCredentialCache::getCredentialsFilteredBy() and not MsalOAuth2TokenCache::load(), or anywhere in MsalCppOAuth2TokenCache as MSAL C++ directly calls getCredentialsFilteredBy() in our AT lookups, and this (as best I can tell) prevents additional work on the Android team to modify all the places load() is called.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,8 @@
 vNext
 ----------
-- Add requested_claims to access tokens, and allow credentials to be filtered by RC
+
+- Add requested_claims to access tokens, and allow credentials to be filtered by RC (#1187)
+- Sends CP version to ESTS and handle WebCP uri. (#1137)
 - Check for eligible for caching when putting command in executing command map
 - Expose IAccountCredentialCache for accessing lower-level cache functions.
 - Adds unit test to verify .trim() behavior of cache keys.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- Add requested_claims to access tokens, and allow credentials to be filtered by RC
 - Check for eligible for caching when putting command in executing command map
 - Expose IAccountCredentialCache for accessing lower-level cache functions.
 - Adds unit test to verify .trim() behavior of cache keys.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- Check for eligible for caching when putting command in executing command map
 - Expose IAccountCredentialCache for accessing lower-level cache functions.
 - Adds unit test to verify .trim() behavior of cache keys.
 - Make CacheRecord immutable, insist on NonNull AccountRecord (#1225).
@@ -8,6 +9,7 @@ vNext
 - Log an informative error message when application redirect_uri does not match the broker's expected value (#1155).
 - Remove connection close from http request to AUTHORIZATION_CODE_NOT_EXCHANGED_FOR_TOKEN issue on emulators.
 - Replace deprecated HttpRequest.sendGet usage with HttpClient.get (#1038)
+- Replace deprecated HttpRequest.sendPost usage with HttpClient.post (#1037)
 
 Version 3.0.8
 ----------
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
@@ -215,9 +215,7 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
 
             if (mustMatchOnRequestedClaims && credential instanceof AccessTokenRecord) {
                 final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                // Reversed order of equalsIgnoreCaseTrim parameters is intentional due to potential null value
-                // from access token, which can cause an exception to be thrown.
-                matches = matches && equalsIgnoreCaseTrim(accessToken.getRequestedClaims(), requestedClaims);
+                matches = matches && equalsIgnoreCaseTrim(requestedClaims, accessToken.getRequestedClaims());
             }
 
             if (matches) {
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
@@ -213,9 +213,13 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
                 matches = matches && authScheme.equalsIgnoreCase(atType);
             }
 
-            if (mustMatchOnRequestedClaims && credential instanceof AccessTokenRecord) {
-                final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                matches = matches && equalsIgnoreCaseTrim(requestedClaims, accessToken.getRequestedClaims());
+            if (mustMatchOnRequestedClaims) {
+                if (credential instanceof AccessTokenRecord) {
+                    final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
+                    matches = matches && equalsIgnoreCaseTrim(requestedClaims, accessToken.getRequestedClaims());
+                } else {
+                    Logger.verbose(TAG, "Query specified requested_claims-match, but attempted to match with non-AT credential type.");
+                }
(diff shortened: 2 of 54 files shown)


common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -230,6 +230,6 @@ public static boolean equalsIgnoreCase(@Nullable final String one, @Nullable fin
      * @return
      */
     public static boolean equalsIgnoreCaseTrim(@Nullable final String one, @Nullable final String two) {
-        return one == two || equalsIgnoreCase(one, two.trim());
+        return one == two || (two != null && equalsIgnoreCase(one, two.trim()));
```
logdog16, line 233: Change to StringUtils made after consulting with @tanmaymanolkar1 (thanks for the fix!)

Previously a null in parameter two would cause an exception to be thrown when two.trim() was called.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 vNext
 ----------
+- Add requested_claims to access tokens, and allow credentials to be filtered by RC
```
iambmelt, line 3:
```suggestion
- Add requested_claims to access tokens, and allow credentials to be filtered by RC (#1187)
```

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java
```diff
@@ -175,6 +177,14 @@ public String generateCacheKey(Credential credential) {
                 cacheKey = cacheKey.replace(AUTH_SCHEME, sanitizeNull(accessToken.getAccessTokenType()));
             }
 
+            if(!StringExtensions.isNullOrBlank(accessToken.getRequestedClaims()))
```
iambmelt, line 180: Minor formatting nit: can we move the curly-brace to previous line?

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java
```diff
@@ -175,6 +177,14 @@ public String generateCacheKey(Credential credential) {
...
+            {
+                // The Requested Claims string has no guarantee it doesn't contain a delimiter, so we hash it
+                cacheKey += CACHE_VALUE_SEPARATOR + REQUESTED_CLAIMS;
+                String reqClaimsHash = String.valueOf(sanitizeNull(accessToken.getRequestedClaims()).hashCode());
```
iambmelt, line 184: Can't recall off the top of my head: do we need a specific hash alg here? I would think we don't want to rely on Java's built-in `String.hashCode()` and instead want something more x-plat friendly?

common/src/main/java/com/microsoft/identity/common/internal/cache/IAccountCredentialCache.java
```diff
@@ -109,7 +110,8 @@
             final String clientId,
             final String realm,
             final String target,
-            final String authScheme
+            final String authScheme,
+            final String requestedClaims
```
iambmelt, line 114: Hmmm we'll likely want to make this an additive API change... we have a few different consumers of common, so it might make sense to introduce a new overload that is called with `null` as a param by the previous (older) version of the method

iambmelt:
@logdog16 Dropped in a couple of thoughts, but I'll need to go through this PR again (with fresher eyes, it's Friday evening and my brain's a bit fried 😆). At a glance, I think we'll want to make the `getCredentialsFilteredBy()` method an API additive change (such that we don't need to rev `AndroidCommon`'s major version number). Let me think about any other changes we might need to support this work

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -157,6 +169,24 @@ public void setKid(@Nullable final String kid) {
...
+    /**
+     * Gets the requested_claims hash.
+     * @return The hashed copy of the claims string.
+     */
+    public String getRequestedClaims() {
+        return mRequestedClaims;
+    }
+
+    /**
+     * Hashes the incoming string (we don't know the length or potential special characters)
```
iambmelt, line 181: I am unclear about the contents of this javadoc -- this being setter, I wouldn't expect any hashing to occur here, nor for it to me mentioned unless this value is expected to be hash prior to being set (in which case this is a statement about the input data)

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -81,8 +82,19 @@
...
         public static final String KID = "kid";
+
+        /**
+         * A hash of claims string sent to server that produced this AT.
+         */
+        public static final String REQUESTED_CLAIMS = "requested_claims";
     }
 
+    /**
+     * The JSON claims string sent to the server that produced this token. Used by MSAL C++
```
iambmelt, line 93: Unclear to me from looking at this class whether or not this is a hashed value (or if it should be)

common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java
```diff
@@ -209,6 +213,11 @@
                 matches = matches && authScheme.equalsIgnoreCase(atType);
             }
 
+            if (mustMatchOnRequestedClaims && credential instanceof AccessTokenRecord) {
```
iambmelt, line 216: (wishlist, not strictly reqd) Can we (similar to above with `mustMatchOnTarget`) log a message if the caller tries to filter on `requestedClaims` for an a credential type this can't apply to?

For instance, if the developer tries to match `target` to an IdToken, they will get the message: `Logger.verbose(TAG, "Query specified target-match, but no target to match.");`. As a "nice to have", a similar message could be useful for non-AT credential types

logdog16: Added in latest commit!

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
logdog16: Fixed the comment. There should be no hashing in AccessTokenRecord.java

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
logdog16: Thanks for pointing this out. I wrote the comment in an early iteration when I was hashing in the setter (I since realized the error in my ways!) 

Fixed in latest commit

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java
logdog16: Fixed in latest commit

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java
logdog16: IMO, HashCode should be fine here. The purpose of adding the claims hash to the cacheKey is to allow 2 "identical" AT's with different claims strings to coexist in the cache. There's no need to be x-plat (Windows and Apple cache implementations both use plat-specific hash algos), so as long as there's a reasonably low chance of collisions (and acceptable perf) I don't think we need to do anything more than HashCode here.

common/src/main/java/com/microsoft/identity/common/internal/cache/IAccountCredentialCache.java
logdog16: Please take a look at what I did in the latest commit, I think this should address your concerns :)

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java
iambmelt: ACK

iambmelt (approved):

PR #1188, opened 2021-01-11
clear cookies in web view

ayushmeghwani:


shahzaibj:
Why?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -198,6 +198,10 @@ private void setUpWebView(@NonNull final View view,
         mWebView.getSettings().setJavaScriptEnabled(true);
         mWebView.requestFocus(View.FOCUS_DOWN);
 
+        // clear cookies from web view session before processing new login request
+        WebViewUtil.removeCookiesFromWebView(getContext());
+        WebViewUtil.removeSessionCookiesFromWebView(getContext());
```
iambmelt, line 203: The documentation for both of these methods state that they [cannot be called from the UI thread](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/WebViewUtil.java#L87)

iambmelt: (In general though, I do not think we should take this PR -- see my other comment)

iambmelt (changes requested):
I am unclear about the purpose/intent of this change; at a glance, I don't believe this is the correct behavior:
- The cookie jar is scoped to _all_ of this applications `WebViews`, not just the instance used for authentication. As a result, performing interactive auth will eject any non-msft cookies set on the `WebView` which will lead to bugs/loss of session state for any sites visited by the user that may be unrelated to MSFT auth

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
shahzaibj: +1 - I also don't know why are making this change. I think we want the cookies to be there. @ayushmeghwani can you explain the rationale behind this change?

iambmelt:
Closing due to inactivity. Do not presently need this change

PR #1189, opened 2021-01-13
Fix a bug in the javascript injection

jbzdarkid:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -136,10 +136,17 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
 
-                        // Inject string from test suites.
-                        // if result already sent don't load url again
-                        if (mAuthResultSent == false && !StringExtensions.isNullOrBlank(mPostPageLoadedUrl)) {
-                            mWebView.loadUrl(mPostPageLoadedUrl);
+                        // Inject the javascript string from testing. This should only be evaluated if we haven't sent
+                        // an auth result already.
+                        if (mAuthResultSent == false && !StringExtensions.isNullOrBlank(mPostPageLoadedJavascript)) {
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+                                mWebView.evaluateJavascript(mPostPageLoadedJavascript, null);
```
jbzdarkid, line 143: This change will cause a build break in MSAL C++, I will fix that when I update which version we're pointing at.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -136,10 +136,17 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+                            } else {
+                                // On earlier versions of Android, javascript has to be loaded with a custom scheme.
```
AdamBJohnsonx, line 145: Does the javascript scheme no longer work?

jbzdarkid: It does work, but it has some interesting properties, which are described in the comment. I prefer to not escape data if it's not necessary.

iambmelt (approved):
Reviewed changes offline, thanks @jbzdarkid

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
AdamBJohnsonx: It's correct from a URL point of view.  Whatever you put in a URL should be URL-safe, so the parser *should* decode it before handing it to the js engine.  I was just curious how this came up and why we didn't notice it until now.

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
jbzdarkid: It came up because one of our redirect URIs had a percent-encoded octet in it, and it was being passed into the js (as part of automation).

PR #1190, opened 2021-01-14
Add refresh_on to access tokens

logdog16:
Implementing the refresh_on optional field per the schema: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=%2FSSO%2FSchema.md

Documentation of the refresh_in header can be found here: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/RefreshAtExpirationPercentage/overview.md&_a=preview

As a note, just as expires_on is a calculated value based on the server value (expires_in) + local time, refresh_on works the same way, being returned from the server as refresh_in, and requiring a calculation to be made by the library to save a unix timestamp.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
@@ -41,6 +41,7 @@ import static com.microsoft.identity.common.internal.dto.AccessTokenRecord.Seria
 import static com.microsoft.identity.common.internal.dto.AccessTokenRecord.SerializedNames.REALM;
 import static com.microsoft.identity.common.internal.dto.AccessTokenRecord.SerializedNames.TARGET;
 import static com.microsoft.identity.common.internal.dto.AccessTokenRecord.SerializedNames.TOKEN_TYPE;
+import static com.microsoft.identity.common.internal.dto.AccessTokenRecord.SerializedNames.REFRESH_ON;
 import static com.microsoft.identity.common.internal.dto.Credential.SerializedNames.EXPIRES_ON;
 
 @EqualsAndHashCode(callSuper = true)
@@ -87,6 +88,11 @@ public class AccessTokenRecord extends Credential {
          * The claims string (if present) that was sent to server to produce this AT.
          */
         public static final String REQUESTED_CLAIMS = "requested_claims";
+
+        /**
+         * String of refresh_in.
+         */
+        public static final String REFRESH_ON = "refresh_on";
     }
 
     /**
@@ -145,6 +151,15 @@ public class AccessTokenRecord extends Credential {
      */
     @SerializedName(EXPIRES_ON)
     private String mExpiresOn;
+
+    /**
+     * Token recommended refresh time. This value should be calculated based on the current UTC time
+     * measured locally and the value refresh_in returned from the service. Measured in milliseconds from
+     * epoch (1970). Note that this value will not always be present, and is only a recommendation to
+     * kick off an async token refresh. The token is still valid until the expires_on value.
+     */
+    @SerializedName(REFRESH_ON)
+    private String mRefreshOn;
 
     /**
      * Gets the kid.
@@ -293,6 +308,24 @@ public class AccessTokenRecord extends Credential {
      */
     public void setExpiresOn(final String expiresOn) {
         mExpiresOn = expiresOn;
+    }
+
+    /**
+     * Gets the refresh_on timestamp.
+     *
+     * @return The refresh_on to get.
+     */
+    public String getRefreshOn() {
+        return mRefreshOn;
+    }
+
+    /**
+     * Sets the refresh_on timestamp.
+     *
+     * @param refreshOn The refresh_on to set.
+     */
+    public void setRefreshOn(final String refreshOn) {
+        mRefreshOn = refreshOn;
     }
 
     private boolean isExpired(final String expires) {
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
@@ -361,6 +361,9 @@ public class AccessTokenRecord extends Credential {
         if (refreshOn != null && !refreshOn.isEmpty()) {
             return isExpired(refreshOn);
         }
-        return isExpired();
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -295,6 +310,24 @@ public void setExpiresOn(final String expiresOn) {
...
+    /**
+     * Sets the refresh_on timestamp.
+     *
+     * @param refreshOn The refresh_on to set.
+     */
+    public void setRefreshOn(final String refreshOn) {
+        mRefreshOn = refreshOn;
+    }
+
     private boolean isExpired(final String expires) {
```
logdog16, line 331: Please let me know if you'd like me to add a shouldRefresh function, based on isExpired.

C++ doesn't use the isExpired logic (we've got our own logic in a higher, x-plat layer), but if Android plans to support refresh_in anytime soon, I'm happy to lay the groundwork as part of this PR :)

iambmelt: @logdog16 If you can add tests for it, I wouldn't decline having this API. I'm assuming it's 1 additional method?

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -344,4 +344,23 @@ private boolean isExpired(final String expires) {
...
+     *
+     * @return Should the library kick off a background token refresh operation for this token
+     * after returning the token to the caller.
+     */
+    public boolean shouldRefresh() {
+        final String refreshOn = getRefreshOn();
+        if(refreshOn != null && !refreshOn.isEmpty()) {
+            return isExpired(refreshOn);
+        }
+        return false;
```
iambmelt, line 364: It seems like it might potentially be safer (from a calling/understanding POV) to call `return isExpired()` instead of `return false` - thoughts?

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -344,4 +344,23 @@ private boolean isExpired(final String expires) {
...
+     *
+     * If the AT does NOT have a refresh_on timestamp, this function will always
+     * return false. Fallback to standard token expiration/refresh logic.
+     *
+     * @return Should the library kick off a background token refresh operation for this token
+     * after returning the token to the caller.
+     */
+    public boolean shouldRefresh() {
+        final String refreshOn = getRefreshOn();
+        if(refreshOn != null && !refreshOn.isEmpty()) {
```
iambmelt, line 361:
```suggestion
        if (refreshOn != null && !refreshOn.isEmpty()) {
```

iambmelt: (Just minor formatting)

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
logdog16: The way I'm writing the logic in C++ works like this:

1. Get AT from cache
2. Check if AT is expired (if yes, break here and go straight to a synchronous RT renewal to get a valid AT for the app)
3. Check if AT should be refreshed (if yes, create an async RT renewal background request, and add it to our BG task queue)
4. Regardless of step 3, return the current AT to the app

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
```diff
@@ -311,4 +344,23 @@ private boolean isExpired(final String expires) {
...
+     *
+     * @return Should the library kick off a background token refresh operation for this token
+     * after returning the token to the caller.
+     */
+    public boolean shouldRefresh() {
+        final String refreshOn = getRefreshOn();
+        if (refreshOn != null && !refreshOn.isEmpty()) {
+            return isExpired(refreshOn);
+        }
+        return false;
```
logdog16, line 364:
```suggestion
        return isExpired();
```

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
logdog16: In either case, I have no horse in this race! Feel free to commit the suggestion I left in a separate comment if you think that's the better approach!

iambmelt: Just seeing this now... let me ponder this a bit and I will plan to finalize your CR tomorrow. Thank you @logdog16!

common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java
iambmelt: I'll take this....

common/src/test/java/com/microsoft/identity/common/AccessTokenTest.java
```diff
@@ -24,19 +24,33 @@
...
+
+    @Test
+    public void testShouldRefreshWhenSet() {
+        final AccessTokenRecord accessToken = new AccessTokenRecord();
+        accessToken.setRefreshOn(getCurrentTimeStr());
+        assertTrue(accessToken.shouldRefresh());
+    }
+
+    @Test
+    public void testShouldRefreshWhenNotSet() {
```
iambmelt, line 51: Is this test named correctly? The naming of the method, paired with `assertFalse` looks inverted

iambmelt: This looks like it may throw a `NullPointerException` with the fallback logic merged in b74482c298bcf0d5f06cd2816d7f3e7ed1422df2 if all properties are uninitialized... I don't expect this should be possible in the real world... may be good to be a little defensive here though if easy/cheap

iambmelt (approved):
Dropped in 2 comments/suggestions in `AccessTokenTest` -- please take a look at those to ensure correctness naming/test expectations. Aside from that, these changes LGTM, thanks @logdog16

PR #1191, opened 2021-01-14
Add logic to allow trusting debug brokers in release builds

shahzaibj:
MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1274
Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1471
ADAL PR: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1578


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -29,14 +29,63 @@
...
+            COMPANY_PORTAL_APP_PACKAGE_NAME,
+            COMPANY_PORTAL_APP_RELEASE_SIGNATURE
+    );
+
+    public static final BrokerData BROKER_HOST = new BrokerData(
+            BROKER_HOST_APP_PACKAGE_NAME,
+            BROKER_HOST_APP_SIGNATURE
+    );
+
+    private static final Set<BrokerData> debugBrokers = new HashSet<BrokerData>() {{
```
AdamBJohnsonx, line 68: If this is a constant, maybe DEBUG_BROKERS, and make it immutable by wrapping it in a Collections.immutableSet

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -29,14 +29,63 @@
...
+        add(MICROSOFT_AUTHENTICATOR_DEBUG);
+        add(BROKER_HOST);
+    }};
+
+    private static final Set<BrokerData> prodBrokers = new HashSet<BrokerData>() {{
+        add(MICROSOFT_AUTHENTICATOR_PROD);
+        add(COMPANY_PORTAL);
+    }};
+
+    private static final Set<BrokerData> allBrokers = new HashSet<BrokerData>() {{
```
AdamBJohnsonx, line 78: ALL_BROKERS and immutable wrapper?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -133,6 +148,14 @@ public boolean verifySignature(final String brokerPackageName) {
         return false;
     }
 
+    public Set<BrokerData> getValidBrokers() {
```
AdamBJohnsonx, line 151: rather than return a list of valid brokerData publicly, could we implement an isBrokerValid(pkgName, signature)?

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
```diff
@@ -55,17 +51,22 @@ private ProcessUtil() {
...
         final String processName = getProcessName(context);
 
-        final String authAppAuthProcess = AZURE_AUTHENTICATOR_APP_PACKAGE_NAME + ":auth";
-        final String cpAuthProcess = COMPANY_PORTAL_APP_PACKAGE_NAME + ":auth";
+        final BrokerValidator brokerValidator = new BrokerValidator(context);
+        final Set<BrokerData> validBrokers = brokerValidator.getValidBrokers();
+
+        for (final BrokerData brokerData : validBrokers) {
+            final String authProcess = brokerData.packageName + ":auth";
+            return authProcess.equalsIgnoreCase(processName);
```
AdamBJohnsonx, line 59: This will only run for the first element in this collection...

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -440,7 +441,8 @@ public EncryptionType getEncryptionType(@NonNull final String data) throws Unsup
 
         if (encryptionType == EncryptionType.USER_DEFINED) {
             if (isBrokerProcess()) {
-                if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
+                if (COMPANY_PORTAL_APP_PACKAGE_NAME.equalsIgnoreCase(packageName) ||
+                        BROKER_HOST_APP_PACKAGE_NAME.equalsIgnoreCase(packageName)) {
```
AdamBJohnsonx, line 445: should we have a static utility method for this?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -60,11 +61,25 @@
...
+import java.util.Set;
 
 public class BrokerValidator {
 
     private static final String TAG = "BrokerValidator";
 
+    private static boolean shouldTrustDebugBrokers = BuildConfig.DEBUG;
+
+    public static void setShouldTrustDebugBrokers(final boolean shouldTrustDebugBrokers) {
+        if (!BuildConfig.DEBUG && shouldTrustDebugBrokers) {
```
AdamBJohnsonx, line 73: shouldTrustDebugBrokers and BuildConfig.DEBUG are identical.  Did you mean to use a different BuildConfig string for that configuration?

AdamBJohnsonx:
Overall, I think we're going to want to have some unit tests here, so that we make sure it's doing what we think it does.

AdamBJohnsonx (commented):
That's the one.  It can get a little clunky to have all those finally blocks lurking around.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -29,14 +29,63 @@
...
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.BROKER_HOST_APP_SIGNATURE;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_RELEASE_SIGNATURE;
+
 /**
  * Represents packageName and SignatureHash of a broker app.
  */
 public class BrokerData {
+
+    public static final BrokerData MICROSOFT_AUTHENTICATOR_DEBUG = new BrokerData(
```
iamgusain, line 49: >MICROSOFT_AUTHENTICATOR_DEBUG [](start = 35, length = 29)

Should these be part of a separate class.. 

public static final class BrokerApps
{
public BrokerData  Authenticator_Release {..};
public BrokerData CompanyPortalRelease {..}
..
..
}

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -57,7 +57,7 @@
 
     private String mBrokerPackageName = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
 
-    private String mBrokerSignature = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_SIGNATURE;
+    private String mBrokerSignature = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_RELEASE_SIGNATURE;
```
iamgusain, line 60: >COMPANY_PORTAL_APP_RELEASE_SIGNATURE [](start = 69, length = 36)

Should this be Debug Signature in case of running in Debug mode?

iamgusain: I see, we use the same signature everywhere.

---
In reply to: [564909272](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1191#discussion_r564909272) [](ancestors = 564909272)

shahzaibj: We don't know what Company Portal debug's signature is and we don't even know how to get a handle on a debug build of CP. In other words, we never expect to be playing with a debug CP. If in the future, we do get access to debug builds of CP then we can find out its signature and include it here.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -29,14 +29,63 @@
...
+            BROKER_HOST_APP_PACKAGE_NAME,
+            BROKER_HOST_APP_SIGNATURE
+    );
+
+    private static final Set<BrokerData> debugBrokers = new HashSet<BrokerData>() {{
+        add(MICROSOFT_AUTHENTICATOR_DEBUG);
+        add(BROKER_HOST);
+    }};
+
+    private static final Set<BrokerData> prodBrokers = new HashSet<BrokerData>() {{
```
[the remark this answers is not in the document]
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/util/ProcessUtil.java
shahzaibj: Addressed

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -141,8 +141,9 @@ public static String getBrokerRedirectUrl(final String packageName, final String
         if (!StringExtensions.isNullOrBlank(packageName)
                 && !StringExtensions.isNullOrBlank(signatureDigest)) {
             // If the caller is the Authenticator, then use the broker redirect URI.
-            if (packageName.equals(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME) &&
-                    signatureDigest.equals(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_SIGNATURE)) {
+            if (packageName.equals(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
+                    && (signatureDigest.equals(BrokerData.MICROSOFT_AUTHENTICATOR_PROD.signatureHash)
+                    || signatureDigest.equals(BrokerData.MICROSOFT_AUTHENTICATOR_DEBUG.signatureHash))) {
```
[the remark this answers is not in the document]
shahzaibj: Since this is specific to just Authenticator, therefore I am not using the complete array here

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
shahzaibj: we could but I would consider it an overkill....besides I don't know what the name that method

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
AdamBJohnsonx: The syntax highlighting got me.  Could we name the method parameter and the field differently?

shahzaibj: No they are not identical....`shouldTrustDebugBrokers` could have any value passed in when this method is called and that would be independent of the value of `BuildConfig.DEBUG`

shahzaibj: Yup done already here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1191/commits/dc80e2672613294a6bc0b3252146938f17f04f36

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
shahzaibj: We could do it if we wanted to....I chose not to do it because it requires creating another class just for this tiny purpose

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
shahzaibj: I've added a method for `isBrokerValid` here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1191/commits/2114def4600f39c7f09390f00bd361a76d13f941 (just in case if someone needs to use it in future)
 
but I've still kept the `getValidBrokers` method because in the places where this is being used we don't know the signature hash....we are basically just using to check if the package name that we have, matches the package name of a valid Broker app.

If we really want to NOT use `getValidBrokers` and rather use `isBrokerValid(pkgName, signature)` then we would have to calculate the signature hash (again) in these places and become aware of it and pass it along to the method.

An alternative would be to create an overload method called as `isBrokerValid(pkgName)` - this would return whether the package name maps to a valid broker app (completely ignores the signature hash aspect of broker validity)

shahzaibj: @AdamBJohnsonx Any thoughts on the above ^^?

AdamBJohnsonx: It's just that it looks like we're mainly getting broker data by package name except for the case where we compute a signature hash and then see if it was the right one for any of the brokers.  I don't want to add unused code, but every time I look at this it seems like we're repeating the same loop over the brokers set.

If nothing is using it, don't include it.

AdamBJohnsonx (approved):
If it works, ship it.

PR #1192, opened 2021-01-14
Validate ATs before writing to MSALCPP cache

iambmelt:
Impetus:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1221073/

In this PR:
- Reenables `MsalOAuth2TokenCache` tests previously disabled due to mysterious failures on Travis-CI infrastructure
- Adds AT validation when saving via `MsalCppOAuth2TokenCache`

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,5 @@
 vNext
 ----------
-
 - Add requested_claims to access tokens, and allow credentials to be filtered by RC (#1187)
 - Sends CP version to ESTS and handle WebCP uri. (#1137)
 - Check for eligible for caching when putting command in executing command map
@@ -13,6 +12,7 @@ vNext
 - Remove connection close from http request to AUTHORIZATION_CODE_NOT_EXCHANGED_FOR_TOKEN issue on emulators.
 - Replace deprecated HttpRequest.sendGet usage with HttpClient.get (#1038)
 - Replace deprecated HttpRequest.sendPost usage with HttpClient.post (#1037)
+- (MSALCPP) Validate ATs before persisting to the cache (#1192)
 
 Version 3.0.8
 ----------
--- a/common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
@@ -314,4 +314,15 @@ public class MsalCppOAuth2TokenCacheTest extends AndroidSecretKeyEnabledHelper {
         Assert.assertTrue(credentials.contains(mTestBundle.mGeneratedIdToken));
         Assert.assertTrue(credentials.contains(mTestBundle.mGeneratedRefreshToken));
     }
+
+    @Test(expected = ClientException.class)
+    public void saveATSansTargetThrowsException() throws ClientException {
+        mTestBundle.mGeneratedAccessToken.setTarget(null);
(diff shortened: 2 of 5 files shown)


AdamBJohnsonx:
Can we just refuse to construct the object if it's not going to be valid?

iambmelt:
> Can we just refuse to construct the object if it's not going to be valid? Or is there some case that we have for Android MSAL that involved Access Tokens without target fields?

@AdamBJohnsonx I reckon this could be done with a similar builder trick to what you used for `CacheRecord`; it might be a somewhat larger change though (versus the changes I've proposed in this PR). So I guess I'll ask... how strongly do you feel about including that as part of this PR? (And accordingly, if we apply that to ATs, we could _go further yet_ to apply it to RTs, IDs, and Accounts so it could increase the scope a fair bit if we want to prioritize the logical consistency of that validation across account & credential types)

AdamBJohnsonx:
> > Can we just refuse to construct the object if it's not going to be valid? Or is there some case that we have for Android MSAL that involved Access Tokens without target fields?
> 
> @AdamBJohnsonx I reckon this could be done with a similar builder trick to what you used for `CacheRecord`; it might be a somewhat larger change though (versus the changes I've proposed in this PR). So I guess I'll ask... how strongly do you feel about including that as part of this PR? (And accordingly, if we apply that to ATs, we could _go further yet_ to apply it to RTs, IDs, and Accounts so it could increase the scope a fair bit if we want to prioritize the logical consistency of that validation across account & credential types)

Yeah, we can make a giant change here, but I was looking at this as more of chipping away at this until we size that down to where we can just bite it off.  I've got a desire to make all objects immutable and always coherent, and I think it would help us when investigating issues, but it's not a thing you can just do in a day.

AdamBJohnsonx (approved):
I take it that the changes from trying to make AccountRecord always valid were too large.  Ship it.  We can always do it later in Common.

iambmelt:
Notes re: suggested Lombok changes

I experimented with these changes and found the changeset getting quite large and starting to snake into lots of files so I backed off
- Because there's an inheritance hierarchy here (`AccountCredentialBase` > `Credential` > `AccessTokenRecord`) and we want to validate properties of both the super/child-class, we'd need to use `@SuperBuilder` when then requires adding `@SuperBuilder` to all this child classes too (or else compilation errors ensue)... so didn't look readily easy to piecemeal make these changes
- Because there are multiple layers of generate Lombok builders (each generated with multiple generic type parameters), the process for writing these classes was a bit error prone. I ended up taking/repeating the following steps: 
    - 1. Add annotation 
    - 2. Delombok the file to get the Builder (because I found writing by hand very error prone with multiple generic params)
    - 3. Override the various builder calls to add validation (like null checks, etc)
- Thinking about the immutability aspect (and maybe I should have started with this in-mind), there's soon to be [support for support](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1221076) merging a new account with existing account data so likely we wouldn't be able to make these objects entirely immutable post-construction
- Something to consider if I/we revisit this suggestion: How does/will Gson behave (and how do we _want_ it to behave) if we attempt to serialize data written by another library, but its lacking fields we've deemed 'required'. Should we fail immediately during construction/serialization? Or later if only a specific field is missing.... just something to consider

iambmelt:
@PetarSDimov If you could review these changes as well

PetarSDimov (approved):
:shipit:

PR #1194, opened 2021-01-16
Authenticator app automation UI fixes

shahzaibj:
- Handle UI for privacy dialog that pops up during shared device registration
- Change logic for setting PIN on the device to handle Authenticator app lock (see javadoc in DevicePinSetupRule to understand why)
  - Only set PIN if CP is being used
  - Remove PIN if Authenticator is being
- Fix for creating powerlift incidents due to UI change in Auth App


kkrishnareddy (approved):
LGTM

PR #1195, opened 2021-01-19
Enabling the MFA setup using Authenticator app

tanmaymanolkar1:
The MFA setup using the webview through Authenticator app was broken. Users had to resort to browser flow to complete the MFA setup as a workaround. This PR aims to fix the mentioned issue.

Related ADAL PR : https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1579

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1079,6 +1079,11 @@ public final class AuthenticationConstants {
         public static final String BROWSER_EXT_WEB_CP = "companyportal://";
 
         /**
+         * Prefix in the redirect from WebCP.
+         */
+        public static final String AUTHENTICATOR_MFA_LINKING = "microsoft-authenticator://activatemfa";
+
+        /**
          * Redirect URL from WebCP that should launch the Intune Company Portal app.
          */
         public static final String WEBCP_LAUNCH_COMPANY_PORTAL_URL = BROWSER_EXT_WEB_CP + "enrollment";
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1170,6 +1170,16 @@ public final class AuthenticationConstants {
         public static final String REDIRECT_SSL_PREFIX = "https://";
 
         /**
+         * Prefix in the redirect for PlayStore.
+         */
+        public static final String PLAY_STORE_INSTALL_PREFIX = "market://details?id=";
+
+        /**
+         * Prefix in the redirect for PlayStore through Browser.
+         */
+        public static final String PLAY_STORE_BROWSER_INSTALL_PREFIX = REDIRECT_SSL_PREFIX + "play.google.com/store/apps/details?id=";
+
+        /**
          * String for expiration buffer.
          * Integer for token expiration buffer. see {@link AuthenticationSettings#mExpirationBuffer}
          */
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1180,11 +1180,6 @@ public final class AuthenticationConstants {
         public static final String PLAY_STORE_INSTALL_PREFIX = "market://details?id=";
 
         /**
-         * Prefix in the redirect for PlayStore through Browser.
-         */
-        public static final String PLAY_STORE_BROWSER_INSTALL_PREFIX = REDIRECT_SSL_PREFIX + "play.google.com/store/apps/details?id=";
-
-        /**
          * String for expiration buffer.
          * Integer for token expiration buffer. see {@link AuthenticationSettings#mExpirationBuffer}
          */
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1079,9 +1079,9 @@ public final class AuthenticationConstants {
         public static final String BROWSER_EXT_WEB_CP = "companyportal://";
 
         /**
-         * Prefix in the redirect from WebCP.
+         * Prefix for the Authenticator MFA linking.
          */
-        public static final String AUTHENTICATOR_MFA_LINKING = "microsoft-authenticator://activatemfa";
+        public static final String AUTHENTICATOR_MFA_LINKING_PREFIX = "microsoft-authenticator://activatemfa";
 
         /**
          * Redirect URL from WebCP that should launch the Intune Company Portal app.
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -335,7 +335,6 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(PLAY_STORE_INSTALL_PREFIX + appPackageName));
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
             getActivity().startActivity(intent);
-            return true;
         } catch (android.content.ActivityNotFoundException e) {
             //if GooglePlay is not present on the device.
             Logger.error(TAG + methodName, "Failed to launch the PlayStore.", e);
@@ -351,7 +350,6 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
             intent.addFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT);
             getActivity().startActivity(intent);
-            return true;
         } catch (android.content.ActivityNotFoundException e) {
             Logger.error(TAG, "Failed to open the Authenticator application.", e);
         }
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -158,34 +158,42 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             return true;
         } else if (isRedirectUrl(formattedURL)) {
             Logger.info(TAG, "Navigation starts with the redirect uri.");
-            return processRedirectUrl(view, url);
+            processRedirectUrl(view, url);
+            return true;
         } else if (isWebsiteRequestUrl(formattedURL)) {
             Logger.info(TAG, "It is an external website request");
-            return processWebsiteRequest(view, url);
+            processWebsiteRequest(view, url);
+            return true;
         } else if (isInstallRequestUrl(formattedURL)) {
             Logger.info(TAG, "It is an install request");
-            return processInstallRequest(view, url);
+            processInstallRequest(view, url);
+            return true;
         } else if (isWebCpUrl(formattedURL)) {
             Logger.info(TAG, "It is a request from WebCP");
-            return processWebCpRequest(view, url);
+            processWebCpRequest(view, url);
+            return true;
         } else if (isPlayStoreUrl(formattedURL)){
             Logger.info(TAG, "Request to open PlayStore.");
             return processPlayStoreURL(view, url);
         } else if (isAuthAppMFAUrl(formattedURL)){
             Logger.info(TAG, "Request to link account with Authenticator.");
-            return processAuthAppMFAUrl(url);
+            processAuthAppMFAUrl(url);
+            return true;
         } else if (isValidRedirectUri(url)){
             Logger.info(TAG, "Check for Redirect Uri.");
-            return processRedirectUriCheck(view, url);
+            processInvalidRedirectUri(view, url);
+            return true;
         } else if (isBlankPageRequest(formattedURL)){
             Logger.verbose(TAG, "It is an blank page request");
             return true;
         } else if (isUriSSLProtected(formattedURL)){
             Logger.verbose(TAG, "Check for SSL protection");
-            return processSSLProtectionCheck(view, url);
+            processSSLProtectionCheck(view, url);
+            return true;
         } else {
-            Logger.info(TAG, "It is an invalid redirect uri.");
-            return processInvalidUrl(url);
+            Logger.info(TAG, "This maybe a valid URI, but no special handling for this mentioned URI, hence deferring to WebView for loading.");
+            processInvalidUrl(url);
+            return false;
         }
     }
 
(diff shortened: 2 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1169,6 +1174,11 @@
          */
         public static final String REDIRECT_SSL_PREFIX = "https://";
 
+        /**
+         * Prefix in the redirect for PlayStore.
+         */
+        public static final String PLAY_STORE_INSTALL_PREFIX = "market://details?id=";
```
AdamBJohnsonx, line 1180: I don't know if we should bank on this URI being this stable.  We could have other query parameters, for sure, that we should probably ignore.  If it was critical, you'd think they'd make it part of the path.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -165,12 +168,48 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            Logger.info(TAG, "Check for Redirect Uri.");
+            return processRedirectUriCheck(view, url);
+        } else if (isBlankPageRequest(formattedURL)){
+            Logger.verbose(TAG, "It is an blank page request");
+            return true;
+        } else if (isUriSSLProtected(formattedURL)){
+            Logger.verbose(TAG, "Check for SSL protection");
+            return processSSLProtectionCheck(view, url);
         } else {
             Logger.info(TAG, "It is an invalid redirect uri.");
```
AdamBJohnsonx, line 187: OK, I hate this message.  Can we make it say something closer to what it means to us: "We have a URI that we know is not unsafe, but we don't have any special handling for.  This is potentially ADFS (or other cases that fall through here, which we should enumerate), deferring to the WebView for loading."

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -280,6 +319,44 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
...
+        final String methodName = "#processPlayStoreURL";
+
+        view.stopLoading();
+        if (!(url.startsWith(PLAY_STORE_INSTALL_PREFIX + COMPANY_PORTAL_PROD_APP_PACKAGE_NAME))
+                && !(url.startsWith(PLAY_STORE_INSTALL_PREFIX + AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))) {
+            return false;
+        }
+        final String appPackageName = (url.contains(COMPANY_PORTAL_PROD_APP_PACKAGE_NAME) ?
+                COMPANY_PORTAL_PROD_APP_PACKAGE_NAME : AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
+        Logger.verbose(TAG + methodName, "Request to open PlayStore.");
```
AdamBJohnsonx, line 332: probably info, and then the package we think we're going to install

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -280,6 +319,44 @@ private boolean processWebsiteRequest(@NonNull final WebView view, @NonNull fina
...
+
+        try {
+            final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(PLAY_STORE_INSTALL_PREFIX + appPackageName));
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+            getActivity().startActivity(intent);
+        } catch (android.content.ActivityNotFoundException e) {
+            //if GooglePlay is not present on the device.
+            Logger.error(TAG + methodName, "Failed to launch the PlayStore.", e);
```
AdamBJohnsonx, line 340: Maybe actually "play store is not present on device"

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -370,32 +447,29 @@ public void run() {
...
-                && url.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX)) {
-            Logger.error(TAG, "The RedirectUri is not as expected.", null);
-            Logger.errorPII(TAG, String.format("Received %s and expected %s", url, mRedirectUrl), null);
-            returnError(ErrorStrings.DEVELOPER_REDIRECTURI_INVALID,
-                    String.format("The RedirectUri is not as expected. Received %s and expected %s", url,
-                            mRedirectUrl));
-            view.stopLoading();
-            return true;
-        }
+    private boolean processRedirectUriCheck(@NonNull final WebView view, @NonNull final String url) {
```
AdamBJohnsonx, line 450: This is a dead-end method, right?  We should name it that way.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -370,32 +447,29 @@ public void run() {
...
+        returnError(ErrorStrings.DEVELOPER_REDIRECTURI_INVALID,
+                String.format("The RedirectUri is not as expected. Received %s and expected %s", url,
+                        mRedirectUrl));
+        view.stopLoading();
+        return true;
+    }
 
-        if (!lowerCaseUrl.startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)) {
-            final String redactedUrl = removeQueryParametersOrRedact(url);
+    private boolean processSSLProtectionCheck(@NonNull final WebView view, @NonNull final String url) {
```
AdamBJohnsonx, line 462: This method only returns one thing.  A lot of these do.  we could pull their return statements into their callers and have them be void.

AdamBJohnsonx:
Can we add some unit tests here?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
tanmaymanolkar1: its not a part of the path as of now, if they do that, other libraries including ours would have to make the changes.
But as of now, since Auth app being the primary choice to handle MFA, we can be certain to get an URI such as this hinting to download authenticator app.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
tanmaymanolkar1: processInvalidUrl(url) is supposed to be the dead-end method.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx: This is more about being humble with respect to query parameters that might appear in that string.  It could change, somebody could decide to add a cache breaker or what-have-you.

As a contrived example:

https://play.google.com/store/apps/details?id=com.zhiliaoapp.musically&foo=bar

This uri works.

We might not want to allow it, but if we do block it, we should say why (we found unknown query paratmeters in the uri).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
AdamBJohnsonx: What I'm saying is that there's no need for it to return.  If you call this method, you know what it's going to return.  Same with all of these, for the most part.  This is a minor comment, but it means that somebody reading the code might think there's the possibility it could return something dependent on its input.  It cannot, and then they have to go read the method to make that determination.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
tanmaymanolkar1: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1195/commits/fa6aa08d511b6cc2d891b8ec8e9e9a13a542de2b

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1078,6 +1078,11 @@
          */
         public static final String BROWSER_EXT_WEB_CP = "companyportal://";
 
+        /**
+         * Prefix in the redirect from WebCP.
```
iamgusain, line 1082: >Prefix in the redirect from WebCP. [](start = 11, length = 34)

This comment matches the above shouldn't it mention the authenticator

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1078,6 +1078,11 @@
...
+         */
+        public static final String AUTHENTICATOR_MFA_LINKING = "microsoft-authenticator://activatemfa";
```
iamgusain, line 1084: >AUTHENTICATOR_MFA_LINKING [](start = 35, length = 25)

I would prefer PREFIX to be added at the end of the constant name AUTHENTICATOR_MFA_LINKING_PREFIX,

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -155,22 +158,66 @@ private boolean handleUrl(final WebView view, final String url) {
...
 
+    private boolean isUriSSLProtected(@NonNull final String url) {
+        return !(url.startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX));
+    }
+
+    private boolean isBlankPageRequest(@NonNull final String url) {
+        return "about:blank".equals(url);
+    }
+
+    private boolean isValidRedirectUri(@NonNull final String url) {
```
iamgusain, line 208: >isValidRedirectUri [](start = 20, length = 18)

if it is a valid redirect uri, why are we processing it as invalidRedirectUri. Did you mean to invert the condition here?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -155,22 +158,66 @@ private boolean handleUrl(final WebView view, final String url) {
...
+        } else if (isAuthAppMFAUrl(formattedURL)){
+            Logger.info(TAG, "Request to link account with Authenticator.");
+            processAuthAppMFAUrl(url);
+            return true;
+        } else if (isValidRedirectUri(url)){
+            Logger.info(TAG, "Check for Redirect Uri.");
+            processInvalidRedirectUri(view, url);
+            return true;
+        } else if (isBlankPageRequest(formattedURL)){
+            Logger.verbose(TAG, "It is an blank page request");
```
iamgusain, line 187: >verbose [](start = 19, length = 7)

info?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -155,22 +158,66 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            return true;
+        } else if (isUriSSLProtected(formattedURL)){
+            Logger.verbose(TAG, "Check for SSL protection");
```
iamgusain, line 190: >verbose [](start = 19, length = 7)

info?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -155,22 +158,66 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            processSSLProtectionCheck(view, url);
+            return true;
         } else {
-            Logger.info(TAG, "It is an invalid redirect uri.");
-            return processInvalidUrl(view, url);
+            Logger.info(TAG, "This maybe a valid URI, but no special handling for this mentioned URI, hence deferring to WebView for loading.");
+            processInvalidUrl(url);
+            return false;
         }
```
iamgusain, line 197: nit: you can also return true here in the end.. and remove it from the else blocks above.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -155,22 +158,66 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            processWebCpRequest(view, url);
+            return true;
+        } else if (isPlayStoreUrl(formattedURL)){
+            Logger.info(TAG, "Request to open PlayStore.");
+            return processPlayStoreURL(view, url);
+        } else if (isAuthAppMFAUrl(formattedURL)){
+            Logger.info(TAG, "Request to link account with Authenticator.");
+            processAuthAppMFAUrl(url);
+            return true;
+        } else if (isValidRedirectUri(url)){
```
iamgusain, line 182: >isValidRedirectUri [](start = 19, length = 18)

if it is a valid redirect uri, why are we processing it as invalidRedirectUri. Did you mean to invert the condition here?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
tanmaymanolkar1: oh yes, thanks for pointing it out, will make the change.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
tanmaymanolkar1: Addressed here : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1195/commits/6f039fe2c3b99c4c6d497cea41431ba645475f83

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -366,39 +438,31 @@ public void run() {
                 view.stopLoading();
             }
         }, threadSleepForCallingActivity);
+    }
 
-        return true;
+    private void processInvalidRedirectUri(@NonNull final WebView view, @NonNull final String url) {
+        final String methodName = "#processRedirectUriCheck";
```
iamgusain, line 444: >processRedirectUriCheck [](start = 36, length = 23)

processInvalidRedirectUri

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -152,23 +155,59 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            Logger.info(TAG, "Request to open PlayStore.");
+            return processPlayStoreURL(view, url);
+        } else if (isAuthAppMFAUrl(formattedURL)){
+            Logger.info(TAG, "Request to link account with Authenticator.");
+            processAuthAppMFAUrl(url);
+        } else if (isInvalidRedirectUri(url)){
+            Logger.info(TAG, "Check for Redirect Uri.");
+            processInvalidRedirectUri(view, url);
+        } else if (isBlankPageRequest(formattedURL)){
+            Logger.info(TAG, "It is an blank page request");
```
iamgusain, line 180: return false;?  //for blank page?

iamgusain (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
tanmaymanolkar1: so in this case, the meaning of the returning true, is we are servicing the urls and we have anticipated them. Whereas false, is when we come across urls which dont necessarily fit in the use cases we want to service.

PR #1203, opened 2021-01-26
Copy logs to sdcard in finally block otherwise we lose logs on failure

shahzaibj:


AdamBJohnsonx (approved):
Since this is in the UIAutomation folder, can we actually take advantage of desugaring to use a try-with-resources?  No deployed code should depend on the automation library, right?

tanmaymanolkar1 (approved):

shahzaibj:
> Since this is in the UIAutomation folder, can we actually take advantage of desugaring to use a try-with-resources? No deployed code should depend on the automation library, right?

Yes deployed code shouldn't depend on automation library. We can probably try it if we wanted to. I believe you are talking about this: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html ?

PR #1204, opened 2021-01-27
Add JUnit Rule to get Factory Version of Chrome if needed

shahzaibj:
Add a Test Rule to downgrade Chrome and WebView to factory default version. The newer versions of Chrome & WebView seem to be not so suitable for automation and we run into resource id not found error. Basically it seems that Chrome rolled out an update due to which automation frameworks are not able to find elements on the web-page (at least this is the case for ESTS UX page). To mitigate this, if we find that Chrome or WebView version on the device is above the known version suitable for automation, then we would factory reset Chrome.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
@@ -93,8 +93,9 @@ public abstract class BaseSettings implements ISettings {
             launchIntent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS, data);
         } catch (final ActivityNotFoundException e) {
             Logger.e(TAG, "Package: " + packageName + " probably not available on device.", e);
-            //Open the generic Apps page:
-            launchIntent(Settings.ACTION_MANAGE_APPLICATIONS_SETTINGS);
+            // we could probably install the package from PlayStore (but for now let's fail the test
+            // to see if this ever happens in the wild)
+            throw new AssertionError(e);
         }
     }
 }
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
@@ -29,6 +29,7 @@ import android.content.Intent;
 import android.net.Uri;
 import android.provider.Settings;
 
+import androidx.annotation.NonNull;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.uiautomator.UiObject;
 
@@ -43,8 +44,6 @@ public abstract class BaseSettings implements ISettings {
 
     private final static String TAG = BaseSettings.class.getSimpleName();
     final static String SETTINGS_PKG = "com.android.settings";
-
-    public static final String TAG = BaseSettings.class.getSimpleName();
 
     @Override
     public void launchDeviceAdminSettingsPage() {
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
@@ -22,18 +22,18 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.ui.automation.device.settings;
 
+import android.content.ActivityNotFoundException;
 import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
+import android.net.Uri;
 import android.provider.Settings;
 
 import androidx.annotation.NonNull;
 import androidx.test.core.app.ApplicationProvider;
-import androidx.test.platform.app.InstrumentationRegistry;
-import androidx.test.uiautomator.UiDevice;
 import androidx.test.uiautomator.UiObject;
-import androidx.test.uiautomator.UiObjectNotFoundException;
 
+import com.microsoft.identity.client.ui.automation.logging.Logger;
 import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
 
 import org.junit.Assert;
@@ -43,6 +43,8 @@ import static com.microsoft.identity.client.ui.automation.utils.CommonUtils.laun
 public abstract class BaseSettings implements ISettings {
 
     final static String SETTINGS_PKG = "com.android.settings";
+
+    public static final String TAG = BaseSettings.class.getSimpleName();
 
     @Override
     public void launchDeviceAdminSettingsPage() {
@@ -77,5 +79,18 @@ public abstract class BaseSettings implements ISettings {
     @Override
     public void launchScreenLockPage() {
         launchIntent(Settings.ACTION_SECURITY_SETTINGS);
+    }
+
+    @Override
+    public void launchAppInfoPage(@NonNull final String packageName) {
+        try {
+            //Open the specific App Info page:
+            final Uri data = Uri.parse("package:" + packageName);
+            launchIntent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS, data);
+        } catch (final ActivityNotFoundException e) {
+            Logger.e(TAG, "Package: " + packageName + " probably not available on device.", e);
+            //Open the generic Apps page:
+            launchIntent(Settings.ACTION_MANAGE_APPLICATIONS_SETTINGS);
+        }
     }
 }
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/ISettings.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/ISettings.java
@@ -103,4 +103,6 @@ public interface ISettings {
      * Launches the security page in Settings app.
      */
     void launchScreenLockPage();
+
+    void launchAppInfoPage(String packageName);
 }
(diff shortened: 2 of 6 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/ISettings.java
```diff
@@ -103,4 +103,6 @@ void addWorkAccount(final ITestBroker expectedBroker,
      * Launches the security page in Settings app.
      */
     void launchScreenLockPage();
+
+    void launchAppInfoPage(String packageName);
```
shahzaibj, line 107: needs javadoc

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
```diff
@@ -0,0 +1,92 @@
+package com.microsoft.identity.client.ui.automation.rules;
```
shahzaibj, line 1: needs license

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
```diff
@@ -0,0 +1,92 @@
...
+import com.microsoft.identity.client.ui.automation.device.TestDevice;
+import com.microsoft.identity.client.ui.automation.logging.Logger;
+import com.microsoft.identity.client.ui.automation.utils.AdbShellUtils;
+import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class FactoryResetChromeRule implements TestRule {
```
shahzaibj, line 22: needs javadoc

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
```diff
@@ -0,0 +1,92 @@
...
+        // confirm disable in dialog
+        UiAutomatorUtils.handleButtonClick("android:id/button1");
+
+        // confirm downgrade to factory version
+        UiAutomatorUtils.handleButtonClick("android:id/button1");
+
+        // Enable Chrome
+        UiAutomatorUtils.handleButtonClick("com.android.settings:id/button1_positive");
+    }
+
```
shahzaibj, line 75: no longer needed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/CommonUtils.java
```diff
@@ -180,4 +181,16 @@ public static void launchIntent(@NonNull final String action) {
         intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_MULTIPLE_TASK);
         context.startActivity(intent);
     }
+
+    /**
+     * Launches an activity specified by the action.
```
shahzaibj, line 186: Fix javadoc

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/ISettings.java
shahzaibj: addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
shahzaibj: addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
shahzaibj: addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/CommonUtils.java
shahzaibj: addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
shahzaibj: addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/BaseSettings.java
```diff
@@ -81,4 +84,17 @@ public void launchScreenLockPage() {
...
+    @Override
+    public void launchAppInfoPage(@NonNull final String packageName) {
+        try {
+            //Open the specific App Info page:
+            final Uri data = Uri.parse("package:" + packageName);
+            launchIntent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS, data);
+        } catch (final ActivityNotFoundException e) {
+            Logger.e(TAG, "Package: " + packageName + " probably not available on device.", e);
+            //Open the generic Apps page:
+            launchIntent(Settings.ACTION_MANAGE_APPLICATIONS_SETTINGS);
```
shahzaibj, line 97: we should just fail here

shahzaibj: we can't do the test if Chrome isn't there? I guess we could install it from PlayStore....but for now let's just fail it

shahzaibj: Addressed

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
```diff
@@ -0,0 +1,103 @@
...
+            @Override
+            public void evaluate() throws Throwable {
+                final String chromeVersion = getPackageMajorVersion(BrowserChrome.CHROME_PACKAGE_NAME);
+                if (!TextUtils.isEmpty(chromeVersion) &&
+                        Integer.parseInt(chromeVersion) > CHROME_MAJOR_VERSION_SUITABLE_FOR_AUTOMATION) {
+                    Logger.w(TAG, "Chrome version on the device is higher than the known version suitable for automation. " +
+                            "We are going to attempt to factory reset Chrome and hope that it will give us the desired version.");
+                    // adb uninstall does factory reset for default apps
+                    AdbShellUtils.removePackage(BrowserChrome.CHROME_PACKAGE_NAME);
+                }
```
AdamBJohnsonx, line 68: Should we check this and see if it worked?

shahzaibj: We can log the new version after downgrade (if it still not the version we want then I am planning to just be optimistic and proceed with the test anyway -- because may be it starts working one day)

AdamBJohnsonx: Hm.  If we're insisting on a version in order to run the tests, and then we don't get it but it succeeds anyway, can we get the framework to give us a signal that our assumptions have changed *other* than just failing the test?

AdamBJohnsonx (approved):
I don't see anything *wrong* here.  Maybe we want to be vocal about the version if we assert that we want a particular number and can't get it, but if we do proceed and the test passes with a new version, we should let ourselves know so that we can update it.

And I love the documentation.  Thank you, Shahzaib!

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/FactoryResetChromeRule.java
shahzaibj: We would have the logs but won't have an alert. We can look into to see how to send the alert. Another thing we can do is that instead of failing or passing the test, we can abort the test if the version we need is not there using assumptions: https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assumptions.html

AdamBJohnsonx: How likely is it that any given engineer is going to pay attention to any of this?

shahzaibj: Not likely...abort is just ignoring the test and likely the engineer won't pay much attention. I don't currently have a mechanism to pass the test and still throw an alert (it might be possible but would require investigation). For the time being then let's just fail the test if we don't get the version that we are looking for even after downgrade. Test failure is something that release engineer should be looking at.

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1204/commits/1561de7500afe0b01c7f865d04b37bdf22a4a4b5

tanmaymanolkar1 (approved):

PR #1207, opened 2021-01-27
Increase temp user creation wait time to 20 seconds

shahzaibj:
15 seconds does not seem to always be enough and AAD doesn't recongize the user...increasing to 20 seconds

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/LoadLabUserTestRule.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/LoadLabUserTestRule.java
@@ -35,6 +35,8 @@ import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * A Test Rule to load lab user for the provided query prior to executing the test case.
  */
@@ -42,7 +44,7 @@ public class LoadLabUserTestRule implements TestRule {
 
     private final static String TAG = LoadLabUserTestRule.class.getSimpleName();
 
-    public static final int TEMP_USER_WAIT_TIME = 15000;
+    public static final long TEMP_USER_WAIT_TIME = TimeUnit.SECONDS.toMillis(20);
 
     private LabUserQuery query;
     private String tempUserType;


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/LoadLabUserTestRule.java
```diff
@@ -35,14 +35,16 @@
...
+
 /**
  * A Test Rule to load lab user for the provided query prior to executing the test case.
  */
 public class LoadLabUserTestRule implements TestRule {
 
     private final static String TAG = LoadLabUserTestRule.class.getSimpleName();
 
-    public static final int TEMP_USER_WAIT_TIME = 15000;
+    public static final long TEMP_USER_WAIT_TIME = TimeUnit.SECONDS.toMillis(20);
```
AdamBJohnsonx, line 47: ... can we poll for it?

shahzaibj: We don't have a way to know if it got created....we basically make interactive acquire token request and AAD page will show error "User not found" that's the only way to know as far as I am aware...so basically what we can do is proceed with the test without waiting and see it fail and then retry (every test will run at least 30 seconds because that's minimum timeout for interactive requests)...so I think instead of doing that we are better off just baking the wait in here and then proceed with the test

AdamBJohnsonx: So asking for the tempUserInfo from the LabApi won't guarantee consistency?

shahzaibj: Nope. The LAB API just returns data from their own db or wherever they are storing it. LAB API returning info for a temp user doesn't ensure that AAD is also aware of the user

AdamBJohnsonx: And we can't directly query AAD for this.

AdamBJohnsonx (approved):

PR #1210, opened 2021-01-28
Throw complete throwable (instead of just message) to get complete stack trace during test failure

shahzaibj:
This was already done in most of the places but seem to have missed a few.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -273,7 +273,7 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
         try {
             pinField.setText(String.valueOf(randomPin));
         } catch (final UiObjectNotFoundException e) {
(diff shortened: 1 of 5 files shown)


AdamBJohnsonx (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -273,7 +273,7 @@ public void handleAppProtectionPolicy() {
         try {
```
iamgusain, line 273: Do we need to try-catch now, or can that be handled at the calling method.

shahzaibj: better to just do it here and fail the test right away if UI Element is not found

iamgusain: I think that's fine. 
nit: This comment was a general comment for other changes as well, in case we want to handle any kind of exception and not just the one that is being caught

iamgusain (approved):

PR #1212, opened 2021-01-29
Increase http socket timeout for create temp user api to 15 seconds

shahzaibj:
The default ok http timeout is 10 seconds....increasing to 15 seconds for create temp user api because we're having some test failures with http timeout. Had a chat with Gladwin and his data says that on average temp user api calls are taking about 7 seconds but in some cases can go up to about 14 seconds...that's probably why we are seeing occasional failures with Socket Timeout as the default is 10 seconds...LAB Team will work on making the API more performant....but for now we are going to increase the timeout to 15 seconds

### DIFF
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
@@ -229,7 +229,6 @@ public class LabUserHelper {
     public static String loadTempUser(final String userType) {
         instance.setupApiClientWithAccessToken();
         CreateTempUserApi createTempUserApi = new CreateTempUserApi();
-        createTempUserApi.getApiClient().setConnectTimeout((int) TimeUnit.SECONDS.toMillis(15));
         createTempUserApi.getApiClient().setReadTimeout((int) TimeUnit.SECONDS.toMillis(15));
 
         TempUser tempUser;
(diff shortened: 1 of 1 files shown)


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabUserHelper.java
```diff
@@ -228,6 +229,8 @@ public static String loadUserForTest(LabUserQuery query) {
     public static String loadTempUser(final String userType) {
         instance.setupApiClientWithAccessToken();
         CreateTempUserApi createTempUserApi = new CreateTempUserApi();
+        createTempUserApi.getApiClient().setConnectTimeout((int) TimeUnit.SECONDS.toMillis(15));
+        createTempUserApi.getApiClient().setReadTimeout((int) TimeUnit.SECONDS.toMillis(15));
```
AdamBJohnsonx, line 233: Can we make the value a CONSTANT?  Also, do we want a 15 *second* connect timeout?  That's a little absurd for things running in the same network.

shahzaibj: yes I can make it a constant.....and yes let me remove it for connect and just do it for read

shahzaibj: Addressed

AdamBJohnsonx (approved):

PR #1213, opened 2021-02-02
Log key thumbprints less often

AdamBJohnsonx:
We are currently attempting to track changes in the encryption key used by MSAL, since changes in that
key will cause failures when reading that are difficult to detect that may be due to other problems.  To isolate
this, we were logging a hashed, fixed-value encryption every time we wrote an entry.

This was noisy for one of our customers, so we went to turn it down, and looking at the logs it wasn't working properly, because the cipher was being run in CBC mode, meaning that every value was different.  

This code corrects that flaw and only logs if it detects that there has been a change in the key.

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
@@ -36,6 +36,7 @@ import com.microsoft.identity.common.adal.internal.AndroidTestHelper;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
 
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -185,6 +186,31 @@ public class StorageHelperTests extends AndroidSecretKeyEnabledHelper {
         assertEquals("Same as initial text", decrypted, decrypted3);
         Log.d(TAG, "Finished testEncryptSameText");
     }
+
+    @Test
+    public void testKeyThumbprint() throws GeneralSecurityException, IOException {
+        Context context = getInstrumentation().getTargetContext();
+        final StorageHelper storageHelper = new StorageHelper(context);
+        for (int i = 0; i < 1000; i++) {
+            Assert.assertEquals(storageHelper.testThumbprint(), storageHelper.testThumbprint());
+        }
+    }
+
+    @Test
+    public void testKeyChange() throws GeneralSecurityException, IOException {
+        Context context = getInstrumentation().getTargetContext();
+        final StorageHelper storageHelper = new StorageHelper(context);
+        Assert.assertFalse(storageHelper.testKeyChange());
+        for (int i = 0; i < 500; i++) {
+            Assert.assertFalse(storageHelper.testKeyChange());
+        }
+        StorageHelper.LAST_KNOWN_THUMBPRINT.set("");
+        Assert.assertTrue(storageHelper.testKeyChange());
+        for (int i = 0; i < 500; i++) {
+            Assert.assertFalse(storageHelper.testKeyChange());
+        }
+    }
+
 
     @Test
     public void testTampering() throws GeneralSecurityException, IOException {
--- a/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
@@ -191,9 +191,7 @@ public class StorageHelperTests extends AndroidSecretKeyEnabledHelper {
     public void testKeyThumbprint() throws GeneralSecurityException, IOException {
         Context context = getInstrumentation().getTargetContext();
         final StorageHelper storageHelper = new StorageHelper(context);
-        for (int i = 0; i < 1000; i++) {
-            Assert.assertEquals(storageHelper.testThumbprint(), storageHelper.testThumbprint());
-        }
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -510,6 +534,18 @@ private String decryptWithSecretKey(@NonNull final byte[] bytes,
         return decrypted;
     }
 
+    private boolean logIfKeyHasChanged(@NonNull SecretKey secretKey, SecretKey hmacKey) throws NoSuchAlgorithmException, NoSuchPaddingException, InvalidKeyException, BadPaddingException, IllegalBlockSizeException {
+        final String keyThumbPrint = getKeyThumbPrint(secretKey, hmacKey);
+        if (!LAST_KNOWN_THUMBPRINT.get().equals(keyThumbPrint)) {
+            LAST_KNOWN_THUMBPRINT.set(keyThumbPrint);
+            if(!FIRST_TIME.compareAndSet(false, true)) {
+                Logger.info(TAG + ":decryptWithSecretKey", "Using key with thumbprint that has changed " + keyThumbPrint);
```
AdamBJohnsonx, line 542:
```suggestion
                Logger.info(TAG + ":logIfKeyHasChanged", "Using key with thumbprint that has changed " + keyThumbPrint);
```

iambmelt:
Can we add a PR description to this? Don't have context for this work

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
```diff
@@ -186,6 +187,31 @@ public void testEncryptSameText() throws GeneralSecurityException, IOException {
         Log.d(TAG, "Finished testEncryptSameText");
     }
 
+    @Test
+    public void testKeyThumbprint() throws GeneralSecurityException, IOException {
+        Context context = getInstrumentation().getTargetContext();
+        final StorageHelper storageHelper = new StorageHelper(context);
+        for (int i = 0; i < 1000; i++) {
+            Assert.assertEquals(storageHelper.testThumbprint(), storageHelper.testThumbprint());
```
tanmaymanolkar1, line 195: would it be better if we checked with here LAST_KNOWN_THUMBPRINT, instead of 2 function calls?

tanmaymanolkar1 (approved):

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
AdamBJohnsonx: It shouldn't change.  I don't really need the loop, even.

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
```diff
@@ -186,6 +187,31 @@ public void testEncryptSameText() throws GeneralSecurityException, IOException {
         Log.d(TAG, "Finished testEncryptSameText");
     }
 
+    @Test
+    public void testKeyThumbprint() throws GeneralSecurityException, IOException {
+        Context context = getInstrumentation().getTargetContext();
+        final StorageHelper storageHelper = new StorageHelper(context);
+        for (int i = 0; i < 1000; i++) {
```
AdamBJohnsonx, line 194:
```suggestion
```

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
```diff
@@ -186,6 +187,31 @@ public void testEncryptSameText() throws GeneralSecurityException, IOException {
...
+            Assert.assertEquals(storageHelper.testThumbprint(), storageHelper.testThumbprint());
+        }
```
AdamBJohnsonx, line 196:
```suggestion
```

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
AdamBJohnsonx, line 195:
```suggestion
        Assert.assertEquals(storageHelper.testThumbprint(), storageHelper.testThumbprint());
```

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
tanmaymanolkar1: okay this then resolves my comment thanks :)

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
tanmaymanolkar1: yeah I asked because of the looping

AdamBJohnsonx: testThumbprint doesn't actually mess with that.

> Can we add a PR description to this? Don't have context for this work

done.  I put you on this mainly because it's messing very lightly with encryption algorithms, and though there doesn't seem to be a better way to do this given the java SecretKey api, you might know one.

AdamBJohnsonx:
> Can we add a PR description to this? Don't have context for this work

done. I put you on this mainly because it's messing very lightly with encryption algorithms, and though there doesn't seem to be a better way to do this given the java SecretKey api, you might know one.  (or know that .equals always works for SecretKey)

iambmelt (approved):
LGTM thanks @AdamBJohnsonx

PR #1214, opened 2021-02-02
Generate and use a persistent device id for telemetry

iambmelt:
Impetus:
- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1050
- https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1282149

Summary of changes:
- Rather than rely on `Settings.Secure.ANDROID_ID`, generate a random GUID that persists for the life of the install and use that value

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
@@ -26,16 +26,12 @@ import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.Build;
-import android.provider.Settings;
 
 import androidx.annotation.NonNull;
 
 import com.microsoft.identity.common.WarningType;
-import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.internal.logging.Logger;
 
-import java.io.UnsupportedEncodingException;
-import java.security.NoSuchAlgorithmException;
 import java.util.TimeZone;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -51,8 +47,12 @@ import static com.microsoft.identity.common.internal.telemetry.TelemetryEventStr
 public class TelemetryContext extends Properties {
     private static final String TAG = TelemetryContext.class.getSimpleName();
 
-    TelemetryContext(final ConcurrentHashMap<String, String> delegate) {
+    private TelemetryPropertiesCache mTelemetryPropsCache;
+
+    private TelemetryContext(final ConcurrentHashMap<String, String> delegate,
+                             @NonNull final Context context) {
         super(delegate);
+        mTelemetryPropsCache = new TelemetryPropertiesCache(context);
     }
 
     /**
@@ -60,8 +60,11 @@ public class TelemetryContext extends Properties {
      * Context}. The {@link Telemetry} client can be called from anywhere, so the returned instances
      * is thread safe.
      */
-    static synchronized TelemetryContext create(final Context context) {
-        final TelemetryContext telemetryContext = new TelemetryContext(new ConcurrentHashMap<String, String>());
+    static synchronized TelemetryContext create(@NonNull final Context context) {
+        final TelemetryContext telemetryContext = new TelemetryContext(
+                new ConcurrentHashMap<String, String>(),
+                context
+        );
         telemetryContext.addApplicationInfo(context);
         telemetryContext.addDeviceInfo(context);
         telemetryContext.addOsInfo();
@@ -92,19 +95,7 @@ public class TelemetryContext extends Properties {
         put(Device.MANUFACTURER, Build.MANUFACTURER);
         put(Device.MODEL, Build.MODEL);
         put(Device.NAME, Build.DEVICE);
-        try {
-            put(
-                    Device.ID,
-                    StringExtensions.createHash(
-                            Settings.Secure.getString(
-                                    context.getContentResolver(),
-                                    Settings.Secure.ANDROID_ID
-                            )
-                    )
-            );
-        } catch (final NoSuchAlgorithmException | UnsupportedEncodingException exception) {
-            Logger.warn(TAG, "Unable to get the device id.");
-        }
+        put(Device.ID, mTelemetryPropsCache.getOrCreateRandomStableDeviceId());
     }
 
     void addOsInfo() {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
```diff
@@ -0,0 +1,74 @@
...
+ * Tracks properties used in telemetry.
+ */
+class TelemetryPropertiesCache {
+
+    private static final String SHARED_PREFS_NAME = "com.microsoft.common.telemetry-properties";
+
+    /**
+     * Cached properties.
+     */
+    private static class Properties {
```
AdamBJohnsonx, line 44: Do we need this extra class?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
```diff
@@ -0,0 +1,74 @@
...
+        mSharedPrefs = new SharedPreferencesFileManager(context, SHARED_PREFS_NAME);
+    }
+
+    /**
+     * Gets or creates the stable device id for this installation.
+     *
+     * @return The String ID used to refer to this device.
+     */
+    synchronized String getOrCreateRandomStableDeviceId() {
+        String deviceId = mSharedPrefs.getString(Properties.DEVICE_ID_GUID);
```
AdamBJohnsonx, line 64: Is this truly per install, or per-install-per-app?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Cached properties.
+     */
+    private static class Properties {
+
+        /**
+         * The randomly generated identifier for this device.
+         */
+        static final String DEVICE_ID_GUID = "device_id_guid";
```
AdamBJohnsonx, line 49: Why not private?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
```diff
@@ -92,19 +95,7 @@ void addDeviceInfo(@NonNull final Context context) {
...
-                            Settings.Secure.getString(
-                                    context.getContentResolver(),
-                                    Settings.Secure.ANDROID_ID
-                            )
-                    )
-            );
-        } catch (final NoSuchAlgorithmException | UnsupportedEncodingException exception) {
-            Logger.warn(TAG, "Unable to get the device id.");
-        }
+        put(Device.ID, mTelemetryPropsCache.getOrCreateRandomStableDeviceId());
```
AdamBJohnsonx, line 98: Why not the ad-id?

AdamBJohnsonx (commented):
Some questions.

tanmaymanolkar1 (commented):
Same question as @AdamBJohnsonx mentioned about Ad-id, and I am guessing since context is passed to create a sharedpref object, it would be specific to the app, we would have one constant id per-install-per-app.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
iambmelt: Ad-ID is just a GUID, not sure why we would add a library dependency on Play Services for this...
https://developers.google.com/android/reference/com/google/android/gms/ads/identifier/AdvertisingIdClient

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
iambmelt: Per-install-per-app. Will get generated once and reused until SharedPrefs are cleared

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
iambmelt: Just lazy, can fix :D

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
iambmelt: By no means required, I'd added thinking if we add lots of properties to this class over time this could be an easy way to organize what the properties are... no strong feelings about this, though

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryContext.java
AdamBJohnsonx: You get two things out of it.  The ad-id is actually per-device, and it is resettable by the user.  If the concern is privacy, then that's better than the per-install flag.

AdamBJohnsonx: This is in support of a public telemetry API.  I think that an application bound GUID is the right approach, but we should be VERY EXPLICIT about it in the documentation so that people aren't confused.

iambmelt: Synced offline, will use a generated GUID

AdamBJohnsonx (approved):
Since we might one-day publish the device id in question to a consuming application, not using ad-id is the right thing to do, since that app might not know what the constraints on handling that value would be and we'd have to change it this way then.  Just do it now.  My minor comments about not having redundant class definitions in the dex are a suggestion.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
iambmelt: de36efe259f1a4ee4d54d5226d384ad8dc9868fa

common/src/main/java/com/microsoft/identity/common/internal/telemetry/TelemetryPropertiesCache.java
iambmelt: Using a region instead: c147f3e8cb256192175951205efaad64a156a4f6

PR #1215, opened 2021-02-02
Fix InteractiveRequest Bound Service backcompat

rpdome:
**Issue**
In older, pre-ContentProvider, version of broker. We're not sending `class name` and `activity name` as part of the Interactive  Request Bundle (this [pr](https://github.com/AzureAD/ad-accounts-for-android/pull/1298), line 255).

In 2.0.4, we regressed this - MSAL 2.0.4+ will throw an error since those name can't be found anywhere in the bundle.

(Pre 2.0.4, we have separate strategies for ContentProvider and BoundService, and BoundService Strategy actually handles this - so this is a new issue.)

**Fix**
In MicrosoftAuthClient (used by Bound Service IPC Strategy in MSAL), extracts class and activity name and put those in bundle if not null.

**Test**
1. Commented out ContentProviderStrategy (so that MSAL uses Bound Service)
2. Remove BROKER_PACKAGE_NAME and BROKER_ACTIVITY_NAME from getBundleForInteractiveRequest on the broker side, and see if AcquireToken flow can be launched properly in MSAL.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
@@ -89,11 +89,11 @@ public class MicrosoftAuthClient extends BoundServiceClient<IMicrosoftAuthServic
                 final Bundle bundle = intent.getExtras();
 
                 //older brokers (pre-ContentProvider) are ONLY sending these values in the intent itself.
-                final String packageName = intent.getPackage();
-                final String className = intent.getComponent().getClassName();
-                if (!TextUtils.isEmpty(packageName) && !TextUtils.isEmpty(className)){
-                    bundle.putString(BROKER_PACKAGE_NAME, packageName);
-                    bundle.putString(BROKER_ACTIVITY_NAME, className);
+                if (intent.getComponent() != null &&
+                        !TextUtils.isEmpty(intent.getPackage()) &&
+                        !TextUtils.isEmpty(intent.getComponent().getClassName())){
+                    bundle.putString(BROKER_PACKAGE_NAME, intent.getPackage());
+                    bundle.putString(BROKER_ACTIVITY_NAME, intent.getComponent().getClassName());
                 }
 
                 return bundle;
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -83,7 +86,17 @@ public MicrosoftAuthClient(@NonNull final Context context,
 
             case MSAL_GET_INTENT_FOR_INTERACTIVE_REQUEST:
                 final Intent intent = microsoftAuthService.getIntentForInteractiveRequest();
-                return intent.getExtras();
+                final Bundle bundle = intent.getExtras();
```
iamgusain, line 89: Should we add UT for this change?

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -83,7 +86,17 @@ public MicrosoftAuthClient(@NonNull final Context context,
...
+
+                //older brokers (pre-ContentProvider) are ONLY sending these values in the intent itself.
+                final String packageName = intent.getPackage();
+                final String className = intent.getComponent().getClassName();
+                if (!TextUtils.isEmpty(packageName) && !TextUtils.isEmpty(className)){
+                    bundle.putString(BROKER_PACKAGE_NAME, packageName);
+                    bundle.putString(BROKER_ACTIVITY_NAME, className);
+                }
```
iamgusain, line 97: do we expect the else case to hit ever; if so should we log something to help in debugging?

rpdome: The else case happens when MSAL talks to the post-ContentProvider (newer) broker. In such case, everything is already packed inside the bundle.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
rpdome: We could.

shahzaibj (approved):

iambmelt (approved):

iamgusain (approved):

tanmaymanolkar1 (approved):

PR #1224, opened 2021-02-09
Implements Support for 'additionalField' Merging

iambmelt:
Impetus:
- https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1221076

Summary:
- If writing an `Account` or `Credential`, combine any `addtionalField` properties into the object we're about to write
- Adds tests

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
@@ -103,21 +103,37 @@ public class SharedPreferencesAccountCredentialCache extends AbstractAccountCred
     }
 
     @Override
-    public synchronized void saveAccount(@NonNull final AccountRecord account) {
+    public synchronized void saveAccount(@NonNull final AccountRecord accountToSave) {
         Logger.verbose(TAG, "Saving Account...");
-        Logger.verbose(TAG, "Account type: [" + account.getClass().getSimpleName() + "]");
-        final String cacheKey = mCacheValueDelegate.generateCacheKey(account);
+        Logger.verbose(TAG, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
+        final String cacheKey = mCacheValueDelegate.generateCacheKey(accountToSave);
         Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(account);
+
+        // Perform any necessary field merging on the Account to save...
+        final AccountRecord existingAccount = getAccount(cacheKey);
+
+        if (null != existingAccount) {
+            accountToSave.mergeAdditionalFields(existingAccount);
+        }
+
+        final String cacheValue = mCacheValueDelegate.generateCacheValue(accountToSave);
         mSharedPreferencesFileManager.putString(cacheKey, cacheValue);
     }
 
     @Override
-    public synchronized void saveCredential(@NonNull Credential credential) {
+    public synchronized void saveCredential(@NonNull Credential credentialToSave) {
         Logger.verbose(TAG, "Saving credential...");
-        final String cacheKey = mCacheValueDelegate.generateCacheKey(credential);
+        final String cacheKey = mCacheValueDelegate.generateCacheKey(credentialToSave);
         Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(credential);
+
+        // Perform any necessary field merging on the Credential to save...
+        final Credential existingCredential = getCredential(cacheKey);
+
+        if (null != existingCredential) {
+            credentialToSave.mergeAdditionalFields(existingCredential);
+        }
+
+        final String cacheValue = mCacheValueDelegate.generateCacheValue(credentialToSave);
         mSharedPreferencesFileManager.putString(cacheKey, cacheValue);
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
@@ -66,8 +66,15 @@ public abstract class AccountCredentialBase {
         }
 
         if (null != other.getAdditionalFields()) {
-            // TODO Ensure merging preserves the property we intend to write, and doesn't overwrite with a existing, possibly conflicting value
-            mAdditionalFields.putAll(other.getAdditionalFields());
+            for (Map.Entry<String, JsonElement> entry : other.getAdditionalFields().entrySet()) {
+                // Only add elements not present in the existing collection, so that old data
+                // does not overwrite new data...
+                if (mAdditionalFields.containsKey(entry.getKey())) {
+                    continue;
+                }
+
+                mAdditionalFields.put(entry.getKey(), entry.getValue());
+            }
         }
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
@@ -26,6 +26,7 @@ import androidx.annotation.NonNull;
 
 import com.google.gson.JsonElement;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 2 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -103,21 +103,37 @@ public SharedPreferencesAccountCredentialCache(
...
         Logger.verbose(TAG, "Saving credential...");
-        final String cacheKey = mCacheValueDelegate.generateCacheKey(credential);
+        final String cacheKey = mCacheValueDelegate.generateCacheKey(credentialToSave);
         Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(credential);
+
+        // Perform any necessary field merging on the Credential to save...
+        final Credential existingCredential = getCredential(cacheKey);
+
+        if (null != existingCredential) {
```
AdamBJohnsonx, line 132: Can we have this create a new copy of the credential rather than mutate it?

common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
```diff
@@ -52,6 +54,21 @@ public void setAdditionalFields(Map<String, JsonElement> additionalFields) {
         mAdditionalFields = additionalFields;
     }
 
+    /**
+     * Adds the additionalFields of the input AccountCredentialBase to this instance.
+     *
+     * @param other The input AccountCredentialBase to merge with this instance.
+     */
+    public void mergeAdditionalFields(@NonNull final AccountCredentialBase other) {
+        if (null == mAdditionalFields) {
```
rpdome, line 63: mAdditionalFields is transient. Wouldn't it be ignored by the serializer?

rpdome: hmm never mind... seems like we're adding these field separately [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/868bb9884d5425691b90b68cbee49d5cd32c465a/common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java#L125) before serialization (and after deserialization)

Do you know any reason why it's implemented this way?

iambmelt: It's implemented this way so that all arbitrary data is written like this:
```json
{
	"arbitrary_prop_1": "value",
	"arbitrary_prop_2": "another_value"
}
```

and not like this:
```json
{
	"mAdditionalFields": {
		"arbitrary_prop_1": "value",
		"arbitrary_prop_2": "another_value"
	}
}
```

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
iambmelt: I'd like to decline this suggestion: the cache doesn't know how to construct `Credentials` as they're an abstract type. Unless I'm mistaken, implementing the behavior you describe would requires adding clone(), copy constructors or a fair bit of code to inspect the type and reconstruct the instance of the subclass, copying all the relevant properties

common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
```diff
@@ -52,6 +54,21 @@ public void setAdditionalFields(Map<String, JsonElement> additionalFields) {
...
+            mAdditionalFields = new HashMap<>();
+        }
+
+        if (null != other.getAdditionalFields()) {
+            mAdditionalFields.putAll(other.getAdditionalFields());
```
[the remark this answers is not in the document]
iambmelt: That's a good question, let me follow up on this and get back

common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
```diff
@@ -52,6 +54,21 @@ public void setAdditionalFields(Map<String, JsonElement> additionalFields) {
...
     }
 
+    /**
+     * Adds the additionalFields of the input AccountCredentialBase to this instance.
+     *
+     * @param other The input AccountCredentialBase to merge with this instance.
+     */
+    public void mergeAdditionalFields(@NonNull final AccountCredentialBase other) {
+        if (null == mAdditionalFields) {
+            mAdditionalFields = new HashMap<>();
```
AdamBJohnsonx, line 64: If we're going to mutate this, this should really be a thread safe collection.

iambmelt: Good call out. I've pushed a change in 14a16c8d658b76089fae071c1398a1808749226b with the goal of being API back-compat friendly (method sigs don't change)

common/src/main/java/com/microsoft/identity/common/internal/dto/AccountCredentialBase.java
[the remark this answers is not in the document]
iambmelt: Confirmed: the data is currently only ever expected to be a string. The last value in should "win". I need to update the code to account for this and will add a test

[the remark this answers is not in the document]
AdamBJohnsonx: Oh, good.  Very good.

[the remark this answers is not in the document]
iambmelt: Added test in 0e7e35801a21e3d357106f90796050fb64045717

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -103,21 +103,37 @@ public SharedPreferencesAccountCredentialCache(
...
+        Logger.verbose(TAG, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
+        final String cacheKey = mCacheValueDelegate.generateCacheKey(accountToSave);
         Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(account);
+
+        // Perform any necessary field merging on the Account to save...
+        final AccountRecord existingAccount = getAccount(cacheKey);
+
+        if (null != existingAccount) {
+            accountToSave.mergeAdditionalFields(existingAccount);
```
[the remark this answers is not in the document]
iambmelt: Although this is more straightforward to do with Accounts (as opposed to Credentials), I'm thinking I will decline this change so that Accounts & Credentials are merged/written using the same logic -- rationale for credentials provided [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1224#discussion_r573143221). I appreciate the desirability of immutable state, though unless there's a compelling reason, I'm thinking leave as is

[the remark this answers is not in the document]
iambmelt: I'll resolve this issue, feel free to unresolve and/or submit proposed changes if you feel strongly about this

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
AdamBJohnsonx: Does the cacheValueDelegate thingy care about the actual type of the class?  It doesn't seem to.  The abstract class *could* just return an anonymous AccountCredentialBase that overrides that method.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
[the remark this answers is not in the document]
AdamBJohnsonx: Yeah, this is too painful to fix.

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesAccountCredentialCache.java
AdamBJohnsonx: OK, too painful.  We'd end up doing custom things to make this happen.  We talked offline, and I know I've seen this before in XML serialization.  In XML there was a lot of painful DOM manipulation to make these extra properties not just still present, but also stable (and comments, too) for all inheritors of a base class.  I want to say that we avoided this for JSON once upon a time by having the serializers gather all of the extra properties into a map on serialization using basically some GSON TypeAdapter trick in one case, and in the other we just treated all objects as generic key-value storage and only enforced typing on a per-key basis.

PR #1225, opened 2021-02-09
Pass requested claims to AT Deletion function when writing an AT

logdog16:
I forgot to pass RC when deleting existing AT's (overwrite scenario) in my original PR.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [PATCH] Use requested claims as a cache key when overwriting an AT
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [PATCH] Use requested claims as a cache key when overwriting an AT
```
iambmelt, line 3:
```suggestion
- [PATCH] Use requested claims as a cache key when overwriting an AT (#1225)
```

iambmelt (approved):
Approving with a minor suggestion.

PR #1227, opened 2021-02-10
Add a thread pool construction method

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/util/ThreadUtils.java
```diff
@@ -50,4 +57,44 @@ public static void sleepSafely(final int sleepTimeInMs, @NonNull final String ta
             }
         }
     }
+
+    /**
+     * Construct a thread pool with specified name and bounded size.
+     * @param corePool The smallest number of threads to keep alive in the pool.
+     * @param maxPool The maximum number of threads to allow in the thread pool, after which RejectedExecutionException will occur.
```
AdamBJohnsonx, line 64: This is only if the queue is also full, parameterize that as well.

PR #1232, opened 2021-02-11
Fix Bug in BrokerApplicationMetadata Tracking

iambmelt:
Impetus:
- [IcM 218697028 (corpnet required)](https://portal.microsofticm.com/imp/v3/incidents/details/218697028/home)

Summary:
- If an app transitions from non-foci -> foci, duplicate instances of its `BrokerApplicationMetadata` can exist
- This issue found after an app that was previously non-foci, became foci -- this caused the broker write tokens to foci cache and then subsequently read them from the processUid cache during silent requests

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
@@ -52,32 +52,6 @@ public class SharedPreferencesBrokerApplicationMetadataCache
     }
 
     @Override
-    public boolean insert(@NonNull final BrokerApplicationMetadata brokerApplicationMetadata) {
-        // Before we insert a record, if we have a existing entry with the same client id, env, and
-        // processUid, we should remove that entry...
-        disposeOfDuplicateRecords(
-                brokerApplicationMetadata.getClientId(),
-                brokerApplicationMetadata.getEnvironment(),
-                brokerApplicationMetadata.getUid()
-        );
-        return super.insert(brokerApplicationMetadata);
-    }
-
-    private void disposeOfDuplicateRecords(@NonNull final String clientId,
-                                           @NonNull final String environment,
-                                           final int uid) {
-        final List<BrokerApplicationMetadata> allMetadata = getAll();
-
-        for (final BrokerApplicationMetadata metadata : allMetadata) {
-            if (clientId.equalsIgnoreCase(metadata.getClientId())
-                    && environment.equalsIgnoreCase(metadata.getEnvironment())
-                    && uid == metadata.getUid()) {
-                remove(metadata);
-            }
-        }
-    }
-
-    @Override
     public Set<String> getAllClientIds() {
         final String methodName = ":getAllClientIds";
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
@@ -52,6 +52,32 @@ public class SharedPreferencesBrokerApplicationMetadataCache
     }
 
     @Override
+    public boolean insert(@NonNull final BrokerApplicationMetadata brokerApplicationMetadata) {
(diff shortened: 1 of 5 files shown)


shahzaibj:
Is there documentation on what is the process for becoming FOCI? As I understand this app was not part of the family but later became one? Why and how?

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
```diff
@@ -51,6 +51,32 @@ public SharedPreferencesBrokerApplicationMetadataCache(@NonNull final Context co
...
+        disposeOfDuplicateRecords(
+                brokerApplicationMetadata.getClientId(),
+                brokerApplicationMetadata.getEnvironment(),
+                brokerApplicationMetadata.getUid()
+        );
+        return super.insert(brokerApplicationMetadata);
+    }
+
+    private void disposeOfDuplicateRecords(@NonNull final String clientId,
+                                           @NonNull final String environment,
```
shahzaibj, line 67: How does this work for cross cloud apps? Could there be records in other environments? In this case I believe they won't be removed? Do we want to remove them though?

iambmelt:
@shahzaibj 
> Is there documentation on what is the process for becoming FOCI?

There is; MSFT employees can view this guidance [here](https://identitydocs.azurewebsites.net/static/v2/foci.html?q=family%20of%20client#how-to-request-being-added-to-a-family-of-client-ids).

>As I understand this app was not part of the family but later became one? Why and how?

I'll refrain from mentioning the specifics on public GitHub. MSFT employees can see this related IcM [here (corpnet required)](https://portal.microsofticm.com/imp/v3/incidents/details/212655831/home).

shahzaibj:
> @shahzaibj
> 
> > Is there documentation on what is the process for becoming FOCI?
> 
> There is; MSFT employees can view this guidance [here](https://identitydocs.azurewebsites.net/static/v2/foci.html?q=family%20of%20client#how-to-request-being-added-to-a-family-of-client-ids).
> 
> > As I understand this app was not part of the family but later became one? Why and how?
> 
> I'll refrain from mentioning the specifics on public GitHub. MSFT employees can see this related IcM [here (corpnet required)](https://portal.microsofticm.com/imp/v3/incidents/details/212655831/home).

Sounds good. Thanks @iambmelt!

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
iambmelt: Since we don't yet support x-cloud apps and I don't have insight into the future caching changes @iamgusain will be making in broker, that's a design question @iamgusain will need to account for in his implementation

iambmelt: Assuming they use the same client registration, [this function](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java#L74-L77) can be used to determine if an app is foci, even x-cloud

shahzaibj: Sounds good...resolving..

shahzaibj (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
iambmelt: Possibly [this function](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java#L137) gets updated to accept a wildcard environment... I don't think this code will be super hard to adapt; will defer to @iamgusain to adapt this implementation as required when the time comes

iambmelt:
Thanks @shahzaibj!

PR #1235, opened 2021-02-16
Stabilize the storage helper key change detection test

AdamBJohnsonx:
Depending on how this test was running, it was unstable.  Make
a change to have it be deterministic.

It was passing when I ran the tests through common, but failing when I ran the connectedAndroidTest target in MSAL.

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
@@ -197,8 +197,9 @@ public class StorageHelperTests extends AndroidSecretKeyEnabledHelper {
     @Test
     public void testKeyChange() throws GeneralSecurityException, IOException {
         Context context = getInstrumentation().getTargetContext();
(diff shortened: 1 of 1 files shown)


tanmaymanolkar1 (approved):

common/src/androidTest/java/com/microsoft/identity/common/adal/internal/cache/StorageHelperTests.java
```diff
@@ -197,8 +197,9 @@ public void testKeyThumbprint() throws GeneralSecurityException, IOException {
     @Test
     public void testKeyChange() throws GeneralSecurityException, IOException {
         Context context = getInstrumentation().getTargetContext();
+        StorageHelper.LAST_KNOWN_THUMBPRINT.set("");
```
shahzaibj, line 200: Should this be executed before each test and move to an `@before`?

AdamBJohnsonx: This is really only important for this test that tries to make sure of our ability to determine that a key has changed, I would think.

PR #1236, opened 2021-02-17
Add support for launch Broker auth Activity without an Activity Context

iambmelt:
Impetus:
- https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1192768

Goal:
- OneAuth-MSAL may not have an `Activity` `Context` available after initialization, but _will_ have a `Context` from the application. To support launching broker in this state, `FLAG_ACTIVITY_NEW_TASK` can be set on the request, if there is no `Activity`

This complete changeset to support this functionality will span:
- AndroidCommon
    * This PR
- MSALCPP
    * Changes to `BrokerImpl`
- OneAuth-MSAL
    * Changes to `BrokerFactory` & `OneAuth`

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@ V.Next
 - [PATCH] Trim() Cache Lookup Input Parameters (#1228)
 - [MINOR] Adds workaround for Mockito.openMocks() desugaring issue (#1229)
 - [PATCH] Fix bug where duplicate BrokerApplicationMetadata entries could be created for a single app (#1232)
+- [MINOR] Adds support for launching Broker auth Activity without an Activity Context from OneAuth-MSAL by setting FLAG_ACTIVITY_NEW_TASK
 
 Version 3.1.0
 ----------
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -236,14 +236,22 @@ public class BrokerMsalController extends BaseController {
         //Get the broker interactive parameters intent
         final Intent interactiveRequestIntent = getBrokerAuthorizationIntent(parameters);
 
+        // To support calling from OneAuth-MSAL, which may be initialized without an Activity
+        // add Flags to start as a NEW_TASK if we are launching from an application Context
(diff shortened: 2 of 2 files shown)


changelog.txt
```diff
@@ -6,6 +6,7 @@ V.Next
 - [PATCH] Trim() Cache Lookup Input Parameters (#1228)
 - [MINOR] Adds workaround for Mockito.openMocks() desugaring issue (#1229)
 - [PATCH] Fix bug where duplicate BrokerApplicationMetadata entries could be created for a single app (#1232)
+- [MINOR] Adds support for launching Broker auth Activity without an Activity Context from OneAuth-MSAL by setting FLAG_ACTIVITY_NEW_TASK
```
iambmelt, line 9:
```suggestion
- [MINOR] Adds support for launching Broker auth Activity without an Activity Context from OneAuth-MSAL by setting FLAG_ACTIVITY_NEW_TASK (#1236)
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -236,14 +236,22 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
         //Get the broker interactive parameters intent
         final Intent interactiveRequestIntent = getBrokerAuthorizationIntent(parameters);
 
+        // To support calling from OneAuth-MSAL, which may be initialized without an Activity
+        // add Flags to start as a NEW_TASK if we are launching from an application Context
+        final boolean startAsNewTask = null == parameters.getActivity();
```
AdamBJohnsonx, line 241: nit: Why create a flag here and branch on it later?  parameters.getActivity() == null is a pretty compelling reason not to call .startActivity on it.  Moving it down also brings the comment into context with the addFlags call.

AdamBJohnsonx (approved):
nitpick.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
iambmelt: 1e7863706333d0a7edc5105ef172e77a2a575539

AdamBJohnsonx:
Still ship it. ⛵

PR #1237, opened 2021-02-18
Logging Improvements

tanmaymanolkar1:
**_Listing the changes made :_** 

1. Added logging indicating the starting and the completion of request.
2. Logging the CommandParameters for each request
    - If the PII is enabled, we serialize the object to JSON String.
    - if the PII is disabled we only the serialize the fields with `@Expose` notation
3. Refactored the `Logger.info` and `Logger.infoPII ` to not log the entire RequestContext.
4. Refactored the logging code which was creating a discrepancy in the categorization of the data fields.
5. Added functionality of displaying `thread_name` instead of `thread_id` in the log statements.
6. Refactored the `CommandResult` object to include `correlation_id` to generate a link between the inflight requests, such that the inflight requests would posses the `correlation_id` of the original request which generated the result.
7. Fix to keep the original correlation id intact, and to not let it mutate with the cascading requests hitting the caches in method `submitSilentReturningFuture()` in the class `CommandDispatcher`.


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -269,6 +269,19 @@ public void run() {
         }
     }
 
+    private static void logParameters(@NonNull String tag, @NonNull String correlationId, @NonNull Object parameters, @Nullable String publicApiId) {
+        final String TAG = tag + ":" + parameters.getClass().getSimpleName();
+
+        //TODO - conversion of PublicApiId in readable form.
+        Logger.info(TAG, DiagnosticContext.getRequestContext().toJsonString(), "Starting request for correlation id : ####" + correlationId + ", with PublicApiId : " + publicApiId);
```
shahzaibj, line 276: Why so many hashtags here? Looks it would be printed like `Starting request for correlation id : #### <GUID>`

That seems weird...I would expect the value for correlation id to be printed right after the colon

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -166,10 +168,6 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
     public static FinalizableResultFuture<CommandResult> submitSilentReturningFuture(@SuppressWarnings(WarningType.rawtype_warning) @NonNull final BaseCommand command) {
 
         final String methodName = ":submitSilent";
-        Logger.verbose(
```
shahzaibj, line 169: Why remove this line?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -232,18 +231,19 @@ public void run() {
                         } else {
                             Logger.info(
                                     TAG + methodName,
-                                    "Silent command result returned from cache."
+                                    "",
```
shahzaibj, line 234: Why are we passing an empty string here?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -515,6 +521,10 @@ public void onReceive(Context context, Intent intent) {
                         // set correlation id on Local Authentication Result
                         setCorrelationIdOnResult(commandResult, correlationId);
 
+                        Logger.info(TAG + methodName, "",
```
shahzaibj, line 524: It just seems confusing to be passing around these empty strings for correlation ids when we do have one actually

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -315,7 +315,8 @@ public static void info(final String tag, @Nullable final String message) {
         getInstance().log(
                 tag,
                 LogLevel.INFO,
-                DiagnosticContext.getRequestContext().toJsonString(),
+                DiagnosticContext.THREAD_ID + " : " + DiagnosticContext.getRequestContext().getOrDefault(DiagnosticContext.THREAD_ID, "UNSET") + ", " +
+                        DiagnosticContext.CORRELATION_ID + " : " + DiagnosticContext.getRequestContext().getOrDefault(DiagnosticContext.CORRELATION_ID, "UNSET"),
```
shahzaibj, line 319: Why making this change?
If we are making it, then why only for INFO level? Why not all levels?
If we are making it, then can this entire block of code move to its own method?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: Its to make it easier to search the entry and exit point of the request from the command dispatcher using a correlation Id.
The hashtags would only be in 2 statements indicating the same. The exit point is marked with ## followed by a correlation Id.
This can be altered to have less number # or any character or remove them completely, but then you would have to either type the complete stuff like
`Starting request for correlation id : <GUID>` or if you choose to search using a substring of this then probably would have to manually check all the hits to find the right one.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: thanks for pointing out, will alter the silent request logging, since some command might not get logged (the ones which are returned right away).

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: This has to do with the way initially the Logger,info/infopii were coded. We dont always to print the entire RequestContext, it just adds to the clutter, hence adding a `""` because as it is we are logging the CorrelationId after the statement, which is essential to main the user readability.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: Noted, will refactor it, again this was done in accordance with how the logging is and was(partly) strucutred.

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
tanmaymanolkar1: Yes, this can be applied to all the logging levels, and would move it in its own method.
To give some context, `DiagnosticContext.getRequestContext().toJsonString()` was working initally because, the `RequestContext ` only contained the CorrelationId and ThreadId. But additional items had to be introduced for tracking the SKUs in the  `RequestContext` so soon the logs started to get spammed with additional metadata(clutter).

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: I made some changes here @shahzaibj, ## for start of request and ** to specify the end.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: replaced it with more structured logging.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: Made changes to the log statements, and changed this part, to not pass the empty string.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: Made the change https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1237/commits/e3620b61cfa0376f89d61d646bfe46340d2f295d

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
tanmaymanolkar1: made the change to all the levels :https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1237/commits/e3620b61cfa0376f89d61d646bfe46340d2f295d

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -166,10 +169,15 @@ public static void submitSilent(@SuppressWarnings(WarningType.rawtype_warning) @
...
 
         final String methodName = ":submitSilent";
-        Logger.verbose(
-                TAG + methodName,
-                "Beginning execution of silent command."
-        );
+
+        final CommandParameters commandParameters = command.getParameters();
+        final String correlationId = initializeDiagnosticContext(commandParameters.getCorrelationId(),
+                commandParameters.getSdkType() == null ? SdkType.UNKNOWN.getProductName() : commandParameters.getSdkType().getProductName(), commandParameters.getSdkVersion());
```
iambmelt, line 175: Style nit: can we format this line to get it down to a line width < 100 char?

(I see there's a few places within this PR where lines get wide, consider breaking these lines up for readability)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -437,27 +466,23 @@ private static boolean eligibleToCacheException(BaseException exception) {
      * @param baseException
      * @param result
      */
-    private static CommandResult getCommandResultFromTokenResult(BaseException baseException, AcquireTokenResult result) {
+    private static CommandResult getCommandResultFromTokenResult(BaseException baseException, AcquireTokenResult result, String correlationId) {
```
iambmelt, line 469: Nit: `final` + nullability annots?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
```diff
@@ -22,23 +22,45 @@
...
+            }
+        }
+    }
 
     private ResultStatus mStatus;
     private Object mResult;
+    private String mCorrelationId;
 
-    public CommandResult(ResultStatus status, Object result){
+    public CommandResult(ResultStatus status, Object result, @Nullable String correlationId) {
```
iambmelt, line 53: API breaking change here - can we make it additive?

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -455,6 +469,8 @@ private void log(final String tag,
         sLogExecutor.execute(new Runnable() {
             @Override
             public void run() {
+                final String dateTimeStamp = getUTCDateTimeAsString();
```
iambmelt, line 472: We should move this up (approx. `L468`) so that the timestamp is generated on the _calling thread_ not the worker

changelog.txt
```diff
@@ -13,6 +13,7 @@ V.Next
 - [MINOR] Adds new API to support ADAL/MSAL migration off of AuthenticationSettings#setSecretKey to a 'managed key' generated by Common (#1244)
 - [PATCH] Fix MSAL-CPP injecting javascript loop (#1238)
 - [MINOR] Device PoP keys are now generated with attestation flags and expose a certificate chain getter (#1247)
+- [MINOR] Added logging functionality to track the request status.(#1237)
```
iambmelt, line 16: (As it stands this change is actually MAJOR (due to API breaks in `CommandResult`), if you make the ctor changes additive this PR will become a MINOR chage)

iambmelt (changes requested):
Requesting changes based on inline feedback, few tips:
- Mind the line widths, let's try to keep <100 char wherever we can (try to wrap someplace reasonable, like between params or on operators)
- Watch out for API breaking changes; if you need to make them, make sure the changelog accurately captures it by marking `MAJOR`

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
```diff
@@ -22,23 +22,45 @@
...
+                return "ERROR";
+            } else {
+                return "CANCEL";
+            }
+        }
+    }
 
     private ResultStatus mStatus;
     private Object mResult;
+    private String mCorrelationId;
```
AdamBJohnsonx, line 51: Can we make this final, and explicitly set it to null?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
AdamBJohnsonx: Yeah, we should add a new telescoping constructor.

tanmaymanolkar1: Yeah my bad, should've done that in the first place. will refactor it.

changelog.txt
tanmaymanolkar1: made the ctor changes additive to make it MINOR

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
iambmelt: LGTM

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -269,13 +283,34 @@ public void run() {
         }
     }
 
+    private static void logParameters(@NonNull String tag, @NonNull String correlationId,
+                                      @NonNull Object parameters, @Nullable String publicApiId) {
+        final String TAG = tag + ":" + parameters.getClass().getSimpleName();
+
+        //TODO - conversion of PublicApiId in readable form.
```
iambmelt, line 290: Leaving this TODO intentionally? Do we have a tracking item? Or something you plan to resolve in this PR?

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -139,6 +140,19 @@ public void setExternalLogger(final ILoggerCallback externalLogger) {
...
+    /**
+     * Get only the required metadata from the DiagnosticContext
+     * to plug it in the log lines.
+     * Here we are considering the correlation_id and the thread_name.
+     * The need for this is because DiagnosticContext contains additional metadata which is not always required to be logged.
+     *
+     * @return String The concatenation of thread_name and correlation_id to serve as the required metadata in the log lines.
+     */
+    public static String getDiagnosticContextMetadata() {
+        return DiagnosticContext.THREAD_NAME + " : " + DiagnosticContext.getRequestContext().getOrDefault(DiagnosticContext.THREAD_NAME, "UNSET") + ", " +
```
iambmelt, line 152: Nit: wide lines; can we condense to <100 chars?

iambmelt (approved):
Approving with suggestions (minor style callouts, line width).

Suggest getting an additional sign off, lest anyone else have feedback

Thanks @tanmaymanolkar1!

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -354,7 +400,8 @@ private static CommandResult executeCommand(@SuppressWarnings(WarningType.rawtyp
      * @param result
      * @param handler
      */
-    private static void returnCommandResult(@SuppressWarnings(WarningType.rawtype_warning) final BaseCommand command, final CommandResult result, Handler handler) {
+    private static void returnCommandResult(@SuppressWarnings(WarningType.rawtype_warning) final BaseCommand command,
+                                            final CommandResult result, Handler handler) {
```
AdamBJohnsonx, line 404: nit:
```suggestion
                                            final CommandResult result, @NonNull final Handler handler) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
```diff
@@ -22,23 +22,49 @@
...
+            } else if (ResultStatus.ERROR == this) {
+                return "ERROR";
+            } else {
+                return "CANCEL";
+            }
+        }
+    }
 
     private ResultStatus mStatus;
     private Object mResult;
```
AdamBJohnsonx, line 50: It looks like these are effectively immutable already - we could make it official at the reference level by adding 'final'

AdamBJohnsonx (approved):
Some minor things.

AdamBJohnsonx:
One thing - did you want to adopt thread pool naming here?

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
tanmaymanolkar1: will do!

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
tanmaymanolkar1: Oh yes, will pin an ADO item

tanmaymanolkar1:
> One thing - did you want to adopt thread pool naming here?

@AdamBJohnsonx I have adopted it inside the logging, keeping the thread ID intact for the reason of not breaking any unknown dependency.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandResult.java
tanmaymanolkar1: Not sure about immutability @AdamBJohnsonx since we do set correlationId on result if its a LocalAuthenticationResult

AdamBJohnsonx: Right, but we don't change the object reference in this class. final doesn't say anything about any references contained in the object, just the one in this class.  We don't reset it, so we could make it final.

AdamBJohnsonx:
> > One thing - did you want to adopt thread pool naming here?
> 
> @AdamBJohnsonx I have adopted it inside the logging, keeping the thread ID intact for the reason of not breaking any unknown dependency.

Right now, thread names are essentially random, depending on the order of creation of the different thread pools.  Feel free to change them.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -81,8 +82,14 @@
...
 
     private static final int SILENT_REQUEST_THREAD_POOL_SIZE = 5;
-    private static final ExecutorService sInteractiveExecutor = Executors.newSingleThreadExecutor();
-    private static final ExecutorService sSilentExecutor = Executors.newFixedThreadPool(SILENT_REQUEST_THREAD_POOL_SIZE);
+    private static final int INTERACTIVE_REQUEST_THREAD_POOL_SIZE = 1;
+    //TODO:1315931 - Refactor the threadpools to not be unbounded for both silent and interactive requests.
+    private static final ExecutorService sInteractiveExecutor = ThreadUtils.getNamedThreadPoolExecutor(
+            1, INTERACTIVE_REQUEST_THREAD_POOL_SIZE, -1, 0, TimeUnit.MINUTES, "interactive"
+    );
+    private static final ExecutorService sSilentExecutor = ThreadUtils.getNamedThreadPoolExecutor(
```
tanmaymanolkar1, line 90: @iambmelt tagging you as a FYI, went over this with @AdamBJohnsonx

iambmelt: thx @tanmaymanolkar1 lgtm

AdamBJohnsonx (approved):

PR #1238, opened 2021-02-19
Fix MSAL-CPP injecting javascript loop

richard1122:
onPageLoaded is executed asynchronously, javascript might be injected into next loaded page, which will cause infinite loading.
We change to onPageCommitVisible callback, it is guarenteed not called if current main frame navigation is canceled.


iambmelt:
@richard1122 Please do not put things like 'hacking' in PR titles. Because this is an auth library, it sends the wrong message

richard1122:
image
image

Hi @jbzdarkid , I'm trying fix the PRT tests signin issue. Here are what I found:
UITestAutomation.js is injected every time when `com.microsoft.identity.common.internal.ui.webview.OnPageLoadedCallback#onPageLoaded` is called, Which causes page reload (form submit?). 

So in this PR I added a set to make sure we will only inject once for the same URL.

Changes to `UIAutomation.js` (only add logs)
```diff
--- a/test/testdata/UIAutomation.js
+++ b/test/testdata/UIAutomation.js
@@ -200,6 +200,7 @@ function SignIn(username, password)
     {
         throw "Couldn't find SignIn button";
     }
+    console.log('signinbutton.click');
     signInButton.click();
 }

@@ -295,6 +296,7 @@ var BACK_BUTTON = [

 function ExitWithError(message)
 {
+    console.log('ExitWithError', message);
     window.location = window.redirect + '?error=UIAutomation.js&error_description=' + encodeURI(message);
     return;
 }
@@ -324,6 +326,7 @@ function ExecuteSignIn()
     }
     catch (e)
     {
+        console.error(e);
         errorDescription = 'Failed attempt ' + attemptNo + '. Last error "' + e + '". ';
     }

@@ -486,6 +489,7 @@ function ClickBackButton()

 function Main()
 {
+    console.log('action: ', window.action);
     switch (window.action)
     {
```

jbzdarkid:
We intentionally re-run javascript on every page load. It is the responsibility of the javascript to take an action which progresses the page.
If the UI is unable to make progress, then that's a scenario bug, not a javascript bug (or a web browser bug).
Can you describe in a bit more detail what pages are being loaded, what action the javascript is taking, and what action a user would take?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -133,15 +134,29 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
                 getActivity(),
                 new AuthorizationCompletionCallback(),
                 new OnPageLoadedCallback() {
+
                     @Override
-                    public void onPageLoaded() {
+                    public void onPageLoaded(String url) {
```
iambmelt, line 139:
```suggestion
                    public void onPageLoaded(final String url) {
```

jbzdarkid (changes requested):
See comment in the thread. I don't think this is the correct approach. Feel free to schedule some time with me if you'd like to discuss in person.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -133,15 +134,29 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+                            String javascriptToExecute;
+                            try {
+                                // onPageLoaded is executed asynchronously, javascript might be
+                                // injected into next loaded page, which will cause infinite loading.
+                                // Here we inject current finished loading url and check if
+                                // the frame executing javascript is equal to expected one in JS.
+                                javascriptToExecute = String.format("window.expectedUrl = '%s';\n%s",
+                                        URLEncoder.encode(url, "UTF-8"),
+                                        mPostPageLoadedJavascript);
+                            } catch (UnsupportedEncodingException e) {
```
iambmelt, line 153:
```suggestion
                            } catch (final UnsupportedEncodingException e) {
```

shahzaibj, line 153:
```suggestion
                            } catch (final UnsupportedEncodingException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -133,15 +134,29 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+                                // Encode url component failed, fallback.
+                                Logger.warn(TAG, "Inject expectedUrl failed.");
```
shahzaibj, line 155: might want to log an error here along with passing the throwable

iambmelt:
> See comment in the thread. I don't think this is the correct approach. Feel free to schedule some time with me if you'd like to discuss in person.

@shahzaibj FYI ^^ -- I requested his review as well, lest this have any impact on our automated tests

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -166,15 +166,20 @@ public void onReceivedSslError(final WebView view,
     }
 
     @Override
-    public void onPageFinished(final WebView view,
-                               final String url) {
-        super.onPageFinished(view, url);
-        mPageLoadedCallback.onPageLoaded();
+    public void onPageCommitVisible(WebView view, String url) {
```
jbzdarkid, line 169:
```suggestion
    public void onPageCommitVisible(final WebView view, final String url) {
```

jbzdarkid (commented):
LGTM, please also get sign-off from someone in android common.

jbzdarkid (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OnPageCommitVisibleCallback.java
```diff
@@ -22,6 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.internal.ui.webview;
 
-public interface OnPageLoadedCallback {
-    void onPageLoaded();
+public interface OnPageCommitVisibleCallback {
```
iambmelt, line 25: This is an API breaking change for common - we should avoid this if we can.

Let's instead:
- Add a new interface
- Add a new ctor to `OAuth2WebViewClient` to accept that interface
- Switch relevant code to call new ctor

iambmelt (changes requested):
API breaking changes

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OnPageCommitVisibleCallback.java
richard1122: OK. Maybe let me just change the caller method.

richard1122: Thanks, added new `OnPageCommitVisibleCallback` and new ctor.

iambmelt (approved):
LGTM, thanks @richard1122!

PR #1242, opened 2021-02-22
Add custom test runner to accept test parameters into runner instead of supplying via BuildConfig

shahzaibj:
Partially address: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1277730

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/runner/IdentityRunnerArgs.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/runner/IdentityRunnerArgs.java
@@ -22,6 +22,10 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.ui.automation.runner;
 
+import android.util.Log;
+
+import androidx.annotation.NonNull;
+
 import com.microsoft.identity.client.ui.automation.installer.AppInstallSource;
 
 /**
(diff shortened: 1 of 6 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/runner/IdentityRunnerArgs.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.microsoft.identity.client.ui.automation.installer.AppInstallSource;
+
+/**
+ * Test runner args to be used with {@link IdentityTestRunner}.
+ */
+public class IdentityRunnerArgs {
+
+    public static boolean preferPreInstalledApks = false;
+    public static String brokerSource = AppInstallSource.PlayStore.getName();
```
AdamBJohnsonx, line 33: This is still creating global state... I wish we could create another way to do this, but it's less hardcoded than the build-config version.  Suggestion - we should expose setters here and log call information whenever the values are changed.

AdamBJohnsonx (approved):
A minor comment that I think would make it easier to see what's happening.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/runner/IdentityRunnerArgs.java
shahzaibj: Yeah unfortunately I couldn't find a way to keep these member fields and still be able to access the runner and its members form other places in the code so had to do it this way for now. Sure, I can add setters

AdamBJohnsonx: The only purpose for these setters would be so that we can log any changes as they happen, mind, so our logs can help us figure it out when we make mistakes.

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1242/commits/a98f1d9f66221f1d1c48ba60a1f431862e0ed3b1

AdamBJohnsonx: 🚢

PR #1205, opened 2021-01-27
Add mock for cloud discovery response + Shadow for HttpClient

gitaumoses4:
Related to https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1279

PR #1246, opened 2021-02-24
Expand the MockHttpClient request matching scenarios.

gitaumoses4:
## What does this PR do?
- Expands the capability of the MockHttpClient to cater for matching of:
1. Request URL
2. Request headers
3. Request method
4. Request body/content.

The original implementation was done here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1205

### DIFF
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/HttpRequestMatcher.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/HttpRequestMatcher.java
@@ -27,6 +27,7 @@ import com.microsoft.identity.common.internal.net.HttpClient;
 import java.net.URL;
 import java.util.Map;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import lombok.Builder;
 import lombok.NonNull;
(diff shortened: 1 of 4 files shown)


testutils/src/main/java/com/microsoft/identity/internal/testutils/HttpRequestMatcher.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public boolean matches(
+            final HttpClient.HttpMethod method,
+            @NonNull final URL url,
+            final Map<String, String> requestHeaders,
+            final byte[] requestContent) {
+        return (null == this.method || this.method.test(method))
+                && (null == this.url || this.url.test(url))
+                && (null == this.headers || this.headers.test(requestHeaders))
+                && (null == this.body || this.body.test(requestContent));
```
AdamBJohnsonx, line 64: We could make certain that none of these are null, and then the conditionals don't need to be so complex (they default to values returning true).

AdamBJohnsonx (approved):
A minor comment.

testutils/src/main/java/com/microsoft/identity/internal/testutils/HttpRequestMatcher.java
```diff
@@ -0,0 +1,66 @@
...
+ * Used to match http request based on headers, body, url and method. A request is considered as matched
+ * if all these 4 parameters are matched.
+ */
+@Builder
+public class HttpRequestMatcher {
+
+    private Predicate<Map<String, String>> headers = header -> true;
+    private Predicate<byte[]> body = content -> true;
+    private Predicate<URL> url = s -> true;
+    private Predicate<HttpClient.HttpMethod> method = s -> true;
```
AdamBJohnsonx, line 44: If we can't use the lambda syntax, android studio can convert them to anonymous classes for you.

gitaumoses4: Do we keep them as lambda syntax? Or will there be compatibility issues?

AdamBJohnsonx: We *should* be able to, but lets be safe and just make them anonymous.

andrewchoi5 (approved):

PR #1247, opened 2021-02-25
Add new API to set pop attestation flags and retrieve chain

iambmelt:
This is preemptive work to support exposing the certificate chain of the device pop key

Note:
- PoP keys initialized _before_ this PR won't have a certificate chain (will return `null` or `1` containing self) -- I'm determining this _not_ to be something I need to account for in this PR, given there is no prod deployment/adoption of these APIs yet


common/src/androidTest/java/com/microsoft/identity/common/internal/platform/DevicePoPManagerTests.java
```diff
@@ -449,4 +451,22 @@ public void testGenerateShr() throws ClientException, MalformedURLException, Par
         Assert.assertNotNull(jwtClaimsSet.getClaim("cnf"));
         Assert.assertEquals(clientClaims, jwtClaimsSet.getClaim("client_claims"));
     }
+
+    @Test
+    public void testHasCertificateChain24() throws ClientException {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
AdamBJohnsonx, line 457: Can you annotate this method and have that work?

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1307,6 +1340,14 @@ private void initialize23(@NonNull final KeyPairGenerator keyPairGenerator,
         keyPairGenerator.initialize(spec);
     }
 
+    @SuppressLint("NewApi")
+    @RequiresApi(Build.VERSION_CODES.N)
+    @NonNull
+    private KeyGenParameterSpec.Builder setAttestationChallenge(
```
AdamBJohnsonx, line 1346: Can we add a test for these?

AdamBJohnsonx (commented):
I feel like we should have a test.

common/src/androidTest/java/com/microsoft/identity/common/internal/platform/DevicePoPManagerTests.java
iambmelt: Annotated - I can revert it if this isn't respected by the test runner

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: The `getCertificateChain()` is the test -- [that function won't return a result unless an attestation challenge is set](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setAttestationChallenge(byte[])) -- if there's some other aspect of this we can test, I'm open to ideas. There didn't seem to be a lot to do here

AdamBJohnsonx: Yeah , this is private.  Oops.

AdamBJohnsonx (approved):

joakley-msft (approved):

PR #1248, opened 2021-02-26
Changes to support removing cross cloud accounts

iamgusain:
- Removed the null check for environment, to allow for the remove account to be environment agnostic.
- Added utility method to create mock response for token requests, this is used in cross cloud UTs Msal PR; https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1316


common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -1269,6 +1269,12 @@ private void setToCacheRecord(@NonNull final CacheRecord.CacheRecordBuilder targ
                     accountRecord
             );
 
+            if(idTokensForAccount == null || idTokensForAccount.size() == 0)
```
shahzaibj, line 1272: When or why might this happen? I think this is an error case and shouldn't really be happening

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -1407,8 +1413,7 @@ public AccountDeletionRecord removeAccount(@Nullable final String environment,
         );
 
         final AccountRecord targetAccount;
-        if (null == environment
-                || null == clientId
+        if (null == clientId
```
shahzaibj, line 1416: Doesn't make sense to me why the check for environment is removed? I think the check needs to stay there and instead the caller should supply `null` I believe to delete records from all environments, right?

shahzaibj (changes requested):
Left some comments

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iamgusain: With this check if environment is passed as null, remove account is no-op and returns early, without removing anything.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iamgusain: I ran into this case while testing, If an application(binary) is using more than one client id (say 2), and have credentials for same user in both the client ids (maybe for different tenants/clouds)
When we return getAccounts we return accounts for all client id, but tokens are filtered based on client id; so in that case some idTokens are filtered out if they do not belong to the clientId that was passed.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
shahzaibj: I think we should make the change **_there_** then

shahzaibj: @iambmelt Feel free to add your thoughts as well as you are the cache expert

iamgusain: This the place to remove the check, remove account happens after this check in this method

shahzaibj: Ok that makes sense.....hadn't looked at the rest of the code earlier......so yeah this check had to be removed then. Thanks for clarifying.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
shahzaibj: Seems a little strange... @iambmelt Do you know if this is expected behavior?

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -1269,6 +1269,12 @@ private void setToCacheRecord(@NonNull final CacheRecord.CacheRecordBuilder targ
                     accountRecord
             );
 
+            if(idTokensForAccount == null || idTokensForAccount.size() == 0)
+            {
```
iambmelt, line 1273: Style nit: needs to move up a line so brackets are

```java
void likeThis() {
    // foo
}
```

and not

```java
void likeThis()
{
    // foo
}
```

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt, line 1272:
```suggestion
            if (idTokensForAccount == null || idTokensForAccount.size() == 0) {
```

testutils/src/main/java/com/microsoft/identity/internal/testutils/mocks/MockServerResponse.java
```diff
@@ -43,6 +43,20 @@ public static HttpResponse getMockTokenSuccessResponse() {
         return new HttpResponse(200, mockResponse, new HashMap<String, List<String>>());
     }
 
+    public static HttpResponse getMockTokenSuccessResponse(final String localAccountId, final String tenant, final String issuer, final String rawClientInfo, final String accessToken) {
+        final String mockResponse = "{\n" +
```
iambmelt, line 47: Rather than strconcat this data together, can we use `JSONObject` or equivalent to construct the data?

iambmelt (changes requested):
Minor formatting/test changes requested - fundamentally these changes LGTM

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: This is expected behavior -- idtokens are scoped to a client_id, same as access tokens. If (in the event of dual-stacking) multiple idtokens are found corresponding to a user, only those matching the current client_id should be returned.

iambmelt (approved):
LGTM

shahzaibj (approved):
LGTM

PR #1253, opened 2021-03-01
Add changes to fix interrupt flow

rpdome:
- Add username (of the account that owns the flow) to BaseException
- Remove the AT check in validateTokensAreInResponse() for interrupt flow.


common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
```diff
@@ -150,6 +154,14 @@ public void setCorrelationId(@Nullable final String correlationId){
         mCorrelationId = correlationId;
     }
 
+    public String getUsername() {
+        return mUsername;
+    }
+
+    public void setUsername(String username) {
+        this.mUsername = username;
+    }
```
AdamBJohnsonx, line 163: Do we want to @Builderize this class or add a constructor with these additional arguments?

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -53,6 +54,16 @@ public static boolean isEmpty(final String message) {
         return message == null || message.trim().length() == 0; //NOPMD  Suppressing PMD warning for new String creation on trim()"
     }
 
+    /**
+     * Returns true if the string contains the given substring.
```
AdamBJohnsonx, line 58:
```suggestion
     * Returns true if the string contains the given substring and.
```
```suggestion
     * @param str the string to search for a substring in
     * @param substr the substring to search for
     * @return true if the string contains the given substring, false if it does not or if it is null
```

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/exception/BaseException.java
rpdome: No to both.

- What we're doing here is to inject username into an existing subclassed object. Using a builder here is sort of overkill IMO.
- This field is not common - it will set only in the interrupt flow the in-line WPJ scenario. Adding a new constructor here (and all the subclasses) would make the code unnecessarily cluttered. 

That said, I'll add Nullable on top of getUsername()

PR #1261, opened 2021-03-10
Initialize connectionAvailable when the network is connecting

gitaumoses4:
The check for internet connection was updated to use the new callback structure. 
https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback

We need to mark the `connectionAvailable` as true before the callback registration and mark it as `false` if an exception was thrown.


paulkagiri (approved):
LGTM

PR #1262, opened 2021-03-10
Implement and use a new method to only include portions of the shared prefences

AdamBJohnsonx:
Uses an `Iterator` for more-optimized perf


iambmelt (approved):
Thanks @AdamBJohnsonx

PR #1267, opened 2021-03-12
Some of these ciphers have very old support - allow it

AdamBJohnsonx:


shahzaibj (approved):
If it works then ship it!

PR #1268, opened 2021-03-13
Track the current interactive task if it's a broker task, terminating it if needed

AdamBJohnsonx:
Currently, we do not recover well from an interactive task that hangs.  We broadcast a message to allow them to stop, but we have no guarantee that this occurred.

I happened to generate a deadlock while testing some code, and noticed that the only way to recover was to quit the application.  This isn't so bad for user's applications, but it terrible for the broker, when what you have to quit is the broker itself.

What I'm doing here is attempting to see that the application has not responded to the cancellation message we sent (in brokered case) or it's been 1/2 second, kill the task that's running.


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +535,28 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 );
             }
 
-            sInteractiveExecutor.execute(new Runnable() {
+            // It is possible that the current task could be stuck irretrievably.  If that's
+            // what's happened, do everything in our power to make certain that it is dead.
+            // Then regenerate the executor service running it if it won't die nicely, so that
+            // we can meaningfully submit new tasks in the expectation that they will succeed.
```
AdamBJohnsonx, line 541:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +535,26 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                try {
+                    // We'll give it 1/2 second to respond to the kill message we broadcast.
+                    sCurrentInteractiveTask.get(500, TimeUnit.MILLISECONDS);
+                } catch (ExecutionException e) {
+                    // OK, it finished with an exception during that time.  Probably OK.
+                    Logger.info(TAG + methodName, null, "Previous task terminated with exception " + e.getMessage());
+                } catch (TimeoutException e) {
+                    // Nope, it's still going.  Send it a thread cancellation.
+                    Logger.warn(TAG + methodName, "Execution still running, attempting to cancel.");
+                    sCurrentInteractiveTask.cancel(true);
```
AdamBJohnsonx, line 550:
```suggestion
                    // This does return a value, but it doesn't tell us much that's actionable.
                    // It's true if the task was cancelled, but false could mean that it hadn't
                    // started yet.
                    sCurrentInteractiveTask.cancel(true);
```

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -360,9 +360,23 @@ public static void info(final String tag, @Nullable final String message) {
...
+        info(tag, correlationID, message, null);
     }
 
+    /**
+     * * Send a {@link LogLevel#INFO} log message without PII.
+     *
+     * @param tag           Used to identify the source of a log message. It usually identifies the
+     *                      class or activity where the log call occurs.
+     * @param correlationID Unique identifier for a request or flow used to trace program execution.
+     * @param message       The message to log.
```
AdamBJohnsonx, line 372:
```suggestion
     * @param message       The message to log.
     * @param exception      A related exception, or null.
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
 
-            sInteractiveExecutor.execute(new Runnable() {
+            // It is possible that the current task could be stuck irretrievably.  If that's
+            // what's happened, do everything in our power to make certain that it is dead.
+            Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
+            if (currentInteractiveTask != null && !(currentInteractiveTask.isDone() || currentInteractiveTask.isCancelled())) {
+                try {
+                    // We'll give it 1/2 second to respond to the kill message we broadcast.
+                    currentInteractiveTask.get(500, TimeUnit.MILLISECONDS);
+                } catch (ExecutionException e) {
```
AdamBJohnsonx, line 546:
```suggestion
                } catch (final ExecutionException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                    // OK, it finished with an exception during that time.  Probably OK.
+                    Logger.info(TAG + methodName, null, "Previous task terminated with exception " + e.getMessage());
+                } catch (TimeoutException e) {
```
AdamBJohnsonx, line 549:
```suggestion
                } catch (final TimeoutException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                    // Nope, it's still going.  Send it a thread cancellation.
+                    Logger.warn(TAG + methodName, "Execution still running, attempting to cancel.");
+                    // This does return a value, but it doesn't tell us much that's actionable.
+                    // It's true if the task was cancelled, but false could mean that it hadn't
+                    // started yet.
+                    currentInteractiveTask.cancel(true);
+                } catch (InterruptedException e) {
```
AdamBJohnsonx, line 556:
```suggestion
                } catch (final InterruptedException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 );
             }
 
-            sInteractiveExecutor.execute(new Runnable() {
+            // It is possible that the current task could be stuck irretrievably.  If that's
+            // what's happened, do everything in our power to make certain that it is dead.
+            Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
```
rpdome, line 541: Do we know when this actually happens? Repro steps?

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 );
             }
 
-            sInteractiveExecutor.execute(new Runnable() {
+            // It is possible that the current task could be stuck irretrievably.  If that's
```
rpdome, line 539: Do we know when this actually happens? Repro steps?

AdamBJohnsonx: When you create a bug in a command (that's how I got it).  You could write a command that does nothing but a thread.sleep, and that would do it.  The idea is that we shouldn't make the user close the software to recover, since we're trying to kill the running command anyway.

AdamBJohnsonx: Thing is that we're not be resilient to bugs in downstream software.

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -360,9 +360,24 @@ public static void info(final String tag, @Nullable final String message) {
...
+     * @param exception      A related exception, or null.
+     */
+    public static void info(final String tag,
+                            @Nullable final String correlationID,
+                            @Nullable final String message,
+                            @Nullable final Throwable exception) {
+        getInstance().log(tag, LogLevel.INFO, correlationID, message, exception, false);
+    }
```
shahzaibj, line 380: If we are doing this, perhaps do it for other levels as well such as WARN, and VERBOSE? Seems strange that throwable is only accepted on INFO and ERROR and not on VERBOSE and WARN

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -523,7 +536,30 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+            // what's happened, do everything in our power to make certain that it is dead.
+            Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
+            if (currentInteractiveTask != null && !(currentInteractiveTask.isDone() || currentInteractiveTask.isCancelled())) {
+                try {
+                    // We'll give it 1/2 second to respond to the kill message we broadcast.
+                    currentInteractiveTask.get(500, TimeUnit.MILLISECONDS);
```
shahzaibj, line 545: Who is sending this broadcast? from where?

AdamBJohnsonx: This method does, if the command is a broker command, on line 521.  This raises an interesting question - what do we think we do right now for users who re-entrantly submit interactive tasks? This could perhaps only be done for the broker, but right now it looks like all interactive tasks enqueue with no limit.

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
AdamBJohnsonx: > If we are doing this, perhaps do it for other levels as well such as WARN, and VERBOSE? Seems strange that throwable is only accepted on INFO and ERROR and not on VERBOSE and WARN

Is there some reason for this distinction?  This is completely optional, honestly, and if there is, I can drop this.

shahzaibj: I think the question is "why isn't it even optional for VERBOSE and WARN"? - seems for these we strictly don't support passing a throwable. I personally think it should be supported on all levels as an optional parameter

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -360,7 +360,14 @@ public static void info(final String tag, @Nullable final String message) {
...
                             @Nullable final String message) {
-        getInstance().log(tag, LogLevel.INFO, correlationID, message, null, false);
+        getInstance().log(
+                tag,
+                LogLevel.INFO,
+                correlationID,
+                message,
+                null,
+                false
+        );
```
AdamBJohnsonx, line 370:
```suggestion
        getInstance().log(tag, LogLevel.INFO, correlationID, message, null, false);
```

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
AdamBJohnsonx: Lets make that separate.

changelog.txt
```diff
@@ -9,6 +9,7 @@ V.Next
 - [MINOR] Device PoP keys are now generated with attestation flags and expose a certificate chain getter (#1247)
 - [PATCH] Fix MSAL-CPP injecting javascript loop (#1238)
 - [MINOR] Adds new API to support ADAL/MSAL migration off of AuthenticationSettings#setSecretKey to a 'managed key' generated by Common (#1244)
+- [PATCH] Track running interactive tasks and attempt to terminate if they do not (#1268)
```
AdamBJohnsonx, line 12:
```suggestion
- [PATCH] Track running interactive tasks in the broker and attempt to terminate them if they do not (#1268)
```

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,49 @@ public void run() {
...
+            @Override
+            public AcquireTokenResult execute() throws Exception {
+                integer.set(1);
+                latch.countDown();
+                return null;
+            }
+        });
+        Assert.assertTrue(latch.await(5, TimeUnit.SECONDS));
+        Assert.assertEquals(1, integer.get());
+
```
AdamBJohnsonx, line 412:
```suggestion

```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
 
-            sInteractiveExecutor.execute(new Runnable() {
+                // It is possible that the current task could be stuck irretrievably.  If that's
+                // what's happened, do everything in our power to make certain that it is dead.
+                Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
+                if (currentInteractiveTask != null && !(currentInteractiveTask.isDone() || currentInteractiveTask.isCancelled())) {
+                    try {
+                        // We'll give it 1/2 second to respond to the kill message we broadcast.
+                        currentInteractiveTask.get(500, TimeUnit.MILLISECONDS);
+                    } catch (ExecutionException e) {
```
AdamBJohnsonx, line 545:
```suggestion
                    } catch (final ExecutionException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                        // OK, it finished with an exception during that time.  Probably OK.
+                        Logger.info(TAG + methodName, null, "Previous task terminated with exception " + e.getMessage());
+                    } catch (TimeoutException e) {
```
AdamBJohnsonx, line 548:
```suggestion
                    } catch (final TimeoutException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                        // Nope, it's still going.  Send it a thread cancellation.
+                        Logger.warn(TAG + methodName, "Execution still running, attempting to cancel.");
+                        // This does return a value, but it doesn't tell us much that's actionable.
+                        // It's true if the task was cancelled, but false could mean that it hadn't
+                        // started yet.
+                        currentInteractiveTask.cancel(true);
+                    } catch (InterruptedException e) {
```
AdamBJohnsonx, line 555:
```suggestion
                    } catch (final InterruptedException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
                 localBroadcastManager.sendBroadcast(
                         new Intent(CANCEL_INTERACTIVE_REQUEST)
                 );
-            }
 
-            sInteractiveExecutor.execute(new Runnable() {
+                // It is possible that the current task could be stuck irretrievably.  If that's
+                // what's happened, do everything in our power to make certain that it is dead.
+                Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
```
AdamBJohnsonx, line 540:
```suggestion
                final Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
+                        // started yet.
+                        currentInteractiveTask.cancel(true);
+                    } catch (InterruptedException e) {
+                        // Something interrupted us.  Log and die.
+                        Logger.error(TAG + methodName, "Interrupted while running, bailing out", e);
+                        Thread.currentThread().interrupt();
+                    }
+                }
+            }
+            Future<?> task = sInteractiveExecutor.submit(new Runnable() {
```
AdamBJohnsonx, line 562:
```suggestion
            final Future<?> task = sInteractiveExecutor.submit(new Runnable() {
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -521,9 +534,32 @@ public static void beginInteractive(final InteractiveTokenCommand command) {
...
-            }
 
-            sInteractiveExecutor.execute(new Runnable() {
+                // It is possible that the current task could be stuck irretrievably.  If that's
+                // what's happened, do everything in our power to make certain that it is dead.
+                final Future<?> currentInteractiveTask = sCurrentInteractiveTask.get();
+                if (currentInteractiveTask != null && !(currentInteractiveTask.isDone() || currentInteractiveTask.isCancelled())) {
+                    try {
+                        // We'll give it 1/2 second to respond to the kill message we broadcast.
+                        currentInteractiveTask.get(500, TimeUnit.MILLISECONDS);
```
iambmelt, line 544: Possibly we put `500` on a named const like `MAX_WAIT` or similar

iambmelt (approved):

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,48 @@ public void run() {
...
+            }
+
+            @Override
+            public void onCancel() {
+
+            }
+        };
+
+        CommandDispatcher.beginInteractive(new InteractiveTokenCommand(BrokerInteractiveTokenCommandParameters.builder().androidApplicationContext(context).build(), (BaseController) null, callback, null) {
```
shahzaibj, line 395: This line is too long and very difficult to read here on GitHub, may be split it across multiple lines

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,48 @@ public void run() {
...
+            @Override
+            public AcquireTokenResult execute() throws Exception {
+                Thread.sleep(1000000);
```
shahzaibj, line 398: constant?

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,48 @@ public void run() {
...
+                return null;
+            }
+        });
+        CommandDispatcher.beginInteractive(new InteractiveTokenCommand(BrokerInteractiveTokenCommandParameters.builder().androidApplicationContext(context).build(), (BaseController) null, callback, null) {
```
shahzaibj, line 402: Long line....can be split to multiple lines

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,48 @@ public void run() {
         }
     }
 
+    @Test
+    public void testCanRecoverFromHangingTask() throws Exception {
+        final AtomicInteger integer = new AtomicInteger(0);
```
shahzaibj, line 374: Can we use a different name than `integer` as that may at first seem like the data type rather than the variable name

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -365,6 +369,48 @@ public void run() {
...
+        final CountDownLatch latch = new CountDownLatch(1);
+        Context context = InstrumentationRegistry.getInstrumentation().getContext();
```
shahzaibj, line 376:
```suggestion
        Context context = ApplicationProvider.getApplicationContext();
```

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
iambmelt: We should document somewhere what our threshold is... if not for ourselves, for external contributors. I usually say 100 chars. AOSP says the same, I think we should set that in 'stone' somewhere

https://source.android.com/setup/contribute/code-style#limit-line-length

Adam and I chatted on briefly a week or two ago and, while most (all?) of us have widescreen monitors, I usually do a lot of v-splitting in my editor and having nice narrow columns usually prevents me having the horizontally scroll. I am less insistent on this _for tests_, though we might want to codify this somewhere

As a random example (just opening some files), the left-pane goes off-screen just a bit in the below screenshot, hence my advocacy for approx. 100 chars
image

tanmaymanolkar1 (approved):

PR #1270, opened 2021-03-16
Try do do something reasonable when we can't connect to custom tabs svc

AdamBJohnsonx:
A customer reported that they had an issue with a percentage of requests
not ever returning a status.  We observed that this happened after there
were failures to register bind in the logs. One thing we could do would be
to try to use our fallback behavior as if the custom tabs service was 
unavailable.  It would also be nice to disambiguate the logs.

Hardcoding this value to false seems to work on my device/emulator.


shahzaibj (approved):
LGTM

PR #1277, opened 2021-03-19
Update dev after release

paulkagiri:


shahzaibj:
Let's use an intermediary branch to perform this merge. This means that instead of merging `master` directly into `dev`, instead create a new branch off of `master`, let's call it something like `paul/mergeMasterToDev` and then resolve the merge conflicts on this branch and then create a new PR to merge this branch into `dev`.

PR #1279, opened 2021-03-19
Logger moving out of internal package, staying back compat

iambmelt:
This PR elevates `Logger` to a non-internal API on AndroidCommon.

Rationale:
- ADAL, MSAL, and now Broker all call this API: keeping it in `internal` is both wrong and dangerous at this point, as API breaks stand to break those callers

This PR:
- Deprecates the internal classes
- Relocates the functionality to a new, public API class
- Preserves internal classes such that callers are not broken forcing a major version change in common, plus additional changes in ADAL, MSAL, and Broker


common/src/main/java/com/microsoft/identity/common/logging/Logger.java
```diff
@@ -0,0 +1,575 @@
...
+     *
+     * @param logLevel The {@link LogLevel} to be enabled for the diagnostic logging.
+     */
+    public void setLogLevel(final LogLevel logLevel) {
+        mLogLevel = logLevel;
+    }
+
+    /**
+     * Set the custom logger. Configures external logging to configure a callback that
+     * the sdk will use to pass each log message. Overriding the logger callback is not allowed.
```
shahzaibj, line 135: Do you mean overriding this method? Or resetting the external logger? Looking at the code it seems that both are actually allowed

iambmelt: This is the exact Logger implementation from common, all I did was move the implementation

iambmelt: From internal. There's no new code here

common/src/main/java/com/microsoft/identity/common/logging/Logger.java
```diff
@@ -0,0 +1,575 @@
...
+                //Format the log message.
+                final String logMessage = formatMessage(correlationID, message, dateTimeStamp, throwable);
+
+                // Send logs into Logcat.
+                if (sAllowLogcat) {
+                    sendLogcatLogs(tag, logLevel, logMessage);
+                }
+
+                // Send logs into external logger callback.
+                final ILoggerCallback externalLogger = mExternalLogger;
```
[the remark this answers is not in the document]
iambmelt: Same as below. I've moved the file. THis isn't new impl

common/src/main/java/com/microsoft/identity/common/logging/Logger.java
shahzaibj: I see - so the original documentation seems wrong then because code clearly allows overriding the method as well as resetting the logger. I think this is something that we need to think about and decide whether we actually want to allow resetting the logger. For apps that consume multiple of our libraries such as ad-accounts and MSAL and they still use the old Logger API from WPJ and MSAL to register their respective external logger, it seems that these loggers would now overwrite each other.....should we allow that?

shahzaibj: This of course assuming that we are delegating WPJ logger to common's logger

AdamBJohnsonx: This is an interesting question - how many applications is that?  I think it's 3 - Authenticator, CP, and BrokerHost.  But there may be more (teams, etc)... we should enumerate these problems.

shahzaibj: If ADAL's also going to use these then that would include basically any app that dual stacks ADAL and MSAL

iambmelt:
Update changelog

common/src/main/java/com/microsoft/identity/common/internal/logging/RequestContext.java
```diff
@@ -22,18 +22,7 @@
...
-import java.util.HashMap;
-
-// TODO I'm not wedded to this name, but the concept may work for tracking correlationIds
-public class RequestContext extends HashMap<String, String> implements IRequestContext {
-
-    private static final long serialVersionUID = -1128593786133166935L;
-    private final Gson mGson = new Gson();
-
-    @Override
-    public String toJsonString() {
```
AdamBJohnsonx, line 36: You might still need this method to exist.

common/src/main/java/com/microsoft/identity/common/internal/logging/DiagnosticContext.java
```diff
@@ -22,74 +22,7 @@
...
-    /**
-     * Clear the local request context thread.
-     */
-    public static void clear() {
-        REQUEST_CONTEXT_THREAD_LOCAL.remove();
-    }
-
+@Deprecated
+public final class DiagnosticContext extends com.microsoft.identity.common.logging.DiagnosticContext {
+    // Class exists for backcompat - originally internal only
```
AdamBJohnsonx, line 27: You probably want implementations of all the public methods that 1-shot errors into the logs

common/src/main/java/com/microsoft/identity/common/logging/RequestContext.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.google.gson.Gson;
+import com.microsoft.identity.common.internal.logging.IRequestContext;
+
+import java.util.HashMap;
+
+public class RequestContext extends HashMap<String, String> implements IRequestContext {
+
+    private static final long serialVersionUID = -2239604897244277047L;
+    private final Gson mGson = new Gson();
```
AdamBJohnsonx, line 33: static

AdamBJohnsonx (approved):
I don't have any real problems here.  We might want to be nicer about the calls we expose, but we can take that later.  We should probably have 3 people sign off on this, though.

common/src/main/java/com/microsoft/identity/common/internal/logging/ILoggerCallback.java
```diff
@@ -36,4 +39,4 @@
      * @param containsPII True if the log message contains PII, false otherwise.
      */
     void log(String tag, Logger.LogLevel logLevel, String message, boolean containsPII);
-}
+}
```
shahzaibj, line 42: can we add back new line ending?

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -76,500 +55,60 @@
...
-                break;
-
+                return com.microsoft.identity.common.logging.Logger.LogLevel.VERBOSE;
             default:
-                throw new IllegalArgumentException("Unknown log level");
+                throw new RuntimeException("Unknown or invalid log level");
         }
     }
+
 }
```
shahzaibj, line 114: new line ending

common/src/main/java/com/microsoft/identity/common/internal/logging/RequestContext.java
iambmelt: How come?

common/src/main/java/com/microsoft/identity/common/logging/RequestContext.java
iambmelt: 33ba0948b540381129f4ce6e71f6e236f3d89c1b

common/src/main/java/com/microsoft/identity/common/logging/Logger.java
```diff
@@ -0,0 +1,575 @@
...
+
+    private static String getUTCDateTimeAsString() {
+        final SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT, Locale.getDefault());
+        dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+
+        return dateFormat.format(new Date());
+    }
+
+    /**
+     * Send logs to logcat as the default logging if developer doesn't turn off the logcat logging.
```
shahzaibj, line 550: I think this just send logs to logcat regardless of whether developer turns it on or off. This method doesn't seem to care....it seems the caller needs to take care of that...we can probably update this javadoc

common/src/main/java/com/microsoft/identity/common/internal/logging/ILoggerCallback.java
iambmelt: whoops

1f6548afc07d15e6fecbc1765f03de535febfe84

shahzaibj (approved):
Left minor comments, else LGTM

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
iambmelt: dff3c9d1ea274e7877c5d6c74caa86aa5920fc3b

common/src/main/java/com/microsoft/identity/common/logging/Logger.java
iambmelt: 0ffd5d6085fa623073dfcebefa57259473648e65

iambmelt:
Update changelog

common/src/main/java/com/microsoft/identity/common/internal/logging/RequestContext.java
rpdome: This one was moved to the base class right?

common/src/main/java/com/microsoft/identity/common/internal/logging/ILoggerCallback.java
```diff
@@ -25,7 +25,10 @@
 /**
  * Interface for apps to configure the external logging and implement the callback to designate the
  * output of the log messages.
+ *
+ * This class is deprecated. Use {@link com.microsoft.identity.common.logging.ILoggerCallback}.
  */
+@Deprecated
 public interface ILoggerCallback {
```
rpdome, line 32: nit: also extends the new ILoggerCallback?

rpdome (approved):

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/logging/ILoggerCallback.java
iambmelt: That would be a breaking change (since the new method from the parent class would be pushed onto the interface)

common/src/main/java/com/microsoft/identity/common/internal/logging/RequestContext.java
iambmelt: That's right - the method will be accessible from the base class (and accordingly, also available on the subclass)

iambmelt:
Related broker PR to pick up these changes
https://github.com/AzureAD/ad-accounts-for-android/pull/1514

common/src/main/java/com/microsoft/identity/common/internal/logging/RequestContext.java
AdamBJohnsonx: I missed the extends clause.  We might want to log, though.

iambmelt: e27764e2e1c9dcf20bd4d009e31c2c396159f33e

tanmaymanolkar1 (approved):

AdamBJohnsonx (approved):

PR #1280, opened 2021-03-19
Make it easier to get the token authority from the response.

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/TokenResponse.java
```diff
@@ -111,6 +111,23 @@
     @SerializedName("id_token")
     private String mIdToken;
 
+    private transient String mTokenAuthority;
+
+    /**
+     * Get the authority that issued this token.
+     * @return the authority that issued this token.
+     */
+    public String getAuthority() {
```
shahzaibj, line 120: Are there usages as of now?

common/src/main/java/com/microsoft/identity/common/exception/ClientException.java
```diff
@@ -27,6 +27,11 @@
     public static final String sName = ClientException.class.getName();
     private static final long serialVersionUID = -2318746536590284648L;
 
+    /**
+     * Indicates that a token request returned an unexpected authority.
+     */
+    public static final String UNEXPECTED_AUTHORITY = "unexpected_authority_token_request";
```
shahzaibj, line 33: Are there usages as of now?

AdamBJohnsonx: This shouldn't be here.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/TokenResponse.java
AdamBJohnsonx: No, but this is an attempt to replace some of the circumlocution that we're going through to figure out authorities.

PR #1282, opened 2021-03-21
Support null task affinity for activities used to launch interactive token requests

shoatman:
When authentication is completed interactive token requests that were started from an activity which do not have a taskAffinity (null taskAffinity) that activity is not brought to the foreground after the interaction with the authorization endpoint is complete.  This is a problem for some apps that leverage null taskaffinity in multi-window scenarios.


shoatman:
This needs changelog information as well as likely a documentation update for configuration as well.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationStrategy.java
```diff
@@ -50,6 +51,7 @@
     private WeakReference<Context> mReferencedApplicationContext;
     private WeakReference<Activity> mReferencedActivity;
     private WeakReference<Fragment> mReferencedFragment;
+    protected int mTaskIdThatStartedAuthorization;
```
shoatman, line 54: Whoops i didn't end up using this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -429,6 +431,38 @@ public void run() {
         });
     }
 
+
+    /**
+     * This method optionally re-orders that task to bring the task that launched
+     * the interactive activity to the foreground.  This is useful when the activity provided
+     * to us does not have a taskAffinity and as a result it's possible that other apps or the home
+     * screen could be in the task stack ahead of the app that launched the interacxtive
```
shoatman, line 439: typo

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,50 @@
...
+        final PackageManager packageManager = applicationContext.getPackageManager();
+        try {
+            ActivityInfo startActivityInfo = packageManager.getActivityInfo(parameters.getActivity().getComponentName(), 0);
+            if(startActivityInfo.taskAffinity == null){
+                mHasTaskAffinity = false;
+                mTaskId = parameters.getActivity().getTaskId();
+            }
+        } catch (PackageManager.NameNotFoundException e) {
+            //TODO: Log properly and then ignore
+            e.printStackTrace();
```
shoatman, line 77: Add correct logging.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 V.Next
 ----------
 - [MINOR] Enables removeAccount api to remove account records from all environments (#1248)
+- [MINOR] Added support for handling null taskAffinity.  InteractiveTokenCommmand now captures whether taskAFfinity is null and records taskId
```
iambmelt, line 4:
```suggestion
- [MINOR] Added support for handling null taskAffinity.  InteractiveTokenCommmand now captures whether taskAFfinity is null and records taskId (#1282)
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationStrategy.java
```diff
@@ -1,3 +1,4 @@
+
```
iambmelt, line 1: stray newline

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,53 @@
...
 
     public InteractiveTokenCommand(@NonNull InteractiveTokenCommandParameters parameters,
                                    @NonNull List<BaseController> controllers,
                                    @SuppressWarnings(WarningType.rawtype_warning) @NonNull CommandCallback callback,
                                    @NonNull final String publicApiId) {
         super(parameters, controllers, callback, publicApiId);
+        checkAndRecordTaskInformation(parameters);
     }
 
+    private void checkAndRecordTaskInformation(InteractiveTokenCommandParameters parameters){
```
iambmelt, line 66:
```suggestion
    private void checkAndRecordTaskInformation(@NonNull final InteractiveTokenCommandParameters parameters){
```

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,53 @@
...
+        final String methodName = ":checkAndRecordTaskInformation";
+        final Context applicationContext = parameters.getAndroidApplicationContext();
+        final PackageManager packageManager = applicationContext.getPackageManager();
+        try {
+            ActivityInfo startActivityInfo = packageManager.getActivityInfo(parameters.getActivity().getComponentName(), 0);
```
iambmelt, line 71: IIRC, OneAuth added some scenarios where the calling `Activity` might be `null`, we may want to be defensive here and check it's `!= null` before requesting the `componentName`

/cc @irving-as

iambmelt, line 71:
```suggestion
            final ActivityInfo startActivityInfo = packageManager.getActivityInfo(parameters.getActivity().getComponentName(), 0);
```

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,53 @@
...
+            if(startActivityInfo.taskAffinity == null){
+                mHasTaskAffinity = false;
+                mTaskId = parameters.getActivity().getTaskId();
+            }
+        } catch (PackageManager.NameNotFoundException e) {
```
iambmelt, line 76:
```suggestion
        } catch (final PackageManager.NameNotFoundException e) {
```

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
shoatman: Not sure how that works... when would activity be null here?

iambmelt: @irving-as Something you can comment on? As I recall, there is behavior where the broker / interactive flows will be invoked using only an `Activity` context

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,53 @@
...
+            }
+        } catch (final PackageManager.NameNotFoundException e) {
+            Logger.warn(
+                    TAG + methodName,
+                    "Unable to get ActivityInfo for activity provided to start authorization."
+            );
+        }
+    }
+
+    public boolean getHasTaskAffinity(){
```
AdamBJohnsonx, line 84:
```suggestion
    public boolean hasTaskAffinity(){
```

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
```diff
@@ -39,21 +43,53 @@
 @EqualsAndHashCode(callSuper = true)
 public class InteractiveTokenCommand extends TokenCommand {
     private static final String TAG = InteractiveTokenCommand.class.getSimpleName();
+    //Normally all tasks have an affinity unless configured explicitly for multi-window support to not have one
+    private boolean mHasTaskAffinity = true;
+    private int mTaskId = 0;
```
AdamBJohnsonx, line 48: nit:

@Getter 
@Accessors(prefix = "m")

?

shoatman: oh boy... let's take this up with Brian and the rest of the team.  I prefer not to do this myself... however it's part of the android naming guidelines

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
shoatman: To be honest I don't care what the standard is... as long as we're all following it.  Unfortunately we have a diversity of things now in the code.

common/src/main/java/com/microsoft/identity/common/internal/commands/InteractiveTokenCommand.java
AdamBJohnsonx: I think you misunderstand.  Putting those two annotations on this class will create the accessors understanding that the member prefix is "m", and name the methods without it.

AdamBJohnsonx (approved):
I don't have problems here.

PR #1285, opened 2021-03-23
Add the ability to access key managers

AdamBJohnsonx:
While working with this code, I discovered that I'd inadvertently
made it difficult to remove keys from the keystore when you were
finished with them.  You could still do it, but you'd need to create
your own separate key manager to make it happen.  It's probably
OK to save the reference to the original manager that created it,
so lets just provide access to accessors that are created in ways
that need cleanup.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,7 +2,7 @@ V.Next
 ----------
 - [MINOR] Hide Switch Account in Broker interactive flows (#1284)
 - [MINOR] Enables removeAccount api to remove account records from all environments (#1248)
-- [MINOR] Adds the ability for KeyAccessors to expose their manager.
+- [MINOR] Adds the ability for KeyAccessors to expose their manager (#1285)
 
 Version 3.2.0
 ----------
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@ V.Next
 ----------
 - [MINOR] Hide Switch Account in Broker interactive flows (#1284)
 - [MINOR] Enables removeAccount api to remove account records from all environments (#1248)
+- [MINOR] Adds the ability for KeyAccessors to expose their manager.
 
 Version 3.2.0
 ----------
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -45,6 +48,11 @@
      */
     public static final String ENCODING_UTF8 = "UTF-8";
 
+    /**
+     * The Constant CHARSET_UTF8.
+     */
+    public static final Charset CHARSET_UTF8 = StandardCharsets.UTF_8; //Charset.forName("UTF-8");
```
iambmelt, line 54: Is our `minSdk` high enough for this?

AdamBJohnsonx: You're right, I mean to leave that out (hence the comment).  If this build passes, we have to figure it out.

changelog.txt
```diff
@@ -2,6 +2,7 @@ V.Next
 ----------
 - [MINOR] Hide Switch Account in Broker interactive flows (#1284)
 - [MINOR] Enables removeAccount api to remove account records from all environments (#1248)
+- [MINOR] Adds the ability for KeyAccessors to expose their manager.
```
iambmelt, line 5:
```suggestion
- [MINOR] Adds the ability for KeyAccessors to expose their manager (#1285)
```

iambmelt (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx, line 54:
```suggestion
    public static final Charset CHARSET_UTF8 = Charset.forName("UTF-8");
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -22,6 +22,9 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.adal.internal;
 
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
```
AdamBJohnsonx, line 26:
```suggestion
```

PR #1287, opened 2021-03-23
Revert change to not pass environment as part of removeAccount bundle

iamgusain:
Reverting the change to not pass  "environment" as part of remove Account bundle request. As this will result in breaking remove account scenarios because the current broker code rejects the request as invalid if environment is passed as null.

Making a separate change in broker to remove the null environment check and also to override the environment value to null [PR 1507](https://github.com/AzureAD/ad-accounts-for-android/pull/1507/files). This will ensure that older versions of broker will continue to work as-is and newer versions of broker will support removing account across all environments.


rpdome (approved):

rpdome (changes requested):
I ran into an issue with this commit, even with this change. (environment still null)

Let's revert the whole PR.

rpdome (approved):
Edit: synced offline. this also needs a change in MSAL.

PR #1288, opened 2021-03-24
change to nullable

rpdome:
Since we move to Lombok, it's now strongly enforced and it starts throwing exception in silent flow. (that flow we shouldn't need authorizationAgent - and we never set one)


AdamBJohnsonx (approved):
As a suggestion, can we add a test that makes sure that the params we think should be nullable are, and the ones we don’t throw an exception?

PR #1290, opened 2021-03-24
Add new broker API support, adopt constant for UTF-8

AdamBJohnsonx:
We're looking at a few new APIs for Broker, and the way it's structured, we need to change common as well.  We can go ahead and put that support in place, doing no harm.

These are supporting four new APIs, three of which are actually implemented in https://github.com/AzureAD/ad-accounts-for-android/pull/1518 

These are getFlights, setFlights, and addFlights.  The thinking is that we allow our caller to change our runtime behavior based on what they know - iff they're the broker host.  There is an experiment system that is in place that we could ask users to integrate with, or they could use other methods to turn on and off broker features in pre-release from.

The interface to this is just a map of string to string, with getFlights returning the map, setFlights overwriting it, and addFlights overlaying its arguments onto it.  In the implementing review, we could adopt other returns, such as including the resulting map into get and setFlights.

New APIs - flight support from broker host apps: setFlights (take a map of id to status and overwrite the current flight information), addFlights (overlay the map onto the existing structure), and getFlights(return the map).  As well as the get SSO token api, which will get the PRTv3 SSO token.

Additionally, I put in a constant for the UTF-8 charset.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1268,7 +1268,7 @@ public final class AuthenticationConstants {
         public static final String SET_FLIGHTS_SUCCEEDED = "set_flights_succeeded";
 
         /**
-         * Boolean to return when a Broker RT is successfully updated.
+         * All of the active flights.
          */
         public static final String GET_FLIGHTS_RESULT = "active_flights";
 
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -23,7 +23,6 @@
 package com.microsoft.identity.common.adal.internal;
 
 import java.nio.charset.Charset;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1425,7 +1425,7 @@ public final class AuthenticationConstants {
                 return mPath;
             }
             public int code() {
-                return mCode == null ? ordinal() + 1: mCode;
+                return mCode == null ? ordinal() + 1 : mCode;
             }
         }
         /**
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1409,7 +1409,8 @@ public final class AuthenticationConstants {
             BROKER_ADD_FLIGHTS(BROKER_API_ADD_FLIGHTS_PATH),
             BROKER_SET_FLIGHTS(BROKER_API_SET_FLIGHTS_PATH),
             BROKER_GET_FLIGHTS(BROKER_API_GET_FLIGHTS_PATH),
-            GET_SSO_TOKEN(GET_SSO_TOKEN_PATH)
+            GET_SSO_TOKEN(GET_SSO_TOKEN_PATH),
+            UNKNOWN(null)
                 ;
             private String mPath;
             private Integer mCode;
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.adal.internal;
 
 import java.nio.charset.Charset;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
@@ -828,6 +829,11 @@ public final class AuthenticationConstants {
         /**
          * String of key for account name.
          */
+        public static final String FLIGHT_INFO = "com.microsoft.identity.broker.flights";
+
+        /**
+         * String of key for account name.
+         */
         public static final String ACCOUNT_NAME = "account.name";
 
         /**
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -67,7 +67,11 @@
         BROKER_API_HELLO,
         BROKER_API_GET_BROKER_ACCOUNTS,
         BROKER_API_REMOVE_BROKER_ACCOUNT,
-        BROKER_API_UPDATE_BRT
+        BROKER_API_UPDATE_BRT,
+        BROKER_SET_FLIGHTS,
+        BROKER_GET_FLIGHTS,
+        BROKER_ADD_FLIGHTS,
+        BROKER_SSO_TOKEN;
```
AdamBJohnsonx, line 74: Here, I don't know what to call this.  It could be MSAL, but we're going to protect it, and only expect to use it for Edge.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/OAuth2Strategy.java
```diff
@@ -225,6 +226,10 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         return response;
     }
 
+    protected String getRequestBody(GenericTokenRequest request) throws UnsupportedEncodingException, ClientException {
```
AdamBJohnsonx, line 229:
```suggestion
    protected String getRequestBody(final GenericTokenRequest request) throws UnsupportedEncodingException, ClientException {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.adal.internal;
 
 import java.nio.charset.Charset;
+import java.util.concurrent.atomic.AtomicInteger;
```
AdamBJohnsonx, line 26:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1373,6 +1389,46 @@
...
+            }
+            API(String path) {
+                this.mPath = path;
+                this.mCode = null;
+            }
+            public String path() {
+                return mPath;
+            }
+            public int code() {
+                return mCode == null ? ordinal() + 1: mCode;
```
AdamBJohnsonx, line 1428:
```suggestion
                return mCode == null ? ordinal() + 1 : mCode;
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1257,6 +1262,16 @@
          */
         public static final String UPDATE_BROKER_RT_SUCCEEDED = "update_broker_rt_succeeded";
 
+        /**
+         * Boolean to return when a Broker RT is successfully updated.
```
tanmaymanolkar1, line 1266: same as `GET_FLIGHTS_RESULT`

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1257,6 +1262,16 @@
...
+         */
+        public static final String SET_FLIGHTS_SUCCEEDED = "set_flights_succeeded";
+
+        /**
+         * Boolean to return when a Broker RT is successfully updated.
```
AdamBJohnsonx, line 1271:
```suggestion
         * All of the active flights.
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx: Yup.

iambmelt (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -54,20 +55,30 @@
...
+        MSAL_GENERATE_SHR(API.GENERATE_SHR, BrokerAccountManagerOperation.GENERATE_SHR),
+        BROKER_GET_KEY_FROM_INACTIVE_BROKER(null, null),
+        BROKER_API_HELLO(API.BROKER_HELLO, null),
+        BROKER_API_GET_BROKER_ACCOUNTS(API.BROKER_GET_ACCOUNTS, null),
+        BROKER_API_REMOVE_BROKER_ACCOUNT(API.BROKER_REMOVE_ACCOUNT, null),
+        BROKER_API_UPDATE_BRT(API.BROKER_UPDATE_BRT, null),
+        BROKER_SET_FLIGHTS(API.BROKER_SET_FLIGHTS, null),
+        BROKER_GET_FLIGHTS(API.BROKER_GET_FLIGHTS, null),
+        BROKER_ADD_FLIGHTS(API.BROKER_ADD_FLIGHTS, null),
+        BROKER_SSO_TOKEN(API.GET_SSO_TOKEN, null);
```
AdamBJohnsonx, line 75:
```suggestion
        MSAL_SSO_TOKEN(API.GET_SSO_TOKEN, null);
```

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
AdamBJohnsonx: I'm going to make it MSAL_SSO_TOKEN, since we're exposing to to MSAL-Cpp.  I'm not sure that this API is cut in quite the right place.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1438,6 +1494,26 @@
          */
         public static final String BROKER_API_UPDATE_BRT_PATH = "/brokerApi/updateBrt";
 
+        /**
+         * Broker api path constant for adding flight information.
+         */
+        public static final String BROKER_API_ADD_FLIGHTS_PATH = "/brokerApi/addFlights";
+
+        /**
+         * Broker api path constant for adding flight information.
```
rpdome, line 1503: @AdamBJohnsonx nit: typo.

PR #1292, opened 2021-03-25
Manipulate the object mapper to provide a mechanism to propagate unknown parameters

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/MicrosoftTokenResponse.java
```diff
@@ -27,6 +27,7 @@
 import com.microsoft.identity.common.internal.providers.oauth2.TokenResponse;
 
 import java.util.Date;
+import java.util.Map;
```
AdamBJohnsonx, line 30:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -47,6 +58,124 @@
      */
     public static final String ENCODING_SCHEME = "UTF-8";
     public static final String TAG = ObjectMapper.class.getSimpleName();
+    public static final Gson GSON = new GsonBuilder()
+            .registerTypeAdapterFactory(new UnknownParamTypeAdapterFactory())
+            .create();
+
+    public static class UnknownParamTypeAdapterFactory implements TypeAdapterFactory {
```
AdamBJohnsonx, line 65:
```suggestion
    /*
     *This likely deserves a comment.  What I'm doing here is hooking the underlying GSON implementation's
     * use of JsonReader.  I KNOW that what's happening in the type adapter that we're delegating to is that if the
     * name of the field isn't in the names that they mapped, they skip it.  What we're doing instead is to keep
     * track of the last seen name, and then when we get called to skip a value, if it will be a string, save it in a map.
     * This map is linked to preserve the order of the parameters for testing use.  At the end of this process, we
     * store the resulting Iterable in the object that can accept it.
     */
    public static class UnknownParamTypeAdapterFactory implements TypeAdapterFactory {
```

changelog.txt
```diff
@@ -3,6 +3,7 @@ V.Next
 - [MINOR] Hide Switch Account in Broker interactive flows (#1284)
 - [MINOR] Enables removeAccount api to remove account records from all environments (#1248)
 - [MINOR] Adds the ability for KeyAccessors to expose their manager (#1285)
+- [MINOR] Propagate unknown parameters from the server
```
AdamBJohnsonx, line 6:
```suggestion
- [MINOR] Propagate unknown parameters from the server (#1292)
```

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -84,7 +224,11 @@ public static String serializeExposedFieldsOfObjectToJsonString(Object object) {
      * @return object
      */
     public static <T> T deserializeJsonStringToObject(String json, Class<T> objectClass) {
-        return new Gson().fromJson(json, objectClass);
+        T value = GSON.fromJson(json, objectClass);
+        if (IHasExtraParameters.class.isAssignableFrom(objectClass)) {
+
```
iambmelt, line 229: empty block?

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -84,7 +224,11 @@ public static String serializeExposedFieldsOfObjectToJsonString(Object object) {
...
+        }
```
AdamBJohnsonx, line 230:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
AdamBJohnsonx: Remnants of a different attempt

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -84,7 +224,8 @@ public static String serializeExposedFieldsOfObjectToJsonString(Object object) {
      * @return object
      */
     public static <T> T deserializeJsonStringToObject(String json, Class<T> objectClass) {
-        return new Gson().fromJson(json, objectClass);
+        T value = GSON.fromJson(json, objectClass);
+        return value;
```
AdamBJohnsonx, line 228:
```suggestion
        return GSON.fromJson(json, objectClass);
```

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
```diff
@@ -47,6 +58,135 @@
...
+                            public String toString() {
+                                return in.toString();
+                            }
+
+                            public String getPath() {
+                                return in.getPath();
+                            }
+                        };
+                        T output = adapter.read(jsonReader);
+                        ((IHasExtraParameters) output).setExtraParameters(Collections.unmodifiableMap(otherKeys).entrySet());
```
iambmelt, line 182: Would it make sense to use a bounded wildcard on the input type param to make sure this cast is going to work? Not sure my syntax is perfect here but something like

```java
TypeToken<T extends IHasExtraParameters> type
```

iambmelt:
Mentioned on Teams, posting here if it gets buried:
> This looks pretty similar (conceptually, anyway) with some JSON parsing going on [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java#L211-L251) - conceivably we could merge these implementations together, picking whichever approach we like best.

We might consider taking a backlog item to combine these implementations

common/src/test/java/com/microsoft/identity/common/unit/ObjectMapperTest.java
```diff
@@ -70,4 +74,26 @@ public void test_JsonToObject() {
...
+
+        Assert.assertEquals("idtokenval", tr.getIdToken());
+        final Iterator<Map.Entry<String, String>> iterator = tr.getExtraParameters().iterator();
+        Map.Entry<String, String> param = iterator.next();
+        Assert.assertEquals("client_id", param.getKey());
+        Assert.assertEquals(CLIENT_ID, param.getValue());
+        param = iterator.next();
+        Assert.assertEquals("other_param", param.getKey());
+        Assert.assertEquals("other_value", param.getValue());
+    }
```
iambmelt, line 98: Negative test? Maybe some malformed JSON? Duplicated fields?

common/src/main/java/com/microsoft/identity/common/internal/net/ObjectMapper.java
AdamBJohnsonx: We should only be in this block:

if (IHasExtraParameters.class.isAssignableFrom(type.getRawType())) {

So unless json breaks the contract...

common/src/test/java/com/microsoft/identity/common/unit/ObjectMapperTest.java
AdamBJohnsonx: Yeah.  Other types, too.

AdamBJohnsonx:
> Mentioned on Teams, posting here if it gets buried:
> 
> > This looks pretty similar (conceptually, anyway) with some JSON parsing going on [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/CacheKeyValueDelegate.java#L211-L251) - conceivably we could merge these implementations together, picking whichever approach we like best.
> 
> We might consider taking a backlog item to combine these implementations

https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1342425

iambmelt (approved):

PR #1294, opened 2021-03-29
Enable optional refresh token in cache

rpdome:
This is part of the "Get PRT from PRT->AT" work - which will be disabled as of now due to the FoCI complication. (I'll put it behind a boolean, so once we resolve foci issue, we can easily turn this on)

Once we're getting back PRT, we'll remove that from the token response and save the result without RT.

I also remove the prefix 'm' from CacheRecord as part of this PR.


AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/CacheRecord.java
```diff
@@ -48,7 +50,7 @@ public CacheRecordBuilder mAccount(final @NonNull AccountRecord account) {
             if (account == null) {
                 throw new NullPointerException("The account record for a CacheRecord may not be null");
```
AdamBJohnsonx, line 51: I think you could write a lombok non-null annotation on the variable, and then the builder will respect it.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -257,8 +256,12 @@ ICacheRecord save(final @NonNull AccountRecord accountRecord,
             throw new ClientException(CREDENTIAL_IS_SCHEMA_NONCOMPLIANT, "[(AT)]");
         }
 
-        if (!isRefreshTokenValid) {
-            throw new ClientException(CREDENTIAL_IS_SCHEMA_NONCOMPLIANT, "[(RT)]");
+        if (refreshTokenRecord != null) {
+            final boolean isRefreshTokenValid = isRefreshTokenSchemaCompliant(refreshTokenRecord);
```
AdamBJohnsonx, line 260: nit: inline?

AdamBJohnsonx (approved):
minor comments

PR #1295, opened 2021-03-29
Move JweResponse, add new class

AdamBJohnsonx:


common4j/src/main/com/microsoft/identity/common/internal/platform/SecretKeyAccessor.java
```diff
@@ -49,37 +35,40 @@
...
+    @Override
+    public byte[] encrypt(@NonNull final byte[] plaintext, @Nullable final Object... args) throws ClientException {
         final String errCode;
         final Exception exception;
         try {
             final KeyStore.SecretKeyEntry entry = mKeyManager.getEntry();
             final SecretKey key = entry.getSecretKey();
-            final Cipher c = Cipher.getInstance(suite.cipher().name());
+            final Cipher c = Cipher.getInstance(mSuite.cipher().name());
+            byte[] iv = new byte[12];
```
AdamBJohnsonx, line 68:
```suggestion

            // The iv length should move into the CipherSuite class
            byte[] iv = new byte[12];
```

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -334,4 +344,34 @@ public void remove(String key) {
     public @NonNull IPlatformUtil getPlatformUtil() {
         return new AndroidPlatformUtil(mContext, mActivity);
     }
+
+    @Override
+    public IKeyStoreAccessor getKeyStore() {
```
rpdome, line 349: fyi: @shahzaibj, this means there'll be an extra work to get PRTv3 to work on linux. Either we implement this, or migrate this logic to use what we already have.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidKeyStoreAccessor.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class AndroidKeyStoreAccessor extends KeyStoreAccessor {
+    protected static final int KEY_PURPOSES =  KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT | KeyProperties.PURPOSE_SIGN ;
+
+    public AndroidKeyStoreAccessor(IPlatformComponents components) {
+        super("AndroidKeyStore", components);
+    }
+
+    @Override
+    public IKeyAccessor newInstance(@NonNull final CryptoSuite cipher, @NonNull final boolean needRawAccess)
```
rpdome, line 36: Note to self: merge IKeyAccessor and IEntryKeyAccessor and the ket generator/usage pattern - to make it linux-compatible.

common4j/src/main/com/microsoft/identity/common/java/crypto/CryptoSuite.java
```diff
@@ -58,4 +61,14 @@
      * @return the signing algorithm desired by this suite.
      */
     SigningAlgorithm signingAlgorithm();
+
+    /**
+     * @return an AlgorithmParameterSpec if needed, null if not required.
+     */
+    AlgorithmParameterSpec cryptoSpec(Object... args);
```
rpdome, line 81: Per our conversation, this will be changed to Map, right? 

If you couldn't get that in in time, please add javadoc on the expected arguments for each implementation / methods that takes Object ... as params. (not just this one)- that would really help.

AdamBJohnsonx: I've put a lot of documentation in.

PR #1296, opened 2021-03-29
Change some versions to try to fix the build

AdamBJohnsonx:
Based on my experimentation so far, this change seems to unblock my local builds.  Corresponding changes in broker, and ADAL and android-common exist


AdamBJohnsonx:
It *just* broke.

PR #1299, opened 2021-03-31
Delete the getMockActivity method

AdamBJohnsonx:
What I was trying to do with activity in the previous commit does not work.  
Seems like the best path just delete the method and do it yourself in the 5 places
that we use these two lines of code, rather than continue to struggle with dependencies.


rpdome (approved):

shahzaibj:
Why not just use Robolectric instead?

shahzaibj (approved):

AdamBJohnsonx:
> Why not just use Robolectric instead?

We talked offline.  Strange errors out of the tests.  If we want to try it later, we can research it.

PR #1300, opened 2021-04-01
Add common-java. Migrating Logger.

rpdome:
- Added a java-only common4j project.
- android common now consumes common4j's source.
- common4j comes with a java-only tests that will also work as a sanity check, to make sure that no android code is accidentally put in there.
- Migrated Logger to common4j


common4j/src/main/com/microsoft/identity/common4j/external/logging/Logger.java
```diff
@@ -0,0 +1,447 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common4j.external.logging;
```
shahzaibj, line 23: Is there a reason we have to be inside `external`?

common4j/src/main/com/microsoft/identity/common4j/util/StringUtil.java
```diff
@@ -0,0 +1,29 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common4j.util;
+
+public class StringUtil {
+    public static boolean isNullOrEmpty(String message) {
```
shahzaibj, line 26: Nullablity annotations and final?

common4j/src/main/com/microsoft/identity/common4j/external/logging/IDetailedLoggerCallback.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.identity.common4j.external.logging;
+
+/**
+ * An extension of ILoggerCallback - so that it prints the discarded log.
+ * This is for testing only (to verify that logs are actually discarded).
+ */
+interface IDetailedLoggerCallback extends ILoggerCallback {
+    /**
+     * Messages that are discarded by the loggers.
```
shahzaibj, line 9: When would log messages be discarded and why?

common4j/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
shahzaibj, line 1: Does this need to be checked in? probably not

common4j/gradlew.bat
```diff
@@ -0,0 +1,89 @@
+@rem
```
shahzaibj, line 1: Does this need to be checked in? probably not

common/src/main/java/com/microsoft/identity/common/internal/logging/IRequestContext.java
```diff
@@ -23,6 +23,6 @@
 package com.microsoft.identity.common.internal.logging;
 
 @Deprecated
-public interface IRequestContext extends com.microsoft.identity.common.logging.IRequestContext {
+public interface IRequestContext extends com.microsoft.identity.common4j.external.logging.IRequestContext {
```
shoatman, line 26: Not sure how i feel about the package name.  com.microsoft.identity.common.java instead?  Also is this the opportunity for the android library and the existing MSAL for Java library to start sharing code?

common/src/main/java/com/microsoft/identity/common/logging/DiagnosticContext.java
```diff
@@ -22,39 +22,21 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.logging;
 
-public class DiagnosticContext {
+import com.microsoft.identity.common4j.external.logging.IRequestContext;
```
shoatman, line 25: What does the external package name signify?

shoatman (commented):
Left a few questions

common4j/src/main/com/microsoft/identity/common4j/util/StringUtil.java
rpdome: There's no Nullable in Lombok's annotation, just NonNull.

common/src/main/java/com/microsoft/identity/common/logging/DiagnosticContext.java
rpdome: External means it will be consumed externally (i.e. by android packages)

common/src/main/java/com/microsoft/identity/common/internal/logging/IRequestContext.java
rpdome: 1. That's the first name I picked 😆. Then I was suggested that common4j might be a better name (since we have msal4j). I don't have any strong preference over this.

2. I'd say that it's not our goal right now. We're not going to introduce any major changes in the code during the refactoring process. If MSAL4J wants to consume it in the future, then we can do it.

common4j/src/main/com/microsoft/identity/common4j/util/StringUtil.java
AdamBJohnsonx: We should pick one of the other nullability annotations and stick to it.

findbugs has some, there is an quiescent jet with this stuff in it, too.

common4j/src/main/com/microsoft/identity/common4j/external/logging/IDetailedLoggerCallback.java
rpdome: PII when allowPII == false, or when log level is below what was set.

common4j/src/main/com/microsoft/identity/common4j/external/logging/Logger.java
rpdome: It's the same idea as what we have in common - anything that is being consumed externally should use external.

common4j/gradlew
rpdome: Added to gitignore and force pushed to resolve this.

common4j/gradlew.bat
rpdome: Added to gitignore and force pushed to resolve this.

common/src/main/java/com/microsoft/identity/common/logging/DiagnosticContext.java
rpdome: Change the scheme to match with android-common (anything internal will be under /internal, anything else is external)

common4j/src/main/com/microsoft/identity/common4j/external/logging/Logger.java
rpdome: Changed the scheme to match with android-common (anything internal will be under /internal, anything else is external)

common4j/src/main/com/microsoft/identity/common4j/util/StringUtil.java
rpdome: I'm going with Lombok's annotation.

common/src/main/java/com/microsoft/identity/common/internal/logging/IRequestContext.java
rpdome: Changed to com.microsoft.identity.common.java

common4j/src/main/com/microsoft/identity/common/java/logging/IDetailedLoggerCallback.java
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * An extension of ILoggerCallback - so that it prints the discarded log.
+ * This is for testing only (to verify that logs are actually discarded).
+ */
+interface IDetailedLoggerCallback extends ILoggerCallback {
+    /**
+     * Messages that are discarded by the loggers.
+     */
+    void discardedLog(String tag, Logger.LogLevel logLevel, String message, boolean containsPII);
+}
```
shahzaibj, line 12: new line

common4j/src/main/com/microsoft/identity/common/java/logging/ILoggerCallback.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Interface method for apps to hand off each log message as it's generated.
+     *
+     * @param tag         The TAG for the log message.
+     * @param logLevel    The {@link Logger.LogLevel} for the generated message.
+     * @param message     The detailed message.
+     * @param containsPII True if the log message contains PII, false otherwise.
+     */
+    void log(String tag, Logger.LogLevel logLevel, String message, boolean containsPII);
+}
```
shahzaibj, line 39: new line

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -0,0 +1,446 @@
...
+            sLoggersLock.writeLock().unlock();
+        }
+    }
+
+    public static void setLogger(@NonNull String identifier,
+                                 ILoggerCallback callback) {
+        sLoggersLock.writeLock().lock();
+        try {
+            if (callback == null &&
+                    sLoggers.containsKey(identifier)) {
```
shahzaibj, line 113: Do we need to `constainsKey`? I don't think so

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -0,0 +1,446 @@
...
+     * The need for this is because DiagnosticContext contains additional metadata which is not always required to be logged.
+     *
+     * @return String The concatenation of thread_name and correlation_id to serve as the required metadata in the log lines.
+     */
+    public static synchronized String getDiagnosticContextMetadata() {
+        String threadName = DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.THREAD_NAME);
+        String correlationId = DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.CORRELATION_ID);
+
+        if (StringUtil.isNullOrEmpty(threadName)) {
+            threadName = "UNSET";
```
shahzaibj, line 136: Can we put UNSET in a constant?

common/src/main/java/com/microsoft/identity/common/logging/DiagnosticContext.java
```diff
@@ -63,33 +45,13 @@ public static void setRequestContext(final IRequestContext requestContext) {
...
-        );
-        REQUEST_CONTEXT_THREAD_LOCAL.get().put(
-                THREAD_NAME,
-                Thread.currentThread().getName()
-        );
-    }
-
-    private static boolean hasThreadMetadata() {
-        return REQUEST_CONTEXT_THREAD_LOCAL.get().containsKey(THREAD_ID);
+        return (IRequestContext) com.microsoft.identity.common.java.logging.DiagnosticContext.INSTANCE.getRequestContext();
```
shahzaibj, line 48: seems strange that we need to cast here...I believe they are same types

shahzaibj (approved):
Left minor comments

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
rpdome: Yes. This means "if this identifier is already in the map, remove the old one *first*"

rpdome: ah.. I got what you're trying to say. nvm.

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
```diff
@@ -0,0 +1,77 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.logging;
+
+public enum DiagnosticContext {
```
iambmelt, line 25: FYI we now have 3 classes called `DiagnosticContext` -- we should javadoc which we actually intend to call from where

common4j/src/main/com/microsoft/identity/common/java/logging/ILoggerCallback.java
```diff
@@ -0,0 +1,39 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.logging;
+
+/**
+ * Interface for apps to configure the external logging and implement the callback to designate the
+ * output of the log messages.
+ */
+public interface ILoggerCallback {
```
iambmelt, line 29: Same feedback here - multiple classes with this name now, distinguished only by the pkg name. Can we javadoc when this should be used directly?

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -0,0 +1,446 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Setter;
+import lombok.Synchronized;
+import lombok.experimental.Accessors;
+
+public class Logger {
```
iambmelt, line 44: Same feedback as other classes - multiple classes with same name and API; can we javadoc which is to be used by whom?

Maybe we want to consider binding all of these implementations together with a shared interface? It seems like these could potentially get out of sync if one were to start adding methods

common4j/src/main/com/microsoft/identity/common/java/logging/RequestContext.java
```diff
@@ -0,0 +1,38 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.logging;
+
+import com.google.gson.Gson;
+
+import java.util.HashMap;
+
+public class RequestContext extends HashMap<String, String> implements IRequestContext {
```
iambmelt, line 29: Same feedback - multiple classes with same name. We should javadoc when to use which

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -0,0 +1,29 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.util;
+
+public class StringUtil {
```
iambmelt, line 25: Do we want to do a wholesale lift of the `StringUtil` from "regular" android common? And maybe `@Deprecate` the old one?

Maybe we even consider elevating this to a _major_ version change and remove the androidcommon impl?

common4j/src/main/com/microsoft/identity/common/java/util/ThrowableUtil.java
```diff
@@ -0,0 +1,36 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.util;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public class ThrowableUtil {
+    public static synchronized String getStackTraceAsString(Throwable throwable) {
```
iambmelt, line 29:
```suggestion
    public static synchronized String getStackTraceAsString(final Throwable throwable) {
```

common4j/src/main/com/microsoft/identity/common/java/logging/ILoggerCallback.java
rpdome: I'll do that and mark the other one as deprecated.

rpdome: Synced offline. We'll create separate PR every once in a while and deprecate a bunch of them - and add Irving to make sure he's aware, since OneAuth doesn't compile when it sees deprecated.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: We'll do a major version change - once the code are completely ported to java-common (so that we only break *once*)

iambmelt (approved):
Synced offline LGTM

PR #1301, opened 2021-04-02
Make version computation less manual

AdamBJohnsonx:
This is an idea.  I like it because it centralizes the API version code and means that we'll always have the correct versions.  The constructor for this things get better after my broker code goes in because we can delete all the _CODE constants and drop a parameter from the enum constructor.

It also gives each individual API its own version, and eventually we can just use that instead of this aggregate computation.


common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -110,7 +112,7 @@ public Bundle getAccountManagerBundle()
     private String getAccountManagerAddAccountOperationKey() throws BrokerCommunicationException{
         final String methodName = ":getAccountManagerAddAccountOperationKey";
 
-        String accountManagerKey = operation.accountManagerOperation;
+        String accountManagerKey = operation.mAccountManagerOperation;
```
AdamBJohnsonx, line 115:
```suggestion
        String accountManagerKey = operation.getAccountManagerOperation();
```

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -126,7 +128,7 @@ private String getAccountManagerAddAccountOperationKey() throws BrokerCommunicat
     public String getContentProviderPath() throws BrokerCommunicationException {
         final String methodName = ":getContentProviderUriPath";
 
-        final API contentApi = operation.contentApi;
+        final API contentApi = operation.mContentApi;
```
AdamBJohnsonx, line 131:
```suggestion
        final API contentApi = operation.getContentApi();
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -662,7 +666,32 @@
         /**
          * The newest Msal-To-Broker protocol version.
          */
-        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "6.0";
+        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE;
+
+        /**
+         * The logging tag for this class.
+         */
+        private static final String TAG =  AuthenticationConstants.class.getSimpleName();
```
AdamBJohnsonx, line 674: obv this would move to top of class

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -672,8 +701,27 @@
         /**
          * The newest BrokerAPI-To-Broker protocol version.
          */
-        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = "1.0";
-
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE;
+
+        static {
+            float protocolVersion = 1.0f;
+            String stringVersion = "1.0";
```
AdamBJohnsonx, line 708: by just tracking the string in parallel, I don't have to try to format the numeric which is filled with edge cases.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1392,35 +1440,37 @@
...
+            BROKER_GET_FLIGHTS(BROKER_API_GET_FLIGHTS_PATH, null, "2.0", "5.0"),
+            GET_SSO_TOKEN(GET_SSO_TOKEN_PATH, null, null, "7.0"),
+            UNKNOWN(null, null, null, null)
                 ;
             private String mPath;
             private Integer mCode;
-            API(String path, int code) {
+            private String mBrokerVersion;
+            private String mMsalVersion;
+            private API(String path, Integer code, String brokerApiVersion, String msalApiVersion) {
```
AdamBJohnsonx, line 1469: an alternate way to write this would be to have a parameter indicating whether or not the api was broker or small, and then a single version string.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1388,43 +1447,66 @@
...
+            ACQUIRE_TOKEN_INTERACTIVE(MSAL_ACQUIRE_TOKEN_INTERACTIVE_PATH, MSAL_ACQUIRE_TOKEN_INTERACTIVE_CODE, null, VERSION_5),
+            ACQUIRE_TOKEN_SILENT(MSAL_ACQUIRE_TOKEN_SILENT_PATH, MSAL_ACQUIRE_TOKEN_SILENT_CODE, null, VERSION_5),
+            GET_ACCOUNTS(MSAL_GET_ACCOUNTS_PATH, MSAL_GET_ACCOUNTS_CODE, null, VERSION_5),
+            REMOVE_ACCOUNTS(MSAL_REMOVE_ACCOUNTS_PATH, MSAL_REMOVE_ACCOUNTS_CODE, null, VERSION_5),
+            GET_CURRENT_ACCOUNT_SHARED_DEVICE(MSAL_GET_CURRENT_ACCOUNT_SHARED_DEVICE_PATH, MSAL_GET_CURRENT_ACCOUNT_SHARED_DEVICE_CODE, null, VERSION_5),
+            GET_DEVICE_MODE(MSAL_GET_DEVICE_MODE_PATH, MSAL_GET_DEVICE_MODE_CODE, null, VERSION_5),
+            SIGN_OUT_FROM_SHARED_DEVICE(MSAL_SIGN_OUT_FROM_SHARED_DEVICE_PATH, MSAL_SIGN_OUT_FROM_SHARED_DEVICE_CODE, null, VERSION_5),
+            GENERATE_SHR(GENERATE_SHR_PATH, MSAL_GENERATE_SHR_CODE, null, VERSION_6),
+            BROKER_HELLO(BROKER_API_HELLO_PATH, BROKER_API_HELLO_URI_CODE, BROKER_VERSION_1, VERSION_5),
+            BROKER_GET_ACCOUNTS(BROKER_API_GET_BROKER_ACCOUNTS_PATH, BROKER_API_GET_BROKER_ACCOUNTS_CODE, BROKER_VERSION_1, VERSION_5),
```
rpdome, line 1479: Why does Broker Operations have MSAL version?

AdamBJohnsonx: Two options, since both APIs are in the same space (BrokerContentProvider) - flag the broker APIs or use separate versions.  I thought the latter was less confusing, but I can take either one.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -672,7 +714,28 @@
         /**
          * The newest BrokerAPI-To-Broker protocol version.
          */
-        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = "1.0";
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = computeMinHostBrokerProtocol();
```
rpdome, line 717: this should be "max", not "min".

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -662,7 +683,28 @@
         /**
          * The newest Msal-To-Broker protocol version.
          */
-        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "6.0";
+        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = computeMinMsalBrokerProtocol();
+
+        @VisibleForTesting
+        public static String computeMinMsalBrokerProtocol() {
```
rpdome, line 689: this should be "max", not "min".

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx, line 717:
```suggestion
        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = computeMaxHostBrokerProtocol();
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -672,7 +714,28 @@
         /**
          * The newest BrokerAPI-To-Broker protocol version.
          */
-        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = "1.0";
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = computeMinHostBrokerProtocol();
+
+        @VisibleForTesting
+        public static String computeMinHostBrokerProtocol() {
```
AdamBJohnsonx, line 720:
```suggestion
        private static String computeMaxHostBrokerProtocol() {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -662,7 +683,28 @@
         /**
          * The newest Msal-To-Broker protocol version.
          */
-        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "6.0";
+        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = computeMinMsalBrokerProtocol();
```
AdamBJohnsonx, line 686:
```suggestion
        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = computeMaxMsalBrokerProtocol();
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx, line 689:
```suggestion
        private static String computeMaxMsalBrokerProtocol() {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -662,7 +683,28 @@
         /**
          * The newest Msal-To-Broker protocol version.
          */
-        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = "6.0";
+        public static final String MSAL_TO_BROKER_PROTOCOL_VERSION_CODE = computeMaxMsalBrokerProtocol();
+
+        @VisibleForTesting
+        private static String computeMaxMsalBrokerProtocol() {
```
AdamBJohnsonx, line 689:
```suggestion
        public static String computeMaxMsalBrokerProtocol() {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -672,7 +714,28 @@
         /**
          * The newest BrokerAPI-To-Broker protocol version.
          */
-        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = "1.0";
+        public static final String BROKER_API_TO_BROKER_PROTOCOL_VERSION_CODE = computeMaxHostBrokerProtocol();
+
+        @VisibleForTesting
+        private static String computeMaxHostBrokerProtocol() {
```
AdamBJohnsonx, line 720:
```suggestion
        public static String computeMaxHostBrokerProtocol() {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1388,46 +1451,56 @@
...
-            BROKER_HELLO(BROKER_API_HELLO_PATH, BROKER_API_HELLO_URI_CODE),
-            BROKER_GET_ACCOUNTS(BROKER_API_GET_BROKER_ACCOUNTS_PATH, BROKER_API_GET_BROKER_ACCOUNTS_CODE),
-            BROKER_REMOVE_ACCOUNT(BROKER_API_REMOVE_BROKER_ACCOUNT_PATH, BROKER_API_REMOVE_BROKER_ACCOUNT_CODE),
-            BROKER_UPDATE_BRT(BROKER_API_UPDATE_BRT_PATH, BROKER_API_UPDATE_BRT_CODE),
-            BROKER_ADD_FLIGHTS(BROKER_API_ADD_FLIGHTS_PATH),
-            BROKER_SET_FLIGHTS(BROKER_API_SET_FLIGHTS_PATH),
-            BROKER_GET_FLIGHTS(BROKER_API_GET_FLIGHTS_PATH),
-            GET_SSO_TOKEN(GET_SSO_TOKEN_PATH),
-            UNKNOWN(null)
+            MSAL_HELLO(MSAL_HELLO_PATH, null, VERSION_5),
```
rpdome, line 1473: Should these be 3?

ref: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2F%5BAndroid%5D%20Broker%20API%2Fbroker_protocol_versions.md

common/src/test/java/com/microsoft/identity/common/unit/AuthenticationConstantsTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AuthenticationConstantsTest {
+    @Test
+    public void testComputeMinBrokerHostVersion() {
+        Assert.assertEquals("2.0", AuthenticationConstants.Broker.computeMinHostBrokerProtocol());
```
rpdome, line 11: probably outdated.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1388,46 +1451,56 @@
          */
         public static final String AUTHORITY = "microsoft.identity.broker";
 
+        private static final String VERSION_1 = ONE_POINT_ZERO;
+        private static final String VERSION_5 = FIVE_POINT_ZERO;
+        private static final String VERSION_6 = "6.0";
```
rpdome, line 1456: nit: Why can this be "6.0"
(Actually, why do we need stuffs like "ONE_POINT_ZERO")?

AdamBJohnsonx: Because if I put more than three copies of "1.0" in the file, PMD fails the build.  I don't want to change our ruleset right now.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
AdamBJohnsonx: fixed.  I plan to copy or this into the ad-accounts package at some point.

PR #1303, opened 2021-04-02
Add an optional alias to RawKeyAccessor

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
```diff
@@ -60,21 +61,40 @@
...
+    /**
+     * The byte array that backs this key.
+     */
+    @NonNull
+    private final byte[] mKey;
+
+    /**
+     * The alias for the stored key, may be null.
+     */
+    private final String mAlias;
```
shahzaibj, line 83: I don't see this ever being set

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
```diff
@@ -193,7 +213,7 @@ public boolean verify(@NonNull final byte[] text,
     }
 
     @Override
-    @Nullable
+    @CheckForNull
```
shahzaibj, line 216: Why change this? How is this different from Nullable and what is the benefit?

AdamBJohnsonx: Well, it's not part of androidx, for one.  Lombok doesn't have a corresponding @Nullable, and this is the one settled on by the JEP.

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
AdamBJohnsonx: Hm.  It wouldn't be, here.  I end up manufacturing these in the PRT code path, but I don't think we communicate a session key alias at that point.  I should set it in the random-key alias part of the KeyStoreAccessor, though.

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
AdamBJohnsonx: Though it's been dormant for a while.  It depends on what we want to do with it.

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
AdamBJohnsonx: 37ed32387b54c8540e6973e8c53ce52aa621eb72

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
```diff
@@ -35,6 +33,7 @@
 import java.security.cert.Certificate;
 import java.util.Arrays;
 
+import javax.annotation.CheckForNull;
```
AdamBJohnsonx, line 36:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
AdamBJohnsonx, line 215:
```suggestion
    @Nullable
```

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
AdamBJohnsonx:
```suggestion
import androidx.annotation.Nullablel;
```

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
```diff
@@ -45,7 +43,9 @@
 import javax.crypto.spec.SecretKeySpec;
```
AdamBJohnsonx, line 44:
```suggestion
import androidx.annotation.Nullable;
```

common/build.gradle
```diff
@@ -138,6 +138,8 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
     implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayoutVersion"
+    implementation 'net.jcip:jcip-annotations:1.0'
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
```
AdamBJohnsonx, line 142:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/platform/RawKeyAccessor.java
```diff
@@ -226,7 +244,7 @@ public SecureHardwareState getSecureHardwareState() {
 
     /**
      * Given this raw key, generate a derived key from it.  If we close on a KDF for hardware keys,
-     * this can get promoted to the symmetric key interface.
+     * this can get promoted to the symmetric key interface.  Derived keys have a null alias.
```
iambmelt, line 247: (mentioned offline, but putting here: let's doc why these are null aliased)

iambmelt (approved):

iambmelt (approved):
Signing off on latest iteration

PR #1305, opened 2021-04-06
Only pass required data to hello bundle creation.

AdamBJohnsonx:
This is basically about not requiring a grab-bag of information to construct the hello bundle.  While I was looking at this, we construct it twice, so maybe don't do that.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -191,19 +191,8 @@ public class BrokerMsalController extends BaseController {
             );
         }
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -170,44 +171,45 @@ public String getActiveBrokerPackageName() {
      * MSAL-Broker handshake operation.
      *
      * @param strategy   an {@link IIpcStrategy}
-     * @param parameters a {@link CommandParameters}
+     * @param minRequestedVersion
```
AdamBJohnsonx, line 174:
```suggestion
     * @param minRequestedVersion the minimum allowed broker protocol version, may be null.
```

shahzaibj (approved):

PR #1306, opened 2021-04-07
Extract package verification methods into a utility class.

AdamBJohnsonx:
Completely mechanical movement of the broker validator private methods into a utility class.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
@@ -143,6 +143,10 @@ public class BrokerValidator {
         return validBrokers;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -160,6 +143,26 @@ public boolean verifySignature(final String brokerPackageName) {
         return validBrokers;
     }
 
+    public Iterator<String> getValidBrokerSignatures() {
```
AdamBJohnsonx, line 146:
```suggestion
    /**
     * Get an iterator of access to valid broker signatures.
     * @return an iterator of access to valid broker signatures.
     */
    public Iterator<String> getValidBrokerSignatures() {
```

shoatman (approved):
lgtm

PR #1308, opened 2021-04-12
Implement equals and hash code for the pop-command paths.

AdamBJohnsonx:
One of our early adopters noticed that these were missing, because their PoP requests that used a new nonce were being returned precisely the same results.  This ends up changing a lot of files, because there are 7 classes changed, three tests added, and one additional test utility class created.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MapBackedPreferencesManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MapBackedPreferencesManager.java
@@ -1,0 +1,100 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.cache;
+
+import androidx.annotation.VisibleForTesting;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.RequiredArgsConstructor;
+
+@Builder
+@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+@RequiredArgsConstructor
+public class MapBackedPreferencesManager implements ISharedPreferencesFileManager {
+
+    private final String mName;
+
+    private final Map<String, String> mBackingStore = new HashMap<>();
+
+    @Override
+    public void putString(String key, String value) {
+        mBackingStore.put(key, value);
+    }
+
+    @Override
+    public String getString(String key) {
+        return mBackingStore.get(key);
+    }
+
+    @Override
+    public void putLong(String key, long value) {
+        mBackingStore.put(key, Long.toString(value));
+    }
+
+    @Override
+    public long getLong(String key) {
+        String s = mBackingStore.get(key);
+        return s == null ? 0 : Long.parseLong(s);
+    }
+
+    @Override
+    public String getSharedPreferencesFileName() {
+        return mName;
+    }
+
+    @Override
+    public Map<String, String> getAll() {
+        return new HashMap(mBackingStore);
+    }
+
+    @Override
+    public Iterator<Map.Entry<String, String>> getAllFilteredByKey(SharedPreferencesFileManager.Predicate<String> keyFilter) {
+        Map<String, String> newMap = new HashMap<>();
+        for (Map.Entry<String, String> entry: mBackingStore.entrySet()) {
+            if (keyFilter.test(entry.getKey())) {
+                newMap.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return newMap.entrySet().iterator();
+    }
+
+    @Override
+    public boolean contains(String key) {
+        return mBackingStore.containsKey(key);
+    }
+
+    @Override
+    public void clear() {
+        mBackingStore.clear();
+    }
+
+    @Override
+    public void remove(String key) {
+        mBackingStore.remove(key);
+    }
+}
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/MapBackedPreferencesManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/MapBackedPreferencesManager.java
@@ -35,6 +35,10 @@ import lombok.RequiredArgsConstructor;
 @Builder
 @VisibleForTesting(otherwise = VisibleForTesting.NONE)
 @RequiredArgsConstructor
+/**
+ * A SharedPreferencesFileManager backed by a HashMap.  This is mainly for testing purposes,
+ * where it doesn't make sense to instantiate shared preferences files.
+ */
 public class MapBackedPreferencesManager implements ISharedPreferencesFileManager {
 
     private final String mName;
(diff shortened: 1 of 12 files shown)


common/src/test/java/com/microsoft/identity/common/internal/commands/parameters/GenerateShrCommandParametersTest.java
```diff
@@ -0,0 +1,71 @@
...
+                        .build())
+                .build();
+
+        Map<CommandParameters, Boolean> map = new HashMap<>();
+        map.put(commandOne, true);
+        Assert.assertEquals(1, map.size());
+        map.put(commandTwo, true);
+        Assert.assertEquals(2, map.size());
+    }
+
```
AdamBJohnsonx, line 70:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/cache/MapBackedPreferencesManager.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Map;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.RequiredArgsConstructor;
+
+@Builder
+@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+@RequiredArgsConstructor
+public class MapBackedPreferencesManager implements ISharedPreferencesFileManager {
```
shahzaibj, line 38: javadoc?

shahzaibj: Is this actually being used anywhere right now?

AdamBJohnsonx, line 38:
```suggestion
/**
 * A SharedPreferencesFileManager backed by a HashMap.  This is mainly for testing purposes,
 * where it doesn't make sense to instantiate shared preferences files.
 */
public class MapBackedPreferencesManager implements ISharedPreferencesFileManager {
```

AdamBJohnsonx: Yes, it is.  I can add javadoc if it's worthwhile.  It's in pretty much every test that I added here.

common/src/test/java/com/microsoft/identity/common/internal/commands/GenerateShrCommandTest.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        @Override
+        public void onTaskCompleted(Object o) {
+
+        }
+    };
+
+    @Test
+    public void testMappability() throws Exception {
```
shahzaibj, line 60: are we just testing here that we can successfully put two different command objects in a hashmap? Do we instead want to test that we can put these in a `Set` and write a test with two different command objects that are essentially equal?

AdamBJohnsonx: Is there a difference between a HashSet and a HashMap?

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/GenerateShrCommandParameters.java
```diff
@@ -24,23 +24,27 @@
...
 
 /**
  * Parameter class for generating SHRs.
  */
 @Getter
-@SuperBuilder
+@SuperBuilder(toBuilder = true)
+@EqualsAndHashCode(callSuper = true)
+@Accessors(prefix = "m")
 public class GenerateShrCommandParameters extends CommandParameters {
```
iambmelt, line 39: FYI this class is also consumed by [OneAuth](https://office.visualstudio.com/OneAuth/_search?action=contents&text=GenerateShr&type=code&lp=code-Project&filters=ProjectFilters%7BOneAuth%7DRepositoryFilters%7BOneAuth%7D&pageSize=25&result=DefaultCollection%2FOneAuth%2FOneAuth%2FGBdev%2F%2Fmsal%2Fsource%2Fandroid%2Fmsal%2Fapp%2Fsrc%2Fmain%2Fjava%2Fcom%2Fmicrosoft%2Fidentity%2Finternal%2Fbroker%2FBrokerRequestConverter.java) -- I don't see anything here that looks breaking, but FYI in case you are

iambmelt (approved):

common/src/test/java/com/microsoft/identity/common/internal/commands/GenerateShrCommandTest.java
shahzaibj: Depends on what we are putting in for the `values` in case of a `Map`...it seems like in your case there is no difference as you are just putting a `boolean` which I believe indicates whether a key is present or not....which is why I am thinking that this should be a `HashSet` as opposed to a `Map`...because it seems like all we care about here is the size of the collection...is that the right understanding?

AdamBJohnsonx: Right.  It doesn't really matter to me.  The bug we're dealing with was a problem with usage of a Map, so I used a Map in the test.  And since a Set is just a Map underneath, I thought it would be clearer that what we're trying to check is that we can put things in a map and it behaves reasonably.

shahzaibj: I think we also want to have some test for the `equals` method of the `GenerateShrCommand` object i.e. if we create two different objects each with the same parameters such that we expect them to be equal and then we call the `equals` method on this, does it actually return true.

AdamBJohnsonx: I'll add tests for equals and hash code.  We should have tests for any method that has code in it, really.

common/src/test/java/com/microsoft/identity/common/internal/commands/GenerateShrCommandTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    public void testHashCode_equals() throws Exception {
+        Assert.assertEquals(COMMAND_ONE.hashCode(), COMMAND_ONE_CLONE.hashCode());
+    }
+
+    @Test
+    public void testHashCode_notEquals() throws Exception {
+        Assert.assertEquals(COMMAND_ONE.hashCode(), COMMAND_ONE_CLONE.hashCode());
```
AdamBJohnsonx, line 136:
```suggestion
        Assert.assertNotEquals(COMMAND_ONE.hashCode(), COMMAND_TWO.hashCode());
```

common/src/test/java/com/microsoft/identity/common/internal/commands/parameters/GenerateShrCommandParametersTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Assert.assertEquals(2, map.size());
+    }
+    @Test
+    public void testHashCode_equals() throws Exception {
+        Assert.assertEquals(PARAMS_ONE.hashCode(), PARAMS_ONE_CLONE.hashCode());
+    }
+
+    @Test
+    public void testHashCode_notEquals() throws Exception {
+        Assert.assertEquals(PARAMS_ONE.hashCode(), PARAMS_ONE_CLONE.hashCode());
```
AdamBJohnsonx, line 118:
```suggestion
        Assert.assertNotEquals(PARAMS_ONE.hashCode(), PARAMS_TWO.hashCode());
```

common/src/test/java/com/microsoft/identity/common/internal/authscheme/PopAuthenticationSchemeTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    
+    @Test
+    public void testHashCode_equals() throws Exception {
+        Assert.assertEquals(AUTHSCHEME_ONE.hashCode(), AUTHSCHEME_ONE_CLONE.hashCode());
+    }
+
+    @Test
+    public void testHashCode_notEquals() throws Exception {
+        Assert.assertEquals(AUTHSCHEME_ONE.hashCode(), AUTHSCHEME_ONE_CLONE.hashCode());
```
AdamBJohnsonx, line 61:
```suggestion
        Assert.assertNotEquals(AUTHSCHEME_ONE.hashCode(), AUTHSCHEME_TWO.hashCode());
```

common/src/test/java/com/microsoft/identity/common/internal/commands/GenerateShrCommandTest.java
AdamBJohnsonx: @shahzaibj Should be in latest revision

common/src/test/java/com/microsoft/identity/common/internal/util/UrlUtils.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A place for gathering utilities for use with URLs.
+ */
+public class UrlUtils {
+    /**
+     * This creates a url from a String, rewriting any malformedUrlExceptions to runtime.
+     * @param urlString the string to convert.
+     * @return the corresponding {@link URL}.
+     */
+    public static URL makeUrlSilent(String urlString) {
```
shahzaibj, line 37: Silent? or was this meant to be something like `makeUrlFromString`

shahzaibj (approved):
LGTM - Thanks @AdamBJohnsonx

common/src/test/java/com/microsoft/identity/common/internal/util/UrlUtils.java
AdamBJohnsonx: The only reason to use this is that I wanted to call it in a static initialization context, and it couldn't have a checked exception declared.  So... it could have been "Sneaky", but...

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/GenerateShrCommandParameters.java
AdamBJohnsonx: This should be completely unnoticeable by OneAuth.

PR #1309, opened 2021-04-13
Minor tweak in JWE response class, promote key derivation to first-class.

AdamBJohnsonx:
This is mainly promotion of an interface from RawKeyAccessor upwards through the key accessor classes to top-level.  We currently don't have a key derivation for inaccessible, hardware-backed keys, but I think it likely enough that we do so in the future that we should prepare for it.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@ V.Next
 - [MINOR] Adds new API support for the broker - SSO token and flight support (#1290)
 - [MINOR] Elevates AndroidCommon Logger out of internal package (#1279)
 - [MINOR] Automate broker protocol version computation. (#1301)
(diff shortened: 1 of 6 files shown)


changelog.txt
```diff
@@ -8,6 +8,7 @@ V.Next
 - [MINOR] Adds new API support for the broker - SSO token and flight support (#1290)
 - [MINOR] Elevates AndroidCommon Logger out of internal package (#1279)
 - [MINOR] Automate broker protocol version computation. (#1301)
+- [MAJOR] New API on KeyAccessor (#1309)
```
AdamBJohnsonx, line 11: Technically, by adding a method to an interface without a default implementation, this is a breaking change.

iambmelt (approved):

PR #1312, opened 2021-04-15
If we end up with a null component name, don't pass it into package manager.

AdamBJohnsonx:
This began by debugging the submodule update in https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1355/ 

It looks like we can end up with null ComponentNames when using Robolectric, that we're passing into the Shadow package manager and it's failing on line 167.  I *think* the best way to deal with this is just to be defensive in this case and treat it as if it had no affinity.

Not all the tests pass locally for me, but with this change, *this* failure does not happen.


iambmelt (approved):
Curious! 


LGTM

tanmaymanolkar1 (approved):

PR #1314, opened 2021-04-15
Fix network connectivity callback

gitaumoses4:
Ensure there's a single instance of the `ConnectivityManager.NetworkCalback` in the `DefaultConnectionService`.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
@@ -50,7 +50,11 @@ public class DefaultConnectionService implements IConnectionService {
 
     private final Context mConnectionContext;
     private static boolean connectionAvailable = false;
-    private ConnectivityManager.NetworkCallback networkCallback;
+    // We need a single instance of the NetworkCallback to prevent multiple callbacks from being
+    // registered for each instance of DefaultConnectionService.
+    private static ConnectivityManager.NetworkCallback networkCallback;
+
+    private static final Object lock = new Object();
 
     /**
      * Constructor of DefaultConnectionService.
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -107,10 +109,12 @@ public void onLost(@NonNull Network network) {
...
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     public void unregisterNetworkCallback() {
-        final ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
-                .getSystemService(Context.CONNECTIVITY_SERVICE);
+        if (networkCallback != null) {
+            final ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
+                    .getSystemService(Context.CONNECTIVITY_SERVICE);
 
-        connectivityManager.unregisterNetworkCallback(networkCallback);
+            connectivityManager.unregisterNetworkCallback(networkCallback);
```
cofitz1, line 116: do we need to `null` out `networkCallback` too?

gitaumoses4: Yes, I think we do. No need for it to be dangling in memory.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -107,10 +109,12 @@ public void onLost(@NonNull Network network) {
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     public void unregisterNetworkCallback() {
-        final ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
-                .getSystemService(Context.CONNECTIVITY_SERVICE);
+        if (networkCallback != null) {
```
AdamBJohnsonx, line 112: Is there some particular reason to lazy-initialize this?  If there is, we need to do it right, meaning we’ll have to synchronize it somehow.  It might be simpler to do in an initializer, though.

gitaumoses4: The only reason is to make sure that `registerNetworkCallback()` does not re-register the callback when invoked multiple times.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -50,7 +50,9 @@
 
     private final Context mConnectionContext;
     private static boolean connectionAvailable = false;
-    private ConnectivityManager.NetworkCallback networkCallback;
+    // We need a single instance of the NetworkCallback to prevent multiple callbacks from being
+    // registered for each instance of DefaultConnectionService.
+    private static ConnectivityManager.NetworkCallback networkCallback;
```
rpdome, line 55: add `synchronized` to both register/unregister function.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -50,7 +50,9 @@
 
     private final Context mConnectionContext;
     private static boolean connectionAvailable = false;
```
rpdome, line 52: If this is private, do we need to register the callback in the first place? Is it being used anywhere?

If not, it's probably safer to just removing it.

gitaumoses4: Yes, it is used in the `isConnectionAvailable` method.

AdamBJohnsonx: The reason to handle this via a callback is that the other mechanism is deprecated and broken in newer versions of android.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
AdamBJohnsonx: Right, but if you registered your callback on static initialization, if you can, then you don't need to do this at all.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
rpdome: Got it. Thanks.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
rpdome: I was looking into the existing code and seems like this function is never called anywhere.

common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
rpdome: My bad. this op should be wrapped inside a synchronize lock with a static lock.

rpdome (approved):

rpdome:
@gitaumoses4 Don't forget to
1. Cherry-pick this into the master branch of common, and cut a new MSAL (sync with @paulkagiri)
2. Cherry-pick this into the release branch of common for the `previous` Broker release, and cut a build for AuthApp, CP.
3. Cherry-pick this into the release branch of the next release of common (that is being bug bashed by Pedro).

AdamBJohnsonx (approved):

PR #1317, opened 2021-04-19
Package validation support for nGMS agent

tanmaymanolkar1:
only let broker host and nGMS agent access the Device Token WPJ API.

Added an API in package util to have a general package validation

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.adal.internal;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.VisibleForTesting;
 
 import com.microsoft.identity.common.internal.logging.Logger;
@@ -1060,6 +1061,23 @@ public final class AuthenticationConstants {
          * Broker Host app package name.
          */
         public static final String BROKER_HOST_APP_PACKAGE_NAME = "com.microsoft.identity.testuserapp";
+
+        /**
+         * Intune nGMS agent app package name.
+         */
+        public static final String INTUNE_NGMS_AGENT_APP_PACKAGE_NAME = "com.microsoft.intune.aospagent";
+
+        /**
+         * Signature info for Azure authenticator app that installs authenticator
+         * component.
+         */
+        public static final String INTUNE_NGMS_AGENT_APP_RELEASE_SIGNATURE = "1L4Z9FJCgn5c0VLhyAxC5O9LdlE=";
+
+        /**
+         * Signature info for Azure authenticator app that installs authenticator
+         * component.
+         */
+        public static final String INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE = "placeholder,correct value inserted here";
 
         /**
          * Azure Authenticator app package name.
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
@@ -40,6 +40,9 @@ import static com.microsoft.identity.common.adal.internal.AuthenticationConstant
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.BROKER_HOST_APP_SIGNATURE;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_APP_RELEASE_SIGNATURE;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.INTUNE_NGMS_AGENT_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.INTUNE_NGMS_AGENT_APP_RELEASE_SIGNATURE;
 
 /**
  * Represents packageName and SignatureHash of a broker app.
@@ -66,6 +69,16 @@ public class BrokerData {
             BROKER_HOST_APP_SIGNATURE
     );
 
+    public static final BrokerData INTUNE_NGMS_AGENT_DEBUG = new BrokerData(
+            INTUNE_NGMS_AGENT_APP_PACKAGE_NAME,
+            INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE
+    );
+
+    public static final BrokerData INTUNE_NGMS_AGENT_PROD = new BrokerData(
+            INTUNE_NGMS_AGENT_APP_PACKAGE_NAME,
+            INTUNE_NGMS_AGENT_APP_RELEASE_SIGNATURE
+    );
+
     private static final Set<BrokerData> DEBUG_BROKERS = Collections.unmodifiableSet(new HashSet<BrokerData>() {{
         add(MICROSOFT_AUTHENTICATOR_DEBUG);
         add(BROKER_HOST);
@@ -79,6 +92,13 @@ public class BrokerData {
     private static final Set<BrokerData> ALL_BROKERS = Collections.unmodifiableSet(new HashSet<BrokerData>() {{
         addAll(DEBUG_BROKERS);
         addAll(PROD_BROKERS);
+    }});
+
+    // This set is specific to whitelist Intune nGMS agent to access Device Token WPJ API.
+    private static final Set<BrokerData> INTUNE_NGMS_AGENT_SPECIFIC = Collections.unmodifiableSet(new HashSet<BrokerData>() {{
+        add(INTUNE_NGMS_AGENT_DEBUG);
+        add(INTUNE_NGMS_AGENT_PROD);
+        add(BROKER_HOST);
     }});
 
     public final String packageName;
@@ -113,5 +133,9 @@ public class BrokerData {
 
     public static Set<BrokerData> getAllBrokers() {
         return ALL_BROKERS;
+    }
+
+    public static Set<BrokerData> getIntuneNgmsAgentSpecific() {
+        return INTUNE_NGMS_AGENT_SPECIFIC;
     }
 }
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -66,6 +69,16 @@
             BROKER_HOST_APP_SIGNATURE
     );
 
+    public static final BrokerData INTUNE_NGMS_AGENT_DEBUG = new BrokerData(
+            INTUNE_NGMS_AGENT_APP_PACKAGE_NAME,
```
AdamBJohnsonx, line 73: Can we drop the static external constants and just supply them locally here?  It doesn't make any sense at all to put these into AuthenticationConstants.

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -202,4 +208,31 @@ public static final X509Certificate getSelfSignedCert(final List<X509Certificate
...
+
+    public static String signatureVerificationAndThrow(final String packageName, final Context context,
+                                                       final Iterator<String> validPackageSignatures)
+            throws ClientException {
+        try {
+            // Read all the certificates associated with the package name. In higher version of
+            // android sdk, package manager will only returned the cert that is used to sign the
+            // APK. Even a cert is claimed to be issued by another certificates, sdk will return
+            // the signing cert. However, for the lower version of android, it will return all the
+            // certs in the chain. We need to verify that the cert chain is correctly chained up.
```
AdamBJohnsonx, line 220: This looks like copy-paste.  Can we research which version of android this is, so we know when to delete the logic?

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -71,31 +72,32 @@
...
-                                                           final Context context)
+                                                                 final Context context)
             throws PackageManager.NameNotFoundException, ClientException, IOException,
             GeneralSecurityException {
 
         final PackageInfo packageInfo = PackageHelper.getPackageInfo(context.getPackageManager(), packageName);
         if (packageInfo == null) {
             throw new ClientException(ErrorStrings.APP_PACKAGE_NAME_NOT_FOUND,
-                    "No broker package existed.");
+                    "No such package existing : " + packageName);
```
AdamBJohnsonx, line 94:
```suggestion
                    "Did not find an installed package named : '" + packageName + "'");
```

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -71,31 +72,32 @@
...
         }
 
-        final Signature [] signatures = PackageHelper.getSignatures(packageInfo);
+        final Signature[] signatures = PackageHelper.getSignatures(packageInfo);
         if (signatures == null || signatures.length == 0) {
             throw new ClientException(BROKER_APP_VERIFICATION_FAILED,
-                    "No signature associated with the broker package.");
+                    "No signature associated with the package : " + packageName);
```
AdamBJohnsonx, line 100:
```suggestion
                    "No signature associated with the package : '" + packageName + "'");
```

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -145,6 +148,7 @@ public static final String verifySignatureHash(final @NonNull List<X509Certifica
             hashListStringBuilder.append(signatureHash);
             hashListStringBuilder.append(',');
 
+            //The next of the iterator should be overridden to get the signatureHash.
```
AdamBJohnsonx, line 151: I don't follow this comment

AdamBJohnsonx (approved):
A bunch of minor suggestions.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -66,6 +69,16 @@
...
+            INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE
+    );
+
+    public static final BrokerData INTUNE_NGMS_AGENT_PROD = new BrokerData(
```
AdamBJohnsonx, line 77: What we really want to vend here is essentially an object that contains a name and a collection of valid signatures for that name.  If we did that, everything would fall nicely out, I think.

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
tanmaymanolkar1: The next() functionality of the iterator should be overridden to retrieve the signatureHash of the entry, for the purpose of comparison in the following code steps.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
tanmaymanolkar1: but I think its better to have a single source of truth for such values.

tanmaymanolkar1: I could still make them local, but then keeping the values consistent across the codebase would be a problem,

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
tanmaymanolkar1: 8a24441

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
tanmaymanolkar1: 8a24441

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1061,6 +1062,23 @@ public static String computeMaxHostBrokerProtocol() {
...
+         * Signature info for Azure authenticator app that installs authenticator
+         * component.
+         */
+        public static final String INTUNE_NGMS_AGENT_APP_RELEASE_SIGNATURE = "1L4Z9FJCgn5c0VLhyAxC5O9LdlE=";
+
+        /**
+         * Signature info for Azure authenticator app that installs authenticator
+         * component.
+         */
+        public static final String INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE = "placeholder,correct value inserted here";
```
tanmaymanolkar1, line 1080: will update this value once I get hold of it. Wont merge the  PR till then.

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
tanmaymanolkar1: 51443b0

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -66,6 +69,16 @@
             BROKER_HOST_APP_SIGNATURE
     );
 
+    public static final BrokerData INTUNE_NGMS_AGENT_DEBUG = new BrokerData(
+            INTUNE_NGMS_AGENT_APP_PACKAGE_NAME,
+            INTUNE_NGMS_AGENT_APP_DEBUG_SIGNATURE
+    );
```
shahzaibj, line 75: INTUNE_NGMS_AGENT afaik is NOT a broker....and therefore these constants should NOT belong here...

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.java
```diff
@@ -81,6 +94,13 @@
         addAll(PROD_BROKERS);
     }});
 
+    // This set is specific to whitelist Intune nGMS agent to access Device Token WPJ API.
+    private static final Set<BrokerData> INTUNE_NGMS_AGENT_SPECIFIC = Collections.unmodifiableSet(new HashSet<BrokerData>() {{
+        add(INTUNE_NGMS_AGENT_DEBUG);
+        add(INTUNE_NGMS_AGENT_PROD);
+        add(BROKER_HOST);
+    }});
```
shahzaibj, line 102: Sane feedback as above...these should NOT belong here..

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -145,6 +148,7 @@ public static final String verifySignatureHash(final @NonNull List<X509Certifica
             hashListStringBuilder.append(signatureHash);
             hashListStringBuilder.append(',');
 
+            //The next() functionality of the iterator should be overridden to retrieve the signatureHash of the entry.
```
shahzaibj, line 151: What does this mean?

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -202,4 +208,41 @@ public static final X509Certificate getSelfSignedCert(final List<X509Certificate
...
+
+            return signatureHash;
+        } catch (PackageManager.NameNotFoundException e) {
+            throw new ClientException(ErrorStrings.APP_PACKAGE_NAME_NOT_FOUND, e.getMessage(), e);
+        } catch (NoSuchAlgorithmException e) {
+            throw new ClientException(ErrorStrings.NO_SUCH_ALGORITHM, e.getMessage(), e);
+        } catch (final IOException | GeneralSecurityException e) {
+            throw new ClientException(ErrorStrings.BROKER_VERIFICATION_FAILED, e.getMessage(), e);
+        }
+    }
```
shahzaibj, line 247: This looks pretty much a copy / paste from the logic in BrokerValidator.java....let's avoid duplicating code. I think this logic can just sit in one place though that may require a bit of refactoring

shahzaibj (changes requested):
Left some comments

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
tanmaymanolkar1: c9898c3

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
AdamBJohnsonx: That's kind of fundamental to the way iterators work.  hasNext() tells you if there is another element to consider, next() fetches it and updates the iterator state. <shrug>

tanmaymanolkar1: oh in this case, the next fetches the signature hash directly and not the element, I can go ahead and remove the comment, but just thought of pointing it out for the next time someone, uses this and passes in an iterator.

AdamBJohnsonx: OK, maybe "This iterator should traverse a list of signature hash elements".  You don't * need * to override next.  I mean, I could write elements.stream().map(e -> e.signatureHash).collect(Collectors.toList()).iterator(), and that shoud be fine, if a bit silly.

PR #1326, opened 2021-04-22
Renaming BrokerData class to AppData

tanmaymanolkar1:
For the purpose of generality, renaming this class to avoid confusion.

This is work is a primer for package validation task

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/AppData.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/AppData.java
@@ -44,39 +44,39 @@ import static com.microsoft.identity.common.adal.internal.AuthenticationConstant
 /**
  * Represents packageName and SignatureHash of a broker app.
  */
-public class BrokerData {
+public class AppData {
 
-    public static final BrokerData MICROSOFT_AUTHENTICATOR_DEBUG = new BrokerData(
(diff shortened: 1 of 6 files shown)


AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/AppData.java
```diff
@@ -44,74 +44,75 @@
...
-                       @NonNull final String hash) {
+    private AppData(@NonNull final String packageName,
+                    @NonNull final String hash) {
         this.packageName = packageName;
         this.signatureHash = hash;
     }
 
     /**
-     * Given a broker package name, verify its signature and return a BrokerData object.
+     * Given a broker package name, verify its signature and return a AppData object.
```
shahzaibj, line 94:
```suggestion
     * Given a broker package name, verify its signature and return an AppData object.
```

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/AppData.java
```diff
@@ -44,74 +44,75 @@
 /**
  * Represents packageName and SignatureHash of a broker app.
  */
-public class BrokerData {
+public class AppData {
```
iambmelt, line 47: Does this class really need to be `public`? If yes, do we want to consider taking it out of the `internal` package space?

PR #1332, opened 2021-04-24
Pedroro/common4j pipeline

p3dr0rv:
set a pipeline up for running tests in common4j

Pipeline here: 
https://identitydivision.visualstudio.com/Engineering/_build?definitionId=1382
pipeline is temporally located in experiments folder
location need to be defined


azure-pipelines/experiments/common4j.yml
```diff
@@ -0,0 +1,39 @@
...
+- job: Phase_1
+  displayName: Phase 1
+  steps:
+  - checkout: self
+    clean: true
+    submodules: recursive
+    persistCredentials: True
+  - template: ../templates/steps/credscan-policheck.yml
+    parameters:
+      policheckCmdLineArgsDir: adal
```
shahzaibj, line 26: what is this adal reference here?

azure-pipelines/experiments/common4j.yml
```diff
@@ -0,0 +1,39 @@
...
+  - template: ../templates/steps/automation-cert.yml
+  - task: Gradle@1
+    name: Gradle1
+    displayName: Assembles the outputs of this project.
+    inputs:
+      tasks: clean common4j:assemble
```
shahzaibj, line 32: can we build a specific flavor? probably `local`?

azure-pipelines/experiments/common4j.yml
p3dr0rv: Good catch, that shouldn't be adal. (this parameter is the folder to check with policheck)

shahzaibj (approved):

rpdome (approved):

PR #1336, opened 2021-04-27
Msal removeAccount shouldn't invoke broker's removeAccount

rpdome:
MSAL's removeAccount was invoking broker's removeAccount (the one that's supposed to be called by broker hosting app).

It doesn't actually remove broker account (the operation will fail). Still, it means that MSAL's removeAccount is now botched.


AdamBJohnsonx (approved):

AdamBJohnsonx:
We should have tests here.

PR #1339, opened 2021-04-28
Resurrect BrowserSelectorTest

rpdome:
This is breaking one of the pr pipeline. (The reason is that Mockito-android cannot mock final class).

I'm switching them to Robolectric.

Note: addResolveInfoForIntent() is already deprecated, but I can't find how to get the alternative work (since it requires ComponentName, ComponentName requires class name and we're not providing one here). I didn't want to spend too much time on it so I just pick the deprecated one for now.

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
@@ -22,7 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common;
 
-import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.pm.ActivityInfo;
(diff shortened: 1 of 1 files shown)


common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
```diff
@@ -253,14 +225,27 @@ public TestBrowserBuilder setVersion(String version) {
...
+        private PackageInfo addSignatures(final PackageInfo packageInfo, final Signature[] signatures) {
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+                packageInfo.signatures = signatures;
+                return packageInfo;
+            }
+
+            final SigningInfo signingInfo = mock(SigningInfo.class);
+            when(signingInfo.hasMultipleSigners()).thenReturn(false);
+            when(signingInfo.getSigningCertificateHistory()).thenReturn(signatures);
+            packageInfo.signingInfo = signingInfo;
```
shahzaibj, line 237: Can this be done with Robolectric too?

common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
```diff
@@ -253,14 +225,27 @@ public TestBrowserBuilder setVersion(String version) {
...
+        }
+
         public TestBrowser build() {
             Signature[] signatures = new Signature[mSignatures.size()];
 
             for (int i = 0; i < mSignatures.size(); i++) {
                 signatures[i] = new Signature(mSignatures.get(i));
             }
 
-            final PackageInfo pi = Util.addSignatures(new PackageInfo(), signatures);
```
shahzaibj, line 263: Is the Util method still there? Can we delete it if there are no usages?

shahzaibj (approved):
just minor comments, else LGTM!

common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
AdamBJohnsonx: Don't know that the methodology matters, but when you're writing code, if you want to make it easy to test - implement an interface, and provide a way to construct your POJOs.

Also, at some point, SigningInfo has been made final.

shoatman (approved):

AdamBJohnsonx (approved):
I don't see an issue here.

common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
rpdome: It's still used.

common/src/test/java/com/microsoft/identity/common/BrowserSelectorTest.java
rpdome: Spent a few mins with it and couldn't think of a way of doing this cleanly with robolectric (this is 'per browser', and there are >1 browsers in each tests), so I'm keeping this as is for now.

PR #1341, opened 2021-04-30
Add existing spotbugs violations as baseline

iamgusain:
Adding existing spotbugs violations as [baseline](https://spotbugs-gradle-plugin.netlify.app/com/github/spotbugs/snom/spotbugsextension#baselineFile).. So that we can catch new violations.
I will create a separate work item to fix these violations and then get rid of the baseline file
Workitem: https://identitydivision.visualstudio.com/Engineering/_backlogs/backlog/Auth%20Client%20-%20Android/Features/?workitem=1381103


tanmaymanolkar1 (approved):
LGTM!

AdamBJohnsonx (approved):

shahzaibj:
These baselines cover both common and common4j, right?

iamgusain:
> These baselines cover both common and common4j, right?

No, its just common.

PR #1347, opened 2021-05-03
Migrate HTTP Layer to common-java

rpdome:
## Changes
1. Migrate UrlConnectionHttpClient, HttpRequest and HttpResponse to common4j.
2. The classes are being consumed by MSAL/Broker, so I'm leaving the old ones as adapters. (Will remove them all in one go once we're ready for a breaking change.)
3. Strip redundant logic in common4j's HttpRequest - now it's just a data container.
4. common's HttpRequest is now built on top of common4j's UrlConnectionHttpClient.getDefaultInstance()

No changes in MSAL, ADAL, Broker should be required.

## Functional Change
1. `READ_TIMEOUT` and `CONNECT_TIMEOUT` in common's HttpRequest (which is set by MSAL) are now ignored.
2. UrlConnectionHttpClient.getDefaultInstance() now contains retry logic and default timeout.

## Test cases
1. Update HttpRequestTest to only handle the legacy HttpRequest methods.
2. Migrate the rest of the code to UrlConnectionHttpClientTest.
3. Migrate HttpResponseBodyTest and HttpUrlConnectionFactoryTest to common4j.

Verified manually with
- MSAL acquireTokenSilent flow (non-brokered auth) 
   - uses UrlConnectionHttpClient.getDefaultInstance()
- joined acquireTokenSilent flow.
   - uses both UrlConnectionHttpClient.getDefaultInstance and deprecated functions in common's HttpClient


common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+ * <p>
+ * There are two ways to supply timeout values to this class, one method takes suppliers, the other
+ * one integers.  If you use both of these, the suppliers method will take precendence over the method
+ * using integers.
+ */
+@AllArgsConstructor
+@Builder
+@ThreadSafe
+@Immutable
+public class UrlConnectionHttpClient extends AbstractHttpClient {
```
AdamBJohnsonx, line 72: Probably this needs some Android naming.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+                client.streamBufferSize);
+    }
+
+    /**
+     * Obtain a static default instance of the HTTP Client class.
+     *
+     * @return a default-configured HttpClient.
+     */
+    public static synchronized UrlConnectionHttpClient getDefaultInstance() {
+        UrlConnectionHttpClient reference = defaultReference.get();
```
AdamBJohnsonx, line 112: Change to lazy-holder pattern?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Supplier.java
```diff
@@ -0,0 +1,35 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.util.ported;
+
+/**
+ * A functional interface modeled off of java.util.function.Supplier for providing
+ * values to callers.
+ * @param <T> the type of value provided.
+ */
+public interface Supplier<T> {
```
AdamBJohnsonx, line 30: Why move this?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: I'm intentionally keeping this the same to avoid breaking change.

Once we're ready for a breaking change, we should not need this adapter class anymore.

rpdome: oops nvm. just noticed that this is the class in common-java.

What do you mean by "Android" naming in that case?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Supplier.java
rpdome: It looks like a util class (not specific to the class that held it).

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
AdamBJohnsonx: Unless you've changed it, there was a lot of code in here written with android-based assumptions in it that we likely don't want in a non-android platform.

common4j/src/main/com/microsoft/identity/common/java/util/ported/Supplier.java
AdamBJohnsonx: Yeah, but non-android, you can use the java.util.function one.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: I'm not seeing anything android-specific here. (I'm aware that there are network checks and stuff, but it's not here).


Could you give me an example?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Supplier.java
rpdome: Seems like this either need Java 1.8 or SDK 24...

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+
+        // Apply special backcompat behaviors for PATCH, if reqd
+        if (HttpClient.HttpMethod.PATCH == httpMethod) {
+            // Because HttpURLConnection predates RFC-5789, we need to fallback on POST w/ a backcompat
+            // workaround. See: https://stackoverflow.com/a/32503192/741827
+            httpMethod = HttpClient.HttpMethod.POST;
+            // This map may be immutable.
+            requestHeaders = new HashMap<>(requestHeaders);
+            requestHeaders.put("X-HTTP-Method-Override", HttpClient.HttpMethod.PATCH.name());
+        }
```
AdamBJohnsonx, line 214: This is android-specific.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+
+            while ((charsRead = reader.read(buffer)) > -1) {
+                stringBuilder.append(buffer, 0, charsRead);
+            }
+
+            return stringBuilder.toString();
+        } finally {
+            safeCloseStream(inputStream);
+        }
+    }
```
AdamBJohnsonx, line 248: This is kind-of android-specific.  Most other places would just deserialize bytes.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
AdamBJohnsonx: > Unless you've changed it, there was a lot of code in here written with android-based assumptions in it that we likely don't want in a non-android platform.

OK, less than I remembered.  One thing.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
AdamBJohnsonx: Nevermind.  It evidently is not.

AdamBJohnsonx (approved):
I don't have anything major here.  Overall, we want to move MSAL into being agnostic to which HttpClient it has been given, I think.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: This is it, right? 

We're using an AtomicReference here, and we're still only init the object during the first call.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -24,72 +24,34 @@
...
-
-import static com.microsoft.identity.common.internal.net.UrlConnectionHttpClient.Supplier;
 
 /**
  * Internal class for handling http request.
  */
-public final class HttpRequest {
-
-    private static final String HOST = "Host";
+public final class HttpRequest extends com.microsoft.identity.common.java.net.HttpRequest {
```
shahzaibj, line 39: Deprecated?

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -22,85 +22,35 @@
...
-
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import lombok.NonNull;
+
 /**
- * Internal class to wrap the raw server response, headers and status code.
+ * Deprecated
```
shahzaibj, line 32: Deprecated annotation?

rpdome: Intentionally not putting it, as @iambmelt called out earlier that it (or any warning) might break OneAuth.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MINOR] Migrate HTTP Layer (#1347)
```
iamgusain, line 3: >Migrate HTTP Layer [](http://example.com/codeflow?start=10&length=18)

Migrate HTTP Layer to common java lib

common/src/androidTest/java/com/microsoft/identity/common/HttpRequestTest.java
```diff
@@ -79,265 +77,137 @@ public void testNullRequestUrl() throws IOException {
     }
 
     /**
-     * Verify that HTTP GET succeed and no retry happens.
+     * Verify that HTTP GET succeed via HttpResponse.sendGet().
```
iamgusain, line 80: >via HttpResponse.sendGet(). [](http://example.com/codeflow?start=36&length=27)

We can add this part to testmethod name..
testHttpGetSucceedViaHttpResponseSendGet() ..

iamgusain: same at other places too, lets make the names do the documentation as much as possible

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -24,72 +24,34 @@
...
 
     /**
      * Value of read timeout in milliseconds.
+     * This is no longer used.
```
iamgusain, line 43: >This is no longer used. [](http://example.com/codeflow?start=7&length=23)

nit: can we also add what are we using instead

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -24,72 +24,34 @@
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import androidx.arch.core.util.Function;
+
+import com.microsoft.identity.common.java.net.HttpConstants;
```
iamgusain, line 28: >com.microsoft.identity.common.java.net.HttpConstants [](http://example.com/codeflow?start=7&length=52)

this is interesting.. Why not just use  com.microsoft.identity.common.HttpConstants.. 
Are we also following the folder structure as per this pacakge name.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -22,85 +22,35 @@
...
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-
-import net.jcip.annotations.Immutable;
-
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import lombok.NonNull;
```
iamgusain, line 29: in the HttpRequest.java we are keeping (not sure if they are used)
import androidx.annotation.NonNull;
import androidx.annotation.Nullable;

But i think we should be consistent?

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -0,0 +1,101 @@
...
+        mRequestContentType = requestContentType;
+    }
+
+    /**
+     * Check if the given status code is the retryable status code(500/503/504).
+     *
+     * @param statusCode The status to check.
+     * @return True if the status code is 500, 503 or 504, false otherwise.
+     */
+    public static boolean isRetryableError(final int statusCode) {
```
iamgusain, line 96: >isRetryableError [](http://example.com/codeflow?start=26&length=16)

This doesnt seem to belong HttpRequest.
Maybe HttpResponse or a separate Utility class if we have one.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+ */
+@AllArgsConstructor
+@Builder
+@ThreadSafe
+@Immutable
+public class UrlConnectionHttpClient extends AbstractHttpClient {
+
+    protected static final int RETRY_TIME_WAITING_PERIOD_MSEC = 1000;
+
+    protected static final int STREAM_BUFFER_SIZE = 1024;
```
iamgusain, line 76: >STREAM_BUFFER_SIZE [](http://example.com/codeflow?start=31&length=18)

STREAM_BUFFER_SIZE_BYTES

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+ * one integers.  If you use both of these, the suppliers method will take precendence over the method
+ * using integers.
+ */
+@AllArgsConstructor
+@Builder
+@ThreadSafe
+@Immutable
+public class UrlConnectionHttpClient extends AbstractHttpClient {
+
+    protected static final int RETRY_TIME_WAITING_PERIOD_MSEC = 1000;
```
iamgusain, line 74: >RETRY_TIME_WAITING_PERIOD_MSEC [](http://example.com/codeflow?start=31&length=30)

Shouldnt this be part of the Retry Policy?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+     * @param response
+     */
+    private static void recordHttpTelemetryEventEnd(final HttpResponse response) {
+        final HttpEndEvent httpEndEvent = new HttpEndEvent();
+
+        if (null != response) {
+            httpEndEvent.putStatusCode(response.getStatusCode());
+        }
+
+        Telemetry.emit(httpEndEvent);
```
iamgusain, line 168: >emit [](http://example.com/codeflow?start=18&length=4)

it will be nice to include requestId here if possible..
So that we can mark the begin and End more precisely by requestId

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+     */
+    private static void safeCloseStream(final Closeable stream) {
+        if (stream == null) {
+            return;
+        }
+
+        try {
+            stream.close();
+        } catch (final IOException e) {
+            //Encountered IO exception when trying to close the stream"
```
iamgusain, line 263: >//Encountered IO exception when trying to close the stream" [](http://example.com/codeflow?start=12&length=59)

Log Something, otherwise it will be harder to debug if we ever run in here. no?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+
+    private Integer getReadTimeoutMs() {
+        return readTimeoutMsSupplier == null ? readTimeoutMs : readTimeoutMsSupplier.get();
+    }
+
+    private Integer getConnectTimeoutMs() {
+        return connectTimeoutMsSupplier == null ? connectTimeoutMs : connectTimeoutMsSupplier.get();
+    }
+
+    private static void setRequestBody(@NonNull final HttpURLConnection connection,
```
iamgusain, line 337: >setRequestBody [](http://example.com/codeflow?start=24&length=14)

I guess this method also sends the request, right?
in this case we should rename the method name..

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+        if (!StringUtil.isNullOrEmpty(requestContentType)) {
+            connection.setRequestProperty("Content-Type", requestContentType);
+        }
+
+        connection.setRequestProperty("Content-Length", String.valueOf(contentRequest.length));
+
+        OutputStream out = null;
+
+        try {
+            out = connection.getOutputStream();
```
iamgusain, line 355: >out = connection.getOutputStream(); [](http://example.com/codeflow?start=12&length=35)

if this fails or throws.. does it set the response/input stream accordingly?
Do we have test cases covering that?

iamgusain: was just looking at this.. [https://stackoverflow.com/questions/4816824/httpurlconnection-post-conn-getoutputstream-throwing-exception](https://stackoverflow.com/questions/4816824/httpurlconnection-post-conn-getoutputstream-throwing-exception)

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+        final Set<Map.Entry<String, String>> headerEntries = request.getRequestHeaders().entrySet();
+
+        for (final Map.Entry<String, String> entry : headerEntries) {
+            urlConnection.setRequestProperty(entry.getKey(), entry.getValue());
+        }
+
+        urlConnection.setConnectTimeout(getConnectTimeoutMs());
+        urlConnection.setReadTimeout(getReadTimeoutMs());
+        urlConnection.setInstanceFollowRedirects(true);
+        urlConnection.setUseCaches(false);
```
iamgusain, line 323: >urlConnection.setUseCaches(false); [](http://example.com/codeflow?start=8&length=34)

this later gets overridden to false in the calling method..
Why are we setting it to true here? and what does it actually do

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+        try {
+            stream.close();
+        } catch (final IOException e) {
+            //Encountered IO exception when trying to close the stream"
+        }
+    }
+
+    private HttpResponse executeHttpSend(HttpRequest request, Consumer<HttpResponse> completionCallback) throws IOException {
+        final HttpURLConnection urlConnection = setupConnection(request);
+        urlConnection.setRequestMethod(request.getRequestMethod());
```
iamgusain, line 269: >setRequestMethod [](http://example.com/codeflow?start=22&length=16)

why not do this in the setupConnection method?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -0,0 +1,361 @@
...
+        urlConnection.setUseCaches(true);
```
iamgusain, line 270: >setUseCaches [](http://example.com/codeflow?start=22&length=12)

this was set to false in the setupConnection method, why change it now, instead of in the setupConnection method itself

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
iamgusain, line 72: >UrlConnectionHttpClient [](http://example.com/codeflow?start=13&length=23)

I am not seeing any telemetry for exceptions/errors in this class... unless it is getting captured in someotherway..
It is very important that we capture telemetry for different kinds of failures here.. to make debugging with logs easier

common4j/src/main/com/microsoft/identity/common/java/net/IRetryPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+ * @param <T> the type of the object on return.
+ */
+public interface IRetryPolicy<T> {
+    /**
+     * Evaluate the object returned from a callable and return the result.
+     * @param supplier an object to call for a result.
+     * @return the result of calling the supplier.
+     * @throws IOException if an IO error occurs.
+     */
+    T attempt (Callable<T> supplier) throws IOException;
```
iamgusain, line 42: >T attempt (Callable<T> supplier) throws IOException; [](http://example.com/codeflow?start=4&length=52)

do we plan to use this for anything other than Http?
If so then maybe we need to move it out of the om.microsoft.identity.common.java.net pacakge

If not I think we can just create a HttpRetryPolicy Otherwise it looks to be over engineering for something that we might not use

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import lombok.NonNull;
+
+/**
+ * Internal class representing a http request.
+ */
+public class HttpRequest {
```
iamgusain, line 36: >HttpRequest [](http://example.com/codeflow?start=13&length=11)

why not use standard java HttpRequest (hopefully there is one)

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
```diff
@@ -22,325 +22,186 @@
...
-    private final int connectTimeoutMs = 30000;
-    @Builder.Default
-    private final int readTimeoutMs = 30000;
-    @Builder.Default
-    private final Supplier<Integer> connectTimeoutMsSupplier = null;
-    @Builder.Default
-    private final Supplier<Integer> readTimeoutMsSupplier = null;
-    @Builder.Default
-    private final int streamBufferSize = 1024;
+    public static class Builder {
```
shahzaibj, line 64: we could probably use lombok here

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
```diff
@@ -22,325 +22,186 @@
...
-    public HttpResponse method(@NonNull final HttpMethod httpMethod,
-                               @NonNull final URL requestUrl,
-                               @NonNull final Map<String, String> requestHeaders,
-                               @Nullable final byte[] requestContent) throws IOException {
-        recordHttpTelemetryEventStart(httpMethod.name(), requestUrl, requestHeaders.get(CLIENT_REQUEST_ID));
-        final HttpRequest request = constructHttpRequest(httpMethod, requestUrl, requestHeaders, requestContent);
-        return retryPolicy.attempt(new Callable<HttpResponse>() {
-            public HttpResponse call() throws IOException {
-                return executeHttpSend(request, new Consumer<HttpResponse>() {
+    private com.microsoft.identity.common.java.net.UrlConnectionHttpClient sAdaptedObject;
```
shahzaibj, line 112: This does not seem static

```suggestion
    private com.microsoft.identity.common.java.net.UrlConnectionHttpClient mClientInstance;
```

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -277,7 +277,7 @@ public MicrosoftStsAccount createAccount(
         }
 
 
-        builder.setLibraryName(DiagnosticContext.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
+        builder.setLibraryName(DiagnosticContext.INSTANCE.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT));
```
shahzaibj, line 280: Did we make change to DiagnosticContext in this PR? If not how was this working until now?

common4j/src/main/com/microsoft/identity/common/java/net/HttpConstants.java
```diff
@@ -50,4 +50,4 @@
          */
         public static final String APPLICATION_JSON = "application/json";
     }
-}
+}
```
shahzaibj, line 53: new line

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -0,0 +1,101 @@
...
+        return mRequestMethod;
+    }
+
+    Map<String, String> getRequestHeaders() {
+        return Collections.unmodifiableMap(mRequestHeaders);
+    }
+
+    private final String mRequestContentType;
+    private final String mRequestMethod;
+    private final Map<String, String> mRequestHeaders = new HashMap<>();
```
shahzaibj, line 66: Can we move these to the top of the class along with other member fields?

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -0,0 +1,101 @@
...
+        return mRequestContentType;
+    }
+
+    String getRequestMethod() {
+        return mRequestMethod;
+    }
+
+    Map<String, String> getRequestHeaders() {
+        return Collections.unmodifiableMap(mRequestHeaders);
+    }
```
shahzaibj, line 62: Can we use lombok's getter annotation?

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+
+import lombok.NonNull;
+
+/**
+ * Internal class representing a http request.
+ */
+public class HttpRequest {
+
+    private static final String HOST = "Host";
```
shahzaibj, line 38: I probably haven't looked deeply into the current http request code...but what is this Host supposed to be represent?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
shahzaibj: I believe you mean Correlation Id? Telemetry is already aware of Correlation Id - it gets it from DiagnosticContext

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
rpdome: DiagContext's change was made in the Logger PR.

shahzaibj: Right. So I believe we are in a state right now where both with and without the `INSTANCE` would work?

rpdome: Yes. I'll eventually retire the old methods though. It's currently kept to avoid making a breaking change.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
rpdome: That one will eventually be retired (once we're ready for a breaking change).

Basically, I'll use the code in common.java for anything that doesn't touch the interface.

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
rpdome: I'm going to put this in UrlConnectionHttpClient.
Once we remove the adapter classes, that will be the only place this function is going to be called.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: This looks like a parameter for initializing that - not really part of the class itself.

rpdome: (The default value seems to be 1000 (See StatusCodeAndExceptionRetry.initialDelay)

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: As for the reason why it's like this, I port the code as-is, so I don't know 😅 

Will make a change as per your comment. It makes sense.

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
rpdome: We can't. Since the constructor of this `adapter` actually initializes common-java's UrlConnectionHttpClient.

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
rpdome: It's not :)

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
rpdome: There is one, but it doesn't map 1:1 with what this one does.

ps. The scope of this PR is to extract the existing code into common-java, so I'm intentionally avoiding making any bigger change. If we need further refactoring, we can wait until this 'guaranteed to work' logic is shipped as part of common-java, and then do the refactoring in that layer later.

common4j/src/main/com/microsoft/identity/common/java/net/IRetryPolicy.java
rpdome: Same reason as below. Intentionally kept as-is to reduce the scope of the work.

We can refactor this later.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: We can add that later. (We don't have a good telemetry story yet tbh).

I'll leave a todo here.

rpdome:
@AdamBJohnsonx  I accidentally re-requested for your review :swearing.

iambmelt:
(Looking at this PR now....)

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
```diff
@@ -99,59 +63,6 @@ public Boolean apply(Exception e) {
...
-
-    String getRequestContentType() {
-        return mRequestContentType;
-    }
-
-    String getRequestMethod() {
-        return mRequestMethod;
-    }
-
-    Map<String, String> getmRequestHeaders() {
```
iambmelt, line 135: I see we've made an effort to be backcompat here -- I see this method was renamed to `getRequestHeaders()` in the superclass so in order to truly maintain backcompat we should preserve `getmRequestHeaders()` here and just call super (since the naming is fixed in the superclass)

iambmelt (changes requested):
See a change that breaks backcompat -- since this PR is clearly making an effort to do this safely, requesting changes to keep it true to that spirit :wink:

https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1347#discussion_r626980927

common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java
rpdome: It looks like this method is package-private, so it shouldn't break MSAL/ADAL/Broker.

iambmelt: ACK

iambmelt (approved):

common4j/src/main/com/microsoft/identity/common/java/util/ported/Supplier.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * @return an instance of a T.x
+     */
+    T get();
+}
```
shahzaibj, line 35: new line

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: We're currently throwing IOException, which is base class of the SocketException.

Those exceptions are not handled in this layer.

PR #1352, opened 2021-05-03
Just rolling back the edge-triggered connection changes, in case.

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -130,22 +65,12 @@ public void unregisterNetworkCallback() {
...
-                TelemetryEventStrings.Key.NETWORK_CONNECTION,
-                String.valueOf(DefaultConnectionService.connectionAvailable)
-        ));
+        ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
+                .getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo activeNetwork = connectivityManager.getActiveNetworkInfo();
 
-        return DefaultConnectionService.connectionAvailable;
+        return activeNetwork != null &&
+                        activeNetwork.isConnectedOrConnecting();
```
AdamBJohnsonx, line 73:
```suggestion
        @SuppressWarnings("deprecation")
        final boolean isConnectionAvailable = activeNetwork != null && activeNetwork.isConnectedOrConnecting();
        Telemetry.emit((BaseEvent) new BaseEvent().put(TelemetryEventStrings.Key.NETWORK_CONNECTION, String.valueOf(isConnectionAvailable)));
        return isConnectionAvailable;
```

PR #1226, opened 2021-02-09
Fix deprecated NetworkInfo class.

gitaumoses4:
Fixes #847 

Updates code to use new callback model. [NetworkInfo](https://developer.android.com/reference/android/net/NetworkInfo)

PR #1353, opened 2021-05-03
Just rolling back the edge-triggered connection changes, in case.

AdamBJohnsonx:
In testing this out, I took an emulator running android 30 and S, and running the MSAL test app, worked through various acquire token flows while enabling or disabling the network.  No abnormal behavior was observed


iambmelt (approved):
This looks to be a revert of the below PR - approving. Can we document (in this PR, if nowhere else) how we expect this to behave on the API 30+/- devices?
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1226/files

Before I merge this, if I do, I'll figure that out.

PR #1361, opened 2021-05-06
Fix #1360

### LINKED ISSUE #1360: Unable to generate Javadoc for Logger setter method
We are unable to generate javadoc for the following method / line in our code: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java#L62

It seems we are using Lombok's experimental method annotations and they do not seem to play very well with the generating of the javadoc: https://projectlombok.org/features/GetterSetter

We get the following error:

`common\common4j\src\main\com\microsoft\identity\common\java\logging\Logger.java:62: error: cannot find symbol
    @Setter(onMethod_ = {@Synchronized})
            ^
  symbol:   method onMethod_()
  location: @interface Setter`

shahzaibj:
Fix #1360

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
@@ -59,13 +59,22 @@ public class Logger {
     @Accessors(prefix = "s")
     private static boolean sAllowPii = false;
 
-    @Setter(onMethod_ = {@Synchronized})
     @Accessors(prefix = "s")
     private static String sPlatformString = "";
 
     private static final ReentrantReadWriteLock sLoggersLock = new ReentrantReadWriteLock();
 
     private static final Map<String, ILoggerCallback> sLoggers = new HashMap<>();
+
+    /**
+     * Set the platform string to be used when generating logs.
+     *
+     * @param platformString the platform string to set
+     */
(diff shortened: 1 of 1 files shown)


shahzaibj:
> Do we want to avoid using getter setter from this point forward?

I think if there is no annotation on them then let's use them. The issue here seems to be using the `onMethod` to put `annotations` and that `onMethod` thing is an experimental feature so it is causing problems. Besides this thing seems to be JDK dependent so we definitely want to avoid this I think:

`	/**
	 * Any annotations listed here are put on the generated method.
	 * The syntax for this feature depends on JDK version (nothing we can do about that; it's to work around javac bugs).<br>
	 * up to JDK7:<br>
	 *  {@code @Setter(onMethod=@__({@AnnotationsGoHere}))}<br>
	 * from JDK8:<br>
	 *  {@code @Setter(onMethod_={@AnnotationsGohere})} // note the underscore after {@code onMethod}.
	 *  
	 * @return List of annotations to apply to the generated setter method.
	 */`

But in general I am fine with getter and setters as long as we are not using experimental features out of them.

rpdome (approved):

AdamBJohnsonx (approved):

PR #1367, opened 2021-05-10
Fix/Suppress SpotBugs issue

rpdome:
Suppress NM_SAME_SIMPLE_NAME_AS_SUPERCLASS. 
- This occurs when a child class and the super class has the same name, but belongs in a different packages.
- This is by design (to avoid breaking change - and I don't want to name the class in java with "Java" suffix either).

Suppress RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE
- This occurs with Lombok AllArgsConstructor.

The rest are fixed.
- Byte array and Date objects are returned as a copy.
- If date time is not provided, HttpResponse will use current Time (new Date()).
- InputStreamReader uses UTF-8. (Still can't use StandardCharsets.UTF_8) 
- Remove getReadTimeoutMs() and getConnectTimeoutMs() unnecessary boxing.
- Set params.getExtraParameters as a constant before iterating through it.
- Add null check for HttpRequest.sendWithMethod()'s response.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
@@ -28,11 +28,14 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 /**
  * Deprecated
  * <p>
  * Currently served as an adapter of {@link com.microsoft.identity.common.java.net.HttpResponse}
  */
+@SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS")
 public final class HttpResponse extends com.microsoft.identity.common.java.net.HttpResponse {
     public HttpResponse(@NonNull final com.microsoft.identity.common.java.net.HttpResponse responseToBeAdapted) {
         super(responseToBeAdapted.getDate(),
@@ -44,7 +47,7 @@ public final class HttpResponse extends com.microsoft.identity.common.java.net.H
     public HttpResponse(final int statusCode,
                         final String responseBody,
                         final Map<String, List<String>> responseHeaders) {
-        super(null, statusCode, responseBody, responseHeaders);
+        super(statusCode, responseBody, responseHeaders);
     }
 
     public HttpResponse(final Date date,
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
@@ -44,6 +44,7 @@ import static com.microsoft.identity.common.java.net.UrlConnectionHttpClient.DEF
  * <p>
  * Currently served as an adapter of {@link com.microsoft.identity.common.java.net.UrlConnectionHttpClient}
  */
+@edu.umd.cs.findbugs.annotations.SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS")
 public class UrlConnectionHttpClient extends AbstractHttpClient {
 
     /**
--- a/common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.java.net;
 
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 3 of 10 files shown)


common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -42,10 +43,17 @@
     @Accessors(prefix = "m")
     private final URL mRequestUrl;
 
-    @Getter
     @Accessors(prefix = "m")
     private final byte[] mRequestContent;
 
+    public byte[] getRequestContent() {
```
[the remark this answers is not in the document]
rpdome: Spotbugs doesn't like the idea of returning the original copy of the array (the value could be modified if an external actor holds a reference to it).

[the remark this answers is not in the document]
AdamBJohnsonx: Mostly, SpotBugs has good ideas.  It might be good to weigh this one - if you adopt this approach, we're going to want to make very sure that it isn't called repeatedly.  I was going to say - if this was a key or some kind of persistent object state, this is a very good idea to do.  For this one... Given what we do, it might be justified to copy it.

common4j/src/main/com/microsoft/identity/common/java/net/HttpRequest.java
```diff
@@ -42,10 +43,17 @@
...
+        if (mRequestContent == null) {
+            return null;
+        }
+        
+        return Arrays.copyOf(mRequestContent, mRequestMethod.length());
```
shahzaibj, line 54: This is probably breaking your tests

```suggestion
        return Arrays.copyOf(mRequestContent, mRequestContent.length);
```

shahzaibj (changes requested):
Left a comment

shahzaibj (approved):

rpdome:
> https://www.javadoc.io/doc/com.github.spotbugs/spotbugs-annotations/3.1.0/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html I was talking about the justification string in the annotation.

ah ok. Fixed

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
```diff
@@ -44,6 +44,7 @@
  * <p>
  * Currently served as an adapter of {@link com.microsoft.identity.common.java.net.UrlConnectionHttpClient}
  */
+@edu.umd.cs.findbugs.annotations.SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS")
```
AdamBJohnsonx:
```suggestion
@edu.umd.cs.findbugs.annotations.SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS",
         justification = "This class kept its original name to avoid breaking change during the refactoring process." +
                 "Once the process is done, this class will be removed entirely. ")
```

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
```diff
@@ -28,11 +28,14 @@
...
 import java.util.Map;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 /**
  * Deprecated
  * <p>
  * Currently served as an adapter of {@link com.microsoft.identity.common.java.net.HttpResponse}
  */
+@SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS")
```
AdamBJohnsonx:
```suggestion
@SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS",
         justification = "This class kept its original name to avoid breaking change during the refactoring process." +
                 "Once the process is done, this class will be removed entirely. ")
```

common/src/main/java/com/microsoft/identity/common/internal/net/UrlConnectionHttpClient.java
rpdome: already fixed in the commit above.

common/src/main/java/com/microsoft/identity/common/internal/net/HttpResponse.java
rpdome: already fixed in the commit above.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -95,8 +97,9 @@
     @Builder.Default
     private final int streamBufferSize = DEFAULT_STREAM_BUFFER_SIZE;
 
-    private static transient AtomicReference<UrlConnectionHttpClient> defaultReference = new AtomicReference<>(null);
+    private static final transient AtomicReference<UrlConnectionHttpClient> defaultReference = new AtomicReference<>(null);
 
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
```
AdamBJohnsonx, line 102:
```suggestion
    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
                        justification = "This is caused by lombok-generated code inserting a check")
```

PR #1372, opened 2021-05-12
Migrate eSTS Telemetry to common4j

rpdome:
## Changes
- Migrates EstsTelemetry and related classes to common4j 
    - The old class is kept as an adapter.
- EstsTelemetry now consumes interfaces (ICommand, ICommandResult, IBaseException, IServiceException) instead of the actual objects.
    - This allows us to only migrate this layer, not every objects that it touches.
    - It's also easier to write unit tests.
- Add a couple of null check and loggers in a few methods.
- Add a handful of unit tests for eSTS telemetry. (@shahzaibj  please double check if the results are valid 😅 Also let me know if I should add anything.)
- Remove @ Deprecated from Logger, so that it doesn't break OneAuth (for now).

Note: MSAL test case is now broken because of this PR. here's the fix.
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1394


common4j/build.gradle
```diff
@@ -140,12 +140,15 @@ dependencies {
 
     implementation "net.jcip:jcip-annotations:1.0"
     implementation "com.google.code.gson:gson:2.8.5"
+    implementation 'org.json:json:20180813'
 
     compileOnly "org.projectlombok:lombok:1.18.12"
     annotationProcessor "org.projectlombok:lombok:1.18.12"
 
+    testCompileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.12'
```
shahzaibj, line 148: Why do we need to add this specifically for test? Isn't it carried over from `compileOnly`?

common4j/src/main/com/microsoft/identity/common/java/commands/ICommandResult.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Returns the result status.
+     */
+    ResultStatus getStatus();
+
+    /**
+     * Gets the result object.
+     */
+    Object getResult();
+}
```
shahzaibj, line 60: new line

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,488 @@
...
+
+    EstsTelemetry() {
+        this(new TelemetryMap(), new SentFailedRequestsMap());
+    }
+
+    // Exposed for testing only.
+    EstsTelemetry(@NonNull final IKeyPairStorage<CurrentRequestTelemetry> telemetryMap,
+                  @NonNull final IKeyPairStorage<Set<FailedRequest>> sendFailedRequestsMap) {
+        mTelemetryMap = telemetryMap;
+        mSentFailedRequests = sendFailedRequestsMap;
```
shahzaibj, line 64:
```suggestion
                  @NonNull final IKeyPairStorage<Set<FailedRequest>> sentFailedRequestsMap) {
        mTelemetryMap = telemetryMap;
        mSentFailedRequests = sentFailedRequestsMap;
```

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        final LastRequestTelemetry lastRequestTelemetryFromCache = mLastRequestTelemetryCache.getRequestTelemetryFromCache();
+        if (lastRequestTelemetryFromCache == null) {
+            // we did not have anything in the telemetry cache for the last request
+            // let's create a new object based on the data available from the current request object
+            // and return the header string formed via that object
+            //
+            // We're trying to send this.. so that if we ever come across this field being null on the server
+            // then we know you have a bug in the client.
```
shahzaibj, line 368: I think this comment falls below on line 376

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/EstsTelemetry.java
```diff
@@ -0,0 +1,488 @@
...
+     **/
+    @Nullable
+    private CurrentRequestTelemetry getCurrentTelemetryInstance(final String correlationId) {
+        if (mTelemetryMap == null || correlationId == null || correlationId.equals("UNSET")) {
+            return null;
+        }
+
+        return mTelemetryMap.get(correlationId);
+    }
+}
```
shahzaibj, line 488: new line

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/SentFailedRequestsMap.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.microsoft.identity.common.java.interfaces.IKeyPairStorage;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import lombok.NonNull;
+
+public class SentFailedRequestsMap implements IKeyPairStorage<Set<FailedRequest>> {
+    final ConcurrentHashMap<String, Set<FailedRequest>> sendFailedRequestsMap = new ConcurrentHashMap<>();
```
shahzaibj, line 33:
```suggestion
    final ConcurrentHashMap<String, Set<FailedRequest>> sendtFailedRequestsMap = new ConcurrentHashMap<>();
```

shahzaibj, line 33: probably can be private

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryMap.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.microsoft.identity.common.java.interfaces.IKeyPairStorage;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+import lombok.NonNull;
+
+public class TelemetryMap implements IKeyPairStorage<CurrentRequestTelemetry>{
+
+    final ConcurrentHashMap<String, CurrentRequestTelemetry> telemetryMap = new ConcurrentHashMap<>();
```
shahzaibj, line 33: probably can be private

common4j/src/main/com/microsoft/identity/common/java/result/AcquireTokenResult.java
```diff
@@ -0,0 +1,44 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.java.result;
+
+public class AcquireTokenResult {
+
+    private ILocalAuthenticationResult mLocalAuthenticationResult;
+
+    private Boolean mSucceeded = false;
```
shahzaibj, line 29: As part of this refactor, can we switch to `boolean`?

common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
```diff
@@ -0,0 +1,158 @@
...
+public final class HashMapExtensions {
+
+    private static final String TAG = HashMapExtensions.class.getSimpleName();
+
+    private HashMapExtensions() {
+        // Intentionally left blank
+    }
+
+    /**
+     * decode url string into a key value pairs with default query delimiter.
```
shahzaibj, line 48:
```suggestion
     * Decode url string into a key value pairs with default query delimiter.
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -22,8 +22,39 @@
...
+import lombok.NonNull;
+
 public class StringUtil {
+    /**
+     * The constant ENCODING_UTF8.
+     */
+    public static final String ENCODING_UTF8 = "UTF-8";
+
+    /**
+     * checks if string is null or empty.
```
shahzaibj, line 37:
```suggestion
     * Checks if string is null or empty.
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -22,8 +22,39 @@
...
+     *
+     * @param message String to check for null or blank.
+     * @return true, if the string is null or blank.
+     */
     public static boolean isNullOrEmpty(String message) {
         return message == null || message.trim().length() == 0;
     }
```
shahzaibj, line 44: We can probably get rid of this method and just use `StringUtils.isNullOrEmpty`

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -0,0 +1,32 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package com.microsoft.identity.common.java.util.ported;
```
shahzaibj, line 24: ported as in ported from Android to java? Do we already have something similar in JAVA that we can use?

common4j/build.gradle
rpdome: nope. I actually had to do the same with Lombok.

common4j/src/main/com/microsoft/identity/common/java/result/AcquireTokenResult.java
rpdome: This is not currently used. I'll migrate in a separate PR.

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
rpdome: This requires including javafx and Java 9
https://docs.oracle.com/javase/9/docs/api/javafx/util/Pair.html

shahzaibj: ok, we certainly don't want that 😁

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: I can't find that method though.

shahzaibj: Ohh sorry, this is the that method 😂

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
AdamBJohnsonx: We could substitute AbstractMap.SimpleEntry<U, V> (which implements Map.Entry) for this, rather than writing our own or needing to so this port.  A thought.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
AdamBJohnsonx: At some point, we should write this to not construct a new string in order to determine if it only contains whitespace.

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
rpdome: @AdamBJohnsonx AbstractMap.SimpleEntry seems to be used only in SharedPrefFileManager, so that will never be ported to java-common.

rpdome: ah... it's the other way around.

rpdome: Looking into the code. I think I'll keep pair. AbstractMap.SimpleEntry would also work, but its naming (key, value) might confuse future readers.

AdamBJohnsonx:
Do we have ADO tasks reminding us to remove the adapters in MSAL 3.0/Common 4.0?

common4j/build.gradle
shahzaibj: I guess it probably works this way in compile

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
AdamBJohnsonx: Don't really care, but it's another class.  Any pure-java developer is likely familiar with map-entry syntax, FWIW.

shahzaibj (approved):
LGTM

common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
```diff
@@ -0,0 +1,158 @@
...
+import org.json.JSONObject;
+
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.StringTokenizer;
+
+public final class HashMapExtensions {
```
AdamBJohnsonx, line 39: Javadoc?

common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * get key value pairs from response.
+     *
+     * @param webResponse {@link HttpResponse} to convert to a map
+     * @return HashMap
+     * @throws JSONException
+     */
+    public static HashMap<String, String> getJsonResponse(HttpResponse webResponse) throws JSONException {
```
AdamBJohnsonx, line 120: Having these things return specifically HashMap is a bit strange.

AdamBJohnsonx (approved):
I think maybe some javadoc on the one class.

common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
rpdome: I'm removing some of the funcs in this class for now.. since we're not using them in this PR.

Including this one.

rpdome: actually.. it's the whole class. I think I might used it initially and scoped the size of the PR down.

rpdome:
Note: MSAL test case is now broken because of this PR
Fix is in: AzureAD/microsoft-authentication-library-for-android#1394

PR #1373, opened 2021-05-12
Revert "Testing change to see build behaviour."

aherciya:
Unintended commit to dev. 
This reverts commit de46964679bd573d419bfa160a8aa2d30f153448.


rpdome (approved):
As a best practice, I would suggest branching out before you start coding to avoid this issue. Always keep dev in your machine clean.

It's not a big deal for now, but we wouldn't want to accidentally slip the change when a release is being cut.

aherciya:
> As a best practice, I would suggest branching out before you start coding to avoid this issue. Always keep dev in your machine clean.
> 
> It's not a big deal for now, but we wouldn't want to accidentally slip the change when a release is being cut.

Yes I agree, I've made local system updates to avoid this in the future.

AdamBJohnsonx (approved):

PR #1375, opened 2021-05-13
Undo changes that breaks MSAL CPP

rpdome:
bring back these classes. Apparently internal classes are not so internal after all 😆 
 
We'll remove them (and introduce breaking changes) when we're ready to do a major release.

(Disclaimer: this is only making sure that it doesn't break MSALCPP, not OneAuth, since I can't get the latter set up on my machine yet)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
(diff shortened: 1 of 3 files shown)


AdamBJohnsonx (approved):

iambmelt:
I'd like to test these changes before signing off.... will ACK once I've done so

common/src/main/java/com/microsoft/identity/common/internal/net/AbstractHttpClient.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.NonNull;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Map;
+
+/**
+ * Deprecated
+ *
+ * Kept for backcompat purpose. Will be remove in the next major release.
```
iamgusain, line 34: >Will be remove in the next major release. [](http://example.com/codeflow?start=32&length=41)

May be better to add the Ado Task link here which is keeping track of files which need to be removed in next major release

iambmelt (approved):
Built, works ✔

PR #1377, opened 2021-05-13
Fixing the Authenticator RedirectURI in broker

tanmaymanolkar1:
Address the issue :
Broker redirect uri is invalid. Expected: urn:ietf:wg:oauth:2.0:oob Actual: msauth://.......


AdamBJohnsonx (approved):
Only question would be if the test coverage we have adequately covers this case.

tanmaymanolkar1:
> Only question would be if the test coverage we have adequately covers this case.


I am passing this RC1 to authenticator team to test this one out

AdamBJohnsonx:
Disabling branch protection to get around spurious test failure.

tanmaymanolkar1:
ran ./gradlew clean common:assemble locally whcih works, hence merging this

PR #1378, opened 2021-05-14
spotbugs task change

p3dr0rv:
* The spotbugs task change 
* I get rid of the values of spotbugsCommand because it will be hard to maintain 

@shahzaibj recently added product flavors (local dist) to the common gradle so now the spotbug task changed
':common'. Candidates are: 'spotbugsDistDebug', 'spotbugsDistRelease', 'spotbugsLocalDebug', 'spotbugsLocalRelease'.
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/90e5edd56c5bb909e928d9576e46e51ef478d7ca

### DIFF
--- a/azure-pipelines/pull-request-validation/common.yml
+++ b/azure-pipelines/pull-request-validation/common.yml
@@ -42,7 +42,6 @@ jobs:
   - template: ../templates/steps/spotbugs.yml
     parameters:
       project: common
-      spotbugsCommand: spotbugsDebug
   - task: Gradle@2
     displayName: Run Unit tests
     inputs:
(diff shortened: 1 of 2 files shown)


AdamBJohnsonx:
Can we have a little bit on the reasoning behind this change?

shoatman (approved):

AdamBJohnsonx (approved):

iamgusain:
Where does it pick the flavor/variant from now?

p3dr0rv:
> Where does it pick the flavor/variant from now?

@iamgusain 
the default is spotbugsLocalDebug if we want want change it we can add a variable and add the logic to pick the value from the variable if a value is set.
if you are interested please let me know and we can make it

PR #1379, opened 2021-05-16
Add new settings to control authorization behavior

shoatman:
- Added 2 new settings to MSAL Config: concurrent_authorization_requests (boolean), authorization_in_current_task (boolean)
- Added the first to InteractiveTokenParameters
- Used the second to modify the behavior of command dispatcher to allow the number of threads in the interactive thread pool to be increased via a new command dispatcher configuration class

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
@@ -85,7 +85,7 @@ public class InteractiveTokenCommandParameters extends TokenCommandParameters {
         return handleNullTaskAffinity;
     }
 
-    public boolean authorziationInCurrentTask(){
+    public boolean authorizationInCurrentTask(){
         return authorizationInCurrentTask;
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
@@ -74,12 +74,19 @@ public class InteractiveTokenCommandParameters extends TokenCommandParameters {
     @Expose()
     private final boolean handleNullTaskAffinity;
 
+    @Expose()
+    private final boolean authorizationInCurrentTask;
+
     private final List<Pair<String, String>> extraQueryStringParameters;
 
     private final List<String> extraScopesToConsent;
 
     public boolean getHandleNullTaskAffinity(){
         return handleNullTaskAffinity;
+    }
+
+    public boolean authorziationInCurrentTask(){
+        return authorizationInCurrentTask;
     }
 
     public List<Pair<String, String>> getExtraQueryStringParameters() {
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -141,6 +141,8 @@ public class CommandDispatcher {
                     1, config.getMaxTheadPoolInteractive(), -1, 0, TimeUnit.MINUTES, "interactive"
             );
             sConfigured = true;
+        }else{
+            Logger.warn(TAG, "CommandDispatcher was already configured.");
         }
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
@@ -135,7 +135,7 @@ public class CommandDispatcher {
         sExecutingCommandMap = newMap;
     }
 
-    public static void configureCommandDispatcher(CommandDispatcherConfiguration config){
+    public static synchronized void configureCommandDispatcher(CommandDispatcherConfiguration config){
         if(!sConfigured) {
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -132,6 +135,17 @@ private static void cleanMap(BaseCommand command) {
         sExecutingCommandMap = newMap;
     }
 
+    public static void configureCommandDispatcher(CommandDispatcherConfiguration config){
```
AdamBJohnsonx, line 138: Since this method is public, we're probably safest to synchronize it.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcherConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+@Getter
+@EqualsAndHashCode
+@SuperBuilder(toBuilder = true)
+/**
+ * Class for encapsulating command dispatcher configuration
+ */
+public class CommandDispatcherConfiguration {
+
+    @Expose()
+    private int maxTheadPoolInteractive;
```
AdamBJohnsonx, line 41: Do we want to follow out recommended naming conventions, and have this be mMaxThreadPoolInteractive or something?

shoatman: I had seen that lombok then use the same thing with the getter... so i'll check if that's the case... if so... i'll leave as is; otherwise i'll use the standard naming convention.

shoatman: Yes.  Lombok then makes the builder method have the m prefix... i'm leaving as is.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Done.  Thanks!

AdamBJohnsonx: OK, now that we've put the writes to this reference under the class-level monitor, we have to do the same thing to all the reads.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
```diff
@@ -82,6 +85,10 @@ public boolean getHandleNullTaskAffinity(){
         return handleNullTaskAffinity;
     }
 
+    public boolean authorziationInCurrentTask(){
```
gitaumoses4, line 88:
```suggestion
    public boolean authorizationInCurrentTask(){
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
shoatman: Sorry Adam.  Would you mind being a bit more verbose in your feedback.  Is the suggestion that we need to synchronize access to the static variable for the thread pool?  Why would need to be synchronized on reads in this case?  Also, what are you suggesting that we need to prevent.  This is not a public method that intended to be called by anyone else than us.  So not part of the public API... 

Strictly speaking this should be set once per process and never changed.  I wasn't prepared at this point to turn command dispatcher into singleton, but I think that's the more correct approach.

AdamBJohnsonx: What I'm saying is that, yes, you should synchronize all access to the static reference, and under the same monitor.  I'd like to avoid any self-enforcing singletons going forward, but that's a bigger configuration->client construction task than I think we're prepared to tackle here, so I'm OK with it.  But this really means finding the place (I think it's one, maybe two) where the access to the static variable is taken and make sure that it's under the same monitor (in this case, the class) where it is written.

And this isn't about necessarily protecting us from our clients, though that helps, it's protecting us from our future selves.  I mean, spotBugs will flag this anyway as inconsistent synchronization.

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcherConfiguration.java
rpdome: @shoatman you can add `@Accessors(prefix = "m")` to avoid that issue.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/InteractiveTokenCommandParameters.java
```diff
@@ -74,6 +74,9 @@
     @Expose()
     private final boolean handleNullTaskAffinity;
 
+    @Expose()
+    private final boolean authorizationInCurrentTask;
```
rpdome, line 78: Would you mind adding comments to describe what this is for? (mainly for our future selves)

common/src/main/java/com/microsoft/identity/common/internal/controllers/CommandDispatcher.java
```diff
@@ -132,6 +135,17 @@ private static void cleanMap(BaseCommand command) {
         sExecutingCommandMap = newMap;
     }
 
+    public static synchronized void configureCommandDispatcher(CommandDispatcherConfiguration config){
```
rpdome, line 138: nit: comments.

PR #1381, opened 2021-05-18
Restore slice and DC special handling to broker

AdamBJohnsonx:
We used to have slice and dc parameters propagate through the broker.  We stopped that, which is confusing.  Users can completely bypass this modeled method, by putting them in the extraQueryParamters which is used as a bag to transmit them across to the broker anyway.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
@@ -39,6 +39,7 @@ import com.microsoft.identity.common.logging.Logger;
 import java.util.List;
 import java.util.Locale;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
@@ -51,6 +52,7 @@ import static com.microsoft.identity.common.internal.authorities.AllAccounts.ALL
 import static com.microsoft.identity.common.internal.authorities.AnyPersonalAccount.ANY_PERSONAL_ACCOUNT_TENANT_ID;
 
 @SuperBuilder
+@SuppressFBWarnings(value = "RCN", justification = "Lombok-generated code creates redundant checks")
 @EqualsAndHashCode
 @NoArgsConstructor
 @AllArgsConstructor
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
@@ -39,9 +39,23 @@ import com.microsoft.identity.common.logging.Logger;
 import java.util.List;
 import java.util.Locale;
 
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.Accessors;
+import lombok.experimental.SuperBuilder;
+
 import static com.microsoft.identity.common.internal.authorities.AllAccounts.ALL_ACCOUNTS_TENANT_ID;
 import static com.microsoft.identity.common.internal.authorities.AnyPersonalAccount.ANY_PERSONAL_ACCOUNT_TENANT_ID;
 
+@SuperBuilder
+@EqualsAndHashCode
+@NoArgsConstructor
+@AllArgsConstructor
+@Accessors(prefix = "m")
+@ToString(callSuper = true)
 public abstract class AzureActiveDirectoryAudience {
 
     private static final String TAG = AzureActiveDirectoryAudience.class.getSimpleName();
@@ -157,7 +171,7 @@ public abstract class AzureActiveDirectoryAudience {
     }
 
     public static AzureActiveDirectoryAudience getAzureActiveDirectoryAudience(final String cloudUrl,
-                                                                               final String tenantId) {
+                                                                               final @NonNull String tenantId) {
         final String methodName = ":getAzureActiveDirectoryAudience";
         AzureActiveDirectoryAudience audience = null;
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
@@ -173,7 +173,7 @@ public abstract class AzureActiveDirectoryAudience {
     }
 
     public static AzureActiveDirectoryAudience getAzureActiveDirectoryAudience(final String cloudUrl,
-                                                                               final @NonNull String tenantId) {
+                                                                               final String tenantId) {
         final String methodName = ":getAzureActiveDirectoryAudience";
         AzureActiveDirectoryAudience audience = null;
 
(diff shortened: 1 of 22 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -46,6 +46,8 @@
     private static final long serialVersionUID = -543392127065130474L;
 
     private class SerializedNames {
+        static final String DC = "data_center";
+        static final String SLICE = "slice";
```
AdamBJohnsonx, line 50:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -105,11 +106,28 @@
...
+            if (slice != null) {
+                extraQueryStringParameters.add(new Pair(AzureActiveDirectorySlice.SLICE_PARAMETER, slice));
+            }
+            String dc = sliceData.getDC();
+            if (dc != null) {
+                extraQueryStringParameters.add(new Pair(AzureActiveDirectorySlice.DC_PARAMETER, dc));
+            }
+        }
+
+        final String extraQueryStringParameter = (queryParams != null || sliceData != null) ?
```
AdamBJohnsonx, line 125:
```suggestion
        final String extraQueryStringParameter = (extraStringQueryParameters.size() > 0) ?
```

titorodi:
val acquireTokenParameters = AcquireTokenParameters.Builder()
                .startAuthorizationFromActivity(activity)
                .withScopes(arrayListOf("UserAuthenticationMethod.Read"))
                .withAuthorizationQueryStringParameters(arrayListOf(android.util.Pair("dc","ESTS-PUB-WCUSC-LZ1-FIRST")))
                .withCallback(callback)
                .build()

^ allowed me to test targeting a DC without slice support.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
titorodi: 👍🏻

titorodi (approved):
:shipit:

shahzaibj:
Do we know when we removed it? Can we link the PR that removed it?

titorodi:
This is the config we expected to work. Is there a config file users can copy paste from for our happier future selves?

  "authorities": [
    {
      "type": "AAD",
      "audience": {
        "type": "AzureADMyOrg",
        "tenant_id": "e603d33d-36ba-4303-89fb-8376a0e36c0e"
      },
      "default": true,
      "slice": {
        "slice": "test",
        "dc": "ESTS-PUB-WCUSC-LZ1-FIRST"
      }
    }
  ]

shahzaibj:
> This is the config we expected to work. Is there a config file users can copy paste from for our happier future selves?
> 
> "authorities": [
> {
> "type": "AAD",
> "audience": {
> "type": "AzureADMyOrg",
> "tenant_id": "e603d33d-36ba-4303-89fb-8376a0e36c0e"
> },
> "default": true,
> "slice": {
> "slice": "test",
> "dc": "ESTS-PUB-WCUSC-LZ1-FIRST"
> }
> }
> ]

@titorodi https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-configuration#example-basic-configuration

AdamBJohnsonx:
> Do we know when we removed it? Can we link the PR that removed it?

As far back as a year ago, this was the state it was in.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -105,11 +106,28 @@
...
+
+        final AzureActiveDirectorySlice sliceData = parameters.getAuthority().getSlice();
+        if (sliceData != null) {
+            String slice = sliceData.getSlice();
+            if (slice != null) {
+                extraQueryStringParameters.add(new Pair(AzureActiveDirectorySlice.SLICE_PARAMETER, slice));
+            }
+            String dc = sliceData.getDC();
+            if (dc != null) {
+                extraQueryStringParameters.add(new Pair(AzureActiveDirectorySlice.DC_PARAMETER, dc));
```
iamgusain, line 121: >extraQueryStringParameters [](http://example.com/codeflow?start=16&length=26)

Should we add a UT that verifies that the final request contains these query params..

iamgusain:
is it possible to ship this change under a flight?

---
In reply to: [843646396](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1381#issuecomment-843646396) [](http://example.com/codeflow?ancestors=843646396,843646396)

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -105,11 +106,28 @@
...
 
-        final String extraQueryStringParameter = parameters.getExtraQueryStringParameters() != null ?
-                QueryParamsAdapter._toJson(parameters.getExtraQueryStringParameters())
+        final List<Pair<String, String>> queryParams = parameters.getExtraQueryStringParameters();
+        final List<Pair<String, String>> extraQueryStringParameters = queryParams != null ?
+                new ArrayList<>(queryParams) : new ArrayList<Pair<String, String>>();
+
+        final AzureActiveDirectorySlice sliceData = parameters.getAuthority().getSlice();
+        if (sliceData != null) {
+            String slice = sliceData.getSlice();
```
rpdome, line 115: nit: final, here and below.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -105,11 +106,28 @@
     public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final InteractiveTokenCommandParameters parameters) {
```
rpdome, line 106: You might want to double check if these values are used everywhere.

For one, constructTokenEndpointForAcquiringPRT() is not using them.

shahzaibj: AFAIK, they are NOT used in most places...that's why we had developed the command line mechanism...those ones are injected everywhere.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
AdamBJohnsonx: This I don't care about here.  I'm writing a UT that the translated InteractiveTokenRequestParamaters contains the relevant values when round-tripped through this code, and that's quite involved enough.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
AdamBJohnsonx: They're not, but that's only because these request construction mechanisms don't use common.  That command line mechanism doesn't work for people who want to use this feature without building it, does it?  FWIW, I was able to recommend an alternate path - just pass the query parameters, and that works without this change.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -157,7 +173,7 @@ public void setTenantId(String tenantId) {
     }
 
     public static AzureActiveDirectoryAudience getAzureActiveDirectoryAudience(final String cloudUrl,
-                                                                               final String tenantId) {
+                                                                               final @NonNull String tenantId) {
```
AdamBJohnsonx, line 176:
```suggestion
                                                                               final String tenantId) {
```

shahzaibj:
Common Auth Client pipeline is failing

PR #1382, opened 2021-05-19
Add method to check if authority belong to same cloud as another authority

iamgusain:
Adding a method to compare if current authority belongs to same cloud as another authority.
Using getAutorityURL to do the comparison
getAuthorityURL returns a url with mapped preferred_network_host_name as the host part of the url
So for two authorities from same cloud getAuthorityURL should return the same host name

Tests:
Added Unit tests to verify failure and success case for the new method


common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -173,4 +173,15 @@ public AzureActiveDirectoryAudience getAudience() {
...
 
+    /**
+     * Checks if current authority belongs to the same cloud as the passed in authority
+     * @param authorityToCheck authority to check against
+     * @return true if the preferred network host name matches for both authorities, otherwise false
+     */
+    public boolean isSameCloudAsAuthority(@NonNull final AzureActiveDirectoryAuthority authorityToCheck){
+        final AzureActiveDirectoryCloud currentAuthorityCloud = AzureActiveDirectory.getAzureActiveDirectoryCloud(this.getAuthorityURL());
+        final AzureActiveDirectoryCloud authorityToCheckCloud = AzureActiveDirectory.getAzureActiveDirectoryCloud(authorityToCheck.getAuthorityURL());
+        return currentAuthorityCloud.getPreferredNetworkHostName().equals(authorityToCheckCloud.getPreferredNetworkHostName());
```
shahzaibj, line 184: we should check if we can do the following:

```java
  return this.getAzureActiveDirectoryCloud().equals(authorityToCheck.getAzureActiveDirectoryCloud())
```

this would mean exposing the `getAzureActiveDirectoryCloud` method and also adding `equals` to the class `AzureActiveDirectoryCloud`

shoatman: I agree with Shahzaib's suggestion.

AdamBJohnsonx: If we're overriding the equals method, I'd like to make sure that we're not going to cause any unforeseen behavior later on.  If we implement equals (and almost every class we create should implement equals unless there is a very good reason), we *must* implement hashCode, and the two *must* be consistent.

shahzaibj: +1...we can probably just use Lombok's annotations in most cases and let it do the work.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -75,32 +90,18 @@ private void getAzureActiveDirectoryCloud() {
...
-
     public boolean getMultipleCloudsSupported() {
         return mMultipleCloudsSupported;
     }
 
+    public void setMultipleCloudsSupported(boolean supported) {
+        mMultipleCloudsSupported = supported;
+    }
+
     @Override
```
iamgusain, line 105: Any idea what different purposes these two methods solve getAuthorityUri vs getAuthorityURL?
I am thinking we can just use one.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: I updated the code to use getAuthorityURL as that includes the preferred_network_host_name.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -173,4 +174,21 @@ public AzureActiveDirectoryAudience getAudience() {
         return mAudience;
     }
 
+    /**
+     * Checks if current authority belongs to the same cloud as the passed in authority
+     *
+     * @param authorityToCheck authority to check against
+     * @return true if host name matches for authorityURL of both authorities, otherwise false
+     */
+    public boolean isSameCloudAsAuthority(@NonNull final AzureActiveDirectoryAuthority authorityToCheck) throws IOException {
```
iambmelt, line 183:
```suggestion
    @WorkerThread
    public boolean isSameCloudAsAuthority(@NonNull final AzureActiveDirectoryAuthority authorityToCheck) throws IOException {
```

`performCloudDiscovery` does network i/o, we should provide some hint you can't call this from any arbitrary thread

iambmelt: This annotation requires an import from `androidx.annotation.WorkerThread`

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -173,4 +174,21 @@ public AzureActiveDirectoryAudience getAudience() {
...
+        if (!AzureActiveDirectory.isInitialized()) {
+            // Cloud discovery is needed in order to make sure that we have a preferred_network_host_name to cloud aliases mappings
+            AzureActiveDirectory.performCloudDiscovery();
+        }
+
+        // getAuthorityURL returns a url with mapped preferred_network_host_name as the host part of the url
+        // So for two authorities from same cloud getAuthorityURL should return the same host name
+        return this.getAuthorityURL().getHost().equals(authorityToCheck.getAuthorityURL().getHost());
```
iambmelt, line 191:
```suggestion
        return getAuthorityURL().getHost().equals(authorityToCheck.getAuthorityURL().getHost());
```

Doesn't appear to be needed

common/src/test/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthorityTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(RobolectricTestRunner.class)
+public class AzureActiveDirectoryAuthorityTests {
+    @Test
+    public void isSameCloudAsAuthority_Returns_True_For_Authority_With_ValidAliases_For_SameCloud() throws IOException {
```
iambmelt, line 37: Nice ✔

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iambmelt, line 183: At a glance, we might want to `synchronize` this call...

iambmelt (changes requested):
Minor edits suggested

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: If you look at getAuthorityURL it returns the preferred_network_host_name, which should take care of alias checking

The thing I am still debating in my head is should we expose getAzureActiveDirectoryCloud from this class or should it be just the preferred_network_host_name which is what is being used for cloud comparison and is already being returned in getAuthorityURL.
Thoughts?

When we want to have equals method in the AzureActiveDirectoryCloud class.. what all should it compare?

shahzaibj: I would personally compare the following:

```java
    @SerializedName("preferred_network")
    private final String mPreferredNetworkHostName;

    @SerializedName("preferred_cache")
    private final String mPreferredCacheHostName;

    @SerializedName("aliases")
    private List<String> mCloudHostAliases;
```

then you know the clouds are truly equal...the worrying this is that `mCloudHostAliases` is a List...we should refactor that info a `Set`

AdamBJohnsonx: Why is that worrying?  Does it actually matter?

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
AdamBJohnsonx: We might like to shrink the scope of use of Android's Uri implementation.  It's cool, but it's not very testable and java has one that's not great, but not terrifyingly difficult to work with.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: I am assuming that CloudDiscovery should have happened by the time we call this method.. But I am not 100% sure if I can make that assumption for all scenarios.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: I am not sure if comparing aliases in needed to ensure if a authority belongs to same cloud as another authority, as long as the preferred_network matches. 
Could there be cases where preferred_network but not the aliases?

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: I am not sure if we want to use both Uri and URL... isnt just one of them sufficient for us?

The thing that adds to my confusion is that both these classes has their own definition of authority (hostname:port). Which we are using at places and it gets confused with our own definition of Authority (hostname/tenant)

So we have code like... authority.getAuthorityURL().getAuthority()

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
iamgusain: @shahzaibj can you think of a scenario where the preferred_network name is same but the alias list is different?  

I am inclined to not expose extra properties if we don’t need to.

iamgusain:
Had an offline discussion with Adam and Shahzaib related to this change, I will be sending out a new PR for this change.
Closing this one.

PR #1384, opened 2021-05-19
Move vsts username and password to project level build.gradle

shahzaibj:
Address: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1402810

cc/ @AdamBJohnsonx 

The username and password needs to be declared in the top-level build.gradle file where the credentials to the ADAL Android repository are specified. For some reason this top level file was referencing constant names that look like they were meant for MSAL and they clearly weren't being supplied in the pipelines (pipelines were supplying a constant name for common - `ENV_VSTS_MVN_ANDROIDCOMMON_ACCESSTOKEN` as opposed to `ENV_VSTS_MVN_ANDROID_MSAL_ACCESSTOKEN`) and this discrepancy was causing common to not be able to fetch `common4j` from internal maven and was causing build failures. So far this wasn't a problem because `common` never consumed anything from internal maven so we didn't notice this issue...but now that we are consuming something from internal maven (common4j) this problem has surfaced up.


rpdome:
@shahzaibj  seems like this breaks pr pipeline.

shahzaibj:
Converted back to draft to fix failing pipeline..

shahzaibj:
> @shahzaibj seems like this breaks pr pipeline.

@rpdome Fixed!

AdamBJohnsonx (approved):

AdamBJohnsonx:
Maybe some commentary in the build files is appropriate, as well.

p3dr0rv (approved):

PR #1271, opened 2021-03-17
Performance: Avoid Repeat Calls to getCredentials()

iambmelt:
This PR aims to improve the performance of `getAccounts()`.

By avoiding repeated calls to `getCredentials()` we can avoid iterating over the cache + keys multiple times.

>Note: this logic will still result in multiple `List` traversals of the cached contents when calling `getCredentialsFilteredByInternal()` however, we avoid any repeat decryption/cache accesses with this change. Refactoring  `getCredentialsFilteredByInternal()` to directly accept a `Set<CredentialType>` is a less straightforward change, so not attempting that as part of this PR.

Cache tests continue to pass with this change.

PR #1229, opened 2021-02-10
Unbreak mockito tests for cache

iambmelt:
Something was broken between incrementing the mockito version and reversion back/forth between Java 1.7 & 1.8 source compat. Something inside of Mockito needs to be desugared when using `Mockito.openMocks()` so for now switching away from an annotation based test implementation

PR #1385, opened 2021-05-19
IdToken & BAM Lookup Uses Faster Cache API to Avoid Extra Decryption

iambmelt:
Moves idtoken lookup over to newer lookup api to avoid hitting cache multiple times (1x per tenant) when an account exists in multiple tenants

Builds on:
- [New cache API to avoid repeat decryption](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1271)
- [Test fixes](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1229)


rpdome (approved):

tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -899,14 +899,17 @@ public RefreshTokenRecord getFamilyRefreshTokenForHomeAccountId(@NonNull final S
                                                            @NonNull AccountRecord accountRecord) {
         final List<IdTokenRecord> result = new ArrayList<>();
 
+        final List<Credential> allCredentials = mAccountCredentialCache.getCredentials();
```
AdamBJohnsonx, line 902: So, thing is, the only time that it should be decrypting these values is when there is more than 256 entries in the cache.  We could tune that value.

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: ACK - I'll go ahead and merge this change in as-is. I would love to get take your input very soon on identifying and weighing bang-for-buck opportunities to squeeze even more performance out of this layer. Work I am happy to do and think it will benefit from more eyes / creative solutions

iambmelt:
Oddly, I cannot merge this PR despite 3 approvals....

iambmelt:
@iamgusain Do you know why this PR cannot be merged? From looking at our code owner enforcement rules, this appears to be acceptable....

iambmelt:
Dismissing prior reviews because I may have gone too crazy in these latest changes 😆

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1201,20 +1204,33 @@ public ICacheRecord getAccountWithAggregatedAccountDataByLocalAccountId(
 
         final List<BrokerApplicationMetadata> allMetadata = mApplicationMetadataCache.getAll();
 
+        // TODO - Everything inside this loop can be parallelized... should it be?
```
AdamBJohnsonx, line 1207: We should NOT do this until we can measure it.  If there aren't enough execution units to manage it, it will be slower than not.

iambmelt: ACK - holding....

shahzaibj: I believe we don't need to wait for Paul's perf work to measure this specific thing...that's more E2E perf measurement...over here I think we can just run a test with both old and new code and just capture numbers in the test and assert that the new code takes < time than old code (we can probably capture and assert on a percentage increase)

Also this is a sort of change that I think is also easy to flight...so if we have ability to do that now then we should try that too.

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1668,26 +1684,9 @@ private static MicrosoftFamilyOAuth2TokenCache initializeFociCache(@NonNull fina
...
-                                                        final int callingProcessUid) {
-        final String methodName = ":getTokenCacheForClient";
-
-        final BrokerApplicationMetadata metadata = mApplicationMetadataCache.getMetadata(
-                clientId,
-                environment,
-                callingProcessUid
-        );
+    private MsalOAuth2TokenCache getTokenCacheForClient(@Nullable final BrokerApplicationMetadata metadata) {
+        final String methodName = ":getTokenCacheForClient (1 arg)";
```
shahzaibj, line 1689: seems like an attempt to distinguish an overload....how do we disntiguish the following?

```java
 private void methodName(TypeA input)
```

```java
 private void methodName(TypeB input)
```

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1718,6 +1717,32 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
         return targetCache;
     }
 
+    /**
+     * Returns the TokenCache to use for supplied client and environment.
```
shahzaibj, line 1721:
```suggestion
     * Returns the {@link OAuth2TokenCache} to use for supplied client id and environment.
```

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
AdamBJohnsonx: We don't really have that ability, we could set it up, but nothing is populating the flight info right now, and we're not integrate with any metrics system that I know of.  The only thing I'm cautioning on is the parallelization.  Do we really know what we're getting into here, and will it actually have the effect we want?  If we want to tackle this, - don't create new thread pools to handle it each time, don't used a cached pool, we MUST measure it, and we need to monitor it.

iambmelt: Reverted these changes in 6c0625b886331da1e58d55f8604965b22d81fc4d -- not going to try this until we can measure/evaluate

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
iambmelt: Made less ambiguous in ceea53560b6510d9d363a9fbc49c7dff5812da74

iambmelt:
Have backed out the more questionable changes from this PR and clarified log tags... rerequesting reviews

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -899,14 +899,17 @@ public RefreshTokenRecord getFamilyRefreshTokenForHomeAccountId(@NonNull final S
...
+
         final List<Credential> idTokens = mAccountCredentialCache.getCredentialsFilteredBy(
```
rpdome, line 904: nit: please add comments on the List<Credential> parameter for this overload

rpdome (approved):

shahzaibj (approved):

iamgusain:
As you are touching the "cache" area, It needs signoff from code owners for cache..  you and Shane at the moment.
And it doesnt allow you to signoff your own PR. 

---
In reply to: [844639652](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1385#issuecomment-844639652) [](http://example.com/codeflow?ancestors=844639652,844639652)

tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt, line 902:
```suggestion
        // Load all the credentials to inspect once, such that we don't need to requery the cache
        // pass these into the new getCredentialsFilteredBy overload, rather than hit disk again
        final List<Credential> allCredentials = mAccountCredentialCache.getCredentials();
```

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
iambmelt: 82e4ab0c9065759ed3aad0bda5b50eec3f3a2b7e

iambmelt:
> As you are touching the "cache" area, It needs signoff from code owners for cache.. you and Shane at the moment.
> And it doesnt allow you to signoff your own PR.
> 
> In reply to: [844639652](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1385#issuecomment-844639652)

@iamgusain I am going to add you as a codeowner then, because in this state I can't merge

iamgusain (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1718,6 +1698,32 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
         return targetCache;
     }
 
+    /**
+     * Returns the {@link MsalOAuth2TokenCache} to use for supplied client id and environment.
+     *
+     * @param clientId          The target client id.
+     * @param environment       The target environment.
+     * @param callingProcessUid The uid of the calling process.
```
shoatman, line 1706: Probably worth calling out somewhere that this identifier is unique and assigned at installation time.  With that said I don't think anything stops these identifiers from being recycled.

shoatman (approved):
Fee free to ignore my comment.

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
iambmelt, line 1706:
```suggestion
     * @param callingProcessUid The uid of the calling process. This value is unique to the calling app and is assigned during installation.
```

iambmelt: b70689bf406336908d169895bc1f7408db842c96

PR #1391, opened 2021-05-26
Tests for DefaultConnectionService

gitaumoses4:
## What does this PR do?
- Adds tests for the `DefaultConnectionService` class.

## Description of tasks to be completed.
- [x] Use robolectric to test the DefaultConnectionService

## How to test this PR?
- Run the test files that were created.

## Any addition comments.
N/A

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapperShadow.java
+++ b/common/src/test/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapperShadow.java
@@ -31,16 +31,13 @@ import org.robolectric.annotation.Implements;
 @Implements(PowerManagerWrapper.class)
 public class PowerManagerWrapperShadow {
 
-    public static boolean deviceIdleMode = true;
-    public static boolean ignoringBatteryOptimizations = true;
-
     @Implementation
     public boolean isDeviceIdleMode(final Context connectionContext) {
-        return deviceIdleMode;
+        return true;
     }
 
     @Implementation
     public boolean isIgnoringBatteryOptimizations(final Context connectionContext) {
-        return ignoringBatteryOptimizations;
+        return false;
     }
 }
--- a/common/src/test/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapperShadow.java
+++ b/common/src/test/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapperShadow.java
@@ -1,0 +1,46 @@
+
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.adal.internal;
+
+import android.content.Context;
+
+import org.robolectric.annotation.Implementation;
+import org.robolectric.annotation.Implements;
+
+@Implements(PowerManagerWrapper.class)
+public class PowerManagerWrapperShadow {
+
+    public static boolean deviceIdleMode = true;
+    public static boolean ignoringBatteryOptimizations = true;
+
+    @Implementation
+    public boolean isDeviceIdleMode(final Context connectionContext) {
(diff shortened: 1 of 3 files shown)


common/src/test/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapperShadow.java
```diff
@@ -0,0 +1,46 @@
...
+import android.content.Context;
+
+import org.robolectric.annotation.Implementation;
+import org.robolectric.annotation.Implements;
+
+@Implements(PowerManagerWrapper.class)
+public class PowerManagerWrapperShadow {
+
+    public static boolean deviceIdleMode = true;
+    public static boolean ignoringBatteryOptimizations = true;
```
shahzaibj, line 35: I don't see any reason for these to be static...in fact I don't see any reason for these to even be defined (based on their current usage)

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    protected Context mContext;
+
+    @Before
+    public void setup() {
+        mContext = ApplicationProvider.getApplicationContext();
+    }
+
+    @Test
+    public void testConnectionWhenConnected() {
+        DefaultConnectionService defaultConnectionService = new DefaultConnectionService(mContext);
```
shahzaibj, line 60: nit: final here and everywhere

shahzaibj:
Why does the following test (circled) take 7 seconds to run? Is this an indication of a problem?

image

gitaumoses4:
> Why does the following test (circled) take 7 seconds to run? Is this an indication of a problem?
> 
> image

Screenshot from 2021-05-26 20-27-24

@shahzaibj Seems the test that's run first always takes longer. I just changed the name so that the one you marked was run last.

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    protected Context mContext;
+
+    @Before
+    public void setup() {
+        mContext = ApplicationProvider.getApplicationContext();
+    }
+
+    @Test
+    public void testConnectionWhenConnected() {
```
shahzaibj, line 59:
```suggestion
    public void testConnectionAvailableWhenConnected() {
```

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+        final DefaultConnectionService defaultConnectionService = new DefaultConnectionService(mContext);
+
+        assertTrue(defaultConnectionService.isConnectionAvailable());
+    }
+
+    @Test
+    public void testConnectionWhenNoActiveNetwork() {
```
shahzaibj, line 66:
```suggestion
    public void testConnectionNotAvailableWhenNoActiveNetwork() {
```

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    public void testNetworkDisabledFromOptimizations() {
+        final DefaultConnectionService defaultConnectionService = new DefaultConnectionService(mContext);
+        ReflectionHelpers.setStaticField(Build.VERSION.class, "SDK_INT", Build.VERSION_CODES.M);
+
+        assertTrue(defaultConnectionService.isNetworkDisabledFromOptimizations());
+    }
+
+    @Test
+    public void testNetworkDisabledFromOptimizationsForAPIsBelow23() {
```
shahzaibj, line 85:
```suggestion
    public void testNetworkNotDisabledFromOptimizationsForAPIsBelow23() {
```

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+                .getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        shadowOf(connectivityManager).setActiveNetworkInfo(null);
+        final DefaultConnectionService defaultConnectionService = new DefaultConnectionService(mContext);
+
+        assertFalse(defaultConnectionService.isConnectionAvailable());
+    }
+
+    @Test
+    public void testNetworkDisabledFromOptimizations() {
```
shahzaibj, line 77:
```suggestion
    public void testNetworkIsDisabledFromOptimizationsForAPIs23AndAbove() {
```

shahzaibj (approved):
Left some suggestions..else lgtm

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    public void testConnectionNotAvailableWhenNoActiveNetwork() {
+        final ConnectivityManager connectivityManager = (ConnectivityManager) mContext
+                .getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        shadowOf(connectivityManager).setActiveNetworkInfo(null);
+        final DefaultConnectionService defaultConnectionService = new DefaultConnectionService(mContext);
+
+        assertFalse(defaultConnectionService.isConnectionAvailable());
+    }
```
AdamBJohnsonx, line 74: What I would suggest here is to trigger a few different network events, if possible.  If a single network is unavailable, we should still be able to connect, or if we lose one connection and gain another we should remain connected.

AdamBJohnsonx: I think the following permutations would be sufficient:

Have 2, lose 1, stay connected.
Have 2, lose both, not connected, gain 1, stay connected.
Have none, not connected, gain 1, are connected.
Have none, gain 2, are connected, lose 1, are connected.

The idea being that we can then bring back the changes to be event driven and have these tests tell us that they're safe.

gitaumoses4: Brilliant!! ... Thanks Adam.

AdamBJohnsonx (approved):

PR #1396, opened 2021-06-01
Add SSL context support to UrlConnectionHttpClient

rpdome:
Breaking change PR in MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1423
DRS Discoverr PR (Broker): https://github.com/AzureAD/ad-accounts-for-android/pull/1581

--- 

Add the following capability to UrlConnectionHttpClient,
- An ability to insert SSLSocket (needed for DRS requests)
- UrlConnectionHttpClient now supports Higher TLS by default, regardless of Android API version. (Please see the added test case).

Also remove the adapter classes, since we're going for a breaking change in the next release.

This, of course, introduces a breaking change, I'll send PR for those later.


common4j/src/main/com/microsoft/identity/common/java/net/AbstractHttpClient.java
```diff
@@ -22,59 +22,91 @@
...
 import lombok.NonNull;
 
 import java.io.IOException;
 import java.net.URL;
 import java.util.Map;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocket;
+
 public abstract class AbstractHttpClient implements HttpClient {
```
AdamBJohnsonx, line 35: javadoc

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -324,8 +319,15 @@ private HttpURLConnection setupConnection(HttpRequest request) throws IOExceptio
             urlConnection.setRequestProperty(entry.getKey(), entry.getValue());
         }
 
-        urlConnection.setRequestMethod(request.getRequestMethod());
+        if (urlConnection instanceof HttpsURLConnection) {
```
AdamBJohnsonx, line 322: Should we check based on the protocol used in the URL?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -324,8 +319,15 @@ private HttpURLConnection setupConnection(HttpRequest request) throws IOExceptio
...
+            ((HttpsURLConnection) urlConnection).setSSLSocketFactory(
+                    SSLSocketFactoryWrapper.getSocketFactory(request.getSslContext())
```
AdamBJohnsonx, line 324: Can we let the SSL level be configured here?  We might want to enforce 1.3 at some point.

rpdome: @AdamBJohnsonx  do you mean enforce per request?

SSLSocketFactoryWrapper now supports TLS1.3, and it should be picked whenever the server supports it.

AdamBJohnsonx: My point is that we may want to be able to set a minimum.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: Yes. This is exactly what it is. URL.openConnection() returns HttpsURLConnection for HTTPS requests, and HttpUrlConnection for HTTP request.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -183,16 +173,18 @@ private static void recordHttpTelemetryEventEnd(final HttpResponse response) {
...
      * @return HttpResponse      The response for this request.
      * @throws IOException If an error is encountered while servicing this request.
      */
     @Override
     public HttpResponse method(@NonNull final HttpClient.HttpMethod httpMethod,
                                @NonNull final URL requestUrl,
                                @NonNull final Map<String, String> requestHeaders,
-                               final byte[] requestContent) throws IOException {
+                               final byte[] requestContent,
+                               final SSLContext sslContext) throws IOException {
```
AdamBJohnsonx, line 185: I've been having to do some research here.  This should be an attribute of the Client, not a per-request feature, probably.

rpdome: The socket(factory) is set as part of HttpsUrlConnection, so it should be per-request.

(For example,  we could potentially reuse the client for 2 DRS requests for different WPJ certs - that's 2 SSL contexts.)

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: We could remove what we don't want to support from SSLSocketFactoryWrapper.SUPPORTED_SSL_PROTOCOLS.

That is... unless we specifically want to do this per-request.

(Of course, we'll have to kill off the non-https support as well).

AdamBJohnsonx: I'm thinking that this is likely a per-client decision.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
AdamBJohnsonx: Hm.  OK, having to go read through this code.  Are we doing this in a way that would allow us to reuse the SSL context for a given session? Right now, it looks to me like we're doing this afresh every time even when we may already have had a connection.  In any event, it's better than what we have.

common4j/src/test/com/microsoft/identity/common/java/net/UrlConnectionHttpClientTest.java
```diff
@@ -1092,4 +1099,82 @@ private void mockRequestBody(final HttpURLConnection mockedConnection) throws IO
...
+        final HttpResponse response = sNoRetryClient.method(
+                HttpClient.HttpMethod.GET,
+                new URL("https://tls-v1-0.badssl.com:1010/"),
+                new LinkedHashMap<String, String>(),
+                null,
+                null
+        );
+
+        Assert.assertEquals(response.getStatusCode(), 200);
+        Assert.assertEquals(SSLSocketFactoryWrapper.getLastHandshakeTLSversion(), "TLSv1");
```
AdamBJohnsonx, line 1134: Assert.assertEquals(<expected>, <actual>), right?

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,203 @@
...
+        return modifyEnabledSockets(mBaseSocketFactory.createSocket(address, port, localAddress, localPort));
+    }
+
+    /**
+     * Returns a Socket with enabled protocols that include TLSv1.2
+     *
+     * @param socket {@link Socket} Socket to be modified to enable TLSv1.2
+     * @return Socket
+     */
+    private Socket modifyEnabledSockets(Socket socket) {
```
gitaumoses4, line 125:
```suggestion
    private Socket modifyEnabledSockets(@NonNull Socket socket) {
```

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,203 @@
...
+        enabledCipherSuites.toArray(array);
+        return array;
+    }
+
+    /**
+     * Returns a {@link SSLSocketFactoryWrapper}.
+     */
+    @NonNull
+    public static synchronized SSLSocketFactory getSocketFactory(@Nullable SSLContext context) {
+        final String methodName = "getSocketFactory";
```
gitaumoses4, line 188:
```suggestion
        final String methodName = ":getSocketFactory";
```

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,203 @@
...
+            return sDefault;
+        }
+
+        final SocketFactory factory = context.getSocketFactory();
+        if (factory == null){
+            Logger.warn(TAG + methodName, "Failed to construct a SSLSocketFactory from SSLContext, returns the default one.");
+            return sDefault;
+        }
+
+        return new SSLSocketFactoryWrapper((SSLSocketFactory) context.getSocketFactory());
```
gitaumoses4, line 201:
```suggestion
        return new SSLSocketFactoryWrapper(context.getSocketFactory());
```

No need to cast since context.getSocketFactory() returns an SSLSocketFactory object.

gitaumoses4 (commented):
Left some few comments. :)

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: @AdamBJohnsonx  If this is not provided, then we will fall back to the default one.
If it's provided, it's up to the caller to specify the same one.

I could also create our own SSLContext that returns a wrapped Factory (instead of creating a new one every time this is provided).

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: added a field in UrlConnectionHttpClientTest's builder.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
rpdome: added a check and 2 test cases.

rpdome:
TLS1 and TLS1.1 test are failing because... the pipeline doesn't support such connection
it's either the JVM itself, or there are something else that blocks it - still have no clue.

javax.net.ssl.SSLHandshakeException: The server selected protocol version TLS11 is not accepted by client preferences [TLS13, TLS12, SSL20Hello]

common4j/src/test/com/microsoft/identity/common/java/net/UrlConnectionHttpClientTest.java
```diff
@@ -1092,4 +1106,129 @@ private void mockRequestBody(final HttpURLConnection mockedConnection) throws IO
...
+
+    @Test
+//    @Ignore("Ignored because our pipeline doesn't support TLS1.0 and TLS1.1. This still can be run locally.")
+    public void testTLS1() throws IOException {
+        final HttpResponse response = sNoRetryClient.method(
+                HttpClient.HttpMethod.GET,
+                new URL("https://tls-v1-0.badssl.com:1010/"),
+                new LinkedHashMap<String, String>(),
+                null,
+                null
```
AdamBJohnsonx, line 1138: Can we run these as methods that take an ssl context?

AdamBJohnsonx (approved):

rpdome:
rebased dev.

common4j/src/test/com/microsoft/identity/common/java/net/UrlConnectionHttpClientTest.java
rpdome: @AdamBJohnsonx  I can add tests for that.

common4j/src/main/com/microsoft/identity/common/java/util/ResultFuture.java
```diff
@@ -58,7 +58,7 @@ public boolean isDone() {
     }
 
     @Override
-    public synchronized T get() throws InterruptedException, ExecutionException {
+    public T get() throws InterruptedException, ExecutionException {
```
AdamBJohnsonx: Why are we removing the synchronized keyword from these methods?

rpdome: I shouldn't have added it. This function calls await and would cause a deadlock with the synchronized keyword.

Removing it here means the code is back where it was before this PR.

AdamBJohnsonx: You know, I think I recall this.  Should probably add documentation so that nobody else tries it.

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * SSL Protocols that our library supports.
+     */
+    public static final List<String> SUPPORTED_SSL_PROTOCOLS =
+            Collections.unmodifiableList(Arrays.asList("SSLv2Hello", "SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"));
+
+    // Gets TLS version of the latest-established socket connection. For testing only.
+    // NOTE: This onMethod thing doesn't generate javadoc, but this method is only exposed for testing only.
+    @Getter(value = AccessLevel.PACKAGE, onMethod_={@Synchronized})
+    @Setter(value = AccessLevel.PACKAGE, onMethod_={@Synchronized})
```
shahzaibj, line 63: we already discussed not to use this `onMethod` notation: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/1360

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,180 @@
...
+ * reference: https://developer.android.com/reference/javax/net/ssl/SSLSocket
+ */
+public class SSLSocketFactoryWrapper extends SSLSocketFactory {
+    private static final String TAG = SSLSocketFactoryWrapper.class.getSimpleName();
+
+    /**
+     * SSL Protocols that our library supports.
+     */
+    public static final List<String> SUPPORTED_SSL_PROTOCOLS =
+            Collections.unmodifiableList(Arrays.asList("SSLv2Hello", "SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"));
```
shahzaibj, line 58: probably can define each of these a constant and then include the constants here

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -324,8 +332,21 @@ private HttpURLConnection setupConnection(HttpRequest request) throws IOExceptio
...
+        if (urlConnection instanceof HttpsURLConnection) {
+            final SSLSocketFactory factory =
+                    request.getSslContext() != null ?
+                            new SSLSocketFactoryWrapper(request.getSslContext().getSocketFactory(), supportedSslProtocol) :
+                            getDefaultWrapper();
+            ((HttpsURLConnection) urlConnection).setSSLSocketFactory(factory);
+        } else if ("https".equalsIgnoreCase(request.getRequestUrl().getProtocol())) {
+            throw new IllegalStateException("Trying to initiate a HTTPS request, but didn't get back HttpsURLConnection");
+        } else if ("http".equalsIgnoreCase(request.getRequestUrl().getProtocol())) {
+            Logger.warn(TAG + methodName, "Making a request for non-https URL.");
```
shahzaibj, line 344: if it is not `https` then is it guaranteed to be http?

common4j/src/test/com/microsoft/identity/common/java/net/UrlConnectionHttpClientTest.java
```diff
@@ -1092,4 +1108,172 @@ private void mockRequestBody(final HttpURLConnection mockedConnection) throws IO
...
+                HttpClient.HttpMethod.GET,
+                new URL("http://www.somewebsite.com/"),
+                new LinkedHashMap<String, String>(),
+                null,
+                null
+        );
+
+        Assert.assertEquals(200, response.getStatusCode());
+    }
+}
```
shahzaibj, line 1279: nit: new line ending

shahzaibj (changes requested):
Using lombok's experimental feature that don't play well with javadoc

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
rpdome: @shahzaibj This is exposed for test case only (as mentioned in the comment above).

shahzaibj: This is not in test directory though...so my understanding is that an attempt will be made to generate javadoc for this class and it will fail. Are you doing anything special to not generate javadoc for these?

common4j/src/main/com/microsoft/identity/common/java/net/SSLSocketFactoryWrapper.java
rpdome: I think that's overkill.

shahzaibj (approved):

PR #1402, opened 2021-06-03
[Automation] Add helper method to get password for lab guest user

iamgusain:
Adding a new helper method `getPasswordForGuestUser `to get password for guest account.
Also added `homeDomain` property in LabGuest class, which is used to fetch the home lab name for the guest user.


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabGuestAccountHelper.java
```diff
@@ -33,9 +33,14 @@ public static LabGuest loadGuestAccountFromLab(final LabUserQuery query) {
...
+                userInfo.getHomeDomain(),
                 userInfo.getHomeTenantID(),
                 guestLabTenants
         );
     }
+
+    public static String getPasswordForGuestUser(final LabGuest guestUser) {
+        final String labName = guestUser.getHomeDomain().split("\\.")[0];
+        return LabHelper.getSecret(labName);
+    }
```
shahzaibj, line 45: Do this existing mechanism for getting password not work for your needs?

iamgusain: No they get password for the msidlab4 (where the user is guest), we need the password for the home lab

shahzaibj: Are you saying that the existing guest account helper code fetches password for the wrong lab? How are they existing guest account tests working then?

iamgusain: I did not say that. 
What I am saying is that existing code fetches the password for the lab, which gets returned in the labConfig part of the config endpoint response. In this case that labconfig is for msidlab4 which is serving the account as it exists as a guest in that lab.
If things are not clear, You can attach fiddler or debug through existing test case to see what I am saying

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabGuest.java
```diff
@@ -6,14 +6,14 @@
  * A Class to facilitate writing tests for Guest Accounts
  */
 public class LabGuest {
-    private String homeUpn;
-    private String homeLabName;
-    private String homeTenantId;
-    private List<String> guestLabTenants;
+    private final String homeUpn;
+    private final String homeTenantId;
+    private final String homeDomain;
```
rpdome, line 11: What's the difference between labName and Domain? 

Also, adding a comment specifying each params in the constructor might help.

iamgusain: labName `(e.g. "arlmsidlab1")` is part of the domainName `(e.g. "arlmsidlab1.us")`

Which parameter is not clear by name, I would prefer to make the name more descriptive than add comment.

testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/LabGuestAccountHelper.java
shahzaibj: So this seems a change in behavior by the lab api...for regular guest users (in the same cloud), the lab info returned was lab of the home cloud / tenant, however, it seem for cross cloud guests, the lab info returned is now of the guest cloud / tenant

shahzaibj: we should probably ask the lab team to fix it

shahzaibj: Resolved offline

shahzaibj (approved):

AdamBJohnsonx (approved):

PR #1389, opened 2021-05-24
Add java only lab api module + Authentication for Lab Api + Tests + Pipeline

shahzaibj:
**Brief Summary**
In this PR, we've added a new JAVA only module for interacting with the LAB API.

**Why doing this work?**
The current LAB API utility module `testutils` is an Android library and thus can't be used in JAVA only gradle projects, and now that we are developing java only projects (common4j, broker4j etc.) we are going to need to writes tests in these modules that are using accounts from LAB API and therefore we need a JAVA only module for LAB API.

Furthermore, `testutils` was a mess...it was taking a dependency on `common` and that meant we can't use `testutils` inside common (circular dependency). Also most of the methods inside `testutils` LAB API wrapper were static methods and very difficult to change implementation. The client secret used to acquire access token for the KeyVault also had to be supplied directly to the BuildConfig of testutils i.e. it must be supplied during build time and that meant consuming a published version of `testutils` was a pain.

In this PR, we've solved all those problems.

**What all is in this PR?**
- A new JAVA module for Lab Api Utilities
- Porting logic from common to create client assertions using certificate (this means some of this code is duplicated in common..but that's okay...we don't want to take a dependency on common when we are trying to test common itself.)
- Add interfaces to perform a confident client flow
- Use msal4j to get a token for confidential client
- Add authentication helpers for KeyVault and LabApi to get access tokens for these
- Add test cases to make sure we can get access tokens for KeyVault and Lab Api
- Add a pipeline to build and test Lab Api Utilities

**What's NOT in this PR?**
We haven't added any code to invoke the Lab Api itself...going to do that in a separate PR as this PR is already big. This PR only covers the authentication aspect of Lab Api...in the next PR, we will take this access token and send it to Lab Api to fetch test accounts.

PR #1406, opened 2021-06-07
Add Lab Api helper methods to Java only Lab Client to create & load accounts for testing purposes

shahzaibj:
Adding (helper) methods to load accounts from Lab Api and create temp users using our new Java only Lab Api / KeyVault helper module. This module was created earlier and authentication helpers were added here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1389

**This PR contains the following changes:**
- Add an interface describing the operations that will be performed by our Lab Api Client (right now only 4 basic operations are there....we can add more over time as expand usage of this module)
- Add an implementation for these operations
- Create a LabQuery class that will be initialized using lombok builder
- Each parameter to LabQuery is an enum now...so it has dedicated type (as opposed to String we've had in the past)
- Create an Account object model that will represent an account used for testing (it will contain username, password)
- Right a couple of tests for these new APIs.

**Note:** I did not add any javadoc to the enums (each param to the Lab Api) because I don't fully understand what each of them are or what they do. We will need to understand the Lab Api to document them and this can be done in the future in collaboration with the lab team.


LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/AltIdSource.java
```diff
@@ -0,0 +1,38 @@
...
+    NONE(LabConstants.AltIdSource.NONE),
+    ON_PREM(LabConstants.AltIdSource.ON_PREM);
+
+    final String value;
+
+    AltIdSource(final String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
```
iamgusain, line 35: >getValue [](http://example.com/codeflow?start=18&length=8)

Use toString() instead of getValue, That way you can use it as AltSource.None instead of AltSource.None.getValue(), and will also not need to do a nul check.

Same for all other enums

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/ILabClient.java
```diff
@@ -0,0 +1,71 @@
...
+public interface ILabClient {
+
+    /**
+     * Load an existing account from Lab Api based on the provided query.
+     *
+     * @param labQuery parameters that determine what kind of account to fetch from lab
+     * @return a {@link LabAccount} object
+     * @throws LabApiException if an error occurs while trying to fetch account from lab
+     */
+    LabAccount fetchUser(LabQuery labQuery) throws LabApiException;
```
iamgusain, line 42: >LabAccount fetchUser [](http://example.com/codeflow?start=4&length=20)

nit: 
LabAccount getLabAccount or LabUser getLabUser

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/ILabClient.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Loads existing account(s) from Lab Api based on the provided query.
+     *
+     * @param labQuery parameters that determine what kind of account(s) to fetch from lab
+     * @return a list of {@link LabAccount} objects
+     * @throws LabApiException if an error occurs while trying to fetch account(s) from lab
+     */
+    List<LabAccount> fetchUsers(LabQuery labQuery) throws LabApiException;
```
iamgusain, line 51: >List<LabAccount> fetchUsers [](http://example.com/codeflow?start=4&length=27)

nit: 
List<LabAccount> getLabAccounts or List<LabUser> getLabUsers

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabAccount.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.experimental.Accessors;
+
+/**
+ * An account object model that will used to represent accounts used for testing purposes.
+ */
+@Getter
+@Accessors(prefix = "m")
+@AllArgsConstructor(access = AccessLevel.PUBLIC)
+@EqualsAndHashCode
+public class LabAccount {
```
iamgusain, line 40: >LabAccount [](http://example.com/codeflow?start=13&length=10)

Why not just expose the UserInfo object, since it contains lot of other useful info about the user

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private List<ConfigInfo> fetchConfigsFromLab(@NonNull final LabQuery query) throws LabApiException {
+        Configuration.getDefaultApiClient().setAccessToken(
+                mLabApiAuthenticationClient.getAccessToken()
+        );
+        try {
+            final ConfigApi api = new ConfigApi();
+            return api.apiConfigGet(
+                    query.getUserType() != null ? query.getUserType().getValue() : null,
```
iamgusain, line 103: >query.getUserType() != null ? query.getUserType().getValue() : null, [](http://example.com/codeflow?start=20&length=68)

if you override the toString() method in the enums this line will simply become.. "query.getUserType()".
See my comment in AltIdSource.java

iamgusain (changes requested):
Added some comments

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
shahzaibj: hmm..actually it wouldn't work because the `apiConfigGet` method wants `Strings`...but `query.getUserType()` would return something of type `UserType` and the method wouldn't accept it:

image

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/AltIdSource.java
shahzaibj: I tried doing it but it doesn't seem to work...see comment here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406#discussion_r648628660

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/ILabClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/b572173845668ac1d2389720f161e185023c11e9

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/ILabClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/b572173845668ac1d2389720f161e185023c11e9

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabAccount.java
shahzaibj: The idea is that the lab team can regress our code by changing their model...for instance they might change the name of the UserInfo type...or they may remove it completely...with this new Lab Api module that I am building, I don't want to take a hard dependency on the lab api specific models...and rather wrap them in a custom model...in this case the Account object..so let's say we regenerate the swagger client in the near future then it is okay if it breaks our internal APIs but it shouldn't break any of our public Lab Api helpers...these should remain unchanged...and that's what I am trying to do here...I've added an interface for the Account model here as well: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/ce755bb47e43a01913f0ae9fb9e4816e8e72c4b2

Right now I've kept only 3 methods...but we may add a method there in future for instance called as `getHomeDomain` and then we can implement it to return `getUserInfo.getHomeDomain()` but if lab team changes it in the future and moves the property to say LabInfo then we can change the implementation to `getLabInfo.getHomeDomain()`...but the signature of the public API remains unchanged in our tests.

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
AdamBJohnsonx: It would work if you were doing string manipulation, but that's because the string + string gets turned into new StringBuilder.append(string).append(string).toString() by the JVM.  There's also More Magic(tm) with strings.  I bet it would if you did "" + foo, as the argument as well.

iamgusain: One way to make it work would be to use String.valueOf()...
https://www.jdoodle.com/ia/ex2

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabAccount.java
iamgusain: I understand your POV. 
My concern was around LabApi returning a lot of other useful info and we seem to trim it down a lot, I think its fine we can just expose what is needed and then add on later.

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/d5e4b3943a0944faaab79d0fecb91d175da48c72 (using a custom valueOf method...because String.valueOf wouldn't work for our purposes)

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/AltIdSource.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/d5e4b3943a0944faaab79d0fecb91d175da48c72

shahzaibj:
@iamgusain All comments have been addressed...please take another look

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -0,0 +1,188 @@
...
+                    valueOf(query.getTokenLifetime())
+            );
+        } catch (final com.microsoft.identity.internal.test.labapi.ApiException ex) {
+            throw new LabApiException(LabError.FAILED_TO_GET_ACCOUNT_FROM_LAB, ex);
+        }
+    }
+
+    private String valueOf(final Object obj) {
+        return obj == null ? null : obj.toString();
+    }
```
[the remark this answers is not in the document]
shahzaibj: You're look at the right link...but they are clearly different code though...String.valueOf does: 

```java
return (obj == null) ? "null" : obj.toString();
```

We need:

```java
return obj == null ? null : obj.toString();
```

iamgusain (approved):
:shipit:

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -0,0 +1,188 @@
...
+    private static final int TEMP_USER_API_READ_TIMEOUT = (int) TimeUnit.SECONDS.toMillis(15);
+
+    @Override
+    public LabAccount getLabAccount(@NonNull final LabQuery labQuery) throws LabApiException {
+        final List<ConfigInfo> configInfos = fetchConfigsFromLab(labQuery);
+        // for each query, lab actually returns a list of accounts..all of which fit the criteria..
+        // usually we only need one such account, and hence over here we are just picking the first
+        // element of the list.
+        final ConfigInfo configInfo = configInfos.get(0);
+        return createAccount(configInfo);
```
shoatman, line 65: In this case you getting an account from the lab and populating a LabAccount DTO with that data.  I think the name here could be a bit confusing to causal readers....perhaps getLabAccountObject?  I think this ia nit overall...

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -0,0 +1,188 @@
...
+            throw new LabApiException(LabError.FAILED_TO_GET_ACCOUNT_FROM_LAB, ex);
+        }
+    }
+
+    private String valueOf(final Object obj) {
+        return obj == null ? null : obj.toString();
+    }
+
+    @Override
+    public LabAccount createTempUser(@NonNull final TempUserType tempUserType) throws LabApiException {
```
shoatman, line 136: It would be good if the lab team had called this create TempAccount... it's hard to create a temporary human agent... and a temporary software agent for that matter... nit.

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/HomeDomain.java
```diff
@@ -0,0 +1,41 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.labapi.utilities.constants;
+
+public enum HomeDomain {
```
shoatman, line 25: Not sure if HomeDomain is a lab concept or something that you created.  I've heard of Home realm... but not Home Domain before.

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/LabConstants.java
```diff
@@ -0,0 +1,199 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.labapi.utilities.constants;
+
+public class LabConstants {
```
shoatman, line 25: Feels like this should have been generatable from the lab swagger, but i'm guessing not.

shoatman (approved):
Nits...

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
shahzaibj: They called it `createTempUser` but we can change the name on our end to call the API createTempAccount

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/HomeDomain.java
shahzaibj: This is a lab concept...it basically is the home tenant represented as a domain such as `msidlab2.com`

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/ad75124d3f31efb65375ebad81874cea4eb55d1e

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1406/commits/de5b2db2c4cb4c4caad3555a12237704ca9ce6b5

PR #1414, opened 2021-06-10
Using a static method to drop an sso challenge handler into the webview.

AdamBJohnsonx:
There are two parts here.  One of them is putting the challenge handler into
the webview.  It's not very clean, relying on there being only one interactive
request at a time.  Then adding a new url handler to deal with the redirect we
get from the server.  There's likely additional armor to add here.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1684,6 +1684,8 @@ public final class AuthenticationConstants {
         public static final String WEB_VIEW_ZOOM_CONTROLS_ENABLED = "com.microsoft.identity.web.view.zoom.controls.enabled";
 
         public static final String WEB_VIEW_ZOOM_ENABLED = "com.microsoft.identity.web.view.zoom.enabled";
+
+        public static final String SSO_TOKEN_HANDLING = "com.microsoft.identity.web.view.sso_token";
     }
 
     public static final class AuthorizationIntentAction {
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -37,11 +37,13 @@ import android.widget.ProgressBar;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
+import androidx.arch.core.util.Function;
 
 import com.microsoft.identity.common.R;
 import com.microsoft.identity.common.WarningType;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
+import com.microsoft.identity.common.exception.ClientException;
 import com.microsoft.identity.common.internal.ui.webview.AzureActiveDirectoryWebViewClient;
 import com.microsoft.identity.common.internal.ui.webview.OnPageCommitVisibleCallback;
 import com.microsoft.identity.common.internal.ui.webview.OnPageLoadedCallback;
@@ -49,7 +51,11 @@ import com.microsoft.identity.common.internal.ui.webview.WebViewUtil;
 import com.microsoft.identity.common.internal.ui.webview.challengehandlers.IAuthorizationCompletionCallback;
 import com.microsoft.identity.common.logging.Logger;
 
+import java.net.URL;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Map;
+import java.util.WeakHashMap;
 
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.AUTH_INTENT;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.POST_PAGE_LOADED_URL;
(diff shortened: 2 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -162,6 +182,26 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            // I do not love this being a HashMap, but I don't dare change the type because of the djinni layer.
+            Map<String, String> headers = new HashMap<>(oldHeaders);
+            final String newHeaderName = mSsoTokenHandler.getHeaderName();
+            final String newCredential = mSsoTokenHandler.getInterruptCredential(ssoNonce);
+            if (headers != null && newHeaderName != null && newCredential != null) {
+                headers.put(newHeaderName, newCredential);
+                view.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        view.loadUrl(url, headers);
```
AdamBJohnsonx, line 198: I believe that this is not correct.

AdamBJohnsonx: Nope, it's actually exactly correct.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1684,6 +1684,8 @@ public static String computeMaxHostBrokerProtocol() {
         public static final String WEB_VIEW_ZOOM_CONTROLS_ENABLED = "com.microsoft.identity.web.view.zoom.controls.enabled";
 
         public static final String WEB_VIEW_ZOOM_ENABLED = "com.microsoft.identity.web.view.zoom.enabled";
+
+        public static final String SSO_TOKEN_HANDLING = "com.microsoft.identity.web.view.sso_token";
```
AdamBJohnsonx, line 1688:
```suggestion
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -165,6 +181,26 @@ private boolean handleUrl(final WebView view, final String url) {
         } else if (isRedirectUrl(formattedURL)) {
             Logger.info(TAG, "Navigation starts with the redirect uri.");
             processRedirectUrl(view, url);
+        } else if (ssoNonce != null && mSsoTokenHandler != null) {
```
AdamBJohnsonx, line 184: This will prevent it from triggering on random other requests.  But it would be better to know the "original url" before we got redirected, since it should "match" (for some definition of match).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/IAuthorizationCompletionCallback.java
```diff
@@ -40,4 +42,6 @@
     void onChallengeResponseReceived(int returnCode, Intent responseIntent);
 
     void setPKeyAuthStatus(boolean status);
+
```
AdamBJohnsonx, line 45: Not needed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -128,6 +134,31 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
         final String methodName = "#onCreateView";
         final View view = inflater.inflate(R.layout.common_activity_authentication, container, false);
         mProgressBar = view.findViewById(R.id.common_auth_webview_progressbar);
+        final SsoCredentialResolver resolver;
+        if (sSsoKeyMap.get("CONSTANT_KEY") != null) {
```
iambmelt, line 138: As a minor cleanup item, `"CONSTANT_KEY"` should probably be a constant itself

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -196,6 +232,30 @@ private boolean handleUrl(final WebView view, final String url) {
...
+            }
+
+        if(realUrl.getQuery().contains("sso_nonce=")) {
+                final Matcher matcher = SSO_PATTERN.matcher(query);
+                if(matcher.matches()) {
+                    if (matcher.groupCount() != 1) {
+                        Logger.warn(TAG, "too many sso_nonce query parameters");
+                        return null;
+                    } else {
+                        String sso_nonce = matcher.group(1);
```
iambmelt, line 249: Is a regex the best way to go here? Can we use [`Uri.getQueryParameter(key)`](https://developer.android.com/reference/android/net/Uri#getQueryParameter(java.lang.String)) to do the lifting for us? I do see a comment in the javadoc re: backcompat behavior around certain chars; I'm wondering if maybe using a regex was a workaround for that....

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -38,6 +38,8 @@
 import org.robolectric.Robolectric;
 import org.robolectric.RobolectricTestRunner;
 
+import java.net.URL;
```
iambmelt, line 41: Stray import?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/IAuthorizationCompletionCallback.java
```diff
@@ -24,6 +24,8 @@
 
 import android.content.Intent;
 
+import java.net.URL;
```
iambmelt, line 27: Stray import?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
iambmelt: Whichever way we go, may be good idea to have a unit test or two covering nonce format

rpdome: Let's not use Android's Uri here....

PR #1418, opened 2021-06-16
Migrate to Common4j Pair

rpdome:
Related PR
Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/1593
ADAL:https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1614
MSAL:https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1441
----

Ran into an issue (regression) while debugging - type erasure. 

An object inside a list that is passed through is not type-checked at compile time until it's being used.

The error I ran into was that we Initially set the value with **_android.util.Pair_**, pass it all the way down to common4j, which expected `com.microsoft.identity.common.java.util.ported.Pair`.

To eliminate the issue, ~~I'm going to migrate our whole stack to the latter. ~~

EDIT: @AdamBJohnsonx  suggested that we use AbstractMap.SimpleEntry<> instead, so I'm going to kill off Common4j's pair and go with that one.

ps. Given that ADFS supports multiple query parameter of the same key, I'm not going to migrate to Map as mentioned earlier. (They're not currently using it though).


shahzaibj:
I read that ADFS supports multiple qp for the same key and therefore you are not planning to migrate to Map and hence keeping Pair. As an alternative, could we model queryParameters as a `Map<String, Set<String>>` where a key (qp) can map to multiples values that are in the set that the key maps to?

shahzaibj:
> I read that ADFS supports multiple qp for the same key and therefore you are not planning to migrate to Map and hence keeping Pair. As an alternative, could we model queryParameters as a `Map<String, Set<String>>` where a key (qp) can map to multiples values that are in the set that the key maps to?

If you wanted, you could also create a custom class that models the C# NameValueCollection: https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection?view=net-5.0

rpdome:
> I read that ADFS supports multiple qp for the same key and therefore you are not planning to migrate to Map and hence keeping Pair. As an alternative, could we model queryParameters as a `Map<String, Set<String>>` where a key (qp) can map to multiples values that are in the set that the key maps to?

@shahzaibj IMO, it's an overkill at the moment. 

ADFS does support it, as in it `recognizes` multiple QP. That said, they're not using it in any scenarios.
eSTS hasn't got back to me, but I know for a fact that they reject duplicated `brkr` param.

shahzaibj:
```java
class NameValueCollection {

}
```

> > I read that ADFS supports multiple qp for the same key and therefore you are not planning to migrate to Map and hence keeping Pair. As an alternative, could we model queryParameters as a `Map<String, Set<String>>` where a key (qp) can map to multiples values that are in the set that the key maps to?
> 
> @shahzaibj IMO, it's an overkill at the moment.
> 
> ADFS does support it, as in it `recognizes` multiple QP. That said, they're not using it in any scenarios.
> eSTS hasn't got back to me, but I know for a fact that they reject duplicated `brkr` param.

Yeah, I would let you pick...but I think this may be cleaner than a `List<Pair>`...would allow multiple values while not allowing duplicates. I just did a quick gist here to try it out: https://gist.github.com/shahzaibj/6d1306cd02586a1ffab3096f8da3f9c0

rpdome:
> > > I read that ADFS supports multiple qp for the same key and therefore you are not planning to migrate to Map and hence keeping Pair. As an alternative, could we model queryParameters as a `Map<String, Set<String>>` where a key (qp) can map to multiples values that are in the set that the key maps to?
> > 
> > 
> > @shahzaibj IMO, it's an overkill at the moment.
> > ADFS does support it, as in it `recognizes` multiple QP. That said, they're not using it in any scenarios.
> > eSTS hasn't got back to me, but I know for a fact that they reject duplicated `brkr` param.
> 
> Yeah, I would let you pick...but I think this may be cleaner than a `List<Pair>`...would allow multiple values while not allowing duplicates. I just did a quick gist here to try it out: https://gist.github.com/shahzaibj/6d1306cd02586a1ffab3096f8da3f9c0

For now, I think it would be better to be on the safe side and not deciding anything on any servers' behalf - unless we're getting an explicit green light. 

(e.g. we know that eSTS will reject brkr=1, so we make sure not to send that, but this might not be true for other values)

This includes eliminating duplicated query params.

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -23,10 +23,54 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AllArgsConstructor;
+import lombok.NonNull;
 
 @AllArgsConstructor
 public class Pair<T, U> {
```
shoatman, line 35: Do you just want to take the code directly from AOSP: https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/util/Pair.java

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -23,10 +23,54 @@
...
+    }
+
+    /**
+     * Add this item to a list if a copy of the same key-value pair doesn't exist.
+     * */
+    public void addToListIfNotExist(@NonNull final List<Pair<T, U>> listToBeAdded) {
+        if (!listToBeAdded.contains(this)) {
+            listToBeAdded.add(this);
+        }
+    }
```
AdamBJohnsonx, line 74: Is what we're trying to prevent duplicate k-v-p's or duplicate keys?

AdamBJohnsonx: i.e. - would the server accept ?brkr=2&brkr=1 or ?brkr=1,2?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -23,10 +23,54 @@
...
 
 @AllArgsConstructor
 public class Pair<T, U> {
     public final T first;
     public final U second;
+
+    @Override
+    @SuppressFBWarnings(value = "NP_METHOD_PARAMETER_TIGHTENS_ANNOTATION",
+            justification = "Equals function should be able to take in a null object.")
+    public boolean equals(@Nullable final Object obj) {
```
AdamBJohnsonx, line 42: Is something wrong with @EqualsAndHashCode?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -23,10 +23,54 @@
...
+
+    /**
+     * Add this item to a list if a copy of the same key-value pair doesn't exist.
+     * */
+    public void addToListIfNotExist(@NonNull final List<Pair<T, U>> listToBeAdded) {
+        if (!listToBeAdded.contains(this)) {
+            listToBeAdded.add(this);
+        }
+    }
+
```
gitaumoses4, line 75: Can we add a convenience method to create a pair.
```java
public static <A, B> Pair <A, B> create(A a, B b){
  return new Pair<>(a, b);
}
```

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
```diff
@@ -23,10 +23,54 @@
...
+    /**
+     * Compute a hash code using the hash codes of the underlying objects
+     *
+     * @return a hashcode of the Pair
+     */
+    @Override
+    public int hashCode() {
+        int result = (first != null ? first.hashCode() : 0);
+        result = 31 * result + (second != null ? second.hashCode() : 0);
```
gitaumoses4, line 63: Using `31 * result`  here will cause these pairs to have different hashcodes,
```java
Pair<String, String> first = new Pair<>("foo", "bar");
Pair<String, String> second = new Pair<>("bar", "foo");
```

But structurally, they have the same pair of values. Is this intentional, and if so, what's the reason for considering them as different pairs?

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
rpdome: ADFS seems to accept them. eSTS, however, might not (haven't heard back from them yet, but they rejects duplicated brkr=1)

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
rpdome: It's intentional.

We're using pair to store query parameter. Let's say brkr=1

brkr=1 should not be equal to 1=brkr.

rpdome: (Maybe I should've renamed this to KeyValuePair)

gitaumoses4: Yeah, it's more of a `KeyValuePair` than a `Pair`.

shahzaibj: Use lombok?

rpdome: @shahzaibj just addressed with EqualsAndHashCode as @AdamBJohnsonx  suggested above.

gitaumoses4 (approved):
Looks good.

common4j/src/main/com/microsoft/identity/common/java/util/ported/Pair.java
rpdome: Decided not to do it for now.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -22,7 +22,7 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.adal.internal.tokensharing;
 
-import android.util.Pair;
+import com.microsoft.identity.common.java.util.ported.KeyValuePair;
```
AdamBJohnsonx, line 25: At this point... you've had to change every usage because the signature changed.  Why not Map.Entry?

rpdome: Replacing List<Pair<,>> with a Map<,> is a breaking change - which would require verification of the scenarios below

1. calls to eSTS (haven't got back answers, only know that they reject requests with duplicated brkr=1)
2. calls to ADFS (they support multiple query parameter values of the same key, but are not using it in any scenarios)
3. calls to on-prem identity solutions customers might be using (as pointed out by @iambmelt in the thread to ADFS, eSTS)

Given the size of coverage, I prefer not to deal with with the moment.

AdamBJohnsonx: I didn't say Map... I said Map.Entry.  List<KeyValyePair<A, B>> is equivalent to Iterable<Map.Entry<A, B>> as far as I can tell, but doesn't require a new class and doesn't provide list-semantics.  Why not?  Do we need List APIs?

rpdome: Not sure if I'm following.... Map.Entry is an interface, are you suggesting that I override a pair class from that?

AdamBJohnsonx: There are current implementations that ship with java.  Pretty sure I put AbstractMap.SimpleEntry into Common for this purpose where there wasn't already a usage of Pair and I didn't want to create a new one.

AdamBJohnsonx: https://docs.oracle.com/javase/7/docs/api/java/util/AbstractMap.SimpleEntry.html

AdamBJohnsonx: And if you change every usage of this thing to refer to the interface rather than the concrete class, just in case we want to change it in the future... you can ship it.  It might be worth having a utility method to make these, if we're calling new a lot, your call.

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -224,9 +222,9 @@ public void saveOrgIdFamilyRefreshToken(@NonNull final String ssoStateSerializer
                     }
                 });
 
-        final Pair<MicrosoftAccount, MicrosoftRefreshToken> resultPair = resultFuture.get();
+        final Map.Entry<MicrosoftAccount, MicrosoftRefreshToken> resultKeyValuePair = resultFuture.get();
```
shahzaibj, line 225: I guess we could just call it `result`

```suggestion
        final Map.Entry<MicrosoftAccount, MicrosoftRefreshToken> result = resultFuture.get();
```

shahzaibj (approved):

AdamBJohnsonx (approved):

PR #1420, opened 2021-06-19
Using singleton-based SharedPrefsFileManager for BAM cache + timing logs

iambmelt:
Playing around with optimizations to the BAM-Cache.

Per recent review of Yammer logs, updating this cache was taking about 1s.


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+        );
+        mKeySingleEntry = singleKey;
+    }
+
+    private interface NamedRunnable<V> extends Callable<V> {
+        String getName();
+    }
+
+    @SneakyThrows
+    private <V> V execWithTiming(@NonNull NamedRunnable<V> runnable) {
```
iambmelt, line 84: @AdamBJohnsonx for your consideration 😁

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
```diff
@@ -38,7 +38,8 @@
 import java.util.Set;
 
 public class SharedPreferencesBrokerApplicationMetadataCache
-        extends SharedPreferencesSimpleCacheImpl<BrokerApplicationMetadata>
+        extends SharedPreferencesFileManagerSimpleCacheImpl<BrokerApplicationMetadata>
+        //extends SharedPreferencesSimpleCacheImpl<BrokerApplicationMetadata>
```
iambmelt, line 42: Can toggle between these interfaces to see how fast/slow tests run.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: nano time not avail @ API 16, but _is avail_ at API 17 (which we should move to)

iambmelt:
Somewhat unscientific test run metrics - inserting 1k records... Seems like a pretty good perf boost

New implementation:
image

Old Implementation:
image

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
AdamBJohnsonx: Should you if(...) else?

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                null // File is not encrypted
+        );
+        mKeySingleEntry = singleKey;
+    }
+
+    private interface NamedRunnable<V> extends Callable<V> {
+        String getName();
+    }
+
+    @SneakyThrows
```
AdamBJohnsonx, line 83: I think we can be a little picky about what it is we might sneakily throw.  Which methods in here can throw?

iambmelt: This is forced by the interface for [Callable#call](https://developer.android.com/reference/java/util/concurrent/Callable#call()) -- nothing in this code can/will `throw`

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
```diff
@@ -0,0 +1,36 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.cache;
+
+import java.lang.reflect.Type;
+
+public interface IListTypeToken {
```
shahzaibj, line 27: javadoc for the interface?

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: Added in 9d8e891e0365ef5576ea736e3dfde7061eea21f2

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Gets the Type token associated with a List.
+     *
+     * @return The type.
```
shahzaibj, line 32: Can link out to Type using `{@link}` notation

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,190 @@
...
+ * @see SharedPreferencesSimpleCacheImpl
+ */
+public abstract class SharedPreferencesFileManagerSimpleCacheImpl<T> implements ISimpleCache<T>, IListTypeToken {
+
+    private static final String TAG = SharedPreferencesFileManagerSimpleCacheImpl.class.getSimpleName();
+    private static final String EMTPY_ARRAY = "[]";
+    private static final String TIMING_TAG = "execWithTiming";
+
+    private ISharedPreferencesFileManager mSharedPrefsFileManager;
+    private final String mKeySingleEntry;
```
shahzaibj, line 63: would like to have some comments about what is this singleKey thing

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,190 @@
...
+    private <V> V execWithTiming(@NonNull final NamedRunnable<V> runnable) {
+        final TimeUnit timeUnit;
+        final long startTime;
+        final long execTime;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            startTime = SystemClock.elapsedRealtimeNanos();
+            timeUnit = TimeUnit.NANOSECONDS;
+        } else {
+            startTime = SystemClock.elapsedRealtime();
```
AdamBJohnsonx, line 96: You could also drop to System.nanoTime().  But I think we don't support less than this version in broker for real, so...

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,190 @@
...
+            timeUnit = TimeUnit.MILLISECONDS;
+        }
+
+        final V v = runnable.call();
```
AdamBJohnsonx, line 100: could/should consider:

try {
 return runnable.call();
} finally {
 //logging
}

This way we get the logs even if there's an exception.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+import lombok.SneakyThrows;
+
+/**
+ * A simple metadata store definition that uses SharedPreferencesFileManager to persist, read,
+ * update, and delete data. Please note that all CRUD actions return success, as the underlying
+ * store's API does not surface a success indicator. If you need stronger guarantees that an
+ * operation was successful, use {@link SharedPreferencesSimpleCacheImpl} which is less performance
```
shahzaibj, line 48: Can we actually use `SharedPreferencesSimpleCacheImpl` (directly)? It seems to be an abstract class

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: Correct, the broker is Android 6+ which IIRC is 23, so whatever I'm doing here is actually.... ornamental. I'll take your suggestion to use `nanoTime()`

AdamBJohnsonx (approved):
My only real concern here is about potentially increased logging.  Do we know how much more it'll generate?

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
iambmelt: 1369aa910cf3728b630fe9069b3698220dd232d4

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
iambmelt: 1369aa910cf3728b630fe9069b3698220dd232d4

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: You'll notice this class is `abstract` too. Neither of them are used directly; instead they are subclassed by `SharedPreferencesBrokerApplicationMetadataCache` and `DefaultBrokerApplicationRegistry` which define the `Type` token for their own separate purposes (caching

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: cef32cd0d38672840aec406e5bb168ce04ae481c

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            startTime = SystemClock.elapsedRealtimeNanos();
+        } else {
+            startTime = System.nanoTime();
+        }
+
+        V v = null;
+        try {
+            v = runnable.call();
+        } catch (final Exception e) {
```
AdamBJohnsonx, line 104:
```suggestion
        } catch (final Exception e) {
            if (e instanceof InterruptedException) {
                Thread.currentThread().interrupt();
            }
```

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            execTime = SystemClock.elapsedRealtimeNanos() - startTime;
+        } else {
+            execTime = System.nanoTime() - startTime;
+        }
+
+        Logger.verbose(TAG + TIMING_TAG,
+                runnable.getName() + " finished in: " + execTime + " " + TimeUnit.NANOSECONDS.name());
```
AdamBJohnsonx, line 118: can we put all of this in a finally block, so we get timing even if there's an exception?

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
```diff
@@ -0,0 +1,39 @@
...
+ * Interface returning {@link Type} tokens for serialization hints to Gson.
+ */
+public interface IListTypeToken {
+
+    /**
+     * Gets the Type token associated with a List.
+     *
+     * @return The {@link Type}.
+     */
+    Type getListTypeToken();
```
shahzaibj, line 37: Looking at the code...I guess the return type here could be very specific to a `TypeToken` or even `TypeToken<List<T>>`

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: c14121ac4462e3f2a96ce2020cd47527e8a02d28

common/src/main/java/com/microsoft/identity/common/internal/cache/IListTypeToken.java
iambmelt: [It's a `Type`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesSimpleCacheImpl.java#L178)

AdamBJohnsonx (approved):
Still ship it.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,199 @@
...
+     * @param singleKey The name of key under which all entries will be cached.
+     */
+    public SharedPreferencesFileManagerSimpleCacheImpl(@NonNull final Context context,
+                                                       @NonNull final String prefsName,
+                                                       @NonNull final String singleKey) {
+        Logger.verbose(
+                TAG + "::ctor",
+                "Init"
+        );
+        mSharedPrefsFileManager = SharedPreferencesFileManager.getSharedPreferences(
```
iamgusain, line 78: >mSharedPrefsFileManager [](http://example.com/codeflow?start=8&length=23)

We can probably get this as constructor dependency. That will make it easier to add Unit Tests for this class..

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,199 @@
...
+ * store's API does not surface a success indicator. If you need stronger guarantees that an
+ * operation was successful, use {@link SharedPreferencesSimpleCacheImpl} which is less performance
+ * oriented.
+ * <p>
+ * Data serializes as JSON.
+ *
+ * @param <T> The type of metadata that will be persisted.
+ * @see SharedPreferencesSimpleCacheImpl
+ */
+public abstract class SharedPreferencesFileManagerSimpleCacheImpl<T> implements ISimpleCache<T>, IListTypeToken {
```
iamgusain, line 53: >SharedPreferencesFileManagerSimpleCacheImpl [](http://example.com/codeflow?start=22&length=43)

Should we also add some UTs for this?

iamgusain (approved):
Approved with suggestions

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: This class is abstract, but there are UTs for the two descendant classes of this:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/bam-cache-optimization-experiment/common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/bam-cache-optimization-experiment/common/src/main/java/com/microsoft/identity/common/internal/cache/registry/DefaultBrokerApplicationRegistry.java

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
iambmelt: Took this suggestion in cd893da2a2b26434d5b5e519f5424a4c4a524f49 and edbd610aeccbf4e31e0d716af3942f200afc85af

Per other comment, subclasses of this impl have UTs, they are here:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/bam-cache-optimization-experiment/common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/bam-cache-optimization-experiment/common/src/main/java/com/microsoft/identity/common/internal/cache/registry/DefaultBrokerApplicationRegistry.java

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -0,0 +1,210 @@
...
+        Logger.verbose(TAG + "::ctor", "Init");
+        mSharedPrefsFileManager = sharedPreferencesFileManager;
+        mKeySingleEntry = singleKey;
+    }
+
+    private interface NamedRunnable<V> extends Callable<V> {
+        String getName();
+    }
+
+    private <V> V execWithTiming(@NonNull final NamedRunnable<V> runnable) {
```
shoatman, line 102: Wondering whether this implementation should be in a separate class ... perhaps accessed via a Timing interface?

shoatman: Also wondering if we should be able to enable/disable timing via a flag.

shoatman (commented):
> Playing around with optimizations to the BAM-Cache.
> 
> Per recent review of Yammer logs, updating this cache was taking about 1s.

I think it's worthwhile to describe in the description what specifically you are trying to optimize and what approach you are taking to optimize that particular thing.

PR #1422, opened 2021-06-22
Fixing removeAccount method for msal cpp when the realm is empty

tongyuze:
Currently, when we call signout in msal cpp, we will pass the realm as an empty string to the removeAccount method in android common. 
However, in MsalOAuth2TokenCache, removeAccount method will only check if the realm is null. 
Therefore, before we pass the realm to MsalOAuth2TokenCache, we set it to null if it's an empty string.


iambmelt (changes requested):
@fengga Thanks -- changes look fine to me. Can we add additional tests to assert the following behaviors:
- When multiple accounts for different `realms` exist in the cache, that all are deleted
- When no matching accounts exist, that nothing is deleted (if there's no test for this already)

iambmelt (approved):

iamgusain (approved):

common/src/test/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -182,6 +183,97 @@ public void removeAccountTest() throws ClientException {
...
+
+        // Call remove with a different realm
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                generatedAccount.getEnvironment(),
+                REALM2
+        );
+
+        // Check the receipt, should remove nothing
+        Assert.assertEquals(0, deletionRecord.size());
```
shahzaibj, line 203: may want to also add a check to actually query the cache and check that the item is still there...this check only asserts that `removeAccount` API returned correctly...it doesn't actually assert that the account wasn't truly removed

tongyuze: Sure, will update.

shahzaibj:
I guess we can't pass `null` directly as opposed to empty strings?

common/src/test/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
```diff
@@ -182,6 +183,97 @@ public void removeAccountTest() throws ClientException {
...
+
+        // Save the second account with a different realm
+        generatedAccount.setRealm(REALM2);
+        mCppCache.saveAccountRecord(generatedAccount);
+
+        // Call remove
+        final AccountDeletionRecord deletionRecord = mCppCache.removeAccount(
+                generatedAccount.getHomeAccountId(),
+                "",
+                ""
```
shahzaibj, line 253: I guess we have empty strings because we can't pass `null` for some reason?

shahzaibj (approved):
Approving...left some comments

common/src/test/java/com/microsoft/identity/common/MsalCppOAuth2TokenCacheTest.java
tongyuze: I don't think there is some special reason. Just because msal cpp is passing it when deleting an account without a realm.

PR #1427, opened 2021-06-23
Avoid needless BAM lookups when the calling app is not foci

iambmelt:
Working on perf scenarios for broker. These changes avoid needlessly reconsulting the BAM-cache when an app is not foci

This PR is tested by:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java

A bit about what this PR is doing....
- Formerly, this code was calling `getTokenCacheForClient()` which is a convenience function for finding the on-disk cache based on the provided `clientId` and `environment`. `getTokenCacheForClient()` first looks up the app in the `BrokerApplicationMetadataCache` (BAM-Cache) and, if the app is FoCI, the fuction returns the FoCI (shared) cache. If the app is _not FoCI_, the function simply called `initializeProcessUidCache()` which returns the app-specific cache for a non-FoCI app (or creates it, if it doesn't exist). This PR shortcuts the BAM-Cache lookup for flows that _cannot be FoCI_ and simply calls the initializer for the non-foci cache, passing the `context` and `uid`


iambmelt:
> Could you write a little bit in this PR about how this works?

yep. will update the description 👍

iambmelt:
> > Could you write a little bit in this PR about how this works?
> 
> yep. will update the description 👍

@AdamBJohnsonx updated

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -278,24 +278,11 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
             );
         } else {
             // Save to the processUid cache... or create a new one
-            MsalOAuth2TokenCache targetCache = getTokenCacheForClient(
-                    idTokenRecord.getClientId(),
-                    idTokenRecord.getEnvironment(),
```
AdamBJohnsonx, line 283: I see.  We know !isFoci, so the checks for the foci-nature inside of getTokenCacheForClient were a complete waste of time.

iambmelt: You got it

AdamBJohnsonx: It's interesting that in this case initializeProcessUidCache was unconditionally called... was there any way we would ever execute the below else statement?  OK, only in the case where the metadata lookup was null.

AdamBJohnsonx (approved):

iamgusain (approved):

PR #1429, opened 2021-06-24
Removing low-perf fix for duplicate BAM data

iambmelt:
The fix originally done in #1232 can be implemented to avoid performing multiple reads of BAM data. This PR ensures the data is read just 1x and relies on a few things...

- `SharedPreferencesFileManagerSimpleCacheImpl` can now be initialized with a flag allowing calls to `insert()` to replace the existing element, even if the existing/new element have the same `hashCode()` - for backcompat, this behavior is disabled by default
- `BrokerApplicationMetadata` no longer considers foci state to be criteria for equality - this likely should have always been this way

These changes are covered by [existing tests here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java#L47)


AdamBJohnsonx:
Changelog?

iambmelt:
> Changelog?

in 5b8807a43eacf870f4d20572d6cc932dbf055db5

iambmelt:
Somewhat related to #1427 - different kinds of optimizations for reading/writing BAM data

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManagerSimpleCacheImpl.java
```diff
@@ -142,6 +173,15 @@ public String getName() {
             @Override
             public Boolean call() {
                 final Set<T> allMetadata = new HashSet<>(getAll());
```
AdamBJohnsonx, line 175: Hm.  Why not translate it into a map, and just separate keys from values?  I mean, since a set is a map…. That seems less strange to me than excluding something from equality and then doing work to change the set contract…

iambmelt: Not sure I follow... this may not be clear, but there are no keys for this data. It's just a collection of values on disk

AdamBJohnsonx: Sure, but we're free to interpret that collection any way we'd like.  If we want to read it in order to create a Map instead of a Set, that's on us.  I was just thinking that instead of making the Set act like a Map, just arrange the data so it's in a Map instead.

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerApplicationMetadata.java
```diff
@@ -41,33 +41,17 @@ public void setFoci(final String mFoci) {
...
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        if (!super.equals(o)) return false;
-
-        BrokerApplicationMetadata metadata = (BrokerApplicationMetadata) o;
-
-        return mFoci != null ? mFoci.equals(metadata.mFoci) : metadata.mFoci == null;
+        // This class intentionally no longer includes foci state to determine equality.
```
AdamBJohnsonx, line 46: If we're making equals behave non-intuitively, then we need to have this called out strongly in the method and class javadoc.  I think there are several places where we've dealt with equals in interesting ways.

iambmelt: 2d12a83145df649da24048f059d15154b6a86b03

AdamBJohnsonx (approved):

PR #1440, opened 2021-06-25
Do not report network errors an authority validation errors

AdamBJohnsonx:
In this ICM:https://portal.microsofticm.com/imp/v3/incidents/details/243581504/home we observe that a failure to contact the network for authority validation will result in a rejection stating that an attempt to make a cross-cloud call was made and rejected.

What we're doing here is clarifying the error. Rather than throwing an ArgumentException though, it was suggested that a RuntimeException would be less invasive, but that would lead to UNKNOWN_ERROR results.  Since at the point that we throw the exception we know what the error code should be, we create an instance of RuntimeException that has this additional information so that it can be properly reported to the caller and in the logs.

Tests were added to verify that the correct handling is happening for the new TerminalException class in the CommandDispatcher.

<strike>This is not intended to merge directly into the release/3.4.4-RC1 branch, but into the next release branch cut after the merge back to master.  Since it doesn't yet exist, this should be equivalent.</strike>
This targets release/3.4.5

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -46,9 +46,7 @@ import com.microsoft.identity.common.internal.result.AcquireTokenResult;
 import com.microsoft.identity.common.internal.result.FinalizableResultFuture;
 import com.microsoft.identity.common.internal.result.GenerateShrResult;
 
-import org.junit.After;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -328,7 +326,7 @@ public class CommandDispatcherTest {
     public void testSubmitSilentWithTerminalException() {
         final String errorCode = "anError";
         final CountDownLatch testLatch = new CountDownLatch(1);
-        CommandDispatcher.submitSilent(new IErrorExceptionCommand(getEmptyTestParams(),
+        CommandDispatcher.submitSilent(new CommandThrowingIErrorInformationException(getEmptyTestParams(),
                 new CommandCallback<String, Exception>() {
                     @Override
                     public void onCancel() {
@@ -506,11 +504,11 @@ public class CommandDispatcherTest {
         }
     }
 
-    static class IErrorExceptionCommand extends BaseCommand<String> {
+    static class CommandThrowingIErrorInformationException extends BaseCommand<String> {
         final String mErrorCode;
 
-        public IErrorExceptionCommand(@NonNull final CommandParameters parameters,
-                                @NonNull final CommandCallback callback, String errorCode) {
+        public CommandThrowingIErrorInformationException(@NonNull final CommandParameters parameters,
+                                                         @NonNull final CommandCallback callback, String errorCode) {
             super(parameters, getTestController(), callback, "test_id");
             mErrorCode = errorCode;
         }
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -27,6 +27,8 @@ import android.content.Intent;
 import androidx.annotation.NonNull;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
+import com.microsoft.identity.common.exception.ClientException;
+import com.microsoft.identity.common.exception.TerminalException;
 import com.microsoft.identity.common.internal.cache.ICacheRecord;
 import com.microsoft.identity.common.internal.commands.BaseCommand;
 import com.microsoft.identity.common.internal.commands.CommandCallback;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/exception/IErrorInformation.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.identity.common.exception;
```
shahzaibj, line 1: LICENSE

common/src/main/java/com/microsoft/identity/common/exception/TerminalException.java
```diff
@@ -0,0 +1,46 @@
+package com.microsoft.identity.common.exception;
```
shahzaibj, line 1: LICENSE

common/src/main/java/com/microsoft/identity/common/exception/TerminalException.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public TerminalException(final @NonNull Throwable cause,
+                             final @Nullable String errorCode) {
+        super(cause);
+        this.mErrorCode = errorCode;
+    }
+
+    @Override
+    public @Nullable String getErrorCode() {
+        return null;
```
shahzaibj, line 44:
```suggestion
        return mErrorCode;
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -246,6 +247,16 @@ public static BaseException baseExceptionFromException(final Throwable exception
             e = exception.getCause();
         }
 
+        if (e instanceof TerminalException) {
+            String errorCode = ((TerminalException) e).getErrorCode();
```
shahzaibj, line 251:
```suggestion
            final String errorCode = ((TerminalException) e).getErrorCode();
```

common/src/main/java/com/microsoft/identity/common/exception/TerminalException.java
AdamBJohnsonx: Or just use lombok

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -478,6 +506,26 @@ public boolean isEligibleForEstsTelemetry() {
         }
     }
 
+    static class IErrorExceptionCommand extends BaseCommand<String> {
+        final String mErrorCode;
+
+        public IErrorExceptionCommand(@NonNull final CommandParameters parameters,
+                                @NonNull final CommandCallback callback, String errorCode) {
+            super(parameters, getTestController(), callback, "test_id");
```
shahzaibj, line 514: Should we pass in `errorCode` here as opposed to `test_id`?

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -478,6 +506,26 @@ public boolean isEligibleForEstsTelemetry() {
         }
     }
 
+    static class IErrorExceptionCommand extends BaseCommand<String> {
```
shahzaibj, line 509: Does not seem an interface 

```suggestion
    static class ErrorExceptionCommand extends BaseCommand<String> {
```

shahzaibj (approved):
left a couple of minor comments

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
AdamBJohnsonx: Yes.

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
AdamBJohnsonx: I've renamed it.

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
AdamBJohnsonx: ... wait, no, that's a correlation id, not an error code.

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/SilentTokenCommandParameters.java
```diff
@@ -72,12 +82,15 @@ private boolean authorityMatchesAccountEnvironment() {
...
             final AzureActiveDirectoryCloud cloud = AzureActiveDirectory.getAzureActiveDirectoryCloudFromHostName(getAccount().getEnvironment());
             return cloud != null && cloud.getPreferredNetworkHostName().equals(getAuthority().getAuthorityURL().getAuthority());
-        } catch (IOException e) {
+        } catch (final IOException e) {
             Logger.error(
                     TAG + methodName,
                     "Unable to perform cloud discovery",
                     e);
-            return false;
+            throw new TerminalException(
```
iambmelt, line 90: Asked this question and got an answer from @AdamBJohnsonx offline - this new exception is to avoid signature changes on `validate()` which would have become more invasive

iambmelt (approved):

AdamBJohnsonx:
https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items?backlogLevel=Backlog%20items

common/src/main/java/com/microsoft/identity/common/exception/TerminalException.java
```diff
@@ -0,0 +1,65 @@
...
+@Getter
+@Accessors(prefix = "m")
+public class TerminalException extends RuntimeException implements IErrorInformation {
+
+    private final String mErrorCode;
+    /**
+     * Construct a TerminalException with a message and cause.
+     * @param message the exception message.
+     * @param cause the causing exception.  Should not be null.
+     * @param errorCode the error code.  May be null.  This should be an error string from {@link ClientException}.
```
AdamBJohnsonx, line 46:
```suggestion
     * @param errorCode the error code.  May not be null.  This should be an error string from {@link ClientException}.
```

PR #1441, opened 2021-06-26
Adding additional logging - perf recording

iambmelt:
Plugging a long gap in logs from a customer report of slow token requests. There is currently no non-verbose logging during cloud discovery

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/Authority.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/Authority.java
@@ -279,13 +279,16 @@ public abstract class Authority {
 
     private static void performCloudDiscovery() throws IOException {
         final String methodName = ":performCloudDiscovery";
-        Logger.verbose(
+        Logger.info(
                 TAG + methodName,
                 "Performing cloud discovery..."
         );
         synchronized (sLock) {
+            Logger.info(TAG + methodName, "Acquired lock.");
             if (!AzureActiveDirectory.isInitialized()) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/authorities/Authority.java
```diff
@@ -279,13 +279,16 @@ protected boolean getKnownToDeveloper() {
 
     private static void performCloudDiscovery() throws IOException {
         final String methodName = ":performCloudDiscovery";
-        Logger.verbose(
+        Logger.info(
                 TAG + methodName,
                 "Performing cloud discovery..."
         );
         synchronized (sLock) {
+            Logger.info(TAG + methodName, "Acquired lock.");
```
shahzaibj, line 287: I guess this can be `verbose`

shahzaibj (approved):

AdamBJohnsonx (approved):
I’m wondering if we can insert perf markers here of if it isn’t supported yet.

common/src/main/java/com/microsoft/identity/common/internal/authorities/Authority.java
iambmelt: CP isn't logging anything below `info` -- in this case, I want to know if the lock is taking a long time to be acquired

iambmelt:
> I’m wondering if we can insert perf markers here of if it isn’t supported yet.

Not on this branch, but @paulkagiri this would be good to have on `dev`

PR #1444, opened 2021-06-28
Use mode private, fixes an issue where malformed json was coming back from the ADAL cache

iambmelt:
This PR corrects a bug discovered in https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1616 that when different `MODES` where passed to `SharedPreferencesFileManager` that multiple "singletons" would exist. The resulting bug was inconsistency in the different references of the cache as to what the underlying file on disk contained.

This PR addresses this by doing the following:
- Where `-1` is passed to the cache, treat this as `MODE_PRIVATE` and give it a cache key name that matches this behavior
- Where `MODE_PRIVATE` is passed to the cache, do the same as above
- Expose a new static constructor that defaults to `MODE_PRIVATE`


AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/ADALOAuth2TokenCache.java
```diff
@@ -119,7 +119,7 @@ protected void initializeSharedPreferencesFileManager(final String fileName) {
                 SharedPreferencesFileManager.getSharedPreferences(
                         getContext(),
                         fileName,
-                        -1,
+                        Context.MODE_PRIVATE,
```
shahzaibj, line 122: hmm...isn't it same as per this code here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/10ccbf9fa3236a980500da68c10063153dd68ace/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java#L163

shahzaibj: I believe that's what's being called internally here

AdamBJohnsonx: It should be.  There's no reason not to be explicit, though.  I was worried that I'd done something really strange that was working and shouldn't have been.

iambmelt: Given that [`testUpdateTokenCacheUsesResultAuthority` fails in this PR](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1616) until I remove this line, it is seemingly not the same -- not very obvious what's going on here...

@AdamBJohnsonx as FYI....

iambmelt: Ahhh I figured this out... I think what's happening here is that the line above[ where the `key` gets created doesn't account for this behavior](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/10ccbf9fa3236a980500da68c10063153dd68ace/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java#L82)... so unless the cache is initialized with the same args everywhere you end up with duplicates "singletons"

AdamBJohnsonx: Realistically, do we just want to remove the option for different modes?

iambmelt: @AdamBJohnsonx Probably, but would be take a breaking change to do it?

iambmelt: Made this change in an additive way...

AdamBJohnsonx (approved):

common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesFileManagerTests.java
```diff
@@ -45,6 +45,8 @@
 import static junit.framework.Assert.assertFalse;
 import static junit.framework.Assert.assertTrue;
 
+import android.content.Context;
```
shahzaibj, line 48: Is this import needed?

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -77,7 +77,7 @@ public static void setIsEnabled(final boolean value) {
     public HelloCache(final @NonNull Context context,
                       final @NonNull String protocolName,
                       final @NonNull String targetAppPackageName) {
-        mFileManager = SharedPreferencesFileManager.getSharedPreferences(context, SHARED_PREFERENCE_NAME, -1, null);
+        mFileManager = SharedPreferencesFileManager.getSharedPreferences(context, SHARED_PREFERENCE_NAME,null);
```
shahzaibj, line 80:
```suggestion
        mFileManager = SharedPreferencesFileManager.getSharedPreferences(context, SHARED_PREFERENCE_NAME, null);
```

shahzaibj (approved):

common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesFileManagerTests.java
iambmelt: nope

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -90,6 +90,20 @@ public static SharedPreferencesFileManager getSharedPreferences(final Context co
...
 
+    /**
+     * Constructs an instance of SharedPreferencesFileManager. Operating mode is always MODE_PRIVATE.
+     *
+     * @param context Interface to global information about an application environment.
+     * @param name The desired {@link android.content.SharedPreferences} file. It will be created if it does not exist.
+     * @param storageHelper The {@link IStorageHelper} to handle encryption/decryption of values.
+     * @return The SharedPreferencesFileManager instance.
+     */
+    public static SharedPreferencesFileManager getSharedPreferences(final Context context,
```
iamgusain, line 101: nit: If we are only going to use this to getSharedPreferences.. we can mark the other one (line78) as deprecated

iamgusain (approved):
Approved with suggestion

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
iambmelt: 15b27a5ca2fa926df66223b7f9380c30b810b20c

PR #1449, opened 2021-07-01
Introduce a more human readable way to specify in tests whether password prompt is expected or not

aherciya:
Adding logic to account for passwordPage for Automation Tests. Associates to [this](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1452/commits) PR in MSAL. 

We used to kind of sort of re-use the prompt parameter for this purpose...however, better to have a dedicated field for this purpose that says:

`passwordPageExpected(true/false)`

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MINOR] Add helper for Automation Tests to account for passwordPageExpected (#1449)
 - [MAJOR] Migrate Exceptions from common to common4j (#1442)
 - [PATCH] Fixing removeAccount method for msal cpp when the realm is empty (#1422)
 - [MINOR] Migrate Requests/Responses (#1424)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,5 @@
 V.Next
 ----------
-- [MINOR] Add helper for Automation Tests to account for passwordPageExpected (#1449)
 - [MAJOR] Migrate Exceptions from common to common4j (#1442)
 - [PATCH] Fixing removeAccount method for msal cpp when the realm is empty (#1422)
 - [MINOR] Migrate Requests/Responses (#1424)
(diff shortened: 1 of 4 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/MicrosoftStsPromptHandler.java
```diff
@@ -76,6 +76,20 @@ public void handlePrompt(@NonNull final String username, @NonNull final String p
...
+                }
+            }
+            if(parameters.isPasswordPageExpected()) {
+                loginComponentHandler.handlePasswordField(password);
+            }
+        }
+
         if (parameters.getPrompt() == PromptParameter.LOGIN || !parameters.isSessionExpected()) {
             loginComponentHandler.handlePasswordField(password);
         }
```
shahzaibj, line 95: Personally don't like this logic being tied into the `if block for WHEN_REQURIED`...in fact, we probably don't need to check for that at all (we are already doing the consent check further below). I think the change that I've suggested here below should do it.

```suggestion
        if (parameters.isPasswordPageExpected() || parameters.getPrompt() == PromptParameter.LOGIN || !parameters.isSessionExpected()) {
            loginComponentHandler.handlePasswordField(password);
        }
```

shahzaibj (changes requested):
Left a comment

changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MINOR] Add helper for Automation Tests to account for passwordPageExpected (#1449)
```
shahzaibj, line 3: We're not shipping this. 

```suggestion
```

shahzaibj (approved):
LGTM - please remove the changelog entry prior to checking in as this is a test only change and is not going to get shipped

iamgusain (approved):

p3dr0rv (approved):

tanmaymanolkar1 (approved):

PR #1455, opened 2021-07-06
Update perf testing documentation

paulkagiri:

PR #1456, opened 2021-07-07
Allow Retry Generation of Device PoP Key without Attempting Attestation-Cert Generation

iambmelt:
Impetus:
Teams has reported the following issues when integrating with PoP...
- [ADO - AT/PoP Failure: "Failed to generate attestation certificate"](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1425870)
- [ADO - AT/PoP Failure: "Failed to generate KeyPair"](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1453707)

#### Issue 1: Failure to generate attestation certificate
This is caused when the [underlying KeyStore does not implement support for attestation certificates](https://android.googlesource.com/platform/frameworks/base/+/ada0fc1/keystore/java/android/security/keystore/AndroidKeyStoreKeyPairGeneratorSpi.java#601). This API was [proposed in the FIDO uaf-v1.1 draft](https://fidoalliance.org/specs/fido-uaf-v1.1-rd-20161005/fido-uaf-reg-v1.1-rd-20161005.html#android-key-attestation) and is not presently a 'locked' part of the spec; as a result, support for it varies across devices/OEMs. The mitigation for this is to simply retry without this flag.

#### Issue 2: Failure to generate KeyPair
This has [multiple potential causes](https://android.googlesource.com/platform/hardware/libhardware/+/master/include/hardware/keymaster_defs.h#437) - on the Huawei Mate 9 Pro I have to repro, these are caused by [`KM_ERROR_SECURE_HW_COMMUNICATION_FAILED (-49)`](https://android.googlesource.com/platform/hardware/libhardware/+/master/include/hardware/keymaster_defs.h#485) which is a general error message from the KeyMaster HAL that communication with the TPM failed.

Also included in this PR:
- I have added an API to `Device` which can be used to test whether PoP can be used on this device _without_ making a token request. I am going to suggest that we eventually expose this on MSAL's `PublicClientApplication` such that Teams (or any other consumer of this API) can test for device support before trying these flows (potentially resulting in double-prompts)


common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1076,8 +1089,22 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
                                 e
                         );
                         tryImport = false;
+
+                        if (null != e.getCause() && e.getCause().getClass().getSimpleName().equals(STRONG_BOX_UNAVAILABLE_EXCEPTION)) {
+                            // On some devices (notably, Huawei Mate 9 Pro), StrongBox errors are
+                            // the cause of the surfaced SecureKeyImportUnavailableException.
+                            tryStrongBox = false;
```
AdamBJohnsonx, line 1096: Do we want a log similar to the above block here?

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1076,8 +1089,22 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
...
+                        }
+
+                        continue;
+                    } else if (trySetAttestationChallenge && FAILED_TO_GENERATE_ATTESTATION_CERTIFICATE_CHAIN.equalsIgnoreCase(e.getMessage())) {
+                        trySetAttestationChallenge = false;
```
AdamBJohnsonx, line 1101: Again, maybe a log message.

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -142,6 +142,16 @@
...
 
+    /**
+     * Error message from underlying KeyStore that StrongBox HAL is unavailable.
+     */
+    public static final String STRONG_BOX_UNAVAILABLE_EXCEPTION = "StrongBoxUnavailableException";
+
+    /**
+     * Error message from underlying KeyStore that
+     */
+    public static final String FAILED_TO_GENERATE_ATTESTATION_CERTIFICATE_CHAIN = "Failed to generate attestation certificate chain";
```
AdamBJohnsonx, line 153: Is this documented somewhere?

common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
```diff
@@ -64,4 +71,25 @@ public static synchronized IDevicePopManager getDevicePoPManagerInstance() throw
         }
     }
 
+    /**
+     * Determines if requests made on this device can successfully use AT/PoP APIs.
+     * @param context The current application's {@link Context}.
+     * @return True if AT/PoP requests can be made from this device. False otherwise.
+     */
+    public static boolean isDevicePoPSupported(@NonNull final Context context) {
```
AdamBJohnsonx, line 79: Code marker?

AdamBJohnsonx (commented):
Some minor comments about not regressing the logs.  Do we want to put some perf markers around some of these things, like doing a key thumbprint and key pair generation in order to determine if we can support POP?

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -142,6 +142,16 @@
      */
     private static final String PRIVATE_KEY_NOT_FOUND = "Not an instance of a PrivateKeyEntry";
 
+    /**
+     * Error message from underlying KeyStore that StrongBox HAL is unavailable.
+     */
+    public static final String STRONG_BOX_UNAVAILABLE_EXCEPTION = "StrongBoxUnavailableException";
+
+    /**
+     * Error message from underlying KeyStore that
```
rpdome, line 151: nit: the rest of the sentence is missing.

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1076,8 +1089,22 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
                                 e
                         );
                         tryImport = false;
+
+                        if (null != e.getCause() && e.getCause().getClass().getSimpleName().equals(STRONG_BOX_UNAVAILABLE_EXCEPTION)) {
```
rpdome, line 1093: missing tryStrongBox check?

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1052,16 +1062,19 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
...
+                    kp = generateNewKeyPair(context, tryStrongBox, tryImport, trySetAttestationChallenge);
                     generated = true;
                 } catch (final ProviderException e) {
                     // This mechanism is terrible.  But there are stern warnings that even attempting to
                     // mention these classes in a catch clause might cause failures. So we're going to look
                     // at the exception names.
-                    if (tryStrongBox && e.getClass().getSimpleName().equals("StrongBoxUnavailableException")) {
+
+
+                    if (tryStrongBox && e.getClass().getSimpleName().equals(STRONG_BOX_UNAVAILABLE_EXCEPTION)) {
```
rpdome, line 1077: nit: it's probably worth turning this
```
returns null != e && e.getClass().getSimpleName().equals(STRONG_BOX_UNAVAILABLE_EXCEPTION)
```
into a function -  for readability.

rpdome (commented):
Left some comments. Mostly minor.

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: Whoops. Fixed in 19fd98fbac5b1202ed8a72f2bab34a9346aec9c4

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: Sadly, [no ](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setAttestationChallenge(byte[])) -- the documentation gives no indication that this can fail on N+ devices when used with RSA; however [this error](https://android.googlesource.com/platform/frameworks/base/+/ada0fc1/keystore/java/android/security/keystore/AndroidKeyStoreKeyPairGeneratorSpi.java#601) is cropping up on certain devices

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: 55f2e3671ee62420e19823627fa4f9141c744453

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: Logging via e07f16c3a30df0e752d531fc831ce301a450e62f

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: a894a2dde5402f61930e875e794f24e9e6694e32

common/src/main/java/com/microsoft/identity/common/internal/platform/Device.java
iambmelt: For me - instructions on doing this here:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/paul/update-perf-testing-documentation/uiautomationutilities/perf_tool/readme.md#steps-to-add-a-new-marker-and-see-it-in-the-report

iambmelt: Took a stab at this here https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1456/commits/5fb8270e059caf42c42bf8a2f7c3cd3557876667

@paulkagiri This gave me a chance to dogfood your instructions (which I [provided feedback on here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1455)!) Can you review my changes and see if I followed your instructions correctly?

AdamBJohnsonx (approved):
This is fine.  The only thing we should make sure we do is surface the same information (basically, is attestation supported, I think) through the KeyStoreAccessor interfaces.

iambmelt:
> This is fine. The only thing we should make sure we do is surface the same information (basically, is attestation supported, I think) through the KeyStoreAccessor interfaces.

I will need to confirm what's the expected functionality in that case - at a glance, the caller will need to support a few different return-values from [`getCertificateChain()`](https://developer.android.com/reference/java/security/KeyStore?hl=en#getCertificateChain(java.lang.String)) incl. `null`, the public portion of the aliased-key's cert, or `ClientException`

iambmelt:
> > This is fine. The only thing we should make sure we do is surface the same information (basically, is attestation supported, I think) through the KeyStoreAccessor interfaces.
> 
> I will need to confirm what's the expected functionality in that case - at a glance, the caller will need to support a few different return-values from [`getCertificateChain()`](https://developer.android.com/reference/java/security/KeyStore?hl=en#getCertificateChain(java.lang.String)) incl. `null`, the public portion of the aliased-key's cert, or `ClientException`

Confirmed the behavior - you will get an array of 1 cert containing the public key portion of the generated key and that's it...
image

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1109,6 +1141,16 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
         );
     }
 
+    private static boolean isStrongBoxUnavailableException(@NonNull final Throwable t) {
+        final boolean isStrongBoxException = t.getClass().getSimpleName().equals(STRONG_BOX_UNAVAILABLE_EXCEPTION);
```
shahzaibj, line 1145: Can we do an instance check here isntead?

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -1109,6 +1141,16 @@ private KeyPair generateNewRsaKeyPair(@NonNull final Context context,
         );
     }
 
+    private static boolean isStrongBoxUnavailableException(@NonNull final Throwable t) {
```
shahzaibj, line 1144: Should we rename the method to `isStrongBoxUnavailable` ? seems like that's the intent? it just uses the exception to figure that out

shahzaibj (approved):
Minor comments

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: We cannot due to wackiness discovered by @AdamBJohnsonx - documented via comment here:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java#L1063-L1065

iambmelt: Basically - hard refs to the `StrongBoxUnavailableException` class can cause issues on certain devices, so this workaround is being used

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: I'll go ahead and leave this as-is, as this method's check accepts the _error_ as the argument, rather than being a 'general purpose' API for discovering whether/not StrongBox is supported

PR #1461, opened 2021-07-09
Migrate classes to common4j as a prereq for AccountManagerStorageHelper work.

rpdome:
Related PR:
ADAL https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1622
MSAL https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1460
Broker https://github.com/AzureAD/ad-accounts-for-android/pull/1622

-----
### 1. Migrate the following layers to common4j
- CacheRecord, ICacheRecord
- classes under 'dto'
- JweResponse
- classes under providers.key
- AccessToken, RefreshToken, IdToken classes.

### 2. Resolve the previously suppressed spotbugs issues.
- Use the cloned object for mutable class variables (arrays, date, etc).
- Turn JweResponse.JweHeader into a "proper" class (with builder, accessors, etc).
- Resolve inconsistent synchronization of MicrosoftAccount's member due to Lombok's equalsAndHashCode

### 3. Kill our Base64 class. Switch dependency from JoseJWT to mesebera's.
- The reason is that the latter is more flexible (support flags such as NO_PADDING, which is required in some flows)


common/src/main/java/com/microsoft/identity/common/adal/internal/util/JsonExtensions.java
```diff
@@ -49,28 +43,6 @@ private JsonExtensions() {
...
-        final Iterator<String> keyIterator = jsonObject.keys();
-
-        final Map<String, String> responseItems = new HashMap<>();
-        while (keyIterator.hasNext()) {
-            final String key = keyIterator.next();
-            responseItems.put(key, jsonObject.getString(key));
-        }
-
-        return responseItems;
-    }
```
shahzaibj, line 72: No longer needed?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerResult.java
```diff
@@ -356,7 +357,9 @@ public String getExceptionType() {
     }
 
     public List<ICacheRecord> getTenantProfileData() {
-        return mTenantProfileData;
+        return mTenantProfileData == null ?
+                null :
+                new ArrayList<>(mTenantProfileData);
```
shahzaibj, line 362:
```suggestion
        return mTenantProfileData == null 
                ? null 
                : new ArrayList<>(mTenantProfileData);
```

common/src/main/java/com/microsoft/identity/common/adal/internal/util/JsonExtensions.java
rpdome: Moved to jsonutil.

common/src/test/java/com/microsoft/identity/common/MicrosoftStsAccountCredentialAdapterTest.java
```diff
@@ -53,7 +53,9 @@
 import java.util.Date;
 import java.util.GregorianCalendar;
 
-import static com.microsoft.identity.common.internal.providers.microsoft.MicrosoftAccount.AUTHORITY_TYPE_MS_STS;
+import cz.msebera.android.httpclient.extras.Base64;
```
shahzaibj, line 56: I'm fine with using msebra over JoseJWT but not a fan of dropping our own Base64 class...which I believe was just a wrapper around an implementation of Base64 (which was JoseJWT until now I believe). Can we keep that wrapper? Or instead of using Base64 directly from msebra, can we create our own interface for Base64 and then provide an implementation of it using msebra? and then probably make a factory that returns Base64 impl? so basically in the future if we decide to something other than mseabra then we just need to change internal implementation and not have to change the imports in a lot of places.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -39,6 +39,22 @@
...
+     */
+    public static final String ENCODING_ASCII_STRING = "ASCII";
+
+    /**
+     * The Constant CHARSET_ASCII.
+     */
+    public static final Charset CHARSET_ASCII = Charset.forName(ENCODING_ASCII_STRING);
+
+    /**
+     * A constant for PMD to be happy with.
```
shahzaibj, line 53: Can we elaborate on this?

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -50,6 +66,26 @@
...
+    public static final class OAuth2 {
+        /**
+         * String of AAD version.
+         */
+        public static final String AAD_VERSION = "ver";
+
+        /**
+         * Constant for  v1 endpoint
+         */
+        public static final String AAD_VERSION_V1 = ONE_POINT_ZERO;
```
shahzaibj, line 82: I kind of feel like `ONE_POINT_ZERO` reduces readability. I kind of read it better if `"1.0"` was hard coded here...may be that's just me

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -50,6 +66,26 @@
...
+
+        /**
+         * Constsnt for v2 endpoint
```
shahzaibj, line 85:
```suggestion
         * Constant for v2 endpoint
```

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -50,6 +66,26 @@
         public static final String CLIENT_REQUEST_ID = "client-request-id";
     }
 
+    /**
+     * Represents the constants value for Active Directory.
+     */
+    @NoArgsConstructor(access = AccessLevel.PRIVATE)
+    public static final class OAuth2 {
```
shahzaibj, line 73: Javadoc says constants for AAD and that's what the class seems to contain, however, the class is named OAuth2. Should it be renamed to something like `AzureActiveDirectory`?

common4j/src/main/com/microsoft/identity/common/java/WarningType.java
```diff
@@ -1,29 +1,6 @@
...
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
 package com.microsoft.identity.common.java;
-
-/**
- * Class to hold names of compile time warning types
+/*
+Class to hold names of compile time warning types
```
shahzaibj, line 3:
```suggestion
Class to hold names of compile time warning types.
```

common4j/src/main/com/microsoft/identity/common/java/cache/CacheRecord.java
```diff
@@ -40,12 +39,12 @@
...
-    private final AccessTokenRecord mAccessToken;
-    private final RefreshTokenRecord mRefreshToken;
-    private final IdTokenRecord mIdToken;
-    private final IdTokenRecord mV1IdToken;
+    @NonNull
+    AccountRecord mAccount;
+    AccessTokenRecord mAccessToken;
+    RefreshTokenRecord mRefreshToken;
+    IdTokenRecord mIdToken;
+    IdTokenRecord mV1IdToken;
```
shahzaibj, line 47: Why remove `private` and `final`?

shahzaibj (approved):
Approving with suggestions

common4j/src/main/com/microsoft/identity/common/java/cache/CacheRecord.java
rpdome: Redundant with @ Value

rpdome: Or I could remove that and add this instead.

common/src/test/java/com/microsoft/identity/common/MicrosoftStsAccountCredentialAdapterTest.java
rpdome: I was debating with myself on this, and eventually decided to remove it.

I could add back our class, but due to its flexibility, the interface would be almost 1:1 mapping with msebra. (i.e. we'll have to create a flag enum that mirrors theirs, or we'll end up creating a lot of helper functions, etc).

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
rpdome: I think it's just there to suppress PMD warning.

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1685,7 +1711,7 @@ public void setSingleSignOnState(@NonNull final String uidStr,
 
         MsalOAuth2TokenCache targetCache;
 
-        final int uid = Integer.valueOf(uidStr);
+        final int uid = Integer.parseInt(uidStr);
```
AdamBJohnsonx, line 1714: This is actually a great change.  It'll result in one fewer auto boxed Integer being created and destroyed - however did you find it?

common4j/src/main/com/microsoft/identity/common/java/providers/keys/ClientCertificateMetadata.java
```diff
@@ -55,7 +58,7 @@ public String getAlias() {
      * @return char[]
      */
     public char[] getPassword() {
-        return mPassword;
+        return mPassword == null ? null : mPassword.clone();
```
AdamBJohnsonx, line 61: Do we want to use clone here, or Arrays.copyOf() ?  I'd default to the Arrays.copyOf method.

common4j/src/main/com/microsoft/identity/common/java/providers/keys/ClientCertificateMetadata.java
```diff
@@ -37,7 +40,7 @@
      */
     public ClientCertificateMetadata(String alias, char[] password) {
         mAlias = alias;
-        mPassword = password;
+        mPassword = password == null ? null : password.clone();
```
AdamBJohnsonx, line 43: Similar comment about Arrays.copyOf rather than clone.

common4j/src/main/com/microsoft/identity/common/java/providers/keys/KeyStoreConfiguration.java
```diff
@@ -43,7 +45,7 @@
     public KeyStoreConfiguration(String keyStoreType, String keyStoreProvider, char[] keyStorePassword) {
         mKeyStoreType = keyStoreType;
         mKeyStoreProvider = keyStoreProvider;
-        mKeyStorePassword = keyStorePassword;
+        mKeyStorePassword = keyStorePassword == null ? null : keyStorePassword.clone();
```
AdamBJohnsonx, line 48: Similar comments about clone.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAccount.java
```diff
@@ -246,7 +253,10 @@ synchronized void setUserId(final String userid) {
      */
     @Override
     public synchronized String getUniqueIdentifier() {
-        return StringExtensions.base64UrlEncodeToString(mUid) + "." + StringExtensions.base64UrlEncodeToString(mUtid);
+
+        return Base64.encodeToString(StringUtil.toByteArray(mUid), Base64.URL_SAFE | Base64.NO_WRAP)
```
AdamBJohnsonx, line 257: nit: should this be a utility method somewhere...?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAccessToken.java
```diff
@@ -41,22 +41,22 @@
     public AzureActiveDirectoryAccessToken(
             @NonNull final AzureActiveDirectoryTokenResponse response) {
         super(response);
-        mExpiresOn = response.getExpiresOn();
-        mExtendedExpiresOn = response.getExtExpiresOn();
+        mExpiresOn = response.getExpiresOn() == null ? null : new Date(response.getExpiresOn().getTime());
```
AdamBJohnsonx, line 44: the response is already doing one defensive copy.  Do we need to copy it again, since we're not storing it in this class?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAccessToken.java
```diff
@@ -41,22 +41,22 @@
...
+        mExtendedExpiresOn = response.getExtExpiresOn() == null ? null : new Date(response.getExtExpiresOn().getTime());
```
AdamBJohnsonx, line 45: This is where it would be convenient to move to java.time.  I believe that Instant instances are immutable, so can be safely returned.

common4j/src/main/com/microsoft/identity/common/java/util/JsonUtil.java
```diff
@@ -0,0 +1,58 @@
...
+     * Extract JSON Object into Map<String, String>.
+     *
+     * @param jsonString String
+     * @return Map
+     * @throws JSONException if JSON string is malformed.
+     */
+    public static Map<String, String> extractJsonObjectIntoMap(final String jsonString)
+            throws JSONException {
+
+        final JSONObject jsonObject = new JSONObject(jsonString);
```
AdamBJohnsonx, line 47: Is there some reason that we're using org.json here rather than Jackson or Gson?  Gson has the ability to directly create a generic-typed map via use of a TypeToken rather than the double creation we have here.

AdamBJohnsonx: OK, I see what's going on here, we're trying to preserve json structures as string values in the map.  We should say that in a comment, since it's tricky.

common4j/src/main/com/microsoft/identity/common/java/util/SchemaUtil.java
```diff
@@ -288,11 +285,11 @@ public static String getTenantId(@Nullable final String clientInfoString,
     }
 
     public static String getDisplayableId(@NonNull final Map<String, ?> claims) {
-        if (!StringExtensions.isNullOrBlank((String) claims.get(MicrosoftStsIdToken.PREFERRED_USERNAME))) {
+        if (!StringUtil.isNullOrEmpty((String) claims.get(MicrosoftStsIdToken.PREFERRED_USERNAME))) {
```
AdamBJohnsonx, line 288: since stringutil is seeing such increased usage, we should rewrite this method to not create an extra object in order to pull this off.

AdamBJohnsonx (approved):
Some minor comments.  Seriously consider ditching clone().

common/src/main/java/com/microsoft/identity/common/internal/cache/BrokerOAuth2TokenCache.java
rpdome: It's all thanks to spotbugs lol

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAccessToken.java
rpdome: i'll create a card to track this. we can do once we go full java 8.

PR #1437, opened 2021-06-25
run credscan and Polichek as its own pipeline ()

p3dr0rv:
The CredScan task is a repo wide assessment, it doesn't need to run as a task under each pipeline within a repo it can run once as its own pipeline...once per each repo.

Update credscan, policheck and publish results to v3

Related PR:
https://github.com/AzureAD/ad-accounts-for-android/pull/1609
https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1617
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1449

Pipeline
https://identitydivision.visualstudio.com/Engineering/_build?definitionScope=%5CAndroid%5CPull%20Request%20Validations%5CCredScan

PR #1462, opened 2021-07-12
Add vsts-release template to help modularize the code

wainaina:
# Summary:
Add a vsts-release template that will aid in the modularization of the vsts release implementation, as currently implemented in broker4j and AADAuthenticator VSTS release pipelines.

Previously, each of these two implementations had their own similar implementation in the broker azure-pipelines section.

## What does this change?
- Improves the modularity of the vsts-release implementation in broker4j / AADAuthenticator and any other module that who's similar functionality would be implemented using this abstraction.
 
## How is the change to be affected?
 - Have the vsts-release abstraction defined in common azure-pipelines and have consuming implementations import it and provide the necessary parameters to start the release process.
 
## How can success be measured?
 - Multiple modules can implement this template with no change to the original template.


azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+# Variable: 'folderName' refers to the project folder
+# Variable: 'gradleTasks' list all the gradle tasks to be used.
+# Variable: 'accessTokenName' refers to the name of the access Token in this project.
+# Variable: 'runUnitTest' boolean: refers to whether the pipeline will run unit tests or not.
+# https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate
+# Variable: 'versionType' type of version (Latest, Snapshot)
+
+parameters:
+- name: versionType
+  default: Latest
```
shoatman, line 14: You may want to say above which variables are required vs optional.  This one is optional since there is a default value selected.

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
+# File: azure-pipelines\templates\steps\nightly-snapshot-releases\vsts-release.yml
+# Description: Template to assemble and publish Snapshot/Latest versions of ${{ parameters.projectName }}
+# Variable: 'ENV_VSTS_MVN_ANDROIDADACCOUNTS_USERNAME' was defined in the Variables tab
+# Variable: 'mvnAccessToken' was defined in the Variables tab
+# Variable: 'folderName' refers to the project folder
+# Variable: 'gradleTasks' list all the gradle tasks to be used.
+# Variable: 'accessTokenName' refers to the name of the access Token in this project.
```
shoatman, line 7: I'd suggest saying that this is the secret used to access the private azure devops maven feed

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    inputs:
+      tasks: ${{ parameters.gradleTasks }}
+      publishJUnitResults: false
+      jdkArchitecture: x86
+      sqAnalysisBreakBuildIfQualityGateFailed: false
+  - task: Gradle@2
+    displayName: Run unit tests
+    condition: eq(variables['parameters.runUnitTests'], 'true')
+    inputs:
+      tasks: testDebugUnitTest
```
shoatman, line 66: Would you want to do this first?  If this fails, then there is no point in creating the release build.

shoatman (approved):
Some nits on documentation and a suggestion on re-order tasks to run unit tests before creating release build.

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+  default: none
+- name: gradleTasks
+  default: none
+- name: accessTokenName
+  default: none
+- name: runUnitTest
+  default: false
+
+jobs:
+- job: Phase_1
```
p3dr0rv, line 30: lets use a more descriptive job name

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+  timeoutInMinutes: 90
+  cancelTimeoutInMinutes: 1
+  pool:
+    name: Hosted Windows 2019 with VS2019
+  steps:
+  - checkout: self
+    clean: true
+    submodules: recursive
+    persistCredentials: True
+  - template: credscan-policheck.yml
```
p3dr0rv, line 41: do not use the credscan-policheck.yml this will be checked separately
check https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1437

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    inputs:
+      SourceFolder: ${{ parameters.projectName }}\build\
+      TargetFolder: $(build.artifactstagingdirectory)
+  - task: PublishPipelineArtifact@1
+    name: PublishPipelineArtifacts1
+    displayName: 'Publish Artifact: ${{ parameters.projectName }} Release'
+    inputs:
+      ArtifactName: ${{ parameters.projectName }}Release
+      TargetPath: $(Build.SourcesDirectory)/${{ parameters.folderName }}/build/
+  - task: ComponentGovernanceComponentDetection@0
```
p3dr0rv, line 84: also we can get rid of this task

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+  - task: Gradle@2
+    displayName: ${{ parameters.versionType }} version update
+    inputs:
+      tasks: ${{ parameters.projectName }}:version${{ parameters.versionType }}
+      publishJUnitResults: false
+  - task: Gradle@1
+    name: Gradle1
+    displayName: Assemble Release
+    inputs:
+      tasks: ${{ parameters.gradleTasks }}
```
p3dr0rv, line 58: you are passing `clean AADAuthenticator:assembleLatest` as parameters.gradleTasks
so versionType is ignored here, also you are passing the project name that we already have

my recommendation it is to change it this way and remove the parameter gradleTasks
```suggestion
      tasks: clean  ${{ parameters.projectName }}:assemble${{ parameters.versionType }}
```

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    displayName: Copy Files to Artifact Staging Directory
+    inputs:
+      SourceFolder: ${{ parameters.projectName }}\build\
+      TargetFolder: $(build.artifactstagingdirectory)
+  - task: PublishPipelineArtifact@1
+    name: PublishPipelineArtifacts1
+    displayName: 'Publish Artifact: ${{ parameters.projectName }} Release'
+    inputs:
+      ArtifactName: ${{ parameters.projectName }}Release
+      TargetPath: $(Build.SourcesDirectory)/${{ parameters.folderName }}/build/
```
p3dr0rv, line 83: this is a personal preference
maybe we can use  `parameters.projectName`  as the folder name and get rid of  `parameters.folderName`

p3dr0rv (changes requested):

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      publishJUnitResults: false
+  - task: CopyFiles@2
+    name: CopyFiles1
+    displayName: Copy Files to Artifact Staging Directory
+    inputs:
+      SourceFolder: ${{ parameters.projectName }}\build\
+      TargetFolder: $(build.artifactstagingdirectory)
+  - task: PublishPipelineArtifact@1
+    name: PublishPipelineArtifacts1
+    displayName: 'Publish Artifact: ${{ parameters.projectName }} Release'
```
p3dr0rv, line 80: Nit:
```suggestion
    displayName: 'Publish Artifact: ${{ parameters.projectName }} ${{ parameters.versionType }}'
```

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      filename: echo
+      arguments: '##vso[task.setvariable variable=${{ parameters.accessTokenName }}]$(mvnAccessToken)'
+  - task: Gradle@2
+    displayName: ${{ parameters.versionType }} version update
+    inputs:
+      tasks: ${{ parameters.projectName }}:version${{ parameters.versionType }}
+      publishJUnitResults: false
+  - task: Gradle@1
+    name: Gradle1
+    displayName: Assemble Release
```
p3dr0rv, line 56:
```suggestion
    displayName: Assemble ${{ parameters.versionType }}
```

azure-pipelines/templates/steps/vsts-release.yml
wainaina: Makes sense.

azure-pipelines/templates/steps/vsts-release.yml
wainaina: The optional variable marker is not yet supported in devops pipelines. Have a look at this thread https://github.com/MicrosoftDocs/azure-devops-docs/issues/7765

azure-pipelines/templates/steps/vsts-release.yml
wainaina: Fair point, however, there is a challenge that comes up. The above suggestion works very well with assemble tasks that have the android gradle plugin e.g. (in AADAuthenticator), when you use this task with the java-library gradle plugin, <project>:assemble task has no variants specifically targeting the flavour.

p3dr0rv: One thing you may want to try is  add  '  ' to parameters.versionType or make it default to empty :)

wainaina: parameters.versionType is a mandatory field. Secondly, I think the specifying the gradleTasks directly is more preferrable as compared to having the tasks be resolved. There are instances one may want to build using different tasks e.g. `<projectName>:build` over `<projectName:>assemble` and also assists on the readability of the caller script.

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,79 @@
...
+
+parameters:
+- name: versionType
+  default: ''
+  values:
+  - Snapshot
+  - Latest
+- name: projectName
+  default: none
+- name: gradleTasks
```
shahzaibj, line 20: I believe this is really `assembleGradleTask`?

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,79 @@
...
+  - task: Gradle@2
+    displayName: Run unit tests
+    condition: eq(variables['parameters.runUnitTests'], 'true')
+    inputs:
+      tasks: testDebugUnitTest
+  - task: Gradle@2
+    displayName: ${{ parameters.versionType }} version update
+    inputs:
+      tasks: ${{ parameters.projectName }}:version${{ parameters.versionType }}
+      publishJUnitResults: false
```
shahzaibj, line 53: Can we put a condition here and only run this if a versionType is provided?

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,79 @@
+# File: azure-pipelines\templates\steps\nightly-snapshot-releases\vsts-release.yml
+# Description: Template to assemble and publish Snapshot/Latest versions of ${{ parameters.projectName }}
+# Variable: 'ENV_VSTS_MVN_ANDROIDADACCOUNTS_USERNAME' was defined in the Variables tab
+# Variable: 'mvnAccessToken' was defined in the Variables tab
+# Variable: 'folderName' refers to the project folder
+# Variable: 'gradleTasks' list all the gradle tasks to be used.
```
shahzaibj, line 6: Not sure how we pass a list here...and what is the usage?

azure-pipelines/templates/steps/vsts-release.yml
wainaina: Sure. Makes sense.

azure-pipelines/templates/steps/vsts-release.yml
wainaina: I pass a string(charsequence) in this case gradleTasks can accept: <projectName>: clean projectName:assemble<VARIANT>. The build tasks could be a bit different in the case where java-libraries assembly differ from android assembly as the android plugin has different assemble tasks for each variant.

azure-pipelines/templates/steps/vsts-release.yml
wainaina: The commands passed here depend on the library invoking this e.g. projectName: clean projectName:assemble<VARIANT>

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,80 @@
...
+- name: gradleTasks
+  default: none
+- name: accessTokenName
+  default: none
+- name: runUnitTest
+  default: false
+
+jobs:
+- job: Phase_1
+  displayName: Initiating Release Process
```
shahzaibj, line 29:
```suggestion
  displayName: Publish to Internal Maven (VSTS)
```

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,80 @@
...
+  timeoutInMinutes: 90
+  cancelTimeoutInMinutes: 1
+  condition: and(ne(variables['parameters.versionType'], ''), ne(variables['parameters.gradleTasks'], ''))
```
shahzaibj, line 32: Looks like now the entire pipeline is going to run based on condition that versionType is specified? that's probably not what we want...I think we should be able to repurpose this pipeline for regular VSTS releases that aren't of the snapshot or latest type...hence I was asking to make the `version update` task conditional (not the entire pipeline)

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,80 @@
...
+    inputs:
+      filename: echo
+      arguments: '##vso[task.setvariable variable=${{ parameters.accessTokenName }}]$(mvnAccessToken)'
+  - task: Gradle@2
+    displayName: Run unit tests
+    condition: eq(variables['parameters.runUnitTests'], 'true')
+    inputs:
+      tasks: testDebugUnitTest
+  - task: Gradle@2
+    displayName: ${{ parameters.versionType }} version update
```
shahzaibj, line 51: I was referring to make this task conditional

shahzaibj: because we have the following types of VSTS releases:

- snapshot
- latest
- (none of the above i.e. a regular VSTS releases where we don't perform a version update task)

I feel like this template can be powerful enough to support that

azure-pipelines/templates/steps/vsts-release.yml
```diff
@@ -0,0 +1,80 @@
+# File: azure-pipelines\templates\steps\nightly-snapshot-releases\vsts-release.yml
+# Description: Template to assemble and publish Snapshot/Latest versions of ${{ parameters.projectName }}
+# Variable: 'ENV_VSTS_MVN_ANDROIDADACCOUNTS_USERNAME' was defined in the Variables tab
+# Variable: 'mvnAccessToken' was defined in the Variables tab
+# Variable: 'folderName' refers to the project folder
```
shahzaibj, line 5: Is this needed anymore?

shahzaibj (approved):
LGTM

p3dr0rv (approved):

PR #1463, opened 2021-07-13
Code changes for new OneAuth cache key migration API

iambmelt:
(Technically this is a major API break)

Reimplements the `ADAL#setSecretKey` migration API exposed for OneAuth.

Functions:
- Allows for behavioral control via `ReencryptionParams`
    * `abortOnFailure`
    * `eraseEntryOnFailure`
    * `eraseAllOnFailure`
- Data will not be written/overwritten unless explicitly overridden to do so
- Returns a new `IMigrationOperationResult` class to surface information about successes/failures

PR #1464, opened 2021-07-14
Code changes for new OneAuth cache key migration API

iambmelt:
Redux of changes in #1463, but this time they're additive.

Reimplements the `ADAL#setSecretKey` migration API exposed for OneAuth.

Changes:
- Deprecates classes formerly used for `setSecretKey` migration
- Allows for behavioral control via `ReencryptionParams`
    * `abortOnFailure`
    * `eraseEntryOnFailure`
    * `eraseAllOnFailure`
- Data will not be written/overwritten unless explicitly overridden to do so
- Returns a new `IMigrationOperationResult` class to surface information about successes/failures

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
+++ b/common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
@@ -1,0 +1,209 @@
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.migration;
+
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.common.internal.cache.ISharedPreferencesFileManager;
+import com.microsoft.identity.common.internal.controllers.TaskCompletedCallback;
+import com.microsoft.identity.common.logging.Logger;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Default implementation of {@link ISharedPrefsFileManagerReencrypter}.
+ */
+public class DefaultSharedPrefsFileManagerReencrypter implements ISharedPrefsFileManagerReencrypter {
+
+    private static final String TAG = DefaultSharedPrefsFileManagerReencrypter.class.getSimpleName();
+    private static final ExecutorService executor = Executors.newSingleThreadExecutor();
+
+    @Override
+    public IMigrationOperationResult reencrypt(@NonNull final ISharedPreferencesFileManager fileManager,
+                                               @NonNull final IStringEncrypter encrypter,
+                                               @NonNull final IStringDecrypter decrypter,
+                                               @NonNull final ReencryptionParams params) {
+        final String methodName = ":reencrypt (sync)";
+        final Map<String, String> cacheEntries = new HashMap<>(fileManager.getAll());
+        Logger.verbose(TAG + methodName,
+                "Attempting to migrate cache entries: " + cacheEntries.size());
+        final MigrationOperationResult result = new MigrationOperationResult();
+        result.setCountOfTotalRecords(cacheEntries.size());
+        final Set<String> keysMarkedForRemoval = new HashSet<>();
+        final AtomicBoolean shouldAbort = new AtomicBoolean(false);
+        final Set<String> skipKeys = new HashSet<>();
+
+        // Decrypt everything
+        applyCacheMutation(
+                cacheEntries,
+                new Callable<Map.Entry<String, String>>() {
+                    @Override
+                    public void call(final Map.Entry<String, String> entry) throws Exception {
+                        final String decryptedText = decrypter.decrypt(entry.getValue());
+                        entry.setValue(decryptedText);
+                    }
+                },
+                result,
+                params,
+                keysMarkedForRemoval,
+                skipKeys,
+                shouldAbort
+        );
+
+        // Clear any keys marked for removal...
+        clearEntriesMarkedForRemoval(fileManager, cacheEntries, keysMarkedForRemoval);
+
+        if (shouldAbort.get()) {
+            Logger.info(TAG + methodName, "Aborting after decrypt.");
+            return result;
+        }
+
+        // Reencrypt everything
+        applyCacheMutation(
+                cacheEntries,
+                new Callable<Map.Entry<String, String>>() {
+                    @Override
+                    public void call(final Map.Entry<String, String> entry) throws Exception {
+                        final String reencryptedText = encrypter.encrypt(entry.getValue());
+                        entry.setValue(reencryptedText);
+                    }
+                },
+                result,
+                params,
+                keysMarkedForRemoval,
+                skipKeys,
+                shouldAbort
+        );
+
+        // Clear any keys marked for removal...
+        clearEntriesMarkedForRemoval(fileManager, cacheEntries, keysMarkedForRemoval);
+
+        if (shouldAbort.get()) {
+            Logger.info(TAG + methodName, "Aborting after reencrypt.");
+            return result;
+        }
+
+        // Write the newly encrypted records...
+        Logger.info(TAG + methodName, "Writing reencrypted cache entries.");
+        for (final Map.Entry<String, String> cacheEntry : cacheEntries.entrySet()) {
+            fileManager.putString(cacheEntry.getKey(), cacheEntry.getValue());
+        }
+
+        return result;
+    }
+
+    private interface Callable<T> {
+        void call(T t) throws Exception;
+    }
+
+    /**
+     * Utility function for reencryption operations. Accepts the contents of a {@link Map} and
+     * applies the mutation defined by {@link Callable} to it.
+     *
+     * @param cacheEntries         The Cache entries to which the supplied mutation is applied.
+     * @param callable             The callable definining the mutation.
+     * @param inputResult          A {@link MigrationOperationResult} used to track error states which may
+     *                             occur during a mutation.
+     * @param params               The {@link com.microsoft.identity.common.migration.ISharedPrefsFileManagerReencrypter.ReencryptionParams}
+     *                             defining how errors should be handled/surfaced.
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,186 @@
...
+                skipKeys.add(cacheEntry.getKey());
+
+                if (params.eraseEntryOnError()) {
+                    Logger.warn(TAG + methodName, "Marking key for removal.");
+                    keysMarkedForRemoval.add(cacheEntry.getKey());
+                }
+
+                if (params.eraseAllOnError()) {
+                    Logger.warn(TAG + methodName, "Marking all keys for removal.");
+                    keysMarkedForRemoval.addAll(cacheEntries.keySet());
```
AdamBJohnsonx, line 151: I mean, you could return here, right?

iambmelt: Yes (and tests continue to pass)

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,188 @@
...
+                shouldAbort
+        );
+
+        // Clear any keys marked for removal...
+        clearEntriesMarkedForRemoval(fileManager, cacheEntries, keysMarkedForRemoval);
+
+        if (shouldAbort.get()) {
+            Logger.info(TAG + methodName, "Aborting after decrypt.");
+            return result;
+        }
```
AdamBJohnsonx, line 84: So if this is remove all on error, this will remove everything from the map and the Shared Prefs file, then not abort, but write nothing (since there will be nothing in cacheEntries and nothing will happen).  What's the difference between these options?

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
iambmelt: I pushed an update based on your [previous comment](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1464#discussion_r669195712); I believe this is now resolved

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
```diff
@@ -0,0 +1,50 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.migration;
+
+import java.util.Map;
+
+public interface IMigrationOperationResult {
```
shahzaibj, line 27: javadoc?

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
```diff
@@ -0,0 +1,50 @@
...
+     * @return The total number of records.
+     */
+    int getTotalRecords();
+
+    /**
+     * Gets the number of failed records.
+     *
+     * @return The number of failed records.
+     */
+    int getFailedRecords();
```
shahzaibj, line 41: I think we need the method to indicate that this does NOT return the failed records, but rather the number of failed records....not sure what is the best way to do that

```suggestion
    int getNumberOfFailedRecords();
```

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Gets an {@link Map} of Exception types + message -> total count of errors.
```
shahzaibj, line 44: Seems like Exception type + message is the key....can we put an example in the javadoc of what this key might look like?

common/src/main/java/com/microsoft/identity/common/migration/ISharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,143 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.migration;
+
+import com.microsoft.identity.common.internal.cache.ISharedPreferencesFileManager;
+import com.microsoft.identity.common.internal.controllers.TaskCompletedCallback;
+
+/**
+ * Interface describing an object that can reencrypt instances of
```
shahzaibj, line 29: Re-encryption is essentially encryption? or something different? I mean if this interface can reencrypt instances of ISharedPreferencesFileManager...can it also `encrypt` instances of ISharedPreferencesFileManager?

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,188 @@
...
+
+        return result;
+    }
+
+    private interface Callable<T> {
+        void call(T t) throws Exception;
+    }
+
+    private void applyCacheMutation(@NonNull final Map<String, String> cacheEntries,
```
shahzaibj, line 124: Even though this is private, I think javadoc would be nice, primarily is because there are a lot of params and it seems to be doing a lot of things

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,188 @@
...
+                                    @NonNull final Callable<Map.Entry<String, String>> callable,
+                                    @NonNull final MigrationOperationResult inputResult,
```
shahzaibj, line 126: Not sure why it needs to accept result as input...can't we just create it within this method and return it?

iambmelt: The same result is mutated across multiple executions of this method: 1st time for decryption, 2nd time for reencryption. It is passed in to update the existing object with any additional exceptions that need to be added to the error.

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
iambmelt: ACK

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,188 @@
...
+        final String methodName = ":applyCacheMutation";
+        for (final Map.Entry<String, String> cacheEntry : cacheEntries.entrySet()) {
+            try {
+                if (skipKeys.contains(cacheEntry.getKey())) {
+                    Logger.warn(TAG + methodName, "Skipping entry.");
+                    continue;
+                }
+                callable.call(cacheEntry);
+            } catch (final Exception e) {
+                Logger.error(TAG + methodName, "Error during mutation", e);
```
shahzaibj, line 140: Do we need to log the key that caused the error? I guess it could be pii/oii?

common/src/main/java/com/microsoft/identity/common/migration/ISharedPrefsFileManagerReencrypter.java
iambmelt: Reencryption here means that data encrypted with KeyA will be decrypted and reencrypted with KeyB. This class is narrowly focused on rotating the encryption key that was used to write the cache's records

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
iambmelt: ACK

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
iambmelt: It will be both PII and OII - I can look at logging it via the PII/OII safe methods

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
shahzaibj: So if I understand correctly, you need the end results to contain all the errors..which is essentially that some keys would run into errors during decryption and some would get errors during re-encryption? and you want the result to contain these all?

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
iambmelt: To keep this work easy to `cherry-pick` for @AdamBJohnsonx - I have applied your feedback in a single amended commit via `32fdc9cdf4a9357bd40cca2e0e08f6daf1d50157`

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
iambmelt: To keep this work easy to `cherry-pick` for @AdamBJohnsonx - I have applied your feedback in a single amended commit via `32fdc9cdf4a9357bd40cca2e0e08f6daf1d50157`

common/src/main/java/com/microsoft/identity/common/migration/IMigrationOperationResult.java
iambmelt: To keep this work easy to `cherry-pick` for @AdamBJohnsonx - I have applied your feedback in a single amended commit via `32fdc9cdf4a9357bd40cca2e0e08f6daf1d50157`

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
iambmelt: To keep this work easy to `cherry-pick` for @AdamBJohnsonx - I have applied your feedback in a single amended commit via `32fdc9cdf4a9357bd40cca2e0e08f6daf1d50157`

common/src/main/java/com/microsoft/identity/common/migration/DefaultSharedPrefsFileManagerReencrypter.java
iambmelt: To keep this work easy to `cherry-pick` for @AdamBJohnsonx - I have applied your feedback in a single amended commit via `32fdc9cdf4a9357bd40cca2e0e08f6daf1d50157`

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/migration/ISharedPrefsFileManagerReencrypter.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Performs reencryption of the provided {@link ISharedPreferencesFileManager}, delegating to
+     * the suppplied {@link IStringEncrypter} and {@link IStringDecrypter} to perform content
+     * transformations.
+     * <p>
+     * Please note: this method does not lock the underlying store during reencryption. Users of
+     * this API are advised to ensure the designated store is not mutated during the reencryption
+     * process otherwise undefined behavior/results may occur.
```
iamgusain, line 109: nit:
Should this be specified in the methodName..
reencryptWithoutLock?

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/migration/ISharedPrefsFileManagerReencrypter.java
iambmelt: This is a good suggestion to be explicit - I will decline it for now such that this PR remains a single cherry-pickable commit and due to the fact that this method is _essentially_ private API as it will be called internally from OneAuth. The fact that this storage cannot be locked will be captured in two places that hopefully provide adequate warning/context for those executing this code: first, here via javadoc, and again via the OneAuth migration documentation that will be updated to assist developers transitioning from ADAL -> OneAuth

PR #1470, opened 2021-07-15
Add method to check if preferred_network_host_name matches for two authorities

iamgusain:
Adding a method to compare if an instance of AzureActiveDirectoryAuthority has same preferred_network_host_name as another AzureActiveDirectoryAuthority instance.

Tests:
Added Unit tests to verify failure and success case for the new method

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
@@ -186,7 +186,7 @@ public class AzureActiveDirectoryAuthority extends Authority {
     }
 
     /**
-     * Checks if current authority cloud has same preferred network host name as the passed in authority cloud
+     * Checks if current authority cloud has same preferred network host name as the passed in authority cloud.
      *
      * @param authorityToCheck authority to check against
      * @return true if preferred network host name matches for both authorities, otherwise false
--- a/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
@@ -25,19 +25,21 @@ package com.microsoft.identity.common.internal.authorities;
 import android.net.Uri;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.WorkerThread;
 
 import com.google.gson.annotations.SerializedName;
-import com.microsoft.identity.common.java.WarningType;
-import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectory;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectoryCloud;
 import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Configuration;
 import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Strategy;
 import com.microsoft.identity.common.internal.providers.oauth2.OAuth2Strategy;
 import com.microsoft.identity.common.internal.providers.oauth2.OAuth2StrategyParameters;
+import com.microsoft.identity.common.java.WarningType;
+import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectorySlice;
 import com.microsoft.identity.common.logging.Logger;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -183,4 +185,21 @@ public AzureActiveDirectoryAudience getAudience() {
...
 
+    /**
+     * Checks if current authority cloud has same preferred network host name as the passed in authority cloud
+     *
+     * @param authorityToCheck authority to check against
+     * @return true if preferred network host name matches for both authorities, otherwise false
+     */
+    @WorkerThread
+    public synchronized boolean isSamePreferredNetworkHostAsAuthority(@NonNull final AzureActiveDirectoryAuthority authorityToCheck) throws IOException {
+        if (!AzureActiveDirectory.isInitialized()) {
```
AdamBJohnsonx, line 196: I don't love this API, but sure.  We have two function calls where one might do.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -56,7 +58,7 @@
 
     private AzureActiveDirectoryCloud mAzureActiveDirectoryCloud;
```
AdamBJohnsonx, line 59: This can't be final because we're occasionally looking for updates...

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
AdamBJohnsonx: ... so this can change, however unlikely.  We should be open to it being stale and needing a refresh.

AdamBJohnsonx (approved):
I don't see anything being made worse here.

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -183,4 +185,21 @@ public AzureActiveDirectoryAudience getAudience() {
...
+     * @return true if preferred network host name matches for both authorities, otherwise false
+     */
+    @WorkerThread
+    public synchronized boolean isSamePreferredNetworkHostAsAuthority(@NonNull final AzureActiveDirectoryAuthority authorityToCheck) throws IOException {
+        if (!AzureActiveDirectory.isInitialized()) {
+            // Cloud discovery is needed in order to make sure that we have a preferred_network_host_name to cloud aliases mappings
+            AzureActiveDirectory.performCloudDiscovery();
+        }
+
+        final String preferredNetworkHostName = getAzureActiveDirectoryCloud().getPreferredNetworkHostName();
```
shahzaibj, line 201: I guess you could've just used the member variable directly

common/src/test/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthorityTests.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(RobolectricTestRunner.class)
+public class AzureActiveDirectoryAuthorityTests {
+    @Test
+    public void isSamePreferredNetworkHostAsAuthority_Returns_True_For_Authority_With_ValidAliases_For_SameCloud() throws IOException {
+        String[] cloudAliasesWW = new String[]{"https://login.microsoftonline.com", "https://login.windows.net", "https://login.microsoft.com", "https://sts.windows.net"};
+        String[] cloudAliasesCN = new String[]{"https://login.chinacloudapi.cn", "https://login.partner.microsoftonline.cn"};
+        String[] cloudAliasesUSGov = new String[]{"https://login.microsoftonline.us", "https://login.usgovcloudapi.net"};
```
shahzaibj, line 40: nit: final

common/src/main/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -183,4 +185,21 @@ public AzureActiveDirectoryAudience getAudience() {
         return mAudience;
     }
 
+    /**
+     * Checks if current authority cloud has same preferred network host name as the passed in authority cloud
```
shahzaibj, line 189:
```suggestion
     * Checks if current authority cloud has same preferred network host name as the passed in authority cloud.
```

common/src/test/java/com/microsoft/identity/common/internal/authorities/AzureActiveDirectoryAuthorityTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+        AzureActiveDirectoryAuthority authorityWW = new AzureActiveDirectoryAuthority(new AllAccounts(cloudAliasesWW[0]));
```
shahzaibj, line 42: nit: final here and everywhere

shahzaibj (approved):

PR #1471, opened 2021-07-16
Update method to compare if two authorities belong to same cloud

iamgusain:
Renamed method "isSamePreferredNetworkHostAsAuthority" to "isSameCloudAsAuthority" in AzureActiveDirectoryAuthority.
Added equals method in AzureActiveDirectoryCloud class to compare two instances based on PreferredNetworkHostName and PreferredCacheHostName.
Updated the tests to reflect the naming change.

isSameCloudAsAuthority will be used to check if two authorities belong to same cloud or not. This is useful  in cross cloud scenarios where we want to make sure that we only use PRT for home authority, and if a token requests comes for foreign authority we don't use PRT and trigger normal cross cloud auth flow.

Related PR: #1470

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryCloud.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryCloud.java
@@ -22,12 +22,13 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory;
 
-import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import com.google.gson.annotations.SerializedName;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * This class contains information about a specific Azure Active Directory Cloud.  Azure Active Directory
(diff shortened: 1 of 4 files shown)


shoatman (approved):
Thanks Amit!

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/microsoft/azureactivedirectory/AzureActiveDirectoryCloud.java
```diff
@@ -111,4 +114,25 @@ void setIsValidated(final boolean isValidated) {
...
+        }
+
+        return  mPreferredNetworkHostName.equals(((AzureActiveDirectoryCloud) cloudObjectToCheck).mPreferredNetworkHostName) &&
+                mPreferredCacheHostName.equals(((AzureActiveDirectoryCloud) cloudObjectToCheck).mPreferredCacheHostName);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(mPreferredNetworkHostName, mPreferredCacheHostName);
+    }
```
shahzaibj, line 137: Can use Lombok's equalAndHashCode that allows to include and exclude specific fields from equals...any reason not to use it?

iamgusain: Updated to use lombok.

shahzaibj (approved):

PR #1475, opened 2021-07-20
Migrate Strategies part 1/2

rpdome:
Here comes (what I think is) one of the most convoluted part of our code base.
Once I finish part 2, we should be able to start integrating Linux WebView :)

## Changes

1. Migrate from `LocalBroadcastManager` to our own `LocalBroadcaster`.
2. Replace Intent/Bundle with `DataBag` (a class with maps)
3. Draw a clear line between "Authorization Request" and "Broker Interactive session" - to improve readability.
   -  The result returned from broker will have a separate set of return code (AuthenticationConstants.BrokerResponse).
   - The result returned from broker will has its own broadcast alias (RETURN_BROKER_INTERACTIVE_ACQUIRE_TOKEN_RESULT)
   - Due to above, the returned result will no longer take a long trip from BrokerActivity -> BroadcastManager -> CommandDispatcher -> InteractiveTokenCommand -> BrokerMsalController. We will register/unregister the broadcast event in `BrokerMsalController.acquireToken()` directly. 
   - Renamed `BaseController.notify()` to `BaseController.onFinishAuthorizationSession()`, because that's what it really does.
   - Renamed `BaseController.completeAcquireToken()` to `BaseController.onFinishAuthorizationSession()`, because Authorization is a subset of acquireToken.
   - Created a PBI for a long-term change https://identitydivision.visualstudio.com/Engineering/_backlogs/backlog/Auth%20Client%20-%20CPP%20Linux/Epics/?workitem=1470135
4. Consolidate the shared logic in `MicrosoftStsAuthorizationResultFactory` and `AzureActiveDirectoryAuthorizationResultFactory` into the base class `AuthorizationResultFactory`
5. (Try to) consolidate the 'redirect URL interpreter' logic from `BrowserAuthorizationFragment` and `AzureActiveDirectoryWebViewClient` into `RawAuthorizationResult`.
   - The idea is... you pass in the result URL in the factory method, we'll intepret it.
   - This will also abstract out the bundle/dataBag from those classes - you don't need to know how the data is packed/sent throughout the code, we'll do that for you.
   - eSTS error in redirect uri is now handled in only one place - `AzureActiveDirectoryWebViewClient`. The current logic is associating those error with the error code that should be used for client-side error (this logic is not consistent in all path).
6. Migrate multiple classes under common/internal/providers

## Other smaller changes
1. Migrate IDevicePopManager to common4j.
   - DevicePopManager will take context in via its constructor, as the method interface is now java-only.
2. Add IDevicePopManager and ClockSkewManager to ICommonComponents.
3. Delete android-common's `Device` class, along with its global IDevicePopManager.
4. OAuth2StrategyParameters now takes ICommonComponents instead of Context.
5. Turns `AuthorizationStrategy` into `AndroidAuthorizationStrategy` and `IAuthorizationStrategy` (because we'll need one for Linux).
6. In `AuthorizationStrategy`, always use the state value from the request. 
   - Previously, one of the subclasses is using it from the request, while the other is looking for it in the bundle - and the value in the bundle is never used.
7. Consolidate duplicated logic in `CurrentTaskBrowserAuthorizationStrategy` and `BrowserAuthorizationStrategy` 
   - by converting `BrowserAuthorizationStrategy` into the shared base class, and have `CurrentTaskBrowserAuthorizationStrategy` and `DefaultBrowserAuthorizationStrategy` inherit from it. 
8. Add/update/migrate test cases. esp. around `RawAuthorizationResult` and `AuthorizationResultFactories`
9. Turns `UIResponse` into an enum in `RawAuthorizationResult`


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -163,7 +164,7 @@ public void onPageCommitVisible() {
             public void run() {
                 Logger.info(TAG + methodName, "Launching embedded WebView for acquiring auth code.");
                 Logger.infoPII(TAG + methodName, "The start url is " + mAuthorizationRequestUrl);
-                mWebView.loadUrl(mAuthorizationRequestUrl, mRequestHeaders);
+                mWebView.loadUrl("https://login.microsoftonline.com/organizations/oAuth2/v2.0/authorize?prompt=select_account&client-request-id=6abae31d-3a08-4b95-8299-7d7ace789a59&x-client-CPU=arm64-v8a&x-client-DM=Pixel+XL&x-client-OS=29&x-client-SKU=MSAL.Android&x-client-Ver=1.5.9&login_hint=&instance_aware=false&code_challenge=iGjnawmjUbXXZHjybOkopPBiwOa30ovHH6WX9Cv7ZAE&code_challenge_method=S256&claims=%7B%7D&client_id=4b0db8c2-9f26-4417-8bde-3f0e3656f8e0&redirect_uri=msauth%3A%2F%2Fcom.msft.identity.client.sample.local%2F1wIqXSqBj7w%252Bh11ZifsnqwgyKrY%253D&scope=user.read++openid+offline_access+profile&state=1254%3A0ea28b21-8a94-4e5d-a0ff-3a5d3cc07720-9536ce32-aabc-4b0e-98f8-e55cf00e7cfe", mRequestHeaders);
```
shahzaibj, line 167: Is this intended?

common/src/main/java/com/microsoft/identity/common/internal/ui/Changes
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+Other smaller changes
+1. Migrate IDevicePopManager to common4j.
+   - DevicePopManager will take context in via its constructor, as the method interface is now java-only.
+2. Add IDevicePopManager and ClockSkewManager to ICommonComponents.
+3. Delete android-common's `Device` class, along with its global IDevicePopManager.
+4. OAuth2StrategyParameters now takes ICommonComponents instead of Context.
+5. Turns `AuthorizationStrategy` into `AndroidAuthorizationStrategy` and `IAuthorizationStrategy` (because we'll need one for Linux).
```
shahzaibj, line 25: Is this meant to go under a `docs` folder? and should this be in markdown?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
rpdome: nope. i was testing. good catch.

common/src/main/java/com/microsoft/identity/common/internal/ui/Changes
rpdome: i'll remove this.

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
```diff
@@ -0,0 +1,100 @@
...
+
+import com.microsoft.identity.common.java.interfaces.INameValueStorage;
+import com.microsoft.identity.common.java.util.ported.DataBag;
+
+import java.io.Serializable;
+import java.util.Map;
+
+import lombok.NonNull;
+
+public class DataBagUtil {
```
shahzaibj, line 35: Can we put some javadoc here?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -242,6 +251,40 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
         final Intent brokerActivityIntent = new Intent(parameters.getAndroidApplicationContext(), BrokerActivity.class);
         brokerActivityIntent.putExtra(BrokerActivity.BROKER_INTENT, interactiveRequestIntent);
 
+        LocalBroadcaster.INSTANCE.registerCallback(RETURN_BROKER_INTERACTIVE_ACQUIRE_TOKEN_RESULT,
+                new LocalBroadcaster.IReceiverCallback() {
+            @Override
+            public void onReceive(@NonNull DataBag dataBag) {
+                /**
+                 * Get the response from the Broker captured by BrokerActivity.
+                 * BrokerActivity will pass along the response to the broker controller
```
shahzaibj, line 260:
```suggestion
                 * BrokerActivity will pass along the response to the broker controller.
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -242,6 +251,40 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
...
+                 * The Broker controller will map th response into the broker result
```
shahzaibj, line 261:
```suggestion
                 * The Broker controller will map the response into the broker result,
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -242,6 +251,40 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
...
+                 * And signal the future with the broker result to unblock the request.
```
shahzaibj, line 262:
```suggestion
                 * and signal the future with the broker result to unblock the request.
```

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -67,7 +68,8 @@
     private static final String TEST_REDIRECT_URL = "ABC12/xyz";
     private static final String TEST_WEBSITE_REQUEST_URL = "browser://abcxyz/a";
     private static final String TEST_BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "browser://abcxyz/xyz&ismdmurl=1";
-    private static final String TEST_INSTALL_REQUEST_URL = "msauth://com.msft.identity.client.sample.local/1wIqXSqBj7w%2Bh11Z";
+    private static final String TEST_INSTALL_REQUEST_URL = "msauth://wpj/?username=someusername%somedomain.onmicrosoft.com&app_link=https%3a%2f%2fplay.google.com%2fstore%2fapps%2fdetails%3fid%3dcom.azure.authenticator%26referrer%3dcom.msft.identity.client.sample.local";
```
shahzaibj, line 71: Why is this changing?

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -67,7 +68,8 @@
...
+    private static final String TEST_DEVICE_REGISTRATION_URL = "msauth://wpj/?username=someusername%somedomain.onmicrosoft.com";
```
shahzaibj, line 72: What is this for?

common4j/src/main/com/microsoft/identity/common/java/interfaces/ICommonComponents.java
```diff
@@ -40,5 +44,24 @@
...
     IKeyAccessor getStorageEncryptionManager(@Nullable final ITelemetryCallback telemetryCallback);
+
+    /**
+     * Gets clock skew manager.
+     */
+    @NonNull
+    IClockSkewManager getClockSkewManager();
+
+    /**
+     * @throws ClientException if it fails to initalize, or if the operation is not supported by the platform.
```
shahzaibj, line 57: Missing method description

common4j/src/main/com/microsoft/identity/common/java/interfaces/ICommonComponents.java
```diff
@@ -40,5 +44,24 @@
...
+     * */
+    @NonNull
+    IDevicePopManager getDefaultDevicePopManager() throws ClientException;
+
+    /**
```
shahzaibj, line 62: Missing method description

common4j/src/main/com/microsoft/identity/common/java/interfaces/INameValueStorage.java
```diff
@@ -55,4 +68,10 @@
      * Clear all data from the storage.
      */
     void clear();
+
+    /**
+     * Get all keys in this storage.
```
shahzaibj, line 73: Interesting...why do we need this? Why not have a getAll method instead that returns values?

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
rpdome: the previous one is not valid.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
rpdome: This is what we got once you click the "Register" button.

rpdome: (The other one is what you got "if the broker is not installed")

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryAuthorizationResultFactory.java
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+            Logger.info(TAG, correlationId, "Auth code is successfully returned from webview redirect.");
+            AzureActiveDirectoryAuthorizationResponse response = new AzureActiveDirectoryAuthorizationResponse(code, state);
+            response.setCorrelationId(correlationId);
+            result = new AzureActiveDirectoryAuthorizationResult(AuthorizationStatus.SUCCESS, response);
+        }
+
+        return result;
+    }
+}
```
shahzaibj, line 131: nit: new line ending

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResultFactory.java
```diff
@@ -0,0 +1,155 @@
...
+            result = new MicrosoftStsAuthorizationResult(
+                    AuthorizationStatus.SUCCESS,
+                    authResponse
+            );
+        }
+
+        return result;
+    }
+
+}
```
shahzaibj, line 155: nit: new line ending

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationResultFactory.java
```diff
@@ -0,0 +1,148 @@
...
+                    Logger.warn(TAG, null, "returned URL is null or empty.");
+                    return createAuthorizationResultWithErrorResponse(
+                            AuthorizationStatus.FAIL,
+                            MicrosoftAuthorizationErrorResponse.AUTHORIZATION_FAILED,
+                            MicrosoftAuthorizationErrorResponse.AUTHORIZATION_SERVER_INVALID_RESPONSE);
+                }
+                return parseUrlAndCreateAuthorizationResult(url, request.getState());
+
+            case NON_OAUTH_ERROR:
+                // This is purely client side error, possible return could be chrome_not_installed or the request intent is
```
shahzaibj, line 90: chrome_not_installed I believe is for browser case? What about webview scenario?

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationStatus.java
```diff
@@ -32,12 +32,13 @@
...
     /**
-     * User press device back button.
+     * User press device back button,
+     * or the authz session was cancelled in order to proceed with the flow (not an error).
      */
     USER_CANCEL,
 
     /**
-     * Sdk cancelled Auth floe
+     * Sdk cancelled Auth flow
```
shahzaibj, line 41:
```suggestion
     * Sdk cancelled Auth flow.
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -45,6 +60,138 @@ public static boolean isNullOrEmpty(String message) {
...
+        String utid = null;
+
+        if (EXPECTED_LENGTH == uidUtidArray.length
+                && !isNullOrEmpty(uidUtidArray[INDEX_UID])
+                && !isNullOrEmpty(uidUtidArray[INDEX_UTID])) {
+            uid = uidUtidArray[INDEX_UID];
+            utid = uidUtidArray[INDEX_UTID];
+        }
+
+        return new AbstractMap.SimpleEntry<>(uid, utid);
```
shahzaibj, line 113: Ok, there's no pair being returned here...and I'm not sure I like it being a `Map.Entry` either..I mean I believe Map.Entry is meant to be a key and value...not sure if `uid` is a valid key for `utid`....isntead of doing this..can we just create a dedicated class for TenantInfo?

```java
class TenantInfo {
   private final String uid;
   private final String utid;
}
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -45,6 +60,138 @@ public static boolean isNullOrEmpty(String message) {
...
+     *
+     * @param items     String
+     * @param delimiter String
+     * @return List<String>
+     */
+    public static List<String> getStringTokens(final String items, final String delimiter) {
+        final StringTokenizer st = new StringTokenizer(items, delimiter);
+        final List<String> itemList = new ArrayList<>();
+        while (st.hasMoreTokens()) {
+            String name = st.nextToken();
```
shahzaibj, line 139:
```suggestion
            final String name = st.nextToken();
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -45,6 +60,138 @@ public static boolean isNullOrEmpty(String message) {
...
+
+    /**
+     * Split the input string into a list of string tokens.
+     *
+     * @param items     String
+     * @param delimiter String
+     * @return List<String>
+     */
+    public static List<String> getStringTokens(final String items, final String delimiter) {
+        final StringTokenizer st = new StringTokenizer(items, delimiter);
```
shahzaibj, line 136:
```suggestion
        final StringTokenizer tokenizer = new StringTokenizer(items, delimiter);
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -45,6 +60,138 @@ public static boolean isNullOrEmpty(String message) {
...
+
+        return itemList;
+    }
+
+    /**
+     * Split the input with delimiter.
+     *
+     * @param input     String
+     * @param delimiter char
+     * @return ArrayList<String>
```
shahzaibj, line 153: Can we add description for these?

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
```diff
@@ -56,16 +66,146 @@ public static String appendPathToURL(@NonNull final URL urlToAppend,
...
         }
 
         builder.setPathSegments(combinedPathSegments);
-        return builder.build().toString();
+        return builder.build().toURL();
     }
+
+
+    /**
+     * create url from given endpoint. return null if format is not right.
```
shahzaibj, line 83:
```suggestion
     * Create url from given endpoint. return null if format is not right.
```

shahzaibj: Do we just want to throw the exception instead?

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
```diff
@@ -56,16 +66,146 @@ public static String appendPathToURL(@NonNull final URL urlToAppend,
...
+
+            return urlFormDecode(response.getQuery());
+        } catch (final URISyntaxException e) {
+            Logger.warn(TAG + methodName, "Failed to extract URL parameters");
+            Logger.errorPII(
+                    TAG + methodName,
+                    "Url is invalid",
+                    e
+            );
+            return Collections.emptyMap();
```
shahzaibj, line 137: Do we just want to throw the exception from here? I guess the question is that whether the entire token request should fail when this happens? What is the caller supposed to be with the empty map? Will they be able to proceed or are they supposed to throw their own exception if/once they realize the got back an empty map?

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
```diff
@@ -56,16 +66,146 @@ public static String appendPathToURL(@NonNull final URL urlToAppend,
...
+        }
+    }
+
+    /**
+     * decode url string into a key value pairs with default query delimiter.
```
shahzaibj, line 142:
```suggestion
     * Decode url string into a key value pairs with default query delimiter.
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: Many of the code are taken/migrated as-is, and i'm refraining myself from modifying them (unless i think it really should be improved right away) - otherwise the scope of change will keep growing (and more test will be required).

LMK if i definitely should pick this up.

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
```diff
@@ -56,16 +66,146 @@ public static String appendPathToURL(@NonNull final URL urlToAppend,
...
+     * @param urlParameter URL query parameter
+     * @return key value pairs
+     */
+    @NonNull
+    public static Map<String, String> urlFormDecode(@NonNull final String urlParameter) {
+        return urlFormDecodeData(urlParameter, "&");
+    }
+
+    /**
+     * decode url string into a key value pairs with given query delimiter given
```
shahzaibj, line 153:
```suggestion
     * Decode url string into a key value pairs with given query delimiter given
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
shahzaibj: Let's leave it as is for now then

common4j/src/main/com/microsoft/identity/common/java/util/ported/DataBag.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Getter
+@Accessors(prefix = "m")
+public class DataBag {
+    private final INameValueStorage<Boolean> mBooleanMap = new MapWithDefault<>();
+    private final INameValueStorage<byte[]> mByteArrayMap = new MapWithDefault<>();
+    private final INameValueStorage<String> mStringMap = new MapWithDefault<>();
+    private final INameValueStorage<Integer> mIntMap = new MapWithDefault<>();
+    private final INameValueStorage<Serializable> mSerializableMap = new MapWithDefault<>();
+}
```
shahzaibj, line 40: Can we put the following utility methods on this class?

```java
void put(String key, Boolean value) {
   mBooleanMap.put(key, value);
}

void put(String key, byte[] value) {
   mByteArrayMap.put(key, value);
}

// do the same for all the other maps as well

// also provider a getter as follows:
Object get(String key) {
 // does a look-up against each of these and returns the right value
}
```

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
```diff
@@ -0,0 +1,100 @@
...
+                result.getBooleanMap().put(key, (Boolean) value);
+            } else if (value instanceof Integer) {
+                result.getIntMap().put(key, (Integer) value);
+            } else if (value instanceof Serializable) {
+                result.getSerializableMap().put(key, (Serializable) value);
+            }
+        }
+
+        return result;
+    }
```
shahzaibj, line 99: See comment here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1475/files#r674201550

If we do that then the code here becomes a lot simple

common4j/src/main/com/microsoft/identity/common/java/util/ported/DataBag.java
shahzaibj: Also see comment here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1475/files#r674202035

shahzaibj: Another thought is that we may able to simplify this even further by having just one map for byte[]

```java
private final INameValueStorage<byte[]> mByteArrayMap = new MapWithDefault<>();
```

This is because all the types here are Serializable anyway...so we should be able to convert them to bytes

common4j/src/test/com/microsoft/identity/common/java/providers/Constants.java
```diff
@@ -0,0 +1,51 @@
+package com.microsoft.identity.common.java.providers;
```
shahzaibj, line 1: LICENSE

common4j/src/test/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationResultFactoryTest.java
```diff
@@ -64,14 +64,78 @@ public void setUp() {
...
+        return new MicrosoftStsAuthorizationRequest.Builder().setState(MOCK_STATE).build();
+    }
+
+    private MicrosoftStsAuthorizationRequest getMstsAuthorizationRequestWithState(@NonNull final String state) {
+        return new MicrosoftStsAuthorizationRequest.Builder().setState(state).build();
+    }
+
+    @Test
+    public void testClientError() {
+        AuthorizationResult result = mAuthorizationResultFactory.createAuthorizationResult(
```
shahzaibj, line 76: nit: `final` here and everywhere when applicable

common4j/src/main/com/microsoft/identity/common/java/crypto/AsymmetricAlgorithm.java
```diff
@@ -20,36 +20,28 @@
...
-        mMap.put(name, value);
-    }
-
-    @Override
-    public void remove(@NonNull String name) {
-        mMap.remove(name);
-    }
-
+/**
+ * Marker interface for asymmetric algorithms.
```
shahzaibj, line 28: What does Marker mean here?

common4j/src/main/com/microsoft/identity/common/java/crypto/IDevicePopManager.java
```diff
@@ -95,11 +88,11 @@
      * requires use of a SHA-1 digest or uses NO_PADDING should not be supported.
      */
     enum Cipher implements AsymmetricAlgorithm {
-        @RequiresApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
+        //@RequiresApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
```
shahzaibj, line 91: Not needed? Should the line be just removed?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsPromptBehavior.java
```diff
@@ -1,4 +1,4 @@
-package com.microsoft.identity.common.internal.providers.microsoft.microsoftsts;
+package com.microsoft.identity.common.java.providers.microsoft.microsoftsts;
```
shahzaibj, line 1: Missing LICENSE

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationErrorResponse.java
```diff
@@ -24,16 +24,35 @@
...
 /**
  * Class to encapsulate Authorization Result errors.
  */
+@Getter
+@Setter
+@Accessors(prefix = "m")
 public class AuthorizationErrorResponse implements IErrorResponse{
 
+    /**
+     * error returned due to Authorization request failure.
```
shahzaibj, line 40: Capitalize first letter in javadoc....here and everywhere

shahzaibj (commented):
Left comments...I think my primary comments are around simplifying usage of the DataBag class...I think we can make the class itself a little more sophisticated to simplify usage for its consumers.

common4j/src/main/com/microsoft/identity/common/java/crypto/IDevicePopManager.java
rpdome: I'm keeping this for reference. It's still useful.. even if it won't show up warning.

(RequiresApi cannot be used in java project)

common4j/src/main/com/microsoft/identity/common/java/util/ported/DataBag.java
rpdome: I could create those helper function (that's actually my original design - it's just that it make this classes cluttered, since we'll need get() put() and getOrDefault() for each of them, 3*5 = 15 methods in this class)

That said, with one map, it's better to clutter thing here than in DataBagUtil, so i'll make the change.

rpdome: (what you suggested is actually similar to how Bundle is implemented)

rpdome: actually i can just use generic getters.

common4j/src/main/com/microsoft/identity/common/java/crypto/AsymmetricAlgorithm.java
rpdome: I don't know lol :D

AdamBJohnsonx: A marker interface generally has no methods.  I see this one evolved since I created it, and the javadoc should probably be corrected.  Also, there's no need to override interface methods when extending...

common4j/src/main/com/microsoft/identity/common/java/crypto/AsymmetricAlgorithm.java
```diff
@@ -20,36 +20,28 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
```
AdamBJohnsonx, line 20: This file has a **WEIRD** diff.  I'm not sure that it won't cause problems down the line.

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
rpdome: Change the method such that it takes in URI only. this exception will be thrown somewhere else.

common4j/src/main/com/microsoft/identity/common/java/util/UrlUtil.java
rpdome: Removed this unused func.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationResultFactory.java
rpdome: as of now, the only place that seems to throw this is the PKeyAuthChallengeFactory.

common4j/src/main/com/microsoft/identity/common/java/interfaces/INameValueStorage.java
rpdome: to iterate and dump the data as-is to the databag/bundle.

common4j/src/main/com/microsoft/identity/common/java/crypto/AsymmetricAlgorithm.java
rpdome: I killed this class... somehow git thought it's close enough to AsymmetricAlgorithm.... probably because of the license?

common4j/src/main/com/microsoft/identity/common/java/util/ported/DataBag.java
```diff
@@ -0,0 +1,76 @@
...
+import lombok.NonNull;
+import lombok.experimental.Accessors;
+
+@Accessors(prefix = "m")
+public class DataBag {
+    private final INameValueStorage<Object> mMap = new InMemoryStorage<>();
+
+    public void put(@NonNull final String name, @Nullable final Object value) {
+        mMap.put(name, value);
+    }
```
shahzaibj, line 41: Should we restrict these to `Serializable` as well?

common4j/src/main/com/microsoft/identity/common/java/util/ported/DataBag.java
```diff
@@ -0,0 +1,76 @@
...
+        if (object == null) {
+            return defaultValue;
+        }
+
+        try {
+            return (T) object;
+        } catch (final ClassCastException e) {
+            return defaultValue;
+        }
+    }
```
shahzaibj, line 71: Seems like we have an opportunity to share code here

```suggestion
    @Nullable
    @SuppressWarnings("unchecked")
    public <T extends Serializable> T get(@NonNull final String name) {
        return getOrDefaultInternal(name, null)
    }

    @NonNull
    @SuppressWarnings("unchecked")
    public <T extends Serializable> T getOrDefault(@NonNull final String name, @NonNull final T defaultValue) {
       return getOrDefaultInternal(name, defaultValue)
    }

    @NonNull
    @SuppressWarnings("unchecked")
    private <T extends Serializable> T getOrDefaultInternal(@NonNull final String name, @Nullable final T defaultValue) {
        final Object object = mMap.get(name);
        if (object == null) {
            return defaultValue;
        }

        try {
            return (T) object;
        } catch (final ClassCastException e) {
            return defaultValue;
        }
    }
```

shahzaibj:
A bunch of tests seem to be failing

common/src/test/java/com/microsoft/identity/common/internal/commands/GenerateShrCommandTest.java
```diff
@@ -53,6 +53,7 @@
                     .url(UrlUtils.makeUrlSilent("https://url"))
                     .nonce("one")
                     .clockSkewManager(new ClockSkewManager(new InMemoryStorage<Long>()))
+                    .clockSkewManager(new ClockSkewManager(new InMemoryStorage<Long>()))
```
shahzaibj, line 56: Seems an exact copy of the previous line

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -252,21 +295,21 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
...
+            // MSAL needs to save this account locally for future token calls.
+            // parameters.getOAuth2TokenCache() will be non-null only in case of MSAL native
+            // If the request is from MSALCPP , OAuth2TokenCache will be null.
+            if (parameters.getOAuth2TokenCache() != null) {
+                saveMsaAccountToCache(resultBundle, (MsalOAuth2TokenCache) parameters.getOAuth2TokenCache());
+            }
+
             result = new MsalBrokerResultAdapter().getAcquireTokenResultFromResultBundle(resultBundle);
-        } catch (BaseException e) {
+        } catch (BaseException | ExecutionException e) {
```
AdamBJohnsonx, line 312:
```suggestion
        } catch (final BaseException | ExecutionException e) {
```

AdamBJohnsonx (approved):

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -0,0 +1,54 @@
...
+public enum  LocalBroadcaster {
+    INSTANCE;
+
+    public interface IReceiverCallback {
+        void onReceive(@NonNull final DataBag dataBag);
+    }
+
+    final Map<String, IReceiverCallback> mReceivers = new HashMap<>();
+
+    public void registerCallback(@NonNull final String alias, @NonNull final IReceiverCallback callback){
```
AdamBJohnsonx, line 39: Do we want to let this stomp on existing values?

AdamBJohnsonx: This is OK, since it is only used on interactive flows, and those run in a single thread right now.  As a safety feature, you could throw an exception if an overwrite was attempted and then have the register methods that are run enclosed in a try-finally to make sure listeners release these.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -242,6 +251,40 @@ public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandPar
...
+                );
+
+                mBrokerResultFuture.setResult(DataBagUtil.toBundle(dataBag));
+
+                Telemetry.emit(
+                        new ApiEndEvent()
+                                .putApiId(TelemetryEventStrings.Api.BROKER_COMPLETE_ACQUIRE_TOKEN_INTERACTIVE)
+                );
+
+                LocalBroadcaster.INSTANCE.unregisterCallback(RETURN_BROKER_INTERACTIVE_ACQUIRE_TOKEN_RESULT);
```
AdamBJohnsonx, line 284: This should happen in a finally block in this method.

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
```diff
@@ -261,17 +264,20 @@
         static final String RSA = "RSA";
     }
 
-    DevicePopManager() throws KeyStoreException, CertificateException,
+    private final Context mContext;
+
+    public DevicePopManager(@NonNull final Context context) throws KeyStoreException, CertificateException,
```
shoatman, line 269: I assume this is going to be the android application context rather than an activity context.  One of the safety aspects of passing the context in rather than storing it as a member variable... is that presumably when it's passed in the context is alive and well.  May be worth asking @iambmelt is this was part of his considerations.

common/src/main/java/com/microsoft/identity/common/AndroidCommonComponents.java
```diff
@@ -26,27 +26,55 @@
...
 
+import static com.microsoft.identity.common.java.exception.ClientException.CERTIFICATE_LOAD_FAILURE;
+import static com.microsoft.identity.common.java.exception.ClientException.IO_ERROR;
+import static com.microsoft.identity.common.java.exception.ClientException.KEYSTORE_NOT_INITIALIZED;
+import static com.microsoft.identity.common.java.exception.ClientException.NO_SUCH_ALGORITHM;
+
 /**
  * Android implementations of platform-dependent components in Common.
  */
 public class AndroidCommonComponents implements ICommonComponents {
```
shoatman, line 62: I'm not a big fan of this interface name.  IPlatformComponents or IOperatingSystemComponents makes more sense to me.  ICommonComponents is very generic and i think potentially confusing.

shoatman: Then you get: AndroidPlatformComponents or AndroidOperatingSystemComponents / LinuxPlatformComponents, LinuxOperatingSystemComponents

common/src/main/java/com/microsoft/identity/common/AndroidCommonComponents.java
```diff
@@ -26,27 +26,55 @@
...
+     */
+    private static final String SKEW_PREFERENCES_FILENAME =
+            "com.microsoft.identity.client.clock_correction";
+
     protected final Context mContext;
+    private IClockSkewManager mClockSkewManager;
+    private IDevicePopManager mDefaultDevicePoPManager;
 
-    public AndroidCommonComponents(@NonNull final Context context){
+    public AndroidCommonComponents(@NonNull final Context context) {
```
shoatman, line 75: I think the intention here is that this be Application context rather than activity context.  We may want to check that we're getting the kind of context that we expect or at least not getting a context that we don't expect.

shoatman: I see that below the clock skew manager and device pop manager are lazy loaded.  Do you want to instantiate those in the constructor instead?  I'm curious why lazy loading is being used in this case?  If something is wrong with initializing either of those when do you want to find out?

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
```diff
@@ -20,36 +20,46 @@
...
 
-import java.util.concurrent.ConcurrentHashMap;
+import java.io.Serializable;
+import java.util.Map;
 
 import lombok.NonNull;
 
-public class TelemetryMap implements INameValueStorage<CurrentRequestTelemetry> {
+/**
+ * {@link DataBag} utilities.
```
shoatman, line 36: I noticed that DataBag itself doesn't have a meaningful description in teh code comments.  What is a DataBag?  Generally i've heard of Property bags... but a databag is less clear to me.

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
```diff
@@ -20,36 +20,46 @@
...
+     * Converts a {@link Bundle} to a {@link DataBag}
+     * */
+    public static DataBag fromBundle(@NonNull final Bundle bundle) {
+        final DataBag result = new DataBag();
 
-    @Override
-    public void clear() {
-        telemetryMap.clear();
+        for (final String key : bundle.keySet()) {
+            result.put(key, bundle.getSerializable(key));
```
shoatman, line 60: DataBag.put is marked with a NonNull annotation; however it's valid for getSerializeable to return a null.  See: https://developer.android.com/reference/android/os/Bundle#getSerializable(java.lang.String)

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -255,6 +255,15 @@ public void setDeviceCertificateProxyClass(@SuppressWarnings(WarningType.rawtype
         return mClazzDeviceCertProxy;
     }
 
+
+    /**
+     * remove class for work place join related API. This is only used from
+     * Authenticator side.
+     */
+    public void removeDeviceCertificateProxy() {
```
shoatman, line 263: This is definitely strange... likely some additional comments here would help reviewer understand why this is required.

common/src/main/java/com/microsoft/identity/common/internal/authscheme/PopAuthenticationSchemeInternal.java
```diff
@@ -165,16 +152,14 @@ public String getAccessTokenForScheme(@NonNull final String accessToken) throws
...
-                .getDevicePoPManagerInstance()
-                .mintSignedAccessToken(
-                        getHttpMethod(),
-                        timestampMillis / ONE_SECOND_MILLIS,
-                        getUrl(),
-                        accessToken,
-                        getNonce(),
-                        getClientClaims()
-                );
+        return mPopManager.mintSignedAccessToken(
```
shoatman, line 155: I didn't catch this earlier... i think we should try to stick to a limited number of verbs for conveying create.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
```diff
@@ -26,14 +26,18 @@
 import android.content.Intent;
 import android.os.Bundle;
 
-import androidx.localbroadcastmanager.content.LocalBroadcastManager;
-
-import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.DataBagUtil;
+import com.microsoft.identity.common.java.util.ported.DataBag;
```
shoatman, line 30: What does the "ported" namespace convey?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -80,7 +80,7 @@ public static BaseException exceptionFromAcquireTokenResult(final AcquireTokenRe
                                 return new DeviceRegistrationRequiredException(
                                         microsoftAuthorizationErrorResponse.getError(),
                                         microsoftAuthorizationErrorResponse.getErrorDescription(),
-                                        microsoftAuthorizationErrorResponse.getUserName()
+                                        microsoftAuthorizationErrorResponse.getUpnToWpj()
```
shoatman, line 83: Is this always a UPN?

shoatman (changes requested):
I think we need to review the state paramemter... i read your overview, but the sentence there end s half way... so not sure what's happening.  the state send in the request needs to match the state in the response.

rpdome:
> I think we need to review the state paramemter... i read your overview, but the sentence there end s half way... so not sure what's happening. the state send in the request needs to match the state in the response.

It should be "never used". I made the change so that the comparison source is always the request itself.

common/src/main/java/com/microsoft/identity/common/AndroidCommonComponents.java
rpdome: I'll rename it in a separate PR. (Already have 2 PRs on top of this, and renaming it now would end up in a conflict mess).

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
rpdome: I can change this to PropertyBag.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
rpdome: yes

shoatman (approved):
Approving based on our discussion this morning.

common/src/main/java/com/microsoft/identity/common/DataBagUtil.java
rpdome: the value field is nullable.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
rpdome: Classes that exists in Android or newer version of java.

common/src/main/java/com/microsoft/identity/common/internal/platform/DevicePopManager.java
iambmelt: Yes, this was an intentional design consideration of this API.... I see I'm 5 days too late to give this feedback :(

PR #1480, opened 2021-07-26
Adjoh/kill mockito

AdamBJohnsonx:
This is one of a trio of changes unblocking the automation tests once more.  We saw a failure design objenesis (dependency of mockito), and it looks like a versioning problem.  So we bump the mockitol versions and sever the test dependency on mockito so as to avoid including mockito-inline and mockito-android in the same closure (See: https://github.com/mockito/mockito/pull/2024 ). Additional changes accommodate it.

### DIFF
--- a/common/versioning/version.properties
+++ b/common/versioning/version.properties
@@ -1,4 +1,4 @@
 #Tue Apr 06 22:55:08 UTC 2021
-versionName=3.5.0-RC2
+versionName=3.5.0-RC3
 versionCode=1
 latestPatchVersion=180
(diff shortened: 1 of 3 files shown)


common/versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Tue Apr 06 22:55:08 UTC 2021
-versionName=3.5.0-RC2
+versionName=3.5.0-RC3
```
AdamBJohnsonx, line 2:
```suggestion
versionName=3.5.0
```

PR #1483, opened 2021-07-27
Migrate AccountManager Account

rpdome:
ref: https://github.com/AzureAD/ad-accounts-for-android/pull/1643


common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -0,0 +1,149 @@
...
+     * Logs and throw {@link ClassCastException} if it fails to parse.
+     */
+    @NonNull
+    public static AndroidBrokerAccount cast(@NonNull final IBrokerAccount account) {
+        final String methodName = ":cast";
+        try {
+            return (AndroidBrokerAccount) account;
+        } catch (final ClassCastException e) {
+            Logger.error(TAG + methodName,
+                    "Expected an AndroidBrokerAccount, but got something else.", e);
```
shahzaibj, line 73: Can we log what we got instead?

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @NonNull
+    public static AndroidBrokerAccount load(@NonNull final Account account) {
+        return AndroidBrokerAccount.builder()
+                .account(account)
+                .build();
+    }
+
+    @NonNull
+    public static AndroidBrokerAccount create(@NonNull final AccountManager accountManager,
```
shahzaibj, line 86: Why does it need to be supplied `accountManager` as a param? Can't we just initialize that in this class?

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -0,0 +1,149 @@
...
+                                              @NonNull final String accountName,
+                                              @NonNull final String accountType) {
+        final String methodName = ":create";
+
+        Account account = getAccount(accountManager, accountName, accountType);
+        if (account == null) {
+            account = new Account(accountName, accountType);
+            Logger.verbose(TAG + methodName, "Creating account.");
+            Logger.verbosePII(TAG + methodName, ACCOUNT_NAME + ":" + account.name);
```
shahzaibj, line 95:
```suggestion
            Logger.verbosePII(TAG + methodName, "Creating account with " + ACCOUNT_NAME + ":" + account.name);
```

common4j/src/main/com/microsoft/identity/common/java/broker/IBrokerAccount.java
```diff
@@ -0,0 +1,34 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.broker;
+
+import lombok.NonNull;
+
+public interface IBrokerAccount {
```
shahzaibj, line 27: javadoc

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
```diff
@@ -0,0 +1,59 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.java.cache;
+
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+public class HttpCache {
```
shahzaibj, line 29: javadoc

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private static IHttpCacheCallback sHttpCache;
+
+    private static final ReentrantReadWriteLock sLock = new ReentrantReadWriteLock();
+
+    public interface IHttpCacheCallback {
+        void flush();
+    }
```
shahzaibj, line 37: javadoc

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
```diff
@@ -0,0 +1,59 @@
...
+    public static void setHttpCache(@Nullable IHttpCacheCallback httpCache) {
+        sLock.writeLock().lock();
+        try {
+            sHttpCache = httpCache;
+        } finally {
+            sLock.writeLock().unlock();
+        }
+    }
+
+    public static void flush() {
```
shahzaibj, line 48: javadoc on all public methods

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -0,0 +1,149 @@
...
+    @NonNull
+    public String getUsername() {
+        return mAccount.name;
+    }
+
+    /**
+     * Logs and throw {@link ClassCastException} if it fails to parse.
+     */
+    @NonNull
+    public static AndroidBrokerAccount cast(@NonNull final IBrokerAccount account) {
```
shoatman, line 67: This throws a runtime exception... wonder if we need this wrapper if we're just going to throw a runtime exception anyway.

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
AdamBJohnsonx: +1

shoatman (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
AdamBJohnsonx, line 67: Can we declare an unwrap method instead on a generic IBrokerAccount interface?

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -0,0 +1,149 @@
...
+            return (AndroidBrokerAccount) account;
+        } catch (final ClassCastException e) {
+            Logger.error(TAG + methodName,
+                    "Expected an AndroidBrokerAccount, but got something else.", e);
+            throw e;
+        }
+    }
+
+    @NonNull
+    public static AndroidBrokerAccount load(@NonNull final Account account) {
```
shahzaibj, line 79: Is there a still a 0-arg constructor in this class? If yes, should it be made private explicitly?

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
shoatman: I'd also add that you can check instance of before attempting to cast....

shahzaibj (approved):
Left some minor comments

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/SilentTokenCommandParameters.java
```diff
@@ -73,25 +77,36 @@ private boolean isAuthorityB2C() {
...
+                cause,
+                errorCode);
     }
 
-    private static void performCloudDiscovery() throws IOException {
+    private static void performCloudDiscovery()
+            throws IOException, URISyntaxException {
         final String methodName = ":performCloudDiscovery";
         Logger.verbose(
                 TAG + methodName,
```
AdamBJohnsonx, line 112: OK, so we're statically locked for cloud discovery.  Interesting.

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -0,0 +1,29 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.java.authorities;
+
+public class AzureActiveDirectoryAudience {
```
AdamBJohnsonx, line 25: I could wish this was something iterable, so we could potentially try them all.

common4j/src/main/com/microsoft/identity/common/java/broker/IBrokerAccount.java
AdamBJohnsonx: Should this interface be generic in the type that it wraps?

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
AdamBJohnsonx: Really thinking we need to hide this in the client, and it needs to be non-static.  Since we're using factories to access things anyway, should we have a static platform factory class as well as the *Components classes?

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+public class HttpCache {
+
+    private static IHttpCacheCallback sHttpCache;
+
+    private static final ReentrantReadWriteLock sLock = new ReentrantReadWriteLock();
```
AdamBJohnsonx, line 33: Yeah, we should really get this from a static factory, so we can change it without needing the synchronization,

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -219,4 +228,36 @@ public static String fromByteArray(@NonNull final byte[] bytes) {
...
+     * Converts a Date object into a RFC3339 formatted date String
+     *
+     * @param date Date
+     * @return String
+     */
+    @NonNull
+    public static String RFC3339DateToString(@NonNull final Date date) {
+        final String methodName = "RFC3339DateToString";
+        Logger.verbose(TAG + methodName, "RFC3339DateToString is called.");
+        final SimpleDateFormat RFC3339DateFormat = new SimpleDateFormat(RFC3339_DATE_FORMAT, Locale.US);
```
AdamBJohnsonx, line 242: Are fully desugared at this point?  Can we move to java.time?

AdamBJohnsonx (approved):
Minor things.

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
AdamBJohnsonx: I mean, alternately have the factory store an atomic or get a Configuration from somewhere on start.

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
rpdome: I prefer not to have an account object holds an accountmanager.

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
rpdome: This is not a full class. I partially move these three variables here first to..... keep the size of PR... somewhat in control.

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
rpdome: I primarily add this for the log message.

common4j/src/main/com/microsoft/identity/common/java/broker/IBrokerAccount.java
rpdome: @AdamBJohnsonx I'd say no, as in, I don't see the benefit that would make it worth having to suppress "raw use of parameter" (or supply <?> everywhere we use this).

common4j/src/main/com/microsoft/identity/common/java/cache/HttpCache.java
rpdome: the way this is being used is similar to Logger, Telemetry, Device class, as in, they are being called out of nowhere.

Having to pass a factory class everywhere would make them harder to use.

We could have a static platform factory class, but I prefer having only 2 patterns instead of 3. If we are going that route, then we should move Logger, Telemetry, Device to that one as well.... probably something to think about post P0.

rpdome: https://identitydivision.visualstudio.com/Engineering/_queries/edit/1446143/?triage=true&workitem=1485244

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: I prefer dealing with this later. For now i'll keep this as-is.

PR #1486, opened 2021-07-31
Move Account Manager User Data look-up constants from common, AADAuthenticator to common4j

shahzaibj:
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1647

This PR achieves the following:

**- Move Account Manager User Data look-up constants from common, AADAuthenticator to common4j**
Move constants used for account manager storage data, and relocate them from common and AADAuthenticator to common4j. The reason is because we need to access them in LinuxBroker as well. The previous original constants now just references them from the new file created in this PR (BrokerAccountDataName). At a later point, we will just delete the older ones once the refactoring is complete.

**- Replicate createHash method in common4j's StringUtil.java (from StringExtensions.java)**
Add a `createHash` method to `StringUtils.java` because we need it in Linux too...this is the same code take from `StringExtensions.java`...leaving that old method there for now because that's being used in a lot of places too. Over the course of refactoring the usage will go away and we will delete the old one.

**- Add tests for AndroidBrokerAccount**
Add tests for AndroidBrokerAccount.java (also added EqualsAndHashCode to AndroidBrokerAccount to get the tests to pass)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/broker/BrokerAccountDataName.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/broker/BrokerAccountDataName.java
@@ -1,0 +1,116 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


common4j/src/main/com/microsoft/identity/common/java/broker/BrokerAccountDataName.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private BrokerAccountDataName() {
+    }
+
+    /**
+     * Key for the data used by the PRTv3 protocol.
+     */
+    public static final String PRT_V3 = "workplaceJoin.key.prt3";
+
+    public static final String PRT = "workplaceJoin.key.prt";
```
rpdome, line 35: nit: it would be great to fill the missing comments.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -260,4 +265,20 @@ public static Date RFC3339StringToDate(@NonNull String dateStr) throws ParseExce
...
     }
 
+    /**
+     * Create the Hash string of the message.
+     *
+     * @param msg String
+     * @return String in Hash
+     * @throws NoSuchAlgorithmException throws if no such algorithm.
+     */
+    public static String createHash(final String msg) throws NoSuchAlgorithmException {
```
rpdome, line 275: nit: can you kill its counterpart in StringExtension?

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
shahzaibj: I tried...but that one for some reason also throws an `UnsupportedEncodingException`. I don't know why because it is the same code...it seems when the code is written in an Android module then it uses the java in Android SDK for compilation and that throws an UnsupportedEncodingException when calling `getBytes` on a String...that `createHash` method was being used in a lot of places and the `UnsupportedEncodingException` was being thrown up the hierarchy and I didn't want to touch all that now...so I left it alone. 😁

shoatman (approved):

PR #1488, opened 2021-08-03
Migrate Strategies, Parameters, and some other stuffs.

rpdome:
## Changes
1. Make IPlatformComponents (renamed from ICommonComponents as suggested by @shoatman) a "Required (Nonnull)" parameter in `CommandParameter`
2. Added IAuthorizationStrategyFactory to common4j, and is now consumable via `IPlatformComponents`
    - The current `AuthorizationStrategy` becomes `AndroidAuthorizationStrategy` (generated by `AndroidPlatformComponents`)
3. Extract a handful of 'small functions that varies from platform to platform' into `IPlatformUtil` 
    - Needs an extra eyes on this, especially on `AndroidPlatformUtil.optionallyReorderTasks()` (I moved them out of the Command and CommandDispatcher to here.)
  
~4. Introduced a static class `CommandLineParameter`, which operates the same way as Logger - initialized first (in `AndroidPlatformComponents`), and then accessible anywhere.~
  ~- I would only go with the static route when 1. It's not a critical operation 2. It's being used everywhere, especially in the layer where we're not passing `Context` into right now.~
Killing this off based on Shahzaib's recommendation below.

5. Split `IBrokerRequestAdapter` into 
   1. `IBrokerRequestAdapter`-  an adapter for 'packing the request and send to Broker'
   2. `IBrokerCommandParameterAdapter` - an adapter for 'unpacking the request on the broker side.
   The reason being the latter requires `AndroidBrokerPlatformComponents`, and therefore its implementation has to sit in AADAuthenticator.
6. Migrate a handful of classes from common to common4j.
   - CommandParameter classes.
   - AuthenticationScheme classes.
   - BrowserDescriptor.
      - IMO, BrowserDescriptor doesn't belong in common4j, but I don't want to deal with the config serialization issue right now.
   - Authority classes.
   - PKeyAuthChallenge + its handler.
   - AuthenticationSettings's `get/setDeviceCertificateProxyClass()` (This is being used to save/load cert for PKeyAuth scenario)
   - AzureActiveDirectoryAudience (the rest of the class)
   - OAuth2TokenCache (just the base class only, cc:@AdamBJohnsonx)
7. Migrate the following code from broker to common 
   - IntuneMAMEnrollmentIdGateway
8. Changed usages of URL to URI as suggested by SpotBugs (that it could be perf hog, as URL's equals() and hashCode() does domain name resolution)
   - This includes anything else that uses them, like Map and Set.


common/src/main/java/com/microsoft/identity/common/internal/broker/IntuneMAMEnrollmentIdGateway.java
```diff
@@ -0,0 +1,171 @@
...
+            if (!userId.equals(cacheKey.userId)) return false;
+            return packageName.equals(cacheKey.packageName);
+        }
+
+        @Override
+        public int hashCode() {
+            int result = userId.hashCode();
+            result = 31 * result + packageName.hashCode();
+            return result;
+        }
```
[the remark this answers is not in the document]
rpdome: I could make that change. This is move as-is from broker though.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
```diff
@@ -0,0 +1,58 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java;
+
+import com.microsoft.identity.common.java.challengehandlers.IDeviceCertificate;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public enum AuthenticationSettings {
```
shahzaibj, line 29: I think we already have something called `AuthenticationSettings` (for setting package signature in ADAL)..should we rename this class to something else?

shahzaibj: javadoc

common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
```diff
@@ -0,0 +1,210 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.security.interfaces.RSAPrivateKey;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public class PKeyAuthChallenge implements Serializable {
```
shahzaibj, line 46: javadoc

common4j/src/main/com/microsoft/identity/common/java/util/CommandLineParameter.java
```diff
@@ -0,0 +1,53 @@
...
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.util;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+
+/**
+ * Static class for accessing command line parameters.
+ * */
+public class CommandLineParameter {
```
shahzaibj, line 31: Why do we need this? 

If the class where these (slice, and dc) are set is already moved to common4j, then we shouldn't a platform-dependent way to set these...these are just gradle properties...we should be able to set them directly from the BuildConfig in common4j. Any consumers of common4j (common Android, Linux Broker) shouldn't care about these and they don't need to provide a way to set these.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
rpdome: The below field was actually taken from that class. I can definitely rename it.

common4j/src/main/com/microsoft/identity/common/java/util/CommandLineParameter.java
rpdome: AFAIK, BuildConfig is an Android thing. There's a plugin for java (https://github.com/mfuerstenau/gradle-buildconfig-plugin), but it hasn't been updated since early 2017.

rpdome: We could figure out an xplat way to set this value - if you have any suggestion (and it's small), I can do it now. Otherwise I could create a PBI for this.

shahzaibj: We are using this one: https://github.com/gmazzo/gradle-buildconfig-plugin (updated 20 days ago)

READ THIS: https://github.com/AzureAD/android-complete/blob/master/docs/ProjectBuild/build_config.md

I already did it for broker4j...I might not have done it yet for common4j...but I think we should do that, and then we don't need a class called CommandLineParameter

rpdome: ok. will do. Thanks!

common/src/main/java/com/microsoft/identity/common/internal/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -91,7 +92,7 @@ public void validate() throws ArgumentException {
 
         // Check if SDK is capable of MSA to ensure there is uniformity of logic with SdkType.MSALCPP and SdkType.MSAL
         if (getSdkType().isCapableOfMSA() &&
-                !BrokerValidator.isValidBrokerRedirect(getRedirectUri(), getAndroidApplicationContext(), getCallerPackageName())) {
+                !getPlatformComponents().getPlatformUtil().isValidCallingApp(getRedirectUri(), getCallerPackageName())) {
```
shoatman, line 95: I like this rename.  Thanks

shoatman (approved):
Thanks Dome.

common4j/src/main/com/microsoft/identity/common/java/util/CommandLineParameter.java
rpdome: Done.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -63,7 +63,7 @@
     @Override
     public String toString() {
         if (this == UNSET) {
-            return null;
+            return "";
```
shahzaibj, line 66: Why are we changing the behavior here?

common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
```diff
@@ -0,0 +1,210 @@
...
+            constructor = certClazz.getDeclaredConstructor();
+            deviceCertProxy = (IDeviceCertificate) constructor.newInstance((Object[]) null);
+        } catch (final NoSuchMethodException | InstantiationException | IllegalAccessException
+                | IllegalArgumentException | InvocationTargetException e) {
+            throw new ClientException(ErrorStrings.DEVICE_CERTIFICATE_API_EXCEPTION,
+                    "WPJ Api constructor is not defined", e);
+        }
+        return deviceCertProxy;
+    }
+}
```
shahzaibj, line 210: nit: new line ending

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
```diff
@@ -0,0 +1,58 @@
...
+        if (IDeviceCertificate.class.isAssignableFrom(clazz)) {
+            mClazzDeviceCertProxy = clazz;
+        } else {
+            throw new IllegalArgumentException("clazz");
+        }
+    }
+
+    /**
+     * get class for work place join related API. This is only used from
+     * Authenticator side.
```
shahzaibj, line 50: Authenticator as in Authenticator app? or AADAuthenticator?

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
```diff
@@ -0,0 +1,58 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+public enum AuthenticationSettings {
+    INSTANCE;
+
+    private Class<?> mClazzDeviceCertProxy;
+
+    /**
+     * set class for work place join related API. This is only used from
```
shahzaibj, line 35:
```suggestion
     * Set class for work place join related API. This is only used from
```

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
```diff
@@ -0,0 +1,58 @@
...
+    public void setDeviceCertificateProxyClass(@NonNull final Class<?> clazz) {
+        if (IDeviceCertificate.class.isAssignableFrom(clazz)) {
+            mClazzDeviceCertProxy = clazz;
+        } else {
+            throw new IllegalArgumentException("clazz");
+        }
+    }
+
+    /**
+     * get class for work place join related API. This is only used from
```
shahzaibj, line 49:
```suggestion
     * Get class for work place join related API. This is only used from
```

shahzaibj (approved):
Left some minor comments, else LGTM

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OpenIdConnectPromptParameter.java
rpdome: Spotbugs. toString should be nonnull.

common/src/main/java/com/microsoft/identity/common/internal/broker/IntuneMAMEnrollmentIdGateway.java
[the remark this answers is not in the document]
rpdome: keeping the same as-is for now.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
rpdome: Broker process.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationSettings.java
rpdome: Keeping the same as-is for now.

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
```diff
@@ -95,6 +99,10 @@
             GenericAccount,
             GenericRefreshToken> mAccountCredentialAdapter;
 
+    @Getter
+    @Accessors(prefix = "m")
+    private final Context mContext;
```
AdamBJohnsonx, line 104: This seems like it's common to all of these classes.  Should we interpose an AbstractAndroidOAuth2TokenCache here that just holds the context?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
```diff
@@ -77,9 +79,42 @@ public static Browser select(final Context context, final List<BrowserDescriptor
         throw new ClientException(ErrorStrings.NO_AVAILABLE_BROWSER_FOUND, "No available browser installed on the device.");
     }
 
+    private static boolean matches(@NonNull final BrowserDescriptor browserDescriptor,
```
AdamBJohnsonx, line 82: It is strange that this moved.  Should we have a platform-dependent variant of BrowserDescriptor?  It looks like we're establishing a need for a platform-dependent BrowserSelector instead.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Migrate Strategies, Parameters, and some other stuffs. (#1488)
```
AdamBJohnsonx, line 3: bit: maybe a more customer-friendly changelog entry is desirable.

common/build.gradle
```diff
@@ -37,17 +37,6 @@ android {
 
     compileSdkVersion rootProject.ext.compileSdkVersion
 
-    def sliceParameter = "" // will be blank unless specified by developer
```
AdamBJohnsonx, line 40: Is there a plan for this?

rpdome: this is moved to 4j.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -59,29 +67,82 @@
...
+     */
+    private static boolean sInitialized = false;
+
+    /**
+     * Initializes platform-dependent static classes.
+     */
+    public static synchronized void initializeStaticClasses() {
+        if (!sInitialized) {
+            Device.setDeviceMetadata(new AndroidDeviceMetadata());
+            Logger.setAndroidLogger();
```
AdamBJohnsonx, line 102: It looks like these are not dependent at all on the members of the class.  Do we want to do this initialization in a static context instead of a checked variable at construction time?  If not, it's worth it to say why.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -147,4 +208,28 @@ public synchronized IDevicePopManager getDefaultDevicePopManager() throws Client
...
+                .context(mContext)
+                .activity(mActivity)
+                .fragment(mFragment)
+                .build();
+    }
+
+    @Override
+    public @NonNull IStateGenerator getStateGenerator() {
+        if (mActivity == null) {
+            throw new IllegalStateException("StateGenerator requires an activity");
```
AdamBJohnsonx, line 225: It looks like we're creating, functionally, two separate objects here that are differentiated in behavior by their initialization parameters.  Do we want to throw this exception here, or have our static factory method return a different implementation that might not have an interface that exposes this getStateGenerator method so that we can avoid accidentally hitting this exception?

common/src/main/java/com/microsoft/identity/common/internal/request/AuthenticationSchemeTypeAdapter.java
```diff
@@ -32,25 +34,35 @@
...
  * Gson de/serialization utility class for auth schemes.
  */
-class AuthenticationSchemeTypeAdapter implements
+public class AuthenticationSchemeTypeAdapter implements
         JsonDeserializer<AbstractAuthenticationScheme>,
         JsonSerializer<AbstractAuthenticationScheme> {
 
     private static final String TAG = AuthenticationSchemeTypeAdapter.class.getSimpleName();
 
+    private static Gson sRequestAdapterGsonInstance = new GsonBuilder()
```
AdamBJohnsonx, line 56: (thought, not to be dealt with now) When I've done this with XML using JaxbContext, it helped to have one global JaxbContext for the entire application with all types defined therein.  I'm sure that the Json equivalent did something similar when it came time to shift serialization formats. I'm wondering if a similar pattern should hold for us.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
AdamBJohnsonx: I see this called out in the description.

common4j/build.gradle
```diff
@@ -126,9 +132,22 @@ publishing {
     }
 }
 
+def sliceParameter = "" // will be blank unless specified by developer
```
AdamBJohnsonx, line 135: OK, this moved here.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -78,9 +78,21 @@ public void setSlice(AzureActiveDirectorySlice slice) {
         mSlice = slice;
     }
 
-    public abstract Uri getAuthorityUri();
+    public URI getAuthorityUri() {
+        try {
+            return new URI(mAuthorityUrl);
```
AdamBJohnsonx, line 83: to save people like me from wondering why we're converting a URL to a URI and back again, maybe rename this variable to mAuthorityUrlString?

AdamBJohnsonx (approved):
Generally approve. Maybe fix the changelog entry to be less casual.  Some documentation asks.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
rpdome: Eventually - yes - at the entry point of broker.
Right now we're still working our way there, so I put it here. Will add todo to call this out.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
rpdome: In Android, this seems to be the case. This does/might not be true in Linux though.

IMO separating interface for this might do more harm than good - as in, we'll have to do extra casting work, and at the point we need to use this, and we try to cast the object and it failed, then we'll have to throw an exception anyway.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BaseController.java
```diff
@@ -173,7 +169,7 @@ public abstract AcquireTokenResult acquireDeviceCodeFlowToken(@SuppressWarnings(
         UUID correlationId = null;
 
         try {
-            correlationId = UUID.fromString(DiagnosticContext.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
+            correlationId = UUID.fromString(DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
```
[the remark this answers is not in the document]
rpdome: https://identitydivision.visualstudio.com/Engineering/_queries/edit/1446143/?triage=true&workitem=1493186

common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java
rpdome: Let's keep it here for now to keep the change minimum.(They'll be 'abstracted out' anyway).

PR #1490, opened 2021-08-03
Modify MSAL CPP JS injection flow

sansrivastav:
Some of the tests like the one which require Credential owner Aad have separate pages for username and password.
In such test flows, onPageCommitVisible() is called only once because of which JavaScript injection gets evaluated only once.
This makes the test get stuck at the password page and the test times out.
So, in this PR we move back the JavaScript injection to the onPageLoaded() callback.
In addition, we enhance the JavaScript evaluation by adding expectedUrl check to prevent overloading the server.
Corresponding MSAL changes: https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/2367


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -128,20 +129,24 @@ void extractState(@NonNull final Bundle state) {
...
-                    public void onPageLoaded() {
+                    public void onPageLoaded(final String url) {
                         mProgressBar.setVisibility(View.INVISIBLE);
-                    }
-                },
-                new OnPageCommitVisibleCallback() {
-                    @Override
-                    public void onPageCommitVisible() {
+                        try {
+                            javascriptToExecute[0] = String.format("window.expectedUrl = '%s';%n%s",
```
PetarSDimov, line 143: >javascriptToExecute[0] [](http://example.com/codeflow?start=28&length=22)

This is never used. Shouldn't this be used insted of mPostPageLoadedJavascript in the code below?
Also why is it an array?

sansrivastav: Modified code to use javascriptToExecute[0] instead of mPostPageLoadedJavascript.
As javascriptToExecute[0] needs to be declared as final, so declared it as 1 element array.

PetarSDimov (approved):
:shipit:

AdamBJohnsonx:
From a sanity point of view, could we have a brief discussion of why this fixes things in the pull request?  Also, we'll likely want to get this in a release of common soon-ish, which would mean cherry-picking it.  Could you refashion this into a single commit?

changelog.txt
```diff
@@ -90,7 +90,7 @@ Version 3.2.0
 - [MINOR] Implement an in-memory cache to avoid multiple decryptions of shared preferences (#1254)
 - [PATCH] Fixes deprecated NetworkInfo class (#847)
 - [MINOR] Device PoP keys are now generated with attestation flags and expose a certificate chain getter (#1247)
-- [PATCH] Fix MSAL-CPP injecting javascript loop (#1238)
+- [PATCH] Fix MSAL-CPP injecting javascript loop (#1238, #1490)
```
AdamBJohnsonx, line 93: This needs a new changelog entry for the version that will be released, not version 3.2, where it will not be fides.

AdamBJohnsonx:
This was altered previously here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1238

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -132,6 +133,7 @@ void extractState(@NonNull final Bundle state) {
     public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
         final String methodName = "#onCreateView";
         final View view = inflater.inflate(R.layout.common_activity_authentication, container, false);
+        final String[] javascriptToExecute = new String[1];
```
AdamBJohnsonx, line 136: This is minor, but I'm unclear on the need for this to be an array reference.  Since it's never in use outside of your anonymous OnPageLoadedCallback, why can it not be local?

AdamBJohnsonx (commented):
This was changed previously to a different mechanism.  Would just like you to be certain this isn't going to start failing in a similar manner it was before https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1238

changelog.txt
sansrivastav: Added new changelog entry.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
sansrivastav: Moved the array declaration inside onPageLoaded() callback. Thanks.

sansrivastav:
We discussed the current issue that we are facing with Richard(based on his changes in #1238 ) and made the current changes based on his suggestion. I have also verified locally these changes for test cases which were failing sometimes before the #1238 changes( like InteractiveSignInClickBackButton.SignInInteractively_ClickBackButton_MSA). I have not observed any failure in my local debugging so far.

AdamBJohnsonx (approved):

AdamBJohnsonx:
> We discussed the current issue that we are facing with Richard(based on his changes in #1238 ) and made the current changes based on his suggestion. I have also verified locally these changes for test cases which were failing sometimes before the #1238 changes( like InteractiveSignInClickBackButton.SignInInteractively_ClickBackButton_MSA). I have not observed any failure in my local debugging so far.

Last question - what's the urgency on this?  Currently this is changed against dev, but we're shipping from release branches cut from master.

@iambmelt - you may want to be aware.

PR #1491, opened 2021-08-03
Fixing Perf Report Issues

paulkagiri:
This PR does two main things:
- It aims to fix the intermittent ConcurrentModificationException that has been happening sometimes when running the Perf pipeline
- It also aims to unify the perf reports sent out to be one email with both results of the base run as well as the latest run.

This is how the unified report looks like:

updated_report


common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -35,20 +35,19 @@
...
-        if (codeMarkers.size() > 0) {
-            stringToWrite.append(codeMarkers.get(0).getCsvHeader());
-        }
+        final StringBuilder content = new StringBuilder();
+        content.append(codeMarkers.get(0).getCsvHeader());
 
-        for (final CodeMarker codeMarker : codeMarkers) {
-            stringToWrite.append('\n');
-            stringToWrite.append(codeMarker.getCsvLine());
+        synchronized (codeMarkers) {
```
AdamBJohnsonx, line 45: This is interesting.  Is something else also synchronizing on this monitor?

Om83:
hi Paul, Some suggestions for the report. You can make following changes to your Table headers.
Base Build -> Last Released Build
Latest Build -> Last Dev Build
Target Build -> Current Build 
Job Summary - > Perf Report
Perf Run XXX vs Base Run - > Perf comparison - Current vs. Last Released Build
Perf Run XXX vs Latest Run -> Perf Comparison - Current vs. Last Dev Build

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
paulkagiri: I thought of synchronizing on code markers so that any modifications at this point don't affect the list. I would also appreciate your advice on how best to handle this: [`markCode`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java#L63) adds the code markers, [`clearMarkers`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/CodeMarkerManager.java#L117) removes all while [`getCsvContent`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/c6bb5aaf9d377ba5a40bbdc224cceb81543881a3/common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java#L46) has the loop that sometimes was failing with ConcurrentModificationException

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
```diff
@@ -35,20 +35,19 @@
...
+            for (CodeMarker codeMarker : codeMarkers) {
```
shahzaibj, line 46:
```suggestion
            for (final CodeMarker codeMarker : codeMarkers) {
```

shoatman (approved):

common/src/main/java/com/microsoft/identity/common/CodeMarkerUtil.java
AdamBJohnsonx: OK, this method is counting on receiving a SynchronizedList, so that the synchronization on the list itself blocks all other accesses - we should document that, since any other user will not know it if that's what we're after.  Alternately, since this is all called from within the CodeMarkerManager file, we could just synchronize those methods, rather than having them synchronize internally on the list reference.

Is it possible that 1/2 of a pair or markers could be added while this process goes on?

Is there any reason that this is in a separate class?

AdamBJohnsonx (approved):
My comments are something that would need a larger discussion.  I think we can move on this now and take understanding the synchronization as a follow up.

uiautomationutilities/perf_tool/Program.cs
```diff
@@ -23,19 +23,132 @@
...
+            task.AppName = appName;
+            task.Device = deviceModel;
+            task.FeatureGateOverrides = new Dictionary<string, string>();
+            task.Id = id;
+            task.LogsDir = artifactURL;
+            return task;
+        }
+    }
+
+    class ReportParams
```
shahzaibj, line 115: Can we add javadoc (I mean the equivalent of that in C#)?

shahzaibj (approved):

PR #1493, opened 2021-08-05
Make change to accommodate WPJ migration

rpdome:
ref: https://github.com/AzureAD/ad-accounts-for-android/pull/1650

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -26,11 +26,23 @@ import android.os.Build;
 
 import com.microsoft.identity.common.java.platform.AbstractDeviceMetadata;
 
+import lombok.NonNull;
+
 /**
  * Provides device metadata in Android.
  **/
 public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
+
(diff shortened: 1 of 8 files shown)


common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
```diff
@@ -103,6 +103,18 @@ public static String getProductVersion() {
         }
     }
 
+    public static String getDeviceType() {
+        sLock.readLock().lock();
```
AdamBJohnsonx, line 107: Could we get an @GuardedBy("sLock") annotation on sDeviceMetadata?

AdamBJohnsonx (approved):
This looks straightforward.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -45,16 +57,19 @@ public String getCpu() {
     }
 
     @Override
+    @NonNull
     public String getOs() {
-        return String.valueOf(Build.VERSION.SDK_INT);
+        return android.os.Build.VERSION.RELEASE;
```
shahzaibj, line 62: Interesting that we're changing this. I hope this resolves to the same value

shahzaibj (approved):

PR #1498, opened 2021-08-06
Revert getOS in AndroidDeviceMetadata

rpdome:
Build.VERSION.SDK_INT returns SDK version (i.e. 24)
android.os.Build.VERSION.RELEASE returns user-friendly version (i.e. 7.0)

I'm seeing inconsistency, as in, we're sending 7.0 in the registration call, however, we can patch this later with another request.

The broker/common code seems to be using 24 always.

Therefore, i'm inclining to keeping the same behavior here.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -56,10 +56,11 @@ public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
         return "UNKNOWN";
     }
 
+    // Returns a SDK version (i.e. 24) instead of a user-friendly android version (i.e. 7.0)
     @Override
     @NonNull
     public String getOs() {
-        return android.os.Build.VERSION.RELEASE;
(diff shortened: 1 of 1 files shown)


shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -56,10 +56,11 @@ public String getCpu() {
         return "UNKNOWN";
     }
 
+    // Returns a SDK version (i.e. 24) instead of a user-friendly android version (i.e. 7.0)
```
AdamBJohnsonx, line 59: Do we want to support the other concept as well?

AdamBJohnsonx (approved):

iambmelt (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
rpdome: I prefer not to do that unless we really have to. Will check with satheesh and co.

PR #1501, opened 2021-08-06
Fix for sov cloud TSL scenarios

iambmelt:
Thanks @irving-as for providing this fix.

This PR corrects an issue when setting the home authority on a refresh token; previously this implementation (incorrectly) assumed that all tokens would be issued to the WW cloud authority. In order to support TSL scenarios in sov clouds this updated implementation sets the authority to the appropriate `/common` endpoint for the particular cloud that issued the token. `/common` is used at the tenant on the authority as ADAL's legacy application uses this value when looking up RTs from cache during account/token discovery


irving-as (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -99,6 +97,39 @@ public TokenShareUtility(@NonNull final String clientId,
         mTokenCache = cache;
     }
 
+    @NonNull
+    private static String getCanonicalEnvironment(@NonNull final String environment) {
```
rpdome, line 101: nit: a comment describing what "canonical environment" actually means would be helpful.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -99,6 +97,39 @@ public TokenShareUtility(@NonNull final String clientId,
...
+            case "login.windows.net":
+            case "login.microsoft.com":
+            case "sts.windows.net":
+                return ENVIRONMENT_GLOBAL;
+            case "login.chinacloudapi.cn":
+                return ENVIRONMENT_GALLATIN;
+            case "login.usgovcloudapi.net":
+                return ENVIRONMENT_ITAR;
+            default:
+                return environment;
```
AdamBJohnsonx, line 112: Should we be doing anything else to canonicalize this string?

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -58,22 +58,20 @@
...
-
-    /**
-     * To support caching lookups in ADAL, the following authority is used to signal
-     * that the tokens being yielded belong to the target user's home tenant.
-     */
-    private static final String sHomeTenantAuthority = "https://login.windows.net/common";
+    private static final String ENVIRONMENT_GLOBAL = "login.microsoftonline.com";
+    private static final String ENVIRONMENT_GALLATIN = "login.partner.microsoftonline.cn";
+    private static final String ENVIRONMENT_ITAR = "login.microsoftonline.us";
+    private static final String ENVIRONMENT_BLACKFOREST = "login.microsoftonline.de";
```
AdamBJohnsonx, line 74: Should we make this an enum, and then bake the authority in here?

iambmelt: Applied in 48a2ac32baa9d8a894959676e23876143bd7bff6

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
iambmelt: Refactored out this method and put a `static` method on the enum. The intent should now be clearer

iambmelt: 48a2ac32baa9d8a894959676e23876143bd7bff6

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
```diff
@@ -58,22 +58,53 @@
...
+                case "login.windows.net":
+                case "login.microsoft.com":
+                case "sts.windows.net":
+                    return Environment.WORLDWIDE;
+                case "login.chinacloudapi.cn":
+                    return Environment.GALLATIN;
+                case "login.usgovcloudapi.net":
+                    return Environment.ITAR;
+                default:
+                    return null;
```
AdamBJohnsonx, line 100: Do we want a null return here, or should we fail some other way?

iambmelt: I am thinking this should throw a client exception, log a message, and not return the token to the caller if we can't identify the authority. Will push this in another commit

iambmelt: 47d045ff5a33ac16755b181092a8060cd94610ac

common/src/main/java/com/microsoft/identity/common/adal/internal/tokensharing/TokenShareUtility.java
iambmelt: Should be sufficient

AdamBJohnsonx (approved):

PR #1502, opened 2021-08-09
Remove deprecation logging

gitaumoses4:
Remove deprecation logging in `Logger` to stop the spam logs in Authenticator app.


common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -64,129 +62,108 @@
...
         return com.microsoft.identity.common.logging.Logger.getAllowPii();
     }
 
     public static boolean getAllowLogcat() {
-        logDeprecationWarning();
         return com.microsoft.identity.common.logging.Logger.getAllowLogcat();
     }
 
     public static String getDiagnosticContextMetadata() {
-        logDeprecationWarning();
```
AdamBJohnsonx, line 87: Hm.  Should we add a Telemetry.emit() in the place of the logger?  Eventually we'd like to be able to have things like this in metrics. @shahzaibj for thoughts.

AdamBJohnsonx (approved):
This is pretty simple, but I'd like Shahzaib's thoughts on whether we can go ahead and code a telemetry recording step here.

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
shahzaibj: Yes, let add telemetry. Let's create a new `Event` called something like `DeprecatedApiUsageEvent` (or some similar name), and `emit` that event whenever a deprecated api is used. The event should capture the class and name of the API that was used.

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -256,19 +259,17 @@ private static LogLevel adapt(com.microsoft.identity.common.logging.Logger.LogLe
...
+            Telemetry.emit(new DeprecatedApiUsageEvent().putDeprecatedClassUsage(Logger.class));
         }
     }
 
-    private static void logDeprecationWarning() {
-        if (sLogDeprecationWarning) {
-            sLogDeprecationWarning = false;
-            com.microsoft.identity.common.logging.Logger.warn(TAG, "This class is deprecated. "
-                    + "Migrate usage to: com.microsoft.identity.common.logging.Logger");
+    private static void emitDeprecationEvent() {
```
shahzaibj, line 269: I believe you can can just call this static method everywhere and delete the other Non static method.

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
gitaumoses4: @AdamBJohnsonx  why did you add the non static method?

AdamBJohnsonx: > @AdamBJohnsonx why did you add the non static method?

This was to resolve a spotBugs warning.  You could just as well add an @SuppressFBWarnings annotation.

PR #1504, opened 2021-08-09
Seperate OS field in Device class for eSTS and DRS request

rpdome:
In Android, we send...
- SDK version to ESTS (i.e. 24)
- ANDROID_VERSION to DRS (i.e. 7.0)

Currently, ESTS is relying on SDK version to gate their feature, and (an engineer from) DRS doesn't seem to like the idea of replacing ANDROID_VERSION with SDK version.

Given that the concept of SDK VERSION doesn't make sense for xplat code, i'm separating out the fields based on the endpoint.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -58,9 +58,13 @@ public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
 
     // Returns a SDK version (i.e. 24) instead of a user-friendly android version (i.e. 7.0)
     @Override
-    @NonNull
-    public String getOs() {
+    public @NonNull String getOsForEsts() {
         return String.valueOf(Build.VERSION.SDK_INT);
+    }
+
+    @Override
+    public @NonNull String getOsForDrs() {
+        return android.os.Build.VERSION.RELEASE;
     }
 
     @Override
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/AbstractDeviceMetadata.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/AbstractDeviceMetadata.java
@@ -35,6 +35,7 @@ public abstract class AbstractDeviceMetadata implements IDeviceMetadata {
         return getDeviceModel() + SEPARATOR +
                 getManufacturer() + SEPARATOR +
                 getCpu() + SEPARATOR +
-                getOs();
+                getOsForEsts() + SEPARATOR +
+                getOsForDrs();
     }
 }
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
@@ -81,7 +81,7 @@ public class Device {
 
             if (sDeviceMetadata != null) {
                 platformParameters.put(PlatformIdParameters.CPU_PLATFORM, sDeviceMetadata.getCpu());
-                platformParameters.put(PlatformIdParameters.OS, sDeviceMetadata.getOs());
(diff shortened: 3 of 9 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -58,11 +58,15 @@ public String getCpu() {
 
     // Returns a SDK version (i.e. 24) instead of a user-friendly android version (i.e. 7.0)
     @Override
-    @NonNull
-    public String getOs() {
+    public @NonNull String getOsForEsts() {
         return String.valueOf(Build.VERSION.SDK_INT);
     }
 
+    @Override
```
AdamBJohnsonx, line 65: I would suggest that we don't put endpoint-specific names in the DeviceMetadata class.  I'm not sure how this maps to any other platform, however, and would think that we'd want to decide which one of these is the "official" OS version that we want to discuss with the server (probably SDK version) let that be the getOs method, and then give the other one a different name (getOsReleaseVersion or something).

rpdome: I've thought about that, but I think it'll be more confusing that way.

rpdome: (The issue here arises because two endpoints are taking different OS values, so that's why i put the endpoint in that name. This is to make sure that if, for some reason, in the future, we need this OS value for some other stuffs, we're aware of this particular issue, and therefore pick the value we intend to pick.)

rpdome: (of course, if we can have DRS and eSTS agreed on a single OS format, that would be ideal, but I don't think that'll happen soon, if at all.)

AdamBJohnsonx: I don't think this is less confusing, but that's because it puts this in two places: in the client talking to DRS/ESTS *and* in the device metadata.  Think about this - if we sent something to a different service, what would it need to do?

rpdome: I would say we create a new field for that service.

The reason is that we don't know what that new service wants to know. What are the format, etc. Does it want SDK_VERSION or ANDROID_VERSION? Does it prefer linux version to be sent differently. (Of course this will be a decision to be agreed upon by both client and server team, but nothing is set in stone).

My point being that the following
- ANDROID_VERSION and SDK_VERSION
- eSTS and DRS
are equally ranked. Each requires different values for different purposes. Therefore, I don't think we can really mark one of them as 'official'.

rpdome: (If we want to go the 'official' route, I can do that, but I would still prefer having it as the 3rd field.)

AdamBJohnsonx: Generally, what I would have done with this in the past is create a series of OS properties, and have the clients select from that list what they needed to send - no service client name is present in the OS properties production and the client logic is free to take whatever it needs and do whatever it wants to with it to communicate with that server.  Since it seems like both of these properties are legitimate attributes of the android os, I wonder whether that model makes sense.

AdamBJohnsonx: Something to think about, just make sure you weigh these options when you decide how to move forward.

AdamBJohnsonx (approved):

shahzaibj (approved):

rpdome:
The pipeline is botched, so I ran the tests locally and didn't see any issue.

PR #1505, opened 2021-08-09
Clear the asymmetric key if it cannot be accessed when attempting to sign ATs

iambmelt:
If (for whatever reason) the device pop key cannot be accessed when attempting to sign a token the entry will be dropped the from the `AndroidKeystore` and will be regenerated on the next token request.

Per [Android's docs](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder.html#setUserAuthenticationRequired(boolean)) we shouldn't see this error except when `setUserAuthenticationRequired` is set... however, Teams reported seeing this error message pop up in their telemetry.

More information is required to understand when/why this error this might be thrown; [this workitem](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1380806) links to a few bugtracker entry which may explain it on some devices.

For now, we'll assume that the key is forever lost and we'll clean up the entry in the keystore. It will automatically recreated on the next token request


AdamBJohnsonx (approved):
Is it safe to import the class on an older device?  I'd hope so.

iambmelt:
> Is it safe to import the class on an older device? I'd hope so.

Since our target/compile SDK ver is >23 this [should be fine](https://stackoverflow.com/a/10885286)

rpdome (approved):

PR #1511, opened 2021-08-11
Migrate Joined flow part 1

rpdome:
Mostly 1:1 migration, except the following.

1. Extracts OAuth2ErrorCode into its own class (will do the same for SubError and other classes in `AuthenticationConstants` that we haven't started migrating)
2. Merge `ILocalAuthenticationResultBase` into `ILocalAuthenticationResult`.

Also get rid of some duplicated functions/constants.


AdamBJohnsonx (approved):
This looks straightforward.  I think we're in for a fun merge.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -351,4 +356,38 @@ public static String sanitizeNullAndLowercaseAndTrim(final String input) {
 
         return outValue;
     }
+
+    /**
+     * encode string with url form encoding. Space will be +.
```
shahzaibj, line 361:
```suggestion
     * Encode string with url form encoding. Space will be +.
```

shahzaibj: "Space will be plus dot" ~ What does this mean?

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -351,4 +356,38 @@ public static String sanitizeNullAndLowercaseAndTrim(final String input) {
...
+     *
+     * @param source the string to encode.
+     * @return the decoded
```
shahzaibj, line 364:
```suggestion
     * @return the decoded String
```

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -351,4 +356,38 @@ public static String sanitizeNullAndLowercaseAndTrim(final String input) {
...
+    }
+
+    /**
+     * Joins the string in the given set into a single string with the provided delimiter.
+     *
+     * @param delimiter     the delimiter string.
+     * @param stringsToJoin set of strings to join.
+     */
+    public static String join(@NonNull final String delimiter,
+                              @NonNull final Set<String> stringsToJoin) {
```
shahzaibj, line 378: If we move to java 8 then we can just do String.join(delimeter, String args...)

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: @shahzaibj I think it meant "space will be replaced by '+'". dot is just a fullstop.

iambmelt (approved):
I don't see any issues

PR #1519, opened 2021-08-13
Migrate CommandDispatcher

rpdome:
## Changes
1. Migrate prerequisite classes 
    - BaseCommand & co.
    - LibraryConfiguration
    - CommandResult & CommandResultCache

2. Migrate CommandDispatcher
   - Add `IPlatformUtil.postCommandResult()` to let the platform decide how it would post the result back.
       - In android, we're using Handler/Looper to post the result back on main thread.
   - change `CommandDispatcher.eligibleToCacheException()` to `BaseException.isCacheable()`
       - Let each exception decide if it can/should be caught.
       - This is primarily because `BrokerCommunicationException` doesn't belong in 4j.

## Tests
- Perform an acquireToken smoke test on MSAL (brokered + non-brokered auth)
- Didn't test much on caching, since it's disabled (was reminded the hard way ... after spending time trying to re-enable the test..)


shahzaibj (approved):
LGTM

AdamBJohnsonx (approved):

PR #1524, opened 2021-08-17
Delete get storage file name method from IMultiTypeNameValueStorage

shahzaibj:
Per @rpdome's comments on https://github.com/AzureAD/ad-accounts-for-android/pull/1667 it seems having this `getStorageFileName` method on this interface creates a lot of confusion because unique ways need to be developed to try to implement this, so deleting it. It was only being used in a test so it was useless anyway.


rpdome (approved):

shahzaibj:
Marking as no-changelog as the interface never shipped to production yet

AdamBJohnsonx (approved):

iambmelt (approved):

PR #1529, opened 2021-08-19
Wainaina/update common4j version

wainaina:
Small PR to basically update the common4j version.

### DIFF
--- a/common4j/build.gradle
+++ b/common4j/build.gradle
@@ -33,7 +33,6 @@ repositories {
 }
 
 apply from: './versioning/version_tasks.gradle'
-apply from: '../../gradle/versions.gradle'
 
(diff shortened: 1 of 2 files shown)


common4j/build.gradle
```diff
@@ -33,6 +33,7 @@ repositories {
 }
 
 apply from: './versioning/version_tasks.gradle'
+apply from: '../../gradle/versions.gradle'
```
shahzaibj: This seems interesting, because I think it was working without it earlier

shahzaibj: looks like you removed it already

shahzaibj (approved):

common4j/build.gradle
wainaina: Yes just removed it. I just figured that it was resolving a small bug I had on dev environment.

iambmelt (approved):

PR #1540, opened 2021-08-25
Fix authorization activity selection to address available user agent on device

shoatman:
Addresses: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1449628

- In the case that no browser is available we don't need to use the CurrentTaskAuthorizationActivity as embeeded webview already works in that way.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
@@ -78,6 +78,10 @@ public class AuthorizationActivityFactory {
         if (ProcessUtil.isBrokerProcess(context)) {
             intent = new Intent(context, BrokerAuthorizationActivity.class);
         } else if (libraryConfig.isAuthorizationInCurrentTask() && !authorizationAgent.equals(AuthorizationAgent.WEBVIEW)) {
+        // We exclude the case when the authorization agent is already selected as WEBVIEW because of confusion
+        // that results from attempting to use the CurrentTaskAuthorizationActivity in that case, because as webview
+        // already uses the current task, attempting to manually simulate that behavior ends up supplying an incorrect
+        // Fragment to the activity.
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -77,7 +77,7 @@ public static Intent getAuthorizationActivityIntent(final Context context,
         final LibraryConfiguration libraryConfig = LibraryConfiguration.getInstance();
         if (ProcessUtil.isBrokerProcess(context)) {
             intent = new Intent(context, BrokerAuthorizationActivity.class);
-        } else if (libraryConfig.isAuthorizationInCurrentTask()) {
+        } else if (libraryConfig.isAuthorizationInCurrentTask() && !authorizationAgent.equals(AuthorizationAgent.WEBVIEW)) {
```
AdamBJohnsonx, line 80: A comment here in the code explaining why this is the case would be super-helpful.  It'll be almost impossible to reverse-engineer this from the code.

shahzaibj: +1

It is not clear how this change solves that crash. It probably does avoid the crash altogether by not even taking that code path, but I'm thinking if we can still take this code path in a safe way and not crash. Do we know what actually caused the crash?

shoatman: Yes.  The crash happens when the code attempts to use the default browser, but none is available on the device.  In that situation we end up providing the incorrect fragment to the authorization activity... which throws illegal state exception.  This code ensures that if we have to fallback to webview... because no browser is installed.... then we'll use the correct authorization activity and fragment.

shahzaibj (approved):

AdamBJohnsonx (approved):
Just a suggestion

PR #1548, opened 2021-08-30
Make changes to accommondate broker4j getAccounts

rpdome:
https://github.com/AzureAD/ad-accounts-for-android/pull/1690

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
@@ -61,6 +61,14 @@ public interface IPlatformComponents {
     IDevicePopManager getDefaultDevicePopManager() throws ClientException;
 
     /**
+     * Gets a {@link IDevicePopManager} associated to the alias.
+     *
+     * @throws ClientException if it fails to initalize, or if the operation is not supported by the platform.
+     */
+    @NonNull
+    IDevicePopManager getDevicePopManager(@Nullable final String alias) throws ClientException;
+
+    /**
      * Retrieve a name-value store with a given identifier.
      *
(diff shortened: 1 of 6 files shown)


shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
```diff
@@ -60,6 +60,14 @@
     @NonNull
     IDevicePopManager getDefaultDevicePopManager() throws ClientException;
 
+    /**
+     * Gets a {@link IDevicePopManager} associated to the alias.
+     *
+     * @throws ClientException if it fails to initalize, or if the operation is not supported by the platform.
+     */
+    @NonNull
+    IDevicePopManager getDevicePopManager(@Nullable final String alias) throws ClientException;
```
AdamBJohnsonx, line 69: We probably don't want an IDevicePopManager here.

rpdome: 1. I moved it up from below.
2. If you meant this class doesn't belong to this object, you may be right. That said, we decided that we won't be touching the PoP layer in P0. Once we are ready to support PoP in Linux broker, we'll come back here and actually figure out what, in this class, actually belongs to IPlatformComponents.

AdamBJohnsonx: I just didn't find a usage.

AdamBJohnsonx: And, OK, I see it now.  Which is why I didn't see the usage.  Nevermind. :headdesk

AdamBJohnsonx (approved):

PR #1549, opened 2021-08-30
Log additional messages, swallow and log failures to unbind.

AdamBJohnsonx:
https://portal.microsofticm.com/imp/v3/incidents/details/257678975/home

The customer in this incident is seeing a crash due to multiple unbind attempts. At this point, I think swallowing and logging the error is preferable, since it probably shouldn't crash their application, and I don't see a path to actually cause this.  I'm also injecting a couple of logging statements in the CustomTabsActivity callback to alert us to potential issues.

At this point, however, I have been unable to reproduce this crash. Creating this review off of master, we may have to cherry-pick it into dev if we want it.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
@@ -78,6 +78,20 @@ public class CustomTabsManager {
             mCustomTabsClient.set(null);
             mClientLatch.countDown();
         }
+
+        @Override
+        public void onBindingDied(final ComponentName name) {
+            Logger.warn(TAG, "Binding died callback on custom tabs service, there will likely be failures. " +
+                    " Component class that failed: " + ((name == null) ? "null" : name.getClassName()));
+            super.onBindingDied(name);
+        }
+
+        @Override
+        public void onNullBinding(final ComponentName name) {
+            Logger.warn(TAG, "Null binding callback on custom tabs service, there will likely be failures."
+                    + " Component class that failed: " + ((name == null) ? "null" : name.getClassName()));
+            super.onNullBinding(name);
+        }
     };
 
     public CustomTabsIntent getCustomTabsIntent() {
@@ -160,8 +174,16 @@ public class CustomTabsManager {
      * Method to unbind custom tabs service {@link androidx.browser.customtabs.CustomTabsService}.
      */
     public synchronized void unbind() {
-        if (mContextRef.get() != null && mCustomTabsServiceIsBound) {
-            mContextRef.get().unbindService(mCustomTabsServiceConnection);
+        final Context context = mContextRef.get();
+        if (context != null && mCustomTabsServiceIsBound) {
+            try {
+                context.unbindService(mCustomTabsServiceConnection);
+            } catch(final Exception e) {
+                Logger.warn(TAG, "Error unbinding custom tabs service, likely failed to bind or previously died: " + e.getMessage());
+                if (e instanceof InterruptedException) {
+                    Thread.currentThread().interrupt();
+                }
+            }
         }
 
         mCustomTabsServiceIsBound = false;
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
```diff
@@ -160,8 +174,13 @@ public CustomTabsClient getClient() {
      * Method to unbind custom tabs service {@link androidx.browser.customtabs.CustomTabsService}.
      */
     public synchronized void unbind() {
-        if (mContextRef.get() != null && mCustomTabsServiceIsBound) {
-            mContextRef.get().unbindService(mCustomTabsServiceConnection);
+        final Context context = mContextRef.get();
+        if (context != null && mCustomTabsServiceIsBound) {
+            try {
+                context.unbindService(mCustomTabsServiceConnection);
```
shoatman, line 180: The documentation doesn't indicate that "unbindService" throws an exception.

shoatman (changes requested):

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
AdamBJohnsonx: The documentation might not indicate it.  Two things:

1) Runtime exceptions aren't checked exceptions, and can be thrown from anywhere without appearing in the method signature.  IllegalArgumentException is a runtime exception.

2) This is clearly occurring:

8-06 18:39:00.814 10299 27760 27760 E AndroidRuntime: 	at android.app.LoadedApk.forgetServiceDispatcher(LoadedApk.java:1901)
08-06 18:39:00.814 10299 27760 27760 E AndroidRuntime: 	at android.app.ContextImpl.unbindService(ContextImpl.java:1954)
08-06 18:39:00.814 10299 27760 27760 E AndroidRuntime: 	at android.content.ContextWrapper.unbindService(ContextWrapper.java:810)
08-06 18:39:00.814 10299 27760 27760 E AndroidRuntime: 	at com.microsoft.identity.common.internal.ui.browser.CustomTabsManager.unbind(CustomTabsManager.java:2)
08-06 18:39:00.814 10299 27760 27760 E AndroidRuntime: 	at com.microsoft.identity.common.internal.ui.browser.BrowserAuthorizationStrategy.dispose(BrowserAuthorizationStrategy.java:3)

3) In the context implementation this throws here:

https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/LoadedApk.java;drc=875888eff38e3a6c1369bbb9c2336e7b97758c51;l=1793

Given this, do you have a suggestion?

shoatman:
Feels like something strange is happening here inter of service connection lifecycle.  It's like the connection is being closed before it's actually bound/connected.... i found similar exceptions when things happen out of order.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
shoatman: Thanks for writeup and offline discussion.  Approving.  Thanks

shoatman (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/CustomTabsManager.java
AdamBJohnsonx: > Feels like something strange is happening here inter of service connection lifecycle. It's like the connection is being closed before it's actually bound/connected.... i found similar exceptions when things happen out of order.



> Feels like something strange is happening here inter of service connection lifecycle. It's like the connection is being closed before it's actually bound/connected.... i found similar exceptions when things happen out of order.

I agree.  There's something fishy here.  Ideas for either logging to catch the issue or methods to reproduce are welcome.

PR #1550, opened 2021-08-30
Add method to clear the singleton cache for shared prefs

AdamBJohnsonx:
Another group has observed that post-account-data migration office seems to have difficulty getting accounts until an app restart.
Experimentally, performing a clear on the singleton cache resolves that issue.  This PR just implements that method.

https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1533351

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -80,7 +80,8 @@ public class SharedPreferencesFileManager implements ISharedPreferencesFileManag
                                                                     final String name,
                                                                     final int operatingMode,
                                                                     final IStorageHelper storageHelper) {
-        String key = name + "/" + context.getPackageName() + "/" + (operatingMode == -1 ? Context.MODE_PRIVATE : operatingMode);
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -91,6 +91,14 @@ public static SharedPreferencesFileManager getSharedPreferences(final Context co
         return cachedFileManager;
     }
 
+    /**
+     * This method clears the singleton cache in use by the system, in the case that unsafe operations
+     * have been performed on disk and the actual data needs to be removed.
+     */
+    public static void clearSingletonCache() {
```
shahzaibj, line 98: Where does this get called?

matthewyu-ms: This will get called from OneAuth side after migrateAdalCacheKey is called. I tested this locally with my Office local enlistment and clearing the objectCache seems to fix the issue of getting accounts.

matthewyu-ms (approved):

shahzaibj (approved):

shoatman (approved):

PR #1553, opened 2021-09-01
Add more operation names to ArgumentException

shahzaibj:
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1693

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
@@ -22,8 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.exception;
 
-import com.microsoft.identity.common.java.exception.BaseException;
-
 public class ArgumentException extends BaseException {
 
     public static final String sName =  ArgumentException.class.getName();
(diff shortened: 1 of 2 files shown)


rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
```diff
@@ -32,6 +30,10 @@
     public final static String ACQUIRE_TOKEN_OPERATION_NAME = "acquireToken";
     public final static String ACQUIRE_TOKEN_SILENT_OPERATION_NAME = "acquireTokenSilent";
     public final static String BROKER_TOKEN_REQUEST_OPERATION_NAME = "brokerTokenRequest";
+    public final static String GET_ACCOUNTS_OPERATION_NAME = "getAllAccounts";
```
AdamBJohnsonx, line 33: Can we do something else, so that we're not curating a separate list of these?

AdamBJohnsonx (approved):
There's nothing broken here... but do we need to curate a separate list of constants for this, or can ewe derive it from something.

common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
shahzaibj: I think this is fine for now.

PR #1557, opened 2021-09-03
Fix extracting http response body

shahzaibj:
The response from server looks like this:

```json
{
  "error": "some-string",
  "error_description": "some-string",
  "error_codes": [700003],
  "timestamp": "2021-09-03 19:50:35Z",
  "trace_id": "some-string",
  "correlation_id": "some-string",
  "error_uri": "some-string",
  "suberror": "some-string"
}
```

So the **error_codes** field is NOT a String, and with the existing code it throws a JSON Exception:

**JSONException("JSONObject[" + quote(key) + "] not a string.");**

So I've updated the code to accommodate any type that isn't a String. 

I don't know why we're running into this now but I suspect may be something got changed somewhere as part of the refactoring, but it isn't clear what changed. At least the code in this file seemed to be the same even prior to refactoring.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
@@ -43,10 +43,10 @@ public class HashMapExtensions {
         final HashMap<String, String> response = new HashMap<>();
         if (webResponse != null && !StringUtil.isNullOrEmpty(webResponse.getBody())) {
             final JSONObject jsonObject = new JSONObject(webResponse.getBody());
-            final Iterator<?> i = jsonObject.keys();
-            while (i.hasNext()) {
-                String key = (String) i.next();
-                response.put(key, jsonObject.getString(key));
+            final Iterator<String> keyIterator = jsonObject.keys();
+            while (keyIterator.hasNext()) {
+                String key = keyIterator.next();
+                response.put(key, jsonObject.get(key).toString());
             }
         }
         return response;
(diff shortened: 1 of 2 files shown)


common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
```diff
@@ -43,10 +43,10 @@
...
         if (webResponse != null && !StringUtil.isNullOrEmpty(webResponse.getBody())) {
             final JSONObject jsonObject = new JSONObject(webResponse.getBody());
-            final Iterator<?> i = jsonObject.keys();
-            while (i.hasNext()) {
-                String key = (String) i.next();
-                response.put(key, jsonObject.getString(key));
+            final Iterator<String> keyIterator = jsonObject.keys();
+            while (keyIterator.hasNext()) {
+                String key = keyIterator.next();
+                response.put(key, jsonObject.get(key).toString());
```
AdamBJohnsonx, line 49: It could have been a json lib change that brought this on.  Did the server schema change to become error_codes:list instead of error_code:string?  Can we take on schema validation here?

AdamBJohnsonx (approved):
Nothing wrong with this, but it would be nice if the response from whatever server this is indicated what's in the object.

common4j/src/main/com/microsoft/identity/common/java/util/HashMapExtensions.java
AdamBJohnsonx: What call is being made that is interpreted in this way?

shahzaibj: Getting an AT in exchange for a PRT

shahzaibj: Our own tests with mocked responses indicate that this was always an array on the server: https://github.com/AzureAD/ad-accounts-for-android/blob/dev/AADAuthenticator/src/sharedTestClasses/com/microsoft/broker/sharedtestclasses/TestUtils.java#L397

rpdome (approved):

PR #1560, opened 2021-09-07
Accommodate changes to key interface refactoring

rpdome:
ref: https://github.com/AzureAD/ad-accounts-for-android/pull/1700

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
@@ -20,54 +20,33 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-package com.microsoft.identity.common.java.challengehandlers;
+package com.microsoft.identity.common.java.crypto;
 
+import com.microsoft.identity.common.java.exception.ClientException;
+
+import java.security.Key;
 import java.security.PrivateKey;
-import java.security.PublicKey;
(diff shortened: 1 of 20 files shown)


common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Decrypt the given byte array.
+     *
+     * @param key               the key to decrypt with.
+     * @param decryptAlgorithm  algorithm to decrypt with.
+     * @param iv                an initialization vector (IV).
+     * @param dataToBeDecrypted the data to be encrypted.
+     * @return a decrypted byte array.
+     */
+    byte[] decrypt(@NonNull final Key key,
```
shahzaibj, line 48: I think this is the 3rd place we have a decrypt method on an interface

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -231,7 +231,7 @@ public static String getStackTraceAsString(@NonNull final Exception exception) {
         return pw.toString();
     }
 
-    public static String fromByteArray(@NonNull final byte[] bytes) {
```
shahzaibj, line 234: So Nullable?

shahzaibj: what happens if this is NULL?

shahzaibj (approved):
Left minor comments, but LGTM

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
rpdome: a warning is shown if we put NonNull here, so I removed it.

AdamBJohnsonx (approved):

PR #1562, opened 2021-09-09
Create space in CommandParameters for flight information

AdamBJohnsonx:
In order add flight info to all broker commands efficiently, the only safe place to add them is in the base class.  So we create a new transient field for flight parameters, and then don't mention it in common anymore.  Additionally, the mechanism for sharing shared preferences relies on a different mode, and we absolutely cannot cache them, so create a new vending mechanism for a shared preferences file with MODE_MULTI_PROCESS and use it to supply the flight manager.

To me clear here, the purpose of using MODE_MULTI_PROCESS is so that we have a chance to read the flight data that may be written by another process at any time.  According to the documentation on shared preferences, android will attempt to cache the data.

I have validated that this is true.  If I use MODE_PRIVATE in a separate process, I can't read writes from the other process.  Additionally, if I use our home-grown shared pref manager, I can't read the writes from another process.


common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
```diff
@@ -102,6 +102,8 @@
      */
     IMultiTypeNameValueStorage getFileStore(String storeName);
 
+    INameValueStorage<String> getMultiProcessStringStore(@NonNull String storeName);
```
shahzaibj, line 105: Remind me..why do we need it to be shared among process?

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -242,6 +246,64 @@ public IMultiTypeNameValueStorage getFileStore(final @NonNull String storeName)
...
+            @Override
+            public void clear() {
+                sharedPreferences.edit().clear().commit();
+            }
+
+            @Override
+            public void remove(String key) {
+                sharedPreferences.edit().remove(key).commit();
+            }
+        });
```
shahzaibj, line 304: I believe we already have implementations of SharedPreferences for INameValueStorage, can't we just use those here? (just with a different operating mode)

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
AdamBJohnsonx: Because it's written by Authenticator and read in the :auth process.  This isn't actually great - I'm asking them to just go ahead and code the call to setFlights, and we can ditch this in that release.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
AdamBJohnsonx: Let me see if I can find it.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
AdamBJohnsonx: Also, this mode is deprecated.  Seems that multiprocess writing to shared preferences is unsafe.  There's a jetpack library called DataStorage that does this better.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
AdamBJohnsonx: The SharedPreferencesFileManager is a bit of a Swiss-army-knife right now.  The main thing it does, I have to defeat - I absolutely must create new instances of this every time I read, or we risk not having cross-process visibility.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -242,6 +246,64 @@ public IMultiTypeNameValueStorage getFileStore(final @NonNull String storeName)
         return SharedPreferencesFileManager.getSharedPreferences(mContext, storeName, null);
     }
 
+    @Override
+    public INameValueStorage<String> getMultiProcessStringStore(final @NonNull String storeName) {
+        final SharedPreferences sharedPreferences = mContext.getSharedPreferences(storeName, Context.MODE_MULTI_PROCESS);
```
shoatman, line 251: We don't need the checking that's performed for some versions of Android configured by this flag.  We'll only be writing to this file from 1 process.  Rather than using this deprecated flag... let's just add a warning/comment to this code that writing to shared preferences from more than one process is not a good idea.  A problem exclusive to the broker for our work to date.

Adding link to docs: [Context.MODE_MULTI_PROCESS](https://developer.android.com/reference/android/content/Context#MODE_MULTI_PROCESS)

shoatman: If we don't use this.  I assume we can use the existing implementation.

shoatman (changes requested):
Let's just go without the flag that apparently doesn't work everywhere.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
AdamBJohnsonx: We cannot.

AdamBJohnsonx: " the file on disk will be checked for modification even if the shared preferences instance is already loaded in this process" - this is precisely what we will need.

shoatman: Have we tested to confirm that there is a problem here?  In other words... have we tested that process B is unable to read what process A has written?  I don't want to use a flag which is deprecated, because they discovered it doesn't work as intended...

shoatman: NOTE: Is the issue here about how to notify/update the file at runtime?

shoatman: I believe a cheap and dirty solution would be use an intent to launch an activity associated with the :auth process.  That activity doesn't have do have an UI per se....

AdamBJohnsonx: The issue is that the Authenticator process will write the file, and then it will be read in the auth process.  Authenticator will be updating on a daily cadence, so there will be writes at runtime.

AdamBJohnsonx: I've altered the test app to add a button that manipulates the flight manager in the android process instead of calling through the IPC api in order to see the results.  I put the change in the broker code review.  By altering what I was doing, I can establish that I do not read the writes of the other process unless I set the value to MODE_MULTI_PROCESS once I read.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
pablobaxter: As an FYI, DataStore is not process-safe either. As for `SharedPreferences`, the reason that the `MODE_MULTI_PROCESS` became unsupported is due to the risk of corrupting the preference file if multiple processes attempt to write to the prefs file. Having a single process write to the `SharedPreferences` while another reads it using `MODE_MULTI_PROCESS` _should_ work, but you won't be notified of updates, and you have to continuously call `getSharedPreferences()` to have the underlying cache refresh.

I don't mean this to be a plug (I was just browsing through this library), but I have written a process-safe shared preference library that doesn't use ContentProviders or any other IPC Android class. It's completely file based. My approach was to use the filesystem as you would normally (files are atomic and available to all processes), and listen for changes to the file using `FileObserver`. This is the approach I took with Harmony (https://github.com/pablobaxter/Harmony). I'm more than happy to answer any questions about the approach I took with Harmony if you folks are looking for a similar approach, or questions about Harmony altogether.

shoatman (approved):
Discussed offline.  approving.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
AdamBJohnsonx: We're taking this ability as a short-term measure while we move to an IPC model for a particular piece of information in a derivative product.  As you correctly infer, only one process will actually write this file.  We could engineer a filesystem implementation, it just seems overly complex for what we're using it for.

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
pablobaxter: Yea, IPC on Android (and in general) isn't all that easy, and the constructs Android provides assumes both processes need to be running for data going back and forth, which can be impactful to the system (especially on battery). If you want a true short-term approach, I would suggest having your `SharedPrefStringNameValueStorage` call `getSharedPreferences()` instead of caching the shared prefs object. Otherwise, data won't update if you continue using the same `SharedPrefStringNameValueStorage`.

PR #1568, opened 2021-09-20
Handle the scenario where broker activity is killed wrongly.

rpdome:
Scenario:
1. MSAL app launches an interactive request.
2. A broker activity (AccountChooserActivity) is launched.
3. Somehow, that Activity is prematurely killed. (i.e. you could go to the app switcher and kill it)
4. In the MSAL app, try launching an interactive session again. You won't be able to, because it's still waiting for the result from the previous interactive request.

Fix:
If the BrokerActivity is destroyed BEFORE receiving any result, return a UserCanceledException.

Tests:
1. if the broker process is killed via adb, then onActivityResult is still triggered. We just need to handle the error code = 0.
2. if i kill the activity from the activity switcher, then onActivityResult will not be triggered. in that case, onDestroy logic is triggered.
3. If Broker throws an unhandled exception, same as 2)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
@@ -29,14 +29,13 @@ import android.os.Bundle;
 import com.microsoft.identity.common.PropertyBagUtil;
 import com.microsoft.identity.common.internal.result.BrokerResultAdapterFactory;
 import com.microsoft.identity.common.internal.result.IBrokerResultAdapter;
+import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.exception.ErrorStrings;
-import com.microsoft.identity.common.java.exception.UserCancelException;
 import com.microsoft.identity.common.java.request.SdkType;
 import com.microsoft.identity.common.java.util.ported.PropertyBag;
 import com.microsoft.identity.common.java.util.ported.LocalBroadcaster;
 import com.microsoft.identity.common.logging.Logger;
 
-import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
```diff
@@ -80,6 +87,26 @@ protected void onStop() {
         super.onStop();
     }
 
+    @Override
+    protected void onDestroy() {
+        // If result hasn't been received. It means that this activity is prematurely killed.
+        // In this case, we want to properly cancel the activity so that it doesn't block MSAL from making a subsequent interactive call.
```
AdamBJohnsonx, line 93: Don't we broadcast a message that is supposed to cancel the activity already in the broker?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
```diff
@@ -80,6 +87,26 @@ protected void onStop() {
...
+        if (!mBrokerResultReceived) {
+            final IBrokerResultAdapter resultAdapter = BrokerResultAdapterFactory.getBrokerResultAdapter(SdkType.MSAL);
+            final Bundle resultBundle = resultAdapter.bundleFromBaseException(
+                    new UserCancelException(ErrorStrings.USER_CANCELLED,
+                            "Activity is killed by user"), null);
```
AdamBJohnsonx, line 98: Does it require a user to kill it, or can android or the OOM killer take it out randomly?

rpdome: In the IcM, the team does not know who the killer is.

I repro this by killing it myself though. 

Maybe I should change the error code.

AdamBJohnsonx (approved):
I think there's only one comment I have, and it's about an edge case.

shahzaibj (approved):

shoatman (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
rpdome: We do in the broker process. This is on MSAL side.

PR #1569, opened 2021-09-20
Handle the scenario where broker activity is killed wrongly

rpdome:
The pre-broker-refactored version of  https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1568

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
@@ -29,6 +29,11 @@ import android.os.Bundle;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.exception.ClientException;
+import com.microsoft.identity.common.exception.ErrorStrings;
+import com.microsoft.identity.common.internal.request.SdkType;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
```diff
@@ -112,9 +146,11 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
             data.putExtra(RESULT_CODE, resultCode);
 
             LocalBroadcastManager.getInstance(getApplicationContext()).sendBroadcast(data);
+        } else {
+            // This means the broker is unexpectedly killed. (tested by killing the broker process via adb).
+            returnsExceptionOnActivityUnexpectedlyKilled();
```
shahzaibj, line 151: but the result is still received? I see it being set to true above

rpdome: yes. this function will be triggered, but the result code will be 0, and the intent will be null.

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
shahzaibj: I suppose it just depends on when it is unexpectedly killed. I would think there would be cases where it is unexpectedly killed and also a response is NOT received.

shahzaibj (approved):

PR #1579, opened 2021-09-30
Changes to handle IntuneAppProtectionException in common rather than …

shoatman:
Migrated logic in broker library to the common library to ensure that errors related to Intune App Protection are handled in both joined and non-joined scenarios.  

NOTE: I don't plan to commit this change right away; however I will be creating builds out of this branch to provide to partners and customers for testing.  Your feedback is very welcome.


common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
-                            errorResponse.getResponseHeadersJson(),
-                            errorResponse.getResponseBody()
-                    )
+
+    public static ServiceException getExceptionFromTokenErrorResponse(@Nullable final CommandParameters commandParameters,
+                                                                      @NonNull final TokenErrorResponse errorResponse) {
+
+        if(isIntunePolicyRequiredError(errorResponse)){
+            if(commandParameters == null || !(commandParameters instanceof BrokerSilentTokenCommandParameters)){
+                Logger.warn(TAG, "In order to properly construct the IntuneAppProtectionPolicyRequiredException we need the command parameters to be supplied.  Returning as service exception instead.");
```
AdamBJohnsonx, line 211: Just WARN?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+        }
+
+        com.microsoft.identity.common.internal.logging.Logger.warn(TAG + methodName,
+                "Home Account id doesn't have uid or tenant id information, returning null "
+        );
+
+        return null;
+    }
+
+    private static void setIntuneExceptionProperties(
```
AdamBJohnsonx, line 299: Could we move this to a specific constructor for IntuneAppProtectionPolicyException instead?

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
shoatman: This shouldn't really happen... I suppose this is better as an error though... not ready to commit to throwing a full blow exception here though.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
shoatman: We can later.  ;-)

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+
+        if (homeAccountId != null) {
+            final Pair<String, String> tenantInfo = StringUtil.getTenantInfo(homeAccountId);
+            tenantId = tenantInfo.second;
+        }
+
+        if (TextUtils.isEmpty(tenantId) && authority instanceof AzureActiveDirectoryAuthority) {
+            tenantId = ((AzureActiveDirectoryAuthority) authority).mAudience.getTenantId();
+
+        }
```
AdamBJohnsonx, line 335: Since it looks like we can get the tenant id here from multiple different places, would logging where we got it help solve any problems?

AdamBJohnsonx (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
AdamBJohnsonx: Yeah.  Just something to draw somebody's eye when they're looking at logs.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
shoatman: No immediate problems related to what we're trying to fix here, but we can certainly review this all up.  I think it's worthwhile to do a code walkthrough in our weekly meeting to identify similar potentially pervasive issues.  I feel like there was some WPJ code still hanging around in here.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
AdamBJohnsonx: Since it means that whatever they were trying to do is about to fail, horribly.

shahzaibj:
What is the outcome of this change? I think this causes the apps to call into MAM and then it will make the device mam complaint, right? but what happens after that? I suspect the app will be caught into the same loop i.e. eSTS will keep giving back the same error? I'm asking because AFAIK, we only attach enrollment id in joined scenario - I don't know why. Do you know? But since we do it only in joined, that means non-joined accounts cannot send the enrollment id so possibly they get into a loop here. I don't know what the current behavior / impact is to the apps for non-joined cases.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
 
-        try {
-            outErr.setHttpResponse(
-                    synthesizeHttpResponse(
-                            errorResponse.getStatusCode(),
-                            errorResponse.getResponseHeadersJson(),
-                            errorResponse.getResponseBody()
-                    )
+
+    public static ServiceException getExceptionFromTokenErrorResponse(@Nullable final CommandParameters commandParameters,
```
rpdome, line 206: nit: it might be cleaner to rename this to "getIntunePolicyRequiredException", and in the original getExceptionFromTokenErrorResponse
- invoke this function
- if it doesn't return anything, proceed with the original logic.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+    /**
+     * Helper method to get uid from home account id
+     * V2 home account format : <uid>.<utid>
+     * V1 : it's stored as <uid>
+     *
+     * @param homeAccountId
+     * @return valid uid or null if it's not in either of the format.
+     */
+    @Nullable
+    public static String getUIdFromHomeAccountId(@Nullable String homeAccountId) {
```
rpdome, line 266: fyi: in the dev branch, this function is moved to BrokerUtil.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +198,143 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+        );
+
+        return null;
+    }
+
+    private static void setIntuneExceptionProperties(
+            @NonNull final IntuneAppProtectionPolicyRequiredException exception,
+            @NonNull final BrokerSilentTokenCommandParameters requestParameters) {
+
+        com.microsoft.identity.common.internal.logging.Logger.info(TAG, "Setting properties to IntuneAppProtectionPolicyRequiredException ");
```
rpdome, line 303: nit: remove extra `com.microsoft.identity.common.internal.logging` import. Here and below.

rpdome (approved):
Left some minor nit suggestions, otherwise LGTM.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+
+        if (homeAccountId != null) {
+            final Pair<String, String> tenantInfo = StringUtil.getTenantInfo(homeAccountId);
+            tenantId = tenantInfo.second;
+        }
+
+        if (TextUtils.isEmpty(tenantId) && authority instanceof AzureActiveDirectoryAuthority) {
+            tenantId = ((AzureActiveDirectoryAuthority) authority).mAudience.getTenantId();
+
+        }
```
shahzaibj, line 99:
```suggestion
        }
        
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +201,70 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+    public static ServiceException getExceptionFromTokenErrorResponse(@Nullable final CommandParameters commandParameters,
+                                                                      @NonNull final TokenErrorResponse errorResponse) {
+
+        if(isIntunePolicyRequiredError(errorResponse)){
+            if(commandParameters == null || !(isBrokerTokenCommandParameters(commandParameters))){
+                Logger.warn(TAG, "In order to properly construct the IntuneAppProtectionPolicyRequiredException we need the command parameters to be supplied.  Returning as service exception instead.");
+                return getExceptionFromTokenErrorResponse(errorResponse);
+            }
+            IntuneAppProtectionPolicyRequiredException policyRequiredException;
+            if(commandParameters instanceof BrokerInteractiveTokenCommandParameters) {
```
AdamBJohnsonx, line 218:
```suggestion
            if (commandParameters instanceof BrokerInteractiveTokenCommandParameters) {
```

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+                originalParameters.getAccountManagerAccount().name :
+                originalParameters.getLoginHint();
+
+        setAccountUpn(upn);
+
+        String uId = originalParameters.getLocalAccountId();
+
+        if (TextUtils.isEmpty(uId)) {
+            com.microsoft.identity.common.internal.logging.Logger.info(TAG, "Local account id is empty, attempting get user id from home account id");
+            uId = getUIdFromHomeAccountId(
```
AdamBJohnsonx, line 78: Log if you failed.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+     * V2 home account format : <uid>.<utid>
+     * V1 : it's stored as <uid>
+     *
+     * @param homeAccountId
+     * @return valid uid or null if it's not in either of the format.
+     */
+    @Nullable
+    private String getUIdFromHomeAccountId(@Nullable String homeAccountId) {
+        //TODO: This method is from BrokerOperationParameterUtils...
+        // seems like this is not broker specific per se and should move to somewhere better
```
shahzaibj, line 154: seems like this todo is now resolved? Do we want it in this class though? Seems like it needs to become `public` and in go into some util class

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+    public IntuneAppProtectionPolicyRequiredException(final String errorCode,
+                                                      final String errorMessage,
+                                                      final BrokerSilentTokenCommandParameters originalParameters) {
+        super(errorCode, errorMessage, null);
+
+        final String upn = (originalParameters.getAccountManagerAccount() != null) ?
+                originalParameters.getAccountManagerAccount().name :
+                originalParameters.getLoginHint();
+
+        setAccountUpn(upn);
```
AdamBJohnsonx, line 112: Let us know if it was empty or null.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+
+        String uId = originalParameters.getLocalAccountId();
```
AdamBJohnsonx, line 114: Let us know I it was empty or null.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+        String tenantId = null;
+
+        if (homeAccountId != null) {
+            final Pair<String, String> tenantInfo = StringUtil.getTenantInfo(homeAccountId);
+            tenantId = tenantInfo.second;
+        }
+
+        if (TextUtils.isEmpty(tenantId) && authority instanceof AzureActiveDirectoryAuthority) {
+            tenantId = ((AzureActiveDirectoryAuthority) authority).mAudience.getTenantId();
+
```
AdamBJohnsonx, line 138: If it is possible for tenantId toned up being empty, log that.

In the interest brevity, maybe hold all these logging tasks until the end, and log a single-line summary of what is not available.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+        }
+        setTenantId(tenantId);
+    }
+
+    public IntuneAppProtectionPolicyRequiredException(final String errorCode,
```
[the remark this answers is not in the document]
AdamBJohnsonx: The difference is that BrokerInteractiveTokenCommandParameters and BrokerSilentTokenCommandParameters are sibling classes.  There's no convenient way to address either of them in a method.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
shoatman: Done.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
shoatman: I ran the code formatter in Android Studio.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +201,70 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+                policyRequiredException = new IntuneAppProtectionPolicyRequiredException(
+                        errorResponse.getError(),
+                        errorResponse.getErrorDescription(),
+                        (BrokerInteractiveTokenCommandParameters) commandParameters
+                        );
+            }else{
```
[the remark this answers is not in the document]
shoatman: I ran the code formatter in Android Studio.

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +201,70 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+        setHttpResponseUsingTokenErrorResponse(outErr, errorResponse);
+        return outErr;
+    }
+
+
+    public static ServiceException getExceptionFromTokenErrorResponse(@Nullable final CommandParameters commandParameters,
+                                                                      @NonNull final TokenErrorResponse errorResponse) {
+
+        if(isIntunePolicyRequiredError(errorResponse)){
+            if(commandParameters == null || !(isBrokerTokenCommandParameters(commandParameters))){
```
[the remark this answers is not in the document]
shoatman: I ran the code formatter in Android Studio.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+                                                      final String errorMessage,
+                                                      final BrokerInteractiveTokenCommandParameters originalParameters) {
+        super(errorCode, errorMessage, null);
+
+        final String upn = (originalParameters.getAccountManagerAccount() != null) ?
+                originalParameters.getAccountManagerAccount().name :
+                originalParameters.getLoginHint();
+
+        setAccountUpn(upn);
```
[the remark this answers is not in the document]
shoatman: Done.

common/src/main/java/com/microsoft/identity/common/exception/IntuneAppProtectionPolicyRequiredException.java
```diff
@@ -45,6 +60,128 @@ public IntuneAppProtectionPolicyRequiredException(final String errorCode,
...
+        String tenantId = null;
+
+        if (homeAccountId != null) {
+            final Pair<String, String> tenantInfo = StringUtil.getTenantInfo(homeAccountId);
+            tenantId = tenantInfo.second;
+        }
+
+        if (TextUtils.isEmpty(tenantId) && authority instanceof AzureActiveDirectoryAuthority) {
+            tenantId = ((AzureActiveDirectoryAuthority) authority).mAudience.getTenantId();
+
```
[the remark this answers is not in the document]
shoatman: Done

common/src/main/java/com/microsoft/identity/common/internal/controllers/ExceptionAdapter.java
```diff
@@ -188,24 +201,70 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+                Logger.warn(TAG, "In order to properly construct the IntuneAppProtectionPolicyRequiredException we need the command parameters to be supplied.  Returning as service exception instead.");
+                return getExceptionFromTokenErrorResponse(errorResponse);
```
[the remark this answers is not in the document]
shoatman: I'm going to leave this one for future consideration.... as frankly, I'm not sure in what scenarios for MAM those would be absent.... I need to think about this a bit more.

PR #1581, opened 2021-10-05
Make it clear if adding a query param should overwrite or leave as is

rpdome:
Replace URIBuilder with our CommonURIBuilder,  which will force the caller to either
1. Overwrites query parameter if exists with `setParameter`
2. Only append the query param if there is no matching query parameter (key) with `addParameterIfAbsent`

This is to prevent us from sending duplicate query parameter, which eSTS doesn't like.

calling addParameters, which permits duplicate query params, will throw UnsupportedOperationException.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
@@ -1,0 +1,162 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.util;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Map;
+
+import cz.msebera.android.httpclient.NameValuePair;
+import cz.msebera.android.httpclient.client.utils.URIBuilder;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.NonNull;
+
+/**
+ * Our URIBuilder.
+ * We want to make sure we never send duplicated parameters to the server.
+ * This is done by
+ * 1. disabling {@link cz.msebera.android.httpclient.client.utils.URIBuilder#addParameter(String, String)} and
+ *    {@link cz.msebera.android.httpclient.client.utils.URIBuilder#addParameters(List))}
+ * 2. adding {@link CommonURIBuilder#addParametersIfAbsent}
+ */
+public class CommonURIBuilder extends cz.msebera.android.httpclient.client.utils.URIBuilder {
+
+    public CommonURIBuilder(){
+        super();
+    }
+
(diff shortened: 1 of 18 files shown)


common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Our URIBuilder.
+ * We want to make sure we never send duplicated parameters to the server.
+ * This is done by
+ * 1. disabling {@link cz.msebera.android.httpclient.client.utils.URIBuilder#addParameter(String, String)} and
+ *    {@link cz.msebera.android.httpclient.client.utils.URIBuilder#addParameters(List))}
+ * 2. adding {@link CommonURIBuilder#addParametersIfAbsent}
+ */
+public class CommonURIBuilder extends cz.msebera.android.httpclient.client.utils.URIBuilder {
```
shahzaibj, line 43: I think an alternative to doing this would be to use lombok extensions: https://projectlombok.org/features/experimental/ExtensionMethod

but that's in Experimental -- does seem promising though

common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+    public CommonURIBuilder setParameters(NameValuePair... nvps) {
+        super.setParameters(nvps);
+        return this;
+    }
+
+    @Override
+    public CommonURIBuilder setParameter(String param, String value) {
+        super.setParameter(param, value);
+        return this;
+    }
```
shahzaibj, line 83: seems like we're just calling `super` - do we still need to override these?

rpdome: it's just to return CommonURIBuilder instead of URIBuilder

common4j/src/test/com/microsoft/identity/common/java/util/CommonURIBuilderTest.java
```diff
@@ -0,0 +1,143 @@
...
+        final List<Map.Entry<String, String>> map = new ArrayList<>();
+        map.add(new AbstractMap.SimpleEntry<String, String>("Test1", "Value1"));
+
+        builder.addParametersIfAbsent(map);
+
+        Assert.assertEquals(1, builder.getQueryParams().size());
+        Assert.assertEquals("Test1", builder.getQueryParams().get(0).getName());
+        Assert.assertEquals("Value1", builder.getQueryParams().get(0).getValue());
+    }
+}
```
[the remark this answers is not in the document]
rpdome: I use `setParameters` because that's what URIBuilder provides. (and I assume that URIBuilder works as expected - the test cases here is for testing our code specifically)

In those tests, we want to verify if adding the same param to an existing object would yield the right result. 
Using `addParametersIfAbsent` as an initial add seems to defeat the purpose.

That said, there are tests, such as `testCallingAddParametersIfAbsent` that tests adding `addParametersIfAbsent` into an empty object - if that's what you're looking for.

common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
shahzaibj: but `CommandURIBuilder` is `URIBuilder` -  I guess we would need to cast in that case though. Fine.

shahzaibj (approved):
This looks fine to me. Left a suggestion around using Lombok extensions. That's experimental but I'm personally fine with it if it works. - We don't have to try that out right now though.

common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+     * The parameter name and value are expected to be unescaped and may contain non ASCII characters.
+     * <p>
+     * Please note query parameters and custom query component are mutually exclusive. This method
+     * will remove custom query if present.
+     * </p>
+     *
+     * @param params list of parameters.
+     * @return {@link CommonURIBuilder}
+     */
+    public CommonURIBuilder addParametersIfAbsent(@Nullable final Map<String, ?> params) {
```
AdamBJohnsonx, line 96: Can we make this follow the pattern from the superclass?

AdamBJohnsonx: That is to say

```
   public URIBuilder addParametersIfAbsent(final List <NameValuePair> nvps) {
 ```
and

```
   public URIBuilder addParametersIfAbsent(final NameValuePair... nvps) {
 ```

rpdome: will do. This is an extension from that though.

AdamBJohnsonx: I know... just asking to use the list of NameValuePair instead of Map.

rpdome: @AdamBJohnsonx I'm inclined not to add these two method... because
1. They're not being used. adding methods means more maintainance (adding tests, etc). If we need to use them (which I doubt), then we can always add them later.
2. List <NameValuePair> has the same type erasure as List<Map.Entry<String, String>>

AdamBJohnsonx:
> > Just to be clear, the base class is or is not using NameValuePair for their queryParamters? I really don't want to create a method that doesn't fit into the class we're extending. We might want to upstream this change at some point. I don't think it _matters_ right now.
> 
> The base class supports it, we don't use it in our code.
> 
> (we replaced Android's Pair with Map.Entry)

⛵

PR #1584, opened 2021-10-07
Fix a typo in a class name, leave a breadcrumb for porters

AdamBJohnsonx:
When we changed the name of this constant, we left no path for a user to discover where it had gone.  We also misspelled local, but nobody would ever have known that if I hadn't had to go on a great search for the missing constant.

In general, if you're renaming public constants, it would be nice to leave the previous name in a comment, so that at least people using code search could discover where it went.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
@@ -630,7 +630,7 @@ public class AuthenticationConstants {
         public static final String CANCEL_AUTHORIZATION_REQUEST = "cancel_authorization_request";
 
         /**
-         * an alias that the intent contains authorization results, was formerly RETURN_INTERACTIVE_REQUEST_RESULT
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -622,15 +622,15 @@
...
+    public static final class LocalBroadcasterAliases {
 
         /**
          * an alias specifying that the current authorization action should be cancelled.
          */
         public static final String CANCEL_AUTHORIZATION_REQUEST = "cancel_authorization_request";
 
         /**
-         * an alias that the intent contains authorization results.
+         * an alias that the intent contains authorization results, was formerly RETURN_INTERACTIVE_REQUEST_RESULT
```
shahzaibj, line 633:
```suggestion
         * An alias that the intent contains authorization results, was formerly RETURN_INTERACTIVE_REQUEST_RESULT.
```

rpdome (approved):

PR #1587, opened 2021-10-08
QueryParam Backcompat

rpdome:
Older MSAL/Common serializes broker request request from Pair.

Even though we're ditching Pair already (for common4j - Pair is an android class), we need to be still able to support it.


common/src/main/java/com/microsoft/identity/common/internal/util/QueryParamsAdapter.java
```diff
@@ -40,9 +44,14 @@
...
 /**
  * Class to serialize and deserialize query parameters from List<Pair<String, String>> to json String
- * and vice versa
+ * and vice versa.
+ *
+ * NOTE: Even we no longer use Pair (Since it's android-only), we are keeping this the same
+ *       to maintain backcompat with serialized value from older common that still uses it.
  */
-public class QueryParamsAdapter extends TypeAdapter<List<Map.Entry<String, String>>> {
+public class QueryParamsAdapter extends TypeAdapter<List<Pair<String, String>>> {
```
shahzaibj, line 52: How does this now work with new MSAL? I suppose that's going to send Map.Entry?

common/src/test/java/com/microsoft/identity/common/internal/util/QueryParamsAdapterTest.java
```diff
@@ -0,0 +1,103 @@
...
+    public void testConvertFromTruncatedJson(){
+        final String input = "[{\"key1\"";
+        try {
+            QueryParamsAdapter._fromJson(input);
+            Assert.fail();
+        } catch (final ClientException e){
+            Assert.assertEquals(ClientException.JSON_PARSE_FAILURE, e.getErrorCode());
+        }
+    }
+}
```
shahzaibj, line 103: nit: new line ending

rpdome:
> > @AdamBJohnsonx Unfortunately, [we shipped a bug from the beginning](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blame/master/common/src/main/java/com/microsoft/identity/common/internal/util/QueryParamsAdapter.java#L51) - we're supposed to specify the TypeToken of List or List instead, this means that this adapter is never used, so gson has always used it default adapter from the get go.
> > This means we're stuck with it, unless we want to make a broker-msal protocol breaking change.
> > If we migrate to the right format `{"eqp1":"1", "eqp2":"2"}`, then it will break when MSAL is trying to talk to the older broker (which expects the pair format)
> > To prevent the confusion, I'll just kill the adapter.
> 
> What does GSON default adapter do? I suppose it is handling Pair correctly? Can the default not handle Map.Entry?

you get something like this 

pair: [{"first":"eqp1","second":"1"},{"first":"eqp2","second":"2"}]

map: [{"key":"eqp1","value":"1"},{"key":"eqp2","value":"2"}]

common/src/main/java/com/microsoft/identity/common/internal/util/QueryParamsAdapter.java
```diff
@@ -39,64 +43,60 @@
...
-    @Override
-    public List<Map.Entry<String, String>> read(final JsonReader in) throws IOException {
-        in.beginObject();
-        final List<Map.Entry<String, String>> result = new ArrayList<>();
-        while (in.hasNext()) {
-            final String key = in.nextName();
-            final String value = in.nextString();
-            final Map.Entry<String, String> keyValuePair = new AbstractMap.SimpleEntry<>(key, value);
-            result.add(keyValuePair);
+    public static String _toJson(final List<Map.Entry<String, String>> extraQueryStringParameters) {
```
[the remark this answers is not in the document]
rpdome: no. this is being called by us.

AdamBJohnsonx:
> Alternatively, i could make the adapter of List<Map.Entry> to return List's format.

Bleah.  So the thing you *can * do is to code the read side of this adapter so that it can understand any of these formats, rev the broker protocol version for the compact format, and choose how you want to send it.  We should *definitely* break any tie between the structure of code being used to hold the data and how that data is transmitted and understood, however it happens.

shahzaibj (approved):
Looks good to me.

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+        in.beginArray();
+        while (in.hasNext()) {
+            in.beginObject();
+
+            String key = "";
+            String value = "";
+
+            while (in.hasNext()) {
+                final String name = in.nextName();
+                if (StringUtil.equalsIgnoreCase(name, "first")) {
```
AdamBJohnsonx, line 155: nit: do we want this to be case insensitive?

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+     * */
+    boolean mWriteProperFormat = false;
+
+    static {
+        final GsonBuilder gsonBuilder = new GsonBuilder();
+        gsonBuilder.registerTypeAdapter(
+                getListType(),
+                // Turn off the "proper" format for now, because it'll break backcompat.
+                new QueryParamsAdapter(false)
+        );
```
shahzaibj, line 73: I'm wondering if this entire thing needs to live in some other class that bootstraps gson with custom adapters? Are we already doing that somewhere? If this is is the only instance then I guess it is fine to leave it here.

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+                    key = in.nextString();
+                } else if (StringUtil.equalsIgnoreCase(name, "second")) {
+                    value = in.nextString();
+                } else{
+                    throw new JsonSyntaxException("Unexpected NAME field: " + name);
+                }
+            }
+
+            final Map.Entry<String, String> keyValuePair = new AbstractMap.SimpleEntry<>(key, value);
+            result.add(keyValuePair);
```
AdamBJohnsonx, line 165: nit: inline?

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
rpdome: Right now we're using the static class to serialize (and get ClientException - which i just added), so this will be the only place.

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+            case BEGIN_ARRAY:
+                return readListPairFormat(in);
+
+            // i.e. {"eqp1":"1","eqp2","2"}, the "proper" one.
+            // We don't use it now because it's not compatible with the old Broker.
+            case BEGIN_OBJECT:
+                return readProperFormat(in);
+        }
+
+        return new ArrayList<>();
```
shahzaibj, line 126: does this need to get in the switch default statement?

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private List<Map.Entry<String, String>> readProperFormat(final JsonReader in) throws IOException {
+        final List<Map.Entry<String, String>> result = new ArrayList<>();
+
+        in.beginObject();
+        while (in.hasNext()) {
+            final String key = in.nextName();
+            final String value = in.nextString();
+            final Map.Entry<String, String> keyValuePair = new AbstractMap.SimpleEntry<>(key, value);
+            result.add(keyValuePair);
```
shahzaibj, line 137: Could probably convert these into one line

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+            final String errorMessage = "malformed json string:" + jsonString;
+            Logger.error(TAG + methodName, errorMessage, e);
+            throw new ClientException(ClientException.JSON_PARSE_FAILURE, errorMessage, e);
+        }
+    }
+
+    /**
+     * Create a Type for the List of query params
+     *
+     * @return a Type object representing the type of the query params in this case List<Pair<String, String>>
```
shahzaibj, line 196: seems to be Map.Entry instead of List<Pair> - I suppose the javadoc is wrong?

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+
+            final Map.Entry<String, String> keyValuePair = new AbstractMap.SimpleEntry<>(key, value);
+            result.add(keyValuePair);
+            in.endObject();
+        }
+        in.endArray();
+        return result;
+    }
+
+    public static String _toJson(final List<Map.Entry<String, String>> extraQueryStringParameters) {
```
shahzaibj, line 172: javadoc

common4j/src/main/com/microsoft/identity/common/java/util/QueryParamsAdapter.java
```diff
@@ -0,0 +1,201 @@
...
+        return mGson.toJson(extraQueryStringParameters, getListType());
+    }
+
+    public static List<Map.Entry<String, String>> _fromJson(final String jsonString)
```
shahzaibj, line 176: javadoc

common/src/main/java/com/microsoft/identity/common/internal/util/QueryParamsAdapter.java
iamgusain: We should document this in the release notes for MSAL/ADAL.. As part of how to upgrade to new major version.
right? 
<Changing the code to send Map.Entry instead of Pair>

shahzaibj: I think with this change now even new one sends Pair..so I don't expect there to be any changes needed for upgrade. @rpdome can confirm.

rpdome: the new one will serialize Map.Entry into the very same format as Pair.

iamgusain: I had to make this change in Authenticator code, with the first RC build, that's why I thought..
Maybe I will need to revert that with the new RC build :)

PR #1594, opened 2021-10-12
Errors in the web view don't crash the process without responding.

AdamBJohnsonx:
When we have a situation that an unknown exception is thrown in this view, we throw it out into the web view, which
has poor characteristics about handling the error.  Instead, if it is not a critical error, catch it and report it to the callback
listener.  If it is a critical Error, report it and rethrow it.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
@@ -182,9 +182,6 @@ public class PKeyAuthChallenge implements Serializable {
                     throw new ClientException(ErrorStrings.KEY_CHAIN_PRIVATE_KEY_EXCEPTION);
                 }
                 final PublicKey publicKey = deviceCertProxy.getPublicKey();
-                if (publicKey == null) {
-                    throw new ClientException(ErrorStrings.KEY_CHAIN_PUBLIC_KEY_EXCEPTION);
-                }
                 final X509Certificate certificate = deviceCertProxy.getCertificate();
                 if (certificate == null) {
                     throw new ClientException(ErrorStrings.KEY_CHAIN_CERTIFICATE_EXCEPTION);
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/exception/ErrorStrings.java
```diff
@@ -220,6 +220,16 @@ private ErrorStrings() {
      */
     public static final String KEY_CHAIN_PRIVATE_KEY_EXCEPTION = "Key Chain private key exception";
 
+    /**
+     * Key Chain public key exception.
+     */
+    public static final String KEY_CHAIN_PUBLIC_KEY_EXCEPTION = "Key Chain public key exception";
+
```
AdamBJohnsonx, line 227:
```suggestion
```

AdamBJohnsonx:
> Do we know why the public key was null?

In the key information refactoring that was done, it was not persisted.  There are no current uses, and it should be part of the persisted certificate.

common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
```diff
@@ -179,12 +181,17 @@ public String getSubmitUrl() {
                 if (privateKey == null) {
                     throw new ClientException(ErrorStrings.KEY_CHAIN_PRIVATE_KEY_EXCEPTION);
                 }
+                final PublicKey publicKey = deviceCertProxy.getPublicKey();
+                final X509Certificate certificate = deviceCertProxy.getCertificate();
+                if (certificate == null) {
```
shahzaibj, line 186: What if the device isn't joined? Or we are not going to even get here in that case?

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/challengehandlers/PKeyAuthChallenge.java
rpdome: the cert proxy should not be set in the first place

PR #1615, opened 2021-10-25
Log a message when we fail to extract information from homeAccountId

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -135,6 +137,15 @@ public static String join(final char delimiter, @NonNull final Iterable<String>
                 && !StringExtensions.isNullOrBlank(uidUtidArray[INDEX_UTID])) {
             uid = uidUtidArray[INDEX_UID];
             utid = uidUtidArray[INDEX_UTID];
+        } else {
+            Logger.warnPII(TAG, "We had a home account id that could not be split correctly, " +
```
shahzaibj, line 141: I don't see any PII here

```suggestion
            Logger.warn(TAG, "We had a home account id that could not be split correctly, " +
```

AdamBJohnsonx: Home account id is never PII?

shahzaibj: I mean all you were logging was the length of the homeAccountId. I don't think String length is PII.

shahzaibj (approved):

p3dr0rv (approved):

PR #1616, opened 2021-10-28
Porting the home account id logging to master.

AdamBJohnsonx:


common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -132,6 +134,15 @@ public static String join(final char delimiter, @NonNull final Iterable<String>
                 && !StringExtensions.isNullOrBlank(uidUtidArray[INDEX_UTID])) {
             uid = uidUtidArray[INDEX_UID];
             utid = uidUtidArray[INDEX_UTID];
+        } else {
+            Logger.warnPII(TAG, "We had a home account id that could not be split correctly, " +
```
AdamBJohnsonx, line 138:
```suggestion
            Logger.warn(TAG, "We had a home account id that could not be split correctly, " +
```
```suggestion
            Logger.warnPII(TAG, "We had a home account id that could not be split correctly, " +
```

common/versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Tue Apr 06 22:55:08 UTC 2021
-versionName=3.6.3
+versionName=3.6.3-debugAccountSplit
```
AdamBJohnsonx, line 2:
```suggestion
versionName=3.6.5-RC1
```

shahzaibj:
Why are we porting this to `master`?

AdamBJohnsonx:
> Why are we porting this to `master`?

Because I need something to potentially release off of before the 4.0 train leaves?

shahzaibj:
> > Why are we porting this to `master`?
> 
> Because I need something to potentially release off of before the 4.0 train leaves?

I didn't realize we were doing another release off of master here...are we doing it?

rpdome (approved):
Generally, if we're cutting a hotfix for specific issues, i would be more inclined to ship with that fix only.

It's just adding logger, so it's probably fine.

rpdome:
> > > Why are we porting this to `master`?
> > 
> > 
> > Because I need something to potentially release off of before the 4.0 train leaves?
> 
> I didn't realize we were doing another release off of master here...are we doing it?

This is for unblocking OneAuth development.

PR #1623, opened 2021-11-04
Allow for the cache directory to be null.

AdamBJohnsonx:
While debugging an ADAL test failure I tripped across this oddity.  It looks like our mock contexts there don't have this method implemented, and so explode if you call it when in debugging mode.  I just thought it would be better to tolerate the behavior.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
+++ b/common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
@@ -63,6 +63,7 @@ import com.microsoft.identity.common.java.util.ported.Predicate;
 import com.microsoft.identity.common.logging.Logger;
 import com.microsoft.identity.common.java.strategies.IAuthorizationStrategyFactory;
 
+import java.io.File;
 import java.io.IOException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -110,7 +111,10 @@ private static synchronized void initializeStaticClasses(@NonNull final Context
         if (!sInitialized) {
             Device.setDeviceMetadata(new AndroidDeviceMetadata());
             Logger.setAndroidLogger();
-            HttpCache.initialize(context.getCacheDir());
+            final File cacheDir = context.getCacheDir();
+            if (cacheDir != null) {
+                HttpCache.initialize(cacheDir);
+            }
```
shahzaibj, line 117: we should log a warning in the `else` case.

shahzaibj (approved):

AdamBJohnsonx:
> it shouldn't happen with Robolectric, These are probably mocked tests?

Yes, they are.

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
```diff
@@ -110,7 +111,12 @@ private static synchronized void initializeStaticClasses(@NonNull final Context
         if (!sInitialized) {
             Device.setDeviceMetadata(new AndroidDeviceMetadata());
             Logger.setAndroidLogger();
-            HttpCache.initialize(context.getCacheDir());
+            final File cacheDir = context.getCacheDir();
+            if (cacheDir != null) {
+                HttpCache.initialize(cacheDir);
+            } else {
+                Logger.warn(TAG, "Http caching is not enabled because the cache dir is null");
```
iamgusain, line 118: Can this happen in production ? and should this be an error in that case?

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
AdamBJohnsonx: I'm unsure.  We **should not** depend on http caching for anything other than performance...

PR #1626, opened 2021-11-11
[WIP] Return an error if the interactive thread pool is maxed out

rpdome:
Issue:
1) Launch MSAL test app, launch a new task A, click acquireToken. webview is launched
2) launch another task B, click acquireToken, nothing happened (because the thread pool is full.)

I accidentally stumbled across this issue during bugbash issue investigation, and from task B, there is no clear way to unblock myself. (You'll have to go to the activity hub, then manually kill/finish task A's interactive session).

--

Proposed fix: Throw an exception when this happens.

Alternatively, we could
1. cancel the ongoing session.
2. Increase thread pool size and support simultaneous interactive session (might still run into the same issue if someone abuses the api).

Merging this to dev because this is a behavior-changing (ish) change.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -632,7 +632,9 @@ public class CommandDispatcher {
                                 commandParameters.getSdkVersion()
                         );
                         try {
-                            sRunningInteractiveSessionCount++;
+                            synchronized (sLock) {
+                                sRunningInteractiveSessionCount++;
+                            }
 
                             // set correlation id on parameters as it may not already be set
                             commandParameters.setCorrelationId(correlationId);
@@ -674,7 +676,9 @@ public class CommandDispatcher {
                             Telemetry.getInstance().flush(correlationId);
                             returnCommandResult(command, commandResult);
                         } finally {
-                            sRunningInteractiveSessionCount--;
+                            synchronized (sLock) {
+                                sRunningInteractiveSessionCount--;
+                            }
                             DiagnosticContext.INSTANCE.clear();
                         }
                     }
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -29,6 +29,7 @@ import static com.microsoft.identity.common.java.AuthenticationConstants.LocalBr
 import static com.microsoft.identity.common.java.AuthenticationConstants.SdkPlatformFields.PRODUCT;
 import static com.microsoft.identity.common.java.AuthenticationConstants.SdkPlatformFields.VERSION;
 import static com.microsoft.identity.common.java.commands.SilentTokenCommand.ACQUIRE_TOKEN_SILENT_DEFAULT_TIMEOUT_MILLISECONDS;
+import static com.microsoft.identity.common.java.exception.ClientException.INTERACTIVE_SESSION_IN_PROGRESS;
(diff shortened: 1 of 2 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -85,13 +86,16 @@
...
+    private static final int INTERACTIVE_REQUEST_THREAD_POOL_SIZE = 1;
     private static final int SILENT_REQUEST_THREAD_POOL_SIZE = 5;
-    private static final ExecutorService sInteractiveExecutor = Executors.newSingleThreadExecutor();
+    private static final ExecutorService sInteractiveExecutor = Executors.newFixedThreadPool(INTERACTIVE_REQUEST_THREAD_POOL_SIZE);
     private static final ExecutorService sSilentExecutor = Executors.newFixedThreadPool(SILENT_REQUEST_THREAD_POOL_SIZE);
     private static final Object sLock = new Object();
     private static InteractiveTokenCommand sCommand = null;
     private static final CommandResultCache sCommandResultCache = new CommandResultCache();
 
+    private static int sRunningInteractiveSessionCount = 0;
```
AdamBJohnsonx, line 97: Maybe make this an atomic.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -85,13 +86,16 @@
 public class CommandDispatcher {
 
     private static final String TAG = CommandDispatcher.class.getSimpleName();
+    private static final int INTERACTIVE_REQUEST_THREAD_POOL_SIZE = 1;
     private static final int SILENT_REQUEST_THREAD_POOL_SIZE = 5;
-    private static final ExecutorService sInteractiveExecutor = Executors.newSingleThreadExecutor();
+    private static final ExecutorService sInteractiveExecutor = Executors.newFixedThreadPool(INTERACTIVE_REQUEST_THREAD_POOL_SIZE);
```
shoatman, line 91: I believe this returns a ThreadPoolExecutor... which has a getActiveCount method... I mention this because I'm not sure we should keep track of interactive account.... unless we're planning to support more than one in the future.... and I think at that point we want only 1 interactive request per Android Task....

shoatman (changes requested):
I'd suggest we just use ActiveCount of threadPoolExecutor... and I think that even newSingleThreadExecutor will return that type actually.

common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
```diff
@@ -188,6 +188,10 @@
      */
     public static final String DUPLICATE_COMMAND = "duplicate_command";
 
+    /**
+     * An interactive session is already in progress.
+     */
+    public static final String INTERACTIVE_SESSION_IN_PROGRESS = "interactive_session_in_progress";
```
AdamBJohnsonx, line 194: I think that if we're breaking things, we should allow the application author the ability to have their software recover the situation, by either waiting on the current interactive request to complete or by terminating it.  This means that we need to give them a way to discover what it is, and wait on it/kill it.  We don't have to do it all for them, but it shouldn't be impossible.  And we have to rethink this in the context of the broker, where the calculus is completely different, given that we will be serving requests from multiple applications potentially representing different interests.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
AdamBJohnsonx: getActiveCount doesn't tell you what you think it might.  There *will* be race conditions where the active count is zero and there is a thread in queue.

PR #1630, opened 2021-11-15
Adjust signing parameters, add failure checking

AdamBJohnsonx:
Courtesy of the deprecation of ubuntu 16, we seem to be running with a new version of gpg.  We need to do a few things
differently in this version, and I think we want to fail more thoroughly when things go wrong.

Fix command error, check for signature existence

There is no set in the --pinentry-mode paramteter

Attempt to configure the now-required agent

unquote home directory character

Gpg.conf files use dashes, not underscores

Add a few more options to the gpg.conf


p3dr0rv (approved):
lgtm!
Thanks @AdamBJohnsonx

PR #1632, opened 2021-11-15
merge release/4.0.0 to dev

iamgusain:
Merging changes from release/4.0.0 back to dev

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
@@ -34,6 +34,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+import cz.msebera.httpclient.android.BuildConfig;
 import lombok.NonNull;
 
 /**
@@ -101,8 +102,8 @@ public class Device {
 
         final String version = DiagnosticContext.INSTANCE.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION);
         if (StringUtil.isNullOrEmpty(version)) {
-            Logger.warn(TAG + methodName, "Product version is not set.");
-            return NOT_SET;
+            Logger.warn(TAG + methodName, "Product version is not set.", null);
+            return StringUtil.isNullOrEmpty(BuildConfig.VERSION_NAME) ? "1.5.9-default" : BuildConfig.VERSION_NAME + "-default";
         } else {
             return version;
         }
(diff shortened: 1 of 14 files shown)


shahzaibj (approved):

AdamBJohnsonx (approved):
Need to cherry-pick more things back from master... but good.

common4j/src/test/com/microsoft/identity/common/java/platform/DeviceTest.java
```diff
@@ -93,7 +93,7 @@ public void testGetOs(){
 
     @Test
     public void testGetProductVersion_DiagContextNotSet(){
-        Assert.assertEquals(NOT_SET, Device.getProductVersion());
+        Assert.assertEquals("1.5.9-default", Device.getProductVersion());
```
rpdome, line 96: This is probably an actual regression.

common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
```diff
@@ -101,8 +102,8 @@ public static String getProductVersion() {
 
         final String version = DiagnosticContext.INSTANCE.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.VERSION);
         if (StringUtil.isNullOrEmpty(version)) {
-            Logger.warn(TAG + methodName, "Product version is not set.");
-            return NOT_SET;
+            Logger.warn(TAG + methodName, "Product version is not set.", null);
+            return StringUtil.isNullOrEmpty(BuildConfig.VERSION_NAME) ? "1.5.9-default" : BuildConfig.VERSION_NAME + "-default";
```
rpdome, line 106: Why are we hardcoding 1.5.9-default here?

NOT_SET means we're not initializing this properly. This version value is also platform-dependent. Not sure where 1.5.9 comes from.

iamgusain: This was changed in #1607

common4j/src/test/com/microsoft/identity/common/java/platform/DeviceTest.java
iamgusain: This is due to the change in #1607

common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
rpdome: Let's add a comment + create an PBI to fix this properly.

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
rpdome: to 
1. undo the default value to NOT_SET (if we're getting this value, it means the upper layer didn't initialize the value properly).
2. Fix it in the upper layer(s).

PR #1634, opened 2021-11-16
Fix Continuous delivery pipeline

iamgusain:
### What
Update continuous integration pipeline to use version number in the format 0.0.YYYYMMDD.ver
Update common4j dependency in common module to be 0.0.+ , such that it can pick up the latest build published from continuous-delivery piepeline

### Why
With latest merge from release/4.0.0 branch the common4j dep in common module got updated to a fixed version, we need to keep it dynamic such that it can consume the latest common4j code.


p3dr0rv:
I see the pipeline is failing, do we need something else here
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=846411&view=results

common/build.gradle
```diff
@@ -109,7 +109,7 @@ dependencies {
         transitive = false
     }
 
-    distApi("com.microsoft.identity:common4j:1.0.0-RC5") {
+    distApi("com.microsoft.identity:common4j:0.0.+") {
```
shahzaibj, line 112: So this would no longer be something like 0.0.n and rather be `0.0.$(Date:yyyyMMdd)$(Rev:.r) # $(Build.BuildNumber)` - have we verified that Android Studio can automatically pick up new ones correctly? Just want to make sure we have it verified

iamgusain:
> I see the pipeline is failing, do we need something else here https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=846411&view=results

There are two new ropc unit tests which are failing 
cc: @shahzaibj

common/build.gradle
iamgusain: Yes,

shahzaibj:
> > I see the pipeline is failing, do we need something else here https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=846411&view=results
> 
> There are two new ropc unit tests which are failing cc: @shahzaibj

At a glance it looks like you don't have the lab cert on this new pipeline...so you're not able to pull accounts from lab api

p3dr0rv:
We might need to add the following task under [steps](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/fea9efaf87b328575bb86d721ec53bef685ffb5d/azure-pipelines/continous-delivery/common.yml#L36) in the common_phase


```
  - template: ../templates/steps/automation-cert.yml
    parameters:
      envVstsMvnAt: ENV_VSTS_MVN_ANDROIDCOMMON_ACCESSTOKEN
```

p3dr0rv (approved):

shahzaibj (approved):

PR #1636, opened 2021-11-17
Hook telemetry to LocalAuthenticationResult and BaseException to be a…

shahzaibj:
…ble to supplied to caller

Related broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1779

Basically we need to send telemetry back to OneAuth-MSAL from Broker. Some of this telemetry is produced via the [common4j Event Based Telemetry infra](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java), however, that telemetry wasn't actually being returned as part of any result objects...it is tied to lifecycle of commands and is flushed to observers upon completion of the command, but not provided as part of say the Authentication Result for a particular request. OneAuth-MSAL wants the telemetry back for a particular request along with the result data. To achieve this we grab the telemetry (prior to it being flushed), and we attach it to the result objects (LocalAuthenticationResult or BaseException), so that broker has it available in the result objects and can then run the [OneAuthTelemetryExtractor on it and send the sanitized version back to OneAuth](https://github.com/AzureAD/ad-accounts-for-android/pull/1779).

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -277,8 +277,6 @@ public class CommandDispatcher {
                                 + correlationId + ", with the status : " + commandResult.getStatus().getLogStatus()
                                 + " is cacheable : " + command.isEligibleForCaching());
                         // TODO 1309671 : change required to stop the LocalAuthenticationResult object from mutating in cases of cached command.
-                        // set correlation id on Local Authentication Result
-                        setCorrelationIdOnResult(commandResult, correlationId);
                         final List<Map<String, String>> telemetryMap = Telemetry.getInstance().getMap(correlationId);
                         commandResult.setTelemetryMap(telemetryMap);
                         Telemetry.getInstance().flush(correlationId);
@@ -354,8 +352,6 @@ public class CommandDispatcher {
                         Logger.info(TAG + methodName, "Completed as owner for correlation id : **"
                                 + correlationId + statusMsg(commandResult.getStatus().getLogStatus())
                                 + " is cacheable : " + command.isEligibleForCaching());
-                        // set correlation id on Local Authentication Result
-                        setCorrelationIdOnResult(commandResult, correlationId);
                         Telemetry.getInstance().flush(correlationId);
                         EstsTelemetry.getInstance().flush(command, commandResult);
                         finalFuture.setResult(commandResult);
@@ -491,7 +487,23 @@ public class CommandDispatcher {
             }
         }
 
+        // set correlation id on Local Authentication Result
+        setCorrelationIdOnResult(commandResult, correlationId);
+        setTelemetryOnResult(commandResult, correlationId);
         return commandResult;
+    }
+
+    private static void setTelemetryOnResult(@NonNull final CommandResult commandResult,
+                                             @NonNull final String correlationId) {
+        final List<Map<String, String>> telemetryMap = Telemetry.getInstance().getMap(
+                correlationId
+        );
+
+        if (commandResult.getResult() instanceof LocalAuthenticationResult) {
+            ((LocalAuthenticationResult) commandResult.getResult()).setTelemetry(telemetryMap);
+        } else if (commandResult.getResult() instanceof BaseException) {
+            ((BaseException) commandResult.getResult()).setTelemetry(telemetryMap);
+        }
     }
 
         /**
@@ -649,9 +661,6 @@ public class CommandDispatcher {
                             sCommand = null;
 
                             LocalBroadcaster.INSTANCE.unregisterCallback(RETURN_AUTHORIZATION_REQUEST_RESULT);
-
-                            // set correlation id on Local Authentication Result
-                            setCorrelationIdOnResult(commandResult, correlationId);
 
                             Logger.info(TAG + methodName,
                                     "Completed interactive request for correlation id : **" + correlationId +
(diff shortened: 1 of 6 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -491,9 +487,25 @@ private static CommandResult executeCommand(@SuppressWarnings(WarningType.rawtyp
...
+                                             @NonNull final String correlationId) {
+        final List<Map<String, String>> telemetryMap = Telemetry.getInstance().getMap(
+                correlationId
+        );
+
+        if (commandResult.getResult() instanceof LocalAuthenticationResult) {
+            ((LocalAuthenticationResult) commandResult.getResult()).setTelemetry(telemetryMap);
+        } else if (commandResult.getResult() instanceof BaseException) {
+            ((BaseException) commandResult.getResult()).setTelemetry(telemetryMap);
+        }
```
rpdome, line 506: adding log for some other type of results?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -491,9 +487,25 @@ private static CommandResult executeCommand(@SuppressWarnings(WarningType.rawtyp
             }
         }
 
+        // set correlation id on Local Authentication Result
+        setCorrelationIdOnResult(commandResult, correlationId);
+        setTelemetryOnResult(commandResult, correlationId);
         return commandResult;
     }
 
+    private static void setTelemetryOnResult(@NonNull final CommandResult commandResult,
```
rpdome, line 496: I'm wondering if it's better to create a function that do all the telemetry work in one place 
(attach to command Result AND flush)

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
AdamBJohnsonx: yeah, you should do something in the other else case.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
AdamBJohnsonx: I'm actually wondering if we should just lean on thread scope to gather broker telemetry rather than attempting to attach it to all these exceptions and result types.  It's probably not time to worry about that, but it would be nicer, I think.

shahzaibj: Yes, the current design is not great. I want to actually refactor CommandDispatcher and start returning CommandResult out of that...that way only command result contains telemetry. Currently we don't return command result anywhere, it's weird. Refactoring all this CommandDispather is complex and I'll save that for later.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
AdamBJohnsonx: Something like... unless this actually happens all the time.
```suggestion
        } else if (commandResult.getResult() != null) {
          Logger.warn(TAG + ":setTelemetryOnResult", "Unknown result type " + commandResult.getResult().getCanonicalName());
        }
```

common4j/src/main/com/microsoft/identity/common/java/exception/BaseException.java
```diff
@@ -189,12 +195,26 @@ public void setUsername(@NonNull final String username) {
...
         //TODO : ADO 1373343 Add the whole transient exception category.
         return !nonCacheableErrorCodes.contains(mErrorCode);
     }
+
+    /**
+     * Set the telemetry on base exception.
+     *
+     * @param telemetry the {@link List<Map<String, String>>} containing telemetry data
+     */
+    public void setTelemetry(@NonNull final List<Map<String, String>> telemetry) {
```
AdamBJohnsonx, line 212: Two things... take this on the constructor, and make a safe-copy.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
shahzaibj: It would happen frequently in MSAL..because we have LoadAccountCommand, RemoveAccountCommand, DeviceModeCommand and their corresponding results coming through. So I don't think we need warn log here...we could do a `verbose` level log if we still want to know that. 

Ideally we also put telemetry on results for these other commands, however it is irrelevant right now for broker because broker getAccounts, removeAccount operations are actually not using commands and the dispatcher. It's only used by MSAL, and since my telemetry work was specific to broker, therefore I didn't bother about that, but at some point we need to be consistent with MSAL and Broker and probably start using Dispatcher in the same way in each operation.

common4j/src/main/com/microsoft/identity/common/java/exception/BaseException.java
shahzaibj: Changed to `add`. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1636/commits/4f18eff7cf4b93131c2ebd4ec44908688f0efe0b

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
shahzaibj: Created separate `setOnResultAndFlush` method. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1636/commits/4f18eff7cf4b93131c2ebd4ec44908688f0efe0b

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
shahzaibj: Added verbose level log. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1636/commits/4f18eff7cf4b93131c2ebd4ec44908688f0efe0b

rpdome (approved):

AdamBJohnsonx (approved):

PR #1638, opened 2021-11-19
Modified base class for BrokerAppMetadata cache.  Implemented as sing…

shoatman:
- Added ReentrantReadWriteLock to base class
- Added member variable to in-memory list to reduce the number of times reading from SharedPreferences and deserializing from JSON
- Added readLock around read operations
- Added writeLock around update/remove/clear operations
- Converted getAll to use a copy constructor and return a copy of the in-memory list/set
-

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
@@ -29,6 +29,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.google.gson.reflect.TypeToken;
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.logging.Logger;
 
 import java.lang.reflect.Type;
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
```diff
@@ -78,7 +96,14 @@ public void remove(@NonNull final String clientId,
     private void disposeOfDuplicateRecords(@NonNull final String clientId,
                                            @NonNull final String environment,
                                            final int uid) {
-        final List<BrokerApplicationMetadata> allMetadata = getAll();
+        Set<BrokerApplicationMetadata> allMetadata = null;
+        readLock.lock();
```
AdamBJohnsonx, line 100: You'll need to hold the write lock for the entire duration of this activity. Read locks aren't exclusive.

AdamBJohnsonx (changes requested):
You're going to need to be far more pessimistic about your locking here.  Read locks aren't exclusive, so you're not actually preventing concurrent, overlapping modification unless you hold the write lock.  You have to hold that write lock through the entire read-modify-write operation unless you want to implement an optimistic locking optimization... which is almost surely unjustified.

shoatman:
@tanmaymanolkar1 - Can you take a look at this PR as discussed and propose some tests for this work.  Thanks

tanmaymanolkar1:
Will have look at it @shoatman, thanks for working on this fix :)

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesBrokerApplicationMetadataCache.java
```diff
@@ -41,31 +42,36 @@
...
     private static final String DEFAULT_APP_METADATA_CACHE_NAME = "com.microsoft.identity.app-meta-cache";
 
     private static final String KEY_CACHE_LIST = "app-meta-cache";
 
-    public SharedPreferencesBrokerApplicationMetadataCache(@NonNull final Context context) {
-        super(context, DEFAULT_APP_METADATA_CACHE_NAME, KEY_CACHE_LIST);
+    public synchronized static SharedPreferencesBrokerApplicationMetadataCache getInstance(@NonNull final Context context){
+        if(INSTANCE == null){
+            INSTANCE = new SharedPreferencesBrokerApplicationMetadataCache(context);
+        }
```
AdamBJohnsonx, line 56: nit: consider lazy-holder pattern instead.

PR #1639, opened 2021-11-19
Make LocalBroadcaster.broadcast method to be asynchronous

iamgusain:
### What
Making the "broadcast" method under LocalBroadcaster class to do the broadcasting operation asynchronously.

### Why
With refactoring work we have replaced the androidx.localbroadcastmanager.content.LocalBroadcastManager class with our own implemenation in LocalBroadcaster to make it shareable across platforms as part of common4j. 
The LocalBroadcastManager's sendBroadcast method was used to broadcast the given intent asynchronously.
```
/**
     * Broadcast the given intent to all interested BroadcastReceivers.  This
     * call is asynchronous; it returns immediately, and you will continue
     * executing while the receivers are run.
     *
     * @param intent The Intent to broadcast; all receivers matching this
     *     Intent will receive the broadcast.
     *
     * @see #registerReceiver
     *
     * @return Returns true if the intent has been scheduled for delivery to one or more
     * broadcast receivers.  (Note tha delivery may not ultimately take place if one of those
     * receivers is unregistered before it is dispatched.)
     */
    public boolean sendBroadcast(@NonNull Intent intent) {
```
However in LocalBroadcaster class we are sending all the broadcast synchronously. In this change we are making this method to execute asynchronously. 
We found an issue while testing in oneauth due to above method being synchronous. When back key is pressed the AuthorizationFragment  goes into a bad state and eventually causes a crash. 

### How
Executing the broadcast operation in a separate thread

### Testing
Verified that it fixes the oneauth crash.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
@@ -26,6 +26,8 @@ import com.microsoft.identity.common.java.logging.Logger;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import lombok.NonNull;
 
@@ -33,6 +35,7 @@ public enum  LocalBroadcaster {
     INSTANCE;
 
     private static final String TAG = LocalBroadcaster.class.getSimpleName();
+    private static final ExecutorService sBroadcastExecutor = Executors.newSingleThreadExecutor();
 
     public interface IReceiverCallback {
         void onReceive(@NonNull final PropertyBag propertyBag);
@@ -59,16 +62,19 @@ public enum  LocalBroadcaster {
         mReceivers.remove(alias);
     }
 
-    public void broadcast(@NonNull final String alias, @NonNull final PropertyBag propertyBag){
+    public void broadcast(@NonNull final String alias, @NonNull final PropertyBag propertyBag) {
         final String methodName = ":broadcast";
-        final IReceiverCallback receiver = mReceivers.get(alias);
-
-        if (receiver != null){
-            Logger.info(TAG + methodName, "broadcasting to alias: " + alias);
-            receiver.onReceive(propertyBag);
-        } else {
-            Logger.info(TAG + methodName, "No callback is registered with alias: " + alias +
-                    ". Do nothing.");
-        }
+        sBroadcastExecutor.execute(new Runnable() {
+            public void run() {
+                final IReceiverCallback receiver = mReceivers.get(alias);
+                if (receiver != null) {
+                    Logger.info(TAG + methodName, "broadcasting to alias: " + alias);
+                    receiver.onReceive(propertyBag);
+                } else {
+                    Logger.info(TAG + methodName, "No callback is registered with alias: " + alias +
+                            ". Do nothing.");
+                }
+            }
+        });
     }
 }
--- a/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
@@ -31,7 +31,7 @@ import java.util.concurrent.Executors;
 
 import lombok.NonNull;
 
-public enum  LocalBroadcaster {
+public enum LocalBroadcaster {
     INSTANCE;
 
     private static final String TAG = LocalBroadcaster.class.getSimpleName();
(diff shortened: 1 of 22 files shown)


common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -26,13 +26,16 @@
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import lombok.NonNull;
 
 public enum  LocalBroadcaster {
```
AdamBJohnsonx, line 34:
```suggestion
public enum LocalBroadcaster {
```

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -26,13 +26,16 @@
...
     INSTANCE;
 
     private static final String TAG = LocalBroadcaster.class.getSimpleName();
+    private static final ExecutorService sBroadcastExecutor = Executors.newSingleThreadExecutor();
```
AdamBJohnsonx, line 38: Di we care that we only have one thread to broadcast from?

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -59,16 +62,19 @@ public void unregisterCallback(@NonNull final String alias){
...
-        final IReceiverCallback receiver = mReceivers.get(alias);
-
-        if (receiver != null){
-            Logger.info(TAG + methodName, "broadcasting to alias: " + alias);
-            receiver.onReceive(propertyBag);
-        } else {
-            Logger.info(TAG + methodName, "No callback is registered with alias: " + alias +
-                    ". Do nothing.");
-        }
+        sBroadcastExecutor.execute(new Runnable() {
```
AdamBJohnsonx, line 67: Is it possible to generate some waiting-time metrics here, just in case something gets stuck in queue for the executor?

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
rpdome: Should the register/unregister function also utilize this executor?

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -26,13 +26,16 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import lombok.NonNull;
 
 public enum  LocalBroadcaster {
     INSTANCE;
 
     private static final String TAG = LocalBroadcaster.class.getSimpleName();
+    private static final ExecutorService sBroadcastExecutor = Executors.newSingleThreadExecutor();
```
rpdome, line 39: Should the register/unregister function also utilize this executor?

rpdome: Or maybe switch mReceivers to concurrentHashMap?

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
iamgusain: I guess not

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
iamgusain: switched to ConcurrentHashMap

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -24,21 +24,23 @@
...
 
     public interface IReceiverCallback {
         void onReceive(@NonNull final PropertyBag propertyBag);
     }
 
-    final Map<String, IReceiverCallback> mReceivers = new HashMap<>();
+    final ConcurrentHashMap<String, IReceiverCallback> mReceivers = new ConcurrentHashMap<>();
```
AdamBJohnsonx, line 43: ConcurrentMap may be all you need to declare as the reference type here.

AdamBJohnsonx (approved):

PR #1644, opened 2021-11-30
Add wrapper to allow injecting http response without using Robolectric

rpdome:
Inspired by (and took some code from) https://github.com/AzureAD/ad-accounts-for-android/pull/1719/files

This will allow us to directly inject http response to test cases.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IHttpClientWrapper.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IHttpClientWrapper.java
@@ -31,6 +31,11 @@ import lombok.NonNull;
  */
 public interface IHttpClientWrapper {
 
+    /**
+     * Wrap the given {@link HttpClient} object.
+     * e.g. For test cases, we could wraps the given {@link HttpClient} object with a Mock HttpClient.
+     *      while we can just return the given object in prod code.
+     */
     @NonNull
     HttpClient wrap(@NonNull final HttpClient client);
 }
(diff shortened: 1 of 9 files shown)


AdamBJohnsonx:
> Another alternative is to have the instantiated `InterceptedHttpClientWrapper` (a new one will be created for each PlatformComponent object) to take in MockHttpClient (or interceptors) directly.
> 
> We should just need one - having them for each endpoint is somewhat overkill imo.
> 
> Then we can kill the static part of the class.

Whichever, really.  From a test POV it doesn't really matter, though from an actual-usage POV there will be different configurations depending on the service you're talking to (timeouts, etc).

AdamBJohnsonx (approved):
I'm remarkable not picky here, now that I think about it.  Any way we choose to do it is fine, though I'd bias towards one that created fewer class files.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IHttpClientWrapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import lombok.NonNull;
+
+/**
+ * A wrapper interface around {@link HttpClient}, allowing us to inject responses in test cases.
+ */
+public interface IHttpClientWrapper {
+
+    @NonNull
+    HttpClient wrap(@NonNull final HttpClient client);
```
shahzaibj, line 40: probably javadoc here too

shahzaibj (approved):

shoatman (approved):
lgtm

iamgusain (approved):

PR #1650, opened 2021-12-08
Test cases for metadata cache synchronization

tanmaymanolkar1:


common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
```diff
@@ -86,6 +100,106 @@ public void testInsert() {
...
+        executionLatch.await();
+        latch.countDown();
+
+        for(Future<Boolean> f : list){
+            Assert.assertTrue(f.get());
+        }
+
+        final int size = mMetadataCache.getAll().size();
+
+        assertEquals("Expected 1 record, but got too many",
```
AdamBJohnsonx, line 129: Probably say how many, it could be 0.

tanmaymanolkar1: 3977441

common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
```diff
@@ -226,4 +340,15 @@ private static BrokerApplicationMetadata generateRandomMetadata() {
 
         return randomMetadata;
     }
+
+    private static BrokerApplicationMetadata generatePredictiveMetadata() {
```
AdamBJohnsonx, line 344: nit: maybe DuplicateMetadata?

common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
```diff
@@ -86,6 +100,106 @@ public void testInsert() {
...
+        final Set<BrokerApplicationMetadata> set = Collections.synchronizedSet(new HashSet<BrokerApplicationMetadata>());
+
+        mMetadataCache.clear();
+        //inserting one entry initially
+        mMetadataCache.insertOrUpdate(generatePredictiveMetadata());
+
+        assertEquals(1, mMetadataCache.getAll().size());
+
+        for (int i = 0; i <5; i++){
+            executor.submit(new Callable<Boolean>() {
```
AdamBJohnsonx, line 160: No point in using a Callable if you're not checking the return.

common/src/androidTest/java/com/microsoft/identity/common/SharedPreferencesBrokerApplicationMetadataCacheTest.java
```diff
@@ -86,6 +100,106 @@ public void testInsert() {
...
+    public boolean insertOrUpdateHelper
+            (BrokerApplicationMetadata metadata, CountDownLatch latch, CountDownLatch executionLatch)
+            throws InterruptedException {
+        executionLatch.countDown();
+        latch.await();
+        return mMetadataCache.insertOrUpdate(metadata);
+    }
+
+    @Test
+    public void testReadsWithInsertOrUpdate() throws Exception {
```
AdamBJohnsonx, line 146: I think the is the seconds test I discussed with you.  The thing is, you kind of want to test multiple possible interleaving of this mess, and there's actually not a really good way to control it.  The best we're going to be able to do is get a "probably covered it".  Meaning that you want to set this up to strobe the cache structure until you tell it to stop - probably a set number of iterations or total ops per reader/writer pool, like 100000 or so.

So you don't need to have all the latching, necessarily, though controlling the start doesn't hurt you.  You will want to execute your read in a loop, potentially with a micro/milli second sleep in there to make sure you don't starve the writers while you continually insert/update the data store.

PR #1651, opened 2021-12-09
Configure debug broker support in release mode

gitaumoses4:
We currently support MSAL trusting debug brokers; however, we need to be sure that people don't set this in error for release builds that we'll be shipped to production.

> When the build-type is RELEASE and `trust debug brokers` is enabled throw a runtime exception.  

## Implementation 

We need to check that the application trusting debug brokers is allowed to do so, hence the `BrokerValidator` hosts a variable that keeps a record of applications with the ability to trust debug brokers.


rpdome (dismissed):

shahzaibj (changes requested):
Looks like this PR effectively reverts the work we had done to enable debug broker support in release mode to allow automation to be run with prod MSAL and BrokerHost. With this PR we can no longer to do that. I would still like to have support for that. I think `RuntimeException` is not the right thing to do here. I suggest you schedule a meeting to discuss this work.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -59,10 +59,11 @@
...
 
     public static void setShouldTrustDebugBrokers(final boolean shouldTrustDebugBrokers) {
-        if (!BuildConfig.DEBUG && shouldTrustDebugBrokers) {
-            Logger.warn(TAG, "You are forcing to trust debug brokers in non-debug builds.");
+        if (BuildConfig.DEBUG || BuildConfig.SHOULD_TRUST_DEBUG_BROKERS) {
+            BrokerValidator.sShouldTrustDebugBrokers = shouldTrustDebugBrokers;
+            return;
         }
-        BrokerValidator.sShouldTrustDebugBrokers = shouldTrustDebugBrokers;
+        throw new RuntimeException("Cannot trust debug brokers in non-debug builds.");
```
shahzaibj: What if the value supplied to the method is actually `false`? I don't think we need to throw RuntimeException in that case.

gitaumoses4: Thanks for catching this. Fixed in 0e7188f

tanmaymanolkar1 (approved):

common/build.gradle
```diff
@@ -52,12 +57,14 @@ android {
...
             testCoverageEnabled enableCodeCoverage
             minifyEnabled false
+            buildConfigField("boolean", "SHOULD_TRUST_DEBUG_BROKERS", "$shouldTrustDebugBrokers")
         }
         release {
             minifyEnabled false
             debuggable false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
             consumerProguardFiles 'consumer-rules.pro'
+            buildConfigField("boolean", "SHOULD_TRUST_DEBUG_BROKERS", "$shouldTrustDebugBrokers")
```
shahzaibj: No need to set this twice (in both debug and release). Just set this once in the `defaultConfig`

shahzaibj:
So what is the plan here? It seems like the only way this is now actually configurable is to first declare trust via BuildConfig, right? So that means the builds that are in Maven central are not actually ever going to be able to trust debug brokers because they are not in debug mode and also don't have trust declared in Build Config. I think that is fine... 

My question is how is this debug broker support useful now? I suspect that the plan would be to publish two versions of common going forward to at least the internal maven? One that has debug support declared via BuildConfig and one that doesn't. Is that the correct understanding? If that's the case, I'm not sure if we even need the `setShouldTrustDebugBroker` method in `BrokerValidator` at all. I'm not seeing any utility since the functioning is really controlled by build config.

gitaumoses4:
@shahzaibj  as discussed, we are now checking the application's package name and signature hash before allowing it to trust debug brokers.
For a start we allow only MSAL_TEST_APP and MSAL_AUTOMATION_APP to do so.

PR #1652, opened 2021-12-10
Fixing possible IndexOutOfBoundsException when retrieving tenantProfile and none exists

iamgusain:
### What
Fixing possible IndexOutOfBoundsException when retrieving tenantProfile and none exists.
This happens in method `loadByFamilyIdWithAggregatedAccountData` where we try to load tenantProfiles for given account Record. After loading all the tenantProfiles from cache we remove the first entry in the returned tenantProfiles as that is the original account record. However if there is no tenatProfile record this results in IndexOutOfBoundsException and fail the whole request.

### Why
Due to this issues user's silent request is failed for FOCI apps and they are prompted for Interactive login even though we have the account in cache. Link to the ICM https://portal.microsofticm.com/imp/v3/incidents/details/276821709/home

### How
Moved the code to remove first entry in tenantProfiles record inside the code block which first checks it the list is not empty

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftFamilyOAuth2TokenCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftFamilyOAuth2TokenCache.java
@@ -22,6 +22,8 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.cache;
 
+import static com.microsoft.identity.common.java.cache.AbstractAccountCredentialCache.targetsIntersect;
+
 import com.microsoft.identity.common.java.BaseAccount;
 import com.microsoft.identity.common.java.WarningType;
 import com.microsoft.identity.common.java.authscheme.AbstractAuthenticationScheme;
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftFamilyOAuth2TokenCache.java
```diff
@@ -212,13 +212,13 @@ public ICacheRecord loadByFamilyId(@Nullable final String clientId,
...
 
-        // Ignore the first element, as it will contain the same result as loadByFamilyId()....
-        accountsInOtherTenants.remove(0);
-
         if (!accountsInOtherTenants.isEmpty()) {
             // We need the IdToken of each of these accounts... we can reuse the RT, since it is
             // an FRT...
 
+            // Ignore the first element, as it will contain the same result as loadByFamilyId()....
+            accountsInOtherTenants.remove(0);
```
AdamBJohnsonx, line 220: Relatively minor, but maybe something to think about: if this result in an IndexOutOfBounds exception, that means that this is backed by an ArrayList, so that a remove operation will result in copying the entire array... It would be nice to avoid that, if we can, since we're already copying it once.

AdamBJohnsonx (approved):
It would be best if we could get a unit test for this.

iamgusain:
On digging deeper and taking a close look at the code, this doesn't seem to be the real problem. and this code is behaving as expected. Closing this PR and will continue to investigate more

Issue #1646, opened 2021-12-02
Google Play Store console flagging Unsafe Cipher Mode for StorageHelper.getKeyThumbPrint()


vicalcazar:
The Google Help Center Article link is:
https://support.google.com/faqs/answer/10046138

According to the article, a configuration mode of "AES/ECB/PKCS5Padding" is insecure when used for `Cipher.getInstance()`.

This configuration is used for `StorageHelper.CIPHER_ALGORITHM_FOR_KEY_TRACKING`

https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/44768f5a94616ff631f76461e89167e28d785749/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java#L146

PR #1654, opened 2021-12-14
Remove unsafe key thumbprint generator

rpdome:
Resolve : #1646 

We've been using an outdated encryption mode to generate thumbprints - to track the potential key change issue.

This pr is to get rid of this outdated logic, and make it call the new one.

(ps. We no longer use the StorageHelper in the main line scenario - eventually, the entire class will be gone).

PR #1655, opened 2021-12-15
Remove unsafe key thumbprint generator

rpdome:
Resolve : #1646 

Similar to #1654 - but this change is made on the master branch, and the code in #1654 does not exist here.


shoatman (approved):
Thanks Dome.

AdamBJohnsonx (approved):

PR #1659, opened 2021-12-21
Synchronize updating refresh token in cache (saving new and removing old) to avoid race condition

iamgusain:
### What
Synchronize writing new refresh tokens to Foci cache (saving new and removing old) to avoid race condition

### Why
Currently when saving Family Refresh tokens (FRT) we first save the new FRT to cache and then remove any non-matching RTs from cache. In case of multiple threads updating the cache concurrently, this leads to all RTs getting deleted from the cache. as each thread deletes the RT from the other thread. This manifests in user needing to do an interactive signin as there are no RTs in cache. 

### How
To fix the issue we need to synchronize updating Refresh Tokens in cache, to ensure that we always have the latest RT in the cache.

### Testing
Added a UT to mock below scenario
Save token in cache for Some Client Application
Spawn 2 threads and from each thread save new tokens to the cache
Verify that we have tokens available in the cache after execution of all the above task

Ensured that this test fails with reason no refresh token found without the change and passes once we add synchronization logic.

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/MicrosoftFamilyOAuth2TokenCacheTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/MicrosoftFamilyOAuth2TokenCacheTest.java
@@ -66,6 +66,7 @@ import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
 import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -573,7 +574,7 @@ public class MicrosoftFamilyOAuth2TokenCacheTest extends MsalOAuth2TokenCacheTes
         when(mockCredentialAdapter.createAccount(mockStrategy, mockRequest, mockResponse)).thenAnswer(new Answer<AccountRecord>() {
             @Override
             public AccountRecord answer(InvocationOnMock invocation) throws Throwable {
-                String currentThread = Thread.currentThread().getName();
+                final String currentThread = Thread.currentThread().getName();
                 if(currentThread == thread1[0]) {
                     return frtTestBundle1.mGeneratedAccount;
                 } else if(currentThread == thread2[0]) {
@@ -585,7 +586,7 @@ public class MicrosoftFamilyOAuth2TokenCacheTest extends MsalOAuth2TokenCacheTes
         when(mockCredentialAdapter.createAccessToken(mockStrategy, mockRequest, mockResponse)).thenAnswer(new Answer<AccessTokenRecord>() {
             @Override
             public AccessTokenRecord answer(InvocationOnMock invocation) throws Throwable {
-                String currentThread = Thread.currentThread().getName();
+                final String currentThread = Thread.currentThread().getName();
                 if(currentThread == thread1[0]) {
                     frtTestBundle1.mGeneratedAccessToken.setSecret(SECRET + "AT2" + currentThread);
                     return frtTestBundle1.mGeneratedAccessToken;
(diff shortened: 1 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -367,10 +366,11 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
 
         // Save the Account and Credentials...
         saveAccounts(accountToSave);
-        saveCredentialsInternal(accessTokenToSave, refreshTokenToSave, idTokenToSave);
-
-        // Remove old refresh tokens (except for the one we just saved) if it's MRRT or FRT
-        removeAllRefreshTokensExcept(accountToSave, refreshTokenToSave);
+        synchronized(this) {
```
AdamBJohnsonx, line 369: You probably want to synchronize this on a static monitor rather than an instance.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -367,10 +366,11 @@ public ICacheRecord save(@NonNull final GenericOAuth2Strategy oAuth2Strategy,
...
+            saveCredentialsInternal(accessTokenToSave, refreshTokenToSave, idTokenToSave);
+            // Remove old refresh tokens (except for the one we just saved) if it's MRRT or FRT
+            removeAllRefreshTokensExcept(accountToSave, refreshTokenToSave);
```
rpdome, line 372: I'm wondering if this is enough. 

It would protect this specific method from race-condition itself, but it won't protect any other potential cache-related race condition issue. 

What if saveCredentialsInternal is invoked by other methods when this synchronize block is operating? (i.e setSingleSignOnState is also calling both functions).

Maybe we need some sort of read-write lock for the class' public interfaces?

iamgusain: That probably is a bigger work item and out of scope for current fix, 
Shared Preference does provide some support for that, but we might need to think from our usage perspective, especially for cases where we expect two separate operations (add and remove) to happen atomically

common/src/test/java/com/microsoft/identity/common/MicrosoftFamilyOAuth2TokenCacheTest.java
```diff
@@ -525,4 +534,146 @@ public void testDualStackIdTokenRetrieval() throws ClientException {
...
+                    mOauth2TokenCache.save(mockStrategy, mockRequest, mockResponse);
+                } catch (ClientException e) {
+                    throw new AssertionError(e.getMessage());
+                }
+            }
+        };
+
+        ExecutorService executorService = Executors.newFixedThreadPool(5);
+        Future runnableTask1 = executorService.submit(runnable1);
+        Future runnableTask2 = executorService.submit(runnable2);
```
AdamBJohnsonx, line 653: There's no guarantee that these are actually not happening sequentially here.  You'll want to put a latch in place to insure that we have problems.

iamgusain: done

AdamBJohnsonx (approved):
I'm going to approve this, but to be truly correct, you really need two latches - one to make sure that all tasks are actually in an executing state, and one that they are waiting on.  It's *probably* not an issue.  I'd usually do that when doing concurrency testing.

rpdome (approved):
I think this is good for the current scope, but let's create a tracking work item and bring this up in the Q3 discussion next week to prevent race condition from happening in other places in the future.

cc: @shoatman

PR #1660, opened 2021-12-27
Add application identifier to access token

shoatman:
- Adding an application identifier field (string) to the access token cache record.
- The application identifier is a combination of package name of the application and signature of the application
- This is only required for use by the android auth broker as this is the only shared cache on Android
- Addition of this field distinguishes legitimate applications which use the same client id from one another where were we need to know if particular policies have been applied locally in a physical application

From a scenario perspective this change is required to address a True MAM scenario around legacy recommendations to applications for Single Sign On... when they were encouraged to use the same client id.

Please review the token cache documentation in the MSAL repo prior to reviewing.


common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -182,7 +182,11 @@ public static Signature[] getSignatures(final PackageInfo packageInfo) {
             if (packageInfo.signingInfo == null) {
                 return null;
             }
+            //Adding linke here: https://developer.android.com/reference/android/content/pm/SigningInfo#getSigningCertificateHistory()
+            //if the package has multiple signers the certificates may never be changed....
```
moumighosh, line 186: What is the relevance of who is signing these certificates? This is the broker package, the certificate used to sign this package would be issued by the Microsoft's internal root CA? Which I am assuming eSTS trusts. Is my assumption incorrect?

shoatman: I think you followed up in Teams chat on this one.  The reason I linked to this... is that if the signature did change that would mean the app would no longer have access to any prior tokens from the cache.  Multiple signatures is uncommon but alos a potential complication.... when we chatted with Google... they basicallyi told us they don't test all of the AOSP features relative to multi-signing.... which seems to indicate that folks don't rely on it in practice.

rpdome:
@shoatman  would you mind describing in detail
1. the reason/scenario that drives this change? 
2. How to navigate through the flow affected by the change (IMHO, the cache layer is pretty convoluted).

I think it'll be helpful for reviewers, and as a future reference.

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
```diff
@@ -158,6 +160,7 @@ public MsalOAuth2TokenCache getTokenCache(final IPlatformComponents context,
                 EXPIRES_ON,
                 SECRET,
                 CLIENT_ID,
+                APPLICATION_IDENTIFIER,
                 SECRET,
                 MOCK_ID_TOKEN_WITH_CLAIMS,
                 "1",
```
rpdome, line 169: We might want to add coverage of the following cases.

This is what I see - please correct me if i'm wrong.

1. populate cache *with* the identifier -> try to read with a specific identifier = covered. (seems like you kind of converted every tests in OAuth2TokenCache test classese in to this)
2. populate cache without the identifier -> try to read with a specific identifier = not covered.
3. populate cache *with* the identifier -> try to read without a specific identifier = not covered. (do we need this? This is applicable to probably only rollout scenario.)

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -184,6 +188,15 @@ protected List<Credential> getCredentialsFilteredByInternal(@Nullable final Stri
                 matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(clientId, credential.getClientId());
             }
 
+            if (mustMatchOnApplicationIdentifier) {
+                if (credential instanceof AccessTokenRecord) {
+                    final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
+                    matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(applicationIdentifier, accessToken.getApplicationIdentifier());
```
rpdome, line 199: if applicationIdentifier is set (and therefore mustMatchOnApplicationIdentifier == true), but accessToken.getApplicationIdentifier() is null, would this fail?

This seems like the upgrade scenario.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -582,6 +585,7 @@ public ICacheRecord load(@NonNull final String clientId,
         } else {
             resultRecord = targetCache.load(
                     clientId,
+                    applicationIdentifier,
```
rpdome, line 591: where are we saving this value into the cache?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
```diff
@@ -79,11 +81,27 @@ public class CommandParameters {
     @Expose()
     private boolean powerOptCheckEnabled;
 
+    //Marking as final excludes from lombok generated builder
+    @Expose()
+    private final String applicationIdentifier;
```
rpdome: nit: do we need this?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -56,12 +55,6 @@ public void validate() throws ArgumentException {
                     "mCallerUId", "Caller Uid is not set"
             );
         }
-        if (StringUtil.isNullOrEmpty(callerPackageName)) {
```
rpdome, line 80: why removing this check?

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
shoatman: It's going to be a cache miss... which will result in a new token request.  I didn't think a "migration" solution made sense in this case.  thoughts?

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
shoatman: Agree.  I'll add some additional coverage.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
shoatman: > @shoatman would you mind describing in detail
> 
> 1. the reason/scenario that drives this change?
> 2. How to navigate through the flow affected by the change (IMHO, the cache layer is pretty convoluted).
> 
> I think it'll be helpful for reviewers, and as a future reference.

Happy to provide additional context.  I think I added some design documentation on the cache to common already as a separate PR that Brian reviewed and approved.  I'll add a link to that.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
shoatman: I'm actually not sure at this point.  I'll go back and check if there was a reason.  It may have been a whoops.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
shoatman: Nope.  I'll double check that it's not actually be used somewhere before I remove it.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
rpdome: If we decide to go that route, I think we should call out that this is sort of a breaking change - so that it doesn't come as a surprise.

(Personally, I don't think doing an additional sign in is a big deal, but I can't speak for others, and I don't know if there is any weird scenario out there that would make this a painful experience for user.)

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
rpdome: Dug into the code... this is very not obvious... the modification to the "save" happens in generateCacheKey().

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
shoatman: I wasn't thinking it would be an additional sign in... as we're talking about access tokens, not refresh tokens at the moment.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
shoatman: Confirmed.  Not needed.  Removed.

shoatman:
> @shoatman would you mind describing in detail
> 
> 1. the reason/scenario that drives this change?
> 2. How to navigate through the flow affected by the change (IMHO, the cache layer is pretty convoluted).
> 
> I think it'll be helpful for reviewers, and as a future reference.

I added additional information to the description.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
shoatman: I moved these all the way down to CommandParameters.... so if we want a validation check on these It should be there.  I'll see if it makes sense to add it there.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -140,6 +142,8 @@ protected List<Credential> getCredentialsFilteredByInternal(@Nullable final Stri
         final boolean mustMatchOnRealm = !StringUtil.isNullOrEmpty(realm);
         final boolean mustMatchOnTarget = !StringUtil.isNullOrEmpty(target);
         final boolean mustMatchOnClientId = !StringUtil.isNullOrEmpty(clientId);
+        final boolean mustMatchOnApplicationIdentifier = !StringUtil.isNullOrEmpty(applicationIdentifier);
+        final boolean mustMatchOnMamEnrollmentIdentifier = !StringUtil.isNullOrEmpty(mamEnrollmentIdentifier);
```
rpdome, line 146: Once the MAM wipe happens, would CP return an empty MAM ID?

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
shoatman: That's correct.  No enrollment id would be returned.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
```diff
@@ -41,9 +41,6 @@
 public class BrokerInteractiveTokenCommandParameters extends InteractiveTokenCommandParameters
         implements IHasExtraParameters, IBrokerTokenCommandParameters {
 
-    @Expose
-    private final String callerPackageName;
```
shahzaibj, line 45: Why remove this?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
```diff
@@ -111,12 +108,6 @@ public void validate() throws ArgumentException {
                         "mCallerUId", "Caller Uid is not set"
                 );
             }
-            if (StringUtil.isNullOrEmpty(callerPackageName)) {
```
shahzaibj, line 114: Why remove this?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -38,9 +38,6 @@
 @EqualsAndHashCode(callSuper = true)
 public class BrokerSilentTokenCommandParameters extends SilentTokenCommandParameters implements IBrokerTokenCommandParameters {
 
-    @Expose
-    private final String callerPackageName;
```
shahzaibj, line 42: Why remove this?

shahzaibj: Ok, moved it to parent class. #Resolved.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
shahzaibj: Ok, moved it to parent class. #Resolved.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
shahzaibj: Ok, moved it to parent class. #Resolved.

shahzaibj (approved):

PR #1539, opened 2021-08-24
Add ropc command and ropc flow to BaseController

shahzaibj:
Related ADAL PR: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1638
Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1496

Also added E2E tests at the controller level to test ropc flow.

PR #1661, opened 2021-12-28
Fix accidental code change that disabled PoP for auth code grant flow

shahzaibj:
Accidentally had made the change in wrong location while addressing the following comment on original PR: 
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1539#discussion_r750633862


shoatman (approved):
lgtm.

AdamBJohnsonx (approved):
I do think we should have a test here that would break if this happened again.

PR #1662, opened 2022-01-04
Avoid key stomping

shoatman:
Multiple apps as of Today (1/4/2022) can still share a linux user id, by configuring the sharedUserId attribute in their
            Android Manifest file.  If multiple apps reference the same value for sharedUserId and are signed with the same keys
            they will use the same AndroidKeyStore and may obtain access to the files and shared preferences of other applications
            by invoking createPackageContext.
            Support for sharedUserId is deprecated, however some applications still use this Android capability.
            See: https://developer.android.com/guide/topics/manifest/manifest-element
            To address apps in this scenario we will attempt to load an existing KeyPair instead of immediately generating
            a new key pair.  This will use the same keypair to encrypt the symmetric key generated separately for each
            application using a shared linux user id... and avoid these applications from "stomping"/overwriting
            one another's keypair

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
@@ -715,6 +715,29 @@ public class StorageHelper implements IStorageHelper {
             final Locale currentLocale = Locale.getDefault();
             applyKeyStoreLocaleWorkarounds(currentLocale);
 
+            /*
+            !!WARNING!!
+
+            Multiple apps as of Today (1/4/2022) can still share a linux user id, by configuring the sharedUserId attribute in their
+            Android Manifest file.  If multiple apps reference the same value for sharedUserId and are signed with the same keys
+            they will use the same AndroidKeyStore and may obtain access to the files and shared preferences of other applications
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -715,6 +715,29 @@ private synchronized KeyPair generateKeyPairFromAndroidKeyStore()
...
+            Support for sharedUserId is deprecated, however some applications still use this Android capability.
+            See: https://developer.android.com/guide/topics/manifest/manifest-element
+
+            To address apps in this scenario we will attempt to load an existing KeyPair instead of immediately generating
+            a new key pair.  This will use the same keypair to encrypt the symmetric key generated separately for each
+            application using a shared linux user id... and avoid these applications from "stomping"/overwriting
+            one another's keypair
+             */
+
+            KeyPair existingPair = readKeyPair();
```
shahzaibj, line 735: nit: final

shahzaibj (approved):

iambmelt (approved):
Nice find - didn't know anybody was actually using `sharedUserId`!

PR #1664, opened 2022-01-04
Add a code formatter action to our repository

AdamBJohnsonx:
Add a code formatter that runs on every pr action.

This is roughly equivalent to what happens in the OneAuth/MSAL-cpp codebase.  Every time you check in, we run a formatter on the code and if there is a difference it creates another commit and checks it in.  You have to review these commits, because we don't trust the formatter, but with the exception of this giant review they should be small and easy to follow.

In addition to the formatter running, moving lines around broke several suppressions.  Rather than move them, I just either fixed them in the case of the extra-string-creation pieces, or added a suppression comment that won't break when the format runs.

The benefits of using code formatters is that they will help with merges.  If you run the same code formatter on your branch, then the merge difference will be smaller, particularly when you're messing with import statements and other silly things.  This also puts us closer to having AOSP style enforcement.

If you're working on something that's spanning this change, an easy way to work around it if you don't want to just get the command line tool and run it is to just take the formatter yml file into your branch, altered to run on your branch instead of dev, check it in, pull the result, and remove the formatter file.


shahzaibj:
What's the best way to review this PR? Specific files I should be looking at?

AdamBJohnsonx:
> What's the best way to review this PR? Specific files I should be looking at?

I'd generally scan it for sanity.  The only change *I* made was in the .yml file, and the action reformatted everything else.

PR #1667, opened 2022-01-05
Testing license addition

AdamBJohnsonx:


shahzaibj:
It is adding as a double comment...by wrapping comment in a comment. I suggesting updating the HEAEDER file to make it uncommented text.

AdamBJohnsonx:
> It is adding as a double comment...by wrapping comment in a comment. I suggesting updating the HEAEDER file to make it uncommented text.

This was a test... it looked like this is what it was going to do.  We'll have to fork the action and change the code, since it is only capable of writing block comments and we use line comments.

PR #1668, opened 2022-01-12
Update exception name to match older (pre-refactoring) value to avoid breaking older msal clients

iamgusain:
### What
Update exception name to match older (pre-refactoring) value to avoid breaking older msal clients

### Why
With refactoring the exception classes have moved to a new package, but older clients still use older package name to compare the exception type from broker result bundle. So this breaks the scenario when Broker is using the new package  (refactored code0 but Msal client is on older version

### How
Changing the exception name to match older (pre-refactoring) value to avoid breaking older msal clients

### Testing
Verified locally with Msal test app and Broker Host app

### Other
MsalBrokerResultAdapter class code for comparing (for reference)
image

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
@@ -24,7 +24,7 @@ package com.microsoft.identity.common.java.exception;
 
 public class ArgumentException extends BaseException {
 
-    public static final String sName =  ArgumentException.class.getName();
+    public static final String sName =  "com.microsoft.identity.common.exception.ArgumentException";
     private static final long serialVersionUID = -6399451133831073876L;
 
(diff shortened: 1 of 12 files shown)


shahzaibj (approved):
Good find. Thanks @iamgusain

common4j/src/main/com/microsoft/identity/common/java/exception/ArgumentException.java
```diff
@@ -24,7 +24,7 @@
 
 public class ArgumentException extends BaseException {
 
-    public static final String sName =  ArgumentException.class.getName();
+    public static final String sName =  "com.microsoft.identity.common.exception.ArgumentException";
```
rpdome, line 31: nit: also add comment here.

rpdome (approved):
LGTM. Could you please also add the problematic MSAL adapter code to the PR description for reference? Thanks!

Issue #1669, opened 2022-01-16
"withFragment" does not work, I need help / a sample of how to use "withFragment" to embed MSAL login into my app


VansonLeung:
This issue only occurs when using: 

    parametersBuilder.withFragment(fragment)

```

        var parametersBuilder = AcquireTokenParameters.Builder()
            .startAuthorizationFromActivity(activity)
            .fromAuthority(authorityUrl)
            .withScopes(scopes)
            .withPrompt(Prompt.LOGIN)
            .withCallback(acquireTokenCallback)

            parametersBuilder = parametersBuilder.withLoginHint(loginHint)
            parametersBuilder = parametersBuilder.withFragment(fragment)   // key factor of reproducing the issue
            parametersBuilder = parametersBuilder.withPrompt(msalPromptType)
            parametersBuilder = parametersBuilder.withAuthorizationQueryStringParameters(extraQueryParameters)

        val parameters = parametersBuilder.build()


        getB2CApp { b2cApp ->
            b2cApp.acquireToken(parameters)
        }
```

The result is that the AuthorizationActivity was "finish()"ed immediately after AuthorizationFragment runs "onCreate()"

Here's the MSAL debug log:

```
Performing cloud discovery Android 28
Performing cloud discovery... Android 28
Acquired lock. Android 28
Cloud discovery complete. Android 28
Cloud is known. Android 28
Use webView for authorization. Android 28
Creating AuthorizationRequestBuilder Android 28
Creating AuthorizationRequestBuilder... Android 28
Perform the authorization request with embedded webView. Android 28
No stored state. Unable to handle response Android 28
N/A Android 28
Hosting Activity is destroyed before Auth request is completed, sending request cancel Android 28
Sending result from Authorization Activity, resultCode: 2008 Android 28
Launching embedded WebView for acquiring auth code. Android 28
The start url is null Android 28
Error is returned from webview redirect Android 28
```

By removing:

            parametersBuilder = parametersBuilder.withFragment(fragment)

, the MSAL login works immediately

```
Performing cloud discovery Android 28
Performing cloud discovery... Android 28
Acquired lock. Android 28
Cloud discovery complete. Android 28
Cloud is known. Android 28
Use webView for authorization. Android 28
Creating AuthorizationRequestBuilder Android 28
Creating AuthorizationRequestBuilder... Android 28
Perform the authorization request with embedded webView. Android 28
Launching embedded WebView for acquiring auth code. Android 28
[...]

PR #1672, opened 2022-01-21
[Fix] enable custom fragment logic

VansonLeung:
The custom fragment logic should fulfill the intent bundle parameters 

Otherwise the fragment will destroy itself immediately after auth creation, thus rendering this function useless.

Related issue: #1669


shoatman:
@rpdome - Please review when you have a chance.  I'll take a look today.  Thanks

shoatman:
There seems to be an existing method to set the state; however it does not appear be used currently.  getAuthorizationFragmentFromStartIntentWithState (found in AuthorizationActivityFactory)

VansonLeung:
It does not appear that end users could directly use this function publicly and it was not documented either.

```
getAuthorizationFragmentFromStartIntentWithState
```

I am not sure the use of this function.

Unless someone has figured out what 

```
getAuthorizationFragmentFromStartIntentWithState
```


 is for - I think my fix is not a dirty fix.

PR #1681, opened 2022-02-21
Broker telemetry with Aria 

gitaumoses4:


common/src/main/java/com/microsoft/identity/common/internal/telemetry/relay/AriaTelemetryRelayClient.java
```diff
@@ -0,0 +1,107 @@
...
+import com.microsoft.identity.common.java.telemetry.relay.AbstractTelemetryRelayClient;
+import com.microsoft.identity.common.java.telemetry.relay.TelemetryRelayClientException;
+
+import java.util.Map;
+
+/**
+ * A relay client to send events to Aria
+ */
+public class AriaTelemetryRelayClient extends AbstractTelemetryRelayClient {
+    private static final String ARIA_TABLE = "android_event";
```
bktowett, line 42: Will all the events go to this table? We should also have names that can aid in identifying which auth client team is sending the events as we are using a shared cluster. Consider changing this to something like "msal.android.android_events"

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/TelemetryRelayClientException.java
```diff
@@ -0,0 +1,50 @@
...
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.experimental.Accessors;
+
+@EqualsAndHashCode(callSuper = true)
+@Data()
+@Accessors(prefix = "m")
+public class TelemetryRelayClientException extends Exception {
+    private static final long serialVersionUID = -1543623857511895210L;
```
bktowett, line 36: Is this value,-1543623857511895210L, arbitrary?

shahzaibj:
Please add a description to the PR.

PR #1685, opened 2022-03-01
Merge common v4.0.1 release to dev

iamgusain:
Merging release/4.0.0 branch back to dev


common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
```diff
@@ -715,6 +715,29 @@ private synchronized KeyPair generateKeyPairFromAndroidKeyStore()
             final Locale currentLocale = Locale.getDefault();
             applyKeyStoreLocaleWorkarounds(currentLocale);
 
+            /*
```
rpdome, line 718: Hm. This does not seem to be ported to common4j (AndroidWrappedKeyLoader, line 249)

StorageHelper is already marked as deprecated and won't be used in 4j.

common/src/main/java/com/microsoft/identity/common/internal/util/StringUtil.java
```diff
@@ -135,6 +137,15 @@ public static String join(final char delimiter, @NonNull final Iterable<String>
                 && !StringExtensions.isNullOrBlank(uidUtidArray[INDEX_UTID])) {
             uid = uidUtidArray[INDEX_UID];
             utid = uidUtidArray[INDEX_UTID];
+        } else {
```
rpdome, line 140: This method has been migrated to 4j (same class name), and this change should go there as well.

iamgusain: added

common/src/main/java/com/microsoft/identity/common/adal/internal/cache/StorageHelper.java
iamgusain: Lets do that in a separate change

shoatman (approved):

rpdome (approved):

PR #1688, opened 2022-03-08
Melissaahn/clear cert pref

melissaahn:
*Please read the "Important Note" section below*
## The problem 
Addressing bug: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1776683 
On an Android device without broker, if the first user account is authenticated with CBA, attempting to authenticate subsequent user accounts with CBA fails. Additionally, if a user account ever backs out of the cert picker (the "deny" button), subsequent CBA attempts will result in an error. 
## Why?
There are three main [`ClientCertRequest`](https://developer.android.com/reference/android/webkit/ClientCertRequest) methods that can be called on the client side when the user makes a choice on the cert picker: [`proceed`](https://developer.android.com/reference/android/webkit/ClientCertRequest#proceed(java.security.PrivateKey,%20java.security.cert.X509Certificate[])), [`cancel`](https://developer.android.com/reference/android/webkit/ClientCertRequest#cancel()), and [`ignore`](https://developer.android.com/reference/android/webkit/ClientCertRequest#ignore()). We currently only use `proceed` and `cancel`. For `proceed` and `cancel`, Chromium caches the client's choice for the remainder of the app's session. This results in the last choice (whether that's proceeding with a particular cert or choosing to cancel the flow) will be automatically selected whenever the user attempts CBA afterwards. 
## What I did 
The static WebView method [`clearClientCertPreferences`](https://developer.android.com/reference/android/webkit/WebView#clearClientCertPreferences(java.lang.Runnable)) clears the cached selection made with [`ClientCertRequest`](https://developer.android.com/reference/android/webkit/ClientCertRequest). 
It took some trial and error to find the right place for the method call.  
- Calling `clearClientCertPreferences` right after a `proceed` or `cancel` call (which are both running asynchronously) resulted in either the cache being cleared too early or an error where the cert picker would always appear a second time after authentication. 
- Calling the method at the end of `onReceivedClientCertRequest` resulted in the cache being cleared too early.
- Calling the method in `onChallengeResponseReceived` results in the cert picker appearing twice (similar to first bullet point)

~~It seemed that I needed to call the method as late as possible in order to ensure that `proceed`/`cancel` would finish first.
The most successful option thus far is clearing the cert preferences cache whenever the WebView reloads a url with a prefix of "https://certauth". The method is called in `onPageLoaded` callback in `onCreateView` in the fragment `WebViewAuthorizationFragment`. 
However, calling `clearClientCertPreferences` introduced a new problem: whenever `cancel` was called, the cert picker would pop up rapidly no less than three times (with the last time being cached). I found that replacing `cancel` with `ignore` seemed to mostly fix the issue. According to [Chromium Code Search](https://source.chromium.org/chromium/chromium/src/+/main:android_webview/java/src/org/chromium/android_webview/AwContentsClientBridge.java;drc=48a3eaf6f8ae83f7c99851f0b9088365e3edf6e6;l=120?q=clientcertrequest%20ignore&ss=chromium), the only difference between the two methods on the client side is that `cancel` is cached in a lookup table.~~  **See Important Note below.**
### Important Note
~~I have noticed that sometimes, the cert picker will exhibit weird behavior (popping up three times) even with `ignore`. This happened for a period of time while I was testing one day. On that same day (but an hour later) the cert picker was working normally again. I do suspect that this might be some concurrency issue on the client side, but I'm not totally sure. 
Therefore, if you have suggestions on how to improve this fix, *please* let me know!~~
Update: Per Dome's suggestion, `clearClientCertPreferences` is now called when the `WebViewAuthorizationFragment` is created, which is always _before_ CBA starts. To account for a failed authentication or exiting out of the cert picker, `clearClientCertPreferences` is also called in `onBackPressed`.
## How I tested
In AzureSample:
1. Go to Multiple Account Mode
2. Select "Get Graph Data Interactively"
3. Use "vimala@vimrang1.onmicrosoft.com" as upn
4. Select "Sign in with a certificate"
5. From cert picker, select cert with CN=vimala
6. Graph data should show. Click "Get Graph Data Interactively" again
7. Use "mel@vimrang1.onmicrosoft.com" as upn
8. Select "Sign in with a certificate"
9. From cert picker, select "Deny"
10. Should get an "Invalid Request" error. Click back button
11. Use "mel@..." as upn again and select "Sign in with a certificate"
12. From cert picker, select cert with CN=mel
13. Graph data for mel should show

Any time "Sign in with a certificate" is selected, a cert picker should show.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -28,7 +28,7 @@ codeCoverageReport {
     coverage.enabled = enableCodeCoverage
 }
 
-def common4jVersion = "4.0.1-CBAFix"
+def common4jVersion = "0.0.+"
 if (project.hasProperty("distCommon4jVersion") && project.distCommon4jVersion != '') {
     common4jVersion = project.distCommon4jVersion
 }
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -28,7 +28,7 @@ codeCoverageReport {
     coverage.enabled = enableCodeCoverage
 }
 
-def common4jVersion = "0.0.+"
+def common4jVersion = "4.0.1-CBAFix"
 if (project.hasProperty("distCommon4jVersion") && project.distCommon4jVersion != '') {
     common4jVersion = project.distCommon4jVersion
 }
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -63,7 +63,7 @@ import static com.microsoft.identity.common.adal.internal.AuthenticationConstant
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_CONTROLS_ENABLED;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_ENABLED;
 
-import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.CERTAUTH_PREFIX;
+import static com.microsoft.identity.common.java.AuthenticationConstants.AAD.CERTAUTH_PREFIX;
 
 /**
  * Authorization fragment with embedded webview.
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -96,6 +96,7 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        final String methodTag = TAG + ":onCreate";
         final FragmentActivity activity = getActivity();
         if (activity != null) {
             WebViewUtil.setDataDirectorySuffix(activity.getApplicationContext());
@@ -106,7 +107,7 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             WebView.clearClientCertPreferences(null);
         } else {
-            Logger.info(TAG, null, "onCreate:" + "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
+            Logger.warn(methodTag, "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
         }
     }
 
@@ -215,7 +216,7 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
                 WebView.clearClientCertPreferences(null);
             } else {
-                Logger.info(TAG, null, "onBackPressed:" + "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
+                Logger.warn(methodTag, "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
             }
         } else {
             cancelAuthorization(true);
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -36,6 +36,7 @@ import android.widget.ProgressBar;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.FragmentActivity;
 
@@ -61,6 +62,8 @@ import static com.microsoft.identity.common.adal.internal.AuthenticationConstant
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.REQUEST_URL;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_CONTROLS_ENABLED;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_ENABLED;
+
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.CERTAUTH_PREFIX;
 
 /**
  * Authorization fragment with embedded webview.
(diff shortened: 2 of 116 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -167,6 +171,11 @@ public void onPageLoaded(final String url) {
                                 mWebView.loadUrl("javascript:" + javascriptToExecute[0].replace("%", "%25"));
                             }
                         }
+
+                        //clearcerts if url has certauth
```
iamgusain: You can add more context for why we are clearing the certPreference here ( copy paste from your PR description here, or add the bug link)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -69,7 +69,7 @@ public Void processChallenge(@NonNull final ClientCertRequest request) {
                     public void alias(String alias) {
                         if (alias == null) {
                             Logger.info(TAG, "No certificate chosen by user, cancelling the TLS request.");
-                            request.cancel();
+                            request.ignore();
```
iamgusain: What is the difference between cancel and ignore?

common/build.gradle
```diff
@@ -28,7 +28,7 @@ codeCoverageReport {
     coverage.enabled = enableCodeCoverage
 }
 
-def common4jVersion = "0.0.+"
+def common4jVersion = "4.0.1-CBAFix"
```
iamgusain: remove this before merging your PR.

versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Tue Apr 06 22:55:08 UTC 2021
-versionName=4.0.0-RC6
+versionName=4.0.1-CBAFix2
```
iamgusain: remove changes to this file before merging PR.

iamgusain (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: On client side, only difference is that the user choice isn't cached. For some reason, the combination of `cancel` and `clearClientCertPreferences` resulted in multiple popups of the cert picker upon selection (maybe some sort of concurrency issue?). Changing to `ignore` resulted in better behavior. https://source.chromium.org/chromium/chromium/src/+/main:android_webview/java/src/org/chromium/android_webview/AwContentsClientBridge.java;drc=48a3eaf6f8ae83f7c99851f0b9088365e3edf6e6;l=120?q=clientcertrequest%20ignore&ss=chromium

shoatman (approved):
lgtm - Let's add the test case you mentioned in the PR description to our manual test plan for MSAL and Broker release.

rpdome:
I'm wondering if we can clear the cert pref in onCreate, or in AzureActiveDirectoryWebViewClient's constructor.

in other words - can we do this as a part of webview bootstrapping process - no need to even look up for https://certAuth. That way - we shouldn't have to deal with potential race condition, since it'll be done before everything starts.

(unless there is a scenario which allows user to "switch cert" in the middle of a webview session. Please let me know if that's the case.)

melissaahn:
> I'm wondering if we can clear the cert pref in onCreate, or in AzureActiveDirectoryWebViewClient's constructor.
> 
> in other words - can we do this as a part of webview bootstrapping process - no need to even look up for https://certAuth. That way - we shouldn't have to deal with potential race condition, since it'll be done before everything starts.
> 
> (unless there is a scenario which allows user to "switch cert" in the middle of a webview session. Please let me know if that's the case.)

Clearing the cache beforehand definitely solves the race condition issue, which is great. The one minor issue I've encountered with clearing in the beginning is if the user either gets an error while authenticating or cancels out of the cert picker, the choice will still be cached until the user is able to back out of the webview session and start a new one. 
That being said, I still think your suggestion is cleaner, so I'm willing to make the changes. @rpdome what do you think?

rpdome:
> > I'm wondering if we can clear the cert pref in onCreate, or in AzureActiveDirectoryWebViewClient's constructor.
> > in other words - can we do this as a part of webview bootstrapping process - no need to even look up for https://certAuth. That way - we shouldn't have to deal with potential race condition, since it'll be done before everything starts.
> > (unless there is a scenario which allows user to "switch cert" in the middle of a webview session. Please let me know if that's the case.)
> 
> Clearing the cache beforehand definitely solves the race condition issue, which is great. The one minor issue I've encountered with clearing in the beginning is if the user either gets an error while authenticating or cancels out of the cert picker, the choice will still be cached until the user is able to back out of the webview session and start a new one. That being said, I still think your suggestion is cleaner, so I'm willing to make the changes. @rpdome what do you think?

would (also) invoking the method in onBackPressed() solve that issue?

rpdome:
not sure if there will be an unintended side effects though.. (i.e. the user gets through the CBA, and is far deep into the flow, and somehow accidentally click back button - in which case they'll be prompted again if any subsequent CBA is required).

@shoatman what do you think? Am I overthinking this? 🤣

melissaahn:
calling the method in onBackPressed() within WebViewAuthorizationFragment works! I also tested clicking the back button before the authentication can complete - the next attempt to authenticate does get messed up, but the third attempt will work. 
One thing i forgot to mention is that the annotation `@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP) ` needs to be added to every method that contains clearClientCertPreferences. I know that the current min api is 16, so would this be ok to add to onBackPressed() and the WebViewAuthorizationFragment constructor?

rpdome:
> calling the method in onBackPressed() within WebViewAuthorizationFragment works! I also tested clicking the back button before the authentication can complete - the next attempt to authenticate does get messed up, but the third attempt will work. One thing i forgot to mention is that the annotation `@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP) ` needs to be added to every method that contains clearClientCertPreferences. I know that the current min api is 16, so would this be ok to add to onBackPressed() and the WebViewAuthorizationFragment constructor?

Could you please elaborate on how the 2nd attempt is messed up? Is it still suffering from race condition?

If it's so, maybe we could try clearing the cache right after cancel() or ignore() is invoked? (Ignore shouldn't persist anything... but you still observe the issue occasionally - maybe it's an OS bug)

re: RequiresApi - It should be fine, but you could also put it in a block that only executes if the API version is > X.

```Java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
   // do work.
} else {
   // log. shouldn't happen though.
}
```

melissaahn:
For testing the back button during flow, I tested both with and without clearing the cert choice cache. In both cases, no matter if the cert picker comes up the second time or not, I get an "invalid request. Request is malformed or invalid" error. I think this error is actually independent of the clearClientCertPreferences call, so from my end, making the calls in the two places we discussed would be a complete solution :o
Btw, calling clearClientCertPreferences immediately after cancel() or ignore() resulted in that pesky cert picker pop up bug.
I'll add in that if/else block instead of the annotation. Thanks!

PR #1689, opened 2022-03-08
port key stomping fix

rpdome:
Port https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1662 into the new common4j class - since StorageHelper is no longer used.


iamgusain (approved):
Nit: Update changelog as well

tanmaymanolkar1 (approved):

PR #1692, opened 2022-03-11
Fix error in pipelines that build common using Dist

p3dr0rv:
## What is this

remove vsts-maven-adal-android in buidl.gradle
## Why

[CD pipelines fail](https://identitydivision.visualstudio.com/Engineering/_build?definitionId=1515&_a=summary) at building common using Dist variant because the  vsts-maven-adal-android repository does not have some dependencies

## Test
image

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=890406&view=results

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,38 +22,19 @@ allprojects {
     Order of repositories is important.
      */
     repositories {
-        /*
-        This is the project main repository and contains all of the packages we publish.
-        This azure devops feed also has an upstream to Maven Central as well as several other
-        internal azure devops respositories
-
-        NOTE: This repository is not currently available to our public CI (Travis)
-         */
+        google()
+        mavenCentral()
+        // Adding here because Travis cannot access AndroidADAL feed.
+        maven {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -37,14 +37,6 @@ allprojects {
                 password project.vstsPassword
             }
         }
-        maven {
-            name "vsts-maven-adal-android"
```
iamgusain, line 41: This where our packages are published, we wouldn't be able to get internal packages from here, if we remove.
Is the issue because.. the name "vsts-maven-adal-android" is already used?

iamgusain:
I think changing the order for now, should unblock this.

build.gradle
iamgusain: I think changing the order for now should unblock this

iamgusain (approved):

rpdome:
@p3dr0rv  could you please update this PR to reflect the change in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1695/files?

This is to make all of our repo sources consistent across ad-accounts, common and MSAL.

rpdome (approved):

PR #1700, opened 2022-03-23
Fixed setAttestationChallenge to pass null rather than an empty byte …

shoatman:
An attestation challenge should be provided or set to null.  Prior to this fix, it was set to an empty byte array and in some strict KeyMaster implementations the empty byte array resulted in a runtime exception.  Specfically the KeyMaster implementations for the Google Pixel 6 and 6 Pro.


shoatman:
@iambmelt - FYI

iamgusain (approved):

rpdome (approved):

PR #1701, opened 2022-03-23
Ensure consistent tags in our logs

gitaumoses4:
## What does this PR do?
- Updates all TAGs in common to use the format. `{className}:{methodName}`

PR #1702, opened 2022-03-24
IllegalStateException when removing AuthorizationFragment from fragment manager (1695851)

shoatman:
Some clients provide their own activity and FragmentManager for us to render our authorization fragment into.  When we are finished with the authorization fragment we attempt to remove ourselves from the provided FragmentManager.  Telemetry provided by some of our consumers shows that in some cases an IllegalStateException is thrown on the call to the FragmentManager to "commit()" the removal of our authorization fragment.  

Adding a link to Android document on [Fragment transactions](https://developer.android.com/guide/fragments/transactions)

Android documentation indicates that commit() should work in the "vast majority" of cases; however in this case it appears to fail as indicated above for some clients.  As a result, this PR adopts the usage of the "commitNow()" method and in addition wraps the calling code in an exception handler in order to be able to log any exception and swallow it.  We're swallowing the exception since this is a clean up step and should not negatively impact the provided activity / fragment manager if it's already been disposed of.


rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java
```diff
@@ -104,20 +104,34 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
     }
 
     void finish() {
+        final String methodName = "#finish";
```
gitaumoses4, line 107: For consistency, we updated the tags to use a colon to define the `methodName`.
Moreso, we can have:
```java
final String methodTag = TAG +":finish";
```
in case the tag is needed in multiple locations.

gitaumoses4: See: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1701

iamgusain (approved):

PR #1706, opened 2022-03-29
Fix null pop manager reference when request is made to broker for pop authentication scheme

shoatman:
Like the clock skew manager the device pop manager needs to be set when a broker request is deserialized from the intent bundle.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,11 +8,12 @@ V.Next
 - [PATCH] Fix accidental code change that disabled PoP for auth code grant flow (#1661)
 - [MINOR] Add flighting parameters to commmandParameters (#1562)
 - [MINOR] Add ropc command and ropc flow to BaseController (#1539)
+- [PATCH] Ensure a device pop manager is provided when PoPAuthenticationScheme is requested of the broker (#1814728)
 
 Version 4.0.4
 ----------
 - [PATCH] Adding orientation flag to BrokerActivity android:configChanges to prevent it from getting restarted on orientation change (#1705)
-- [PATCH] Correct setAttestationChallenge to provide null rather than empty byte array (#1700)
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -8,11 +8,12 @@ V.Next
 - [PATCH] Fix accidental code change that disabled PoP for auth code grant flow (#1661)
 - [MINOR] Add flighting parameters to commmandParameters (#1562)
 - [MINOR] Add ropc command and ropc flow to BaseController (#1539)
+- [PATCH] Ensure a device pop manager is provided when PoPAuthenticationScheme is requested of the broker (#1814728)
```
shahzaibj: What does this number mean? (#1814728)

This number generally represents the pull request id however, that doesn't seem to be the case here

rpdome (approved):

changelog.txt
```diff
@@ -8,11 +8,12 @@ V.Next
...
 
 Version 4.0.4
 ----------
 - [PATCH] Adding orientation flag to BrokerActivity android:configChanges to prevent it from getting restarted on orientation change (#1705)
-- [PATCH] Correct setAttestationChallenge to provide null rather than empty byte array (#1700)
+- [PATCH] Correct setAttestationChallenge to provide null rather than empty byte array (#1814725)
```
shahzaibj: Is this change meant to be in this PR?

changelog.txt
shoatman: Sorry, I was using the bug id... which I thought made more sense.  Sorry about that.

changelog.txt
shoatman: See above for my confusion.

shahzaibj (approved):

iamgusain (approved):

PR #1711, opened 2022-04-04
Implement cert loader for both multiple and legacy WPJ data store in PKeyAuth

rpdome:
spec: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2726

### Background:
This particular change is dealing with one of the device authentication challenge method that ADAL/MSAL/Broker supports - this is called [PKeyAuth](https://microsoft.sharepoint.com/teams/aad/devex/Shared%20Documents/Core%20Design/Device%20authentication%20using%20PKeyAuth.docx?web=1). The gist of the protocol is that eSTS would send us a nonce, we created a signed JWT with the keys+cert associated to the device - which associates to the tenants the sign-in accounts was created (home tenant) - and return it to eSTS.

Currently, the broker assumes that there is only one certificate (if exists), and therefore it would preload the cert into the memory in the broker code, such that the code in common could use it to create the PKeyAuth response.

(NOTE: this also means that on Android, PKeyAuth can only be used by the broker itself - since no other apps will not have access to certificate data).

In order to support multiple WorkplaceJoin Certificates, we should be able to respond to device authentication challenge with a corresponding cert. The server would return tenant Id as part of the PKeyAuth challenge (will be done later in another PR), and we should be able to use the tenant Id to perform a cert lookup - therefore we can no longer preload the cert.

### Changes:
1. Instead of preloading, we are going to preload an `IDeviceCertificateLoader` object instead. This object will give the common code an access to the WPJ certificates when it needs to.
     - In this change, we're going to introduce this new interface, but we will not turn it on yet (since the Multiple WPJ work is not done). We'll return different instances based on the feature flag in `WorkplaceJoinControllerFactory`.
     - This means that after this change is checked in, there is no change feature-wise (yet), and the existing flow should work as expected.
2. `DeviceCertProxy` and associated codes are removed.
    - this also means we'll officially stop supporting PKeyAuth in ADAL (which should not break anyone).
3. Add tests around `PKeyAuthChallenge` - this is mostly migrated from ADAL (due to the code removal in 2.)

### Tests
Manual tests - making sure the existing PKeyAuth flow works as expected
1. Perform WPJ in brokerhost
2. launch MSAL and perform acquireToken with..
     - NO login hint
     - device id claim
5. In WebView, enter the login hint you used to do WPJ and enter password.
6. MSAL should return a token with device id claim.


mohitc1 (approved):

p3dr0rv (approved):
LGTM

PR #1713, opened 2022-04-12
Add null check to avoid NPE when checking for AccountTypesWithManagementDisabled

iamgusain:
Adding null checks to avoid NPE when checking for AccountTypesWithManagementDisabled

The method getAccountTypesWithManagementDisabled is defined as Nullable. Adding a null check before using the return value in for loop to avoid NullPointerException being thrown in the for loop. as it results in a crash.

For more details see oneauth crash bug: https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20CPP/Backlog%20items/?workitem=1853626

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [PATCH] Add null check to avoid NPE when checking for AccountTypesWithManagementDisabled (#1713)
 - [MINOR] Add prompt=create support. (#1707)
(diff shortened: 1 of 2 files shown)


fadidurah (approved):
LGTM

somalaya (approved):

shahzaibj (approved):

PR #1718, opened 2022-04-15
Improve common Logger class

mohitc1:
**What**
Improvements in common java Logger.

**Why**
As part of effort to improve logging for better diagnosis and debugging.

**How**
1. Refactoring method to use "correct" param name and improve formatting diagnostic logs.
2. Some logs only have correlationId (for APIs which take correlationId as parameter). Add other diagnostic info as well (at this point it's thread_name) so that all logs are uniform.
3. Fixed a couple of callers which passed debugging message as part correlationId. This seemed like an improvement.

**Test**
Ran UTs.
Ran various broker scenarios like acquireToken from MSALTestApp, User based join from brokerHost, getdeviceMode, GetAccounts flows to see new logging 

Old example:
[2022-04-26 15:33:33 - thread_name : Thread-2, correlation_id : 627cb6d3-bfa6-4e83-8063-479b3ddb0a55]  Android 30 Initialized new DiagnosticContext
 [2022-04-26 15:33:33 - {"thread_id":"512","thread_name":"Thread-2","correlation_id":"627cb6d3-bfa6-4e83-8063-479b3ddb0a55","x-client-Ver":"3.0.1","x-client-SKU":"MSAL.Android"}]  Android 30 Starting request for correlation id : ##627cb6d3-bfa6-4e83-8063-479b3ddb0a55, with PublicApiId : 1200

New Example:
2022-04-26 15:36:03 - thread_name: Thread-2, correlation_id: fcfdf279-7622-42cc-af82-817f4fc17eb0 - Android 30] Initialized new DiagnosticContext
[2022-04-26 15:36:03 - thread_name: Thread-2, correlation_id: fcfdf279-7622-42cc-af82-817f4fc17eb0 - Android 30] Starting request with request context: {"thread_id":"518","thread_name":"Thread-2","correlation_id":"fcfdf279-7622-42cc-af82-817f4fc17eb0","x-client-Ver":"3.0.1","x-client-SKU":"MSAL.Android"}, with PublicApiId: 1200


common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -138,26 +138,14 @@ public static boolean setLogger(@NonNull String identifier,
...
-        }
-        if (StringUtil.isNullOrEmpty(correlationId)) {
-            correlationId = UNSET;
-        }
-
-        return DiagnosticContext.THREAD_NAME + " : "
-                + threadName + ", "
-                + DiagnosticContext.CORRELATION_ID + " : "
-                + correlationId;
+        return getDiagnosticContextMetadata(DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
```
shahzaibj, line 149: Seems like now this method only returns correlation id? If that's the case then we should probably update javadoc as well as the method name.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -138,26 +138,14 @@ public static boolean setLogger(@NonNull String identifier,
     /**
      * Get only the required metadata from the DiagnosticContext
      * to plug it in the log lines.
-     * Here we are considering the correlation_id and the thread_name.
+     * Here we are considering the correlation_id and the thread_name. Calls private getDiagnosticContextMetadata with
```
rpdome, line 141: seems like this one should also return thread name?

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -457,5 +446,26 @@ private static String getUTCDateTimeAsString() {
...
     }
+
+    /**
+     * Get only the required metadata from the DiagnosticContext to plug it in the log lines.
+     * Here thread_name is taken from DiagnosticContext, and correlationId from input parameter.
+     * The need for this is because DiagnosticContext contains additional metadata which is not always required to be logged.
+     *
+     * @return String The concatenation of thread_name and correlation_id to serve as the required metadata in the log lines.
+     */
+    private static synchronized String getDiagnosticContextMetadata(String correlationId) {
```
rpdome: suggest marking correlationId nullable, and we can use this method everywhere. (we don't need the other one).

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
mohitc1: It still returns correlationId and thread_name like it used to. As result of refactoring, it now calls into private method passing correlationId from request context. Documentation is updated to reflect this.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
mohitc1: From code it seems, a mainstream usecase is to not pass correlationId, and expect it to be taken from DiagnosticContext's requestContext. If we keep this method, many callers will simply make getDiagnosticContextMetadata(null) call by design. So, I felt it's better to keep two separate methods. Thoughts? 
getDiagnosticContextMetadata() is public (from both common and common4j). Is that supposed to be part of public API of this lib? If so probably both versions of this method need to be public (or just one method when correlationId is marked as nullable).

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -875,8 +875,7 @@ protected AccountRecord getCachedAccountRecord(
                     "No accounts found for clientId ["
                             + clientId
                             + ", "
-                            + "]",
-                    null
+                            + "]"
```
iamgusain: We can probably get rid of this whole Logger.info block.
As same info is getting logged again in line 880

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -380,8 +380,8 @@ private static void logParameters(@NonNull String tag, @NonNull String correlati
         final String TAG = tag + ":" + parameters.getClass().getSimpleName();
 
         //TODO:1315871 - conversion of PublicApiId in readable form.
-        Logger.info(TAG, DiagnosticContext.INSTANCE.getRequestContext().toJsonString(),
-                "Starting request for correlation id : ##" + correlationId
+        Logger.info(TAG, "Starting with request with request context: "
```
iamgusain: nit: Starting request with..

iamgusain (approved):
LGTM, added few minor comments
Please add the log output in the PR description before and after these changes.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
mohitc1: One will get filtered out if PII is turned off. Is that okay?

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -457,5 +446,26 @@ private static String getUTCDateTimeAsString() {
...
+        String threadName = DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.THREAD_NAME);
```
rpdome, line 459: nit: final

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
rpdome: nit: mark parameter with nullable final

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -434,19 +422,20 @@ private static void logDiscardedLogIfApplicable(String logMessage, ILoggerCallba
...
-     * If correlation id exists:
-     * <library_version> [<timestamp> - <correlation_id>] <log_message>
-     * If correlation id doesn't exist:
+     * If diagnosticMetadata (diagnosticMetadata contains thread name and correlationId) exists:
+     * <library_version> [<timestamp> - <diagnosticMetadata>] <log_message>
+     * If diagnosticMetadata doesn't exist:
      * <library_version> [<timestamp>] <log_message>
      */
-    private static String formatMessage(final String correlationID,
+    private static String formatMessage(final String diagnosticMetadata,
```
rpdome: while you're here, could you mark the params with @ Nonnull or @ Nullable?

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
mohitc1: can't make threadName final, as it can be set to "UNSET" in the code that follows.

PR #1721, opened 2022-04-21
ignore failing unit tests which talks to server

iamgusain:


p3dr0rv (approved):

shahzaibj:
We have a lot of tests that talk to the server. They shouldn't fail. If they are failing then we should probably try to fix them.

What error are we getting back?

PR #1725, opened 2022-04-25
Fix APPLICATION_CANCELLED by handling back button press

nickbopp:
`AuthorizationActivity` no longer needs to directly handle back button presses, and it's all standard machinery.  `(Web)AuthorizationFragment` handle the press, and lifetime is tied to that of its view.

This change also updates the required androidx appcompat version from 1.0.2 -> [1.1](https://developer.android.com/jetpack/androidx/releases/appcompat#version_110_3).  I expect most end-user applications are actually using an even newer version of this package already, as even 1.1 was released in 2019.


nickbopp:
I'm still working on getting a good MSAL and OneAuth test pass using these Common changes, but I wanted to give folks a chance to look at the code.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -198,14 +198,10 @@
...
 
-    /**
-     * NOTE: Fragment-only mode will not support this, as we don't own the activity.
-     * This must be invoked by AuthorizationActivity.onBackPressed().
-     */
     @Override
-    public boolean onBackPressed() {
+    public void handleBackButtonPressed() {
         final String methodTag = TAG + ":onBackPressed";
-        Logger.info(methodTag, "Back button is pressed");
```
iamgusain, line 210: nit: update the methodTag to include the new name for the method

iamgusain:
nit: Can you also add a link to any documentations on what changed between appCompat version 1.1.0 and 1.0.2 corresponding to the change in this PR, for future reference.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -51,11 +51,4 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         }
         setFragment(mFragment);
     }
-
```
iamgusain, line 54: do we also need to remove this in "CurrentTaskAuthorizationActivity,java"

iamgusain (commented):
added few minor comments

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
nickbopp: Yes, thanks for the catch.

nickbopp:
When this PR is ingested to MSAL/OneAuth, the `addToBackStack` calls in the two EmbeddedBrowser classes should be removed.  These changes together solve APPLICATION_CANCELLED + fragment lifetime being too long.

irving-as (approved):

iamgusain (approved):

shahzaibj (approved):

nickbopp:
Okay, I'm back from OOF so this is ready for merge now, when someone with permissions is available to do so.

PR #1726, opened 2022-04-26
Split Global and PCA-Specific Configuration in Common

fadidurah:
**What**
Adding support for global configuration, where developers designate global fields that they'd want all PCA instances to have through a global configuration file. Refer to the design document PR below.
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/4763

Pre-req for: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1604

| Global Fields   |      PCA-Specific      |  Unexposed (Default File) |
|-----------------|------------------------|---------------------------|
|minimum_required_broker_protocol_version|client_id|environment|
|web_view_zoom_controls_enabled|redirect_uri|http<br>- read_timeout<br>- connect_timeout|
|web_view_zoom_enabled|authorization_user_agent||
|power_opt_check_for_network_request_enabled|multiple_clouds_supported||
|handle_null_taskaffinity|broker_redirect_uri_registered||
|authorization_in_current_task|account_mode||
|telemetry|client_capabilities||
|logging<br>- pii_enabled<br>- log_level<br>- logcat_enabled|authorities||
|browser_safe_list<br>- browser_package_name<br>- browser_signature_hashes<br>- browser_version_lower_bound<br>- browser_version_upper_bound|||

**Why**
At the moment, every `PublicClientApplication` (PCA) needs a configuration file to explicitly set all of its configuration fields, including fields that are not designed to be different across multiple PCA instances within the same application. A developer would need to create a different configuration file if they wanted to exclude a field from a PCA's configuration, and reapply whatever fields they wanted to keep.

**How**
Creating a `GlobalSettings` class used to initialize the global configuration from a global configuration file that the developer passes in as a parameter. In this design, the global settings initialization is done before the PCA is created in the developer's app. Once initialization is done, the global settings will be available as a static singleton field, which can be used later by PCA instances to supply global fields as command parameters. These settings are called "Global" because as mentioned above, fields included in these files are not meant to be changed between different PCA instances within the same session.

**Testing**
Added unit testing in this PR through `GlobalSettingsTest` to check global configuration file reading and merging with default global configuration. Interaction with PCA testing done in MSAL (see other PR).


common/src/main/java/com/microsoft/identity/common/globalsettings/GlobalSettings.java
```diff
@@ -0,0 +1,256 @@
...
+    @WorkerThread
+    private static void loadDefaultGlobalConfiguration(@NonNull final Context context) {
+        setDefaultGlobalConfiguration(
+                initializeGlobalConfiguration(context)
+        );
+    }
+
+    @WorkerThread
+    private static void setGlobalConfiguration(@NonNull final GlobalSettingsConfiguration globalConfiguration,
+                                               @NonNull final GlobalSettingsListener listener) {
```
mohitc1, line 166: Just to understand, why don't we return the result as return value? do we expect this method (and other similar methods in this class) to ever be async and be passed result in listener?

PR #1727, opened 2022-04-28
Setting Up Common for YubiKey Support

melissaahn:
### Links to Dependent PRs
(coming soon....)

## The Problem
Two main issues were brought about at the start of this EO YubiKey support work:

- When the YubiKey was plugged in, the host app, common, and broker code reacted adversely. While trying to authenticate via WebView, the WebView would seem to refresh upon YubiKey plug in, and this would result in an authentication flow that didn't seem to complete despite providing a correct upn and certificate/password. Additionally, if the YubiKey was plugged in or removed while in AzureSample's fragments, the app would crash.
- [YubiKit for Android](https://github.com/Yubico/yubikit-android) has a minimum API level of 19. The current minimum level of Common is 16.

## Why This Was Happening
The first issue was due to Android recognizing a YubiKey device as a keyboard, which I suspect has to do with the [YubiKey's Touch Triggered OTP feature](https://developers.yubico.com/Developer_Program/Guides/Touch_triggered_OTP.html#:~:text=The%20YubiKey%20has%20an%20integrated,input%20field%20or%20command%20prompt.). Android sees an external keyboard being plugged in as a configuration change [`configChanges`](https://developer.android.com/guide/topics/manifest/activity-element#config). The default behavior for `configChanges` is to restart the current Activity. This behavior checks out with the observed WebView behavior. In the case of the host app crashing, the account Spinner was attempting to reference a null context before the current fragment could get reattached.

In the second issue, bumping up the `minSdkVersion` of Common needed to be avoided in order to still support our customers workings with a min API level less than 19. 

## Changes Made
For the first issue, in all Manifests with Activities that could be exposed to a plugged in YubiKey, `keyboard` and `keyboardHidden` need to be added to the `configChanges` for each activity listed. This overrides the default behavior and allows the activities themselves to handle the configuration change (in this case, we do nothing). 

For the second issue, it was decided that the code directly handling smartcard CBA was going to be housed in the Broker repo. due to both Authenticator and Company Portal being on API levels higher than 19, it is safe to bump the `minSdkVersion` of broker to `brokerMinSdkVersion` (19). Since our main customer for this project has users who will all have Authenticator, requiring a broker app to complete smartcard CBA will not be an issue. 
In order to port over the smartcard CBA handling from Broker to Common, a number of changes were made to both Broker and Common. Below is a summary of the more important changes made to Common: 

-  **`IChallengeHandler`** was moved from `com.microsoft.identity.common.internal.ui.webview.challengehandlers` to `com.microsoft.identity.common.java.challengehandlers` (moved to common4j). Needed to move it because `IPlatformComponents` is in common4j and needed to reference it for a new method's return value. All `IChallengeHandler` references were updated.
- **`IPlatformComponents`**: Added `IChallengeHandler<?,?> getClientCertAuthChallengeHandler()`. Only notable method implementations are in `AndroidPlatformComponents` (returns a `ClientCertAuthChallengeHandler`) and `AndroidBrokerPlatformComponents` (returns `BrokerClientCertAuthChallengeHandler`, the new `ChallengeHandler` created to handle smartcard CBA). The method in the other components does nothing.
- **`LocalMSALController`:** In `performAuthorizationRequest`, if found authorization strategy is of type `EmbeddedWebViewAuthorizationStrategy`, set the strategy's `ClientCertAuthChallengeHandler` to `ClientCertAuthChallengeHandler`.
- **`IAuthorizationStrategy`**: added method `setClientCertAuthChallengeHandler` for `EmbeddedWebViewAuthorizationStrategy`. This method in other `AuthorizationStrategy` implementations does nothing.
- **`AndroidAuthorizationStrategy`**: `launchIntent` method is overloaded to take an `IChallengeHandler` as a parameter. In the case where an `AuthorizationFragment` is being started, the `ChallengeHandler` is passed to the `AuthorizationFragment` if the fragment is an instance of a `WebViewAuthorizationFragment`.
- **`EmbeddedWebViewAuthorizationStrategy`:** has a local private variable `mClientCertAuthChallengeHandler` that can be set with `setClientCertAuthChallengeHandler` method. `mClientCertAuthChallengeHandler` is passed as a parameter to `AuthorizationActivityFactory.getAuthorizationActivityIntent` within `buildAuthorizationActivityStartIntent`.
- **`AuthorizationActivityFactory`:** `getAuthorizationActivityIntent` is overloaded to take an `IChallengeHandler` parameter. This `ChallengeHandler` is passed to `AuthorizationActivity` via setting `AuthorizationActivity`'s static `ChallengeHandler` variable.
- **`AuthorizationActivity`**: Added static variables `sClientCertAuthChallengeHandler` and `sLock` (acts as a mutex for setting and reading `sClientCertAuthChallengeHandler`). Injects `sClientCertAuthChallengeHandler` into the local `mFragment` if it is an instance of `WebViewAuthorizationFragment`. 
- **`WebViewAuthorizationFragment`:** added local variable `mClientCertAuthChallengeHandler` and pass it to constructor of `AzureActiveDirectoryWebViewClient`.
- **`AzureActiveDirectoryWebViewClient`:** added `IChallengeHandler` parameter to constructor, which sets local private variable. `onReceivedClientCertRequest` is altered to call `processChallenge `on `mClientCertAuthChallengeHandler` rather than creating a new instance of `ClientCertAuthChallengeHandler`. This is key to allowing `BrokerClientCertAuthChallengeHandler.processChallenge` to be called when a Broker app is present. 

## Testing
(Coming soon....)


common/src/main/AndroidManifest.xml
```diff
@@ -1,22 +1,23 @@
...
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.CurrentTaskAuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
 
         <!-- Activity to invoke an interactive request to the intent passed by ad-accounts(Broker) -->
         <activity
             android:name="com.microsoft.identity.common.internal.broker.BrokerActivity"
-            android:configChanges="orientation|screenSize|screenLayout" />
+            android:configChanges="orientation|screenSize|screenLayout|keyboardHidden|keyboard" />
```
rpdome, line 20: Let's add a comment here that this is because yubikey is perceived as a keyboard.

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -225,6 +227,10 @@ private AuthorizationResult performAuthorizationRequest(@NonNull final OAuth2Str
         parameters.getPlatformComponents().getAuthorizationStrategyFactory();
 
         mAuthorizationStrategy = parameters.getPlatformComponents().getAuthorizationStrategyFactory().getAuthorizationStrategy(parameters);
+        //if authorizationStrategy is an EmbeddedWebViewAuthorizationStrategy, set clientCertAuthChallengeHandler variable in strategy.
+        if (mAuthorizationStrategy instanceof EmbeddedWebViewAuthorizationStrategy) {
```
rpdome: I think this could be done in the AuthorizationFactory itself, so that we don't have to do this check in multiple places.

pass in the component object into the factory, and use it in getAuthorizationStrategyFactory()

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -23,19 +23,29 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
 public class AuthorizationActivity extends DualScreenActivity {
 
     public static final String TAG = AuthorizationActivity.class.getSimpleName();
 
     private AuthorizationFragment mFragment;
 
+    //Holds ChallengeHander that handles CBA.
+    public static IChallengeHandler<ClientCertRequest, Void> sClientCertAuthChallengeHandler;
```
rpdome, line 46: I'm wondering if we can do something like this

https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1728
not the cleanest solution, but IMO still better than using static obj.

(The code above will definitely break. I just dished it out as a guideline.)

rpdome: (I also think that we're handling state bundle incorrectly all along).

rpdome (commented):
left a few comments.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
rpdome: I assume making the class serializable is not an option. If it is, then that is always the best way to go.

PR #1728, opened 2022-04-28
tmp - please ignore

rpdome:


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -135,6 +129,26 @@ void extractState(@NonNull final Bundle state) {
...
+
+    @SuppressWarnings("unchecked")
+    private IChallengeHandler<ClientCertRequest, Void> extractClientCertAuthHandler(@NonNull final Bundle state){
+        final String challengeHandlerClassName = state.getString(CLIENT_CERT_REQUEST_CLASS_NAME, null);
+        if (!StringUtil.isNullOrEmpty(challengeHandlerClassName)) {
+            try {
+                final Class<?> challengeHandlerClass = Class.forName(challengeHandlerClassName);
+                mClientCertAuthChallengeHandler = (IChallengeHandler<ClientCertRequest, Void>) challengeHandlerClass.newInstance();
+            } catch (ClassNotFoundException e) {
+                // Log. Do nothing
```
rpdome, line 143: on second thought.. probably better to throw. we want to make sure any "dev errors" like this are caught during test/bugbash.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -111,19 +118,6 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         }
     }
 
-    @Override
-    public void onSaveInstanceState(@NonNull Bundle outState) {
```
rpdome, line 115: do not delete this, and also save the cert challenge handler class name as well.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java
```diff
@@ -54,11 +54,6 @@ public abstract class AuthorizationFragment extends Fragment {
 
     private static final String TAG = AuthorizationFragment.class.getSimpleName();
 
-    /**
-     * The bundle containing values for initializing this fragment.
-     */
-    private Bundle mInstanceState;
```
rpdome, line 60: I think we still can use this, but moving to getArguments() is probably less confusing IMO.

PR #1729, opened 2022-05-02
Melissaahn/yubi kit override integration

melissaahn:
### Links to Dependent PRs
- `keyboard` config changes in Broker Manifest: https://github.com/AzureAD/ad-accounts-for-android/pull/1869
- `keyboard` config changes in MSAL Manifest: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1618
- Adding YubiKit Version to versions.gradle in android complete: https://github.com/AzureAD/android-complete/pull/135
- Adding YubiKit Version to versions.gradle in ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1689

## The Problem
Two main issues were brought about at the start of this EO YubiKey support work:

- When the YubiKey was plugged in, the host app, common, and broker code reacted adversely. While trying to authenticate via WebView, the WebView would seem to refresh upon YubiKey plug in, and this would result in an authentication flow that didn't seem to complete despite providing a correct upn and certificate/password. Additionally, if the YubiKey was plugged in or removed while in AzureSample's fragments, the app would crash.
- [YubiKit for Android](https://github.com/Yubico/yubikit-android) has a minimum API level of 19. The current minimum level of Common is 16.

## Why This Was Happening
The first issue was due to Android recognizing a YubiKey device as a keyboard, which I suspect has to do with the [YubiKey's Touch Triggered OTP feature](https://developers.yubico.com/Developer_Program/Guides/Touch_triggered_OTP.html#:~:text=The%20YubiKey%20has%20an%20integrated,input%20field%20or%20command%20prompt.). Android sees an external keyboard being plugged in as a configuration change [`configChanges`](https://developer.android.com/guide/topics/manifest/activity-element#config). The default behavior for `configChanges` is to restart the current Activity. This behavior checks out with the observed WebView behavior. In the case of the host app crashing, the account Spinner was attempting to reference a null context before the current fragment could get reattached.

In the second issue, bumping up the `minSdkVersion` of Common needed to be avoided in order to still support our customers workings with a min API level less than 19. 

## Changes Made
For the first issue, in all Manifests with Activities that could be exposed to a plugged in YubiKey, `keyboard` and `keyboardHidden` need to be added to the `configChanges` for each activity listed. This overrides the default behavior and allows the activities themselves to handle the configuration change (in this case, we do nothing). 

For the second issue, the `minSdkVersion` issue is avoided by [overriding <uses-sdk> for yubico.yubikit.android](https://developer.android.com/studio/build/manage-manifests#override_uses-sdk_for_imported_libraries). The library will only be referenced within blocks that ensure the device's api level is at 19 or above. 

~~Most of the smartcard CBA handling will be contained within a new `IChallengeHandler` `EoClientCertAuthChallengeHandler`. At the moment, the handler still handles on-device CBA the same (it extends `ClientCertAuthChallengeHandler`),~~ `ClientCertAuthChallengeHandler` still handles on device CBA the same, and now, as an example, it allows the device to detect whether or not a YubiKey is plugged in while the authentication WebView is shown.
Below is a summary of other notable file changes:

- `WebViewAuthorizationFragment`: A class variable was created to store the instance of `AzureActiveDIrectoryWebViewClient` so that it could be referenced later when the YubiKitManager needs to end Usb discovery.
- `AzureActiveDirectoryWebViewClient`: A new ~~`IChallengeHandler`~~ `ClientCertAuthChallengeHandler` class variable is created ~~and is set to either `EoClientCertAuthChallengeHandler` or `ClientCertAuthChallengeHandler` depending on the device's api level.~~ The `IChallengeHandler` is created in the constructor in order to ensure that the `YubiKitManager` (which is created in the `ClientCertAuthChallengeHandler`) has enough time to detect a plugged in YubiKey.
- When `WebViewAuthorizationFragment` is getting destroyed, it calls a wrapper method that stops the `YubiKitManager` from scanning for new Usb connections. Without this, the manager will continue to scan for Usb devices for the duration of the app session.

## Testing
Current tests in _Cert-Based Auth (TLS Challenge)_ in the manual test plans cover testing to make sure that on-device CBA still works with `ClientCertAuthChallengeHandler`. Detailed manual tests will be added to the same category once Smartcard CBA is fully complete.
When a YubiKey is plugged in, an alertDialog pops up that asks the user for permission to access the key. This is implemented by the YubiKit SDK. If the user accepts, the YubiKitManager now has access to the key. If the user denies, the YubiKitManager ignores the key.
To make sure that the YubiKey was being detected by the YubiKit SDK, I plugged in the YubiKey at various stages of the CBA flow while testing AzureSample, with and without BrokerHost:

- Plug in before clicking "Acquire Token"
    1. Plug in YubiKey.
    2. No logs that device was connected initially.
    3. Click "Acquire Token".
    4. Should be prompted by AlertDialog to authorize connection with YubiKey. Click "Accept".
    5. Log should show that device was connected.
    6. Finish on-device CBA.
    7. Upon successful login and return to AzureSample, log should show that device was disconnected.
- Plug in YubiKey after clicking "AcquireToken" (while WebView is showing)
    1. Click "Acquire Token".
    2. Once WebView is visible, plug in YubiKey.
    3. Should be prompted by AlertDialog to authorize connection with YubiKey. Click "Accept".
    4. Log should show that device was connected.
    5. Finish on-device CBA.
    6. Upon successful login and return to AzureSample, log should show that device was disconnected.
 - Plug in YubiKey after clicking "AcquireToken", and unplug while WebView is showing.
    1. Click "Acquire Token".
    2. Once WebView is visible, plug in YubiKey.
    3. Should be prompted by AlertDialog to authorize connection with YubiKey. Click "Accept".
    4. Log should show that device was connected.
    5. Unplug YubiKey.
    6. Log should show that device was disconnected.
    7. Finish on-device CBA.
    8. CBA should finish successfully with no crashes.

Another alternative flow that could be included in each of the scenarios above is denying connection with YubiKey in AlertDialog (No logs that show device is connected or disconnected). Due to the addition of the `keyboard` and `keyboardHidden` to Activity `configChanges` in the Manifests, none of the scenarios above should result in crashing. 

(Thank you Fadi and Dome for the testing advice :))


common/src/main/AndroidManifest.xml
```diff
@@ -1,22 +1,26 @@
...
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+    <uses-sdk tools:overrideLibrary="com.yubico.yubikit.android" />
+
     <application>
+        <!-- Because Android detects plugged in YubiKeys as keyboard config changes, configChanges must have `keyboardHidden` and `keyboard` -->
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
```
shahzaibj, line 14: Can you share official Android documentation on what this does? I think you said in the description that Android handles this themselves but not entirely clear what Android does here

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -264,6 +266,16 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.setWebViewClient(webViewClient);
     }
 
+    /**
+     * NOTE: In the case that AADWebViewClient is using EoClientCertAuthChallengeHandler to handle CBA,
```
shahzaibj: This seems more like a comment than a JavaDoc. I think this is specifically for line 276

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -264,6 +266,16 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+     * the YubiKitManager needs to stop discovering Usb devices upon fragment destroy.
+     */
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        mAADWebViewClient.stopYubiKitManagerUsbDiscovery();
```
shahzaibj: Where/when do we start the discovery?

common/src/main/AndroidManifest.xml
melissaahn: https://developer.android.com/guide/topics/resources/runtime-changes#HandlingTheChange
The default handling of config changes is that the activity is destroyed and created again. I didn't want this to happen whenever a YubiKey was plugged in, hence the addition of the `keyboard` flag, so that nothing happens instead (until we're able to start up the YubikitManger in the EoClientCertAuthChallengeHandler).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
```diff
@@ -0,0 +1,132 @@
...
+                    @Override
+                    public void run() {
+                        Log.i(TAG, "A YubiKey device was disconnected");
+                    }
+                });
+            }
+        });
+
+    }
+    //processChallenge is the same for now.
```
shahzaibj: If it is the same then we don't need to duplicate all this code. Let's create an `AbstractChallengeHandler` and move this code there so it can be shared between both EoClientCertChallengeHandler as well as the regular ClientCertChallengeHandler.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String TAG = EoClientCertAuthChallengeHandler.class.getSimpleName();
+    private static final String ACCEPTABLE_ISSUER = "CN=MS-Organization-Access";
+    private Activity mActivity;
+    private final YubiKitManager mYubiKitManager;
+
+    public EoClientCertAuthChallengeHandler(@NonNull final Activity activity) {
+        mActivity = activity;
+        //Create and start YubiKitManager
+        mYubiKitManager = new YubiKitManager(mActivity.getApplicationContext());
+        mYubiKitManager.startUsbDiscovery(new UsbConfiguration(), new Callback<UsbYubiKeyDevice>() {
```
shahzaibj: Do we need to be discovering USB devices for the entire lifetime of this object or for the entire duration of the WebView being open? Or should we only be discovering when there is actually a client cert request?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
shahzaibj: This seems during IChallengeHandler constructor. I left comments there and we can continue the discussion there.

shahzaibj (commented):
I left some comments / suggestions

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: In the constructor for EoClientCertAuthChallengeHandler: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1729/files#diff-75fd24e46d0e710da05bf7629216b41c67633fd3f59f5d3169e94ff8c01a7eb9R56
A EoClientCertAuthChallengeHandler is created in the constructor of AzureActiveDirectoryWebViewClient.
Do you think it would be better to have a corresponding method in AzureActiveDirectoryWebViewClient for starting the discovery?

shahzaibj: I actually think we may want to start / stop the discovery when there is a client cert request. I left a comment in the ctor there.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
melissaahn: I did try to just start discovery in the processChallenge method initially, but I noticed that the YubiKey wasn't being detected at all in time to choose a certificate. Allowing discovery for the duration of the WebView being open seems to be more effective at the moment. Soldiers will be plugging in their keys early, most likely at some point when the WebView is showing, and maybe they want a notification telling them that it's been connected as soon as they connect it. But containing the discovery to the challengehandler is something I will still explore as the structure of the EoClientCertAuthChallengeHandler builds.
Maybe calling startDiscovery in the onClientCertRequest method would give enough time for the key to be recognized? Something I'll try.

shahzaibj: > Maybe calling startDiscovery in the onClientCertRequest method would give enough time for the key to be recognized?

Let's try this out.

> Soldiers will be plugging in their keys early, most likely at some point when the WebView is showing, and maybe they want a notification telling them that it's been connected as soon as they connect it.

Is our code sending a System notification? Or is the YubiKey sdk doing it? I didn't realize a notification was being sent.

melissaahn: When in discovery mode, the YubiKit SDK shows an alertDialog when a YubiKey is plugged in, asking if it's ok for the host or broker app to access the key. If the user selects ok, then we have access to the key's contents. If they select cancel, the sdk ignores it and does not recognize it as a yubikey anymore.

shahzaibj: If this is the case then I think the current model of early discovery makes sense.

melissaahn: Attempt to start discovery right away in onReceivedClientCertRequest resulted in the same outcome as starting discovery in processChallenge. I think there might be race conditions involved, since I can plug in my yubikey and get the alertDialog after canceling the regular CBA flow.

shahzaibj: I think the way you have the code currently is good. Now that you explained the notification scenario I think we should keep the code as is. I'll resolve this comment.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
melissaahn: Is it possible to not make ClientCertAuthChallengeHandler final, and then have EoClientCertAuthChallengeHandler extend it? Asking since ClientCertAuthChallengeHandler would basically be AbstractChallengeHandler, since EoClientCertAuthChallengeHandler will handle on-device CBA in the same way. 
My plan was to have ClientCertAuthChallengeHandler have a method that takes care of on-device CBA and call it in processChallenge. Then, EoClientCertAuthChallengeHandler can inherit this method and call it when it needs to (when a smartcard is not plugged in).

shahzaibj: I think it should be okay to remove `final` from that class.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: Made change to regular comment.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
melissaahn: Latest commit implements the plan described above.

shahzaibj (approved):

common/src/main/AndroidManifest.xml
```diff
@@ -1,22 +1,26 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="com.microsoft.identity.common">
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
+    <uses-sdk tools:overrideLibrary="com.yubico.yubikit.android" />
```
rpdome, line 8: nit: also add comment why we have to do this.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -264,6 +266,14 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.setWebViewClient(webViewClient);
     }
 
+    // In the case that AADWebViewClient is using EoClientCertAuthChallengeHandler to handle CBA,
+    // the YubiKitManager needs to stop discovering Usb devices upon fragment destroy.
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        mAADWebViewClient.stopYubiKitManagerUsbDiscovery();
```
rpdome: let's add null check just in case.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -77,13 +79,22 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
...
 
     public AzureActiveDirectoryWebViewClient(@NonNull final Activity activity,
                                              @NonNull final IAuthorizationCompletionCallback completionCallback,
                                              @NonNull final OnPageLoadedCallback pageLoadedCallback,
                                              @NonNull final String redirectUrl) {
         super(activity, completionCallback, pageLoadedCallback);
         mRedirectUrl = redirectUrl;
+        //If api level is 19 or above, use EoClientCertAuthChallengeHander
+        //Otherwise, use regular ClientCertAuthChallengeHandler
+        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
rpdome: ClientCertRequest object seems to be requiring API level 21... onReceivedClientCertRequest as well...

So.. I'm thinking that we might not need this check.

Can you try building MSAL test app with targetApi < 21 and and test the CBA flow? what's the behavior there?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/EoClientCertAuthChallengeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public EoClientCertAuthChallengeHandler(@NonNull final Activity activity) {
+        super(activity);
+        mActivity = activity;
+        //Create and start YubiKitManager
+        mYubiKitManager = new YubiKitManager(mActivity.getApplicationContext());
+        mYubiKitManager.startUsbDiscovery(new UsbConfiguration(), new Callback<UsbYubiKeyDevice>() {
+            @Override
+            public void invoke(UsbYubiKeyDevice device) {
+                Log.i(TAG, "A YubiKey device was connected");
```
rpdome: nit: use common4j's Logger.

rpdome (commented):
left a few comments.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: Tested with a device Api level 19 (KitKat).
Couldn't build msalTestApp: `Invoke-customs are only supported starting with Android O (--min-api 26)`
So I tried with AzureSample and the regular ClientCerthAuthChallengeHandler.
Sometimes the WebView crashes right away, but after the third try, I was able to try to sign in. When I try to login with a cert, the picker doesn't show up and it takes me directly to the error page. 
The logcat shows many DalvikVm errors, but the one pertaining to ClientCertAuthRequest is: `Could not find class 'android.webkit.ClientCertRequest', referenced from method com.microsoft.identity.common.internal.ui.webview.challengehandlers.ClientCertAuthChallengeHandler.processChallenge', referenced from method com.microsoft.identity.common.internal.ui.webview.challengehandlers.EoClientCertAuthChallengeHandler.processChallenge` 

This check is definitely not needed anymore, since it seems like any device <21 can't do regular CBA anyway.
Additionally, seems like I can just edit the ClientCertAuthChallengeHandler class directly (no need for the Eo version)

rpdome: @melissaahn we should probably let vimala know (if request comes from android and x-client-OS < 21, do not trigger CBA)

cc @shoatman

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -454,7 +457,17 @@ private void returnError(final String errorCode, final String errorMessage) {
     @Override
     public void onReceivedClientCertRequest(WebView view,
                                             final ClientCertRequest clientCertRequest) {
-        final ClientCertAuthChallengeHandler clientCertAuthChallengeHandler = new ClientCertAuthChallengeHandler(getActivity());
-        clientCertAuthChallengeHandler.processChallenge(clientCertRequest);
+        if (mClientCertAuthChallengeHandler != null) {
+            mClientCertAuthChallengeHandler.processChallenge(clientCertRequest);
+        }
+    }
```
rpdome, line 466: nit: also add logging if this is null.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -264,6 +266,16 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.setWebViewClient(webViewClient);
     }
 
+    // For ClientCertAuthChallengeHandler within AADWebViewClient,
+    // the YubiKitManager needs to stop discovering Usb devices upon fragment destroy.
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        if (mAADWebViewClient != null) {
```
rpdome, line 275: let's add logging if this is null.

rpdome (approved):
Let's add comment on top of the `ClientCertAuthChallengeHandler` class that

Yubikey requires API>19
CBA requires API>21

PR #1732, opened 2022-05-06
UI Automation changes to support local install source for WordApp

somalaya:
**The Problem**
We currently support WordApp installation from the playstore; however, we now see that on API 29 on emulators, word app has stopped support for installation from playstore.

In this PR, Added support to install wordApp using a localAPK

**Implementation**
We need to add a WordAppSource build config field, which will be LocalApk when we need to install using the localApk. This is similar to how we install broker apps using the config

By default, we set it to playstore.

The -pWordAppSource build config field will be set while running the test cases.

image

**Changes Made**

- Made changes in build.gradle to set the wordApp install source property
- Note : Also made changes to some first run UI automation for WordApp and CP as we no longer see some screens like Privacy consent and grant storage permission. 

**Testing**
Added UI test cases in BrokerAutomationApp module to verify the changes..  test case PR https://github.com/AzureAD/ad-accounts-for-android/pull/1879

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -71,11 +71,6 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
         localApkFileName = COMPANY_PORTAL_APK;
     }
 
-    public BrokerCompanyPortal(@NonNull final IAppInstaller appInstaller) {
-        super(COMPANY_PORTAL_APP_PACKAGE_NAME, COMPANY_PORTAL_APP_NAME, appInstaller);
-        localApkFileName = COMPANY_PORTAL_APK;
-    }
-
     @Override
     public void performDeviceRegistration(@NonNull final String username,
                                           @NonNull final String password) {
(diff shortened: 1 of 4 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -71,11 +71,6 @@ public BrokerCompanyPortal() {
         localApkFileName = COMPANY_PORTAL_APK;
     }
 
-    public BrokerCompanyPortal(@NonNull final IAppInstaller appInstaller) {
-        super(COMPANY_PORTAL_APP_PACKAGE_NAME, COMPANY_PORTAL_APP_NAME, appInstaller);
-        localApkFileName = COMPANY_PORTAL_APK;
-    }
```
shahzaibj: Why is this code removed?

shahzaibj:
> We currently support WordApp installation from the playstore; however, we now see that on API 29, word app has stopped support for installation from playstore.

This is not entirely true. WordApp should install fine from PlayStore on physical API 29 devices. The issue is only with emulators.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
somalaya: I could not find any code referencing this one. We are installing CP using using another overload constructor. Let me know if i am missing something!

shahzaibj: Android Studio Find Usages feature sometimes doesn't correctly report all usages, so therefore I suggest that we add constructor back even if it is not being used. There is no harm in keeping there for now.

gitaumoses4 (approved):
Looks good.

shahzaibj (approved):

PR #1733, opened 2022-05-11
Add RequestContext autocloseable utility

mohitc1:
**What**
Adding AutoCloseable for setting and clearing RequestContext on DiagnosticContext
**Why**
This will help callers to wrap code in try () specifying a request context to be used for current DiagnosticContext, and automatically clearing after execution. This will help all Logger log statements use data from the request context (primarily correlationId for now).
**How**
- Added new class RequestContextGuard which implements AutoCloseable interface.
- It creates RequestContext from correlationId or allows to specify RequestContext in constructor.
- Initialize method adds the request context to DiagnosticContext.

**Test**
Added unit tests. No usage inside common yet.

Note: Current intention of introducing this is use it in StoreManager.java in msalcpp. This will help in using the correlationId passed from OneAuth/MSAL to common.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
@@ -1,0 +1,27 @@
+package com.microsoft.identity.common.java.logging;
+
+import com.microsoft.identity.common.java.util.StringUtil;
+
+public class RequestContextGuard implements AutoCloseable {
+
+    private final IRequestContext requestContext;
+    public RequestContextGuard(IRequestContext requestContext) {
+        this.requestContext = requestContext;
+    }
+
+    public RequestContextGuard(String correlationId) {
+        this.requestContext = new RequestContext();
+        if (!StringUtil.isNullOrEmpty(correlationId)) {
+            this.requestContext.put(DiagnosticContext.CORRELATION_ID, correlationId);
+        }
+    }
+
+    public void initialize() {
+        DiagnosticContext.INSTANCE.setRequestContext(this.requestContext);
+    }
+
+    @Override
+    public void close()  {
+        DiagnosticContext.INSTANCE.clear();
+    }
+}
--- a/common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
@@ -1,25 +1,69 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
 package com.microsoft.identity.common.java.logging;
 
 import com.microsoft.identity.common.java.util.StringUtil;
+import edu.umd.cs.findbugs.annotations.Nullable;
 
+/**
+ * This class provide way to run piece of code with a request context as resource.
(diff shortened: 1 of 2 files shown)


common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
```diff
@@ -0,0 +1,27 @@
+package com.microsoft.identity.common.java.logging;
```
shahzaibj, line 23: Missing LICENSE

common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.microsoft.identity.common.java.util.StringUtil;
+
+public class RequestContextGuard implements AutoCloseable {
```
shahzaibj, line 34: Javadoc

common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private final IRequestContext requestContext;
+    public RequestContextGuard(IRequestContext requestContext) {
```
shahzaibj: nit: add empty line between fields and constructor

shahzaibj: Add NonNull final to param

common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public RequestContextGuard(String correlationId) {
+        this.requestContext = new RequestContext();
+        if (!StringUtil.isNullOrEmpty(correlationId)) {
+            this.requestContext.put(DiagnosticContext.CORRELATION_ID, correlationId);
+        }
+    }
+
+    public void initialize() {
```
shahzaibj, line 60: javadoc

common4j/src/test/com/microsoft/identity/common/java/logging/LoggerTest.java
```diff
@@ -148,6 +148,27 @@ public void execute() {
         }, false);
     }
 
+    @Test(timeout = TEST_TIME_OUT_IN_MILLISECONDS)
+    public void logWithRequestContextGuard() throws InterruptedException {
+        final Logger.LogLevel logLevel = Logger.LogLevel.VERBOSE;
+        final boolean containsPII = false;
+        final String newCorrelationId = "NEW_CORRELATION_ID";
```
shahzaibj:
```suggestion
        final String correlationId = UUID.randomUUID().toString();
```

common4j/src/test/com/microsoft/identity/common/java/logging/LoggerTest.java
```diff
@@ -148,6 +148,27 @@ public void execute() {
...
+        final String newThreadName = Thread.currentThread().getName();
```
shahzaibj: Let's not use `new` as a prefix for variable / constant names. This is just a thread name as I see it...there is nothing "new" about it

shahzaibj: and `new` is a reserved word in java

common4j/src/main/com/microsoft/identity/common/java/logging/RequestContextGuard.java
shahzaibj, line 34: This class seems to not do anything special. Why do we need this class? Why can't we do what it does directly in the DiagnosticContext class?

common4j/src/test/com/microsoft/identity/common/java/logging/LoggerTest.java
```diff
@@ -197,6 +218,58 @@ public void execute() {
...
+                    try (RequestContextGuard guard2 = new RequestContextGuard(correlationId_2)) {
+                        guard2.initialize();
+                        final RequestContext requestContext = new RequestContext();
+                        requestContext.put(DiagnosticContext.CORRELATION_ID, correlationId_2);
+                        DiagnosticContext.INSTANCE.setRequestContext(requestContext);
+
+                        testLogger(tag, logLevel, correlationId_2, threadName_2[0], containsPII, new IOperationToTest() {
+                            @Override
+                            public void execute() {
+                                Logger.verbose(tag, "");
```
shahzaibj, line 252: Probably better to log something that is not an empty string...a logic could be introduced in the future that says discards empty string messages sent to the logger and then this test becomes useless

common4j/src/test/com/microsoft/identity/common/java/logging/LoggerTest.java
mohitc1: made it just "CORRELATION_ID"

rpdome:
Could you please give a concrete (real life) example on where this will be used?

RequestContext is used as a per-thread global variable - it could be invoked anywhere in the flow, so its lifetime = the flow's lifetime.

(the alternative would be to pass the object when Logging.xxxx() is invoked - my understanding is that it was intentionally designed this way because passing this object all around the code would be painful).

This introduces a usage whose lifetime ends within a try block.

mohitc1:
> Could you please give a concrete (real life) example on where this will be used?
> 
> RequestContext is used as a per-thread global variable - it could be invoked anywhere in the flow, so its lifetime = the flow's lifetime.
> 
> (the alternative would be to pass the object when Logging.xxxx() is invoked - my understanding is that it was intentionally designed this way because passing this object all around the code would be painful).
> 
> This introduces a usage whose lifetime ends within a try block.

This is done with intention that if piece of code/flow is run under diagnostic context, once it is completed, the context is automatically cleared. While debugging an MSALCPP flow which uses common's cache via StoreManager.java (also part of msalcpp), I saw diagnostic context is not set. StoreManager's methods are called from native CPP code in different sign in/acquire token flows. It seems like it would be helpful for such methods to set context when execution starts and clear when they end. These methods have are entry point into to common's functionality.

PR #1734, opened 2022-05-12
Directly support DayNight in http_auth_dialog.xml

fadidurah:
Had a recent ICM (https://portal.microsofticm.com/imp/v3/incidents/details/300034065/home) where a user was running into NTLM prompt on teams device and text entered in username/password boxes was matching colors with the background of the TextEdit object, so the text was invisible to the user. This looked like the font color was set from dark mode, but background color was not. Could not reproduce this locally, think this would fix the problem of mishandling DayNight Mode.

### DIFF
--- a/common/src/main/res/layout/http_auth_dialog.xml
+++ b/common/src/main/res/layout/http_auth_dialog.xml
@@ -8,6 +8,7 @@
     android:paddingRight="@dimen/http_auth_dialog_padding_right" >
 
     <EditText
+        style="@style/Theme.AppCompat.DayNight"
         android:inputType="text"
         android:id="@+id/editUserName"
         android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


melissaahn (approved):

rpdome:
Have you tried popping up this dialog with system's dark theme (before and after this change)? What did you observe?

(also, do we know if the affected device is pre or post android 10? have we tried this with both?)

fadidurah:
> Have you tried popping up this dialog with system's dark theme (before and after this change)? What did you observe?
> 
> (also, do we know if the affected device is pre or post android 10? have we tried this with both?)

I tested Night/Day themes before and after this change, The prompt was adapting as expected to both themes and there were no issues. I could not actually reproduce the issue, so the behavior did not change for me after adding this change. I also tried a different approach where we specify background and text color manually based on what theme it is, but that led to some inconsistent looking prompts. I would try that approach if this fix does not solve the issue, but I can't tell for sure since I was never able to reproduce this. Logically, this has a good chance to fix the problem since the issue is the EditText object is inconsistently setting its colors, so deleting the xml file in MSAL and adding a declaration for DayNight support.

PR #1737, opened 2022-05-16
changing changelog entry from minor to major with details

melissaahn:
This is a small change to the already merged PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1729
Due to the addition of YubiKit SDK, all host apps that consume libraries that use common will need to support Java Version 8, which is a MAJOR change instead of a MINOR change.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 V.Next
 ----------
-- [MINOR] Adding YubiKit SDK dependency; added keyboard flag to android:configChanges for all activities that could interact with a YubiKey. (#1729)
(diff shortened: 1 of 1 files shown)


shahzaibj:
For context, YubiKit declared Java 8 in their build.gradle: https://github.com/Yubico/yubikit-android/blob/master/build.gradle#L46-L49

and that's why we're doing this i.e. that's why we're making this a MAJOR change on our end because we're a library and NOT an application and therefore even desugaring is of no use to us because desugaring happens on use and not on publication. We can declare desugaring in our build.gradle and publish our AAR but consuming apps will still need to bump themselves directly to Java 8 because our AAR is actually NOT desugared by virtue of how desugaring works: https://stackoverflow.com/questions/48099802/android-desugar-classes-in-aar-package

shahzaibj (approved):

p3dr0rv (approved):

fadidurah (approved):

PR #1743, opened 2022-05-21
Fix SDK Cancel when using authorization in current task

shahzaibj:
### What?
This PR fixes [ADO bug 1864115](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1864115)

We seem to have completely broken the `authorization_in_current_task` in MSAL at some point. I don't know if this was broken when we added the authorization in new task support or if it was broken at some point after adding that support. Regardless, I've fixed that here.

### How?
It seems we had a lot of duplicated code here in [`CurrentTaskAuthorizationFragment`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CurrentTaskAuthorizationFragment.java#L49) and all of this code was already present on the parent class i.e.  [`AuthorizationFragment`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L53). I've removed the `CurrentTaskAuthorizationFragment` class altogether, which essentially removes the duplicated code and resolves the problem. 

You might be wondering how just having duplicated code led to the problem? Well, the answer is that on the parent class i.e. `AuthorizationFragment` there is a state variable called as [`mInstanceState`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L60) and [state is set on this variable](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L77) and then read inside the[ `onCreate`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L90) method of this class. We ended up re-declaring the[ `mInstanceState`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CurrentTaskAuthorizationFragment.java#L56) variable in the child class and also override the [`setInstanceState` method](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CurrentTaskAuthorizationFragment.java#L68) and this essentially means that state was NOT updated on the parent class and only updated on the child class. So when the parent class tried to access the state then it performed the [check on `mInstanceState` inside its own class](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L90) and this was still set to `null` because `setInstanceState` wasn't invoked on this at all, and as a result our code [invoked `finish`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/1c8f0fd72dfeb5c594f120e2e717eff06e2f1605/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java#L92) on the Activity and the activity got destroyed and resulted in an SDK cancelled result code.

### Testing
Tested with MSAL Test App using authorization in current task as well as different task.


shoatman (approved):
Thanks Shahzaib.

melissaahn (approved):

PR #1745, opened 2022-05-23
Melissaahn/remove new config changes

melissaahn:
Creating this for the new release branch.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,5 @@
 Version 5.0.0
 ----------
-- [MAJOR] Adding YubiKit SDK, which requires Java Version 8 and will thus bump up Java version overall to 8 (#1729)
 - [MAJOR] Move IKeyStoreKeyManager and IDevicePopManager to common4j (#1683)
 - [PATCH] Add exception handling in Content Provider strategy, for broker communication (#1722)
 - [MINOR] Support TenantID value from eSTS in PKeyAuth flows (#1712)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
-V.Next
+Version 5.0.0
 ----------
-- [MINOR] Support WordApp local apk install in UI automation flows (#1732)
+- [MAJOR] Adding YubiKit SDK, which requires Java Version 8 and will thus bump up Java version overall to 8 (#1729)
+- [MAJOR] Move IKeyStoreKeyManager and IDevicePopManager to common4j (#1683)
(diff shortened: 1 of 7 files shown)


gitaumoses4:
@shahzaibj seems here we'll be removing all of the Yubikit changes.

shahzaibj:
@melissaahn I thought we didn't need to remove all the changes?

melissaahn:
I had to remove all changes that included or used the YubiKit Sdk. There are some remaining changes from my PR that just shifted around some sections (the ClientCertAuthChallengeHandler class, for example), but those don't depend on YubiKit.

melissaahn:
If the question is about bumping to 5.0.0.... since Yubikit is not included, and Java version 8 is not included. nor needed to be included in the host app anymore, I'm thinking that the version number does not need to be moved to 5 anymore. I forgot to remove my changelog line, so I'll do that right now.

shahzaibj:
> I had to remove all changes that included or used the YubiKit Sdk. There are some remaining changes from my PR that just shifted around some sections (the ClientCertAuthChallengeHandler class, for example), but those don't depend on YubiKit.

Why was this needed?

melissaahn:
> > I had to remove all changes that included or used the YubiKit Sdk. There are some remaining changes from my PR that just shifted around some sections (the ClientCertAuthChallengeHandler class, for example), but those don't depend on YubiKit.
> 
> Why was this needed?

When the Authenticator team attempted to add a work account in their testing, they got a NoClassDefFoundError in regard to the YubiKitManager class. I'm not sure why this is happening, as they shouldn't need to include any of the YubiKit dependencies (should be taken care of by common). It seems that I need to do more testing before including these dependencies, so I was thinking that the quickest fix to unblock the release is to take out all the YubiKit code.

shahzaibj:
> > > I had to remove all changes that included or used the YubiKit Sdk. There are some remaining changes from my PR that just shifted around some sections (the ClientCertAuthChallengeHandler class, for example), but those don't depend on YubiKit.
> > 
> > 
> > Why was this needed?
> 
> When the Authenticator team attempted to add a work account in their testing, they got a NoClassDefFoundError in regard to the YubiKitManager class. I'm not sure why this is happening, as they shouldn't need to include any of the YubiKit dependencies (should be taken care of by common). It seems that I need to do more testing before including these dependencies, so I was thinking that the quickest fix to unblock the release is to take out all the YubiKit code.

Okay, sounds good.

I think the version number for our release will still remain at `5.0.0` because I see there is another MAJOR change in the changelog.

shahzaibj (approved):

PR #1750, opened 2022-05-26
Add base64 encode for the state parameter

gitaumoses4:
## The Problem
When generating the `state` parameter in Android, we are appending the `taskId` which is an Integer and two UUIDs as follows
```java
String.format("%d%s%s", taskId, SPLITTER, uuids)
```

The generated state is then passed to the `AuthorizationRequest` which builds an authorization URL as a request parameter. Once the authorization is complete, we compare the returned state parameter with the parameter generated on the SDK.

This works fine for most device languages. 
But when the device language is set to "Ar", there is a difference in the returned state parameter and the generated parameter.

The `requestStateParameter` is the one generated by the SDK, the `state` is returned from the authorization redirect.

state

## Why is this happening?
The `String.format` above that generates the state parameter makes use of the device's Locale. This means that the `taskId` will not make use of ASCII characters in some cases e.g. when the device's language is "ar". 

## Changes Made
1. We need to ensure we use a single Locale when generating the state parameter in Android. 
2. It may be hard for us to keep track of which state generators create a URL-unsafe state parameter, hence we will Base64 encode the state parameter value before sending it in the Authorization request. This ensures that we always have a state parameter that is URL safe.

### Reproducing the bug.
1. Change your device language to "AR"
2. Checkout to dev branch and run MSAL test app
3. Perform interactive login with any account
4. You will get this error "Returned state from authorize endpoint is not the same as the one sent"

https://user-images.githubusercontent.com/25629064/170463766-d2be8a16-e391-4341-92c6-6666342a2e6e.mp4

## Testing
1. Checkout to the branch `git/state-mismatch-exception`
2. Change your device language to "AR"
3. Run MSAL test app and ensure broker is not installed in the device.
4. Perform interactive login with any account


shoatman (approved):
Thanks Moses.  This looks right to me.

somalaya (approved):

changelog.txt
```diff
@@ -8,6 +8,7 @@ V.Next
 - [PATCH] Fix SDK Cancel when using authorization in current task (#1743)
 - [MINOR] Added support for broadcasting to applications installed on the device. (#1744)
 - [PATCH] Fix APPLICATION_CANCELLED by handling back button press (#1725)
+- [MINOR] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
```
gitaumoses4:
```suggestion
- [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
```

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
```diff
@@ -241,6 +258,11 @@ public URI getAuthorizationRequestAsHttpRequest() throws ClientException {
             final CommonURIBuilder builder = new CommonURIBuilder(getAuthorizationEndpoint());
             builder.addParametersIfAbsent(ObjectMapper.serializeObjectHashMap(this));
             builder.addParametersIfAbsent(mExtraQueryParams);
+
+            if (!StringUtil.isNullOrEmpty(mState)) {
+                builder.setParameter("state", getEncodedState()); // pass the URL safe state to the URL builder.
+            }
```
shahzaibj: Were we not even attaching state previously?

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -342,6 +342,10 @@ public static String encodeUrlSafeString(@NonNull final String stringToEncode) {
         return Base64.encodeToString(toByteArray(stringToEncode), Base64.NO_WRAP | Base64.NO_PADDING | Base64.URL_SAFE);
     }
 
+    public static String decodeUrlSafeString(@NonNull final String stringToDecode) {
```
shahzaibj: javadoc

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
gitaumoses4: We were, but it was only URL encoded. We need to Base64 encode it when building the URL.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
shahzaibj: probably add some tests for this method

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
shahzaibj: Was that happening in a different place in the code? Can you share the existing code where that was happening?

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
```diff
@@ -219,6 +227,15 @@ public B setWebViewZoomControlsEnabled(boolean webViewZoomControlsEnabled) {
         public abstract AuthorizationRequest build();
     }
 
+    /**
+     * Base64 encode the state parameter to make it URL safe.
+     *
+     * @return a string representing a Base64 encoded state parameter.
+     */
+    public String getEncodedState() {
```
shahzaibj: I'm wondering if rather than having this public accessor, if we should just set the state to encoded internally and keep using the same getState accessor

shahzaibj: Essentially we set `mState` to encodedUriSafeString version and then we don't need this method at all

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
shahzaibj: I'm asking because I'm not seeing any code replaced in this PR that was doing the following:

```java
builder.setParameter("state", mState);
```

shahzaibj: It seems like what we're doing here is first attaching the only URL encoded state to the builder and then later in the flow we're patching (overriding) it with Base64 encoded variant. I think this would make debugging harder. Probably a better idea to set the state only once and correctly in the first place on the authorization request (builder).

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
gitaumoses4: Yes, we can do that. 
I was essentially ensuring `setState` and `getState` receive and retrieve the same value respectively. But I agree it comes with some added complexity. 
I'll just set `mState` to the encodedUriSafeString and have javadoc in the methods.

gitaumoses4: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1750/commits/49bf80b27b2e2c5a3e93047f563df5603050beb9

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
gitaumoses4: Not needed at the moment, so I have removed it.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
gitaumoses4: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1750/commits/49bf80b27b2e2c5a3e93047f563df5603050beb9

shahzaibj (approved):

PR #1752, opened 2022-05-30
BrokerHost - utility functions added

somalaya:
**What** : Added utility functions like getAllAccounts, removeAccounts, confirmNoncePresent which are useful for testing the Broker APIs

**Why** : The utility functions added are emulating the user action to click on the getAccounts, getFlights and other buttons in BrokerHost app and also to test the response displayed in the dialog box after the functionality is completed.
These are some common functions used across test cases. Hence added these in BrokerHost class.
The utility functions added here are being used in the testcases introduced in the following PR : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1633

**Testing** :  Tested the utility functions by invoking them using the test cases introduced in this PR : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1633
The test cases ran successfully

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -171,7 +171,7 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
     }
 
     public void enrollDevice(@NonNull final String username,
-                             @NonNull final String password, @NonNull final Boolean isFederated) {
+                             @NonNull final String password, final boolean isFederated) {
         Logger.i(TAG, "Enroll Device for the given account..");
         launch(); // launch CP app
 
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -32,6 +32,9 @@ import androidx.test.uiautomator.UiSelector;
 
 import com.microsoft.identity.client.ui.automation.TestContext;
 import com.microsoft.identity.client.ui.automation.installer.IAppInstaller;
+import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.AdfsLoginComponentHandler;
+import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.MicrosoftStsPromptHandler;
+import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.MicrosoftStsPromptHandlerParameters;
 import com.microsoft.identity.client.ui.automation.powerlift.IPowerLiftIntegratedApp;
 import com.microsoft.identity.client.ui.automation.constants.DeviceAdmin;
 import com.microsoft.identity.client.ui.automation.device.settings.ISettings;
@@ -169,9 +172,8 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
         // UiAutomatorUtils.handleButtonClick("com.microsoft.windowsintune.companyportal:id/privacy_notice_agree_button");
     }
 
-    @Override
     public void enrollDevice(@NonNull final String username,
-                             @NonNull final String password) {
+                             @NonNull final String password, @NonNull final boolean isFederated) {
         Logger.i(TAG, "Enroll Device for the given account..");
         launch(); // launch CP app
 
@@ -180,19 +182,40 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
         // click Sign In button on CP welcome page
         UiAutomatorUtils.handleButtonClick("com.microsoft.windowsintune.companyportal:id/sign_in_button");
 
-        final PromptHandlerParameters promptHandlerParameters = PromptHandlerParameters.builder()
-                .prompt(PromptParameter.LOGIN)
-                .consentPageExpected(false)
-                .expectingLoginPageAccountPicker(false)
-                .sessionExpected(false)
-                .loginHint(null)
-                .build();
+        if (isFederated) {
+            final MicrosoftStsPromptHandlerParameters promptHandlerParameters = MicrosoftStsPromptHandlerParameters.builder()
+                    .prompt(PromptParameter.LOGIN)
+                    .consentPageExpected(false)
+                    .expectingLoginPageAccountPicker(false)
+                    .sessionExpected(false)
+                    .isFederated(true)
+                    .loginHint(null)
+                    .build();
 
-        final AadPromptHandler aadPromptHandler = new AadPromptHandler(promptHandlerParameters);
+            final MicrosoftStsPromptHandler microsoftStsPromptHandler = new MicrosoftStsPromptHandler(promptHandlerParameters);
 
-        Logger.i(TAG, "Handle prompt in AAD login page for enrolling device..");
-        // handle AAD login page
-        aadPromptHandler.handlePrompt(username, password);
+            Logger.i(TAG, "Handle prompt in ADFS login page for enrolling device..");
+            // handle AAD login page for username
+            UiAutomatorUtils.handleInput("i0116", username);
+            UiAutomatorUtils.handleButtonClick("idSIButton9");
+
+            // handle ADFS login page for password
+            microsoftStsPromptHandler.getLoginComponentHandler().handlePasswordField(password);
+        } else {
+            final MicrosoftStsPromptHandlerParameters promptHandlerParameters = MicrosoftStsPromptHandlerParameters.builder()
+                    .prompt(PromptParameter.LOGIN)
+                    .consentPageExpected(false)
+                    .expectingLoginPageAccountPicker(false)
+                    .sessionExpected(false)
+                    .loginHint(null)
+                    .build();
+
+            final MicrosoftStsPromptHandler microsoftStsPromptHandler = new MicrosoftStsPromptHandler(promptHandlerParameters);
+
+            Logger.i(TAG, "Handle prompt in AAD login page for enrolling device..");
+            // handle AAD login page
+            microsoftStsPromptHandler.handlePrompt(username, password);
+        }
 
         // click the activate device admin btn
         final UiObject accessSetupScreen = UiAutomatorUtils.obtainUiObjectWithText("Access Setup");
@@ -256,6 +279,12 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
 
         // Enrollment has been performed successfully
         enrollmentPerformedSuccessfully = true;
+    }
+
+    @Override
+    public void enrollDevice(@NonNull final String username,
+                             @NonNull final String password) {
+        enrollDevice(username, password, false);
     }
 
     @Override
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -39,9 +39,7 @@ import com.microsoft.identity.client.ui.automation.powerlift.IPowerLiftIntegrate
 import com.microsoft.identity.client.ui.automation.constants.DeviceAdmin;
 import com.microsoft.identity.client.ui.automation.device.settings.ISettings;
 import com.microsoft.identity.client.ui.automation.device.settings.SamsungSettings;
-import com.microsoft.identity.client.ui.automation.interaction.PromptHandlerParameters;
 import com.microsoft.identity.client.ui.automation.interaction.PromptParameter;
-import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.AadPromptHandler;
 import com.microsoft.identity.client.ui.automation.logging.Logger;
 import com.microsoft.identity.client.ui.automation.utils.CommonUtils;
 import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
@@ -53,7 +51,7 @@ import java.util.Random;
 import lombok.Getter;
 
 import static com.microsoft.identity.client.ui.automation.utils.CommonUtils.FIND_UI_ELEMENT_TIMEOUT;
-import static org.junit.Assert.fail;
+
 
 /**
  * A model for interacting with the Company Portal Broker App during UI Test.
@@ -173,7 +171,7 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
     }
 
     public void enrollDevice(@NonNull final String username,
-                             @NonNull final String password, @NonNull final boolean isFederated) {
+                             @NonNull final String password, @NonNull final Boolean isFederated) {
         Logger.i(TAG, "Enroll Device for the given account..");
         launch(); // launch CP app
 
@@ -193,14 +191,7 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
                     .build();
 
             final MicrosoftStsPromptHandler microsoftStsPromptHandler = new MicrosoftStsPromptHandler(promptHandlerParameters);
-
-            Logger.i(TAG, "Handle prompt in ADFS login page for enrolling device..");
-            // handle AAD login page for username
-            UiAutomatorUtils.handleInput("i0116", username);
-            UiAutomatorUtils.handleButtonClick("idSIButton9");
-
-            // handle ADFS login page for password
-            microsoftStsPromptHandler.getLoginComponentHandler().handlePasswordField(password);
+            ((AdfsLoginComponentHandler) microsoftStsPromptHandler.getLoginComponentHandler()).handleEnrollmentPrompt(username, password);
         } else {
             final MicrosoftStsPromptHandlerParameters promptHandlerParameters = MicrosoftStsPromptHandlerParameters.builder()
                     .prompt(PromptParameter.LOGIN)
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
@@ -42,6 +42,8 @@ import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
 import org.junit.Assert;
 
 import static com.microsoft.identity.client.ui.automation.utils.CommonUtils.FIND_UI_ELEMENT_TIMEOUT;
+import java.util.ArrayList;
+import java.util.List;
 
 public class BrokerHost extends AbstractTestBroker {
 
@@ -51,7 +53,7 @@ public class BrokerHost extends AbstractTestBroker {
     public final static String BROKER_HOST_APK = "BrokerHost.apk";
     public final static String BROKER_HOST_APK_PROD = "BrokerHostProd.apk";
     public final static String BROKER_HOST_APK_RC = "BrokerHostRC.apk";
-    
+
     public BrokerHost() {
         super(BROKER_HOST_APP_PACKAGE_NAME, BROKER_HOST_APP_NAME, new LocalApkInstaller());
         localApkFileName = BROKER_HOST_APK;
@@ -185,7 +187,7 @@ public class BrokerHost extends AbstractTestBroker {
         Logger.i(TAG, "Obtain Device Id..");
         final String resourceButtonId = "com.microsoft.identity.testuserapp:id/buttonDeviceId";
         final String textId = "DeviceId";
-        return basicButtonHandler(resourceButtonId,textId);
+        return basicButtonHandler(resourceButtonId, textId);
     }
 
     @Override
@@ -232,7 +234,7 @@ public class BrokerHost extends AbstractTestBroker {
         Logger.i(TAG, "Get Account Upn..");
         final String resourceButtonId = "com.microsoft.identity.testuserapp:id/buttonGetWpjUpn";
         final String textId = "UPN";
-        return basicButtonHandler(resourceButtonId,textId);
+        return basicButtonHandler(resourceButtonId, textId);
     }
 
     @Nullable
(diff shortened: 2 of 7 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -275,7 +277,40 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        final String resourceButtonId = "com.microsoft.identity.testuserapp:id/getFlightsButton";
+        final String text = "Get Flights";
+        launch();
+        // scroll to find the get flights button
+        UiAutomatorUtils.obtainChildInScrollable(text);
+        UiAutomatorUtils.handleButtonClick(resourceButtonId);
+        try {
+            final UiObject flightsObj = UiAutomatorUtils.obtainUiObjectWithResourceId("com.microsoft.identity.testuserapp:id/editTextFlights");
+            String flightsJson = flightsObj.getText();
+            return flightsJson;
```
shahzaibj:
```suggestion
            return flightsObj.getText();
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -275,7 +277,40 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        } catch (UiObjectNotFoundException e) {
```
shahzaibj:
```suggestion
        } catch (final UiObjectNotFoundException e) {
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
             UiAutomatorUtils.handleButtonClick("android:id/button1");
         }
     }
-
-    @Override
-    public void setFlights(@Nullable final String flightsJson) {
-        Logger.i(TAG, "Set Flights..");
+    /**
```
shahzaibj, line 337: Add empty line between previous method and this one

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        Assert.assertTrue(dialogBox.exists());
+
+        // As of now we are testing only for 2 accounts.
+        List<String> accounts = new ArrayList<>();
+        final String accountName = getDialogBoxText(dialogBox, textId);
+        if (accountName != null) {
+            accounts.add(accountName);
+        }
+        if (expectingMultipleAccounts) {
+            final String accountName2 = getDialogBoxText(dialogBox, textId);
```
shahzaibj, line 362: This is interesting...this is calling the same method with same params i.e. `getDialogBoxText(dialogBox, textId);` but then how does it return a different account compared to the first invocation?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+            if (accountName2 != null)
+                accounts.add(accountName2);
+        }
+        return accounts;
+    }
+
+    /**
+     * Removes the added account
+     */
+    public void removeAccount(String username) {
```
shahzaibj: nit: NonNull final

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        final UiObject removeAccount = UiAutomatorUtils.obtainUiObjectWithResourceId(
+                "com.microsoft.identity.testuserapp:id/buttonRemoveAccount"
+        );
+        UiAutomatorUtils.obtainChildInScrollable("someone@contoso.com");
```
shahzaibj: Is this intentional?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        UiAutomatorUtils.handleInput("com.microsoft.identity.testuserapp:id/editTextUsername", username);
+        try {
+            removeAccount.click();
+        } catch (UiObjectNotFoundException e) {
```
shahzaibj:
```suggestion
        } catch (final UiObjectNotFoundException e) {
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        }
+    }
+
+    /**
+     * Decode the sso token and look for the requested nonce
+     */
+    public void confirmNoncePresent() {
+        // fill the nonce
+        UiAutomatorUtils.obtainChildInScrollable("nonce");
+        UiAutomatorUtils.handleInput("com.microsoft.identity.testuserapp:id/editTextNonce", "testNonce");
```
shahzaibj: Looks like we're hard coding `testNonce` here...what's the utility of that?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+                final String[] str = decodedTxt.split("request_nonce => ");
+                if (str.length > 1) {
+                    Assert.assertEquals(str[1].trim(), "testNonce");
+                } else {
+                    Assert.fail("decoded token does not contain correct nonce");
+                }
+            } else {
+                Assert.fail("decoded token does not contain correct nonce");
+            }
+        } catch (UiObjectNotFoundException e) {
```
shahzaibj:
```suggestion
        } catch (final UiObjectNotFoundException e) {
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        // Look for the dialog box
+        final UiObject dialogBox = UiAutomatorUtils.obtainUiObjectWithResourceId(
+                "android:id/message"
+        );
+        Assert.assertTrue(dialogBox.exists());
+        try {
+            if (!dialogBox.getText().contains("Calling app could not be verified")) {
+                Assert.fail("Could not find the string 'calling app could not be verified' in the msg displayed in the dialog");
+            }
+        } catch (UiObjectNotFoundException e) {
```
shahzaibj:
```suggestion
        } catch (final UiObjectNotFoundException e) {
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: do you mean searching for the input box with hint as "someone@contoso.com"?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: It returns 2 accounts when we tap on the getAccounts button. which will show 2 dialog boxes one on top of the other. So when we do getDialogBoxText for the first time, we read the txt, get the first account and click on ok to dismiss the dialog. now we will still see the 2nd dialog box with a different account.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: testNonce is a random string entered in the edit box to get the ssoToken. Once we get it, we decode it to see if the random nonce (testNonce) is present in the result as well.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -275,7 +277,40 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+        try {
+            final UiObject flightsObj = UiAutomatorUtils.obtainUiObjectWithResourceId("com.microsoft.identity.testuserapp:id/editTextFlights");
+            String flightsJson = flightsObj.getText();
+            return flightsJson;
+        } catch (UiObjectNotFoundException e) {
+            throw new AssertionError(e);
+        }
+    }
 
+    private String getDialogBoxText(UiObject dialogBox, String textId) {
```
fadidurah: Nit: NonNull final

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -292,16 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
         }
     }
-
-    @Override
-    public void setFlights(@Nullable final String flightsJson) {
-        Logger.i(TAG, "Set Flights..");
+    /**
+     * Gets all the accounts added
+     */
+    public List<String> getAllAccounts(boolean expectingMultipleAccounts) {
```
fadidurah: Nit: final?

fadidurah (approved):
Looks good from my end. Agree with shahzaib's comments. Left some nits. Thanks for adding this functionality!

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: Yes. I'm not sure what we're trying to do here. It seems we're looking for some UiObject and then totally forgetting about it.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: How do you plan to use this method? It seems this method figures out a nonce by itself, fills it in, and then asserts for its existence. What's the benefit of such a method?

I think instead we need to have the following two methods that replace this method:

```java
/**
* Acquires an SSO token using the provided nonce.
*/
public String acquireSsoToken(String nonce);

/**
* Decodes the SSO Token that was acquired earlier.
*/
public Map<String, ?> decodeSsoToken(String expectedNonce);
```

Note: I don't think we necessarily need that second method to decode the SSO Token since we can just do that without UI using the JWT Parser and therefore I suggest that you only implement one method here to just acquire the SSO Token.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
shahzaibj: Okay, so we're just looping through the dialogs here. I think the assumption around 2 accounts is okay for now thought we may need to refine it later.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -293,15 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+            throw new AssertionError(e);
+        }
+    }
+
+    /**
+     * Decode the sso token and look for the requested nonce
+     */
+    public void confirmNoncePresent() {
+        // fill the nonce
+        UiAutomatorUtils.obtainChildInScrollable("nonce");
```
shahzaibj: I'm not sure what we're doing here. It seems we're trying to obtain a UiObject and then forgetting about it. I don't think we even need to obtain this object though

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -293,15 +327,111 @@ private final String basicButtonHandler(@NonNull final String resourceButtonId,
...
+            }
+        } catch (final UiObjectNotFoundException e) {
+            throw new AssertionError(e);
+        }
+    }
+
+    /**
+     * Confirm that the calling app is not verified
+     */
+    public void confirmCallingAppNotVerified() {
```
shahzaibj, line 438: It seems we're trying to verify an error scenario. When is someone supposed to call this method?

shahzaibj (changes requested):
Left some comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: This method will be called in the testcase https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1561652
I have added this new testcase to send the nonce and verify the nonce after decoding. Even if we pass the nonce from the testcase, we would be doing the same thing. But yes, sending it as a param is a better option.

I agree with your suggestion that we can break it into 2 functions, but if we just keep the first one and skip the decodeSSoToken call, we will have to duplicate the logic executed when decodeSSoToken button is pressed (MainActivity.java in testuserApp has the decode logic)

The testCase needs to do the following steps:
1. Assign random string to nonce
2. Call acquireSSOToken(nonce)
3. Call decodeSSoToken(expectedNonce /*same as nonce*/)

shahzaibj: I guess what I'm saying is that we don't need to click the decode button to perform the actual decode since we can just do it in code ourselves; and yes the logic will be duplicated however, the test will become faster due to one less UI interaction. With that said, doing it via the UI is also fine..so feel free to add both methods.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: this was done to first scroll and then fill with resourceId. Changed it to reuse the UIObject obtained after scrolling.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: This is being called in the test case 1600567 https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1600567

Here we perform multiple actions like getAccounts, removeAccount, getFlights etc and verify that the dialog always has the "calling App not verified" in the dialog.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: handled this in commit 060dcf041f3b0ff63dd5271687503dda4e4efc11

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
somalaya: added changes in 060dcf041f3b0ff63dd5271687503dda4e4efc11

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -258,6 +281,12 @@ public void enrollDevice(@NonNull final String username,
         enrollmentPerformedSuccessfully = true;
     }
 
+    @Override
+    public void enrollDevice(@NonNull final String username,
+                             @NonNull final String password) {
+        enrollDevice(username, password, false);
```
somalaya, line 278: The reason for making this change is because the flow for federated user to enroll using CP is different. the resourceIds of username and pwd fields are different.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -180,19 +182,40 @@ public void enrollDevice(@NonNull final String username,
...
+
+            final MicrosoftStsPromptHandler microsoftStsPromptHandler = new MicrosoftStsPromptHandler(promptHandlerParameters);
+
+            Logger.i(TAG, "Handle prompt in ADFS login page for enrolling device..");
+            // handle AAD login page for username
+            UiAutomatorUtils.handleInput("i0116", username);
+            UiAutomatorUtils.handleButtonClick("idSIButton9");
+
+            // handle ADFS login page for password
+            microsoftStsPromptHandler.getLoginComponentHandler().handlePasswordField(password);
```
shahzaibj: Why can't we just do 
`microsoftStsPromptHandler.handlePrompt(username, password)`

just like we do in the non-federated case?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/GoogleSettings.java
```diff
@@ -307,7 +307,7 @@ private void handleDoneButton() throws UiObjectNotFoundException {
         if (android.os.Build.VERSION.SDK_INT == 28) {
             UiAutomatorUtils.handleButtonClick("com.android.settings:id/redaction_done_button");
         } else {
-            final UiObject doneButton = UiAutomatorUtils.obtainUiObjectWithExactText("Done");
+            final UiObject doneButton = UiAutomatorUtils.obtainUiObjectWithExactText("DONE");
```
shahzaibj: Which device are you testing on? I don't expect this to change for API 29

somalaya: Pixel 4 emulator

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
somalaya: handlePrompt calls handlePasswordField method, which has a different resourceId for pwd field from what we see in non-federated case.
 UiAutomatorUtils.handleInput("i0118", password) is used in AadLoginComponentHandler and UiAutomatorUtils.handleInput("passwordInput", password) is used in AdfsLoginComponentHandler for autofilling  pwd  input.

shahzaibj: Okay, in that case what you need to do is that for federated case, instead of using `MicrosoftStsPromptHandler` rather use the `AdfsPromptHandler`

somalaya: Yes Shahzaib, I have sent the promptParameters as federated = true to create an object of AdfsPromptHandler only. 
MicrosoftStsPromptHandler takes this boolean and creates an obj of AADHandler or AdfShandler accordingly

shahzaibj: then in that case it should be working already. I mean internally the AdfsPromptHandler would perform handleInput with the correct resource Id. Is that not working as expected?

somalaya: i think the reason why it was not working earlier is because we don't have an override method of handlePrompt in ADFS Handler. 
 Let's say I add this method and call handleEmail and handlePasswordField which are already present in the ADFS Handler, it will still fail. That is because we first get the CP prompt, which is AAD prompt. Here we still need to use AADHandler's handleEmail method for email. Once I fill this, it will realize that it is a federated user, so takes us to the next screen where we have the username auto-populated but password still needs to be filled. This is the reason why i used UiAutomatorUtils.handleInput("i0116", username); for username but handlePasswordField method of ADFSHandler seperately.

I can add this in ADFSHandler 
void handlePrompt(username, pwd, isFederated) {
// To handle AAD prompt for email
UiAutomatorUtils.handleInput("i0116", username); -> This one is email on AAD screen not ADFS
// To handle ADFS prompt for pwd
handlePasswordField(password);
}

shahzaibj: We discussed offline that this should be improved and we should be reusing the Login Component Handler and Prompt Handler code for this rather than hard coding resource ids. It can be improved in a different PR or this one. Totally up to @somalaya

shahzaibj (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/GoogleSettings.java
somalaya: I see that this is still Done in the physical device. reverting this. For emulator on Pixel 4, i have to make this change locally.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -173,7 +171,7 @@ public void handleFirstRun() {
     }
 
     public void enrollDevice(@NonNull final String username,
-                             @NonNull final String password, @NonNull final boolean isFederated) {
+                             @NonNull final String password, @NonNull final Boolean isFederated) {
```
shahzaibj: Why `Boolean` as opposed to `boolean`?

somalaya: I hve now reverted it to boolean and also removed NonNull annotation

PR #1753, opened 2022-06-06
Fix ignored authority_url when creating the authority audience

gitaumoses4:
## The Problem
When the authority `type` is AAD, `authority_url` value of the `authority` is not being considered when deserializing the msal config object.  This results in the audience created for the authority to use the public cloud. 

## Why is this happening
When deserializing the authority, through the `AuthorityDeserializer` we generate the audience's cloudUrl by invoking `aadAuthority.getAuthorityUri()`. This method is overriden in the `AzureActiveDirectoryAuthority` class to return the cloud URL from the authority. This flows as follows.

1. We deserialize the authority object into an `AzureActiveDirectoryAuthority` object.
2. We try to create the audience  for this object by invoking `getAuthorityUri()` 
3. The `getAuthorityUri()` method builds the authority uri by making use of the audience's cloudUrl (which we are trying to create in the first place)
4. since the audience's cloudUrl is null, it is instantiated to the public cloud and returned as the authority URL.

## Changes made
Revert back to the original implementation before the regression.

## Reproducing the bug

1. Checkout to the `dev` branch on common. 
2. Run MSAL test app.
3. Choose `ARLINGTON` config file. This has its authority_url set to the US Gov cloud.
4. Try to acquire token with device code flow.

###**Expectation**
You should get a message with the device user code as shown below.
Screenshot_20220606_130755

###**Actual**
The device code flow fails with an error as shown below.
Screenshot_20220606_131249

## Testing
1. Checkout to `git/fix-authority-url` branch
2. Run MSAL test app
3. Choose `ARLINGTON` config file. This file has its authority_url set to the US Gov cloud
4. You should be able to acquire token with device code flow successfully. The verification URL should be "https://microsoft.com/deviceloginus"


common4j/src/main/com/microsoft/identity/common/java/authorities/AuthorityDeserializer.java
```diff
@@ -60,12 +61,15 @@ public Authority deserialize(JsonElement json, Type typeOfT, JsonDeserialization
 
                     if (aadAuthority != null) {
                         try {
-                            final CommonURIBuilder uri = new CommonURIBuilder(aadAuthority.getAuthorityUri());
+                            // Use authority_url provided by the developer, otherwise, use the audience cloudUrl from aadAuthority.getAuthorityUri()
```
rpdome: The fix should be in AzureActiveDirectoryAuthority.

rpdome:
My understanding is that this is a regression. 

If that's the case, could you please go back to the non-regressed one and see what was the previous behavior? 

If the app sets BOTH authority and authority_url, which one would have a higher precedence?

common4j/src/main/com/microsoft/identity/common/java/authorities/AuthorityDeserializer.java
gitaumoses4: > My understanding is that this is a regression.
> 
> If that's the case, could you please go back to the non-regressed one and see what was the previous behavior?
> 
> If the app sets BOTH authority and authority_url, which one would have a higher precedence?

Yes, I based my implementation on the non-regressed one. Here's a snipped of what we had before in the `AuthorityDeserializer`

```java
                    // The developer might supply authority_url instead of audience.
                    // In that case, we'll try our best to map the audience here.
                    if (aadAuthority != null && aadAuthority.mAuthorityUrl != null) {
                        final Uri uri = Uri.parse(aadAuthority.mAuthorityUrl);
                        final String cloudUrl = uri.getScheme() + "://" + uri.getHost();
                        final String tenant = uri.getLastPathSegment();
                        if (!TextUtils.isEmpty(tenant)) {
                            aadAuthority.mAudience = AzureActiveDirectoryAudience.getAzureActiveDirectoryAudience(cloudUrl, tenant);
                        }
                    }
```

gitaumoses4:
> My understanding is that this is a regression.
> 
> If that's the case, could you please go back to the non-regressed one and see what was the previous behavior?
> 
> If the app sets BOTH authority and authority_url, which one would have a higher precedence?

Yes, I based my implementation on the non-regressed one. Here's a snippet of what we had before in the `AuthorityDeserializer`

```java
// The developer might supply authority_url instead of audience.
// In that case, we'll try our best to map the audience here.
if (aadAuthority != null && aadAuthority.mAuthorityUrl != null) {
    final Uri uri = Uri.parse(aadAuthority.mAuthorityUrl);
    final String cloudUrl = uri.getScheme() + "://" + uri.getHost();
    final String tenant = uri.getLastPathSegment();
    if (!TextUtils.isEmpty(tenant)) {
        aadAuthority.mAudience = AzureActiveDirectoryAudience.getAzureActiveDirectoryAudience(cloudUrl, tenant);
    }
}
```
Link: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/release/3.6.7/common/src/main/java/com/microsoft/identity/common/internal/authorities/AuthorityDeserializer.java

rpdome:
> > My understanding is that this is a regression.
> > If that's the case, could you please go back to the non-regressed one and see what was the previous behavior?
> > If the app sets BOTH authority and authority_url, which one would have a higher precedence?
> 
> Yes, I based my implementation on the non-regressed one. Here's a snippet of what we had before in the `AuthorityDeserializer`
> 
> ```java
> // The developer might supply authority_url instead of audience.
> // In that case, we'll try our best to map the audience here.
> if (aadAuthority != null && aadAuthority.mAuthorityUrl != null) {
>     final Uri uri = Uri.parse(aadAuthority.mAuthorityUrl);
>     final String cloudUrl = uri.getScheme() + "://" + uri.getHost();
>     final String tenant = uri.getLastPathSegment();
>     if (!TextUtils.isEmpty(tenant)) {
>         aadAuthority.mAudience = AzureActiveDirectoryAudience.getAzureActiveDirectoryAudience(cloudUrl, tenant);
>     }
> }
> ```
> 
> Link: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/release/3.6.7/common/src/main/java/com/microsoft/identity/common/internal/authorities/AuthorityDeserializer.java

Ok I see what's going on now... thanks for the context.

Would you mind reverting the code to be exactly like this? 

What happened here is that if the serialized mAuthorityUrl exists, then it'll take a higher precedence and will overwrite mAudience. 

If the value is not set, then we don't have to do this whole thing in the first place.

(please also add the description as comments in the code too).

common4j/src/test/com/microsoft/identity/common/java/authorities/AuthorityDeserializerTest.java
```diff
@@ -0,0 +1,86 @@
...
+        gson = new GsonBuilder()
+                .registerTypeAdapter(Authority.class, new AuthorityDeserializer())
+                .create();
+    }
+
+    @Test
+    public void testDeserializeB2C() {
+        final Authority authority = gson.fromJson(B2C_AUTHORITY, Authority.class);
+
+        Assert.assertTrue(authority instanceof AzureActiveDirectoryB2CAuthority);
```
shahzaibj, line 58: Let's also add an assert for the authority URL here

common4j/src/main/com/microsoft/identity/common/java/util/CommonURIBuilder.java
```diff
@@ -159,4 +158,14 @@ private boolean containsParam(@NonNull final String param){
         return false;
     }
 
+    /**
+     *
+     * @return the last segment in the URI. Returns an empty string if there are no path segments.
+     */
+    public String getLastPathSegment() {
+        final List<String> pathSegments = getPathSegments();
+        if (pathSegments.isEmpty()) return "";
```
shahzaibj: Please follow the code format that we have in the library for `if` statements

gitaumoses4: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1753/commits/f58ff6ec019f1ebe2587d702c6e465d7acbbea0f

common4j/src/test/com/microsoft/identity/common/java/authorities/AuthorityDeserializerTest.java
gitaumoses4: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1753/commits/f58ff6ec019f1ebe2587d702c6e465d7acbbea0f

shahzaibj (approved):

rpdome (approved):

PR #1687, opened 2022-03-03
Fix silent flow pkeyauth, add build param to disable silent flow timeout during debugging

rpdome:
Related: https://github.com/AzureAD/ad-accounts-for-android/pull/1827, https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1597

## Why?
1. We're not sending x-ms-PKeyAuth to both auth and token endpoint. 
       - we're only sending it to ADAL's token endpoint (which has a totally separate PKeyAuth stack).
2. However, for some reason, the auth endpoint sends us back PKeyAuth (tested in Broker) - that's why we're not aware that we're not sending x-ms-PKeyAuth there.
3. We do not get PKeyAuth in token endpoint's response, though.

## What I did
1. Make sure we send x-ms-PKeyAuth in both interactive and silent flow.
2. Make sure we're handling the content properly[ based on the spec](https://microsoft.sharepoint.com/:w:/t/aad/devex/EfksUgoBpPNGnACpeQ-_puwBj5JZ616HHxQYxvKJm3JTMg?e=USnvFj).
4. Add tests

## How I tested
1. Install broker
2. Install MSAL test app, signs in interactively* with idlab@msidlab4.onmicrosoft.com 
3. verify the token with jwt.ms, it should NOT have any deviceid claim in it.
5. Perform silent request with idlab@msidlab4.onmicrosoft.com **_with device id claim_**
6. You should get the following error back...

MsalUiRequiredException.
AADSTS50187: Failed to perform device authentication.
Trace ID: ed0290e9-2175-4f5c-85a5-1551076a8900
Correlation ID: b7bad247-913d-499d-b5de-3e94dd474484
Timestamp: 2022-03-03 23:58:12Z

7. signs in interactively with idlab1@msidlab4.onmicrosoft.com **_with device id claim_**
8. User is prompted for a WPJ registration
9. Click "Register"
10. Get a token back, the token should have deviceid claim in it.
11. Perform silent request with idlab@msidlab4.onmicrosoft.com **_with device id claim_**
    - We have a non-deviceid claim RT in the token cache for this account from step 2, so this will trigger pkeyauth in silent flow.
12. You should get a token back, now the device id claim should be in it.

* by signing in interactively, I mean do not enter login_hint, and select "Use another account" if prompted.
-----

I also added build param to disable silent flow timeout during debugging.
 
[...]

PR #1755, opened 2022-06-09
Do not send x-ms-PKeyAuth header to both auth and token endpoint.

p3dr0rv:
Sending the x-ms-PKeyAuth to both auth and token endpoint, included in the PR #1687 is considered as a "breaking change" fix. 

### Why
The clients in this particular scenario depend on that we are not sending the PkeyAuthHeader for the request to fail and fallback to the interruptflow without going through the performPKeyAuthRequest. (see [performTokenRequest](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/e1f1a640c612f650764158eab23ef1c9968e09a4/common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java#L492))

As up today we do not know how many customers rely on this broken flow. We decide to revert this change in this current release (5.0.0) and add it to the next one as a major release with breaking changes.

related ICM: [31201066](https://portal.microsofticm.com/imp/v3/incidents/details/312010665/home)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -306,7 +306,8 @@ public abstract class BaseController {
             completeRequestHeaders.put(AuthenticationConstants.AAD.APP_VERSION,
                     parameters.getApplicationVersion()
             );
-            completeRequestHeaders.put(PKEYAUTH_HEADER, PKEYAUTH_VERSION);
+            // ADO:TODO:1934500 - Reverting this change as this is considered a "breaking change" fix.
+            //completeRequestHeaders.put(PKEYAUTH_HEADER, PKEYAUTH_VERSION);
 
             // Add additional fields to the AuthorizationRequest.Builder to support interactive
             setBuilderProperties(builder, parameters, interactiveTokenCommandParameters, completeRequestHeaders);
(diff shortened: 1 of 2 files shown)


p3dr0rv:
@rpdome @gitaumoses4

shahzaibj:
We want to revert both auth and token or only token?

p3dr0rv:
> We want to revert both auth and token or only token?

Token endpoint is the one that is affecting the customer.
Regarding to auth endpoint, from Dome comment in #1687 "However, for some reason, the auth endpoint sends us back PKeyAuth (tested in Broker) " so I decided to revert all.

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -306,7 +306,8 @@ protected final AuthorizationRequest.Builder initializeAuthorizationRequestBuild
             completeRequestHeaders.put(AuthenticationConstants.AAD.APP_VERSION,
                     parameters.getApplicationVersion()
             );
-            completeRequestHeaders.put(PKEYAUTH_HEADER, PKEYAUTH_VERSION);
```
rpdome: we should still send this as this is going to the auth endpoint, not token endpoint.

rpdome (approved):

PR #1765, opened 2022-06-17
Fix UI automation for authenticator app's device registration screen

somalaya:
**What** : UI Automation with authenticator app was broken in the device registration flow. This was because the resourceIds were removed from the settings screen on the authenticator app. Now Authenticator team has added some new resourceIds and minor string changes were made in the settings screen.

This PR fixes the automation flow by changing the resouceIds when we are testing on updated versions
We will still keep the previous files with old resourceIds to be able to test the old versions of authenticator.

**How** : Created a factory to generate the BrokerAuthenticatorPreviousVersionImpl when we are testing on previous versions of authenticator app and to generate the BrokerAuthenticatorUpdatedVersionImp when we are testing with the latest version of authenticator app. The factory does a version compare against a hardcoded version which is 6.2206.3949 (This version has the new resourceIds added)
The method setBrokerImpl in the factory will be invoked immediately after install() method in App.java . This will ensure that the app is already installed (during the rules evaluation) and the version is available to decide on the implementation to be used.

Note : Since the factory class is a generic one, we can extend it to other broker apps as well.

**Testing** : Tested existing test cases that were earlier breaking with the latest auth app are now running successfully.
Existing testcases are also running successfully with the old version of authenticator app (version before the resourceIds were removed)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorPreviousVersionImpl.java
```diff
@@ -0,0 +1,124 @@
...
+            Assert.assertTrue(
+                    "Microsoft Authenticator - Registered account upn matches provided upn.",
+                    currentRegistration.getText().equalsIgnoreCase(username)
+            );
+        } catch (final UiObjectNotFoundException e) {
+            throw new AssertionError(e);
+        }
+    }
+
+    public void performSharedDeviceRegistration(@NonNull final String username,
```
shahzaibj, line 98: Is this overriden? Override annotation?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerImplFactory.java
```diff
@@ -0,0 +1,64 @@
...
+package com.microsoft.identity.client.ui.automation.broker;
+
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+
+import androidx.test.core.app.ApplicationProvider;
+
+/**
+ * Factory class to create broker implementations based on the version of Broker app under test
```
shahzaibj:
```suggestion
 * Factory class to create broker implementations based on the version of Broker app under test.
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerImplFactory.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class BrokerImplFactory {
```
shahzaibj: I thought you were going to make this an AppImplFactory as opposed to just BrokerImpl

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerImplFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public final static String LATEST_VERSION_NUMBER = "6.2206.3949";
+
+    /**
+     * Set broker implementation based on the version of the app installed
+     * @param appName appName of the app installed that acts as brokerImpl
+     */
+    public static void setBrokerImpl(String appName) {
```
shahzaibj: I don't think we should do this based on appName...probably something more specific than that

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -63,9 +64,12 @@ public class BrokerMicrosoftAuthenticator extends AbstractTestBroker implements
 
     private final static String INCIDENT_MSG = "Broker Automation Incident";
 
-    public static final String TAG = BrokerMicrosoftAuthenticator.class.getSimpleName();
+    private static final String TAG = BrokerMicrosoftAuthenticator.class.getSimpleName();
+
+    protected boolean isInSharedDeviceMode = false;
+
+    public static BrokerMicrosoftAuthenticator brokerMicrosoftAuthenticatorImpl;
```
shahzaibj: I don't think we want this to be a static that is set from a different class

shahzaibj: This should really be a Non-Static, `protected` member variable that is set directly in the `install` method. It doesn't need to be set from anywhere else

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerImplFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            final Context context = ApplicationProvider.getApplicationContext();
+            try {
+                PackageInfo packageInfo = context.getPackageManager().getPackageInfo("com.azure.authenticator", 0);
+                // authenticator app follows version number format - V.YYMM.XXXX
+                // V is the major version, YYMM are last two digits of an year followed by month
+                // XXXX is number of hours passed after Jan1st 00.00 of current year
+                // String comparison of versions should work for this format
+                if (packageInfo.versionName.compareTo(LATEST_VERSION_NUMBER) >= 0) {
+                    // Use latest automation code as this is the new updated authenticator app
+                    BrokerMicrosoftAuthenticator.brokerMicrosoftAuthenticatorImpl = new BrokerAuthenticatorUpdatedVersionImpl();
```
shahzaibj: See my other comments. We don't want static state. We should try to rather return an instance of `IApp` from the factory. Factories generally create and return objects. Factories aren't meant to have this sort of `void` methods that assign a static variable in a different class.

shahzaibj (changes requested):
Left some comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerImplFactory.java
somalaya: Ideally App.java should not be assigning anything to instance variable of child classes. So, I changed the design to add an abstract method setAppImpl() in IApp.java now. And everytime an app needs to decide on the implementation, it will override this method and add its own criteria to decide on the implementation.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
somalaya: I am now using a private member and setting it directly from BrokerMicrosoftAuthenticator only.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorPreviousVersionImpl.java
somalaya: added this annotation in the latest commit 409b0ee9b1ff1f1b4f95466b8300946dcf71e57b

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/IApp.java
```diff
@@ -83,4 +83,9 @@ public interface IApp {
      * Copy Installed APK to supplied destination.
      */
     void copyApk(String destApkFileName);
+
+    /**
+     * Determines the implementation to use based on the app version
+     */
+    void setAppImpl();
```
shahzaibj: I was actually hoping that it would be a "get" method rather than a "set"

Also I think this probably doesn't need to go in `IApp`, probably rather into the `App.java` as an `abstract` method. Or we can actually define a default implementation that just returns `this`

shahzaibj: then inside install you can assign as 

`mAppImpl = getAppImpl()`

somalaya: Shahzaib, can you elaborate on why it should be get method? My intention to write setter is to set the impl based on the version. I can move it to app.java instead of IApp.java but all methods like install, forceStop are currently in IApp.java. So i thought this would be consistent with other methods.

shahzaibj: The rest of the methods are inside `IApp` because those are actual functionalities provided by definition of an app. Methods that are declared on an interface are `public` by nature and can be invoked by anyone from anywhere. `getAppImpl` method however, is sort of a custom functionality that I believe doesn't actually belong to an app definition, and therefore I think it shouldn't be publicly accessible i.e. not anyone should be able to call that method to get or set the App Impl. For this reason, I think we can move this method to `App.java` which is an abstract class and abstract classes can allow to declare both implemented as well as abstract methods with different visibility modifiers such as `public`, `protected` as well as `package-private`. I believe the `getAppImpl` method can be `protected` so that child classes can define their implementation of the impl, but only child classes can now see this method and no outsider class can see that method. I hope that makes sense.

Furthermore, regarding the "get" vs "set" - Setters generally accept an argument and set that. I don't see your method doing that so it is not really a setter. The way I see it, what your method is doing is really "getting or creating" an App Impl and then actually sets that impl somewhere as well, but I wouldn't call it a setter.

I mean a setter would be the following:

`public void setAppImpl(IApp appImpl) {
   mAppImpl = appImpl
}
`

Your method does the following however:
`public void setAppImpl() {
   mAppImpl = createAppImpl();
}`

So I think it can really refactor to the following:

```
class App {
 // the below is actually going to be assigned inside install method I believe
  mAppImpl = getAppImpl(); // this can be renamed to createAppImpl

  protected App getAppImpl() {
     return this; // or create impl based on version for child classes
  }
}
```

somalaya: Shahzaib, I moved it to App.java and made it an abstract method in my local. But I wanted to close on the get/set part of it. If setAppImpl is giving the impression of a setter, then we can probably change it to initializeAppImpl. The issue I see with getAppImpl is that, I dont want to get it right away. I want to initialise the impl in install() and then when I call the individual methods like performDeviceRegistration, the initialised impl will simply be used. LMK if you think this could be an alternate solution?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -401,4 +304,24 @@ public void handleFirstRun() {
...
         shouldHandleFirstRun = false;
     }
-}
+
+    @Override
+    public void setAppImpl() {
+        final Context context = ApplicationProvider.getApplicationContext();
+        try {
+            PackageInfo packageInfo = context.getPackageManager().getPackageInfo("com.azure.authenticator", 0);
+            // authenticator app follows version number format - V.YYMM.XXXX
```
shahzaibj, line 313: Let's log the package info and version number in the logs

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/IApp.java
somalaya: I also tried implementing this with generics like we discussed offline. But I hit a blocker, I must be knowing the type of implementation while creating an object of BrokerMicrosoftAuthenticator, which is not possible in this case. To keep this simple, I left it as is for now. I will give it another try in another PR seperately

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
somalaya: Added logs in 813cf421a5097ac44bacba65961fc6f3689603e5

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/IApp.java
shahzaibj: Sounds good

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -101,8 +101,16 @@ public void install() {
...
         shouldHandleFirstRun = true;
+
+        // setting app implementation for the installed app based on version
+        initialiseAppImpl();
     }
 
+    /**
+     * Initialise implementation of app based on version
+     */
+    abstract protected void initialiseAppImpl();
```
shahzaibj, line 112: Thinking about it more, now that since we're not doing it with generics, I think this leading to the following code in a lot of places:

```java
 @Override
    public void initialiseAppImpl() {
       // nothing required
    }
```

I think this is going to cause some confusion for folks reading the code. I suggest that we either remove it from `App.java` and move it only to the authenticator class or if we want to keep it on `App.java` then we can provide a default implementation that sets `mAppImpl` to `this`

somalaya: If we go with first approach (" I suggest that we either remove it from App.java and move it only to the authenticator class"), then this is not scalable to support multiple versions for all apps. 
I think the second approach is what I tried to do earlier, but the problem is mAppImpl is not defined in App.java. It is defined in each of the app 's child classes. So, I went with the approach of having private impl variable defined in each class and in each class, when we want to support running on multiple app versions, we can remove the //nothing required here comment and add the relevant code. If "nothing required here" is causing the confusion, I can add a To-Do in each class to say that when we want to support multiple versions, we can add it in this function (initialiseAppImpl)

shahzaibj, line 112: Spelling error

```suggestion
    abstract protected void initializeAppImpl();
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -401,4 +304,30 @@ public void handleFirstRun() {
         UiAutomatorUtils.handleButtonClick("com.azure.authenticator:id/frx_skip_button");
         shouldHandleFirstRun = false;
     }
-}
+
+    @Override
+    protected void initialiseAppImpl() {
+        final Context context = ApplicationProvider.getApplicationContext();
+        try {
+            PackageInfo packageInfo = context.getPackageManager().getPackageInfo("com.azure.authenticator", 0);
```
shahzaibj, line 312:
```suggestion
            final PackageInfo packageInfo = context.getPackageManager().getPackageInfo("com.azure.authenticator", 0);
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -401,4 +304,30 @@ public void handleFirstRun() {
...
+                brokerMicrosoftAuthenticatorImpl = new BrokerAuthenticatorPreviousVersionImpl();
+                Logger.i(TAG, "Using previous implementation of authenticator");
+            } else {
+                Assert.fail("Authenticator app version is not supported");
+            }
+        } catch (final PackageManager.NameNotFoundException e) {
+            throw new AssertionError(e);
+        }
+    }
+}
```
shahzaibj, line 333: nit: new line ending

shahzaibj (approved):

wainaina (approved):
LGTM. Approved.

PR #1751, opened 2022-05-27
Implementing Cert Picker and PIN Dialogs for Smartcard CBA

melissaahn:
### Note
New smartcard CBA features will be merged into the feature branch `melissaahn/smartcardCBAFeature` until the feature is ready for public previews.
## The Feature
This PR builds on [my first PR](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1729) by using the established connection with the YubiKey device to start the process of smartcard CBA. The parts of the smartcard CBA feature that are being address in this PR are (with corresponding PBIs linked):

- Recognizing that a YubiKey is plugged in, and thus, the user wants to try to use certificates on their YubiKey.
    - [Smartcard Certificate Picker Should Pop Up When Smartcard is Initially Plugged In](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1872882)
- Showing the user a dialog that lists the certificates available and prompts them to select one for authentication.
    - [Design Smartcard Certificate Picker](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1879582) 
- Prompting the user to enter the PIN that will unlock their YubiKey.
    - [Design PIN Prompt](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1872500)
    - [Get PIN Prompt to Pop Up When Certificate is Selected in Smartcard Cert Picker](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1879583)
- Handling cases when user wants to exit the flow.
    - [Handle Cancel Scenarios in Smartcard Cert Picker and PIN Prompt](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1879636)
- Handling the case when the user becomes locked out of their YubiKey due to too many incorrect PIN attempts.
    - [Design 'Too many attempts' PIN Dialog](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1914775)
- Ensuring that on-device CBA will remain functional.
    - [Main Success Scenario Works For On-Device Certificates](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1874347)
    - [On-Device Cert Picker should Pop Up When Smartcard is Not Plugged In](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1873478)  

This PR does NOT include the actual authentication with a private key, which is planned to be added next sprint cycle.
 
## Changes Made
[...]

PR #1766, opened 2022-06-17
Smartcard CBA Completed Flow

melissaahn:
### Note
New smartcard CBA features will be merged into the feature branch `melissaahn/smartcardCBAFeature` until the feature is ready for the next Broker and MSAL releases.
## The Feature
This PR builds on [my second PR](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1751) by using the information retrieved by the certificate picker and the PIN dialog to attempt CBA with a certificate located on a YubiKey device. The parts of the smartcard CBA feature that are being address in this PR are (with corresponding PBIs linked):

- Given the slot where a chosen certificate resides and a verified PIN, authentication via a modified private key can be attempted.
     - [Get main auth success scenario working for smartcard certificates](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1879631)
-  Testing and handling alternative flows, including early unplugging and no certificates on device (exceptions are handled with a general error dialog, but they have not been tested yet).
    - [Test For User Case Extensions and Alternate Flows](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1879653)
 
## Changes Made
Main changes:
- YubiKit and PIV version bumped up to `2.1.0-alpha.1`, which include methods that help facilitate client TLS authentication without having to copy the private key from the YubiKey device.
- `useSmartcardCertForAuth` method
    - Adds `PivProvider` in `Security` static list of security providers
    - Uses a `PivProvider` to set up KeyStore instance 
    - Sets up `PivPrivateKey` and certificate chain
    - Attempts to authenticate using `PivPrivateKey` object and certificate chain as parameters to `request.proceed()`
- `SmartcardErrorDialog` class takes in two string resource ids and displays a dialog when an error occurs.

## Result
When the user chooses a cert in the picker and successfully enters their PIN, they are either logged in successfully or met with an error page from the server.

If, at any point when a dialog is showing, the user decides to unplug their key, an error dialog will be shown reminding the user to replug their key in. The current CBA flow will be cancelled.
If the user does not have any certificates on their YubiKey device, they will be shown an error dialog informing them they do not have any relevant certificates on their device. The current CBA flow will be cancelled.

## Follow Ups
A private build of Authenticator with the smartcard CBA feature will be tested in private preview.
Need to ensure before GA that none of the logging statements will contain PII.

## Testing
Below are the general testing steps I'm planning on using to help form manual tests for Broker and MSAL. I'll need to contact Labs to see if they can provide test AAD accounts for CBA.
Post private preview, Robolectric will be used to test scenarios where exceptions can be thrown.

**Beginning of all scenarios:**
1. Download AzureSample.
2. Go to Multiple Account Mode.
3. Click "Get Graph Data Interactively".
4. Type the upn associated with the cert on key (or basically any other upn that brings up the "sign in with a certificate" link on the next page). Click "Next".
5. Plug in the YubiKey.
6. When the permission dialog pops up, click "OK".
7. Click "Sign in with a certificate". The custom cert picker should pop up with a listing of the certs currently held on the YubiKey.

**Success Scenario:**
8. Select the cert associated with the upn and click "Continue". The pin dialog should pop up.
9. Enter the YubiKey's pin and click "Unlock". The dialog should dismiss, and the page should be redirected to the Multiple Account fragment, where the user should successfully be logged in.
10. Repeat steps 3-9 with a second account.

**Incorrect Pin Scenario:**
8. Select the cert associated with the upn and click "Continue". The pin dialog should pop up.
9. Enter an incorrect YubiKey pin and click "Unlock". The EditText should clear and should have a red message underneath that says "The PIN you entered was incorrect".
10. Enter the YubiKey's pin and click "Unlock". The dialog should dismiss, and the page should be redirected to the Multiple Account fragment, where the user should successfully be logged in.

**Max Pin Attempts Allowed Scenario:**
8. Select the cert associated with the upn and click "Continue". The pin dialog should pop up.
9. Enter an incorrect YubiKey pin and click "Unlock". The EditText should clear and should have a red message underneath that says "The PIN you entered was incorrect". 
10. Repeat step 9 two more times. The pin dialog should be replaced with another dialog that states "Too many failed attempts, try again later."
11. Click "Ok" on the dialog and it should dismiss.
12. Click the back button.
13. Retype the upn and click "next".
14. Click "Sign in with a certificate". The same "Too many failed attempts" dialog should pop up, and the error page should be seen.

**Early YubiKey Unplugging Scenario:**
8. Unplug the YubiKey. The cert picker dialog should dismiss and a "Try Again" dialog should be shown.
9. Replug the YubiKey and click "Ok" when the permission prompt comes up.
10. Click the back button.
11. Retype the upn and click "next".
12. Click "Sign in with a certificate". The custom cert picker should pop up with a listing of the certs currently held on the YubiKey.
13. Select any cert and click "Next". The Pin dialog should pop up.
14. Unplug the YubiKey. The pin dialog should dismiss and a "Try Again" dialog should be shown.
16. Repeat steps 9-12.
17. Follow the "Max Pin Attempts Allowed Scenario", and when the error dialog shows up, unplug the YubiKey. The dialog should dismiss and a "Try Again" dialog should be shown.

**Switch To On-Device Certs Scenario:**
8. Click "Cancel". The dialog should dismiss and the error page should be shown.
9. Unplug the YubiKey.
10. Click the back button.
11. Retype the upn and click "next".
12.  Click "Sign in with a certificate". The default Android on-device cert picker should pop up.

All of the scenarios above should be repeated with brokerHost installed as well.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -526,10 +595,10 @@ public void onClick() {
                     }
                 },
                 mActivity);
-        //set current dialog to maxFailedAttemptsDialog and show.
+        //set current dialog to errorDialog and show.
         synchronized (sSmartcardDialogLock) {
-            mCurrentDialog = maxFailedAttemptsDialog;
-            maxFailedAttemptsDialog.show();
+            mCurrentDialog = errorDialog;
```
rpdome: if there is already a dialog, it should be dismissed (and also log - so that we know what dialog is dismissed.)

common/src/main/res/values/strings.xml
```diff
@@ -21,7 +21,17 @@
...
     <string name="smartcard_pin_layout_edittext_hint">PIN</string>
 
+    <string name="smartcard_error_dialog_positive_button">Ok</string>
+
     <string name="smartcard_max_attempt_dialog_title">Try again later</string>
     <string name="smartcard_max_attempt_dialog_message">Too many failed attempts, try again later.</string>
-    <string name="smartcard_max_attempt_dialog_positive_button">Ok</string>
+
+    <string name="smartcard_early_unplug_dialog_title">Try again</string>
+    <string name="smartcard_early_unplug_dialog_message">Smartcard device was removed too early. Please reinsert device before attempting to log in again.</string>
```
rpdome: nit: I would remove "too early".

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -138,19 +148,27 @@ public void invoke(@NonNull UsbYubiKeyDevice device) {
...
+                                //Note that Security.removeProvider will silently return if YKPiv doesn't exist.
+                                Security.removeProvider("YKPiv");
                                 //Call onCancel on current dialog, if dialog is showing.
                                 synchronized (sSmartcardDialogLock) {
                                     if (mCurrentDialog != null) {
                                         mCurrentDialog.onCancelCba();
-                                        //Reset dialog to null
-                                        mCurrentDialog = null;
-                                        //TODO: maybe show a dialog that tells user they unplugged too early? I'm not sure this is necessary since I'm thinking in a lot of cases, the user wanted to cancel current flow by unplugging.
+                                        //Show dialog informing users that they have unplugged their device too early.
```
rpdome: Wondering if we can do something like this

``` Java


@ThreadSafe
class DialogHolder {

    private Dialog mCurrentDialog = null;

    public synchronized void showErrorDialog(@NonNull final String title, @NonNull final String message){
        showDialog(new SmartcardErrorDialog(title, message));
    }
    
    public synchronized void dismissDialog(){
        showDialog(null);
    }
    
    public synchronized void showDialog(@Nullable final SmartcardDialog dialog){
        if (mCurrentDialog != null) {
            mCurrentDialog.dismiss();
        }
    
        mCurrentDialog = dialog;
    
        if (mCurrentDialog != null){
            // dispatch to main thread if not on main thread
            mCurrentDialog.show();
        }
    }
}

```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -512,11 +510,82 @@ public void run() {
...
+    /**
+     * Used to provide PivProvider constructor a Callback that will establish a new PivSession when it is needed.
+     * @return A Callback which returns a Callback that will return a new PivSession instance.
+     */
+    private Callback<Callback<Result<PivSession, Exception>>> getPivProviderCallback() {
+        return new Callback<Callback<Result<PivSession, Exception>>>() {
+            @Override
+            public void invoke(@NonNull final Callback<Result<PivSession, Exception>> callback) {
+                synchronized (sDeviceLock) {
+                    mDevice.requestConnection(UsbSmartCardConnection.class, new Callback<Result<UsbSmartCardConnection, IOException>>() {
```
rpdome, line 549: mDevice could be unplugged during the process.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -208,9 +228,16 @@ public void invoke(@NonNull final Result<UsbSmartCardConnection, IOException> va
                     final PivSession piv = new PivSession(c);
                     //Check if there are no PIN attempts remaining.
                     if (piv.getPinAttempts() == 0) {
-                        showMaxAttemptsDialog();
+                        showErrorDialog(R.string.smartcard_max_attempt_dialog_title, R.string.smartcard_max_attempt_dialog_message);
```
rpdome: Not related to this line

I'm wondering if we need to call requestConnection twice (on line 222 and 423). Can we reuse the  PivSession object (since this whole ChallengeHandler is ONE session) on line 222 in 423's operation?

If we can do this, then we'll be able to get rid of mDevice null checks and duplicated connection codes.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -512,11 +510,82 @@ public void run() {
...
+     */
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private void useSmartcardCertForAuth(@NonNull final X509Certificate cert,
+                                         @NonNull final char[] pin,
+                                         @NonNull final String slotAlias,
+                                         @NonNull final PivSession piv,
+                                         @NonNull final ClientCertRequest request) {
+        final String methodTag = TAG + "useSmartcardCertForAuth:";
+        //Need to add a PivProvider instance to the beginning of the array of Security providers in order for signature logic to occur.
+        //Note that this provider is removed when the UsbYubiKeyDevice connection is closed.
```
rpdome, line 476: Very likely that we'll need to invoke [clearClientCertPreferences](https://developer.android.com/reference/android/webkit/WebView#clearClientCertPreferences(java.lang.Runnable)) when the connection is closed as well.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -512,11 +510,82 @@ public void run() {
...
+        Security.insertProviderAt(new PivProvider(getPivProviderCallback()), 1);
+        try {
+            //Using KeyStore methods in order to generate PivPrivateKey.
+            //Loading null is needed for initialization.
+            KeyStore keyStore = KeyStore.getInstance("YKPiv", new PivProvider(piv));
```
rpdome: nit: final, here and below.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
rpdome: Even if that's not the case... we probably could still refactor the connection code and its error handler to be a separate method like this.

```Java

public interface IPivSessionCallback {
    void onGetSession(@NonNull final PivSession p);
}


private void getActivePivSessionAsync(@NonNull final ClientCertRequest request, @NonNull final IPivSessionCallback callback){
    if (mDevice == null) {
        showErrorDialog();
        request.cancel();
        return;
    }

    mDevice.requestConnection(UsbSmartCardConnection.class, new Callback<Result<UsbSmartCardConnection, IOException>>() {
        @Override
        public void invoke(@NonNull final Result<UsbSmartCardConnection, IOException> value) {
            try {
                final SmartCardConnection c = value.getValue();
                final PivSession piv = new PivSession(c);

                // validate is piv session is valid, i.e. check pin attempt, etc. 

                callback.onGetSession(piv);
            } catch (.... Exceptions){
                showErrorDialog();
                request.cancel();
            }
        }
    });
}
```

Usage
``` Java
private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
    getActivePivSessionAsync(request, new IPivSessionCallback() {
        @Override
        void onGetSession(@NonNull final PivSession p){
            //load certs.
            //display certs.
        }
    });
}
```

rpdome (commented):
Leave some comments.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
rpdome: Would you mind playing around with the 'result cached' scenarios? (IIRC, you've tried this with on device CBA)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
rpdome: getActivePivSessionAsync() should also be reusable in getPivProviderCallback();

rpdome:
Can you also verify the following in the static pivprovider callback?

- what happened if you force throw an exception (runtime/non-runtime). Does the webview crashes? Or is it handled gracefully?

(Regardless, we should still log every exception in that block, and rethrow if needed)

- if it's handled gracefully, what's the behavior? Is it the same as calling cancel()?

- Can we display a dialog in that operation?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: hmm I don't detect any issues when testing with logging in multiple accounts in a row with both types of CBA and a mix. Since `clearClientCertPreferences` is called every time `WebViewAuthorizationFragment` is created or a back button is pressed while in `WebViewAuthorizationFragment`, I think it should be able to handle both scenarios.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: I think this is a good idea. Adding methods for showing picker and pin dialogs, as well as methods to help facilitate the onCancelCBA callback.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: > Not related to this line
> 
> I'm wondering if we need to call requestConnection twice (on line 222 and 423). Can we reuse the PivSession object (since this whole ChallengeHandler is ONE session) on line 222 in 423's operation?
> 
> If we can do this, then we'll be able to get rid of mDevice null checks and duplicated connection codes.

I'm not totally sure if ChallengeHandler is entirely in one session. 
My experience with this before was that I attempted to create an instance variable to hold the PivSession, but when I attempted to use it outside of the invoke method via a button click callback (which I believe all of the dialog button callbacks are outside of the initial invoke method), I got the `Failed to send 25 bytes` error we've seen before. I don't 100% remember if I tried passing the PivSession object as a parameter, but I wouldn't be surprised if I already tried it (I recall spending a long time trying to debug this).

I will do the `IPivSessionCallback` refactoring though.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Also noting here (for the future and myself) that calling `dismiss()` on a dialog multiple times doesn't seem to cause any errors. Relevant in this case since positive and negative buttons on dialogs automatically call `dismiss()`, but calling it again in the DialogHolder method doesn't cause issues.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: added null check and invoke Result.failure on callback in such case, which throws an Exception with the message "Exception: Instance UsbYubiKitDevice variable (mDevice) is null."

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: taken care of by DialogHolder.showDialog

melissaahn:
> 

Tried a few things:
- Attempted to show dialog straight in callback (within and outside of `runOnUiThread` -> crashed webview and app (for the app part, probably has to do with UI for adapter)
- Attempted to cancel Request, then invoke `Result.failure` on callback -> crashed webview and app due to request being cancelled twice.
- Attempted to just invoke `Result.failure` in callback -> webview closes, see MsalServiceException "ERR_SSL_CLIENT_AUTH_SIGNATURE_FAILED" in body of AzureSample. No fatal errors, but notably this is NOT the same as calling request.cancel().

For private preview, the last option is probably the best. I probably have to experiment and debug a bit more to see what can be done otherwise.

melissaahn:
Addressed feedback by implementing changes and responding to comments.
If anyone has a YubiKey and wants to test:
- I need to provide you a cert and account, so ping me.
- MSAL branch: melissaahn/YubiKeyConfigChanges
- broker branch: melissaahn/YubiKitOverrideIntegration
- I've been using AzureSample and BrokerHost.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -71,10 +83,9 @@ public final class ClientCertAuthChallengeHandler implements IChallengeHandler<C
     private final Activity mActivity;
     private final YubiKitManager mYubiKitManager;
     private UsbYubiKeyDevice mDevice;
-    private SmartcardDialog mCurrentDialog;
-    //Locks to help facilitate synchronization
+    private final DialogHolder dialogHolder;
```
shahzaibj:
```suggestion
    private final DialogHolder mDialogHolder;
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -138,22 +283,27 @@ public void invoke(@NonNull UsbYubiKeyDevice device) {
...
-                                        Logger.verbose(TAG, "YubiKey was disconnected while dialog was still displayed.");
-                                    }
+                                //Remove the YKPiv security provider if it was added.
+                                //Note that Security.removeProvider will silently return if YKPiv doesn't exist.
+                                Security.removeProvider("YKPiv");
+                                //Show an error dialog only if a dialog is still showing.
+                                if (dialogHolder.isDialogShowing()) {
+                                    //Call Cancel Cba callback
+                                    dialogHolder.onCancelCba();
+                                    //Show dialog informing users that they have unplugged their device too early.
```
shahzaibj, line 190: How do we know at this layer that it was too early? Is it possible that the flow was actually successfully completed?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -194,53 +344,98 @@ private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
         final String methodTag = TAG + ":handleSmartcardCertAuth";
         //Show error dialog and cancel flow if mDevice is null.
         if (mDevice == null) {
-            //TODO: show error dialog.
+            Logger.info(methodTag, "Instance UsbYubiKitDevice variable (mDevice) is null.");
```
shahzaibj: log as error?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -194,53 +344,98 @@ private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
...
+                        dialogHolder.showErrorDialog(R.string.smartcard_max_attempt_dialog_title, R.string.smartcard_max_attempt_dialog_message);
                         request.cancel();
                         Logger.info(methodTag,  "User has reached the maximum failed attempts allowed.");
+                        // NOTE: for testing purposes, reset pin attempts remaining.
+                        // Take lines below out when in review
+                        try {
+                            piv.unblockPin("12345678".toCharArray(), "123456".toCharArray());
+                        } catch (InvalidPinException invalidPinException) {
+                            //for testing, so nothing needed rn
+                        }
```
shahzaibj: Do these need to be removed now?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -194,53 +344,98 @@ private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
...
+                                @Override
+                                public void onCancel() {
+                                    request.cancel();
+                                }
+                            });
+                } catch(final IOException | ApduException | BadResponseException e) {
                     Logger.error(methodTag, e.getMessage(), e);
+                    //Show general error dialog.
+                    dialogHolder.showErrorDialog(R.string.smartcard_general_error_dialog_title, R.string.smartcard_general_error_dialog_message);
                     request.cancel();
```
shahzaibj: Does the order of showing dialog and canceling the request matter?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -194,53 +344,98 @@ private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
...
+     * Requests a connection from UsbYubiKeyDevice instance in order to run IPivSessionCallback code that depends on a PivSession instance.
+     * @param request ClientCertRequest received from AzureActiveDirectoryWebViewClient.onReceivedClientCertRequest.
+     * @param callback Callback code that utilizes PivSession.
+     */
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private void getActivePivSessionAsync(@NonNull final ClientCertRequest request, @NonNull final IPivSessionCallback callback){
+        final String methodTag = TAG + "getActivePivSessionAsync:";
+        //If mDevice is null, show a general error dialog and cancel request.
+        if (mDevice == null) {
+            dialogHolder.showErrorDialog(R.string.smartcard_general_error_dialog_title, R.string.smartcard_general_error_dialog_message);
```
shahzaibj: We probably want to log an error here as well

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,43 +605,91 @@ public void run() {
...
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private void useSmartcardCertForAuth(@NonNull final X509Certificate cert,
+                                         @NonNull final char[] pin,
+                                         @NonNull final String slotAlias,
+                                         @NonNull final PivSession piv,
+                                         @NonNull final ClientCertRequest request) {
+        final String methodTag = TAG + "useSmartcardCertForAuth:";
+        //Need to add a PivProvider instance to the beginning of the array of Security providers in order for signature logic to occur.
+        //Note that this provider is removed when the UsbYubiKeyDevice connection is closed.
+        Security.insertProviderAt(new PivProvider(getPivProviderCallback(request)), 1);
```
shahzaibj: Can you add a comment here indicating that `1` maps to index `0`

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,43 +605,91 @@ public void run() {
...
+        try {
+            //Using KeyStore methods in order to generate PivPrivateKey.
+            //Loading null is needed for initialization.
+            final KeyStore keyStore = KeyStore.getInstance("YKPiv", new PivProvider(piv));
+            keyStore.load(null);
+            //PivPrivateKey implements PrivateKey
+            final PivPrivateKey pivPrivateKey = (PivPrivateKey) keyStore.getKey(slotAlias, pin);
```
shahzaibj: Can we run into ClassCastException here? Should we be checking type first?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,43 +605,91 @@ public void run() {
...
+    private Callback<Callback<Result<PivSession, Exception>>> getPivProviderCallback(ClientCertRequest request) {
+        final String methodTag = TAG + "getPivProviderCallback:";
+        return new Callback<Callback<Result<PivSession, Exception>>>() {
+            @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+            @Override
+            public void invoke(@NonNull final Callback<Result<PivSession, Exception>> callback) {
+                synchronized (sDeviceLock) {
+                    //Show error dialog and cancel flow if mDevice is null.
+                    if (mDevice == null) {
+                        Logger.info(methodTag, "Instance UsbYubiKitDevice variable (mDevice) is null.");
```
shahzaibj: error?

gradle/versions.gradle
```diff
@@ -43,7 +43,7 @@ ext {
     uiAutomatorVersion = "2.2.0"
     msal4jVersion = "1.10.0"
     mseberaApacheHttpClientVersion = "4.5.8"
-    yubikitAndroidVersion = "2.0.0"
+    yubikitAndroidVersion = "2.1.0-alpha.1"
```
shahzaibj, line 46: Is the alpha version of their library considered stable?

melissaahn: no, but I don't think any major changes will be added before their next stable release. This CBA feature will use the alpha release for private preview, and then the final stable release will replace it before GA (when it'll be merged into dev).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: The current logic is that if any dialog is showing (that includes error dialogs), then we are in the middle of a flow. 
In the case of a successful flow, the pin dialog is first dismissed, and then `proceed` is called on the request object (this is done in the `useSmartcardCertForAuth` method). In this situation, `dialogHolder.isDialogShowing()` should return false, so the cancel callback nor the error dialog should run. 
It should be noted that `setOnClosed` is automatically called when the fragment is destroyed. So I do need to do some more testing/debugging to see the outcome when different exceptions occur, or other errors occur within the `proceed` method.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Doesn't the error logger need an exception object as a parameter? Should I be throwing an exception in this case?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Yes! Oops.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: I don't think so, but I tried to keep consistent (show dialog, then cancel). I'll check again and make sure I'm consistent here.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: I'll add a type check, just in case. If the key retrieved is not of type PivPrivateKey, an error will be logged, a generic error dialog will be displayed, and the request will be cancelled.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Change to logging as error and passing null for Exception parameter.

common4j/versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Wed May 12 20:08:39 UTC 2021
-versionName=2.0.0
+versionName=1.0.6-CBATest
```
rpdome, line 2: Probably not intended to check in?

common/src/main/res/values/strings.xml
```diff
@@ -21,7 +21,17 @@
     <string name="smartcard_pin_dialog_error_message">The PIN you entered was incorrect</string>
     <string name="smartcard_pin_layout_edittext_hint">PIN</string>
 
+    <string name="smartcard_error_dialog_positive_button">Ok</string>
```
rpdome: Gentle reminder that we'll need these translated before GA. Not sure if you've reached out to authapp team.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardErrorDialog.java
```diff
@@ -78,9 +85,7 @@ public void onClick(DialogInterface dialog, int which) {
                 dialog.setOnCancelListener(new DialogInterface.OnCancelListener() {
                     @Override
                     public void onCancel(DialogInterface dialog) {
-                        if (mPositiveButtonListener != null) {
-                            mPositiveButtonListener.onClick();
-                        }
+                        mPositiveButtonListener.onClick();
```
rpdome: this probably is not a positive dialog callback... it's more like an "on dismissed dialog callback".

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -121,12 +134,146 @@ public Slot getSlot() {
         }
     }
 
+    /**
+     * Builds and shows SmartcardDialog instances while keeping track of the current dialog being shown to the user.
+     */
+    @ThreadSafe
+    private class DialogHolder {
```
rpdome: nit: can we move this into a separate file?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -121,12 +134,146 @@ public Slot getSlot() {
...
+        public synchronized void showPinDialog(@NonNull final SmartcardPinDialog.PositiveButtonListener positiveButtonListener,
+                                               @NonNull final SmartcardPinDialog.NegativeButtonListener negativeButtonListener,
+                                               @NonNull final SmartcardPinDialog.CancelCbaCallback cancelCbaCallback) {
+            final SmartcardPinDialog pinDialog = new SmartcardPinDialog(
+                    positiveButtonListener,
+                    negativeButtonListener,
+                    cancelCbaCallback,
+                    mActivity);
+            //Show PinDialog, which should always be called after a positive button press.
+            mDialogHolder.showDialog(pinDialog);
```
rpdome: just showDialog?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -298,88 +488,38 @@ private void getAndPutCertDetailsInList(@NonNull final Slot slot, @NonNull final
...
-        //final String methodTag = TAG + ":getSmartcardCertPickerDialogPositiveButtonListener";
         return new SmartcardCertPickerDialog.PositiveButtonListener() {
             @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
             @Override
             public void onClick(@NonNull final YubiKitCertDetails certDetails) {
                 //Need to prompt user for pin and verify pin. The positive button listener will handle the rest of the CBA flow.
-                showPinDialog(certDetails, request);
+                mDialogHolder.showPinDialog(
+                        getSmartcardPinDialogPositiveButtonListener(certDetails, request),
+                        new SmartcardPinDialog.NegativeButtonListener() {
```
rpdome: nit: NegativeButtonListener and CancelCbaCallback looks somewhat redundant. Wondering if we just need only one OnCancelledCallback that can be used in both cases.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -194,53 +346,92 @@ private void handleSmartcardCertAuth(@NonNull final ClientCertRequest request) {
         final String methodTag = TAG + ":handleSmartcardCertAuth";
         //Show error dialog and cancel flow if mDevice is null.
         if (mDevice == null) {
```
rpdome: probably don't need this anymore, since this is covered in getActivePivSessionAsync()

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -396,55 +536,31 @@ public void onClick(@NonNull final String pin) {
                 synchronized (sDeviceLock) {
                     //Show error dialog and cancel flow if mDevice is null.
                     if (mDevice == null) {
-                        //TODO: show error dialog.
+                        Logger.error(methodTag, MDEVICE_NULL_ERROR_MESSAGE, null);
```
rpdome: probably don't need this anymore, since it's covered in getActivePivSessionAsync()

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -467,27 +583,16 @@ private void verifySmartcardPin(@NonNull final char[] pin, @NonNull final YubiKi
             piv.verifyPin(pin);
             //If pin is successfully verified, we will get the certificate and perform the rest of the logic for authentication.
             final X509Certificate cert = piv.getCertificate(certDetails.getSlot());
```
rpdome: Sorry for a late feedback: I'm wondering if it's possible to store X509Certificate itself into YubiKitCertDetails.
(seems like we only need an X509 and a slot there).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -467,27 +583,16 @@ private void verifySmartcardPin(@NonNull final char[] pin, @NonNull final YubiKi
...
-                    synchronized (sSmartcardDialogLock) {
-                        mCurrentDialog = null;
-                    }
-                }
-            });
-            request.cancel();
-            synchronized (sSmartcardDialogLock) {
-                mCurrentDialog.dismiss();
+            //Do a confirmation that the cert retrieved above is the same cert that was picked earlier by comparing with YubiKitCertDetails.
+            if (!cert.getSubjectDN().getName().equals(certDetails.getSubjectText()) || !cert.getIssuerDN().getName().equals(certDetails.getIssuerText())) {
```
rpdome: was going to say that you should use getEncoded() to compare certs (or compare their thumbprints)... but I don't think you actually need to compare it - if you store X509 in YubiKitCertDetails in the first place.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,43 +601,100 @@ public void run() {
...
+    private void useSmartcardCertForAuth(@NonNull final X509Certificate cert,
+                                         @NonNull final char[] pin,
+                                         @NonNull final String slotAlias,
+                                         @NonNull final PivSession piv,
+                                         @NonNull final ClientCertRequest request) {
+        final String methodTag = TAG + "useSmartcardCertForAuth:";
+        //Need to add a PivProvider instance to the beginning of the array of Security providers in order for signature logic to occur.
+        //Note that this provider is removed when the UsbYubiKeyDevice connection is closed.
+        //The position parameter is 1-based (1 maps to index 0).
+        Security.insertProviderAt(new PivProvider(getPivProviderCallback()), 1);
```
rpdome, line 497: Probably a good idea to check if there's any PivProvider being loaded in the static list (and log/telemet it).

rpdome (commented):
left a few comments. Super close to being done :)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: YubiKitCertDetails now contains an X509Certificate and its corresponding slot

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Removed this comparison

common/src/main/res/values/strings.xml
melissaahn: PR for PhoneFactor with updated strings.xml file is currently in review

melissaahn: PR is merged. After translations are complete, I'll transfer over the strings to Common in a separate PR

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: NegativeButtonListener has been removed and replaced with CancelCbaCallback. CancelCbaCallback is now called when the negative button is pressed, the back button is pressed, and when the yubikey is unplugged in the middle of the flow.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardErrorDialog.java
melissaahn: changed name to DismissCallback

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Added both Telemetry and Logging for the case where PivProvider is already in the static list, which is unexpected behavior. May create custom Telemetry Event in a subsequent PR for CBA.

melissaahn:
Added Telemetry to report if: 
- PivProvider already exists in the Security static list
- PivProvider is being added to static list
- PivProvider is being removed from static list
Please let me know if I'm going about this the correct/wrong way. I'm thinking about creating a custom Event for CBA in a subsequent PR post private preview.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -68,49 +81,40 @@
 public final class ClientCertAuthChallengeHandler implements IChallengeHandler<ClientCertRequest, Void> {
     private static final String TAG = ClientCertAuthChallengeHandler.class.getSimpleName();
     private static final String ACCEPTABLE_ISSUER = "CN=MS-Organization-Access";
+    private static final String MDEVICE_NULL_ERROR_MESSAGE = "Instance UsbYubiKitDevice variable (mDevice) is null.";
+    private static final String YKPIV = "YKPiv";
```
rpdome: nit: rename to YubiKeyProvider

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,41 +440,116 @@ public void run() {
...
+            Telemetry.emit((BaseEvent) new BaseEvent().put(
+                    TelemetryEventStrings.Key.IS_EXISTING_PIVPROVIDER_PRESENT,
+                    String.valueOf(true)
+            ));
+            Logger.info(methodTag, "Existing PivProvider was present in Security static list.");
         }
+        //The position parameter is 1-based (1 maps to index 0).
+        Security.insertProviderAt(new PivProvider(getPivProviderCallback()), 1);
+        //Telemet and log that PivProvider is being added to Security static list.
+        Telemetry.emit((BaseEvent) new BaseEvent().put(
```
rpdome: We probably won't need this - since it'll be emitted 100% of the time.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -138,21 +155,34 @@ public void invoke(@NonNull UsbYubiKeyDevice device) {
...
-                                        mCurrentDialog = null;
-                                        //TODO: maybe show a dialog that tells user they unplugged too early? I'm not sure this is necessary since I'm thinking in a lot of cases, the user wanted to cancel current flow by unplugging.
-                                        Logger.verbose(TAG, "YubiKey was disconnected while dialog was still displayed.");
-                                    }
+                                //Remove the YKPiv security provider if it was added.
+                                if (Security.getProvider(YKPIV) != null) {
+                                    //Remove provider.
+                                    Security.removeProvider(YKPIV);
+                                    //Telemet and log that PivProvider is being removed from Security static list.
+                                    Telemetry.emit((BaseEvent) new BaseEvent().put(
```
rpdome, line 173: Here, we probably want to emit an even that
- If the provider is set, then we emit "true"
- If the provider is NOT there, then we emit "false".

The idea is to see if the app hosting MSAL is messing with YKPIV when the webview session is happening.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,41 +440,116 @@ public void run() {
...
+        final String methodTag = TAG + "useSmartcardCertForAuth:";
+        //Need to add a PivProvider instance to the beginning of the array of Security providers in order for signature logic to occur.
+        //Note that this provider is removed when the UsbYubiKeyDevice connection is closed.
+        //First check if a PivProvider instance is already present in the static list.
+        if (Security.getProvider(YKPIV) != null) {
+            //Remove existing PivProvider.
+            Security.removeProvider(YKPIV);
+            //Telemet and log this, as the PivProvider instance is either unexpectedly being added elsewhere
+            // or it isn't being removed properly upon CBA flow termination.
+            Telemetry.emit((BaseEvent) new BaseEvent().put(
```
rpdome, line 483: I would suggest emitting "false" also when the provider is not there - so that we can calculate the percentage.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,41 +440,116 @@ public void run() {
...
+            }
+            //PivPrivateKey implements PrivateKey
+            final PivPrivateKey pivPrivateKey = (PivPrivateKey) key;
+            //Cert chain only needs the cert to be used for authentication.
+            final X509Certificate[] chain = new X509Certificate[]{cert};
+            //Clear current dialog.
+            mDialogHolder.dismissDialog();
+            //Call proceed on ClientCertRequest with PivPrivateKey and cert chain.
+            request.proceed(pivPrivateKey, chain);
+        } catch (UnrecoverableKeyException | CertificateException | KeyStoreException | IOException | NoSuchAlgorithmException e) {
```
rpdome: nit: final

rpdome (approved):
We should also emit results of the operation (success, failure + causes).

This can be done in a different PR.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Renaming to YUBIKEY_PROVIDER

common/build.gradle
```diff
@@ -28,7 +28,7 @@ codeCoverageReport {
     coverage.enabled = enableCodeCoverage
 }
 
-def common4jVersion = "0.0.+"
+def common4jVersion = "1.0.8-CBATest"
```
shahzaibj, line 31: I believe this is for your feature branch only

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
shahzaibj: YKPIV == YubiKitProvider? just want to make sure

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -138,21 +155,41 @@ public void invoke(@NonNull UsbYubiKeyDevice device) {
...
-                                        Logger.verbose(TAG, "YubiKey was disconnected while dialog was still displayed.");
-                                    }
+                                //Remove the YKPiv security provider if it was added.
+                                if (Security.getProvider(YUBIKEY_PROVIDER) != null) {
+                                    //Remove provider.
+                                    Security.removeProvider(YUBIKEY_PROVIDER);
+                                    //Telemet and log that PivProvider is being removed from Security static list.
+                                    Telemetry.emit((BaseEvent) new BaseEvent().put(
+                                            TelemetryEventStrings.Key.PIVPROVIDER_REMOVED,
+                                            String.valueOf(true)
```
shahzaibj, line 175: We can probably expose a method that allows to emit booleans directly...we don't have to do this right now

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -496,41 +447,119 @@ public void run() {
...
         }
+        //The position parameter is 1-based (1 maps to index 0).
+        Security.insertProviderAt(new PivProvider(getPivProviderCallback()), 1);
+        //Log that PivProvider is being added to Security static list.
+        Logger.info(methodTag, "An instance of PivProvider was added to Security static list.");
+        try {
+            //Using KeyStore methods in order to generate PivPrivateKey.
+            //Loading null is needed for initialization.
+            final KeyStore keyStore = KeyStore.getInstance(YUBIKEY_PROVIDER, new PivProvider(piv));
+            keyStore.load(null);
```
shahzaibj, line 504: What if it is already initialized?

shahzaibj (approved):
I suggest running code formatter on all classes in this PR.

common/build.gradle
melissaahn: yep this was for a private preview build. I'll change this along with the versions.properties variables back when my feature branch is ready to be merged into dev

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Not sure if this completely answers the question:
https://docs.oracle.com/javase/7/docs/api/java/security/KeyStore.html#load(java.security.KeyStore.LoadStoreParameter)
"Note that if this KeyStore has already been loaded, it is reinitialized and loaded again from the given parameter."

PR #1767, opened 2022-06-20
Remove lab api classes from common4j

rpdome:
## Issue:
MicrosoftClientAssertion is using SHA1, and since it's shipped with common4j, library customers are now prompted by play store (that they're using insecure algorithm).

## Fix:
Remove MicrosoftClientAssertion and associated classes from common4j. Use the one in LabApiUtilities instead. (This means testutils will now dependent on LabApiUtilities)


shahzaibj:
Looks like tests in MSAL are broken because of this.

> Unable to get access token for automation:clientId is marked non-null but is null

rpdome:
> Looks like tests in MSAL are broken because of this.
> 
> > Unable to get access token for automation:clientId is marked non-null but is null

Hmm... I thought every tests were already executed locally.... seems like i missed something. thanks.

It should work now.

shahzaibj (approved):
Thanks Dome!

melissaahn (approved):
Makes sense, thanks Dome

PR #1757, opened 2022-06-13
Add a telemetry relay client.

gitaumoses4:
## What does this PR do?

We need to initialize telemetry on the broker, therefore in this PR we are setting up the infrastructure to do so. The following additions and reactors are necessary:

- Adds support for telemetry relay - this can be implemented to relay telemetry events to a database
- Moves event classes to common4j - some base events are needed in both android and java/linux.
- Ensures there are no duplicates for telemetry observers - to prevent a single instance of a telemetry observer from being invoked multiple times.
- Add client-side filters for telemetry - we are setting up filters to the telemetry relay client to allow client side filtering of telemetry events.

## Relevant PRs.

1. https://github.com/AzureAD/ad-accounts-for-android/pull/1903 - Adds aria relay client to the broker

PR #1768, opened 2022-06-21
Add Error Event and extra events to broker telemetry

gitaumoses4:
## What
We need to capture error events in telemetry when an exception is thrown. 

## How
In order to capture an error event, we shall emit an `ErrorEvent` to telemetry every time an instance of `BaseException` is created.

We then capture the exception information in the error event class. 

#### Error Tag
An error-tag is a unique string that allows us to identify exceptions that were thrown from the same location. This helps us keep track of the number of hits of an exception that is thrown. 

> NOTE: We shall need to implement a different tagging method to enhance debugging.

#### BrokerTelemetryAdapter
This extends the `TelemetryAggregationAdapter` which basically does a merge of all the events with the same correlation id into a single event. If we just make use of this adapter, we will lose important error information since all the error fields will be merged into the aggregated event. 

The `BrokerTelemetryAdapter` sends out error events captured in the flow as independent events. It includes a hit count of the error during the flow.

## Related PRs.
1. Common infrastructure to integrate telemetry to the broker: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1757
2. Broker infrastructure for telemetry: https://github.com/AzureAD/ad-accounts-for-android/pull/1903

## Design Documents
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Telemetry&version=GBgit/android-broker-telemetry

### DIFF
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -11,11 +11,11 @@
         <!-- Because Android detects plugged in YubiKeys as keyboard config changes, configChanges must have `keyboardHidden` and `keyboard` -->
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
 
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.CurrentTaskAuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
 
         <!-- Activity to invoke an interactive request to the intent passed by ad-accounts(Broker) -->
         <activity
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -11,11 +11,11 @@
         <!-- Because Android detects plugged in YubiKeys as keyboard config changes, configChanges must have `keyboardHidden` and `keyboard` -->
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
 
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.CurrentTaskAuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
 
         <!-- Activity to invoke an interactive request to the intent passed by ad-accounts(Broker) -->
         <activity
--- a/common/src/main/AndroidManifest.xml
+++ b/common/src/main/AndroidManifest.xml
@@ -1,26 +1,23 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
     package="com.microsoft.identity.common">
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <!-- Min Sdk version for YubiKit is 19, so override is needed to avoid build errors with Common's min sdk version -->
-    <uses-sdk tools:overrideLibrary="com.yubico.yubikit.android" />
 
     <application>
         <!-- Because Android detects plugged in YubiKeys as keyboard config changes, configChanges must have `keyboardHidden` and `keyboard` -->
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
 
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.CurrentTaskAuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
 
         <!-- Activity to invoke an interactive request to the intent passed by ad-accounts(Broker) -->
         <activity
             android:name="com.microsoft.identity.common.internal.broker.BrokerActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|screenSize|screenLayout" />
     </application>
 
     <!-- Required for API Level 30 to make sure we can detect browsers and other apps we want to
--- a/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.internal.telemetry;
 
 import android.content.Context;
+import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.Build;
@@ -39,7 +40,7 @@ import lombok.NonNull;
 /**
  * TelemetryContext for Android.
  * Containing Android Metadata. It also persists data in SharedPreferences.
- * */
+ */
 public class AndroidTelemetryContext extends AbstractTelemetryContext {
 
     private static final String TAG = AndroidTelemetryContext.class.getName();
@@ -60,9 +61,15 @@ public class AndroidTelemetryContext extends AbstractTelemetryContext {
             final PackageInfo packageInfo = packageManager.getPackageInfo(context.getPackageName(), 0);
             long versionCode = PackageInfoCompat.getLongVersionCode(packageInfo);
 
-            super.addApplicationInfo(packageInfo.applicationInfo.packageName,
+            final ApplicationInfo applicationInfo = context.getApplicationInfo();
+            final String applicationName = applicationInfo.labelRes == 0 ? applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
+
+            super.addApplicationInfo(
+                    applicationInfo.packageName,
+                    applicationName,
                     packageInfo.versionName,
-                    String.valueOf(versionCode));
+                    String.valueOf(versionCode)
+            );
         } catch (final PackageManager.NameNotFoundException e) {
             //Not throw the exception to break the auth request when getting the app's telemetry
             Logger.warn(methodTag, "Unable to find the app's package name from PackageManager.");
--- a/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
@@ -59,10 +59,12 @@ public class AndroidTelemetryContext extends AbstractTelemetryContext {
         try {
             final PackageManager packageManager = context.getPackageManager();
             final PackageInfo packageInfo = packageManager.getPackageInfo(context.getPackageName(), 0);
-            long versionCode = PackageInfoCompat.getLongVersionCode(packageInfo);
+            final long versionCode = PackageInfoCompat.getLongVersionCode(packageInfo);
 
             final ApplicationInfo applicationInfo = context.getApplicationInfo();
-            final String applicationName = applicationInfo.labelRes == 0 ? applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
+            // we should capture the application name as well.
+            final String applicationName = applicationInfo.labelRes == 0 ?
+                    applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
 
             super.addApplicationInfo(
                     applicationInfo.packageName,
--- a/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
@@ -63,11 +63,17 @@ public class AndroidTelemetryContext extends AbstractTelemetryContext {
 
             final ApplicationInfo applicationInfo = context.getApplicationInfo();
             // we should capture the application name as well.
-            final String applicationName = applicationInfo.labelRes == 0 ?
-                    applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
+            String applicationName = "";
+            String packageName = "";
+
+            if (applicationInfo != null) {
+                applicationName = applicationInfo.labelRes == 0 && applicationInfo.nonLocalizedLabel != null ?
+                        applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
+                packageName = applicationInfo.packageName;
+            }
 
             super.addApplicationInfo(
-                    applicationInfo.packageName,
+                    packageName,
                     applicationName,
                     packageInfo.versionName,
                     String.valueOf(versionCode)
--- a/common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
@@ -238,32 +238,6 @@ public class Telemetry {
     }
 
     /**
-     * Flush telemetry error events. These do not require a correlation id, so we pick all events that have their
-     * type as TelemetryEventStrings.EventType.ERROR_EVENT
-     */
-    public void flushErrors() {
-        if (!mIsTelemetryEnabled) {
-            return;
-        }
-
-        if (!mDefaultConfiguration.isDebugEnabled() && mIsDebugging) {
-            return;
-        }
-
-        final List<Map<String, String>> finalRawMap = new CopyOnWriteArrayList<>();
-
-        for (Iterator<Map<String, String>> iterator = mTelemetryRawDataMap.iterator(); iterator.hasNext(); ) {
-            Map<String, String> event = iterator.next();
-            if (TelemetryEventStrings.EventType.ERROR_EVENT.equalsIgnoreCase(event.get(Key.EVENT_TYPE))) {
-                finalRawMap.add(applyPiiOiiRule(event));
-                iterator.remove();
-            }
-        }
-
-        processRawMap(finalRawMap);
-    }
-
-    /**
      * Flush the telemetry data based on the correlation id to the observers.
      *
      * @param correlationId The correlation id should either passed by the client app through the API call
--- a/common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
@@ -238,6 +238,36 @@ public class Telemetry {
     }
 
     /**
+     * Flush telemetry error events. These do not require a correlation id, so we pick all events that have their
+     * type as TelemetryEventStrings.EventType.ERROR_EVENT
+     */
+    public void flushErrors() {
+        if (!mIsTelemetryEnabled) {
+            return;
+        }
+
+        if (!mDefaultConfiguration.isDebugEnabled() && mIsDebugging) {
+            return;
+        }
+
+        final List<Map<String, String>> finalRawMap = new CopyOnWriteArrayList<>();
+
+        for (Iterator<Map<String, String>> iterator = mTelemetryRawDataMap.iterator(); iterator.hasNext(); ) {
+            Map<String, String> event = iterator.next();
+            if (TelemetryEventStrings.EventType.ERROR_EVENT.equalsIgnoreCase(event.get(Key.EVENT_TYPE))) {
+                finalRawMap.add(applyPiiOiiRule(event));
+                // Remove errors that do not have a correlation id.
+                // Those that have, we will flush them along events with the same correlation id
+                if (StringUtil.isNullOrEmpty(event.get(Key.CORRELATION_ID))) {
+                    iterator.remove();
+                }
+            }
+        }
+
+        processRawMap(finalRawMap);
+    }
+
+    /**
      * Flush the telemetry data based on the correlation id to the observers.
      *
      * @param correlationId The correlation id should either passed by the client app through the API call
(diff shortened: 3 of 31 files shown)


common/src/main/AndroidManifest.xml
```diff
@@ -11,11 +11,11 @@
         <!-- Because Android detects plugged in YubiKeys as keyboard config changes, configChanges must have `keyboardHidden` and `keyboard` -->
         <activity
             android:name="com.microsoft.identity.common.internal.providers.oauth2.AuthorizationActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout" />
```
gitaumoses4, line 14:
```suggestion
            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard" />
```

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -60,9 +61,15 @@ private void addApplicationInfo(@NonNull final Context context) {
             final PackageInfo packageInfo = packageManager.getPackageInfo(context.getPackageName(), 0);
             long versionCode = PackageInfoCompat.getLongVersionCode(packageInfo);
 
-            super.addApplicationInfo(packageInfo.applicationInfo.packageName,
+            final ApplicationInfo applicationInfo = context.getApplicationInfo();
+            final String applicationName = applicationInfo.labelRes == 0 ? applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
```
shahzaibj: Can you add some comments here? 

I believe for our purposes this is the name of the broker hosting application?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -60,9 +61,15 @@ private void addApplicationInfo(@NonNull final Context context) {
...
+
+            super.addApplicationInfo(
+                    applicationInfo.packageName,
+                    applicationName,
```
shahzaibj: What's the different between these two?

common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
```diff
@@ -258,16 +258,25 @@ public void flush(@NonNull final String correlationId) {
...
 
-        List<Map<String, String>> finalRawMap = new CopyOnWriteArrayList<>();
+        final List<Map<String, String>> finalRawMap = new CopyOnWriteArrayList<>();
 
         for (Iterator<Map<String, String>> iterator = mTelemetryRawDataMap.iterator(); iterator.hasNext(); ) {
             Map<String, String> event = iterator.next();
-            if (correlationId.equalsIgnoreCase(event.get(Key.CORRELATION_ID))) {
+            // flush error events too
+            if (correlationId.equalsIgnoreCase(event.get(Key.CORRELATION_ID))
+                    || TelemetryEventStrings.EventType.ERROR_EVENT.equals(event.get(Key.EVENT_TYPE))) {
```
shahzaibj: Why we need this type of explicit check for ErrorEvent? We never seemed to need it for other events

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+
+        // count the number of hits of the errors
+        final List<Map<String, String>> countedErrors = countErrors(errorEvents);
+        // publish the event
+        for (Map<String, String> errorEvent : countedErrors) {
+            getObserver().onReceived(errorEvent);
+        }
+
+        // send out the aggregated event separately. This contains the tag of the error event that was
+        // fired last, hence in case of a failure, we know which error caused it.
```
shahzaibj, line 64: So we only know tag for one error event?

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+ * and add combine it with the aggregated event. This means that the event sent by {@link TelemetryAggregationAdapter} only contains
+ * the last error that was dispatched.
  */
 public class BrokerTelemetryAdapter extends TelemetryAggregationAdapter {
     public BrokerTelemetryAdapter(@NonNull IBrokerTelemetryObserver observer) {
         super(observer);
     }
+
+    @Override
+    public void process(@NonNull List<Map<String, String>> rawData) {
```
shahzaibj, line 50: Can you check where (on which thread) does all the processing happen? I'm thinking this probably doesn't need to happen on the same thread that was doing the AT / ATS request and this processing can be done on a separate thread. We don't have to solve the problem right now but something to look at to come up with a plan to improve this going forward.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+        final Map<String, String> aggregatedMap = aggregateEvent(rawData);
+
+        // Filter out the error events first
+        final List<Map<String, String>> errorEvents = filterErrorEvents(rawData, aggregatedMap);
+
+        // count the number of hits of the errors
+        final List<Map<String, String>> countedErrors = countErrors(errorEvents);
+        // publish the event
+        for (Map<String, String> errorEvent : countedErrors) {
```
shahzaibj:
```suggestion
        for (final Map<String, String> errorEvent : countedErrors) {
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
     }
+
+    @Override
+    public void process(@NonNull List<Map<String, String>> rawData) {
+        final Map<String, String> aggregatedMap = aggregateEvent(rawData);
+
+        // Filter out the error events first
+        final List<Map<String, String>> errorEvents = filterErrorEvents(rawData, aggregatedMap);
+
+        // count the number of hits of the errors
```
shahzaibj: this is really collecting than just counting

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+            getObserver().onReceived(errorEvent);
+        }
+
+        // send out the aggregated event separately. This contains the tag of the error event that was
+        // fired last, hence in case of a failure, we know which error caused it.
+        getObserver().onReceived(aggregatedMap);
+    }
+
+    /**
+     * Filters out error events from the list of events
```
shahzaibj:
```suggestion
     * Filters out error events from the list of events.
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+     */
+    private List<Map<String, String>> filterErrorEvents(List<Map<String, String>> rawData, Map<String, String> aggregatedMap) {
```
shahzaibj: NonNull and final to params

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+        final List<Map<String, String>> errorEvents = new ArrayList<>();
+
+        for (Map<String, String> event : rawData) {
```
shahzaibj:
```suggestion
        for (final Map<String, String> event : rawData) {
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
     public BrokerTelemetryAdapter(@NonNull IBrokerTelemetryObserver observer) {
         super(observer);
     }
+
+    @Override
+    public void process(@NonNull List<Map<String, String>> rawData) {
+        final Map<String, String> aggregatedMap = aggregateEvent(rawData);
+
+        // Filter out the error events first
+        final List<Map<String, String>> errorEvents = filterErrorEvents(rawData, aggregatedMap);
```
shahzaibj, line 54: I don't fully get why we need to do this and why can't we just flush everything altogether

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,93 @@
...
+
+                errorEvents.add(errorEvent);
+            }
+        }
+        return errorEvents;
+    }
+
+    /**
+     * Errors with the same tag are essentially the same error, we don't need to send them to telemetry as different errors. Therefore
+     * we shall add an {@link TelemetryEventStrings.Key#ERROR_COUNT} key that denotes the number of hits of the error.
```
shahzaibj, line 88: Ok, I think I see now what we're trying to do here...

Can you add javadoc to the other method to link here?

shahzaibj: I'm actually curious, when do we anticipate ourselves to hold onto multiple ErrorEvents with same error code and tag in memory and not having flushed them already?

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/TelemetryAggregationAdapter.java
```diff
@@ -76,10 +77,11 @@ protected Map<String, String> aggregateEvent(@NonNull final List<Map<String, Str
                 );
             }
 
-            if (!StringUtil.isNullOrEmpty(event.get(Key.IS_SUCCESSFUL))) {
+            if (!StringUtil.isNullOrEmpty(event.get(Key.IS_SUCCESSFUL)) || eventName.contains(END)) {
```
shahzaibj: END?

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ApiStartEvent.java
```diff
@@ -152,11 +157,25 @@ public ApiStartEvent putProperties(
...
         if (parameters instanceof BrokerInteractiveTokenCommandParameters) {
-            //TODO when integrate the telemetry with broker.
+            final BrokerInteractiveTokenCommandParameters interactiveParameters = (BrokerInteractiveTokenCommandParameters) parameters;
+
+            put(Key.BROKER_VERSION, interactiveParameters.getBrokerVersion());
+            put(Key.BROKER_PROTOCOL_VERSION, interactiveParameters.getNegotiatedBrokerProtocolVersion());
+
+            put(Key.CALLER_APP_VERSION, interactiveParameters.getCallerAppVersion());
+            put(Key.CALLER_APP_PACKAGE_NAME, interactiveParameters.getCallerPackageName());
+            put(Key.CALLER_APP_UUID, String.valueOf(interactiveParameters.getCallerUid()));
```
shahzaibj: I think we discussed in the meeting that we want to avoid sending these for now

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ApiStartEvent.java
```diff
@@ -152,11 +157,25 @@ public ApiStartEvent putProperties(
         }
 
         if (parameters instanceof BrokerInteractiveTokenCommandParameters) {
-            //TODO when integrate the telemetry with broker.
+            final BrokerInteractiveTokenCommandParameters interactiveParameters = (BrokerInteractiveTokenCommandParameters) parameters;
+
+            put(Key.BROKER_VERSION, interactiveParameters.getBrokerVersion());
```
shahzaibj: Can you also check that broker version is populated correctly on the command parameters? So this capturing is not in vain.

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ApiStartEvent.java
```diff
@@ -182,6 +201,11 @@ public ApiStartEvent putValidationStatus(@NonNull final String validationStatus)
         return this;
     }
 
+    public ApiStartEvent putWorkPlaceJoined(final boolean isWorkPlaceJoined) {
```
shahzaibj, line 209: This is fine for now, but ultimately we want some of these things to probably move out of API Event

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/BaseEvent.java
```diff
@@ -37,7 +39,10 @@ public BaseEvent() {
 
     @Override
     public Properties put(String key, String value) {
-        return super.put(key, value);
+        if(!StringUtil.isNullOrEmpty(value)) {
```
shahzaibj, line 42: What if the key itself is null?

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/BrokerEvent.java
```diff
@@ -0,0 +1,47 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.telemetry.events;
+
+import com.microsoft.identity.common.java.telemetry.TelemetryEventStrings;
+
+import lombok.NonNull;
+
+/**
+ * A generic broker event. This can be used to emit any kind of event in the broker. e.g. an event with the IPC strategy
```
shahzaibj:
```suggestion
 * A generic broker event. This can be used to emit any kind of event in the broker. e.g. an event with the IPC strategy.
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/BrokerEvent.java
```diff
@@ -0,0 +1,47 @@
...
+        super();
+        types(TelemetryEventStrings.EventType.BROKER_EVENT);
+        names(eventName);
+    }
+
+    public BrokerEvent() {
+        this(null);
+    }
+
+    public BrokerEvent putIPCStrategy(@NonNull String ipcStrategy) {
```
shahzaibj, line 43: So this is more for MSAL telemetry than broker telemetry

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ErrorEvent.java
```diff
@@ -0,0 +1,77 @@
+package com.microsoft.identity.common.java.telemetry.events;
```
shahzaibj, line 23: LICENSE

common4j/src/main/com/microsoft/identity/common/java/telemetry/rules/TelemetryPiiOiiRules.java
```diff
@@ -56,7 +56,10 @@ final public class TelemetryPiiOiiRules {
             Key.REDIRECT_URI,
             Key.HTTP_PATH,
             Key.AUTHORITY,
-            Key.IDP_NAME
+            Key.IDP_NAME,
+            Key.CALLER_APP_PACKAGE_NAME,
+            Key.CALLER_APP_UUID,
+            Key.CALLER_APP_VERSION
```
shahzaibj, line 62: Ok, I see you flagged them as PII. This should work.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
gitaumoses4: The first one gives the package name. e.g. `com.microsoft.authenticator` the second one gives the application label e.g. `Authenticator`

common4j/src/main/com/microsoft/identity/common/java/telemetry/Telemetry.java
gitaumoses4: Not sure why I added this. Lemme remove it.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
gitaumoses4: Yes that's right. But in line 59
```java
        for (Map<String, String> errorEvent : countedErrors) {
            getObserver().onReceived(errorEvent);
        }
```

We emit error events separately, hence we can query all errors for a particular event by correlation id.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/TelemetryAggregationAdapter.java
gitaumoses4: Need to remove that.

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ApiStartEvent.java
gitaumoses4: We actually don't need this line. I forced every event to contain the `Broker version` by putting the value when creating the TelemetryContext.

```java
        final AndroidTelemetryContext telemetryContext = new AndroidTelemetryContext(mContext);
        telemetryContext.put(TelemetryEventStrings.Key.BROKER_VERSION, WorkplaceJoin.API_VERSION);
```

The telemetry context is merged with every event that is sent.

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/BaseEvent.java
gitaumoses4: Well, the `BaseEvent` extends the `Properties` object which will just ignore if the key is null. I should probably mark this as @NonNull

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/BrokerEvent.java
gitaumoses4: It could be both. We can observe how MSAL interfaces with broker, in terms of which IPC strategy is mostly used.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
gitaumoses4: The event sent to aria is a key-value pair, but there may be multiple errors that were reported in the flow. These errors have the following keys `tag`, `exception class` , `error code` etc.

This means that if we combine all the errors into a single event, we can only store one error value for `tag`, `exception class`, `error code` etc. Meaning we would have lost important error metrics.

Therefore, the main event should be flushed out with the error keys `tag`, `exception class`, `error code` having values for the last error event that was fired in the flow. 

Then we send out all the error events that were captured in the same flow as single events. (Note that they all have the same correlation id), so we can quickly perform a `join` on the database side to see all errors emitted for a particular flow.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
gitaumoses4: This could happen from utility functions. E.g. we have a function that checks for internet connectivity. 
We could check for internet connectivity multiple times in the acquire token flow. 

If an exception occurs inside this method, it means we shall have an error emitted for every single time this method is invoked.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
shahzaibj: Interestingly. I wasn't aware that we were throwing multiple errors in one flow. Doesn't throwing one error just halt the flow?

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
shahzaibj: Hmm..I'm not sure how that happens. I mean the first time network check fails, the flow is supposed to halt. How do we keep throwing the network check failed error multiple times in a flow?

gitaumoses4: Well, the network one was just an example. I probably should have said an exception that doesn't necessarily lead to a direct failure.

common4j/src/main/com/microsoft/identity/common/java/telemetry/TelemetryEventStrings.java
```diff
@@ -28,6 +28,7 @@ public class TelemetryEventStrings {
     public static final class App {
         public static final String BUILD = EVENT_PREFIX + "application_build";
```
iamgusain, line 29: Let's remove this EVENT_PREFIX from all properties, if it is not must have

shahzaibj: It would definitely need some checking in the code if anything is relying on these currently. This must only be done if can be safely done. If we cannot guarantee safety immediately, I suggest we do this in the next iteration.

gitaumoses4:
@shahzaibj @iamgusain I've addressed all the comments.

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,92 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import lombok.NonNull;
 
 /**
  * Adapter to aggregate events for broker telemetry. We make use of the {@link TelemetryAggregationAdapter} which
```
iamgusain, line 37: Are we using the AggregationAdapter for the current broker telemetry work?
Or the Default one, which doesnt aggregate?

iamgusain (approved):

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
shahzaibj: Okay, that makes sense

common4j/src/main/com/microsoft/identity/common/java/telemetry/adapter/BrokerTelemetryAdapter.java
```diff
@@ -21,16 +21,92 @@
...
  * aggregates the events into MATS format.
+ * <p>
+ * We then send out the error events as single events since the {@link TelemetryAggregationAdapter} will pick the last error event
+ * and add combine it with the aggregated event. This means that the event sent by {@link TelemetryAggregationAdapter} only contains
+ * the last error that was dispatched.
  */
 public class BrokerTelemetryAdapter extends TelemetryAggregationAdapter {
```
shahzaibj, line 44: Let's add some unit tests for the methods in this class

shahzaibj (approved):

PR #1771, opened 2022-06-22
Update msal-to-broker protocol in GET_ACCOUNTS to V8 

p3dr0rv:
Related PBI: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1923544
Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1914

* Bumped Max MSAL Broker Protocol version to 8.0.
* GET_ACCOUNTS endpoint requires minimum_required_broker_protocol_version of 8.0+ to return an account constructed from PRT id token to FOCI apps.
* Add helper method to validate if negotiated msal-broker protocol allows FOCI apps to construct accounts from PRT Id Token.

(FOCI) Family of Client IDs

Update [Android Auth Broker Protocol Versions](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Broker%20API/broker_protocol_versions.md&version=GBdev&_a=preview)


common/src/main/java/com/microsoft/identity/common/internal/util/BrokerProtocolVersionUtil.java
```diff
@@ -42,4 +43,15 @@ public static boolean canCompressBrokerPayloads(@Nullable String negotiatedBroke
                 MSAL_TO_BROKER_PROTOCOL_COMPRESSION_CHANGES_MINIMUM_VERSION);
 
     }
+
+    public static boolean canFociAppsConstructAccountsFromPrtIdTokens(@Nullable String negotiatedBrokerProtocol) {
```
iamgusain: Can we add a test for this.

iamgusain (approved):
Approved with suggestion

common/src/main/java/com/microsoft/identity/common/internal/util/BrokerProtocolVersionUtil.java
shahzaibj: javadoc

shahzaibj: This method seems to be doing almost same thing as `canCompressBrokerPayloads`...should some of this code move to a helper method?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1401,7 +1402,7 @@ public enum API {
             MSAL_HELLO(MSAL_HELLO_PATH, null, VERSION_3),
             ACQUIRE_TOKEN_INTERACTIVE(MSAL_ACQUIRE_TOKEN_INTERACTIVE_PATH, null, VERSION_3),
             ACQUIRE_TOKEN_SILENT(MSAL_ACQUIRE_TOKEN_SILENT_PATH, null, VERSION_3),
-            GET_ACCOUNTS(MSAL_GET_ACCOUNTS_PATH, null, VERSION_3),
+            GET_ACCOUNTS(MSAL_GET_ACCOUNTS_PATH, null, VERSION_8),
```
shahzaibj: So this is minimum required broker protocol version? 

After this change the only way to get accounts would be to upgrade MSAL to a version that supports the new version?

rpdome: This shouldn't be 8. 

Only apps that requires WPJ account from PRT needs to set to 8.

p3dr0rv: From my understanding, if we want to increment the msal-broker protocol version in broker we need to increment the version here, because the version is calculated from the max number in the BrokerContentProvider.API
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/b8ea54b8ea2e37f7cc40939752a6dd582c6ae21c/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java#L599
hence the change here in the GET_ACCOUNTS API.

still old msal versions will work, see:
https://github.com/AzureAD/ad-accounts-for-android/blob/43aa86750ea846015e4c73a50f05a08cdbe5296a/AADAuthenticator/src/msal/java/com/microsoft/identity/client/BrokerUtils.java#L96
the change will only affect if the msal msal-broker protocol version is bigger that the msal-broker protocol version in broker, and the client sets a minimum_required_broker_protocol_version for their aplication to work properly.

common/src/main/java/com/microsoft/identity/common/internal/util/BrokerProtocolVersionUtil.java
p3dr0rv: added common4j/src/test/com/microsoft/identity/common/java/util/BrokerProtocolVersionUtilTest.java

p3dr0rv: added here cb84a73eee5e8907dc89588ee566623e99383865

p3dr0rv: done cb84a73eee5e8907dc89588ee566623e99383865

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
p3dr0rv: After an offline chat with @rpdome I have decided to remove the computeMaxMsalBrokerProtocol method.
This method only works properly when a new endpoint is introduced to the API, for changes in exiting endpoints it is confusing to update the version number, due this number reflects the min required msal-broker protocol version number to support a new endpoint.

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -478,4 +478,57 @@ public static void throwIfArgumentIsNullOrEmpty(final @Nullable String argument,
             throw new NullPointerException(argumentName + " is null or empty.");
         }
     }
+
+
+    /**
+     * Returns true if the first semantic version is smaller or equal to the second version.
+     */
+    public static boolean isFirstVersionSmallerOrEqual(@NonNull final String first,
```
rpdome: Could probably just move these 3 into BrokerProtocolVersionUtil.

common4j/src/main/com/microsoft/identity/common/java/util/BrokerProtocolVersionUtil.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String MSAL_TO_BROKER_PROTOCOL_ACCOUNT_FROM_PRT_CHANGES_MINIMUM_VERSION = "8.0";
+
+    /**
+     * Verifies if negotiated broker protocol version allows to decompressing/compressing broker payloads.
+     *
+     * @param negotiatedBrokerProtocol negotiated protocol version, result of hello handshake.
+     * @return true if the negotiated protocol version is larger or equal than
+     * the {@link BrokerProtocolVersionUtil#MSAL_TO_BROKER_PROTOCOL_COMPRESSION_CHANGES_MINIMUM_VERSION}.
+     */
+    public static boolean canCompressBrokerPayloads(@Nullable String negotiatedBrokerProtocol) {
```
rpdome: nit: add final to arguments, here and below.

iamgusain: Can we call this installedBrokerProtocolVersion?

iamgusain (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -676,6 +656,11 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String NEGOTIATED_BP_VERSION_KEY = "common.broker.protocol.version.name";
 
+        /**
+         * The Boolean to send when FOCI apps are allowed to construct accounts from PRT id token in getAccounts.
+         */
+        public static final String FOCI_APPS_CAN_CONSTRUCT_ACCOUNTS_FROM_PRT_ID_TOKEN_KEY = "can.construct.accounts.from.prt.id.token";
```
shahzaibj: This seems like a statement rather than a question...

FOCI_APPS_CAN_CONSTRUCT_ACCOUNTS_FROM_PRT_ID_TOKEN_KEY means foci apps **CAN** construct accounts from Prt ID token

I think what you want is the following:

CAN_FOCI_APPS_CONSTRUCT_ACCOUNTS_FROM_PRT_ID_TOKEN_KEY - this would be a true boolean question

common4j/src/main/com/microsoft/identity/common/java/util/BrokerProtocolVersionUtil.java
p3dr0rv: for this method is the negotiated protocol version, that's decided after the handshake (hello).

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
p3dr0rv: yeah 237d0dda446bb08baf9fc17372b08813eaa8efd9

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/util/BrokerProtocolVersionUtil.java
p3dr0rv: cb46fe6a8049a3c1e0c57f76d83c3e5b01e5cd7d

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
p3dr0rv: cb46fe6a8049a3c1e0c57f76d83c3e5b01e5cd7d

shahzaibj (approved):

PR #1772, opened 2022-06-22
[UI automation] support app update scenarios

somalaya:
**What** : Currently we install the app and perform UI automation testcase run on one particular version of the app. But we do not have testcases that test update scenarios. 
For ex : AcquireToken interactive is called on an app and the next update of the broker app is available. The user updates the app and acquireToken silent is called. To test whether this scenario is successful, I added update support.

**How** : Added update() method to IApp to update with the given installer. By default the app will be updated using the  playstore version if the installer is not explicitly provided.  In each of the broker apps like BrokerMicrosoftAuthentictor, BrokerCompanyPortal, BrokerHost and AzureSampleApp - Added the name of the localApk which will be used for updating the app.

Note : Internally update method added in App.java calls adb install only. But this will still be an app update as adb install -r is the closest you get, that is actually an update as it keeps the database and stored preferences. If you uninstall/re-install both the app database and preferences is deleted. 

**Testing** : Added testcases to test the acquireTokenSilent call after updating the app in https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1640

### DIFF
--- a/uiautomationutilities/build.gradle
+++ b/uiautomationutilities/build.gradle
@@ -42,9 +42,13 @@ android {
 
     def final installSourcePlayStore = "PlayStore"
     def final installSourceLocalApk = "LocalApk"
+    def final updateSourcePlayStore = "PlayStore"
+    def final updateSourceLocalApk = "LocalApk"
 
     // defaults to LocalApkFile - the file must be pushed to the device before tests
     def brokerInstallationSource = installSourcePlayStore
+
+    def brokerUpdationSource = updateSourcePlayStore
 
     def wordAppInstallationSource = installSourcePlayStore
 
@@ -56,6 +60,18 @@ android {
         } else {
             throw new InvalidUserDataException("Unspported broker source provided: "
                     + brokerSource
+                    + " Only the following broker sources are supported: PlayStore , LocalApk");
+        }
+    }
+
+    if (project.hasProperty("brokerUpdateSource")) {
+        if (brokerUpdateSource.equalsIgnoreCase(updateSourcePlayStore)) {
+            brokerUpdationSource = updateSourcePlayStore
+        } else if (brokerUpdateSource.equalsIgnoreCase(updateSourceLocalApk)) {
+            brokerUpdationSource = updateSourceLocalApk
+        } else {
+            throw new InvalidUserDataException("Unspported update broker source provided: "
+                    + brokerUpdateSource
                     + " Only the following broker sources are supported: PlayStore , LocalApk");
         }
     }
@@ -81,10 +97,14 @@ android {
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "BROKER_INSTALL_SOURCE", "\"$brokerInstallationSource\"")
+        buildConfigField("String", "BROKER_UPDATE_SOURCE", "\"$brokerUpdationSource\"")
         buildConfigField("String", "WORD_APP_INSTALL_SOURCE", "\"$wordAppInstallationSource\"")
         buildConfigField("String", "INSTALL_SOURCE_PLAY_STORE", "\"$installSourcePlayStore\"")
         buildConfigField("String", "INSTALL_SOURCE_LOCAL_APK", "\"$installSourceLocalApk\"")
+        buildConfigField("String", "UPDATE_SOURCE_PLAY_STORE", "\"$updateSourcePlayStore\"")
+        buildConfigField("String", "UPDATE_SOURCE_LOCAL_APK", "\"$updateSourceLocalApk\"")
         buildConfigField("boolean", "PREFER_PRE_INSTALLED_APKS", "$usePreInstalledApks")
+
 
         // Specifies a sorted list of flavors that the plugin should try to use from
         // a given dimension. The following tells the plugin that, when encountering
--- a/uiautomationutilities/build.gradle
+++ b/uiautomationutilities/build.gradle
@@ -40,23 +40,21 @@ android {
 
     boolean usePreInstalledApks = project.hasProperty("preferPreInstalledApks")
 
-    def final installSourcePlayStore = "PlayStore"
-    def final installSourceLocalApk = "LocalApk"
-    def final updateSourcePlayStore = "PlayStore"
-    def final updateSourceLocalApk = "LocalApk"
+    def final appSourcePlayStore = "PlayStore"
+    def final appSourceLocalApk = "LocalApk"
 
     // defaults to LocalApkFile - the file must be pushed to the device before tests
-    def brokerInstallationSource = installSourcePlayStore
+    def brokerInstallationSource = appSourcePlayStore
 
-    def brokerUpdationSource = updateSourcePlayStore
+    def brokerUpdationSource = appSourcePlayStore
 
-    def wordAppInstallationSource = installSourcePlayStore
+    def wordAppInstallationSource = appSourcePlayStore
 
     if (project.hasProperty("brokerSource")) {
-        if (brokerSource.equalsIgnoreCase(installSourcePlayStore)) {
-            brokerInstallationSource = installSourcePlayStore
-        } else if (brokerSource.equalsIgnoreCase(installSourceLocalApk)) {
-            brokerInstallationSource = installSourceLocalApk
+        if (brokerSource.equalsIgnoreCase(appSourcePlayStore)) {
+            brokerInstallationSource = appSourcePlayStore
+        } else if (brokerSource.equalsIgnoreCase(appSourceLocalApk)) {
+            brokerInstallationSource = appSourceLocalApk
         } else {
             throw new InvalidUserDataException("Unspported broker source provided: "
                     + brokerSource
@@ -65,10 +63,10 @@ android {
     }
 
     if (project.hasProperty("brokerUpdateSource")) {
-        if (brokerUpdateSource.equalsIgnoreCase(updateSourcePlayStore)) {
-            brokerUpdationSource = updateSourcePlayStore
-        } else if (brokerUpdateSource.equalsIgnoreCase(updateSourceLocalApk)) {
-            brokerUpdationSource = updateSourceLocalApk
+        if (brokerUpdateSource.equalsIgnoreCase(appSourcePlayStore)) {
+            brokerUpdationSource = appSourcePlayStore
+        } else if (brokerUpdateSource.equalsIgnoreCase(appSourceLocalApk)) {
+            brokerUpdationSource = appSourceLocalApk
         } else {
             throw new InvalidUserDataException("Unspported update broker source provided: "
                     + brokerUpdateSource
@@ -77,10 +75,10 @@ android {
     }
 
     if (project.hasProperty("wordAppSource")) {
-        if (wordAppSource.equalsIgnoreCase(installSourcePlayStore)) {
-            wordAppInstallationSource = installSourcePlayStore
-        } else if (wordAppSource.equalsIgnoreCase(installSourceLocalApk)) {
-            wordAppInstallationSource = installSourceLocalApk
+        if (wordAppSource.equalsIgnoreCase(appSourcePlayStore)) {
+            wordAppInstallationSource = appSourcePlayStore
+        } else if (wordAppSource.equalsIgnoreCase(appSourceLocalApk)) {
+            wordAppInstallationSource = appSourceLocalApk
         } else {
             throw new InvalidUserDataException("Unsupported word app source provided: "
                     + wordAppSource
@@ -99,10 +97,10 @@ android {
         buildConfigField("String", "BROKER_INSTALL_SOURCE", "\"$brokerInstallationSource\"")
         buildConfigField("String", "BROKER_UPDATE_SOURCE", "\"$brokerUpdationSource\"")
         buildConfigField("String", "WORD_APP_INSTALL_SOURCE", "\"$wordAppInstallationSource\"")
-        buildConfigField("String", "INSTALL_SOURCE_PLAY_STORE", "\"$installSourcePlayStore\"")
-        buildConfigField("String", "INSTALL_SOURCE_LOCAL_APK", "\"$installSourceLocalApk\"")
-        buildConfigField("String", "UPDATE_SOURCE_PLAY_STORE", "\"$updateSourcePlayStore\"")
-        buildConfigField("String", "UPDATE_SOURCE_LOCAL_APK", "\"$updateSourceLocalApk\"")
+        buildConfigField("String", "INSTALL_SOURCE_PLAY_STORE", "\"$appSourcePlayStore\"")
+        buildConfigField("String", "INSTALL_SOURCE_LOCAL_APK", "\"$appSourceLocalApk\"")
+        buildConfigField("String", "UPDATE_SOURCE_PLAY_STORE", "\"$appSourcePlayStore\"")
+        buildConfigField("String", "UPDATE_SOURCE_LOCAL_APK", "\"$appSourceLocalApk\"")
         buildConfigField("boolean", "PREFER_PRE_INSTALLED_APKS", "$usePreInstalledApks")
 
 
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
@@ -136,18 +136,14 @@ public abstract class App implements IApp {
     }
 
     @Override
-    public void update () {
+    public void update() {
         if (updateAppInstaller instanceof LocalApkInstaller && !TextUtils.isEmpty(localUpdateApkFileName)) {
             Logger.i(TAG, "Installing the " + this.appName + " from local apk..");
-            updateAppInstaller.installApp(localUpdateApkFileName);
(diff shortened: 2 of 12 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -108,6 +127,21 @@ public void launch() {
...
+            Logger.i(TAG, "Installing the " + this.appName + " from local apk..");
+            updateAppInstaller.installApp(localUpdateApkFileName);
+        } else {
+            Logger.i(TAG, "Installing the " + this.appName + " from Play Store..");
+            updateAppInstaller.installApp(packageName);
+        }
+
+        // the app is just installed, first run should be handled
+        // this value can (should) be changed to false by child classes as appropriate
+        shouldHandleFirstRun = true;
```
shahzaibj: Is this actually required for update?

shahzaibj:
You mentioned that `-r` flag is needed to re-install. I don't see that in the PR though. Is our code somewhere already passing the `-r` flag to ADB? Can you point to that code?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
somalaya: I tested the update scenario and this is not required as first run flag is stored in storage in authenticator and this is not reset on update. Hence removing this piece of code

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/AbstractTestBroker.java
```diff
@@ -63,6 +63,9 @@ public abstract class AbstractTestBroker extends App implements ITestBroker {
     public final static IAppInstaller DEFAULT_BROKER_APP_INSTALLER = BuildConfig.INSTALL_SOURCE_LOCAL_APK
             .equalsIgnoreCase(BuildConfig.BROKER_INSTALL_SOURCE)
             ? new LocalApkInstaller() : new PlayStore();
+    public final static IAppInstaller DEFAULT_BROKER_APP_UPDATE_INSTALLER = BuildConfig.INSTALL_SOURCE_LOCAL_APK
```
iamgusain: Should this be managed via a different BuildConfig?
In case we want to first install playstore version and then update using local Apk

iamgusain (approved):
Left minor comment, can be addressed separately

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -76,6 +81,8 @@ public App(@NonNull final String packageName, @NonNull final String appName) {
     public App(@NonNull final String packageName, @NonNull final IAppInstaller appInstaller) {
         this.appInstaller = appInstaller;
         this.packageName = packageName;
+        // update installer is PlayStore by default
+        this.updateAppInstaller = new PlayStore();
```
shahzaibj: I think tis is fine for now, but it would actually make more sense for LocalApk installer to be the default here as usually we would be testing update scenarios with a side loaded APK

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -108,6 +127,17 @@ public void launch() {
         CommonUtils.launchApp(packageName);
     }
 
+    @Override
+    public void update() {
+        if (updateAppInstaller instanceof LocalApkInstaller && !TextUtils.isEmpty(localUpdateApkFileName)) {
+            Logger.i(TAG, "Updating the " + this.appName + " from local apk..");
+            AdbShellUtils.updatePackage(LocalApkInstaller.LOCAL_APK_PATH_PREFIX + localApkFileName, "-t", "-r", "-d");
```
shahzaibj: Why not put this as a method on the `LocalApkInstaller`? I mean create a new method in that file called as `updateApp` - similar to what you've done for PlayStore, and then over here you can probably just invoke `updateAppInstaller.updateApp(identifier)`

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/AbstractTestBroker.java
shahzaibj: +1

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
```diff
@@ -187,4 +187,49 @@ public void installApp(@NonNull final String searchHint) {
         installAppFromMarketPage();
     }
 
+    public void updateApp(@NonNull final String searchHint) {
```
shahzaibj, line 194: Consider my other suggestion and put this as interface method also on the `IAppInstaller`

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/AdbShellUtils.java
```diff
@@ -99,6 +99,28 @@ public static void installPackage(@NonNull final String packageName, @NonNull St
         Assert.assertEquals("Success", result.trim());
     }
 
+    /**
+     * Updates the supplied package on the device with the supplied flags.
+     *
+     * @param packageName the name of the package to update
+     * @param flags       the flags to use during for update of the app
+     */
+    public static void updatePackage(@NonNull final String packageName, @NonNull String... flags) {
```
shahzaibj: Looks like it is doing the same thing as the `installPackage` method. Why do we need this method?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/AbstractTestBroker.java
somalaya: Added a buildConfig for update source as well

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
somalaya: Changed it to use LocalApkInstaller instead

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
somalaya: addressed this in f9ecbb25a67397feee33a63807075d46e1b84038

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
somalaya: addressed in f9ecbb25a67397feee33a63807075d46e1b84038

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/AdbShellUtils.java
somalaya: removed this, I added this method to add few differences but in the end removed them.

uiautomationutilities/build.gradle
```diff
@@ -42,10 +42,14 @@ android {
 
     def final installSourcePlayStore = "PlayStore"
     def final installSourceLocalApk = "LocalApk"
+    def final updateSourcePlayStore = "PlayStore"
+    def final updateSourceLocalApk = "LocalApk"
```
shahzaibj: There is duplication of constants now. I think you can reduce these to just 2 as opposed to 4:

- appSourcePlayStore = "PlayStore"
- appSourceLocalApk = "LocalApk"

uiautomationutilities/build.gradle
```diff
@@ -81,11 +97,15 @@ android {
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         buildConfigField("String", "BROKER_INSTALL_SOURCE", "\"$brokerInstallationSource\"")
+        buildConfigField("String", "BROKER_UPDATE_SOURCE", "\"$brokerUpdationSource\"")
         buildConfigField("String", "WORD_APP_INSTALL_SOURCE", "\"$wordAppInstallationSource\"")
         buildConfigField("String", "INSTALL_SOURCE_PLAY_STORE", "\"$installSourcePlayStore\"")
         buildConfigField("String", "INSTALL_SOURCE_LOCAL_APK", "\"$installSourceLocalApk\"")
+        buildConfigField("String", "UPDATE_SOURCE_PLAY_STORE", "\"$updateSourcePlayStore\"")
+        buildConfigField("String", "UPDATE_SOURCE_LOCAL_APK", "\"$updateSourceLocalApk\"")
```
shahzaibj: See my recommendation above about reducing duplicated constants

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -84,20 +91,43 @@ public App(@NonNull final String packageName,
...
+        else {
+            //TODO: make it build time configurable to specify the installer that should be used.
+            // Ideally we can specify different installers on app basis
+            if (appInstaller instanceof LocalApkInstaller && !TextUtils.isEmpty(localApkFileName)) {
+                Logger.i(TAG, "Installing the " + this.appName + " from local apk..");
+                appInstaller.installApp(localApkFileName);
+            } else {
+                Logger.i(TAG, "Installing the " + this.appName + " from Play Store..");
+                appInstaller.installApp(packageName);
+            }
```
shahzaibj: I think you can consider reducing the number of `if/else` blocks here. As a hint, null check and `TextUtils.isEmpty` check are similar with the latter being a more comprehensive check

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -116,6 +146,19 @@ public void launch() {
         CommonUtils.launchApp(packageName);
     }
 
+    @Override
+    public void update() {
+        String appHint;
+        if (updateAppInstaller instanceof LocalApkInstaller && !TextUtils.isEmpty(localApkFileName)) {
```
shahzaibj: What if the initial install is also done using the same `localApkFileName`?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
```diff
@@ -47,12 +47,21 @@ public class AzureSampleApp extends App {
...
         localApkFileName = AZURE_SAMPLE_APK;
     }
 
+    public AzureSampleApp(final boolean isOldApk) {
+        super(AZURE_SAMPLE_PACKAGE_NAME, AZURE_SAMPLE_APP_NAME, new LocalApkInstaller());
+        localApkFileName = AZURE_SAMPLE_APK;
+        if (isOldApk) {
+            localOldApkFileName = OLD_AZURE_SAMPLE_APK;
+        }
+    }
```
shahzaibj: There is no javadoc and/or comments here and it is hard to tell what this method is achieving

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -77,6 +78,14 @@ public BrokerCompanyPortal(@NonNull final IAppInstaller appInstaller) {
         localApkFileName = COMPANY_PORTAL_APK;
     }
 
+    public BrokerCompanyPortal(final boolean isOldApk) {
```
shahzaibj: I think it is difficult for the readers of this code to figure out what would happen if we pass a true vs false to this constructor. It seems that we've got some boolean here that is governing how to pick the APK file name when doing install.

I think rather than having `oldApkFileName` and dynamically setting it based on a boolean, I think instead we should have clearly defined `localApkFileNameFreshInstall` and `localApkFileNameUpdate` - something like that and then you don't need a booleans in the constructor to figure these out

shahzaibj: The above feedback to all the places these patterns are used in this PR

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
somalaya: This is not possible in update scenarios. I have pushed changes in this (https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1640/files) PR to make sure that in case of update scenarios - We extend AbstractMsalBrokerUpdateTest in all update scenarios to first install old apk and then install the latest one when update function is called

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
somalaya: The reason I did not add localApkFileNameFreshInstall and localApkFileNameUpdate is because when we say update - we will actually be installing the latest apk (one that is picked from the latest pipeline) and so, when we run update scenarios, we first install an oldApk based on the testcase that is running. 

I could have kept static files like you mentioned. But in that case, I would end up installing localApkFileNameFreshInstall apks everytime for the first install - here it would become tricky what a fresh install means as we cannot differentiate between freshInstall in normal testcase and freshInstall in update testcase (in this case freshInstall means oldApk install)

To fix the ambiguity, I can add some comments to make it clear

shahzaibj: I think I kind of get what you're saying, but still not fully clear. Comments in the code would definitely help. Let's also sync on this offline next week so I can get a better understanding of the flow here.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
shahzaibj, line 194: Is this an overriden method? If yes, then let's add override annotation

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
```diff
@@ -187,4 +187,49 @@ public void installApp(@NonNull final String searchHint) {
...
+            updateAppFromMarketPageInternal();
+        } catch (final UiObjectNotFoundException e) {
+            acceptGooglePlayTermsOfService();
+            try {
+                updateAppFromMarketPageInternal();
+            } catch (UiObjectNotFoundException ex) {
+                throw new AssertionError(e);
+            }
+        }
+    }
```
shahzaibj: Looks like in this entire block, the only thing different from `install` is `line 204` i.e. `updateAppFromMarketPageInternal`. Everything else is same as `install` right? I think there is definitely opportunity for code sharing.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/App.java
```diff
@@ -116,6 +134,19 @@ public void launch() {
...
+    public void update() {
+        String appHint;
+        if (updateAppInstaller instanceof LocalApkInstaller && !TextUtils.isEmpty(localUpdateApkFileName)) {
+            Logger.i(TAG, "Updating the " + this.appName + " from local apk..");
+            appHint = localUpdateApkFileName;
+        } else {
+            Logger.i(TAG, "Updating the " + this.appName + " from Play Store..");
+            appHint = packageName;
+        }
+        updateAppInstaller.updateApp(appHint);
```
shahzaibj, line 147:
```suggestion
            updateAppInstaller.updateApp(localUpdateApkFileName);
        } else {
            Logger.i(TAG, "Updating the " + this.appName + " from Play Store..");
            updateAppInstaller.updateApp(packageName);
        }
```

shahzaibj (approved):
I left minor comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
somalaya: Done

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
somalaya: Line 204 and 208. Rest is just calling updateAppFromMarketPageInternal and catching exception.

shahzaibj: Yes, so I think definitely opportunity for code sharing here

somalaya:
> I left minor comments

addressed all comments. Please refer to this commit for playstore code duplication removal changes 6d285f43f549efdfff4579740bee706abf8b399f

shahzaibj (approved):

PR #1762, opened 2022-06-14
Add additional logging/exception to identify root cause for office union app crash

iamgusain:
The office mobile union app is noticing a crash in the latest version (after upgrading to adal 4.1.0, common 4.2.0)
Adding additional logging/exception to figure out the root cause of the crash.

Stack Trace from the crash: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/990c353bb7e4cc3896857e7fb76e217f0ad34d43/common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java#L203

java.lang.IllegalStateException: 
  at com.microsoft.identity.common.java.crypto.StorageEncryptionManager.decrypt (SourceFile:19)
  at com.microsoft.identity.common.java.crypto.KeyAccessorStringAdapter.decrypt (SourceFile:2)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.encryptDecryptInternal (SourceFile:3)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.decrypt (SourceFile:1)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.getString (SourceFile:6)
  at com.microsoft.identity.common.internal.util.SharedPrefStringNameValueStorage.get (SourceFile:2)
  at com.microsoft.identity.common.internal.util.SharedPrefStringNameValueStorage.get (SourceFile:1)
  at com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCache.getAccount (SourceFile:3)
  at com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCache.saveAccount (SourceFile:5)
  at com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache.saveAccounts (SourceFile:2)
  at com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache.setSingleSignOnState (SourceFile:6)
  at com.microsoft.identity.common.adal.internal.cache.ADALOAuth2TokenCache.save (SourceFile:24)
  at com.microsoft.aad.adal.TokenCacheAccessor.updateTokenCacheUsingCommonCache (SourceFile:19)
  at com.microsoft.aad.adal.TokenCacheAccessor.updateCachedItemWithResult (SourceFile:10)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.acquireTokenWithCachedItem (SourceFile:7)
[...]

PR #1775, opened 2022-06-22
[Adal] Ignore failure in updating unified/msal cache if ignoreKeyLoaderNotFoundError flag is set

iamgusain:
The office mobile app is noticing a crash in the latest version (after upgrading to adal 4.1.0, common 4.2.0)
We provided a hotfix build of common (4.2.0-hf1) with additional logging. Which revealed that the crash is happening because the [UserDefinedKey](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/457f539c5569e303f5dd714559283a33b27dbb5e/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java#L102) is not set. This results in crash when performing decryption of cache content which were encrypted using UserDefinedKey. The key is set by calling setSecretKey method on AuthenticationSettings.INSTANCE.
From the callstack we see that the crash is happening when trying to update SSO state in unified/msal cache (as the content of that cache is still encrypted using USER_DEFINED_KEY)

```java
java.lang.IllegalStateException: 
  at com.microsoft.identity.common.java.crypto.StorageEncryptionManager.decrypt (SourceFile:19)
  at com.microsoft.identity.common.java.crypto.KeyAccessorStringAdapter.decrypt (SourceFile:2)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.encryptDecryptInternal (SourceFile:3)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.decrypt (SourceFile:1)
  at com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager.getString (SourceFile:6)
  at com.microsoft.identity.common.internal.util.SharedPrefStringNameValueStorage.get (SourceFile:2)
  at com.microsoft.identity.common.internal.util.SharedPrefStringNameValueStorage.get (SourceFile:1)
  at com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCache.getAccount (SourceFile:3)
  at com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCache.saveAccount (SourceFile:5)
  at com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache.saveAccounts (SourceFile:2)
  at com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache.setSingleSignOnState (SourceFile:6)
  at com.microsoft.identity.common.adal.internal.cache.ADALOAuth2TokenCache.save (SourceFile:24)
  at com.microsoft.aad.adal.TokenCacheAccessor.updateTokenCacheUsingCommonCache (SourceFile:19)
  at com.microsoft.aad.adal.TokenCacheAccessor.updateCachedItemWithResult (SourceFile:10)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.acquireTokenWithCachedItem (SourceFile:7)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.tryFRT (SourceFile:8)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.tryMRRT (SourceFile:8)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.tryRT (SourceFile:15)
  at com.microsoft.aad.adal.AcquireTokenSilentHandler.getAccessToken (SourceFile:9)
  at com.microsoft.aad.adal.AcquireTokenRequest.tryAcquireTokenSilentLocally (SourceFile:3)
  at com.microsoft.aad.adal.AcquireTokenRequest.acquireTokenSilentFlow (SourceFile:4)
  at com.microsoft.aad.adal.AcquireTokenRequest.tryAcquireTokenSilent (SourceFile:3)
  at com.microsoft.aad.adal.AcquireTokenRequest.performAcquireTokenRequest (SourceFile:1)
  at com.microsoft.aad.adal.AcquireTokenRequest.access$200 (SourceFile:1)
  at com.microsoft.aad.adal.AcquireTokenRequest$1.run (SourceFile:4)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:923)
```

The root cause of why the user defined key is not set is not yet known and is still under investigation
To avoid the crash, in this change we are ignoring the exception when updating unified/msal cache in ADAL flow), 
This will only be applicable if the flag ignoreKeyLoaderNotFoundError is set to true. Which OfficeMobile application will need to explicitly set. by calling `AuthenticationSettings.INSTANCE.SetIgnoreKeyLoaderNotFoundError(true)` method 

Related PR: [#1762](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1762)
Bug: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1944140

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -424,7 +424,7 @@ public enum AuthenticationSettings {
      * Method to check whether to suppress errors where KeyLoader is not found to decrypt
      * the cache content.
      */
-    public boolean getIgnoreKeyLoaderNotFoundError() {
+    public boolean shouldIgnoreKeyLoaderNotFoundError() {
         return mIgnoreKeyLoaderNotFoundError;
     }
 }
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -416,6 +416,7 @@ public enum AuthenticationSettings {
      * Method to suppress errors where KeyLoader is not found to decrypt the cache content
      * @param shouldIgnore if true, ignores keyloader not found errors
      */
+    @SuppressFBWarnings(ME_ENUM_FIELD_SETTER)
     public void setIgnoreKeyLoaderNotFoundError(boolean shouldIgnore) {
         mIgnoreKeyLoaderNotFoundError = shouldIgnore;
     }
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -94,7 +94,7 @@ public enum AuthenticationSettings {
 
     private int mReadTimeOut = DEFAULT_READ_CONNECT_TIMEOUT;
 
-    private boolean mShouldIgnorePredefinedKeyLoaderNotFoundError = false;
+    private boolean mIgnoreKeyLoaderNotFoundError = false;
 
     /**
      * Get bytes to derive secretKey to use in encrypt/decrypt.
@@ -413,18 +413,18 @@ public enum AuthenticationSettings {
     }
 
     /**
-     * Method to suppress errors where predefinedKeyLoader is not found to decrypt the cache content
-     * @param shouldIgnorePredefinedKeyLoaderNotFoundError if true, suppresses the error
+     * Method to suppress errors where KeyLoader is not found to decrypt the cache content
+     * @param shouldIgnore if true, ignores keyloader not found errors
      */
-    public void setShouldIgnorePredefinedKeyLoaderNotFoundError(boolean shouldIgnorePredefinedKeyLoaderNotFoundError) {
-        mShouldIgnorePredefinedKeyLoaderNotFoundError = shouldIgnorePredefinedKeyLoaderNotFoundError;
+    public void setIgnoreKeyLoaderNotFoundError(boolean shouldIgnore) {
+        mIgnoreKeyLoaderNotFoundError = shouldIgnore;
     }
 
     /**
-     * Method to check whether to suppress errors where predefinedKeyLoader is not found to decrypt
+     * Method to check whether to suppress errors where KeyLoader is not found to decrypt
      * the cache content.
      */
-    public boolean getShouldIgnorePredefinedKeyLoaderNotFoundError() {
-        return mShouldIgnorePredefinedKeyLoaderNotFoundError;
+    public boolean getIgnoreKeyLoaderNotFoundError() {
+        return mIgnoreKeyLoaderNotFoundError;
     }
 }
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
@@ -94,6 +94,8 @@ public enum AuthenticationSettings {
 
     private int mReadTimeOut = DEFAULT_READ_CONNECT_TIMEOUT;
 
+    private boolean mShouldIgnorePredefinedKeyLoaderNotFoundError = false;
+
     /**
      * Get bytes to derive secretKey to use in encrypt/decrypt.
      *
@@ -408,5 +410,21 @@ public enum AuthenticationSettings {
      */
     public boolean getDisableWebViewHardwareAcceleration() {
         return mEnableHardwareAcceleration;
+    }
+
+    /**
+     * Method to suppress errors where predefinedKeyLoader is not found to decrypt the cache content
+     * @param shouldIgnorePredefinedKeyLoaderNotFoundError if true, suppresses the error
+     */
+    public void setShouldIgnorePredefinedKeyLoaderNotFoundError(boolean shouldIgnorePredefinedKeyLoaderNotFoundError) {
+        mShouldIgnorePredefinedKeyLoaderNotFoundError = shouldIgnorePredefinedKeyLoaderNotFoundError;
+    }
+
+    /**
+     * Method to check whether to suppress errors where predefinedKeyLoader is not found to decrypt
+     * the cache content.
+     */
+    public boolean getShouldIgnorePredefinedKeyLoaderNotFoundError() {
+        return mShouldIgnorePredefinedKeyLoaderNotFoundError;
     }
 }
--- a/common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
@@ -81,8 +81,6 @@ public class AndroidAuthSdkStorageEncryptionManager extends StorageEncryptionMan
         if (PredefinedKeyLoader.USER_PROVIDED_KEY_IDENTIFIER.equalsIgnoreCase(keyIdentifier)) {
             if (mPredefinedKeyLoader != null) {
                 return Collections.<AbstractSecretKeyLoader>singletonList(mPredefinedKeyLoader);
-            } else if (AuthenticationSettings.INSTANCE.getShouldIgnorePredefinedKeyLoaderNotFoundError()) {
-                return null;
             } else {
                 throw new IllegalStateException(
                         "Cipher Text is encrypted by USER_PROVIDED_KEY_IDENTIFIER, " +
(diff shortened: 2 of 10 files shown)


common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
```diff
@@ -199,8 +199,14 @@ public byte[] decrypt(final byte[] cipherText) throws ClientException {
         }
 
         final List<AbstractSecretKeyLoader> keysForDecryption = getKeyLoaderForDecryption(cipherText);
-        if (keysForDecryption == null || keysForDecryption.size() == 0) {
-            throw new IllegalStateException("KeyLoader list must not be null or empty.");
+        if (keysForDecryption == null) {
```
rpdome, line 202: I'm thinking maybe expose the settings as "shouldNotCrashIfDataNotEncrypted", and then use that flag here.

The other method we can keep it align with the lombok nonnull identifier.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
```diff
@@ -81,6 +81,8 @@ public AndroidAuthSdkStorageEncryptionManager(@NonNull final Context context,
         if (PredefinedKeyLoader.USER_PROVIDED_KEY_IDENTIFIER.equalsIgnoreCase(keyIdentifier)) {
             if (mPredefinedKeyLoader != null) {
                 return Collections.<AbstractSecretKeyLoader>singletonList(mPredefinedKeyLoader);
+            } else if (AuthenticationSettings.INSTANCE.getShouldIgnorePredefinedKeyLoaderNotFoundError()) {
```
rpdome, line 84: and then do this

```Java
if (mPredefinedKeyLoader == null){
    // log
}
return Collections.<AbstractSecretKeyLoader>singletonList(mPredefinedKeyLoader);
```

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
rpdome: If the deta is unencrypted, it'll likely go into line 95.

rpdome: (My understanding of the issue is that Office gets the unencrypted data while it's reading).

rpdome (changes requested):
Re-reviewing. undoing my approval until then.

rpdome:
My understanding of the issue is that Office gets the unencrypted data while it's reading.

In that case, you might want to revert getKeyLoaderForDecryption to
```Java

    @Override
    public @NonNull List<AbstractSecretKeyLoader> getKeyLoaderForDecryption(@NonNull byte[] cipherText) {
        final String methodTag = TAG + ":getKeyLoaderForDecryption";
        if (mPredefinedKeyLoader != null &&
                isEncryptedByThisKeyIdentifier(cipherText, PredefinedKeyLoader.USER_PROVIDED_KEY_IDENTIFIER)) {
            return Collections.<AbstractSecretKeyLoader>singletonList(mPredefinedKeyLoader);
        }

        if (isEncryptedByThisKeyIdentifier(cipherText, AndroidWrappedKeyLoader.KEY_IDENTIFIER)) {
            return Collections.<AbstractSecretKeyLoader>singletonList(mKeyStoreKeyLoader);
        }

        Logger.warn(methodTag, "Cannot find a matching key to decrypt the given blob");
        return Collections.emptyList();
    }
```
(Keep the logging you added, of course).

Then in storageEncryptionManager, do this.

```Java
if (keysForDecryption == null || keysForDecryption.size() == 0) {
       if (shouldNotCrash) {
             // return as is.
       }
       throw new IllegalStateException("KeyLoader list must not be null or empty.");
}
```

(Might also worth printing the stack trace if the issue happens.
https://stackoverflow.com/questions/1069066/how-can-i-get-the-current-stack-trace-in-java)

shahzaibj:
I suggest looking at the feedback from Dome. I think that would be reverting it to pre-refactoring code in the true sense that it can handle un-encrypted data.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -409,4 +411,20 @@ public void setDisableWebViewHardwareAcceleration(boolean enable) {
     public boolean getDisableWebViewHardwareAcceleration() {
         return mEnableHardwareAcceleration;
     }
+
+    /**
+     * Method to suppress errors where predefinedKeyLoader is not found to decrypt the cache content
+     * @param shouldIgnorePredefinedKeyLoaderNotFoundError if true, suppresses the error
+     */
+    public void setShouldIgnorePredefinedKeyLoaderNotFoundError(boolean shouldIgnorePredefinedKeyLoaderNotFoundError) {
```
moumighosh: just a nit: Assuming this is a new method, looking at the method name, it's long. 
If the intent is to ignore cases where predefined key is not set, would "IgnorePredefinedKeyNotFoundErrors"

common/src/test/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManagerTest.java
```diff
@@ -121,6 +121,22 @@ public void testGetDecryptionKey_ForDataEncryptedWithPreDefinedKey() {
         }
     }
 
+    @Test
+    public void testGetDecryptionKey_ForDataEncryptedWithPreDefinedKey_PredefinedKeyNotProvided_returns_null_When_shouldIgnorePredefinedKeyLoaderNotFoundError_is_true() {
```
moumighosh: Nit: Very long method names are unreadable. Is it possible to capture the context in a comment and keep the method name short and meaningful?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -409,4 +411,20 @@ public void setDisableWebViewHardwareAcceleration(boolean enable) {
...
+     */
+    public void setIgnoreKeyLoaderNotFoundError(boolean shouldIgnore) {
+        mIgnoreKeyLoaderNotFoundError = shouldIgnore;
+    }
+
+    /**
+     * Method to check whether to suppress errors where KeyLoader is not found to decrypt
+     * the cache content.
+     */
+    public boolean getIgnoreKeyLoaderNotFoundError() {
```
shahzaibj: Nit:

```suggestion
    public boolean shouldIgnoreKeyLoaderNotFoundError() {
```

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
```diff
@@ -91,6 +91,6 @@ public AndroidAuthSdkStorageEncryptionManager(@NonNull final Context context,
         }
 
         Logger.warn(methodTag, "Cannot find a matching key to decrypt the given blob");
-        throw new IllegalStateException("Unknown keyIdentifier for cipher text: " + keyIdentifier);
+        return Collections.emptyList();
```
shahzaibj, line 94: So this is just reverting to what we have on dev, right?

shahzaibj (approved):

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
iamgusain: reverted this change back to what it is in dev, except for the additional logging

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
iamgusain: Updated to setIgnoreKeyLoaderNotFoundError

common/src/test/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManagerTest.java
iamgusain: I have shortened it.
This is the name of the test and makes it easier to read the Test failure reports to see exactly what test failed
That is the reason I prefer to keep the name in this format "<what_Is_being_Tested>_<With_Preconditions_or_Inputs>_<expected_behavior>"

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
iamgusain: yes

PR #1776, opened 2022-06-27
cacheValues might be null

p3dr0rv:
cacheValues is null in a test case of AunthenticatorAPIHelperTest
Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1916


shahzaibj:
probably one of the implementations of `getAllFilteredBy` is returning `null` - which I guess is okay given that the interface doesn't annotate this as NonNull

shahzaibj (approved):

iamgusain (approved):

PR #1777, opened 2022-06-28
[Ignore] Try tagger

mohitc1:
[Ignore] Try tagger


common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -271,6 +271,7 @@ public static void warnPII(final String tag,
     public static void info(final String tag,
                             final String message) {
         Logger.emit(0, String.format("%s - %s", tag, message));
+        Logger.emit(0, String.format("%s - %s", tag, message));
```
bktowett, line 274: Why are the two lines identical?

PR #1780, opened 2022-07-01
Updating common4j version for common release

mohitc1:
Updating common4j version for common release to 3.0.0-RC1

### DIFF
--- a/common4j/versioning/version.properties
+++ b/common4j/versioning/version.properties
@@ -1,4 +1,4 @@
 #Wed May 12 20:08:39 UTC 2021
-versionName=2.0.0
+versionName=3.0.0-RC1
 versionCode=1
 latestPatchVersion=227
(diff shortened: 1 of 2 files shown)


common4j/versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Wed May 12 20:08:39 UTC 2021
-versionName=2.0.0
+versionName=3.0.0-RC1
```
iamgusain, line 2: You also need to point to this version of common4j in build.gradle for common

iamgusain: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/9c8f19f010acb0315cddea4c2303d74204a4b4a0/common/build.gradle#L31-L34

iamgusain (approved):

PR #1788, opened 2022-07-08
Fix failing msal tests when creating android telemetry context

gitaumoses4:
Failing pipeline: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=952752&view=ms.vss-test-web.build-test-results-tab&runId=2170076&resultId=100270&paneView=debug

## Stacktrace: 
android.content.res.Resources$NotFoundException: String resource ID #0x0
      at android.content.res.Resources.getText(Resources.java:348)
      at android.content.res.Resources.getString(Resources.java:441)
      at android.content.Context.getString(Context.java:578)
      at com.microsoft.identity.common.internal.telemetry.AndroidTelemetryContext.addApplicationInfo(AndroidTelemetryContext.java:71)
      at com.microsoft.identity.common.internal.telemetry.AndroidTelemetryContext.<init>(AndroidTelemetryContext.java:52)
  
## Why is this happening?

This check is wrong: 
```
applicationName = applicationInfo.labelRes == 0 && applicationInfo.nonLocalizedLabel != null ?
                        applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
```

If the `labelRes` is zero and `nonLocalizedLabel` is null, then the code `context.getString()` is executed. But it shouldn't because the labelRes it `zero`.

## Changes made.
Split the check into if statements to ensure that `context.getString()` is not executed when labelRes is 0


shahzaibj:
Can you add a description in the PR explaining what actually causes the resource not found issue?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -67,9 +68,13 @@ private void addApplicationInfo(@NonNull final Context context) {
             String packageName = "";
 
             if (applicationInfo != null) {
-                applicationName = applicationInfo.labelRes == 0 && applicationInfo.nonLocalizedLabel != null ?
-                        applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
                 packageName = applicationInfo.packageName;
+                try {
+                    applicationName = context.getString(applicationInfo.labelRes);
+                } catch (Resources.NotFoundException exception) {
```
shahzaibj: What would actually cause to happen? Can it be prevented altogether?

gitaumoses4: This check is wrong:
```
applicationName = applicationInfo.labelRes == 0 && applicationInfo.nonLocalizedLabel != null ?
                        applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
```
If the labelRes is zero and nonLocalizedLabel is null, then the code context.getString() is executed. But it shouldn't because the labelRes it zero.

shahzaibj: I was actually asking about the new code here:

context.getString(applicationInfo.labelRes);

Seems like this throws Resources.NotFoundException
In what cases does it through that exception? I mean how likely are we to hit this exception in our production code?

gitaumoses4: This cannot happen in production since there is an application label in the application context. 
It will only fail for our robolectric tests.

shahzaibj: Okay, I think we should still keep the try/catch block and log the error when they occur

gitaumoses4: It only throws an exception when `labelRes` is equal to zero. Meaning that the application context does not have an application label. This cannot happen for production apps since they all have labels, but will fail for robolectric tests. 

Instead of executing `context.getString(applicationInfo.labelRes)` in a try-catch block, i've just used an if statement since we alreay know the cases when an exeption will be thrown, and this is when `labelRes == 0`.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -67,9 +68,14 @@ private void addApplicationInfo(@NonNull final Context context) {
             String packageName = "";
 
             if (applicationInfo != null) {
-                applicationName = applicationInfo.labelRes == 0 && applicationInfo.nonLocalizedLabel != null ?
-                        applicationInfo.nonLocalizedLabel.toString() : context.getString(applicationInfo.labelRes);
                 packageName = applicationInfo.packageName;
+
+                if (applicationInfo.labelRes == 0) {
```
shahzaibj, line 73: Consider adding some comments here describing when this might be zero

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -67,9 +68,14 @@ private void addApplicationInfo(@NonNull final Context context) {
...
+                    applicationName = applicationInfo.nonLocalizedLabel == null ? packageName :
```
shahzaibj, line 74: Consider adding some comments describing the difference between label and package name. Also when might the label be null

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
```diff
@@ -67,9 +68,14 @@ private void addApplicationInfo(@NonNull final Context context) {
...
+                            applicationInfo.nonLocalizedLabel.toString();
+                } else {
+                    applicationName = context.getString(applicationInfo.labelRes);
```
shahzaibj, line 77: We probably want to add some logging here...verbose level

common/src/test/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContextTest.java
```diff
@@ -0,0 +1,55 @@
...
+    public void setup() {
+        androidTelemetryContext = new AndroidTelemetryContext(ApplicationProvider.getApplicationContext());
+    }
+
+    @Test
+    public void testAndroidTelemetryContextApplicationProperties() {
+        final ConcurrentHashMap<String, String> properties = androidTelemetryContext.getProperties();
+
+        Assert.assertEquals("org.robolectric.default", properties.get(TelemetryEventStrings.App.NAME));
+        Assert.assertEquals("org.robolectric.default", properties.get(TelemetryEventStrings.App.PACKAGE));
```
shahzaibj, line 52: `org.robolectric.default` can be declared as a constant in this file

shahzaibj (approved):
Left some minor comments

changelog.txt
```diff
@@ -12,6 +12,7 @@ V.Next
 - [PATCH] Fix APPLICATION_CANCELLED by handling back button press (#1725)
 - [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
 - [PATCH] Fix missing authority_url when creating the authority audience (#1753)
+- [PATCH] Fix msal failing tests due to telemetry context
```
iamgusain: Lets add this at the top and also include PR#  #Resolved

iamgusain (approved):

mohitc1 (approved):

bktowett (approved):

common/src/main/java/com/microsoft/identity/common/internal/telemetry/AndroidTelemetryContext.java
somalaya, line 73: Does this refer to the <application>label</application> in AndroidManifest file? If yes, can you give some context on how this label is populated for production apps? I was checking MSAL AndroidManifest file and could not find the label tag like it is added in Broker. Even though it is optional,  is there any harm in adding this tag to MSAL as well? And does it resolve the issue in test apps as well?

PR #1789, opened 2022-07-11
Remove SSLContext From HttpClient signature

rpdome:
This is to make way for an HttpClient instance in Linux Broker that will act as a proxy to the Device Broker, to make a request with Device Private Key (for Client TLS challenge) there.

The idea is that 
1. we'll remove the SSLContext object from HttpClient signature.
2. For non-proxy HttpClient object (UrlConnectionHttpClient), SSLContext will be constructed - or loaded - via its constructor/builder.
3. For proxy HttpClient object, we'll take whatever it requires (i.e. key entry, certificate), route the request to Linux Broker, and then use it to construct an SSLContext over there.

This is a breaking change for broker.. so the broker test will fail.
Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/1926

Validated with WPJ's getDeviceState() - as part of https://github.com/AzureAD/ad-accounts-for-android/pull/1928

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
@@ -32,7 +32,6 @@ import com.microsoft.identity.common.java.util.ported.Consumer;
 import com.microsoft.identity.common.java.util.ported.Function;
 import com.microsoft.identity.common.java.util.ported.Supplier;
 
-import net.jcip.annotations.Immutable;
 import net.jcip.annotations.ThreadSafe;
 
 import java.io.BufferedReader;
@@ -52,6 +51,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicReference;
 
+import javax.annotation.Nullable;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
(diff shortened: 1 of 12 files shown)


common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -75,45 +75,93 @@
...
 
     private static final transient AtomicReference<UrlConnectionHttpClient> defaultReference = new AtomicReference<>(null);
 
+    /**
+     * Retry policy of this HttpClient. Default is {@link NoRetryPolicy}.
+     */
+    private final IRetryPolicy<HttpResponse> retryPolicy;
+
+    /**
+     * Retry policy of this HttpClient. Default is {@link NoRetryPolicy}.
```
shahzaibj: This javadoc seems wrong

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -75,45 +75,93 @@
...
+     * Default Constructor, for constructing Lombok's Builder only. Do not expose.
+     */
+    @Builder
+    private UrlConnectionHttpClient(@Nullable final IRetryPolicy<HttpResponse> retryPolicy,
+                                    @Nullable final Integer streamBufferSize,
+                                    @Nullable final Integer connectTimeoutMs,
+                                    @Nullable final Integer readTimeoutMs,
+                                    @Nullable final Supplier<Integer> connectTimeoutMsSupplier,
+                                    @Nullable final Supplier<Integer> readTimeoutMsSupplier,
+                                    @Nullable final List<String> supportedSslProtocol,
```
shahzaibj:
```suggestion
                                    @Nullable final List<String> supportedSslProtocols,
```

shahzaibj (approved):
lgtm

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -75,45 +75,93 @@
...
-
-    private SSLSocketFactoryWrapper sDefault;
-
-    private synchronized SSLSocketFactoryWrapper getDefaultWrapper(){
-        if (sDefault == null){
-            sDefault = new SSLSocketFactoryWrapper((SSLSocketFactory) SSLSocketFactory.getDefault(), supportedSslProtocol);
-        }
-        return sDefault;
-    }
+    public static final int DEFAULT_STREAM_BUFFER_SIZE_BYTE = 1024;
```
iamgusain: do we need to make it public?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -75,45 +75,93 @@
...
+        this.readTimeoutMs = readTimeoutMs != null ?
+                readTimeoutMs : DEFAULT_READ_TIME_OUT_MS;
+        this.connectTimeoutMsSupplier = connectTimeoutMsSupplier;
+        this.readTimeoutMsSupplier = readTimeoutMsSupplier;
+
+        final List<String> protocol = supportedSslProtocols != null ?
+                supportedSslProtocols : SSLSocketFactoryWrapper.SUPPORTED_SSL_PROTOCOLS;
+
+        if (sslContext == null) {
+            // TODO: Build the default SSLContext ourselves with the defined JSSE provider.
```
iamgusain, line 156: Until the TODO is done, should we add a log statement here for diagnostic purpose?

rpdome: This is a todo for FIPS work. There's no behavior change at the moment.

iamgusain:
As this is a breaking change, can you add in the description how did you test the changes. It would be helpful for later.
Also I don't see changelog entry.

iamgusain (approved):

PR #1790, opened 2022-07-11
Send PKeyAuth Header to token endpoint for required msal-broker protocol version 9.0

p3dr0rv:
### Why
On #1687 we started sending x-ms-PKeyAuth in both interactive and silent flow, fixing the silent flow pkeyauth. 
This change broke a client whose misconfiguration relied on the old flow (see related ICM: [31201066](https://portal.microsofticm.com/imp/v3/incidents/details/312010665/home)).
So, we stopped sending the pkeyauth header to the token endpoint on #1755.
We are bringing back this fix in secure way adding a version lock to avoid breaking customers.
Sending the pkeyauth header to the token endpoint only to clients who configure the minimum_required_broker_protocol_version to version 9.0.  

[Android Auth Broker Protocol Versions](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Broker%20API/broker_protocol_versions.md&version=GBdev&_a=preview)
Related PR: [broker#1927](https://github.com/AzureAD/ad-accounts-for-android/pull/1927)

### What change

* Added a boolean value to the acquire token Interactive/silent request bundle that determines if we should send the pkeauth header to the token endpoint.
* On minimum_required_broker_protocol_version to version 9.0 or higher we set this boolean value to true, otherwise false.
*  propagate the flag to OAuth2Strategy using  Broker[Silent,Interactive]TokenCommandParameters and MicrosoftTokenRequest

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Bumped MSAL Broker Protocol version to 9.0, acquireToken/acquireTokenSilent endpoint requires minimum_required_broker_protocol_version of 9.0+ to Send x-ms-PKeyAuth Header to the token endpoint. (#1790)
 - [MAJOR] Bumped MSAL Broker Protocol version to 8.0, GET_ACCOUNTS endpoint requires minimum_required_broker_protocol_version of 8.0+ to return an account constructed from PRT id token to FOCI apps. (#1771)
 - [MAJOR] [Msal] Remove launching logout endpoint in default browser for shared device mode signout flow
 - [PATCH] [Adal] Ignore failure in updating unified/msal cache if ignoreKeyLoaderNotFoundError flag is set (#1781)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 V.Next
 ----------
-- [MAJOR] Bumped MSAL Broker Protocol version to 9.0, acquireToken/acquireTokenSilent endpoint requires minimum_required_broker_protocol_version of 9.0+ to Send x-ms-PKeyAuth Header to the token endpoint. (#1790)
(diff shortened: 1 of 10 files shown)


common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftTokenRequest.java
```diff
@@ -97,6 +100,11 @@ public MicrosoftTokenRequest() {
     // Sent as part of headers if available, so marking it transient.
     private transient String mBrokerVersion;
 
+    @Getter
```
rpdome, line 104: Nit: add comment

rpdome (approved):

changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Bumped MSAL Broker Protocol version to 9.0, acquireToken/acquireTokenSilent endpoint requires minimum_required_broker_protocol_version of 9.0+ to Send x-ms-PKeyAuth Header to the token endpoint. (#1790)
```
p3dr0rv: https://github.com/AzureAD/ad-accounts-for-android/pull/1927#discussion_r924968243

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1159,6 +1159,11 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String BROKER_REQUEST_V2_SUCCESS = "broker_request_v2_success";
 
+        /**
+         * String to send true if the request should send the PkeyAuth header to the token endpoint, false otherwise.
+         */
+        public static final String SHOULD_SEND_PKEYAUTH_HEADER_TO_THE_TOKEN_ENDPOINT = "should.send.pkeyauth.header";
```
iamgusain, line 1165: Should we add Silent_Flow to this variable, as we are not using this in the Interactive flow

iamgusain (approved):
Please add the ICM link in the PR description to add more context

PR #1793, opened 2022-07-13
Add filter for broker telemetry event fields

gitaumoses4:
## Why 
Since we need to do the privacy review for telemetry, for now, we should add a filter to NOT upload anything for which we aren’t sure whether that is PII or OII. 

## How
Add a filter that only allows specific fields to be published to a telemetry relay client

## Broker PR
Here is the related broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1931

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,6 +13,7 @@ Version 6.0.0
 - [PATCH] Fix APPLICATION_CANCELLED by handling back button press (#1725)
 - [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
 - [PATCH] Fix missing authority_url when creating the authority audience (#1753)
+- [PATCH] Add filter for broker telemetry event fields.
 
 Version 5.0.0
 ----------
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,7 +13,7 @@ Version 6.0.0
 - [PATCH] Fix APPLICATION_CANCELLED by handling back button press (#1725)
 - [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
 - [PATCH] Fix missing authority_url when creating the authority audience (#1753)
-- [PATCH] Add filter for broker telemetry event fields.
+- [PATCH] Add filter for broker telemetry event fields. (#1793)
 
 Version 5.0.0
 ----------
--- a/common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
@@ -25,9 +25,6 @@ package com.microsoft.identity.common.java.telemetry.relay;
 import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.telemetry.observers.ITelemetryObserver;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import lombok.NonNull;
 
 /**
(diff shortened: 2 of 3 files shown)


changelog.txt
```diff
@@ -13,6 +13,7 @@ Version 6.0.0
 - [PATCH] Fix APPLICATION_CANCELLED by handling back button press (#1725)
 - [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
 - [PATCH] Fix missing authority_url when creating the authority audience (#1753)
+- [PATCH] Add filter for broker telemetry event fields.
```
gitaumoses4:
```suggestion
- [PATCH] Add filter for broker telemetry event fields. (#1793)
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
```diff
@@ -39,33 +36,33 @@
...
+                relayEvent(filteredData);
+            } catch (TelemetryRelayException e) {
+                Logger.error(methodTag, "Error relaying telemetry data", e);
+            }
         }
     }
 
     /**
-     * Add an event filter to the list of filters
+     * Add an event filter
```
shahzaibj:
```suggestion
     * Add an event filter.
```

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
```diff
@@ -39,33 +36,33 @@
 public abstract class AbstractTelemetryRelayClient<T> implements ITelemetryObserver<T> {
 
     private static final String TAG = AbstractTelemetryRelayClient.class.getSimpleName();
-    // Filters to apply on the events captured.
-    // This basically performs an AND operation on the filters to determine whether the event will be relayed.
-    private final List<ITelemetryEventFilter<T>> eventFilters = new ArrayList<>();
+    // Filter to apply on the events captured.
+    private ITelemetryEventFilter<T> mEventFilter = null;
```
shahzaibj, line 40: It seems we've changed the model of the code here now and only going to allow one filter. What was the rationale behind this change. Are there pros and cons to this approach over the previous approach?

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
```diff
@@ -39,33 +36,33 @@
...
 
-        try {
-            relayEvent(telemetryData);
-        } catch (TelemetryRelayException e) {
-            Logger.error(methodTag, "Error relaying telemetry data", e);
+        if (filteredData != null) {
+            try {
+                relayEvent(filteredData);
+            } catch (TelemetryRelayException e) {
+                Logger.error(methodTag, "Error relaying telemetry data", e);
```
shahzaibj, line 56: What might cause a relay exception?

shahzaibj: I'm thinking in this case we may want to hold onto the data in some sort of telemetry cache and make another attempt at relaying it in the future....just something to think about...we don't have to solve that problem right now

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/ITelemetryEventFilter.java
```diff
@@ -31,7 +31,7 @@
      * Invoked when a new event is captured by the telemetry.
      * @param telemetryEvent the telemetry event data
      *
-     * @return a boolean representing whether the event should be relayed.
+     * @return an event with filtered fields. Return null if the event is to be ignored completely.
      */
-    boolean shouldRelay(T telemetryEvent);
+    T apply(T telemetryEvent);
```
shahzaibj: Seems like this is `Nullable` -probably want to mark it as such here

shahzaibj (approved):
Some basic comments...but I wouldn't block on them

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/telemetry/relay/AbstractTelemetryRelayClient.java
gitaumoses4: > What might cause a relay exception?

Possibly a deleted aria instance.

PR #1796, opened 2022-07-18
set lab secret in env.file for Instrumented test

p3dr0rv:
new network tests added to the broker Instrumented test pipeline needed the lab secret, so we are providing it here.
https://github.com/AzureAD/ad-accounts-for-android/pull/1935

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=957491


shahzaibj (approved):
Thanks @p3dr0rv

iamgusain (approved):

PR #1770, opened 2022-06-22
Capture telemetry in headers for FLW and Multiple WPJ

shahzaibj:
Related: https://github.com/AzureAD/ad-accounts-for-android/pull/1912

### Description
I would suggest reading the description in the above linked related PR to get a complete sense of what's going on.

In a nutshell, we've made some changes here to facilitate us to capture header based telemetry around FLW and multiple Workplace Join. This is based on Olga's design PR here: https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview/pullrequest/5168

Basically, we've separated Android platform fields with shared platform fields we're going to have with Android and iOS.

There are going to be 3 sets of platform fields:

- Android only platform fields
- Android and iOS shared platform fields for FLW
- Android and iOS shared platform fields for Multiple WPJ

The shared platform fields will always come first in the schema and Android only will come second. 

In addition to that, this PR also contains a bunch of utilities class around http request interception, examination, and re-writing to facilitate tests that involve us to do something with the network request. For example, following interceptors are introduced:

- HttpRequestExaminer - allows to intercept and examine the request URL and headers of an HTTP request
- HttpRequestRewriter - allows to intercept and re-write headers or URL or body of an HTTP request
- HttpResponseRewriter - allows to intercept and re-write the response of an HTTP request

PR #1797, opened 2022-07-18
Move common http test utilities from testutils to common4j test fixtures

shahzaibj:
Description: Move common http test utilities from testutils to common4j test fixtures.

Why we are moving these classes? We are moving these classes to test fixtures because test fixtures is where they truly belong. The primary purpose of test fixtures is to be able to share test code across modules. 

This is to break https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1770 into multiple smaller PRs.


rpdome (approved):

iamgusain:
Would be nice to update the description with what is being changed here

shahzaibj:
> Would be nice to update the description with what is being changed here

I've added the title to description as well. That's basically all there is to this PR. I'm not sure what more I'd put for a simple PR like this that's just relocated some files.

common4j/build.gradle
```diff
@@ -187,10 +213,15 @@ dependencies {
...
+
+    testFixturesCompileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.12'
+    testFixturesCompileOnly "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
+    testFixturesAnnotationProcessor "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
+    resolvableTestFixturesImplementation 'org.robolectric:junit:4.5.1'
 }
 
-sourceCompatibility = "1.7"
-targetCompatibility = "1.7"
+sourceCompatibility = "1.8"
```
iamgusain, line 223: Why is this change needed?

iamgusain (approved):

iamgusain:
> > Would be nice to update the description with what is being changed here
> 
> I've added the title to description as well. That's basically all there is to this PR. I'm not sure what more I'd put for a simple PR like this that's just relocated some files.

Something like why we are moving the classes will add some context, when we look back at the PR later.

shahzaibj:
> > > Would be nice to update the description with what is being changed here
> > 
> > 
> > I've added the title to description as well. That's basically all there is to this PR. I'm not sure what more I'd put for a simple PR like this that's just relocated some files.
> 
> Something like why we are moving the classes will add some context, when we look back at the PR later.

Added this:

Why we are moving these classes? We are moving these classes to test fixtures because test fixtures is where they truly belong. The primary purpose of test fixtures is to be able to share test code across modules.

Test Fixtures is a concept that @p3dr0rv had introduced to the team some time ago, and I think I also covered it again in one of my recent brown-bags as well.

tanmaymanolkar1 (approved):

common4j/build.gradle
shahzaibj: I think I needed it for some other code I wrote in the bigger PR. So this is a remnant from that but since it belonged in `build.gradle` therefore I included it here as all the `build.gradle` changes were here so it was easier for me to organize the PRs. 

Also I think we've actually already technically been on Java 8 since Melissa's YubyKit changes (YubyKit SDK is on Java 8). We just hadn't explicitly made the change in build.gradle here.

PR #1798, opened 2022-07-18
Add http request interceptors that allow to examine, and/or modify re…

shahzaibj:
…quest or response

This PR contains some utilities classes around http request interception, examination, and re-writing to facilitate tests that involve us to do something with the network request. For example, following interceptors are introduced:

HttpRequestExaminer - allows to intercept and examine the request URL and headers of an HTTP request
HttpRequestRewriter - allows to intercept and re-write headers or URL or body of an HTTP request
HttpResponseRewriter - allows to intercept and re-write the response of an HTTP request

Some of these interceptors will be used to write telemetry tests to intercept request and assert for the presence of telemetry headers in the request.

### DIFF
--- a/common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestExaminer.java
+++ b/common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestExaminer.java
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.http;
+
+import com.microsoft.identity.common.java.net.HttpClient;
+import com.microsoft.identity.common.java.net.HttpResponse;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Map;
+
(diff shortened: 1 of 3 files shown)


rpdome (approved):

common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestRewriter.java
```diff
@@ -0,0 +1,89 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+
+/**
+ * An {@link HttpRequestInterceptor} that allows to re-write the http request and then proceeds
+ * with the modified request.
+ */
+@AllArgsConstructor
+public abstract class HttpRequestRewriter implements HttpRequestInterceptor {
```
rpdome, line 41: I can see how the other ones are being used (examiner = check request content, rewriter = modify response), but i'm not sure what's the purpose of this one.

rpdome: That said, if I were to write test, i would probably need both examier and rewriter in one class (check if request is valid, and mock the response).

but i haven't seen how you're going to use it, so i can't tell. signing off because the current code looks ok.

shahzaibj: Intercept the request as it is about to go in the wire, and then modify something in the request and then send it on the wire.

For example, one may choose to remove a particular header from request headers and then send it forward.

shahzaibj: Yes, I think what we need is the ability to chain multiple interceptors together to be able to examine, re-write request as well as rewrite response; all for one request, however, the existing model in the code didn't make it so easy to implement that. What we've done here is provided an improvement to the existing model and an opportunity to ourselves to take it and improve it further.

common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestExaminer.java
```diff
@@ -0,0 +1,72 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+
+/**
+ * An {@link HttpRequestInterceptor} that allows to examine the request and request headers and then
+ * proceeds the original request.
+ */
+@AllArgsConstructor
+public abstract class HttpRequestExaminer implements HttpRequestInterceptor {
```
iamgusain, line 41: nit: We can add default implementations to the examineRequestUrl and examineRequestHeaders methods to match the values or take Predicates to verify those.

shahzaibj: I think we can, but I personally wouldn't spend time on this right now. (Very Limited Capacity). Feel free to create a work-item in the backlog and someone should be able to pick it up in the future.

common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestRewriter.java
iamgusain, line 41: Should we add a new class for this or add these methods to the HttpRequestInterceptor itself..
I don't have strong preference, but feel that having a single place for all this kind of operation makes it more easily discoverable.

iamgusain (approved):
Approved with suggestions

common4j/src/testFixtures/java/com/microsoft/identity/http/HttpRequestRewriter.java
shahzaibj: We can certainly do either. Probably needs a more thorough analysis of pros and cons of each approach. In general, I think a lot of things can be improved with the interceptors. We've improved a few things here and we should definitely revisit this in the future to improve more things.

PR #1800, opened 2022-07-22
Fix for failing foci scenarios

somalaya:
**What?** : AcquireTokenSilent is failing to fetch a token in foci scenarios. During the broker 5.0.0 bug bash, it was found that the FOCI tests are failing. Here's one of the FOCI tests: https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833519

**Why was it failing?**
It was failing at step 6. Word was not getting SSO from the user having already signed into Outlook (another FOCI app).

**Why was it happening?**
Use of TSL has verified foci membership and getAccounts, however, the following acquire token silent request fails with an account_not_found error because no matching account and/or credential is found in the cache. The reason is that [broker verifies that the account supplied for ATS is actually part of the cache](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/client/BrokerLocalController.java#L135). [The caching layer will see if there is an existing token cache for this client application](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java#L904). When TSL is used then there is NO existing token cache for client because the client hasn't got any tokens yet. So that part fails (it succeeds using ESTS request because that request seeds the token into the cache and so it creates an entry in the ApplicationMetadataCache). But the flow does not make an ESTS request because there is a check to verify if the cache instance is of type MsalOauth2TokenCache, which is not the case. It is actually BrokerOauth2TokenCache that is used. _Why did we add this  https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/32156ad74515e300394ee3997754b33e95713e83/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java#L865 condition? Can we safely remove it?_ If we can remove it.. Then we would be making use of this fallback i.e to make a service call to confirm the client is a FOCI app and the account is returned successfully.
**_Answer** :_  _The purpose of this instanceOf MsalOAuth2TokenCache check may be to protect it from being called for AdalOAuth2Cache. This was written during initial broker refactor, so this must have been added without taking into consideration that this flow is applicable for BrokerOAuth2Cache as well._ _I have not changed it to avoid any regressions._https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1999531/

Also, even if there is no existing cache for this particular client,  it should still work by virtue of having a dedicated foci cache, however,  even that doesn't work because we fall down here to look into the foci cache, but [we seem to be doing a look-up that requires a hard match on clientId](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java#L924) i.e. the look-up will only return credentials that were issued to that particular client id. This defeats the purpose of having the foci cache if we're doing a hard look-up on a specific client Id. 

**What is the fix?**
I made a change to make break the getCachedAccountRecords in BaseController into 2 parts
 1. To get accounts using localAccountId and clientId
 2. To get accounts from all caches including the FOCI cache - In baseController, I have not added anything in this function other than what was there earlier. But in a seperate PR I did an override of this method and   used the BrokerUtil.getCacheRecordListFromBrokerCache() method which internally verifies if calling app is eligible to get FOCI accounts. These override changes are in https://github.com/AzureAD/ad-accounts-for-android/pull/1951

**Testing my fix**
Tested my fix on Pixel 3 phone by building Authenticator app with my changes. 
Feel free to test the scenario with the following Authenticator build :  https://msazure.visualstudio.com/One/_build/results?buildId=58053075&view=artifacts&pathAsName=false&type=publishedArtifacts

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -865,6 +865,10 @@ public abstract class BaseController {
         return targetAccount;
     }
 
+    /**
+     * Lookup in app-specific cache.
+     */
+    @Nullable
     private AccountRecord getCachedAccountRecordFromCallingAppCache(
             @NonNull final SilentTokenCommandParameters parameters) {
         final boolean isB2CAuthority = B2C.equalsIgnoreCase(
@@ -909,6 +913,7 @@ public abstract class BaseController {
     /**
      * Lookup in ALL the caches including the foci cache.
      */
+    @Nullable
     protected AccountRecord getCachedAccountRecordFromAllCaches(
             @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
         // TO-DO https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1999531/
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -907,7 +907,7 @@ public abstract class BaseController {
     }
 
     /**
-     * Lookup in ALL the caches including the foci cache
+     * Lookup in ALL the caches including the foci cache.
      */
     protected AccountRecord getCachedAccountRecordFromAllCaches(
             @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -27,6 +27,8 @@ import static com.microsoft.identity.common.java.AuthenticationConstants.Broker.
 import static com.microsoft.identity.common.java.authorities.Authority.B2C;
 
 
+import com.microsoft.identity.common.java.cache.BrokerOAuth2TokenCache;
+import com.microsoft.identity.common.java.cache.CacheRecord;
 import com.microsoft.identity.common.java.commands.parameters.RopcTokenCommandParameters;
 import com.microsoft.identity.common.java.foci.FociQueryUtilities;
 import com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache;
@@ -486,7 +488,7 @@ public abstract class BaseController {
                     Should return existing AT without refreshing it.
                     This way caller will know whether to refresh based on this exception.
                  */
-                if(SERVICE_NOT_AVAILABLE.equals(errorCode)){
+                if (SERVICE_NOT_AVAILABLE.equals(errorCode)) {
                     throw new ServiceException(SERVICE_NOT_AVAILABLE, "AAD is not available.", tokenResult.getErrorResponse().getStatusCode(), null);
                 }
 
@@ -588,7 +590,7 @@ public abstract class BaseController {
                     Should return existing AT without refreshing it.
                     This way caller will know whether to refresh based on this exception.
                  */
-                if(SERVICE_NOT_AVAILABLE.equals(errorCode)){
+                if (SERVICE_NOT_AVAILABLE.equals(errorCode)) {
                     throw new ServiceException(SERVICE_NOT_AVAILABLE, "AAD is not available.", tokenResult.getErrorResponse().getStatusCode(), null);
                 }
 
@@ -625,7 +627,7 @@ public abstract class BaseController {
         return cacheRecords.get(0);
     }
 
-    public OAuth2TokenCache getTokenCache(@NonNull final SilentTokenCommandParameters parameters){
+    public OAuth2TokenCache getTokenCache(@NonNull final SilentTokenCommandParameters parameters) {
         //Extract cache from parameters
         return parameters.getOAuth2TokenCache();
     }
@@ -862,11 +864,14 @@ public abstract class BaseController {
                     );
         }
 
+        // TO-DO can we remove the parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache condition?
         if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
             targetAccount = getAccountWithFRTIfAvailable(
                     parameters,
                     (MsalOAuth2TokenCache) parameters.getOAuth2TokenCache()
             );
+        } else if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof BrokerOAuth2TokenCache) {
+            targetAccount = getAccountFromFociCacheUsingLocalAccountId((BrokerOAuth2TokenCache) parameters.getOAuth2TokenCache(), localAccountId);
         }
 
         if (null == targetAccount) {
@@ -879,7 +884,7 @@ public abstract class BaseController {
             } else {
                 Logger.error(
                         TAG,
-                        "No accounts found for clientId [" + clientId +"]",
+                        "No accounts found for clientId [" + clientId + "]",
                         null
                 );
             }
@@ -935,6 +940,32 @@ public abstract class BaseController {
         }
         return null;
     }
+
+    @Nullable
+    private AccountRecord getAccountFromFociCacheUsingLocalAccountId(@NonNull final BrokerOAuth2TokenCache brokerOAuth2TokenCache,
+                                                                     @NonNull final String localAccountId) {
+        final String methodName = ":getAccountFromFociCacheUsingLocalAccountId";
+        AccountRecord targetAccount = null;
+        final List<CacheRecord> fociCacheRecords = brokerOAuth2TokenCache.getFociCacheRecords();
+
+        if (fociCacheRecords.size() > 0) {
+            Logger.info(
+                    TAG + methodName,
+                    "There is (are) ["
+                            + fociCacheRecords.size()
+                            + "] Foci in the broker cache."
+            );
+
+            for (final ICacheRecord cacheRecord : fociCacheRecords) {
+                if (cacheRecord.getAccount() != null && localAccountId
+                        .equalsIgnoreCase(cacheRecord.getAccount().getLocalAccountId())) {
+                    targetAccount = cacheRecord.getAccount();
+                }
+            }
+        }
+        return targetAccount;
+    }
+
 
     /**
      * Helper method which returns false if the tenant id of the authority
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -26,9 +26,6 @@ import static com.microsoft.identity.common.java.AuthenticationConstants.Broker.
 import static com.microsoft.identity.common.java.AuthenticationConstants.Broker.PKEYAUTH_VERSION;
 import static com.microsoft.identity.common.java.authorities.Authority.B2C;
 
-
-import com.microsoft.identity.common.java.cache.BrokerOAuth2TokenCache;
-import com.microsoft.identity.common.java.cache.CacheRecord;
 import com.microsoft.identity.common.java.commands.parameters.RopcTokenCommandParameters;
 import com.microsoft.identity.common.java.foci.FociQueryUtilities;
 import com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache;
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -936,6 +941,32 @@ private AccountRecord getAccountWithFRTIfAvailable(@NonNull final SilentTokenCom
...
+                    TAG + methodName,
+                    "There is (are) ["
+                            + fociCacheRecords.size()
+                            + "] Foci in the broker cache."
+            );
+
+            for (final ICacheRecord cacheRecord : fociCacheRecords) {
+                if (cacheRecord.getAccount() != null && localAccountId
+                        .equalsIgnoreCase(cacheRecord.getAccount().getLocalAccountId())) {
+                    targetAccount = cacheRecord.getAccount();
```
shahzaibj: should we return here? Do we need to keep going through the loop?

shahzaibj:
> _TO-DO : Still need clarity on whether the calling app needs to be additionally verified by making an ESTS call?
> The reason why I did not add this check is because after we return the account which has an FRT.. we make a call to loadWithAggregatedAccountData (https://github.com/AzureAD/ad-accounts-for-android/blob/63efe400aa8d55bab4cbedb22533a72dac27d741/broker4j/src/main/com/microsoft/identity/client/BrokerLocalController.java#L165) and then retrieve AT with FRT. I think the FRT can only be used to retrieve an AT when it is part of the FOCI family. Also, when getAccounts() is called prior to ATS call.. we already would have verified if it is a Foci app or not.

@somalaya I would highly recommend that we still perform an additional verification at this layer to ensure that calling app is indeed a member of FOCI (before looking into the foci cache). For this we don't need ESTS call. I suggest you just use [`platformComponents.isAuthorizedToShareTokens(uid)` ](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/broker4j/broker/platform/components/IBrokerPlatformComponents.java#L147) - I think it is okay to move those methods from `IBrokerPlatformComponents` in `broker4j` to `IPlatformComponents` in common4j. 

cc/ @shoatman if he has additional thoughts here.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -862,11 +864,14 @@ protected AccountRecord getCachedAccountRecord(
                     );
         }
 
+        // TO-DO can we remove the parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache condition?
```
shoatman: I think as we discussed this was likely there to avoid someone configuring an AdalOAuth2TokenCache.  I'd remove this comment or update it.  You updated the overview comment to discuss this aspect a bit.  I suggest either linking to this PR here or filing a separate work item and then linking to that work item.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -862,11 +864,14 @@ protected AccountRecord getCachedAccountRecord(
...
         if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
             targetAccount = getAccountWithFRTIfAvailable(
                     parameters,
                     (MsalOAuth2TokenCache) parameters.getOAuth2TokenCache()
             );
+        } else if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof BrokerOAuth2TokenCache) {
```
shoatman: I mean implementation here doesn't check that the client is authorized to use the token in the cache if it's found.  I think we have to know whether the client performing the silent request is authorized to share tokens (is a member of the Microsoft Family).  Since the BrokerOAuth2TokenCache relies on the the broker application metadata cache to determine whether or not an application is a member of the Microsoft family... perhaps we should be checking the metadata for the current app client id/application identifier and if not present.... inserting it.

shoatman (changes requested):
I think we have to address the is authorized to share tokens.

rpdome:
Maybe..

1. make getCachedAccountRecord() invoke `getCachedAccountRecordFromCallingAppCache()` (or maybe come up with a better name 😅 ) and `getCachedAccountRecordForFoCI()`

2. In BrokerLocalController, override `getCachedAccountRecordForFoCI()`

```Java
protected AccountRecord getCachedAccountRecordForFoCI(
            @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
      // your new logic here.
      return super.getCachedAccountRecordForFoCI(parameters);
}
```

3. In your new getAccountFromFociCacheUsingLocalAccountId(), utilizes mPlatformComponents.isAuthorizedToShareTokens() 


OR In 2,3.... reuses BrokerUtil.getCacheRecordListFromBrokerCache() (this is probably better, as we don't have to maintain similar logic in multiple places).

```Java
protected AccountRecord getCachedAccountRecordForFoCI(
            @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
      // Get AccountRecord from BrokerUtil.getCacheRecordListFromBrokerCache();
}
```

somalaya:
@shahzaibj If I move the  isAuthorizedToShareTokens(uid) method to IPlatformComponents..   then I should add a dependency on TokenShare library in common..  (implementation "com.microsoft:tokenshare:$rootProject.ext.tokenShare")
LMK if this is what we want to be doing.. 
cc @shoatman

shahzaibj:
> @shahzaibj If I move the isAuthorizedToShareTokens(uid) method to IPlatformComponents.. then I should add a dependency on TokenShare library in common4j.. (implementation "com.microsoft:tokenshare:$rootProject.ext.tokenShare") LMK if this is what we want to be doing.. cc @shoatman

I think we can move the interface method to `common4j` and keep the implementation in broker. This probably means having two implementations, one for `common4j` and one for `AADAuthenticator` and it might be okay to have those different implementations with the one in the broker using the TSL and the other may be just use FociQueryUtilities or more simply just return `false`

somalaya:
But if we only have FociQueryUtilities in common's implementation.. we would end up making the ESTS call everytime?
Also, I did not understand why we have to re-use isAuthorizesToShareTokens method if we simply have to return false. Can you elaborate the last part of it?

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: Removed this function as I made changes in broker repo to override getCachedAccountRecordFromAllCaches. It is in this method, I am reusing the existing BrokerUtil.getCacheRecordListFromBrokerCache() method which internally calls mPlatformComponents.isAuthorizedToShareTokens()

somalaya:
> Maybe..
> 
> 1. make getCachedAccountRecord() invoke `getCachedAccountRecordFromCallingAppCache()` (or maybe come up with a better name 😅 ) and `getCachedAccountRecordForFoCI()`
> 2. In BrokerLocalController, override `getCachedAccountRecordForFoCI()`
> 
> ```java
> protected AccountRecord getCachedAccountRecordForFoCI(
>             @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
>       // your new logic here.
>       return super.getCachedAccountRecordForFoCI(parameters);
> }
> ```
> 
> 3. In your new getAccountFromFociCacheUsingLocalAccountId(), utilizes mPlatformComponents.isAuthorizedToShareTokens()
> 
> OR In 2,3.... reuses BrokerUtil.getCacheRecordListFromBrokerCache() (this is probably better, as we don't have to maintain similar logic in multiple places).
> 
> ```java
> protected AccountRecord getCachedAccountRecordForFoCI(
>             @NonNull final SilentTokenCommandParameters parameters) throws ClientException {
>       // Get AccountRecord from BrokerUtil.getCacheRecordListFromBrokerCache();
> }
> ```

I made these changes and now this PR only breaks the getCachedAccountRecord() method into 2 parts like you mentioned 
1. GetAccounts from app cache 
2. Get accounts from all caches (This includes foci cache) - Resued the BrokerUtil.getCacheRecordListFromBrokerCache() method in this broker PR https://github.com/AzureAD/ad-accounts-for-android/pull/1951

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: Using the same check available in BrokerUtil.getCacheRecordListFromBrokerCache() method to see if the authorized to use the token in the cache if it's found. This change is present in https://github.com/AzureAD/ad-accounts-for-android/pull/1951

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: added a work item for this here f2bb751adde2d05496e3cea965a9068022059612

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -831,6 +830,49 @@ protected AccountRecord getCachedAccountRecord(
...
+                    ErrorStrings.NO_ACCOUNT_FOUND,
+                    "No cached accounts found for the supplied "
+                            + (isB2CAuthority ? "homeAccountId" : "localAccountId")
+            );
+        }
+
+        return targetAccount;
+    }
+
+    private AccountRecord getCachedAccountRecordFromCallingAppCache(
```
rpdome, line 872: nit: javadoc.

rpdome (approved):

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -831,6 +830,49 @@ protected AccountRecord getCachedAccountRecord(
                         .getAuthorityTypeString()
         );
 
+        AccountRecord targetAccount = getCachedAccountRecordFromCallingAppCache(parameters);
+        if (targetAccount != null) {
+            return targetAccount;
+        } else {
+            Logger.info(TAG, "Account not found in app cache..");
+            targetAccount = getCachedAccountRecordFromAllCaches(parameters);
```
rpdome, line 838: Something that probably can be tweaked:

1. rename getCachedAccountRecordFromAllCaches() to getAccountRecordFromCaches()
2. In MsalOAuthTokenCache.getAccountRecordFromCaches(), invoke getCachedAccountRecordFromCallingAppCache() before proceeding with the rest of the logic.
3. In BrokeOAuthTokenCache.getAccountRecordFromCaches(), invoke BrokerUtil.getCacheRecordListFromBrokerCache()

Reasoning: BrokerUtil.getCacheRecordListFromBrokerCache() is already looking into the calling app cache.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -861,37 +903,22 @@ protected AccountRecord getCachedAccountRecord(
                             localAccountId
                     );
         }
+        return targetAccount;
+    }
 
-        if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
-            targetAccount = getAccountWithFRTIfAvailable(
+    /**
+     * Lookup in ALL the caches including the foci cache
```
shahzaibj:
```suggestion
     * Lookup in ALL the caches including the foci cache.
```

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: I think I will keep this as is because BrokeOAuthTokenCache is in common and cannot invoke BrokerUtil.getCacheRecordListFromBrokerCache() which is in broker module. 
Also, I think in getCachedAccountRecordFromAllCaches we are currently doing a lookup with localAccountId specifically whereas in BrokerUtil.getCacheRecordListFromBrokerCache() we are doing a lookup with environment and clientId. So, there is some difference in cache lookup.

shoatman (approved):

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -861,37 +903,22 @@ protected AccountRecord getCachedAccountRecord(
...
         }
+        return targetAccount;
+    }
 
-        if (null == targetAccount && parameters.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
-            targetAccount = getAccountWithFRTIfAvailable(
+    /**
+     * Lookup in ALL the caches including the foci cache.
+     */
+    protected AccountRecord getCachedAccountRecordFromAllCaches(
```
shahzaibj, line 917: probably add Nullable annotation

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
shahzaibj: Nullability annotations would also help

PR #1801, opened 2022-07-25
Add telemetry event for content provider call for getting enrollment id

iamgusain:
### What
Add telemetry event for content provider call for getting enrollment id

### Why
Need a way to measure if company portal was called to get enrollment id and how much time it took. 

### How
Added a new TelemetryEventType for ContentProviderCall, and emitting it when we make a call to Company portal for getting enrollment id

## Testing
Verified locally that the event is getting added to the event map

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ContentProviderCallEvent.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ContentProviderCallEvent.java
@@ -1,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


nickbopp (approved):

nickbopp:
@iamgusain happy to sign off when this is moved to a full review.

bktowett (approved):

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ContentProviderCallEvent.java
```diff
@@ -0,0 +1,36 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.telemetry.events;
+
+import com.microsoft.identity.common.java.telemetry.TelemetryEventStrings;
+
+import lombok.NonNull;
+
+public class ContentProviderCallEvent extends BaseEvent {
```
mohitc1, line 32: Consider adding documentation for the class.

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/ContentProviderCallEvent.java
```diff
@@ -0,0 +1,36 @@
...
+    public ContentProviderCallEvent(@NonNull final String contentUri){
```
mohitc1, line 37: Should we take a name too as parameter, as it seems this event could be used for different ContentProvider scenarios and each with different name would be helpful. e.g. this case GetEnrollmentId ?

iamgusain: content uri should be different for each cases, that should be able to differentiate.

PR #1805, opened 2022-07-29
Improving Telemetry In CBA Flows

melissaahn:
## Summary
This PR attempts to improve Telemetry for smartcard certificate-based authentication (CBA) and add some Telemetry for on-device CBA.
A new BaseEvent class `CertBasedAuthResultEvent` is dedicated for emission of CBA details, which includes when a `ClientCertRequest` is being proceeded with, what the outcome is.
A new BaseEvent class `PivProviderStatusEvent` is dedicated for tracking when a `PivProvider` instance is in the Security static list:
- Whether or not a `PivProvider` instance is already present in the `Security` static Provider list (at the stage in which we are supposed to be adding a `PivProvider` instance ourselves. 
- Whether or not an existing `PivProvider` instance is being removed from the `Security` static Provider list (at the stage in which a YubiKey is being unplugged).

For all exceptions thrown related to CBA, an ErrorEvent is emitted.

To appropriately emit a `CertBasedAuthResultEvent` when a `ClientCertRequest` is being proceeded with, I call a wrapper `emitTelemetryForCertBasedAuthResults` within `onChallengeResponseReceived`, which, from `AADWebViewClient` to `ClientCertAuthChallengeHandler`, ends up emitting an event only when CBA was the flow commenced, which is determined by flags set before `proceed` is called on `ClientCertRequest`.

## Testing
I manually made sure that the correct Telemetry statements were being reached appropriately. Some things I checked:
- When on-device CBA occurs, a `CERT_BASED_AUTH_RESULT_ON_DEVICE_EVENT` `CertBasedAuthResultEvent` should be emitted.
- When smartcard CBA occurs, a `CERT_BASED_AUTH_RESULT_SMARTCARD_EVENT` `CertBasedAuthResultEvent` should be emitted.
- When CBA is proceeded with but fails, an exception should be emitted.
- When CBA does not occur, a`CertBasedAuthResultEvent` should not be emitted.
- The `PivProviderStatusEvent`s are being emitted when interacting with the Security static list.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -477,4 +477,17 @@ public void stopYubiKitManagerUsbDiscovery() {
         }
     }
 
+    /**
+     * A wrapper to telemet results from certificate based authentication (CBA) if CBA occurred.
```
shoatman: I'm not familiar with the use of telemet... is this being used as a verb?  If so, suggest emit telemetry instead.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -477,4 +477,17 @@ public void stopYubiKitManagerUsbDiscovery() {
...
+     * @param response a RawAuthorizationResult object received upon a challenge response received.
+     */
+    public void telemetCertBasedAuthResult(@NonNull final RawAuthorizationResult response) {
+        final String methodTag = TAG + ":telemetCertBasedAuthResult";
+        if (mClientCertAuthChallengeHandler != null) {
+            mClientCertAuthChallengeHandler.telemetCertBasedAuthResults(response);
+        } else {
+            Logger.error(methodTag, "CBA results won't be telemeterized due to mClientCertAuthChallengeHandler being null", null);
```
shoatman: I'm not sure I would log this as an error.  When would this be null in practice?

shoatman (approved):
Some questions about wording/naming.  I'm not that familiar with how the telemetry events get emitted to a collector so would be good to get a review from someone more familiar with the Telemetry system.

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/CertBasedAuthEvent.java
```diff
@@ -0,0 +1,108 @@
...
+        types(TelemetryEventStrings.EventType.CERT_BASED_AUTH_EVENT);
+    }
+
+    /**
+     * Puts a Boolean that describes whether or not a PivProvider instance was already present in the Security static provider list
+     *  at the time of proceeding with smartcard CBA.
+     * @param isPresent true when a PivProvider instance was already present. false otherwise.
+     * @return The Event object.
+     */
+    public CertBasedAuthEvent putIsExistingPivProviderPresent(final boolean isPresent) {
```
rpdome: These 3 methods are specific to yubikey, so I would suggest 

1. extending this class into "YubiKeyCertBasedAuthEvent"
or 
2. create a helper method somewher (maybe inside the challenge handler, that constructs the event based on the parameters).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -165,22 +173,18 @@ public void run() {
                             synchronized (sDeviceLock) {
                                 Logger.verbose(TAG, "A YubiKey device was disconnected");
                                 mDevice = null;
+                                //Creating a CertBasedAuthEvent for Telemetry.
+                                final CertBasedAuthEvent certBasedAuthEvent = new CertBasedAuthEvent(TelemetryEventStrings.Event.CERT_BASED_AUTH_SMARTCARD_EVENT);
```
rpdome: Double check with shahzaib, but i'm wondering if we should just create a new event type specifically for this

i.e.  TelemetryEventStrings.Event.YUBIKEY_PROVIDER_STATUS_EVENT

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -626,4 +631,31 @@ public void stopYubiKitManagerUsbDiscovery() {
...
+    /**
+     * Telemet results from certificate based authentication (CBA) if CBA occurred.
+     * @param response a RawAuthorizationResult object received upon a challenge response received.
+     */
+    public void telemetCertBasedAuthResults(@NonNull final RawAuthorizationResult response) {
+        if (mIsOnDeviceCertBasedAuthProceeding || mIsSmartcardCertBasedAuthProceeding) {
+            //Telemet Results based on which type of CBA occurred.
+            final CertBasedAuthEvent certBasedAuthEvent;
+            if (mIsOnDeviceCertBasedAuthProceeding) {
+               certBasedAuthEvent =  new CertBasedAuthEvent(TelemetryEventStrings.Event.CERT_BASED_AUTH_ON_DEVICE_EVENT);
```
rpdome: nit: maybe rename to CERT_BASED_AUTH_ON_DEVICE_RESULT_EVENT and CERT_BASED_AUTH_SMARTCARD_RESULT_EVENT

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -517,12 +518,15 @@ private void useSmartcardCertForAuth(@NonNull final X509Certificate cert,
...
             mDialogHolder.dismissDialog();
+            //Set mIsSmartcardCertBasedAuthProceeding to true so a result will be telemeterized.
+            mIsSmartcardCertBasedAuthProceeding = true;
             //Call proceed on ClientCertRequest with PivPrivateKey and cert chain.
             request.proceed(pivPrivateKey, chain);
         } catch (final UnrecoverableKeyException | CertificateException | KeyStoreException | IOException | NoSuchAlgorithmException e) {
             Logger.error(methodTag, e.getMessage(), e);
             //Show general error dialog.
             mDialogHolder.showErrorDialog(R.string.smartcard_general_error_dialog_title, R.string.smartcard_general_error_dialog_message);
+            //TODO: Telemet why it was cancelled.
```
rpdome: note: this needs to be a separate event names from the one emitted in telemetCertBasedAuthResults, otherwise the code will emit an error twice (here, and when the webview requests actually fail), which could be confusing / skew the data.

rpdome (commented):
left some comments. I also suggest getting a blessing from our telemetry lord @shahzaibj .

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: replacing all the weird conjugations of telemetry... telemet, telemeted, telemeterize

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/CertBasedAuthEvent.java
```diff
@@ -0,0 +1,108 @@
...
+        put(TelemetryEventStrings.Key.ERROR_DESCRIPTION, exception.getMessage()); // pii
+        if (exception.getCause() != null) {
+            put(TelemetryEventStrings.Key.ERROR_CLASS_NAME, exception.getCause().getClass().getSimpleName());
+        }
+
+        put(TelemetryEventStrings.Key.ERROR_CODE, exception.getErrorCode());
+        put(TelemetryEventStrings.Key.SERVER_ERROR_CODE, exception.getCliTelemErrorCode());
+        put(TelemetryEventStrings.Key.SERVER_SUBERROR_CODE, exception.getCliTelemSubErrorCode());
+
+        return this;
```
shahzaibj: This is all in the ErrorEvent already right? So why duplicate the code here? It seems what you need is to rather emit an ErrorEvent

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: I _think_ it could be null if some YubiKey init method within the `ClientCertAuthChallengeHandler` constructor failed. But I haven't seen this happen yet. iirc I was asked to add logging for these types of situations (for `onReceivedClientCertRequest` and `stopYubiKeyManagerUsbDiscovery`) in a previous review.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: have to be honest, I forgot that I needed to emit telemetry for when the flow is cancelled (not proceeded yet). 
I'm considering creating a separate event name for cancelling.
Should I also be emitting telemetry for when the cancellation is not an error (ex. user hits 'cancel' button, has no certs loaded on key, is locked out from key)?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
melissaahn: Could it potentially be a new event name instead? So that it's under the event type CERT_BASED_AUTH_EVENT but the event name is specifically CERT_BAED_AUTH_YUBIKEY_PROVIDER_STATUS_EVENT? (idk maybe that's too long)

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/CertBasedAuthEvent.java
melissaahn: Based on comments earlier above, I'm wondering if it would be ok to put the `putIsExistingPivProviderPresent` and `putPivProviderRemoved` in their own event name (something like CERT_BASED_AUTH_YUBIKEY_PROVIDER_STATUS_EVENT) and then rename this class to `CertBasedAuthResultEvent` (with corresponding names CERT_BASED_AUTH_{SMARTCARD/ON_DEVICE}_RESULT).

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/CertBasedAuthEvent.java
melissaahn: I wanted to keep the response code and exception together in the same cba event, since they're coming from the same object. I guess if it's better to have separate events emitted in order to not duplicate code, that also makes sense to me. 
This would leave just one method then for this event. I'm wondering if it would be better to just emit a BaseEvent instead.

shahzaibj: Which response code specifically are you talking about?

melissaahn: The response code from RawAuthorizationResult: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/9e579b661b4c40326f61d9b5d1049c66e704c717/common4j/src/main/com/microsoft/identity/common/java/providers/RawAuthorizationResult.java#L133

The RawAuthorizationResult comes from here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/b8ea54b8ea2e37f7cc40939752a6dd582c6ae21c/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java#L280

shahzaibj: but where are you using the RawAuthorizationResult over here?

melissaahn: The exception comes from the RawAuthorizationResult as well. When the result code corresponds with an error, the RawAuthorizationResult will include an exception (the exception variable will be null otherwise)

shahzaibj: I meant over here this method just takes Exception and does nothing to RawAuthorizationResult, so I'm not sure what's the relation here. I think my overall question is around why we need this duplicated `putException` method when we have a similar method in ErrorEvent? From your initial response, it seemed like it had something to do with RawAuthorizationResult but I'm not seeing that being used here.

melissaahn: The resultCode and the exception both come from RawAuthorizationResult, so I wanted to keep the exception and the corresponding resultCode (exception serving as an explanation for a resultCode) in the same event.
If this doesn't really matter, and it would be better to have the resultCode and the exception in separate events, I'm alright to make that change, too.

shahzaibj: What question are we looking to answer with the result code? It could definitely be valuable I think but I just want us to enumerate it as a question. 

From code point of view, I think the result code should be in its own event. Something related to the Auth. may be UiEndEvent. 

I think there could be a lot of things that are currently unclear and we may end-up refactoring some of this code in the near term. So I'm not too worried about where we put it right now but we should avoid code duplication.

melissaahn: The result code should give an idea as to whether the CBA flow successfully completed or not. This could be helpful in determining the origin of unexpected errors pertaining to security providers.
I'm going to use ErrorEvents when possible and rewrite the CertBasedAuthEvent to be CertBasedAuthResultEvent (This is probably the simplest thing to do right now... I'll refactor when needed). I'll create a separate Event named PivProviderStatusEvent as well (totally ok with refactoring this later, too).
Hope this plan sounds ok for the time being,

common4j/src/main/com/microsoft/identity/common/java/telemetry/events/PivProviderStatusEvent.java
```diff
@@ -0,0 +1,40 @@
+package com.microsoft.identity.common.java.telemetry.events;
```
shahzaibj, line 23: LICENSE

shahzaibj (approved):
For the data that you are collecting, you may want to suggest that in the comments on the following PRs, so we can track it there as well:

- https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview/pullrequest/6179
- https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview/pullrequest/6305

PR #1810, opened 2022-08-04
Add a Multi Type separated store

shahzaibj:
This adds a storage interface that allows to store name-value pairs associated to a particular "Separator". This separator can be anything of type T as defined by the implementations of this interface.

**Note:** This PR doesn't introduce a new concept. The interface introduced in this PR should be considered similar to Account Manager in Android (or our [IAccountDataStorage](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/broker4j/broker/platform/components/IAccountDataStorage.java#L301-L321)) that allows to store name-value pairs separated per Account, or even [IPerUidStorage](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/LinuxBroker/src/main/com/microsoft/identity/device/broker/storage/peruid/IPerUidStorage.java) where the separator is an int. This interface does the same thing, however, instead of separating per account, it allows to separate per anything of any type. (I basically just introduced a more generic version of something that already exists in our source code). Furthermore, internal storage mechanism here can also be backed by anything i.e. this interface can be implemented by SharedPreferences on Android or any other mechanism on any platform.

This PR also provides a default implementation where the separator is of type String and also allows the internal storage mechanism to be backed by anything as determined by Platform Components. Tests were also added for this new class.

**Where are we using this storage?**
We're going to be using it for FLW / Multiple WPJ Telemetry work where we need to store some telemetry and persist it on DISK and it needs to be associated to a String separator i.e. the tenant guid of the registration. For more info refer to this PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1960
With that said, the interface added in this PR is generic and should be treated as such and we should leverage it whenever we need to store data by separation (which we do quite often).

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
@@ -46,6 +46,9 @@ import lombok.NonNull;
 public interface IPerSeparatorMultiTypeNameValueStorage<T> {
     /**
      * Gets a {@link String} value from the storage.
+     * A "null" will be returned in the following cases:
+     * - separator not found
+     * - name not found
      *
      * @param separator the key to use for storage separation
      * @param name      A name associated to the value
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
@@ -43,7 +43,7 @@ import lombok.NonNull;
  *
  * @param <T> the type of separator to use
  */
-public interface IPerSeparatorMultiTypeNameValueStore<T> {
+public interface IPerSeparatorMultiTypeNameValueStorage<T> {
     /**
      * Gets a {@link String} value from the storage.
      *
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
@@ -69,7 +69,7 @@ public interface IPerSeparatorMultiTypeNameValueStore<T> {
      * @param name      The name (key) of the long to save
      * @param value     The actual value to persist
      */
-    void putLong(@NonNull T separator, String name, long value);
+    void putLong(@NonNull T separator, @NonNull String name, long value);
 
     /**
      * Retrieves a long value previously stored.
@@ -78,7 +78,7 @@ public interface IPerSeparatorMultiTypeNameValueStore<T> {
      * @param name      The name (key) of the long to retrieve
      * @return The persisted value or 0 if one cannot be found
      */
-    long getLong(@NonNull T separator, String name);
+    long getLong(@NonNull T separator, @NonNull String name);
 
     /**
      * Returns all entries in the named resource.
@@ -118,5 +118,5 @@ public interface IPerSeparatorMultiTypeNameValueStore<T> {
      * @param keyFilter A predicate to use to evaluate the key, return true to include key value pair.
      * @return an iterator as a view on the shared preferences file.
      */
-    Iterator<Map.Entry<String, String>> getAllFilteredByKey(@NonNull T separator, Predicate<String> keyFilter);
+    Iterator<Map.Entry<String, String>> getAllFilteredByKey(@NonNull T separator, @NonNull Predicate<String> keyFilter);
 }
--- a/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
@@ -1,0 +1,122 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.interfaces;
+
+import com.microsoft.identity.common.java.util.ported.Predicate;
+
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.NonNull;
+
+/**
+ * This is a storage that allows to store name-value pairs associated to a particular "Separator".
+ * This separator can be anything of type T as defined by the implementations of this interface.
+ * <p>
+ * Functionally, this interface should be considered similar to Account Manager in Android that
+ * allows to store name-value pairs separated per Account. This interface does the same thing,
+ * however, instead of separating per account, it allows to separate per anything of any type.
+ * Furthermore, internal storage mechanism here can also be backed by anything i.e. this interface
+ * can be implemented by SharedPreferences on Android or any other mechanism on any platform.
+ *
+ * @param <T> the type of separator to use
+ */
+public interface IPerSeparatorMultiTypeNameValueStore<T> {
+    /**
+     * Gets a {@link String} value from the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     */
+    @Nullable
+    String getString(@NonNull T separator, @NonNull String name);
+
+    /**
+     * Puts a value into the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     * @param value     value to be persisted
+     */
+    void putString(@NonNull T separator, @NonNull String name, @Nullable String value);
+
+    /**
+     * Persists a long value to the named resource.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      The name (key) of the long to save
+     * @param value     The actual value to persist
+     */
+    void putLong(@NonNull T separator, String name, long value);
+
+    /**
+     * Retrieves a long value previously stored.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      The name (key) of the long to retrieve
+     * @return The persisted value or 0 if one cannot be found
+     */
+    long getLong(@NonNull T separator, String name);
+
+    /**
+     * Returns all entries in the named resource.
+     *
+     * @param separator the key to use for storage separation
+     * @return a map containing all the values associated to the supplied separator in this structure.
+     */
+    @NonNull
+    Map<String, String> getAll(@NonNull T separator);
+
+    /**
+     * Removes a value from the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     */
+    void remove(@NonNull T separator, @NonNull String name);
+
+    /**
+     * Clear all data from the storage associated to the specified separator.
+     */
+    void clear(@NonNull T separator);
+
+    /**
+     * Get all keys in this storage associated to the specified separator.
+     *
+     * @param separator the key to use for storage separation
+     */
+    @NonNull
+    Set<String> keySet(@NonNull T separator);
+
+    /**
+     * Returns an iterator on the shared preferences entries that views only those entries that
+     * the predicate evaluates to true on the key.
+     *
+     * @param separator the key to use for storage separation
+     * @param keyFilter A predicate to use to evaluate the key, return true to include key value pair.
+     * @return an iterator as a view on the shared preferences file.
+     */
+    Iterator<Map.Entry<String, String>> getAllFilteredByKey(@NonNull T separator, Predicate<String> keyFilter);
+}
--- a/common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
@@ -24,17 +24,14 @@ package com.microsoft.identity.common.java.storage;
 
 import com.microsoft.identity.common.java.cache.IMultiTypeNameValueStorage;
 import com.microsoft.identity.common.java.crypto.IKeyAccessor;
+import com.microsoft.identity.common.java.interfaces.AbstractPerSeparatorMultiTypeNameValueStorage;
 import com.microsoft.identity.common.java.interfaces.IPerSeparatorMultiTypeNameValueStorage;
 import com.microsoft.identity.common.java.interfaces.IPlatformComponents;
-import com.microsoft.identity.common.java.util.ported.Predicate;
 
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Set;
 
-import edu.umd.cs.findbugs.annotations.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.NonNull;
 
@@ -45,7 +42,7 @@ import lombok.NonNull;
  * encrypted variant as determined by {@link IPlatformComponents#getEncryptedFileStore(String, IKeyAccessor)}.
  */
 @AllArgsConstructor
-public class StringSeparatedMultiTypeNameValueStorage implements IPerSeparatorMultiTypeNameValueStorage<String> {
+public class StringSeparatedMultiTypeNameValueStorage extends AbstractPerSeparatorMultiTypeNameValueStorage<String> {
 
     @NonNull
     private final IPlatformComponents mPlatformComponents;
@@ -73,7 +70,8 @@ public class StringSeparatedMultiTypeNameValueStorage implements IPerSeparatorMu
             });
 
     @NonNull
-    private synchronized IMultiTypeNameValueStorage getStoreForSeparator(@NonNull final String separator) {
+    @Override
+    protected synchronized IMultiTypeNameValueStorage getStoreForSeparator(@NonNull final String separator) {
         return sPerTenantStorageCache.computeIfAbsent(
                 separator,
                 key -> {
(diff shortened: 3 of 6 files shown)


common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
```diff
@@ -0,0 +1,122 @@
...
+ * <p>
+ * Functionally, this interface should be considered similar to Account Manager in Android that
+ * allows to store name-value pairs separated per Account. This interface does the same thing,
+ * however, instead of separating per account, it allows to separate per anything of any type.
+ * Furthermore, internal storage mechanism here can also be backed by anything i.e. this interface
+ * can be implemented by SharedPreferences on Android or any other mechanism on any platform.
+ *
+ * @param <T> the type of separator to use
+ */
+public interface IPerSeparatorMultiTypeNameValueStore<T> {
```
rpdome: If I understand correctly, this is pretty much a Map of a Map, right?

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Gets a {@link String} value from the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     */
+    @Nullable
+    String getString(@NonNull T separator, @NonNull String name);
```
rpdome: what would be returned/thrown if the provided separator is not found? 

would it be different from the case where a matching provider is found, but the matching "name" is not.

(we probably do not want this to vary for each implementation - so we should make it clear here.).

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
```diff
@@ -0,0 +1,122 @@
...
+    void putString(@NonNull T separator, @NonNull String name, @Nullable String value);
+
+    /**
+     * Persists a long value to the named resource.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      The name (key) of the long to save
+     * @param value     The actual value to persist
+     */
+    void putLong(@NonNull T separator, String name, long value);
```
rpdome: nit: missing nullable and nonnull. here and below.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
rpdome: In other words... should the developer know the differences (such that they are able to handle each error cases differently).

rpdome: and if that matters, the alternative is to

1. Create IKeyValueStorage<U,T> as a base class of INameValueStorage<String, T>
2. Whenever we need this, we create a nested INameValueStorage<Separator, INameValueStorage<String, T>>

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
rpdome, line 46: The alternative is to
```
1. Create IKeyValueStorage<U,T> as a base class of <String> INameValueStorage<T>

2. Turn this into a class, with a builder/constructor that takes in INameValueStorage<Separator> (or IKeyValueStorage<U,Separator>). (or an abstract class, containing getStoreForSeparator() as you implemented below).


```

or make it throw certain types of exception(s).

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
```diff
@@ -0,0 +1,122 @@
...
+ * Functionally, this interface should be considered similar to Account Manager in Android that
+ * allows to store name-value pairs separated per Account. This interface does the same thing,
+ * however, instead of separating per account, it allows to separate per anything of any type.
+ * Furthermore, internal storage mechanism here can also be backed by anything i.e. this interface
+ * can be implemented by SharedPreferences on Android or any other mechanism on any platform.
+ *
+ * @param <T> the type of separator to use
+ */
+public interface IPerSeparatorMultiTypeNameValueStore<T> {
+    /**
```
rpdome, line 47: The alternative is to
```
1. Create IKeyValueStorage<U,T> as a base class of <String> INameValueStorage<T>

2. Turn this into a class, with a builder/constructor that takes in INameValueStorage<Separator> (or IKeyValueStorage<U,Separator>). (or an abstract class, containing getStoreForSeparator() as you implemented below).

```

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.NonNull;
+
+/**
+ * This is a storage that allows to store name-value pairs associated to a particular "Separator".
```
rpdome: nit: I think the term "Storage" (or something along that line) might be easier to understand (yeah... it might be confusing as we're already using the term "store")

This structure consists of multiple key-value storages.

When a dev is using this interface, they're trying to read/write a value from a certain storage.

Maybe we can call this MultiStorageNameValueStore?

common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
```diff
@@ -0,0 +1,142 @@
...
+import lombok.NonNull;
+
+/**
+ * A platform agnostic implementation of {@link IPerSeparatorMultiTypeNameValueStore} where the
+ * separator is of type {@link String}. The internal storage mechanism here could be backed by
+ * anything as determined by the {@link IPlatformComponents#getFileStore(String)} or even be an
+ * encrypted variant as determined by {@link IPlatformComponents#getEncryptedFileStore(String, IKeyAccessor)}.
+ */
+@AllArgsConstructor
+public class StringSeparatedMultiTypeNameValueStorage implements IPerSeparatorMultiTypeNameValueStore<String> {
```
rpdome: Wondering if we can turn this into a utility class by making getStoreForSeparator() abstract.

by doing that, we should not need the interface. All the implementer need to do is to define the source of storages.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
shahzaibj: Conceptually yes. Implementation however can be backed by anything. It may be a Map in Java or some other data structure called something else that can help achieve the same purpose.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
shahzaibj: Should be getting back `null`.

This whole class is actually almost the same as [IAccountDataStorage](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/broker4j/broker/platform/components/IAccountDataStorage.java#L301-L321)

that one seem to not document what would happen if `Account` isn't found but from the look of the code it seems to return back null. The class added in this PR will do the same thing.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
shahzaibj: The idea is that at the interface level the dev shouldn't care about the internal implementation details. It is just data separated by a separator and that separator can be anything. If you look at `IAccountDataStorage` - that is the same code basically except separator there is replaced with `IBrokerAccount`: https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/broker4j/broker/platform/components/IAccountDataStorage.java#L301-L321

Since that one is called `IAccountDataStorage`, I followed similar naming mechanism here i.e. ISeparatorNameValueStore or something along those lines.

I can definitely rename it from Store to Storage.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStore.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1810/commits/37963fd4f401de368069186dc0425789b97b0efb

common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
shahzaibj: IMO, most classes should be backed by an interface and with this one I think it is needed to allow someone to take it and implement in a completely different way.

I did go ahead and added an abstract implementation as you suggested but did so on top of the interface. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1810/commits/2830fc03360f3cadff042300f228c5d0ca03ca15

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
```diff
@@ -0,0 +1,122 @@
...
+ */
+public interface IPerSeparatorMultiTypeNameValueStorage<T> {
+    /**
+     * Gets a {@link String} value from the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     */
+    @Nullable
+    String getString(@NonNull T separator, @NonNull String name);
```
rpdome, line 57: nit: Let's specify that null will be returned when
1. Separator not found.
2. name not found.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Puts a value into the storage.
+     *
+     * @param separator the key to use for storage separation
+     * @param name      A name associated to the value
+     * @param value     value to be persisted
+     */
+    void putString(@NonNull T separator, @NonNull String name, @Nullable String value);
```
rpdome, line 66: nit: add final, here and everywhere.

rpdome (approved):
I think the naming kind of throw me off - had to read the comment twice to see what "Separator" actually means...

If we can come up with a more well-known term, that would be great.

iamgusain:
Where are we going to use this storage? 
Basically, can you add Why do we need this change to the PR description.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
shahzaibj: This is just interface, and we seem to not add `final` anywhere in our interfaces...so following the same pattern as our current code style. I can add it if/once we decide to change the code style.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1810/commits/1525f050ccc9c137ee7320e1a52c831e9385d077

shahzaibj:
> Where are we going to use this storage? Basically, can you add Why do we need this change to the PR description.

Done. Read last paragraph of the PR Description.

iamgusain:
I am not fully convinced that we need to make the storage too generic. 
I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.

shahzaibj:
> I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.

IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.

iamgusain:
> > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> 
> IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.

As an example, in your current PR, 
StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use

shahzaibj:
> > > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> > 
> > 
> > IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.
> 
> As an example, in your current PR, StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use

I'm not sure I understand your comment. Can you elaborate?

iamgusain:
> > > > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> > > 
> > > 
> > > IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.
> > 
> > 
> > As an example, in your current PR, StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use
> 
> I'm not sure I understand your comment. Can you elaborate?

Basically, if I am implementing a String separated storage, 
Where do I specify, where the data is going to be stored and what type of data is going to be stored in that.
Currently it is assuming that there is a PerTenantStorageCache where this data is getting stored, and also I can store any type of data to it (in KeyValue pair.. long, string etc...)

iamgusain:
> > > > > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> > > > 
> > > > 
> > > > IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.
> > > 
> > > 
> > > As an example, in your current PR, StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use
> > 
> > 
> > I'm not sure I understand your comment. Can you elaborate?
> 
> Basically, if I am implementing a String separated storage, Where do I specify, where the data is going to be stored and what type of data is going to be stored in that. Currently it is assuming that there is a PerTenantStorageCache where this data is getting stored, and also I can store any type of data to it (in KeyValue pair.. long, string etc...)

So insteand of a String Separated store.. it is more like a tenant Id Separated store

shahzaibj:
> > > > > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> > > > 
> > > > 
> > > > IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.
> > > 
> > > 
> > > As an example, in your current PR, StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use
> > 
> > 
> > I'm not sure I understand your comment. Can you elaborate?
> 
> Basically, if I am implementing a String separated storage, Where do I specify, where the data is going to be stored and what type of data is going to be stored in that. Currently it is assuming that there is a PerTenantStorageCache where this data is getting stored, and also I can store any type of data to it (in KeyValue pair.. long, string etc...)

It is NOT assuming that is is tenant. I am using tenant for my purposes only. It is any arbitrary String. The storage interface doesn't care what the String separator actually means. (Just like when you create a HashMap of String keys and values in Java...the keys and values can be anything i.e. a tenant or something else...Java doesn't care)

> Where do I specify, where the data is going to be stored and what type of data is going to be stored in that.

This is done through the Type parameter. Generics and Type Parameter is a fundamental concept in JAVA and we already use in a number of places.  I would highly recommend that you read up online on that. Here's some documentation I found quickly: https://www.geeksforgeeks.org/generics-in-java/#:~:text=Generics%20means%20parameterized%20types.,work%20with%20different%20data%20types.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPerSeparatorMultiTypeNameValueStorage.java
```diff
@@ -0,0 +1,125 @@
...
+ * <p>
+ * Functionally, this interface should be considered similar to Account Manager in Android that
+ * allows to store name-value pairs separated per Account. This interface does the same thing,
+ * however, instead of separating per account, it allows to separate per anything of any type.
+ * Furthermore, internal storage mechanism here can also be backed by anything i.e. this interface
+ * can be implemented by SharedPreferences on Android or any other mechanism on any platform.
+ *
+ * @param <T> the type of separator to use
+ */
+public interface IPerSeparatorMultiTypeNameValueStorage<T> {
```
iamgusain, line 46: We can rename this to TypeSeperatedNameValueStorage<T /\*SeperatorType\*/>

common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
```diff
@@ -0,0 +1,91 @@
...
+    private final boolean mShouldEncrypt;
+
+    private static final int MAX_ITEM_COUNT = 25;
+
+    /**
+     * In-memory cache for storing per-string storage for each String. The idea is to not have
+     * create a new instance of {@link IMultiTypeNameValueStorage} for each call to
+     * {@link IPerSeparatorMultiTypeNameValueStorage#putString(Object, String, String)}.
+     */
+    private static final Map<String, IMultiTypeNameValueStorage> sPerTenantStorageCache =
```
iamgusain: rename to sStringSeparatedStroageCache

iamgusain:
> > > > > > I am not fully convinced that we need to make the storage too generic. I would prefer to use a specific IFlwTelemetryStorage interface. And in its implementation keep it tightly coupled with where and what type of data it is going to store.
> > > > > 
> > > > > 
> > > > > IMHO, we need to move to make this things generic and give ourselves opportunity to re-use code as much as we can. I don't think we want to re-write this logic again a fourth time (this is the third time we have written something like this). Read PR description and I have given examples of other places where we added similar code in the past.
> > > > 
> > > > 
> > > > As an example, in your current PR, StringSeparatedMultiTypeNameValueStorage assumes that it is stored in sPerTenantStorageCache, which might not be true for all StringSeparatedMultiTypeNameValueStorage stroages that we want to use
> > > 
> > > 
> > > I'm not sure I understand your comment. Can you elaborate?
> > 
> > 
> > Basically, if I am implementing a String separated storage, Where do I specify, where the data is going to be stored and what type of data is going to be stored in that. Currently it is assuming that there is a PerTenantStorageCache where this data is getting stored, and also I can store any type of data to it (in KeyValue pair.. long, string etc...)
> 
> It is NOT assuming that is is tenant. I am using tenant for my purposes only. It is any arbitrary String. The storage interface doesn't care what the String separator actually means. (Just like when you create a HashMap of String keys and values in Java...the keys and values can be anything i.e. a tenant or something else...Java doesn't care)
> 
> > Where do I specify, where the data is going to be stored and what type of data is going to be stored in that.
> 
> This is done through the Type parameter. Generics and Type Parameter is a fundamental concept in JAVA and we already use in a number of places. I would highly recommend that you read up online on that. Here's some documentation I found quickly: https://www.geeksforgeeks.org/generics-in-java/#:~:text=Generics%20means%20parameterized%20types.,work%20with%20different%20data%20types.

I am approving the PR for now, 
But would like for us to revisit this again later, to define the way we want to expose the storage in our code, having multiple ways to do the same thing can lead to confusions for the developers down the road, as-in which way to pick.

iamgusain (approved):
Approved with suggestions

PR #1815, opened 2022-08-12
Added more precautions when handling PIN

melissaahn:
## Summary
It's generally not great to use Strings to store passwords/pins because they are immutable (and for other reasons [explained here](https://www.geeksforgeeks.org/use-char-array-string-storing-passwords-java/)). 

I removed logic that sets the PIN grabbed from the EditText to a string, replacing it with logic that turns it into a char array. The parameter declarations in which a string was passed were changed to char array. 

I also added a small method, `clearPin`, which sets the characters in the provided char array to 0. I call it in the success scenario after the PivPrivateKey is created (the constructor of PivPrivateKey creates a copy of the char array, which I have no control over) and in the failure scenario when an incorrect PIN attempt is made.

Of course, these changes aren't necessarily going to prevent an experienced hacker from finding the PIN somehow, but it does reduce the area of code in which the PIN is present from our side.

## Testing
I stepped through the code to make sure that the PIN array was being cleared without affecting the existing functionalities.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
@@ -75,6 +75,7 @@ import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.Callable;
 
@@ -406,7 +407,7 @@ public final class ClientCertAuthChallengeHandler implements IChallengeHandler<C
         return new SmartcardPinDialog.PositiveButtonListener() {
             @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
             @Override
-            public void onClick(@NonNull final String pin) {
+            public void onClick(@NonNull final char[] pin) {
                 synchronized (sDeviceLock) {
                     // Need to request a PivSession in order to access certs on YubiKey.
                     getActivePivSessionAsync(request, new IPivSessionCallback() {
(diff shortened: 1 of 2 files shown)


fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -454,6 +457,8 @@ private void verifySmartcardPin(@NonNull final char[] pin, @NonNull final YubiKi
             useSmartcardCertForAuth(certDetails.getCertificate(), pin, certDetails.getSlot().getStringAlias(), piv, request);
         } catch (final InvalidPinException e) {
             // An incorrect Pin attempt.
+            //Clear PIN attempt from local char array.
+            clearPin(pin);
```
rpdome: nit: maybe rename this to "tryUsingSmartcardWithPin" or something of sorts.
(because it's not just "verify smartcard pin", it's also doing smartcard auth here.)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/ClientCertAuthChallengeHandler.java
```diff
@@ -406,22 +407,24 @@ private SmartcardPinDialog.PositiveButtonListener getSmartcardPinDialogPositiveB
...
+                                verifySmartcardPin(pin, certDetails, request, piv);
                             } catch (final IOException | ApduException | BadResponseException e) {
                                 Logger.error(methodTag, e.getMessage(), e);
                                 //Show general error dialog.
                                 mDialogHolder.showErrorDialog(R.string.smartcard_general_error_dialog_title, R.string.smartcard_general_error_dialog_message);
                                 //Emit telemetry for this exception.
                                 Telemetry.emit(new ErrorEvent().putException(e));
                                 request.cancel();
+                                //Clear PIN attempt from local char array.
+                                clearPin(pin);
```
rpdome, line 428: put this in the final {} block here, and you won't have to put it anywhere else.

rpdome (approved):

PR #1823, opened 2022-08-22
Add implementation for clearAll method to BrokerOAuth2TokenCache to clear all the credentials and metadata

iamgusain:
### What
Adding implemetation for clearAll method in BrokerOAuth2TokenCache to clear all credentials and metadata

### Why
This method will be called from Shared Device Mode SignOut flow. Currently we are removing each credential separately during SDM sign-out, which is not efficient and also leads to not be able to clean up all FOCI credentials if multiple FOCI apps credentials are present.

### How
Cleaning up individual app caches for all apps in the metadata cache, Foci cache and the app metadata cache in the clearAll method implementation

### Testing
Corresponding Broker PR where this method is used in SDM signOut Flow (https://github.com/AzureAD/ad-accounts-for-android/pull/1974). 
Also verified that this change fixes the bug https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1980471/
The issue in the bug was that current SignOut form SDM wasn't clearing the FOCI cache properly, it was only clearing the token credentials corresponding to only one client-id, and token credentials corresponding to other client-ids were not cleared.
This led to Edge receiving multiple accounts entry later as it uses multiple client ids (an issue to follow up with Edge). 
After this change we clear all entries in FOCI cache and on the next launch only one account is returned to Edge and it is able to perform SSO.

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
@@ -1174,4 +1174,61 @@ public class BrokerOAuth2TokenCacheTest {
                 retrievedResult.getRefreshToken()
         );
     }
+
+    @Test
+    public void testClearAll() throws ClientException {
+        int appIndex = 0;
+        for (final OAuth2TokenCache cache : mOtherAppTokenCaches) {
+            configureMocks(mOtherCacheTestBundles.get(appIndex));
+
+            final ICacheRecord cacheRecord = cache.save(mockStrategy, mockRequest, mockResponse);
+
+            final BrokerApplicationMetadata applicationMetadata = new BrokerApplicationMetadata();
+            applicationMetadata.setClientId(cacheRecord.getIdToken().getClientId());
+            applicationMetadata.setEnvironment(cacheRecord.getIdToken().getEnvironment());
+            applicationMetadata.setFoci(cacheRecord.getRefreshToken().getFamilyId());
+            applicationMetadata.setUid(testAppUids[appIndex++]);
+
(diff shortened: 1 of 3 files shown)


common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
```diff
@@ -1174,4 +1174,61 @@ public void testWPJSaveFoci() throws ClientException {
...
+
+        for( final String clientId :clientIds) {
+            assertEquals(true, mBrokerOAuth2TokenCache.isClientIdKnownToCache(clientId));
+        }
+
+        // Clear Broker Cache
+        mBrokerOAuth2TokenCache.clearAll();
+
+        // Verify Broker cache is cleared
+        assertEquals(0, mBrokerOAuth2TokenCache.getAccounts().size());
```
rpdome, line 1227: If we supply in-memory storage via MockPlatformComponents (getEncryptedNameValueStore), we could iterate through them to see if the data is written, and is 100% wiped.

rpdome: (and IBrokerApplicationMetadataCache can be mocked as well).

rpdome (approved):

shahzaibj (approved):

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
iamgusain: We are currently checking for Foci and metadata cache being completely wiped,

PR #1824, opened 2022-08-23
Swallow unbound service exceptions on disconnect

nickbopp:
Errors during cleanup of a broker call were surfacing rather than the results of the operation itself.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
@@ -167,7 +167,7 @@ public abstract class BoundServiceClient<T extends IInterface> {
         if (mHasStartedBinding) {
             try {
                 mContext.unbindService(mConnection);
-            } catch (IllegalArgumentException e) {
+            } catch (final IllegalArgumentException e) {
                 final String errorDescription = "Error occurred while unbinding bound Service with " + getClass().getSimpleName();
                 Logger.error(methodTag, errorDescription, e);
             }
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
@@ -163,8 +163,14 @@ public abstract class BoundServiceClient<T extends IInterface> {
      * Disconnects (unbinds) from the service.
      */
     public void disconnect() {
+        final String methodTag = TAG + ":disconnect";
         if (mHasStartedBinding) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
```diff
@@ -163,8 +163,14 @@ public BoundServiceClient(@NonNull final Context context,
      * Disconnects (unbinds) from the service.
      */
     public void disconnect() {
+        final String methodTag = TAG + ":disconnect";
         if (mHasStartedBinding) {
-            mContext.unbindService(mConnection);
+            try {
+                mContext.unbindService(mConnection);
+            } catch (IllegalArgumentException e) {
```
iamgusain: nit: final

iamgusain (approved):
Can you also add an entry to the changelog file.
Thanks

common/src/main/java/com/microsoft/identity/common/internal/broker/BoundServiceClient.java
```diff
@@ -163,8 +163,14 @@ public BoundServiceClient(@NonNull final Context context,
      * Disconnects (unbinds) from the service.
      */
     public void disconnect() {
+        final String methodTag = TAG + ":disconnect";
         if (mHasStartedBinding) {
-            mContext.unbindService(mConnection);
+            try {
+                mContext.unbindService(mConnection);
+            } catch (final IllegalArgumentException e) {
```
rpdome, line 170: May I ask what was the cause of this error? why IllegalArgumentException specifically?

nickbopp: We only have the callstacks rather than the exception details, but it's one of the three coming from here, relating to trying to bind things that are unbound already:
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/LoadedApk.java#1945

As this is a cleanup function, there isn't much of an action to take here.  It wasn't clear to me how we might be getting in one of these three states -- perhaps registration is failing early enough that unbinding isn't actually necessary?

rpdome: Could you please add the context (as much as possible) as a comment in this error block?  (i.e. OneAuth's error stats, etc) - so that future devs are aware what this error is from - and don't accidentally remove it.

It could be that the connection was terminated mid-way (by android itself) and the attempt to unbind therefore failed.

rpdome (approved):
Please add context to the block as suggested, otherwise LGTM.

PR #1825, opened 2022-08-26
Bundling multiple credential saves in a single SharedPreferences transaction

nickbopp:
The abstractions on top of SharedPreferences currently only allow a single key-value pair to be edited, each resulting in a locking/synchronous copying of the in-memory map, followed by a scheduled asynchronous disk write for the transaction to be committed back to disk.

This change gives the abstraction support for editing multiple key-value pairs and comitting in a single transaction.  We expect this to improve performance, especially in cases where we could be writing up to four credentials serially.  These each resulted in deep copies of the entirety of the SharedPreferences in-memory representation.  For apps popping an Activity after authentication is complete, Activity.onPause() also synchronously waits for all the background writes to be finished.  Waiting for one to finish instead of four should be faster.

One other notable change her was removing the LRU cache and lock.  This layer is not needed as the implementation of SharedPreferences in Android is itself keeping the totality of data in a Map<String, Object> at all times.  That layer is itself threadsafe and supports transactions, so it can be safely read directly.

I'm still measuring and testing here.  If time allows, I'm going to try to funnel some performance/timing information back from these code paths to the calling library/application.  We need this to help further understand where slow StorageRead/Write times are coming from for OneAuth customer applications if this change does not get to parity with ADAL.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
+++ b/common/src/main/java/com/microsoft/identity/common/AndroidPlatformComponents.java
@@ -295,7 +295,7 @@ public class AndroidPlatformComponents implements IPlatformComponents {
                 SharedPreferences.Editor editor = sharedPreferences.edit();
 
                 for (Map.Entry<String, Long> keyValuePair : keyValuePairs.entrySet()) {
-                    editor.putLong(keyValuePair.getKey(), keyValuePair.getValue());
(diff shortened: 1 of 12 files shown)


nickbopp:
Local testing shows CPU bound activity, and in particular the LRU cache saving the unecrypted values is critical to speeding up perf.  Continuing the investigation but no one needs to look at this PR yet.

PR #1826, opened 2022-08-26
Add an interface to Broker Token Command Parameters

shahzaibj:
I've added an interface to broker token command parameters. It was strange they didn't already have one because there were lots of similar functionality in both of them. 

Furthermore, I've added an additional field here on such command parameters that is needed for FLW telemetry purposes.

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1978

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/IBrokerTokenCommandParameters.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/IBrokerTokenCommandParameters.java
@@ -81,8 +81,19 @@ public interface IBrokerTokenCommandParameters {
      */
     String getLocalAccountId();
 
+    /**
+     * Get the {@link BrokerRequestType} for the request.
+     *
+     * @return the {@link BrokerRequestType}
+     */
     BrokerRequestType getRequestType();
 
+    /**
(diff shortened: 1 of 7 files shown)


common4j/src/main/com/microsoft/identity/common/java/commands/parameters/IBrokerTokenCommandParameters.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    @Nullable
+    String getTenantIdRequestingBrt();
+
+    /**
+     * Helper method to identify if the request originated from Broker itself or from client libraries.
+     *
+     * @return : true if request is the request is originated from Broker, false otherwise
+     */
+    default boolean isRequestFromBroker() {
```
rpdome, line 105: This keyword is new to me 😆

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/IBrokerTokenCommandParameters.java
shahzaibj: We used to have these as errors some time ago but now we are in Java 8 and can use default interface methods. This is the power of Java 8.

melissaahn (approved):

PR #1831, opened 2022-08-31
[pipeline] update gradle commands for oneauth build and test

iamgusain:
### What
Updating gradle tasks to run for OneAuth build and test as part of consumers of common pipeline

### Why
Need to update the gradle tasks to run from oneauth project to be able to cosnume local common. The previous composite build (includeBuild) does not work as expected.

### Testing
Verified running the command locally to ensure that it is able to build local common and oneauth


tanmaymanolkar1 (approved):
Thanks for correcting the commands @iamgusain

p3dr0rv (approved):

fadidurah (approved):

PR #1834, opened 2022-09-07
Fix Build Issue that arose from targeting sdk 31 

wainaina:
Common Build & Test pipeline only ran successfully after the upgrade after using Java 11.

Initially I was getting this exception while running the assemble tasks that used java 8
image

The Official documentation does not say much about this bug but on searching for other instances of the bug this is predominantly what I found 

https://developercommunity.visualstudio.com/t/some-android-apps-require-java-11-as-sdk/1549184

Upgrading to Java 11 fixed the build issues.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Use Java 11 to accomodate android sdk 31 tooling. (#1832)
 - [PATCH] Swallow unbound service exceptions on disconnect. (#1824)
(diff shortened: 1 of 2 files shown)


fadidurah (approved):

tanmaymanolkar1 (approved):

changelog.txt
```diff
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Use Java 11 to accomodate android sdk 31 tooling. (#1832)
```
shahzaibj, line 3: This is not a change that's going to be released.

PR #1836, opened 2022-09-07
Logging performance speedup

nickbopp:
Logging (almost exclusively verbose, which is turned off) consumes ~20-25% of CPU time in storageWrite calls for OneAuth.  From a profiling run, I see about 2/3 of the time is spent creating SimpleDateFormats and using them, and about 1/3 is spent putting verbose lines that are going to be filtered away onto the Executor.  The new logic filters as early as possible by loglevel, and if the check is passed, saves an Instant, which is very fast.  It's the job of the executor to string format that Instant, but only one formatter is required, rather than being recreated in response to every log line.  This should save time on the logging thread as well.

The one drawback here is that we need to remove the discardedLog override.  I don't see any consumers using this, only tests.  However, I think it's a good tradeoff to get back ~15% of storageWrite time for lack of a discardedLog callback function.  This should also offer improvements throughout API calls in android-common and its consumers.  We haven't completed a profiling run on the broker yet, but we expect that to be signficantly improved as well.

Log messages from test before change:
```
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:11:11 - thread_name: Thread-3, correlation_id: CORRELATIONID_2 - ] N/A
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: CORRELATIONID_1 - ] N/A
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: NEW_CORRELATION_ID - ] N/A
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:11:11 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
```

Log messages from after change:
```
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:04:18 - thread_name: Thread-3, correlation_id: CORRELATIONID_2 - ] N/A
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: CORRELATIONID_1 - ] N/A
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: NEW_CORRELATION_ID - ] N/A
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
[2022-09-08 19:04:18 - thread_name: Time-limited test, correlation_id: correlationId - ] Message
```


nickbopp:
Any suggestions for how to deal with the SpotBugs error?  It seems to be complaining about the empty catch block inside of Logger.java.  However, that's always been there, so I think this is just a case of a hash of the bug instance changing.  Is there a baseline file of some kind that needs to be updated?  I couldn't find one.

iamgusain (approved):
Thanks for doing this @nickbopp 

Could you add example loglines with your changes to the PR description.
Also for the test, you can add the check that the result_* doesnt contain discarded logs.

mohitc1 (approved):
:shipit:

PR #1837, opened 2022-09-08
Add opt-in flag to skip expensive SecretKey invalidation steps

nickbopp:
20-25% of CPU time in the storageWrite path for OneAuth is spent doing the very expensive `AndroidKeyStoreUtil.canLoadKey` check down to the hardware to then invalidate the in-memory copy of the key and generate a new one.  This change currently removes the check altogether, although if useful it could potentially be added more explicitly once per API call from the user for a smaller penalty.

I expect significant improvements in all paths that touch encryption/decryption, particularly in a loop like `getAllAccounts`/`getAllCredentials`.  The time spent checking the hardware for this key (and to a smaller extent, the filesystem for the on-disk file) is significantly more than is spent actually doing encryption/decryption operations with it.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
@@ -65,6 +65,11 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
     private static final String TAG = AndroidWrappedKeyLoader.class.getSimpleName() + "#";
 
     /**
+     * Should KeyStore and key file check for validity before every key load be skipped.
+     */
+    public static boolean sSkipKeyInvalidationCheck = false;
+
+    /**
      * Alias for this type of key.
      */
     /* package */ static final String KEYSTORE_KEY_ALIAS = "KEYSTORE_KEY";
@@ -99,7 +104,8 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
     private final CachedData<SecretKey> mKeyCache = new CachedData<SecretKey>() {
         @Override
         public SecretKey getData() {
-            if (!AndroidKeyStoreUtil.canLoadKey(mAlias) || !getKeyFile().exists()) {
+            if (!sSkipKeyInvalidationCheck &&
+                    (!AndroidKeyStoreUtil.canLoadKey(mAlias) || !getKeyFile().exists())) {
                 this.clear();
             }
             return super.getData();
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -99,9 +99,6 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
     private final CachedData<SecretKey> mKeyCache = new CachedData<SecretKey>() {
         @Override
         public SecretKey getData() {
-            if (!AndroidKeyStoreUtil.canLoadKey(mAlias) || !getKeyFile().exists()) {
```
iamgusain, line 102: Should we add this check and invalidation logic to where we see decrypt failures in getting the wrapped key.
Thoughts? 
cc: @rpdome

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -99,9 +99,6 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
     private final CachedData<SecretKey> mKeyCache = new CachedData<SecretKey>() {
         @Override
         public SecretKey getData() {
```
shoatman, line 108: Is the override needed anymore here?

nickbopp: The class is marked as abstract, so I think this is still needed.

nickbopp: At least something to make this a new class is instantiatable.  I guess this method in particular is not required.

nickbopp: I learned something about Java here -- even without the override, the braces make this an anonymous class.  And since none of the individual methods are abstract, this is enough to be constructible.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
nickbopp: If we hit a decryption failure, then I would expect the key that's in memory (and was either loaded from persisted assets or generated fresh and then persisted this session) already doesn't match the one that got used on disk.  I think this would delete the persisted assets for the different, new key that isn't able to decrypt the old stuff, and then we might just wind up in the same cycle again.  What do you think?

rpdome (approved):
Thanks Nick!

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
nickbopp: Closing as this one isn't relevant anymore.

gitaumoses4 (approved):

iamgusain (approved):

PR #1840, opened 2022-09-12
Common work to get 2 MSAL Automation pipelines to green

fadidurah:
Pipelines: 
https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1743&_a=summary
https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1742&_a=summary

Looks like a recent change to the consent page resource ids broke our automation, this PR includes a fix for the id as well as reliability fixes as part of a bigger automation pipeline repair effort. There is also a change to DevicePinSetupRule (looks like pin removal is currently broken, removing that part of the rule. Also adding a condition to have a pin setup when using authenticator, not having pin with authenticator was previously leading device registration to fail at cert installation.

There were some transient errors that were causing issues in the tests, those were resolved and the pipelines above have been all-green since the last change made in this PR. Will need to keep monitoring the pipelines are seeing if they are still green.


wainaina (approved):
LGTM

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
```diff
@@ -128,6 +128,12 @@ public void signOutFromSingleAccountFragment() {
      */
     public void confirmSignedIn(@NonNull final String username) {
         Logger.i(TAG, "Confirming account with supplied username is signed in..");
+        try {
+            Thread.sleep(TimeUnit.SECONDS.toMillis(5));
```
shahzaibj, line 132: Looks like a huge sleep. I'm wondering what changed here that caused us to from zero sleep to 5 seconds of sleep here?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -201,8 +201,8 @@ private void performDeviceRegistrationHelper(@NonNull final String username) {
     private void postJoinConfirmHelper(@NonNull final String expectedUpn) throws InterruptedException {
         Logger.i(TAG, "Confirming that Shared Device Registration is successfull or not..");
 
-        // Adding brief wait to increase reliability of test
-        Thread.sleep(TimeUnit.SECONDS.toMillis(5));
+        // Adding wait to increase reliability of test
+        Thread.sleep(TimeUnit.SECONDS.toMillis(15));
```
shahzaibj: Quite a big wait again. I think this is fine for now if it helps make the tests reliable, but we should revisit these techniques in the future and explore alternatives that may offer better performance without compromising on reliability.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AadLoginComponentHandler.java
```diff
@@ -97,7 +100,7 @@ public void handleAccountPicker(@NonNull final String username) {
     }
 
     private UiObject getConsentScreen() {
-        return UiAutomatorUtils.obtainUiObjectWithResourceId("consentHeader");
+        return UiAutomatorUtils.obtainUiObjectWithResourceId("appDomainLinkToAppInfo");
```
shahzaibj, line 103: This is a strange change, but if it works then we can take it.

shahzaibj (approved):
Left comments but approving this for now.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AadLoginComponentHandler.java
fadidurah: I'm thinking of proposing a change to how we handle resource-ids, basically create a way to fall back on multiple resource ids incase the first doesn't work. don't know when I'll get to that but I need to ideate a little more on it.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
fadidurah: I'm not sure how else you could handle something like this, unless we stopped relying on ui to confirm these things and perhaps found a way to do it programatically.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
fadidurah: Some test cases would fail because the signed-in username was not updated in time. We could also solve this by introducing a RetryOnFailure, but then we might rerun the whole test instead of doing the wait here.

fadidurah: We could use waitForExists() instead, actually. Might do a later change for this.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
fadidurah: We could think about using waitForExists() when applicable, that may reduce the need for some of this.

fadidurah: Changed to waitForExists()

PR #1843, opened 2022-09-12
Add BrokerContentProvider path and IpcStrategy for new device registration API

p3dr0rv:
# What is this change.
* Added a new path for device registration protocol operations, this path will be used to route all the protocols of the new device registration API when using the ContentProviderStrategy 
* Added a new type of IIpcStrategy, this strategy is the fallback strategy for the new API in case ContentProvider fails. 

# Related PR
https://github.com/AzureAD/ad-accounts-for-android/pull/1987


rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1518,6 +1519,11 @@ public enum API {
          */
         public static final String GET_SSO_TOKEN_PATH = "/ssoToken";
 
+        /**
+         * Broker api path constant for execute device registration protocols.
+         */
+        public static final String DEVICE_REGISTRATION_PROTOCOLS_PATH = "/deviceRegistration/protocols";
```
shahzaibj, line 1525: So every protocol will go through this path?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IIpcStrategy.java
```diff
@@ -16,7 +16,8 @@ enum Type {
         BOUND_SERVICE("bound_service"),
         ACCOUNT_MANAGER_ADD_ACCOUNT("account_manager_add_account"),
         CONTENT_PROVIDER("content_provider"),
-        LEGACY_ACCOUNT_MANAGER_FOR_BROKER_API("legacy_account_manager_for_broker_api");
+        LEGACY_ACCOUNT_MANAGER_FOR_BROKER_API("legacy_account_manager_for_broker_api"),
+        LEGACY_ACCOUNT_AUTHENTICATOR_FOR_WPJ_API("legacy_account_authenticator_for_wpj_api");
```
shahzaibj, line 20: What is the difference between account manager and account authenticator?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
p3dr0rv: yep

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IIpcStrategy.java
p3dr0rv: legacy_account_manager_for_broker_api is for the LegacyAccountManagerForBrokerApiStrategy class
"Old mechanism to communicate to Broker for operation like: API hello, get broker accounts, and remove accounts"
legacy_account_authenticator_for_wpj_api will be used as a backup strategy for WPJ operations in case the main strategy (Broker content provider) is not available

somalaya (approved):

PR #1844, opened 2022-09-15
Add ICryptoFactory to common4j

rpdome:
Part of https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6170

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/DefaultCryptoFactory.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/DefaultCryptoFactory.java
@@ -1,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


iamgusain (approved):

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/crypto/DefaultCryptoFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+import javax.crypto.Cipher;
+import javax.crypto.Mac;
+
+import lombok.NonNull;
+
+/**
+ * A Factory class that generates/utilizes platform's default crypto objects.
+ */
+public class DefaultCryptoFactory implements ICryptoFactory {
```
iamgusain, line 39: Please add UTs for this class

fadidurah: +1

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/crypto/DefaultCryptoFactory.java
rpdome: @iamgusain 

due to the nature of this class, it's going to be tricky to cover it with UT. (i.e. what algorithms are we going to include as part of the test - every single algorithm exists? algorithms that we use in common/MSAL/broker now? or algorithm that is in scope for FIPS only? (some of the algos are not available in the default provider), How are we going to make sure that the test is updated when someone introduces a new algo randomly in the code?)

thoughts?

rpdome: Alternatively, I can get rid of this class, and in common4j, fall back to the behavior of "If no ICryptoFactory is not provided, always fall back to the default implementation - instantiates the object directly from global provider (which is... essentially.. what this class is doing)"

rpdome: (Going to merge this for now to simply my PR tree... (I have a couple more to send out 🤣 ) but please do tag/ping me when you response, so I can address this later on).

PR #1846, opened 2022-09-19
Expose sign() method in PKeyAuth as opposed to the key itself

rpdome:
This will allow the actual signing operation to happen inside the broker - and we'll use ICryptoFactory to generate a provider, essentially allowing us to switch between non-FIPS and FIPS (tbd) mode in PKeyAuth flow.

related: https://github.com/AzureAD/ad-accounts-for-android/pull/2001

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/challengehandlers/IDeviceCertificate.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/challengehandlers/IDeviceCertificate.java
@@ -22,12 +22,12 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.challengehandlers;
 
-import java.security.PrivateKey;
-import java.security.PublicKey;
+import com.microsoft.identity.common.java.exception.ClientException;
+
 import java.security.cert.X509Certificate;
-import java.security.interfaces.RSAPrivateKey;
-import java.security.interfaces.RSAPublicKey;
 import java.util.List;
+
+import lombok.NonNull;
 
 /**
  * Work place join related certificate is required to respond device challenge.
@@ -47,21 +47,8 @@ public interface IDeviceCertificate {
      *
      * @return {@link X509Certificate}
      */
-    X509Certificate getCertificate();
-
-    /**
-     * Gets a private key.
-     *
-     * @return private key
-     */
-    PrivateKey getPrivateKey();
-
-    /**
-     * Gets a public key.
-     *
-     * @return RSA public key.
-     */
-    PublicKey getPublicKey();
+    @NonNull
+    X509Certificate getX509();
 
     /**
      * Gets thumbPrint for certificate.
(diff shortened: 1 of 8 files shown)


common4j/src/main/com/microsoft/identity/common/java/challengehandlers/IDeviceCertificate.java
```diff
@@ -47,27 +47,22 @@ public interface IDeviceCertificate {
...
     /**
-     * Gets thumbPrint for certificate.
+     * Signs a piece of data with the (private key associated to the) certificate.
      *
-     * @return thumbPrint for certificate.
+     * @param algorithm         algorithm for signing the data.
+     * @param dataToBeSigned    the data to be signed.
      */
-    String getThumbPrint();
+    byte[] sign(@NonNull final String algorithm, final byte[] dataToBeSigned) throws ClientException;
```
somalaya, line 66: can the dataToBeSigned be null?

common4j/src/main/com/microsoft/identity/common/java/util/JWSBuilder.java
```diff
@@ -121,25 +124,19 @@ public String generateSignedJWT(String nonce, String audience, PrivateKey privat
...
-        if (StringUtil.isNullOrEmpty(nonce)) {
-            throw new IllegalArgumentException("nonce");
-        }
-        if (StringUtil.isNullOrEmpty(audience)) {
-            throw new IllegalArgumentException("audience");
-        }
-        if (privateKey == null) {
-            throw new IllegalArgumentException("privateKey");
+        final String methodTag = TAG + ":generateSignedJWT";
+        if (nonce.length() == 0) {
```
somalaya: can be replaced with if (nonce.isEmpty())

common4j/src/main/com/microsoft/identity/common/java/challengehandlers/IDeviceCertificate.java
rpdome: it should not be, but we cannot add NonNull annotation to a byte array (it will be ignored by lombok).

common4j/src/main/com/microsoft/identity/common/java/util/JWSBuilder.java
```diff
@@ -121,25 +124,19 @@ public String generateSignedJWT(String nonce, String audience, PrivateKey privat
...
-        Gson gson = new Gson();
-        Claims claims = new Claims();
+        final Gson gson = new Gson();
+        final Claims claims = new Claims();
         claims.mNonce = nonce;
         claims.mAudience = audience;
-        claims.mIssueAt = System.currentTimeMillis() / SECONDS_MS;
+        claims.mIssueAt = getCurrentTimeInSeconds();
 
         JwsHeader header = new JwsHeader();
```
p3dr0rv: final

p3dr0rv (approved):
LGTM :shipit:

somalaya (approved):

PR #1850, opened 2022-09-22
Release Automation Pipeline Work

fadidurah:
Work included in this PR is done to improve the pass rate and consistency of the Broker-CI Release Pipeline:
https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1490&_a=summary

Common change list:
- Implement Socket Retry Functionality on getAccessToken() (Still seeing some error for socket closed, but looks less frequent)
- Implement waitForExist() when useful (instead of just sleeping)
- Added check for Microsoft 365 subscription page
- CP has a problem where ui elements on password page are not visible (this does not happen with login page), adding some things to make that less of a problem (turn off battery optimization, relaunch app when we get a specific exception for not finding password field)
- We now always have pin set on test device, so added check for app lock popup on authenticator
- Some error message updates

Related PRs:
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1697
https://github.com/AzureAD/ad-accounts-for-android/pull/2006


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -198,42 +201,15 @@ public void enrollDevice(@NonNull final String username,
...
-                    .loginHint(null)
-                    .build();
-
-            final MicrosoftStsPromptHandler microsoftStsPromptHandler = new MicrosoftStsPromptHandler(promptHandlerParameters);
-
-            Logger.i(TAG, "Handle prompt in AAD login page for enrolling device..");
-            // handle AAD login page
-            microsoftStsPromptHandler.handlePrompt(username, password);
-        }
+        // Company portal password page is somewhat inconsistent, found out turning off battery optimization helps from testing
```
shahzaibj: interesting

fadidurah: CP password page ui is very hit or miss. Not sure why, but we always seem to successfully catch username page, but password page elements are not visible sometimes (quite often honestly). Checking them in uiautomatorviewer when this happens, they don't show up in the node tree view. Battery optimization being off seems to make the problem less common.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/WordApp.java
```diff
@@ -135,6 +134,18 @@ private void signIn(@NonNull final String username,
     @Override
     public void confirmAccount(@NonNull final String username) {
         Logger.i(TAG, "Confirming account with supplied username is signed in..");
+
+        // Had a screen for microsoft 365 pop up occasionally
+        final UiObject msft365Object = UiAutomatorUtils.obtainUiObjectWithText("Go Premium with Microsoft 365 Personal");
+        if (msft365Object.exists()) {
```
shahzaibj, line 140: Do we know how much this waits for existence?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/WordApp.java
```diff
@@ -135,6 +134,18 @@ private void signIn(@NonNull final String username,
...
+            final UiObject skipObject = UiAutomatorUtils.obtainUiObjectWithText("SKIP FOR NOW");
+            try {
+                skipObject.click();
+            } catch (UiObjectNotFoundException e) {
+                // Nothing, just want to ignore this
```
shahzaibj: Do we want to log something?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -64,8 +64,12 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
     public final static String COMPANY_PORTAL_APP_NAME = "Intune Company Portal";
     public final static String COMPANY_PORTAL_APK = "CompanyPortal.apk";
     public final static String OLD_COMPANY_PORTAL_APK = "OldCompanyPortal.apk";
+    private final static int PASSWORD_UI_ATTEMPT_COUNT = 3;
+
+    final static long ENROLLMENT_PAGE_TIMEOUT = TimeUnit.SECONDS.toMillis(45);
```
shahzaibj: Can you add some comments on what exactly on the page we're waiting on?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -263,7 +239,7 @@ public void enrollDevice(@NonNull final String username,
                 "com.microsoft.windowsintune.companyportal:id/setup_title"
         );
 
-        if (!setupCompletePage.exists()) {
+        if (!setupCompletePage.waitForExists(ENROLLMENT_PAGE_TIMEOUT)) {
```
shahzaibj: Why does this need a different timeout from the rest of the code that seems to use FIND_UI_ELEMENT_TIMEOUT_LONG?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -64,8 +64,12 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
     public final static String COMPANY_PORTAL_APP_NAME = "Intune Company Portal";
     public final static String COMPANY_PORTAL_APK = "CompanyPortal.apk";
     public final static String OLD_COMPANY_PORTAL_APK = "OldCompanyPortal.apk";
+    private final static int PASSWORD_UI_ATTEMPT_COUNT = 3;
+
+    final static long ENROLLMENT_PAGE_TIMEOUT = TimeUnit.SECONDS.toMillis(45);
 
     private boolean enrollmentPerformedSuccessfully;
+    private boolean batteryOptimizationTurnedOff;
```
shahzaibj, line 73: I think you used the word `disabled` some places and `turnedOff` some places. Let's make these consistent. I think `disabled` might be better as less characters but I'd leave it up to you.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
```diff
@@ -150,18 +150,31 @@ private void installOrUpdateAppFromMarketPageInternal(String playStoreAction) th
                 new UiSelector().className(Button.class).text(playStoreAction).enabled(true)
         );
 
-        uiObjBtn.waitForExists(FIND_UI_ELEMENT_TIMEOUT);
+        // Sometimes, firebase devices take a very long time to download an app (outlook for example),
+        // a subsequent retry on the same test will find the app already downloading, and will not find the install/update button
+        // instead of just checking for install/update button, we can check for cancel button (Without pressing it)
```
shahzaibj: I'm not entirely sure I understand what's happening here. Are we speeding up the test?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
```diff
@@ -313,6 +314,24 @@ public static void handleButtonClick(@NonNull final String resourceId) {
         }
     }
 
+    /**
+     * Clicks the button element associated to the supplied resource id with a longer timeout.
+     * Creating another method to use the longer timeout here, some ui interactions take longer than expected.
+     *
+     * @param resourceId the resource id of the button to click
+     */
+    public static void handleButtonClickLongTimeout(@NonNull final String resourceId) {
```
shahzaibj: We can probably write a method as follows:

```java
public static void handleButtonClick(@NonNull final String resourceId, long timeOut);
```

Something like that. this would allow it to be more easily configurable in the future.

shahzaibj (commented):
Left some comments...

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/WordApp.java
fadidurah: Using UiAutomatorUtils and calling the obtain methods calls a waitForExists as long as FIND_UI_ELEMENT_TIMEOUT. I think here, that would be 25 seconds.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/WordApp.java
fadidurah: I don't see why not

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
fadidurah: Sure thing, for this it was just enrollment taking a bit longer than expected

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
fadidurah: We're actually using this timeout more often than the longer one. The longer one is a bit overkill for this case iirc

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
fadidurah: will add consistency here. I used `turnedoff` because that's the terminology used in CP UI.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
fadidurah: Not really speeding it up, just improving the chances for next retry to pass. What happens is firebase sometimes gives us a device with questionable download speed, so downloading the application takes longer than 5 minutes. By the time the retry starts, it's still downloading, so we check for the cancel button incase it's still downloading by the time we reach the play store page again. We could argue against doing this since we want a "clean" slate at the start of each test, but I also think it might be a waste to run these playstore test cases on firebase since they are quite costly budget wise. I have suggestion I got from OneAuth team about how we could alleviate this. Will include it in the upcoming design doc.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
fadidurah: Yes, that is a much better solution.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/installer/PlayStore.java
fadidurah: I'm going to get rid of this, this is a bad solution regardless of what we decide.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/WordApp.java
```diff
@@ -135,6 +134,18 @@ private void signIn(@NonNull final String username,
...
         Logger.i(TAG, "Confirming account with supplied username is signed in..");
+
+        // Had a screen for microsoft 365 pop up occasionally
+        final UiObject msft365Object = UiAutomatorUtils.obtainUiObjectWithText("Go Premium with Microsoft 365 Personal");
+        if (msft365Object.exists()) {
+            final UiObject skipObject = UiAutomatorUtils.obtainUiObjectWithText("SKIP FOR NOW");
+            try {
+                skipObject.click();
+            } catch (UiObjectNotFoundException e) {
+                Logger.i(TAG, "Ignoring failure to find confrim account UI");
```
wainaina:
```suggestion
                Logger.i(TAG, "Ignoring failure to find confirm account UI");
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -343,4 +358,25 @@ protected void initialiseAppImpl() {
...
+
+            try {
+                okObj.click();
+            } catch (UiObjectNotFoundException e) {
+                // Nothing, just want to ignore this
+                // If the button does not exist, then we do not need to do anything
+            }
+        }
+    }
 }
```
wainaina:
```suggestion
}

```

wainaina (approved):

shahzaibj (approved):

PR #1852, opened 2022-09-23
Storage performance improvements

nickbopp:
Perf data in telemetry shows P90 times in StorageRead/Write/Delete flows in the high hundreds of milliseconds.  Locally profiling, this should nearly always be CPU bound, and there are substantial improvements we can make (combined with #1836 and #1837).  Some of those I suspect will contribute substantially to broker performance improvement.  I'm not sure if these will as well as I'm less familiar with Broker's storage stack.  With this and my other two perf-related PRs, we are seeing local latencies cut by 70-90%, which should be sufficient to unblock partners.

[MsalCpp specific]
* Removes the refreshToken delete's from MsalCpp saveCredentials.  This (along with the AccountRecord lookup it needed).  Is not necessary.  It should be safe to just write the tokens we're writing and leave the rest alone.  This logic was already not accounting for the fact that we were sometimes passing two RefreshToken entries (one family and one non-family).  Profiling shows this is a ~5x speedup.

[Multiple users]
* Requires List<Credential> to be passed in on one of the filter methods.  Without this, it was previously just calling 'getCredentials' internally.  Many of the AccountCredentialCache operations involve a database scan + deserialization because they begin with 'getCredentials' or 'getAccounts'. Higher level operations in the TokenCache classes that call these operations multiple times can then involve several full scans and deserializations of every Credential/Account in the database.  Ideally these operations would not involve even a single full scan/deserialization and instead indexed into the list better.  But for now, this is a 3-5x improvement in most cases.
* removeCredential/removeAccount no longer involve a full scan + deserialization and operate on keys only. This was particularly problematic for operations that did full scans to filter to a set of credentials, and then another full scan for each credential that was eventually deleted.


common4j/src/main/com/microsoft/identity/common/java/cache/IAccountCredentialCache.java
```diff
@@ -157,7 +157,8 @@ List<Credential> getCredentialsFilteredBy(
             final String realm,
             final String target,
             final String authScheme,
-            final String requestedClaims
+            final String requestedClaims,
+            final List<Credential> inputCredentials
```
iamgusain, line 185: Please add to the javadoc as well.

iamgusain (approved):
LGTM

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -410,6 +410,38 @@ public List<Credential> getCredentialsFilteredBy(
...
+            @Nullable final String environment,
+            @Nullable final CredentialType credentialType,
+            @Nullable final String clientId,
+            @Nullable final String realm,
+            @Nullable final String target,
+            @Nullable final String authScheme,
+            @Nullable final String requestedClaims,
+            @Nullable final List<Credential> inputCredentials) {
+        final String methodTag = TAG + ":getCredentialsFilteredBy";
+        Logger.verbose(TAG, "getCredentialsFilteredBy()");
```
mohitc1: >TAG [](http://example.com/codeflow?start=23&length=3)

methodTag

mohitc1: nit: Worth adding better description in message.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1661,7 +1661,8 @@ private void deleteAccessTokensWithIntersectingScopes(
                 referenceToken.getRealm(),
                 null, // Wildcard (*)
                 referenceToken.getAccessTokenType(),
-                referenceToken.getRequestedClaims()
+                referenceToken.getRequestedClaims(),
+                mAccountCredentialCache.getCredentials()
```
mohitc1, line 1665: >mAccountCredentialCache.getCredentials() [](http://example.com/codeflow?start=16&length=40)

This call is still passing all credentials from same object (effectively same as before). 

From your description, do you mean the benefit would be when there are multiple calls to getCredentialsFilteredBy, same list can be passed to each, avoiding internal call to getCredentials?

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
nickbopp: Yes, that's exactly right.  This particular one doesn't benefit -- it's okay because all the paths that call into this function only call it once, and don't call anything else that invokes 'getCredentials'.

nickbopp: The usage of this will come on the OneAuth/MSAL side.  We have a read method that loops over each credential type and calls `getCredentialsFilteredBy` each time.  Since that internally calls `getCredentials`, it winds up getting called many times.  With this new overload, I can call `getCredentials` once at the top of the function, and reuse it for each filter call (which is fast).

common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -139,13 +138,6 @@ public synchronized void saveCredentials(@Nullable final AccountRecord accountRe
             }
         }
 
-        if (accountRecord != null && refreshTokenRecord != null) {
-            // MSAL C++ writes credentials first and then the account.
```
shoatman, line 143: Not sure I understand this comment.  Did we wask Brian why this was done before removing it?

common4j/src/main/com/microsoft/identity/common/java/cache/IAccountCredentialCache.java
```diff
@@ -160,6 +160,31 @@ List<Credential> getCredentialsFilteredBy(
...
+     * @param realm            The realm used to match Credential cache keys.
+     * @param target           The target used to match Credential cache keys.
+     * @param requestedClaims  The requested claims used to match Credential cache keys.
+     * @param inputCredentials The list of credentials to be filtered from.
+     * @return A mutable List of Credentials matching the supplied criteria.
+     */
+    List<Credential> getCredentialsFilteredBy(
+            final String homeAccountId,
+            final String environment,
+            final CredentialType credentialType,
```
shoatman, line 179: These should be marked as Nullable if we do allow them to be set to null.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
iambmelt: As I recall, this was written to ensure that a single RT is present in the cache for a user at any given time (we're removing any existing RT before writing a new one). By removing this, the assertion appears to be either that this will always be the case anyway (absent some kind of enforcement) or that this doesn't matter.

If the newly written RT always has the same cache key as the old one, this can be removed without changing the effect of the method.

common4j/src/main/com/microsoft/identity/common/java/cache/IAccountCredentialCache.java
iambmelt: This is marked on the concrete implementation, but not the interface definition. 

Either
- The subclass dictates the behavior of handling or not handling `null` and the interface has no opinion (current)

Or
- The interface defines how subclasses should be handling or not handling `null` (alternative option)

Re-reviewing my own code here, option 2 looks better from a static analysis and Liskov substitution standpoint

PR #1854, opened 2022-09-26
Add support for client/application managed key in pop flow

iamgusain:
### What
Add support for client/application managed key in pop flow
ADO: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2008118/

### Why
To support scenario where client app wants to pass req_cnf claim with the kid/thumbprint of the key owned by the client application. This is needed as the client wants to create a custom SHR themselves and sign it with the key they own. More details in the [ADO item](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2008118/)

### How
Adding a new AuthScheme PopAuthenticationSchemeWithClientKeyInternal, which is used to pass the "kid/thumbprint" for client/application managed key. This is then used to pass "req_cnf" in the request body for token request, to get a pop token with required claims.
Also added the filtering/matching logic to filter the accesstoken by kid when retrieving from cache, if PopAuthenticationSchemeWithClientKey is used.

### Testing
Verified getting regular Pop token and Pop Token with client key using MsalTestApp
Also added Unit tests,


rpdome:
We'll need a new broker protocol version for this change - to make sure things are properly handled if the active broker is old.

That said, i'm not sure how the protocol versioning works with oneauth, though.

iamgusain:
> We'll need a new broker protocol version for this change - to make sure things are properly handled if the active broker is old.
> 
> That said, i'm not sure how the protocol versioning works with oneauth, though.

Upgraded the broker protocol version to 11.0
OneAuth directly calls BrokerMsalController apis and pass the requiredBrokerProtocolVersion in the token parameters

common4j/src/main/com/microsoft/identity/common/java/authscheme/AuthenticationSchemeFactory.java
```diff
@@ -79,6 +80,25 @@ public static AbstractAuthenticationScheme createScheme(@NonNull final IPlatform
...
+                            "Constructing PoP Authentication Scheme With Client Key."
+                    );
+
+                    final IPoPAuthenticationSchemeParams params = (IPoPAuthenticationSchemeParams) nameable;
+                        return new PopAuthenticationSchemeWithClientKeyInternal(
+                                params.getHttpMethod(),
+                                params.getUrl(),
+                                params.getNonce(),
+                                params.getClientClaims(),
+                                params.getClientClaims()
```
mohitc1: >params.getClientClaims() [](http://example.com/codeflow?start=32&length=24)

is this correct? 
I think it can be cast to PopAuthenticationSchemeWithClientKeyInternal itself.

common4j/src/main/com/microsoft/identity/common/java/authscheme/AuthenticationSchemeFactory.java
```diff
@@ -79,6 +80,25 @@ public static AbstractAuthenticationScheme createScheme(@NonNull final IPlatform
...
+                        );
```
mohitc1: indentation

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -964,4 +969,19 @@ private void saveMsaAccountToCache(final @NonNull Bundle resultBundle,
...
+    /**
+     * Verifies if the token parameters are supported by the required broker protocol version
+     * @param parameters Token Parameters for verify
+     * @throws ClientException if the token parameters are not supported
+     */
+    private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParameters parameters) throws ClientException {
+        final String requiredProtocolVersion = parameters.getRequiredBrokerProtocolVersion();
+        if (parameters.getAuthenticationScheme() instanceof PopAuthenticationSchemeWithClientKeyInternal
+                && !BrokerProtocolVersionUtil.canSupportPopAuthenticationSchemeWithClientKey(requiredProtocolVersion)){
+            throw new ClientException(ClientException.AUTH_SCHEME_MISMATCH,
```
somalaya: AUTH_SCHEME_MISMATCH would be used when response from server does not match requested scheme right? Here I think we are verifying if the scheme requested is supported by current broker version. So should we use a different client exception like AUTH_SCHEME_NOT_SUPPORTED?

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Constructs a new PopAuthenticationSchemeWithClientKeyInternal.
+     *
+     * @param httpMethod   The HTTP method associated with this request. Optional.
+     * @param url          The resource URL of future-recipient of this SHR.
+     * @param nonce        Client nonce value; for replay protection.
+     * @param clientClaims Optional claims provided by the caller to embed in the client_claims
+     *                     property of the resulting SHR.
+     */
```
mohitc1, line 68: description for kid

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
shahzaibj: +1 to Sowmya's comment. Let's use a more clearer and specific error code here

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -331,6 +331,7 @@ public List<Credential> getCredentialsFilteredBy(
         final List<Credential> allCredentials = getCredentials();
 
         final List<Credential> matchingCredentials = getCredentialsFilteredByInternal(
+                allCredentials,
```
shahzaibj, line 331: Did we change the order of some params here? Is this a breaking change?

common4j/src/main/com/microsoft/identity/common/java/authscheme/AuthenticationSchemeFactory.java
```diff
@@ -87,4 +107,13 @@ public static AbstractAuthenticationScheme createScheme(@NonNull final IPlatform
...
     }
 
+    /**
+     * Checks the given authentication scheme is a PoP authentication scheme or not
+     * @param authenticationScheme
+     * @return boolean indicating if the the authentication scheme is a PoP authentication scheme
+     */
+    public static boolean isPopAuthenticationScheme(@NonNull final AbstractAuthenticationScheme authenticationScheme) {
+        return authenticationScheme instanceof PopAuthenticationSchemeInternal
+                || authenticationScheme instanceof PopAuthenticationSchemeWithClientKeyInternal;
```
shahzaibj:
```suggestion
        return authenticationScheme instanceof IPoPAuthenticationSchemeParams;
```

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+    private String mNonce;
+
+    @SerializedName(CLIENT_CLAIMS)
+    private String mClientClaims;
+
+    @SerializedName(KID)
+    private String mKid;
+
+    /**
+     * Constructor for gson use.
```
shahzaibj, line 58: What exactly is happening with GSON here?

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+ * Client application.
+ */
+@SuperBuilder(toBuilder = true)
+@EqualsAndHashCode(callSuper = true)
+@Accessors(prefix = "m")
+public class PopAuthenticationSchemeWithClientKeyInternal
+        extends TokenAuthenticationScheme
+        implements IPoPAuthenticationSchemeParams {
+
+    private static final long serialVersionUID = 788393037295696359L;
```
shahzaibj, line 48: Is there any utility of adding these serial version UID when Java serialization is NOT used? I mean we are using JSON, so the serialVersionUID should be being ignored by Java.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -117,6 +118,8 @@
 
     private final HttpClient httpClient = UrlConnectionHttpClient.getDefaultInstance();
 
+    private AbstractAuthenticationScheme mAuthScheme;
```
shahzaibj: If this cannot be `final` then it should probably be removed, otherwise it can cause state issues.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -379,7 +382,8 @@ public MicrosoftStsTokenRequest createTokenRequest(@NonNull final MicrosoftStsAu
             tokenRequest.setGrantType(TokenRequest.GrantTypes.AUTHORIZATION_CODE);
         }
 
-        if (PopAuthenticationSchemeInternal.SCHEME_POP.equals(authScheme.getName())) {
+        this.setAuthenticationScheme(authScheme);
```
shahzaibj: We shouldn't be using `setters` unless extremely necessary. This is going to make our code prone to state issues. If a `MicrosoftStsOauth2Strategy` object is modified by another piece of code that sets its authentication scheme object then we're going to run into problems.

I know the usage of setters is problem with the existing code. We have been moving away from this model and towards classes that contain `final` variables and provides `Builders` to construct the object to effectively become immutable. We should avoid the use of setters for any new code we write.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Constructs a new PopAuthenticationSchemeWithClientKeyInternal.
+     *
+     * @param httpMethod   The HTTP method associated with this request. Optional.
+     * @param url          The resource URL of future-recipient of this SHR.
+     * @param nonce        Client nonce value; for replay protection.
+     * @param clientClaims Optional claims provided by the caller to embed in the client_claims
+     *                     property of the resulting SHR.
+     */
+    public PopAuthenticationSchemeWithClientKeyInternal(@Nullable final String httpMethod,
```
shahzaibj: I saw the `SuperBuilder` annotation but see the constructor as well. Is this constructor needed? Should we just stick to the builder instead?

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+@Accessors(prefix = "m")
+public class PopAuthenticationSchemeWithClientKeyInternal
+        extends TokenAuthenticationScheme
+        implements IPoPAuthenticationSchemeParams {
+
+    private static final long serialVersionUID = 788393037295696359L;
+    /**
+     * The name of this auth scheme as supplied in the Authorization header value.
+     */
+    public static final String SCHEME_POP_WITH_CLIENT_KEY = "PoP_With_Client_Key";
```
shahzaibj, line 52: Is this something actually accepted by ESTS? I was actually expecting to see `PoP`

common4j/src/main/com/microsoft/identity/common/java/authscheme/AuthenticationSchemeFactory.java
shahzaibj: This looks wrong. I believe this param is meant to be `kid`?

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public String getClientClaims() {
+        return mClientClaims;
+    }
+
+    @Override
+    @Nullable
+    public String getNonce() {
+        return mNonce;
+    }
```
shahzaibj: Looks like we wrote a lot of code here to write the `getters` even though we have lombok. Why not just add the `@getters` annotation to the class?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -716,13 +727,21 @@ private String getCloudSpecificTokenEndpoint(final MicrosoftAuthorizationRespons
...
      * Gets the at/pop device credential's thumbprint.
      *
      * @return The at/pop device credential thumbprint.
      */
     @Nullable
     public String getDeviceAtPopThumbprint() {
+        if (mAuthScheme instanceof PopAuthenticationSchemeWithClientKeyInternal) {
+           return ((PopAuthenticationSchemeWithClientKeyInternal) mAuthScheme).getKid();
+        }
+
```
shahzaibj: Rather than having this `if/else` block now to handle get thumbprint for 2 different types of auth scheme, we can move the getThumbprint method directly to the `IPopAuthSchemeParams` interface and implement it over there. As a hint, `mPopManager` is already present inside `PopAuthenticationSchemeInternal` class so what you need is already there. Once you do that then the code here becomes very simple, you can just invoke:

`mAuthScheme.getKeyThumbprint()` and you wouldn't have to worry about the specific type of pop auth scheme at this layer.

shahzaibj:
Unit tests are good. It would also be good to see some integration tests around caching and OAuth2Strategy for the new code added here.

shahzaibj (dismissed):
Left some comments / suggestions.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
iamgusain: We are now passing the credentials for filtering as first argument in the internal method. Not a breaking change.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
iamgusain: For consistency I followed PopAuthenticationSchemeInternal class for this class.
I think it was done based on the GSON recommendation of using a no-args default constructor.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
iamgusain: This is based on what we have in PopAuthenticationSchemeInternal class which I used as a reference for this class.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: I know, it's not ideal, but there is no other easy way to get the kid/thumbprint value other than accessing it from the authscheme. Otherwise, we have to parse the accessToken jwt and fetch the kid value from inside it, which we don't want to do.
To make the authscheme available at the time of saving token to the cache this was the only easier way without requiring a big breaking change.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
iamgusain: I am fine either way, and also keeping both.
I see in oneauth they are using the constructor today for PopAuthenitcationSchemeInternal.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
iamgusain: This is for our internal naming, ests recieves `pop `only

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: Good suggestion,
Currently IPopAuthSchemeParams  is also implemented on the Msal side in PopAuthenticationScheme class, which doesn't have PopManager.
We can either remove that class and start using authscheme classes from common, (it will require some refactoring and would be a breaking change) or we can keep the getKeyThumbprint as no-op in that class.
For now, I am keeping that kind of refactoring as out of scope for current change

iamgusain:
> Unit tests are good. It would also be good to see some integration tests around caching and OAuth2Strategy for the new code added here.

I have added e2e test in the broker [PR ](https://github.com/AzureAD/ad-accounts-for-android/pull/2010)that hits ests.

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
```diff
@@ -0,0 +1,145 @@
...
+        implements IPoPAuthenticationSchemeParams {
+
+    private static final long serialVersionUID = 788393037295696359L;
+    /**
+     * The name of this auth scheme as supplied in the Authorization header value.
+     */
+    public static final String SCHEME_POP_WITH_CLIENT_KEY = "PoP_With_Client_Key";
+
+    @SerializedName(HTTP_METHOD)
+    private String mHttpMethod;
```
rpdome: nit: extract redundant members into an abstract parent class?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
rpdome: Actually, getting it from the response is not a bad idea.

It's probably better than what we currently have - the persisted data should be based on the source (returned token) as much as possible.

common4j/src/main/com/microsoft/identity/common/java/util/BrokerProtocolVersionUtil.java
```diff
@@ -76,6 +77,19 @@ public static boolean canSendPKeyAuthHeaderToTheTokenEndpoint(@Nullable String c
...
     }
 
+    /**
+     * Verifies if client required broker protocol version supports PoPAuthenticationSchemeWithClientKey.
+     *
+     * @param clientRequiredBrokerProtocolVersion broker protocol version of the calling app.
+     * @return true if the broker protocol version of the calling app is larger or equal than
+     * the {@link BrokerProtocolVersionUtil#MSAL_TO_BROKER_PROTOCOL_POP_SCHEME_WITH_CLIENT_KEY_MINIMUM_VERSION}.
+     */
+    public static boolean canSupportPopAuthenticationSchemeWithClientKey(@Nullable String clientRequiredBrokerProtocolVersion) {
```
rpdome: nit: final

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -693,16 +694,28 @@ public ICacheRecord load(@NonNull final String clientId,
         final List<Credential> allCredentials = mAccountCredentialCache.getCredentials();
 
         // Load the AccessTokens
-        final List<Credential> accessTokens = mAccountCredentialCache.getCredentialsFilteredBy(
+        List<Credential> accessTokens;
+        if (authScheme instanceof PopAuthenticationSchemeWithClientKeyInternal) {
+            accessTokens = mAccountCredentialCache.getPoPAccessTokensFilteredBy(
```
rpdome: it's probably okay to just add the two params to getCredentialsFilteredBy(), so that we don't have to branch out the code here.

The method name is sort of misleading as well.... since the normal PoP scheme won't be going through this code path.

rpdome:
left some comments.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: I am checking with Brian to see the reason behind current implementation.
For now, in latest iteration I have made the authscheme part of the strategy Parameters, and removed the local variable

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: this is removed in latest iteration

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: From Brian: 
We don't want to crack open AT, as this won't work for new 1P services now that token encryption is rolling out.
https://review.learn.microsoft.com/en-us/identity/microsoft-identity-platform/token-encryption?branch=main

common4j/src/main/com/microsoft/identity/common/java/authscheme/PopAuthenticationSchemeWithClientKeyInternal.java
iamgusain: These were automatically added as part of implementing the interface

rpdome (approved):

mohitc1 (approved):
:shipit:

PR #1857, opened 2022-09-30
Added telemetry for teams devices specific scenario

somalaya:
**What**  : Identify telemetry for scenarios that are not applicable to mobile devices and are specific to teams devices.
**How** : There is only one path added for Teams devices which is to add default scope for dm agent for foci request. Objective of the telemetry added here is to check how many times dm agent is hitting this path while it gets the account to acquiring token silently. 
Rest of the teams devices flow follows common code path as that on mobile devices. 
**Testing** : 
- The foci scenario checks with a harcoded clientId to check if the calling app is DM agent.
-  So, I set the clientId in a testcase as that of dm agent's clientId and verified that it was hitting the code path
- Added telemetry in the code path
- Verified in kusto (android-sandbox environment) that the telemetry event is logged correctly

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/foci/FociQueryUtilities.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/foci/FociQueryUtilities.java
@@ -52,6 +52,10 @@ import java.net.URL;
 import java.util.UUID;
 
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
 import lombok.NonNull;
 
 public class FociQueryUtilities {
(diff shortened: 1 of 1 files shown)


moumighosh (approved):

common4j/src/main/com/microsoft/identity/common/java/foci/FociQueryUtilities.java
```diff
@@ -129,9 +132,15 @@ public static boolean tryFociTokenWithGivenClientId(@SuppressWarnings(WarningTyp
         //       for every FoCI apps (and hardcode it here).
         //       https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1222002
         if (ObjectUtils.equals(clientId, "87749df4-7ccf-48f8-aa87-704bad0e0e16")) {
-            scopes = "https://devicemgmt.teams.microsoft.com/.default " + BaseController.getDelimitedDefaultScopeString();
-            Logger.info(TAG + methodName,
-                    "Teams agent client ID - making a test request with teams agent resource.");
+            final Span span = GlobalOpenTelemetry.getTracer(TAG).spanBuilder("setScopeForDMAgentForFoci").startSpan();
```
shahzaibj, line 135: `setScopeForDMAgentForFoci` can be defined in `SpanName.java`

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/foci/FociQueryUtilities.java
somalaya: SpanName.java is present in broker. Shall I move it to common?

shahzaibj: Yes, feel free to do that. Or it is also fine the way it is and we can move it to common later once we have more spans in common.

PR #1859, opened 2022-10-03
Add new exception type for broker protocol not supported exception during msal-broker handshake

rpdome:
see https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1703, https://github.com/AzureAD/ad-accounts-for-android/pull/2015

When MSAL initialization fails due to a protocol handshake error, we will be returning an exception with enough info for the client app to show an actionable prompt to the customer. (i.e. update app XXXX)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -666,6 +666,16 @@ public final class AuthenticationConstants {
         public static final String NEGOTIATED_BP_VERSION_KEY = "common.broker.protocol.version.name";
 
         /**
+         * Code of the error that occurs as a result of MSAL-Broker protocol handshake (hello()).
+         */
+        public static final String HELLO_ERROR_CODE = "error";
+
+        /**
+         * Description of the error that occurs as a result of MSAL-Broker protocol handshake (hello()).
+         */
+        public static final String HELLO_ERROR_MESSAGE = "error_description";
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -665,6 +665,16 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String NEGOTIATED_BP_VERSION_KEY = "common.broker.protocol.version.name";
 
+        /**
+         * Code of the error that occurs as a result of MSAL-Broker protocol handshake (hello()).
+         */
+        public static final String HELLO_ERROR_CODE = "error";
```
rpdome, line 671: introduced new variables, same values, because using OAUTH variables in hello is really confusing.

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -86,10 +89,10 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
     private static final String TAG = MsalBrokerResultAdapter.class.getName();
```
mohitc1: Could you fix this too, to MsalBrokerResultAdapter.class.getSimpleName()?

mohitc1 (approved):

fadidurah (approved):
LGTM!

PR #1860, opened 2022-10-05
Wire telemetry to crypto operations in FIPS flow

rpdome:
PBI: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2000156

Scope: 
Wraps around usages of crypto objects from CryptoFactory with a function that emits an event.

CryptoFactoryTelemetryHelper.performCryptoOperationAndUploadTelemetry

related: https://github.com/AzureAD/ad-accounts-for-android/pull/2020

### DIFF
--- a/common4j/build.gradle
+++ b/common4j/build.gradle
@@ -221,6 +221,7 @@ dependencies {
 
     implementation platform("io.opentelemetry:opentelemetry-bom:1.16.0")
     implementation('io.opentelemetry:opentelemetry-api')
+    implementation('io.opentelemetry:opentelemetry-sdk-trace:1.18.0')
 }
 
 sourceCompatibility = "1.8"
--- a/common4j/build.gradle
+++ b/common4j/build.gradle
@@ -219,8 +219,8 @@ dependencies {
     testFixturesAnnotationProcessor "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
     resolvableTestFixturesImplementation "org.robolectric:junit:$rootProject.ext.robolectricVersion"
 
-    implementation platform("io.opentelemetry:opentelemetry-bom:1.16.0")
-    implementation('io.opentelemetry:opentelemetry-api')
+    implementation platform("io.opentelemetry:opentelemetry-bom:1.18.0")
+    implementation('io.opentelemetry:opentelemetry-api:1.18.0')
     implementation('io.opentelemetry:opentelemetry-sdk-trace:1.18.0')
 }
 
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryClassName.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryClassName.java
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.opentelemetry;
+
+/**
+ * Enum representing classes implementing {@link com.microsoft.identity.common.java.crypto.ICryptoFactory}.
+ */
+public enum CryptoFactoryTelemetryClassName {
+    DefaultCryptoFactory,
+    NonFipsBrokerCryptoFactory,
+    AndroidBrokerFipsCryptoFactory
+}
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
@@ -37,7 +37,7 @@ import lombok.NonNull;
 
 public class CryptoFactoryTelemetryHelper {
 
-    private static final String SPAN_NAME = "CryptoFactoryEvents";
+    private static final String SPAN_NAME = "CryptoFactoryEvent";
 
     /**
      * A helper class that consolidate all the telemetry emitting work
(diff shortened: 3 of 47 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -0,0 +1,87 @@
...
+    public static <T> T performCryptoTaskAndUploadTelemetry(@NonNull final CryptoFactoryOperationName operationName,
+                                                            @NonNull final String algorithmName,
+                                                            @NonNull final ICryptoFactory cryptoFactory,
+                                                            @NonNull final ICryptoOperationCallback<T> cryptoOperation)
+            throws ClientException {
+        final Span span = sTracer.spanBuilder(getSpanName(operationName)).startSpan();
+        try (final Scope scope = span.makeCurrent()) {
+            span.setAttribute(crypto_controller.name(), cryptoFactory.getClass().getSimpleName());
+            span.setAttribute(crypto_algorithm.name(), algorithmName);
+            span.setStatus(StatusCode.OK);
```
shahzaibj, line 60: We're setting status "OK" here - what if it fails on next line `cryptoOperation.perform()`? Do we want to mark status ERROR if that fails?

rpdome: for crypto operation, if an exception is not thrown = success.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -77,6 +86,19 @@ public static <T> T performCryptoTaskAndUploadTelemetry(@NonNull final CryptoFac
         }
     }
 
+    /**
+     * Get name of the current span, if possible.
+     **/
+    @Nullable
+    private static String getCurrentSpanName(){
```
shahzaibj: This can be added in a Util class `OtelUtilities.java`

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import lombok.NonNull;
+
+public class OTelUtility {
+    private static final String TAG = OTelUtility.class.getSimpleName();
+    private static final Tracer sTracer = GlobalOpenTelemetry.getTracer(TAG);
```
shahzaibj: Let's make this non-static and see if that has an effect on the error

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -0,0 +1,83 @@
...
+     * @return result of the crypto operation.
+     * */
+    public static <T> T performCryptoOperationAndUploadTelemetry(@NonNull final CryptoFactoryOperationName operationName,
+                                                                 @NonNull final String algorithmName,
+                                                                 @NonNull final ICryptoFactory cryptoFactory,
+                                                                 @NonNull final ICryptoOperation<T> cryptoOperation)
+            throws ClientException {
+        final Span span = OTelUtility.createSpan(SPAN_NAME);
+        try (final Scope scope = span.makeCurrent()) {
+            span.setAttribute(crypto_controller.name(), cryptoFactory.getClass().getSimpleName());
```
shahzaibj: Do we want to create an enum for `CryptoController` names? Class names can get refactored over time but still fundamentally mean the same thing, and if such refactor occurs then that will affect telemetry. I suggest we create enums for CryptoController names because enums are meant to not be refactored.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -0,0 +1,83 @@
...
+            span.setAttribute(crypto_operation.name(),
+                    getCryptoOperationEventName(operationName, algorithmName));
```
shahzaibj, line 59: The same feedback as above for operation names. The logic inside `getCryptoOperationEventName` can get refactored and that would affect our reporting. I suggest we create enums for operation names so we can stick to those names forever.

shahzaibj: I see you have some enums already for crypto factory operation name. I think we should also have enum for what we ultimately want to put inside `crypto_operation` attribute.

shahzaibj (changes requested):
Left some suggestions on using enums for some fields

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -0,0 +1,83 @@
...
+import com.microsoft.identity.common.java.util.StringUtil;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Scope;
+import lombok.NonNull;
+
+public class CryptoFactoryTelemetryHelper {
+
+    private static final String SPAN_NAME = "CryptoFactoryEvent";
```
shahzaibj: Let's put this inside `SpanName.java` and use the same naming pattern.

shahzaibj: I think SpanName.java is inside broker4j so we can move it to common4j

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
rpdome: I prefer keeping it this way, because.

1. If we want to have enum for every values, then we'll have to list all the combinations. This will be boilerplate work if we introduce more algorithms down the line - and i want to avoid that.

3. Algorithm name is unlikely to change (new one can be introduced, but existing one will remain the same).

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
rpdome: added getTelemetryClassName() in ICryptoFactory.

melissaahn (approved):
Thank you for starting the OTelUtility and SpanName classes.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoObjectTelemetryClassName.java
```diff
@@ -0,0 +1,36 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.opentelemetry;
+
+/**
+ * All the operations/methods provided by CryptoFactory
+ * Used as SpanName in OpenTelemetry
+ * */
+public enum CryptoObjectTelemetryClassName {
```
shahzaibj: I'm not really a fan of the word "Class" in an enum class name. Same feedback for other places.

shahzaibj: Can we just call it `CryptoObjectName` ?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryClassName.java
```diff
@@ -0,0 +1,32 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.opentelemetry;
+
+/**
+ * Enum representing classes implementing {@link com.microsoft.identity.common.java.crypto.ICryptoFactory}.
+ */
+public enum CryptoFactoryTelemetryClassName {
```
shahzaibj: Can we just call it `CryptoFactoryName`?

common4j/build.gradle
```diff
@@ -221,6 +221,7 @@ dependencies {
 
     implementation platform("io.opentelemetry:opentelemetry-bom:1.16.0")
     implementation('io.opentelemetry:opentelemetry-api')
+    implementation('io.opentelemetry:opentelemetry-sdk-trace:1.18.0')
```
shahzaibj, line 224: If we're going with specific version, I wonder if we should also update it for `bom` and `api`

shahzaibj: and do the same in broker as well

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoObjectTelemetryClassName.java
shahzaibj: I think this is my only remaining comment at this point, that let's rename it to `CryptoObjectName`

shahzaibj (approved):
Left one comment, else LGTM

PR #1863, opened 2022-10-07
Accept Map as Param as opposed to HashMap impl in getAuthorizationActivityIntent

shahzaibj:
Method params should generally be interfaces as opposed to implementations until or unless the code has strict dependency on certain arguments being specific types. 

Having them as interfaces give us and our partners flexibility in choosing their own implementation in their code for such fields.

This was reported by @yanchoyanev as they prefer using Map directly inside their source.


shahzaibj:
Closing this for now, as not needed at the moment.

PR #1864, opened 2022-10-10
Fix Office SSO when using SharedUserId

matthewyu-ms:
Office was experiencing SSO issue where account credential cache was wiped after force-closing Office apps and restarting. Looks like The Keystore KeyPair that is shared when SharedUserId is in use was being overwritten. In this PR, during the check for ADAL-KS file and realizing it is not there, don't delete the KeyStore keypair because that check was passed earlier. In this case, only delete the ADAL-KS file and local memcache.

Validated manually that this fixed the issue

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
@@ -206,7 +206,8 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
             final byte[] wrappedSecretKey = FileUtil.readFromFile(getKeyFile(), KEY_FILE_SIZE);
             if (wrappedSecretKey == null) {
                 Logger.warn(methodTag, "Key file is empty");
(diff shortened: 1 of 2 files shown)


iamgusain (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -206,7 +206,8 @@ protected SecretKey generateRandomKey() throws ClientException {
             final byte[] wrappedSecretKey = FileUtil.readFromFile(getKeyFile(), KEY_FILE_SIZE);
             if (wrappedSecretKey == null) {
                 Logger.warn(methodTag, "Key file is empty");
-                deleteSecretKeyFromStorage();
+                FileUtil.deleteFile(getKeyFile());
```
shahzaibj, line 211: Can you add some comments in the code (for future readers) why this case must be treated differently from the other one.

shahzaibj (approved):
Left a suggestion, else LGTM

PR #1866, opened 2022-10-13
Adding cached credential service request id to telemetry

iamgusain:
### What
Adding cached credential service request id to telemetry

### Why
This change is part of the ADO workitem
[Android broker should add telemetry indicating whether CCS was used to serve a token](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2018515/)

### How
Adding the attribute ccs_requestid to current span when parsing token response.

### Testing
Added UTs for getting header value if present

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
@@ -101,7 +101,7 @@ public class HttpResponse {
         }
 
         final List<String> list = mResponseHeaders.get(key);
-        if (list == null){
+        if (list == null || list.size() <= index){
             return null;
         }
 
--- a/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
@@ -94,7 +94,11 @@ public class HttpResponse {
     }
 
     /**
-     * Returns: element "index" of the header list associated to the provided key.
+     * Gets the value from the response header list associated to the provided key at a given index.
+     * @param key   header key for look up.
+     * @param index index of the element to get
+     * @return      element at given index of the header list associated to the provided key.
+     *              if element is not found returns null.
      */
     @Nullable
     public String getHeaderValue(@NonNull final String key, final int index){
--- a/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
@@ -22,6 +22,8 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.net;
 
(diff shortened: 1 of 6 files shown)


rpdome (approved):

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -608,6 +611,12 @@ protected TokenResult getTokenResultFromHttpResponse(@NonNull final HttpResponse
                     tokenResponse.setCliTelemSubErrorCode(cliTelemInfo.getServerSubErrorCode());
                 }
             }
+
+            if (null != responseHeaders.get(XMS_CCS_REQUEST_ID)) {
+                Span.current().setAttribute(
+                        AttributeName.ccs_request_id.name(),
+                        responseHeaders.get(XMS_CCS_REQUEST_ID).get(0));
```
shahzaibj: I think @rpdome found that this way of accessing the headers is unsafe and he created a new method for it I believe. @rpdome Can you share that here?

common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
```diff
@@ -101,7 +101,7 @@ public String getHeaderValue(@NonNull final String key, final int index){
         }
 
         final List<String> list = mResponseHeaders.get(key);
-        if (list == null){
+        if (list == null || list.size() <= index){
```
shahzaibj, line 110: We should probably also add a check for index being >= 0 (otherwise negative indices might make this crash)

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -615,6 +618,12 @@ protected TokenResult getTokenResultFromHttpResponse(@NonNull final HttpResponse
                     tokenResponse.setCliTelemSubErrorCode(cliTelemInfo.getServerSubErrorCode());
                 }
             }
+
+            if (null != responseHeaders.get(XMS_CCS_REQUEST_ID)) {
+                Span.current().setAttribute(
```
shahzaibj: Do we also want to emit this in JoinedAccountRequestHandler? I believe that code path doesn't use the strategies...can you double check?

iamgusain: Yes, we will need to do that, I will send a broker PR for that after this is merged.

shahzaibj (approved):

Issue #1867, opened 2022-10-18
Use of DateTimeFormatter in Logger has raised minimum SDK support to API 26


darrentaft:
Bug caused by commit:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/ea6e819bcbcdc0357c6cd42d6e210dde17cbe38b

This is now using APIs (including DateTimeFormatter) only available in Android API 26 and greater, causing use on older devices to crash:

```
Caused by java.lang.ClassNotFoundException: Didn't find class "java.time.format.DateTimeFormatter" on path: DexPathList[[zip file "/data/app/app/base.apk", zip file "/data/app/com.foo.bar-1/split_config.arm64_v8a.apk", zip file "/data/app/com.foo.bar-1/split_config.xhdpi.apk"],nativeLibraryDirectories=[/data/app/com.foo.bar-1/lib/arm64, /data/app/com.foo.bar-1/base.apk!/lib/arm64-v8a, /data/app/com.foo.bar-1/split_config.arm64_v8a.apk!/lib/arm64-v8a, /data/app/com.foo.bar-1/split_config.xhdpi.apk!/lib/arm64-v8a, /system/lib64, /vendor/lib64]]
       at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:56)
       at java.lang.ClassLoader.loadClass(ClassLoader.java:380)
       at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
       at com.microsoft.identity.common.java.logging.Logger.<clinit>(Logger.java:72)
       at com.microsoft.identity.common.java.logging.Logger.setLogger(Logger.java)
       at com.microsoft.identity.common.logging.Logger.setAndroidLogger(Logger.java:104)
       at com.microsoft.identity.common.logging.Logger.<clinit>(Logger.java:92)
       at com.microsoft.identity.common.internal.logging.Logger.setAllowLogcat(Logger.java:76)
       at com.microsoft.identity.client.Logger.setEnableLogcatLog(Logger.java:184)
       at com.foo.bar.login.MSALManager.setupLogging(MSALManager.kt:720)
       at com.foo.bar.login.MSALManager.initialiseVars(MSALManager.kt:394)
       at com.foo.bar.login.MSALManager.<init>(MSALManager.kt:75)
       at com.foo.bar.login.MSALManager.<init>(MSALManager.kt)
[...]

PR #1868, opened 2022-10-18
Remove java.time.* Java8 APIs

nickbopp:
Java 8 APIs are only available default on min-sdk >= 26.  Desugaring allows using language features that are compiled correctly even for lower min-sdk levels. There is also a feature to make [Java 8 APIs](https://developer.android.com/studio/write/java8-support#library-desugaring) available, but it requires the including *app* to opt-in to library desugaring.  Since common only controls its own library, it cannot transitively push these requirements on to the app.  Until min-sdk level is moved up to 26+, it will not be safe to use Java 8+ APIs in common.

This swap back to the original time APIs should be nearly as fast.  The main benefit from changing to Instant was the more clear time API surface, and some multithreading usage of the formatter that isn't actually needed given the Executor is the only thread using it.  The primary benefits of not formatting before checking logLevel and only creating one formatter will still be in place.

Validations:
* Checked all my PRs, these are the only Java 8 APIs introduced
* Validated that we crashed before this PR on API 24, and now fixed
* Validated that log format is the same

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [PATCH] Remove java.time.* Java8 APIs (#1868)
 - [MAJOR] Add support for client/application managed key in pop flow (#1854)
 - [PATCH] Avoid keystore key overwriting for apps using sharedUserId. (#1864)
 - [PATCH] Moved clearClientCertPreferences to onPageLoaded. (#1855)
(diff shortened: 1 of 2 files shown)


shahzaibj:
Fixes #1867

iamgusain (approved):

shahzaibj (approved):
Thanks Nick!

PR #1870, opened 2022-10-19
adding performSharedDeviceRegistrationDontValidate method that doesn'…

ellymakuba:
Description of this PR
What:

Implementing two end-to-end test cases; one for shared device WPJ with a non-admin account and the other to test In Shared device mode, MSAL should notify the app if the sign-out account is changed.
-https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833511
-https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833517

Why:

-833511; Ascertain that a non-admin account cannot perform shared device WPJ 
-833517; Ascertain that MSAL notifies the app when the sign-out account is changed

How:

-833511; Installed MSAL test app and launched, Installed Authenticator app, Navigated to Authenticator app's settings page and performed a shared device WPJ with a non-admin account

 -833517; Installed MSAL test app and launched, performed shared device registration, Launched MSAL Test app, performed  acquireToken with an account from the same tenant with the WPJed account, Navigated to Android's settings page and removed that account, Opened MSAL test app.

Testing:

-833511; Verified after the steps above that the "shared device mode" does not appear
-833517; Verified that the previous Single Account changed to null


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -177,6 +177,12 @@ public void performSharedDeviceRegistration(String username, String password) {
         }
     }
 
+    @Override
+    public void performSharedDeviceRegistrationDontValidate(@NonNull final String username,
+                                                            @NonNull final String password) {
+        throw new UnsupportedOperationException("Not Supported!");
```
iamgusain: Can you implement this here in Broker Host, as we have the control of it's UI,

iamgusain (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
ellymakuba: links to the test cases https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833517 & https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833511

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/ITestBroker.java
```diff
@@ -67,6 +67,15 @@ public interface ITestBroker extends IApp {
      */
     void performSharedDeviceRegistration(String username, String password);
 
+    /**
+     * Perform shared device registration with supplied username. This user must be a cloud device
+     * admin for the registration to actually succeed.
```
fadidurah: This Javadoc should explain what "DontValidate" in the method name means.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
```diff
@@ -138,6 +138,19 @@ public void performSharedDeviceRegistration(@NonNull final String username,
 
         isInSharedDeviceMode = true;
     }
+
+    @Override
+    public void performSharedDeviceRegistrationDontValidate(@NonNull final String username,
+                                                @NonNull final String password) {
+        Logger.i(TAG, "Performing Shared Device Registration for the given account..");
```
fadidurah: Let's change this log message to indicate the "Don't Validate" aspect.

fadidurah: Why do we need a performSharedDevieRegistration method where we don't check. Is this related to Samsung device automation? Are we not seeing the screen at all or are we failing to find the correct UI element because we don't have the resource id.

fadidurah: If you are having trouble using UiAutomatorViewer on your samsung device still, let's document this in your rampup notes. This may be useful for someone else later down the line.

fadidurah:
Left some comments.

ellymakuba:
Description of this PR
What:

Implementing two end-to-end test cases; one for shared device WPJ with a non-admin account and the other to test In Shared device mode, MSAL should notify the app if the sign-out account is changed.
-https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833511
-https://identitydivision.visualstudio.com/DevEx/_workitems/edit/833517

Why:

Ascertain that a non-admin account cannot perform shared device WPJ 

How:

-Installed MSAL test app and launched, Installed Authenticator app, Navigated to Authenticator app's settings page and performed a shared device WPJ with a non-admin account
 

Testing:

Verified after the steps above that the "shared device mode" does not appear

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
ellymakuba: > Why do we need a performSharedDevieRegistration method where we don't check. Is this related to Samsung device automation? Are we not seeing the screen at all or are we failing to find the correct UI element because we don't have the resource id.

The original performSharedDeviceRegistration method is designed to pass by validating the device is in shared mode, but the test case I'm implementing is supposed to fail when performing shared device with a non-admin account. Thus the need for a new performSharedDevice without validation.

fadidurah: I see, good to know. Thanks for clarifying that

fadidurah (approved):
Looks good to me! By the way, since you are working with testing and not any customer-facing feature, please add the no-changelog label to this PR. This will allow you to pass the changelog check as well.

ellymakuba:
> Looks good to me! By the way, since you are working with testing and not any customer-facing feature, please add the no-changelog label to this PR. This will allow you to pass the changelog check as well.

How do I add a non-changelog label?

PR #1874, opened 2022-10-19
add logs on browser selection logic

rpdome:
Intune/COBO is running into an issue where broker is not able to launch Chrome in their flow.

Adding logs to get more information on those devices.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
@@ -62,7 +62,7 @@ public class BrowserSelector {
     public static Browser select(final Context context, final List<BrowserDescriptor> browserSafeList) throws ClientException {
         final String methodTag = TAG + ":select";
         final List<Browser> allBrowsers = getAllBrowsers(context);
-        Logger.verbose(methodTag, "Select the browser to launch.");
+        Logger.info(methodTag, "Select the browser to launch from safeList size=" + browserSafeList.size());
 
         for (final Browser browser : allBrowsers) {
             for (final BrowserDescriptor browserDescriptor : browserSafeList) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
```diff
@@ -82,31 +82,34 @@ public static Browser select(final Context context, final List<BrowserDescriptor
...
-            return false;
-        }
+        final String methodTag = TAG + ":matches";
 
         if (!StringUtil.equalsIgnoreCase(descriptor.getPackageName(), browser.getPackageName())) {
             return false;
         }
 
         if (!descriptor.getSignatureHashes().equals(browser.getSignatureHashes())) {
+            Logger.info(methodTag, "Signature hash does not match. " +
```
rodomsft, line 94: Can you add the packageName in these? Or is it safe to assume there's just one safe package for now.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
```diff
@@ -62,7 +62,7 @@ public class BrowserSelector {
     public static Browser select(final Context context, final List<BrowserDescriptor> browserSafeList) throws ClientException {
         final String methodTag = TAG + ":select";
         final List<Browser> allBrowsers = getAllBrowsers(context);
-        Logger.verbose(methodTag, "Select the browser to launch.");
+        Logger.info(methodTag, "Select the browser to launch from safeList size=" + browserSafeList.size());
```
rodomsft, line 65: Log the size of allBrowsers as well. I'd prefer you log the actual package names but if that's a privacy concern I understand.

rpdome: Not a privacy concern - i just didn't want to flush the logs. 

We definitely can do that in the private test build though.

rpdome:
log added as part of https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1957

PR #1876, opened 2022-10-20
Adding OTel to CertBasedAuth

melissaahn:
## Summary
Certificate Based Authentication (CertBasedAuth or CBA) now has support for creating and adding attributes to an OpenTelemetry `Span` named `CertBasedAuth`. 
Some things to look at in this PR:
- I've added the OpenTelemetry dependencies to Common's build.gradle.
- I created a helper class named `CertBasedAuthTelemetryHelper` to handle telemetry related tasks for the CBA classes. 
- I'm passing a `Span` as a parameter into a method because there are YubiKit tasks that need to run in a different thread, and this thread needs a reference to the same `Span`, otherwise `Span.current()` in this different thread returns null.
- The attribute `cert_based_auth_user_choice` is currently useless, but it will be useful in an upcoming update to CBA, so I'm setting it to be empty in order to make the column appear in Aria (hoping this will allow a dashboard object to be worked on in advance).
- I've deleted old telemetry CBA related classes and strings.


shahzaibj (changes requested):
> I'm passing a Span as a parameter into a method because there are YubiKit tasks that need to run in a different thread, and this thread needs a reference to the same Span, otherwise Span.current() in this different thread returns null.

There is a different way to propagate the Span across threads. I did cover it in my design doc. Let me follow up with you to get that integrated into your changes.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
```diff
@@ -0,0 +1,152 @@
...
+            final Span span = Span.current();
+            span.setAttribute(
+                    AttributeName.cert_based_auth_user_choice.name(),
+                    choice);
+        }
+    }
+
+    /**
+     * Checks if the current Span is related to CBA.
+     * Used to ensure CBA attributes are not being set for other unrelated Spans.
```
shahzaibj: It would be really good to know how often the CBA code can run outside of the CBA span

shahzaibj: Can you enumerate the edge cases here?

Or if this has high likelihood of happening then we have a bigger problem in the code

melissaahn: It has a pretty slim to none likelihood at the moment of happening (based on my latest testing). 
But I did include it because CBA can't be encapsulated by a single try/catch block, so the calls to the helper class may seem scattered to someone unfamiliar to the code. Also, the check does gives me some peace of mind by ensuring that CBA attributes will never ever be set for any other unrelated span.
But if this check does seem more confusing or unnecessary than helpful, I'm ok with leaving it out too.

melissaahn:
Thank you @shahzaibj for figuring out a better way to pass the OpenTelemetry Context to a new runnable started within the YubiKit library. A new method `wrapCallback` was added to `YubiKitCertBasedAuthManager`.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -461,6 +462,10 @@ public void onReceivedClientCertRequest(@NonNull final WebView view,
         if (mCertBasedAuthChallengeHandler != null) {
             mCertBasedAuthChallengeHandler.cleanUp();
         }
+        CertBasedAuthTelemetryHelper.createSpanAndMakeCurrent();
+        //Span attribute for user_choice will be utilized soon in a future update, but
+        // setting it here to be empty in order to prepare for Aria dashboards.
+        CertBasedAuthTelemetryHelper.setUserChoice("");
```
shahzaibj:
```suggestion
        CertBasedAuthTelemetryHelper.setUserChoice(null);
```

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
```diff
@@ -0,0 +1,166 @@
...
+ *  telemetry-related tasks.
+ */
+public class CertBasedAuthTelemetryHelper {
+
+    /**
+     * Creates a new Span and sets it to be the current Span.
+     */
+    public static void createSpanAndMakeCurrent() {
+        final Span span = OTelUtility.createSpan(SpanName.CertBasedAuth.name());
+        span.makeCurrent();
```
shahzaibj: The scopes created by calls to `makeCurrent` must be closed at the end of those Spans. See this: https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/main/java/io/opentelemetry/context/ImplicitContextKeyed.java#L31

This is the reason Otel Java recommends using this in conjunction with `try-with-resources`: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

Otherwise this is going to lead to memory leaks.

shahzaibj (changes requested):
Left a comment about potential memory leak

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
melissaahn: A static variable (mScope) was created to hold the current Scope once span.makeCurrent() is called. 
mScope.close() is called within a null check in all of the setResult methods.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
```diff
@@ -0,0 +1,134 @@
...
+import io.opentelemetry.context.Scope;
+import lombok.NonNull;
+
+/**
+ * Assists classes associated with Certificate Based Authentication (CBA) with
+ *  telemetry-related tasks.
+ */
+public class CertBasedAuthTelemetryHelper {
+
+    private static Scope mScope;
```
shahzaibj: I'm not sure why this needs to be static. Why can't it be a member variable? Perhaps you intended to make it member given that you've prefixed it with `m`?

melissaahn: Making `mScope` (should actually be renamed to `sScope` - my bad) a member variable would mean that I would need to create an instance of this class and pass it through as a new parameter to classes such as the challenge handlers, the cert based auth managers, and the smartcard sessions as well. Alternatively, I could make `sScope` a member variable of the challenge handlers, but that would also require passing it in as a parameter to the manager and and smartcard sessions, and I might as well rewrite this PR to instead collect variable values until the very end (where the span can be created and ended in one shot).
This is why I think having a static scope variable would be more appropriate, but let me know if there's something I'm missing here.

shahzaibj: Statics in general are considered bad because they represent global state and as far as I can see your CBA spans are not something global but rather each interaction with CBA is one Session. You can read more why statics are bad over the internet, here's one example: https://stackoverflow.com/questions/7026507/why-are-static-variables-considered-evil#:~:text=Static%20variables%20are%20generally%20considered,assumptions%20of%20object%2Doriented%20programming.

Here are some questions for you:

- How does a static scope variable represent the scope for one given CBA session? 
- What happens when a second session launches? 
- Does the scope get overridden? 
- How do we know which scope corresponds to which CBA session?

PR #1878, opened 2022-10-20
Merge release 8.0.2 to dev

iamgusain:
Merging release/8.0.2 to dev

### DIFF
--- a/common4j/versioning/version.properties
+++ b/common4j/versioning/version.properties
@@ -1,4 +1,4 @@
 #Wed May 12 20:08:39 UTC 2021
-versionName=5.0.0
+versionName=5.0.1
 versionCode=1
 latestPatchVersion=227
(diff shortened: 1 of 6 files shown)


common4j/versioning/version.properties
```diff
@@ -1,4 +1,4 @@
 #Wed May 12 20:08:39 UTC 2021
-versionName=5.0.0
+versionName=5.0.1
```
shahzaibj, line 2: Just want to make sure this is right. Did we not end up needing a 5.0.2?

shahzaibj (approved):
Left a comment, to double check on something.

common4j/versioning/version.properties
iamgusain: This is correct.
For common8.0.1 there was no common4j release required, so it consumed 5.0.0.
Common 8.0.2 consumes common4j 5.0.1

somalaya (approved):

PR #1880, opened 2022-10-24
Enable signing testFixtureJar for maven release

iamgusain:
Enabling the support for signing testfixture jar when creating maven package for publishing.
Currently it's not signed, and it fails verification when releasing to external maven with error "Missing Signature" for the testfixture jar

### DIFF
--- a/azure-pipelines/maven-release/common4j.yml
+++ b/azure-pipelines/maven-release/common4j.yml
@@ -28,3 +28,4 @@ jobs:
     gpgSourcesJar: true
     gpgJavadocJar: true
     gpgJar: true
+    testFixtureJar: true
--- a/azure-pipelines/maven-release/common4j.yml
+++ b/azure-pipelines/maven-release/common4j.yml
@@ -28,4 +28,4 @@ jobs:
     gpgSourcesJar: true
     gpgJavadocJar: true
     gpgJar: true
-    testFixtureJar: true
+    gpgTestFixtureJar: true
--- a/azure-pipelines/templates/steps/maven-release/gpg-signing.sh
+++ b/azure-pipelines/templates/steps/maven-release/gpg-signing.sh
@@ -73,6 +73,18 @@ if [ $JAVADOCJAR == "true" ]; then
     fi
 fi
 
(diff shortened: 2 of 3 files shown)


azure-pipelines/maven-release/common4j.yml
```diff
@@ -28,3 +28,4 @@ jobs:
     gpgSourcesJar: true
     gpgJavadocJar: true
     gpgJar: true
+    testFixtureJar: true
```
p3dr0rv:
```suggestion
    gpgTestFixtureJar: true
```

azure-pipelines/templates/steps/maven-release/gpg-signing.sh
```diff
@@ -73,6 +73,18 @@ if [ $JAVADOCJAR == "true" ]; then
     fi
 fi
 
+if [ $TESTFIXUTREJAR == "true" ]; then
```
p3dr0rv:
```suggestion
if [ $TESTFIXTUREJAR == "true" ]; then
```

p3dr0rv (changes requested):

shahzaibj:
It seems like the bigger issue here is that testFixtures are being released to external maven? Should we try to fix that rather than signing these?

iamgusain:
> It seems like the bigger issue here is that testFixtures are being released to external maven? Should we try to fix that rather than signing these?

testFixture jar is by default generated by the testFixture plugin, in our maven release we are copying all the generated jars as part of the maven package. We can exclude it there but I thought we included these as testFixture so the consumers could use them for testing. If these are internal only, should these be part of our other testUtil libs.

shahzaibj:
> > It seems like the bigger issue here is that testFixtures are being released to external maven? Should we try to fix that rather than signing these?
> 
> testFixture jar is by default generated by the testFixture plugin, in our maven release we are copying all the generated jars as part of the maven package. We can exclude it there but I thought we included these as testFixture so the consumers could use them for testing. If these are internal only, should these be part of our other testUtil libs.

Interesting question. @p3dr0rv Any thoughts here?

somalaya (approved):

PR #1885, opened 2022-10-26
Melissaahn/cba usb service

melissaahn:
## Summary
This change is being made to address an issue brought up regarding Android emulators that are not supporting the usb_service system service at the moment. 
In the case described above, `context.getSystemService(Context.USB_SERVICE)` will return null. This is more or less an unusual edge case that YubiKit does not currently handle, so it will cause a NullPointerException upon starting up the authentication WebView (which starts startUsbDiscovery).
It will take some time for Yubico and the emulator company to evaluate changes that may/may not need to be made on their sides, so making some null checks directly in our code to unblock customers.
This change will also be cherry-picked into upcoming releases of Authenticator and Company Portal.

## Testing
Tested on my side by explicitly setting the YubiKitManager variable to null and making sure the app doesn't crash. Service engineer also tested with test version of Authenticator and confirmed the fix.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitCertBasedAuthManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitCertBasedAuthManager.java
@@ -64,7 +64,13 @@ public class YubiKitCertBasedAuthManager extends AbstractSmartcardCertBasedAuthM
      * @param context current application context.
      */
     public YubiKitCertBasedAuthManager(@NonNull final Context context) {
-        mYubiKitManager = new YubiKitManager(context);
+        //Edge case where device is an emulator that doesn't account for USB_SERVICE.
+        if (context.getSystemService(Context.USB_SERVICE) != null) {
+            mYubiKitManager = new YubiKitManager(context);
+        } else {
+            mYubiKitManager = null;
+            Logger.info(TAG, "Certificate Based Authentication via YubiKey not enabled due to device not supporting USB_SERVICE.");
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitCertBasedAuthManager.java
```diff
@@ -74,42 +80,44 @@ public YubiKitCertBasedAuthManager(@NonNull final Context context) {
...
-        mYubiKitManager.startUsbDiscovery(new UsbConfiguration(), new Callback<UsbYubiKeyDevice>() {
-            @Override
-            public void invoke(@NonNull UsbYubiKeyDevice device) {
-                Logger.verbose(TAG, "A YubiKey device was connected");
-                synchronized (sDeviceLock) {
-                    mDevice = device;
-                    if (mConnectionCallback != null) {
-                        mConnectionCallback.onCreateConnection();
-                    }
+        if (mYubiKitManager != null) {
```
rpdome: nit: 

```Java
if (mYubiKitManager == null) {
     // log
     return;
}

// do work.
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitCertBasedAuthManager.java
```diff
@@ -64,7 +64,13 @@ public class YubiKitCertBasedAuthManager extends AbstractSmartcardCertBasedAuthM
      * @param context current application context.
      */
     public YubiKitCertBasedAuthManager(@NonNull final Context context) {
-        mYubiKitManager = new YubiKitManager(context);
+        //Edge case where device is an emulator that doesn't account for USB_SERVICE.
```
rpdome: nit: Edge case where device doesn't support USB_SERVICE.

rpdome (approved):

iamgusain (approved):

PR #1887, opened 2022-10-26
Use android.net.NetworkCapabilities to check internet connectivity

iamgusain:
### What
Use android.net.NetworkCapabilities to check internet connectivity
ADO Item: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2017766/

### Why
[NetworkInfo](https://developer.android.com/reference/android/net/NetworkInfo) class is marked as deprecated. It is recommended that callers should instead use the [ConnectivityManager.NetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback) API to learn about connectivity changes, or switch to use [ConnectivityManager#getNetworkCapabilities](https://developer.android.com/reference/android/net/ConnectivityManager#getNetworkCapabilities(android.net.Network)) or [ConnectivityManager#getLinkProperties](https://developer.android.com/reference/android/net/ConnectivityManager#getLinkProperties(android.net.Network)) to get information synchronously.
Since we are checking the connectivity info synchronously  before doing a network operation switching to use getNetworkCapabilities for that purpose.


shahzaibj (approved):
This looks good to me based on documentation.

rpdome (approved):

rpdome:
I would suggest adding telemetry around this change and monitor it with the next broker roll out. 

I don't expect it to break - but just in case it doesn't work on some heavily-modified OEM.

also. nit: changelog.

somalaya (approved):

PR #1888, opened 2022-10-27
Cherry-pick CBA fix for 7.0.2

rpdome:
Cherry-pick the change @melissaahn made for 7.0.1.

The fix addressed the same issue in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1885, but the code was refactored in dev, hence the difference.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,10 @@
 V.Next
 ----------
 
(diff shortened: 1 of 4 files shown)


gitaumoses4 (approved):
LGTM

PR #1890, opened 2022-10-27
Turn PlatformComponent into a pluggable module

rpdome:
Convert PlatformComponent into a pluggable module such that
- We could avoid creating dependencies issue between classes inside PlatformComponent - as it needs to be initialized "in order" (via AndroidPlatformComponentFactory classes)
     - This issue is not obvious at this level, but is *very* obvious in Broker (such that I'll have to send out multiple PR to detangle them. you'll see that soon).
- Extract methods in IPlatformComponents that requires parameter into separate modules (IPopManagerLoader, IStorageLoader). 
      - I still keep IPlatformComponent untouched though, otherwise this PR will be way bigger.
- We could reuse PlatformComponent directly in mock classes. 

ps. Tried my best to keep it as small as possible - this is the best I can do.
ps2. Believe it or not... this is still backcompat with current broker dev :D. I'm keeping breaking changes in separate PRs


common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
```diff
@@ -22,24 +22,24 @@
...
-import com.microsoft.identity.common.java.strategies.IAuthorizationStrategyFactory;
 
-import edu.umd.cs.findbugs.annotations.Nullable;
 import lombok.NonNull;
 
 /**
- * Common components for each platforms.
+ * An interface for providing components for each platforms.
+ *
+ * TODO: make getPopManagerLoader() and getStorageLoader() part of this interface.
```
iamgusain, line 40: Once getPopManagerLoader() and getStorageLoader() are part of this interface, 
Then I assume that we would remove the `IPlatformComponents extends IPopManagerLoader, IStorageLoader` and just keep it IPlatformComponents interface, right?

common4j/src/main/com/microsoft/identity/common/java/interfaces/PlatformComponents.java
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * An class which provide components for each platforms.
+ */
+@Getter
+@SuperBuilder
+@Accessors(prefix = "m")
+public class PlatformComponents implements IPlatformComponents {
+
+    @NonNull
+    private IKeyAccessor mStorageEncryptionManager;
```
iamgusain: mKeyAccessor?

rpdome: intentionally keeping it the same name as the existing IPlatformComponents interface.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPlatformComponents.java
rpdome: Eventually, we could expose IPopManagerLoader and IStorageLoader directly via this interface (as opposed to extending it) - as that would come with a breaking change (Planned to do this in a separate PR).

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -0,0 +1,123 @@
...
+public class AndroidPlatformComponentsFactory {
+    /**
+     * True if all of the platform-dependent static classes have been initialized.
+     */
+    private static boolean sGlobalStateInitalized = false;
+
+    /**
+     * Initializes platform-dependent static classes.
+     */
+    public static synchronized void initializeGlobalStates(@NonNull final Context context){
```
iamgusain, line 59: private?
unless we want to initialize these without creating a AndroidPlatformComponent

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -0,0 +1,123 @@
...
+                                              @Nullable final Fragment fragment) {
+        initializeGlobalStates(context);
+
+        final PlatformComponents.PlatformComponentsBuilder builder = PlatformComponents.builder();
+        fillBuilder(builder, context, activity, fragment);
+        return builder.build();
+    }
+
+    @SuppressWarnings(WarningType.rawtype_warning)
+    public static void fillBuilder(@NonNull final PlatformComponents.PlatformComponentsBuilder builder,
```
iamgusain: private.

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: This and fillBuilder is required by AndroidBrokerPlatformComponents.

(well... technically not required, but I would prefer not duplicating the code).

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -0,0 +1,123 @@
...
+                                   @NonNull final Context context,
+                                   @Nullable final Activity activity,
+                                   @Nullable final Fragment fragment) {
+        builder.storageEncryptionManager(new AndroidAuthSdkStorageEncryptionManager(context, null))
+                .clockSkewManager(new AndroidClockSkewManager(context))
```
iamgusain: nit: For all the different platform Components can we have similar suffix for consistency in naming.
e.g. 
```
AndroidClockSkewManager
AndroidBroadcastManager
AndroidPoPManager
AndroidStorageManager
AndroidNetworkingManager
```
Also, it would be nice to move methods out of AndroidPlatformUtil to individual component manager classes. (maybe in future)

common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageLoader.java
```diff
@@ -0,0 +1,87 @@
...
+import com.microsoft.identity.common.java.crypto.IKeyAccessor;
+
+import javax.annotation.Nullable;
+
+import lombok.NonNull;
+
+/**
+ * An interface for loading {@link INameValueStorage}
+ */
+public interface IStorageLoader {
```
iamgusain: nit: IStorageManager

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
iamgusain: protected?

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
shahzaibj: I don't really understand the purpose of this method. Can you add some javadoc?
Also is this meant to be `public`?

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -0,0 +1,123 @@
...
+                .broadcaster(new AndroidBroadcaster(context))
+                .popManagerLoader(new AndroidPopManagerLoader(context))
+                .storageLoader(new AndroidStorageLoader(context))
+                .platformUtil(new AndroidPlatformUtil(context, activity))
+                .httpClientWrapper(new DefaultHttpClientWrapper());
+
+        if (activity != null){
```
shahzaibj, line 129: Should Activity rather be required?

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.HashMap;
+
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+
+@AllArgsConstructor
+public class AndroidPopManagerLoader implements IPopManagerLoader {
+    private final Context mContext;
+
+    private final HashMap<String, IDevicePopManager> mCachedPopManager = new HashMap<>();
```
shahzaibj: cached per alias it seems

common4j/src/main/com/microsoft/identity/common/java/interfaces/IPopManagerLoader.java
```diff
@@ -0,0 +1,50 @@
...
+import com.microsoft.identity.common.java.crypto.IDevicePopManager;
+import com.microsoft.identity.common.java.exception.ClientException;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.NonNull;
+
+/**
+ * An interface for loading {@link IDevicePopManager}
+ */
+public interface IPopManagerLoader {
```
shahzaibj: Loader to me seems like something that loads something from a storage of some sort. This is rather just instantiating some objects.

I would probably rename these to `IPopManagerProvider` or `IPopManagerSupplier` because that's what these are doing.

The same feedback to all the places similar model is followed.

common4j/src/main/com/microsoft/identity/common/java/interfaces/PlatformComponents.java
```diff
@@ -0,0 +1,168 @@
...
+    private IAuthorizationStrategyFactory mAuthorizationStrategyFactory;
+
+    @Nullable
+    private IStateGenerator mStateGenerator;
+
+    @NonNull
+    private IPlatformUtil mPlatformUtil;
+
+    @NonNull
+    private IHttpClientWrapper mHttpClientWrapper;
```
shahzaibj: Can we make these `final`?

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -0,0 +1,123 @@
...
+                                   @NonNull final Context context,
+                                   @Nullable final Activity activity,
+                                   @Nullable final Fragment fragment) {
+        builder.storageEncryptionManager(new AndroidAuthSdkStorageEncryptionManager(context, null))
+                .clockSkewManager(new AndroidClockSkewManager(context))
+                .broadcaster(new AndroidBroadcaster(context))
+                .popManagerLoader(new AndroidPopManagerLoader(context))
+                .storageLoader(new AndroidStorageLoader(context))
+                .platformUtil(new AndroidPlatformUtil(context, activity))
+                .httpClientWrapper(new DefaultHttpClientWrapper());
```
shahzaibj: If we invoke one of the create methods in this class such as `createFromActivity` multiple (n) times, then I believe that leads to n number of objects created for each of these i.e. `AndroidBroadcaster`, `AndroidPopManagerLoader` etc.?

Do we want that?

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
shahzaibj: It seems that you are trying to gain some perf benefits from caching these and not having to re-initialize these?

I don't think it is going to be too helpful given that this is a member variable. The cache is going to become per instance of `AndroidPopManagerLoader`. I think you rather want this to be a `static` cache.

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public @NonNull IDevicePopManager getDevicePopManager(@Nullable String alias) throws ClientException {
+        final IDevicePopManager cachedManager = mCachedPopManager.get(alias);
+        if (cachedManager != null) {
+            return cachedManager;
+        }
+
+        final IDevicePopManager popManager = initializeDevicePopManager(alias);
+        mCachedPopManager.put(alias, popManager);
+        return popManager;
```
shahzaibj:
```suggestion
        return mCachedPopManager.computeIfAbsent(alias, k -> {
            return initializeDevicePopManager(alias);
        });
```

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
```diff
@@ -0,0 +1,102 @@
...
+import lombok.NonNull;
+
+@AllArgsConstructor
+public class AndroidPopManagerLoader implements IPopManagerLoader {
+    private final Context mContext;
+
+    private final HashMap<String, IDevicePopManager> mCachedPopManager = new HashMap<>();
+
+    @Override
+    public @NonNull IDevicePopManager getDefaultDevicePopManager() throws ClientException {
```
shahzaibj: This entire implementation of this method should probably move to `IPopManagerLoader` as a `default` implementation.

common/src/main/java/com/microsoft/identity/common/components/AndroidStorageLoader.java
```diff
@@ -0,0 +1,142 @@
...
+@AllArgsConstructor
+public class AndroidStorageLoader implements IStorageLoader {
+    private final Context mContext;
+
+    @Override
+    @NonNull
+    public <T> INameValueStorage<T> getNameValueStore(@NonNull final String storeName,
+                                                      @NonNull final Class<T> clazz) {
+        return getEncryptedNameValueStore(storeName, null, clazz);
+    }
```
shahzaibj: This can move to `IStorageLoader` as `default` implementation - that way this can be leveraged in Linux as well.

common/src/main/java/com/microsoft/identity/common/components/AndroidStorageLoader.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+        throw new UnsupportedOperationException("Only Long and String are natively supported as types");
+    }
+
+    @Override
+    @NonNull
+    public IMultiTypeNameValueStorage getEncryptedFileStore(@NonNull final String storeName,
+                                                            @NonNull final IKeyAccessor helper) {
+        return SharedPreferencesFileManager.getSharedPreferences(mContext, storeName, helper);
+    }
```
shahzaibj, line 76: This can move to `IStorageLoader` as `default` implementation - that way this can be leveraged in Linux as well.

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: I made AndroidBrokerPlatformComponentFactory invoke this method when it constructs its own object.

Didn't make them inherit each other since they're static factory classes - so that's why i made this method public.

rpdome: javadoc added.

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: not for silent flow.

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: AndroidBrokerPlatformComponentFactory does not inherit from AndroidPlatformComponentFactory.

(tried. doesn't make much sense since they don't share the exact same factory method signatures)

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: Meant to be invoked by the broker factory as well. will add javadoc.

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
rpdome: I think it's okay _for the scope of the factory_.

If we want to optimize, we should be optimizing this on the higher level - making sure that we only create it once per flow. 

Selectively (statically) cache some objects could lead to weird behavior, if someone modify the code such that the cached class is no longer sharable.

(On a side note, these classes take a context as parameter, and I'm not sure if having static objects that rely on a certain context object is going to be safe)

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
rpdome: API 24 required lol

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
rpdome: I think i'll remove this from this PR. Don't want to prematurely optimize this.

If we want to make a static cache (or any cache)... we probably want to have a separate PR with some stress test to make sure it's not going to behave erratically.

rpdome: removed.

common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageLoader.java
rpdome: will rename to supplier as suggested by shahzaib below.

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerLoader.java
shahzaibj: cc/ @negoe  Since she is trying to drive us move minSdk to 24

We are already there for Brokers, just need to figure it out for MSAL given that common lib is shared.

shahzaibj (approved):
Thanks Dome!

common/src/main/java/com/microsoft/identity/common/components/AndroidPopManagerSupplier.java
```diff
@@ -0,0 +1,83 @@
...
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+
+import lombok.AllArgsConstructor;
+import lombok.NonNull;
+
+@AllArgsConstructor
+public class AndroidPopManagerSupplier implements IPopManagerSupplier {
+    private final Context mContext;
```
iamgusain, line 51: @NonNull

iamgusain: With Lombok AllArgsConstructor, it is important that we specify the nullability of the members this way, 
Otherwise, we should explicitly write the constructor to not accidently miss it.

rpdome: added here and in the StorageSupplier

iamgusain (approved):

PR #1894, opened 2022-11-01
Fixes MSAL Issue #1715

### LINKED ISSUE #1715: Failed resolution of Lio/opentelemetry/api/trace/Span
I have an error in the authentication process of an account, it occurs with msal version 4.0.3 (and also 4.0.4), but not with 4.0.2

We have a two step authentication process as we utilize both EWS and Graph API.
The authentication with the ews scope is successful, afterwards we try to get a second access token for the Graph API, which fails.

The account in question is missing administrator consent to a new graph scope, User.Read.All, that we added recently.
If i use version 4.0.2 the oauth webpage shows the missing administrator consent (correct), but if i change to 4.0.3 or 4.0.4 i get a MsalClientException with mErrorCode unknown_error and detailMessage "failed resolution of Lio/opentelemetry/api/trace/Span"

see [screenshot](https://i.imgur.com/7RXtsTZ.png)

The authentication process is successful for other accounts where the new graph permission/scope was already granted.
It is also successful if i remove the new scope.

To summarize: Adding a new scope to graph on an account where it needs administrator consent causes an error in msal since 4.0.3

shahzaibj:
Fix https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1715

This basically provides a custom, safe implementation of `Span.current()` method to be used in MSAL for scenarios where minSdkVersion of calling application < 24. The reason for this is because the default no-op implementation uses "static interface methods" and these get left out of the APK for applications whose MIN SDK is < 24 and minification through R8 is DISABLED because the consumers-rules are NOT honored when minification is disabled and R8 applies some default proguard rules that leaves static interface methods out of the APK. This causes a "NoSuchMethodError" when such methods are invoked.

We are just catching the exception and return the Noop based on that because there isn't a way to know about application's minSdk AND their minification settings. 

This is not a problem for Broker Hosting applications as the MIN SDK for those is already >= 24. The issue only arises for the some MSAL consumers falling into the above situation. Tracing is disabled by default anyway for MSAL (and only turned on for Broker) and Open Telemetry uses its own Noop implementations, however, here we are just providing our own that DON'T use static interface methods.

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2047

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@ V.Next
 - [MINOR] Leverage Otel Utility create span method (#1877)
 - [MINOR] Add Open Telemetry explicitly in common as transitive is set to false (#1882)
 - [MINOR] Add open telemetry to consumer rules (#1883)
+- [PATCH] Fixes MSAL Issue #1715 (#1894)
 
 V.8.0.3
 ----------
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/NoopSpan.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/NoopSpan.java
@@ -51,7 +51,42 @@ public class NoopSpan implements Span {
     private final SpanContext spanContext;
 
     @Override
+    public Span setAttribute(String key, String value) {
+        return this;
+    }
+
+    @Override
+    public Span setAttribute(String key, long value) {
+        return this;
+    }
+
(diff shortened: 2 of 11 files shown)


gitaumoses4 (approved):
Nice.

wainaina (approved):

PR #1895, opened 2022-11-01
Add support to reset broadcast Executor service

sansrivastav:
Currently there is no way of synchronizing between MSAL CPP code and Android common broadcast service. 
We have seen MSAL Android nightly tests fail when the broadcast to cancel a web flow is delayed and in the meantime another test starts executing.
The cancel broadcast is triggered after the first test has ended and the second test starts executing resulting in the second test failing.
Logs for this case can be found here: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1990045

So, with the proposed change we intend to call the resetBroadcast() method before starting a new test to make sure that the stale broadcast does not interfere with the new test.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
@@ -82,6 +82,9 @@ public enum LocalBroadcaster {
         });
     }
 
+    /**
+     * Resets the broadcast executor service.
+     */
     public void resetBroadcast() {
         shutdownAndAwaitTerminationForBroadcasterService();
         sBroadcastExecutor = Executors.newSingleThreadExecutor();
--- a/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
@@ -27,6 +27,7 @@ import com.microsoft.identity.common.java.logging.Logger;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import lombok.NonNull;
 
@@ -34,7 +35,7 @@ public enum LocalBroadcaster {
     INSTANCE;
 
     private static final String TAG = LocalBroadcaster.class.getSimpleName();
-    private static final ExecutorService sBroadcastExecutor = Executors.newSingleThreadExecutor();
+    private static ExecutorService sBroadcastExecutor = Executors.newSingleThreadExecutor();
 
     public interface IReceiverCallback {
         void onReceive(@NonNull final PropertyBag propertyBag);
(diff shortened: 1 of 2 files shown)


iamgusain:
Do you know the reason for the delay in broadcast?
Is it possible that this is happening because the callback is unregistered before the broadcast happens?

sansrivastav:
> Do you know the reason for the delay in broadcast? Is it possible that this is happening because the callback is unregistered before the broadcast happens?

The reason for delay is not very clear. It can be because Android synchronizes broadcast for power optimization.
Callback is not unregistered, and we can see this in the log flows clearly.

iamgusain:
> > Do you know the reason for the delay in broadcast? Is it possible that this is happening because the callback is unregistered before the broadcast happens?
> 
> The reason for delay is not very clear. It can be because Android synchronizes broadcast for power optimization. Callback is not unregistered, and we can see this in the log flows clearly.

It could be that there are multiple broadcast requests, and they are being queued, as this is a SingleThreadExecutor.
Does adding some delay in the test help?

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -80,4 +81,25 @@ public void run() {
             }
         });
     }
+
+    public void resetBroadcast() {
```
iamgusain: Can you add Javadoc to this method

sansrivastav: done

sansrivastav:
> > > Do you know the reason for the delay in broadcast? Is it possible that this is happening because the callback is unregistered before the broadcast happens?
> > 
> > 
> > The reason for delay is not very clear. It can be because Android synchronizes broadcast for power optimization. Callback is not unregistered, and we can see this in the log flows clearly.
> 
> It could be that there are multiple broadcast requests, and they are being queued, as this is a SingleThreadExecutor. Does adding some delay in the test help?

I have not tried adding delay because it is not certain how much delay will be sufficient. It is much cleaner to reset the executor service.

iamgusain (approved):
LGTM

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -80,4 +81,28 @@ public void run() {
...
+
+    /**
+     * Resets the broadcast executor service.
+     */
+    public void resetBroadcast() {
+        shutdownAndAwaitTerminationForBroadcasterService();
+        sBroadcastExecutor = Executors.newSingleThreadExecutor();
+    }
+
+    void shutdownAndAwaitTerminationForBroadcasterService() {
```
mohitc1: >void [](http://example.com/codeflow?start=4&length=4)

private?

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
sansrivastav: done

nickbopp:
Merging for Sandesh.

PR #1896, opened 2022-11-05
Adding NFC compatibility to CBA

melissaahn:
## Summary
This PR adds in NFC support to the smartcard CBA feature.
The user experience for CBA as a whole changes for both on-device and smartcard certificate users. Instead of smartcard plug-in status determining the user's choice of authentication, there will be a new dialog, `UserChoiceDialog`, that will let the user select if they would like to proceed with on-device or smartcard (though an existing plugged-in smartcard still remains a way to go straight to the cert picker). 
If the user decides to proceed with smartcard, they will be prompted (`SmartcardPromptDialog`) to either plug in their smartcard or press it to the back of their phone. When the user does either of these things, they will continue with that specific method of the authentication. 
The usb path remains the same, but the NFC path will include an extra smartcard tap for every PIN attempt. 

<b>Notable New files and Changes</b>:
- `SmartcardCertBasedAuthManagerFactory`
    - Instantiates managers for smartcard (usb and nfc) certificate based authentication. It can show the `UserChoiceDialog`, `SmartcardPromptDialog`, and `SmartcardNfcLoadingDialog` based on a user's choice of authentication. 
- `UserChoiceDialog`
    - Prompts user to choose either to proceed with smartcard or on-device CBA. This dialog shows up if a user does not already have a smartcard plugged in.
- `SmartcardNfcLoadingDialog`
    - Dialog to be shown to users to notify their NFC smartcard has been connected, but they need to keep holding it to the back of their phone until processes are complete.
- `SmartcardNfcPromptDialog`
    - Dialog to be shown to users after typing in their PIN attempt. 
- `SmartcardPromptDialog`
    - Dialog shown to users after selecting smartcard as their method of auth on the `UserChoiceDialog`.
- `AbstractNfcSmartcardCertBasedAuthChallengeHandler` and `AbstractUsbSmartcardCertBasedAuthChallengeHandler`
    - Logic for handling specific authentication with nfc vs usb is separated, with `AbstractSmartcardCertBasedAuthChallengeHandler` containing shared logic.
    
```mermaid
    classDiagram
        class IChallengeHandler {
            +processChallenge(genericChallenge)
        }
        IChallengeHandler <|-- ICertBasedAuthChallengeHandler
        class ICertBasedAuthChallengeHandler{
            +emitTelemetryForCertBasedAuthResults(response)
            +cleanUp()
        }
        ICertBasedAuthChallengeHandler <|-- AbstractSmartcardCertBasedAuthChallengeHandler
        class AbstractSmartcardCertBasedAuthChallengeHandler{
            #Activity mActivity
            #T mCbaManager
            #DialogHolder mDialogHolder
            #CertBasedAuthTelemetryHelper mTelemetryHelper 
            #boolean mIsSmartcardCertBasedAuthProceeding

            #onPausedSmartcardDiscovery()
            #promptTooManyFailedPinAttempts(methodTag)
            #indicateGeneralException(methodTag, e)
            -getSmartcardCertPickerDialogPositiveButtonListener(request)
            #getSmartcardPinDialogPositiveButtonListener(certDetails, request)
            #tryUsingSmartcardWithPin(pin, certDetails, request, session)
            #setPinDialogForIncorrectAttempt(certDetails, request)
            #useSmartcardCertForAuth(certDetails, pin, session, request)
            #clearPin(pin)
        }
        AbstractSmartcardCertBasedAuthChallengeHandler <|-- UsbSmartcardCertBasedAuthChallengeHandler
        AbstractSmartcardCertBasedAuthChallengeHandler <|-- NfcSmartcardCertBasedAuthChallengeHandler
        class UsbSmartcardCertBasedAuthChallengeHandler{
            #AbstractUsbSmartcardCertBasedAuthManager mCbaManager
        }  
        class NfcSmartcardCertBasedAuthChallengeHandler{
            #AbstractNfcSmartcardCertBasedAuthManager mCbaManager
        }  
    
```
- `AbstractNfcSmartcardCertBasedAuthManager` and `AbstractUsbSmartcardCertBasedAuthManager`
    - Logic for managing discovery and connections is shared by the base class `AbstractSmartcardCertBasedAuthManager`.
    - Implementations of both were created for YubiKit specificially.
```mermaid
    classDiagram
        class AbstractSmartcardCertBasedAuthManager {
            #T mConnectionCallback

            +startDiscovery(activity)
            +stopDiscovery(activity)
            +requestDeviceSession(callback)
            +isDeviceConnected()
            +initBeforeProceedingWithRequest(telemetryHelper)
            +onDestroy(activity)
            +setConnectionCallback(connectionCallback)
            +clearConnectionCallback()
        }

        AbstractSmartcardCertBasedAuthManager *-- ISessionCallback
        AbstractSmartcardCertBasedAuthManager <|-- AbstractUsbSmartcardCertBasedAuthManager
        AbstractSmartcardCertBasedAuthManager <|-- AbstractNfcSmartcardCertBasedAuthManager

        class ISessionCallback {
            +onGetSession(session)
            +onException(e)
        }

        class AbstractUsbSmartcardCertBasedAuthManager {
            #IUsbConnectionCallback mConnectionCallback
        }

        AbstractUsbSmartcardCertBasedAuthManager <|-- YubiKitUsbSmartcardCertBasedAuthManager
        
        class YubiKitUsbSmartcardCertBasedAuthManager{
            -UsbYubiKeyManager mUsbYubiKeyManager
            -UsbYubiKeyDevice mUsbDevice
            -Object sDeviceLock

            +getPivProviderCallback()
        }
        
        class AbstractNfcSmartcardCertBasedAuthManager {
            #IConnectionCallback mConnectionCallback
            #boolean isDeviceChanged
        }
        AbstractNfcSmartcardCertBasedAuthManager <|-- YubiKitNfcSmartcardCertBasedAuthManager
        class YubiKitNfcSmartcardCertBasedAuthManager{
            -NfcYubiKeyManager mNfcYubiKitManager
            -NfcYubiKeyDevice mNfcDevice
            -mbyte[] mTagId
            -Object sDeviceLock

            +getPivProviderCallback()
        }  
```
- Telemetry has been added for CBA using OpenTelemetry.

<b>Other PRs merged into this one</b>:
- [x] Implement and add additional dialog that informs users to turn on NFC capabilities on their device if they would like to use NFC. This dialog will be shown right before the `SmartcardPromptDialog`, if it is detected that the user's device has NFC disabled.
    - https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1899
- [x] Add Telemetry to CBA.
    - https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1903
- [x] Replace strings with final versions from the UX designer. 
    - https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1918


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,18 +45,21 @@ public class CertBasedAuthFactory {
...
+            return;
+        }
         mSmartcardCertBasedAuthManager.setDiscoveryExceptionCallback(new AbstractSmartcardCertBasedAuthManager.IDiscoveryExceptionCallback() {
             @Override
-            public void onException() {
+            public void onException(@NonNull final Exception exception) {
                 //Logging, but may also want to emit telemetry.
                 //This method is not currently being called, but it could be
                 // used in future SmartcardCertBasedAuthManager implementations.
                 Logger.error(methodTag, "Unable to start smartcard usb discovery.", null);
```
fadidurah: Perhaps include part of the exception in the log

rpdome:
I'm wondering if eSTS can send us a hint if the CBA is for smart card or on-device. (not sure if it has such information during provisioning).

If yes, maybe we can use such hint to determine what UX we should display.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,18 +45,21 @@ public class CertBasedAuthFactory {
     public CertBasedAuthFactory(@NonNull final Activity activity) {
```
rpdome: Wondering if we can have 3 managers.
1. USB.
2. NFC.
3. On-device.

Each flow, irrespective of device manufacturer, should have its own UX flow. You can extract and reuse some implementation among them (either via subclassing, or using the [strategy pattern](https://www.tutorialspoint.com/design_pattern/strategy_pattern.htm)).

---

and this is sorted based on the order of preferences. 

1. keep asking the user to tap (NFC) or plug in the device (USB) - at this time you can start the discovery for both. No need to explicitly ask the user if they're going to use NFC or USB.
2. If device is detected via whatever means, go that route.
3. If the user explicitly request on-device (this could be the only button on the UI), then fall back to on-device.

maybe the CBA selection dialog could stay in createCertBasedAuthChallengeHandler?

rpdome: with this, we wouldn't need startUsbDiscovery() and startNfcDiscovery() - just startDiscovery() would suffice.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitCertBasedAuthManager.java
```diff
@@ -146,35 +193,61 @@ public void stopDiscovery() {
...
+                            final SmartCardConnection c = value.getValue();
+                            final PivSession piv = new PivSession(c);
+                            final YubiKitSmartcardSession session = new YubiKitSmartcardSession(piv);
+                            callback.onGetSession(session);
+                        } catch (final Exception e) {
+                            callback.onException(e);
+                        }
+                    }
+                });
+            } else if (isNfcDeviceConnected()) {
```
rpdome: when a class big if-else like this all over the place, IMO it's a sign that it needs to be splitted.

rpdome (commented):
left some suggestions.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: Splitting up the SmartcardCertBasedAuthManager into two makes sense to me, and correspondingly, I'll split up the SmartcardCertBasedAuthChallengeHandler into two as well. The USB manager will remain in CertBasedAuthFactory, but I think the NFC manager can be created in UserChoiceCertBasedAuthChallengeHandler, since that's where the first NFC discovery will always start. USB discovery needs to remain active for the duration of the WebView being shown, while NFC discovery will only be on whenever we explicitly ask the user to tap.

In response to the second part of the feedback, the design you're proposing is similar to the initial sample I showed to the UX designer, and it was rejected in favor of having that additional dialog where the user first chooses between general smartcard or on-device. Choosing the smartcard option will bring the user to a dialog like 1. described above, but with a cancel button. The UI flow as shown in the video I attached is what the designer has approved.

Also, as I recall in my previous refactoring PR, we were trying to keep control of the dialogs (via DialogHolder) separate from CertBasedAuthFactory. I also think that, since there are at least 3 dialogs involved with the process of choosing a type of CBA, it's been helpful to have a separate challenge handler to take care of those dialogs and then route the ClientCertRequest to a more specific challenge handler.

rpdome: I don't see any attached video. Would you mind sharing the link?

melissaahn: I'll send thru chat

rpdome: actually never mind. just saw it in the PR review chat room.

rpdome: After watching the video... here's my feedback.

- We probably want to handle the flow selection in this class. Once it's decided, there is no way for the user to change it. This means that the first dialog, the "please tap" dialog and the cancel dialog (or any UI that would let the user switch between the flow) should probably be handled here. (you can still use DialogHolder, I meant the wiring part should be done here)

- createCertBasedAuthChallengeHandler might be turned into an asynchronous method (taking in a callback object) - which eventually trigger one of the three flows, USB, NFC, On-Device. Maybe it could perform asynchronous polling to check if any devices (of any manufacturers, not just yubikey) is detected (or if the user clicks "cancel"), then proceed with a corresponding challengeHandler + CBA Manager

With the current structure, I do see some problems,  i.e.
- What would happen if the user plugs in a non-NFC yubikey when we ask the user to tap?
- What if we need to support more than YubiKey?

melissaahn: To answer the two concerns:
- So I think I've tested this scenario in the past, but testing it now, when plugging in the key, the plugging in acts as a tap lol. I believe what I had before (but somehow got lost) was an if statement that only runs the connection callback if  `isNfc` parameter is true. The result of that would be if the user plugs in a non-NFC yubikey while the user is already on the NFC path, the consent dialog pops up, but nothing else should happens, and the user can unplug and still proceed with authentication via NFC. 
- The manager classes will hold lists of the smartcard-specific managers and will control which discoveries are on and off. I believe the USB discovery method of YubiKit will only call its callback after identifying the connected device as a YubiKey. I'm not as certain for NFC, so that's something I'll need to follow up with Yubico with. 
Best case scenario, supposing other companies would do the same thing for both USB and NFC, we would be able to keep all discoveries on in the "please tap" dialog, make a note of the initial manager that made this connection, and then use this manager when it comes to reading future information off the smartcard. 
If the other smartcard brands cannot or refuse to implement a vendor_id check, and supposing we also won't be able to implement that ourselves for NFC... then we would either need some sort of hint from eSTS (not certain this would be possible) or (more likely) I will need to create an additional dialog that asks the user which smartcard vendor they're working with. 

In terms of moving the code in UserChoiceCertBasedAuthChallengeHandler to CertBasedAuthFactory, I understand logically why we would do this, so despite my hesitancy with introducing UI as well as telemetry (what I'm working on rn) into it, I'll make the changes.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -461,8 +462,18 @@ public void onReceivedClientCertRequest(@NonNull final WebView view,
         if (mCertBasedAuthChallengeHandler != null) {
             mCertBasedAuthChallengeHandler.cleanUp();
         }
-        mCertBasedAuthChallengeHandler = mCertBasedAuthFactory.createCertBasedAuthChallengeHandler();
-        mCertBasedAuthChallengeHandler.processChallenge(clientCertRequest);
+        mCertBasedAuthFactory.createCertBasedAuthChallengeHandler(new CertBasedAuthFactory.CertBasedAuthChallengeHandlerCallback() {
+            @Override
+            public void onReceived(@Nullable ICertBasedAuthChallengeHandler challengeHandler) {
```
rpdome: nit: final

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -23,59 +23,232 @@
...
+
+    /**
+     * Asynchronously chooses and returns a ICertBasedAuthChallengeHandler.
+     * @param callback logic to run after a ICertBasedAuthChallengeHandler is chosen.
+     */
+    public void createCertBasedAuthChallengeHandler(@NonNull final CertBasedAuthChallengeHandlerCallback callback) {
+        final CertBasedAuthTelemetryHelper telemetryHelper = new CertBasedAuthTelemetryHelper();
+        telemetryHelper.setUserChoice(CertBasedAuthChoice.NON_APPLICABLE);
+        telemetryHelper.setCertBasedAuthChallengeHandler(NON_APPLICABLE);
+        if (mUsbSmartcardCertBasedAuthManager == null) {
```
rpdome: This if block (and maybe the comment from line 59-61) doesn't make much sense to me.  (Not sure if we've talked about this - it's been a while 😅)

If the device doesn't support USB - why are we blocking NFC flow? We can probably just proceed to the user choice dialog.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -23,59 +23,232 @@
...
+                    mUsbSmartcardCertBasedAuthManager,
+                    mDialogHolder,
+                    telemetryHelper));
+            return;
+        }
+
+        if (mNfcSmartcardCertBasedAuthManager != null
+                && mNfcSmartcardCertBasedAuthManager.startDiscovery(mActivity)) {
+            //Inform user to turn on NFC if they want to use NFC.
+            mDialogHolder.showSmartcardNfcReminderDialog(new SmartcardNfcReminderDialog.DismissCallback() {
```
rpdome: can this be consolidated with showSmartcardPromptDialog? (remove line 151-170, and add the NFC message in showSmartcardPromptDialog()

rpdome: (NFC is supposed to be easy to use. I would just use the USB if i have to click through 3 prompts + 1 pin prompt every time lol)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -23,59 +23,232 @@
...
+        });
+
+        if (mNfcSmartcardCertBasedAuthManager == null) {
+            return;
+        }
+        mNfcSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
+            @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+            @Override
+            public void onCreateConnection() {
+                mUsbSmartcardCertBasedAuthManager.setConnectionCallback(null);
```
rpdome: maybe we could create a "clearSmartcardCBACallback()" method and just invoke it in line 193 and here.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,170 @@
...
+                                                     @NonNull final CertBasedAuthTelemetryHelper telemetryHelper) {
+        super(activity, nfcSmartcardCertBasedAuthManager, dialogHolder, telemetryHelper, NfcSmartcardCertBasedAuthChallengeHandler.class.getSimpleName());
+    }
+
+    /**
+     * Additional logic to run upon before the end of processChallenge.
+     */
+    @Override
+    protected void onGetSessionFinished() {
+        mSmartcardCertBasedAuthManager.stopDiscovery(mActivity);
```
rpdome: suggestion: can rename mSmartcardCertBasedAuthManager to just mCbaManager. I think it already give enough context.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,170 @@
...
+                        mDialogHolder.dismissDialog();
+                        mTelemetryHelper.setResultFailure(USER_CANCEL_MESSAGE);
+                        request.cancel();
+                    }
+                });
+                mSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
+                    @Override
+                    public void onCreateConnection() {
+                        mDialogHolder.showDialog(new SmartcardNfcLoadingDialog(mActivity));
+                        if (mSmartcardCertBasedAuthManager instanceof AbstractNfcSmartcardCertBasedAuthManager
```
rpdome: Suggestion: could use generic to avoid this check

```Java
abstract class AbstractSmartcardCertBasedAuthChallengeHandler<T extends AbstractNfcSmartcardCertBasedAuthManager> { 
       T mCbaManager;
}

public class NfcSmartcardCertBasedAuthChallengeHandler extends AbstractSmartcardCertBasedAuthChallengeHandler<AbstractNfcSmartcardCertBasedAuthManager>{
...
}
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractNfcSmartcardCertBasedAuthManager.java
```diff
@@ -0,0 +1,42 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.ui.webview.challengehandlers;
+
+/**
+ * An abstract smartcard CertBasedAuth manager specifically for NFC.
+ */
+public abstract class AbstractNfcSmartcardCertBasedAuthManager extends AbstractSmartcardCertBasedAuthManager {
+
+    protected boolean mIsCurrDiffFromPrevDevice;
```
rpdome: isDeviceChanged?

rpdome: (also, you can use lombok's `@Getter` in line 29 and remove line 38-40

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthManager.java
```diff
@@ -96,7 +106,7 @@ interface IConnectionCallback {
         void onCreateConnection();
 
         /**
-         * Logic to be run upon disconnection of a smartcard device.
+         * Logic to be run upon disconnection of a smartcard device via usb.
```
rpdome: Maybe we should extract this into a separate interface, and expose it only in the usb manager?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -23,59 +23,232 @@
...
     /**
      * Creates an instance of CertBasedAuthFactory.
      * Instantiates relevant implementations of ISmartcardCertBasedAuthManagers.
      * @param activity current host activity.
      */
     public CertBasedAuthFactory(@NonNull final Activity activity) {
-        final String methodTag = TAG + ":CertBasedAuthFactory";
         mActivity = activity;
-        mSmartcardCertBasedAuthManager = new YubiKitCertBasedAuthManager(mActivity.getApplicationContext());
-        mSmartcardCertBasedAuthManager.setDiscoveryExceptionCallback(new AbstractSmartcardCertBasedAuthManager.IDiscoveryExceptionCallback() {
```
rpdome, line 49: do we still need this setDiscoveryExceptionCallback()?

If not, remove it. we can add it back later when needed.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -23,59 +23,232 @@
...
+        mUsbSmartcardCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createUsbSmartcardCertBasedAuthManager(mActivity.getApplicationContext());
+        mNfcSmartcardCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createNfcSmartcardCertBasedAuthManager(mActivity.getApplicationContext());
+        mDialogHolder = new DialogHolder(mActivity);
+        if (mUsbSmartcardCertBasedAuthManager == null) {
```
rpdome:
```Java
        if (mUsbSmartcardCertBasedAuthManager == null) {
               mUsbSmartcardCertBasedAuthManager.startDiscovery(activity);
        }
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,170 @@
...
+                    @Override
+                    public void onCancel() {
+                        mDialogHolder.dismissDialog();
+                        mTelemetryHelper.setResultFailure(USER_CANCEL_MESSAGE);
+                        request.cancel();
+                    }
+                });
+                mSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
+                    @Override
+                    public void onCreateConnection() {
```
rpdome: do we ever have to remove the callback once the connection is established? (i.e. if the connection is intermittent, what would happen if onCreateConnection() is invoked multiple time in succession?).

rpdome: (maybe we could turn onCreateConnection setter to clear itself once the callback is fired).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/UserChoiceDialog.java
```diff
@@ -0,0 +1,118 @@
...
+        createDialog();
+    }
+
+    /**
+     * Builds an Android Dialog object and set it to mDialog.
+     */
+    @Override
+    void createDialog() {
+        //Create array for auth choices to pass into the AlertDialog.
+        ArrayList<String> choicesList = new ArrayList<>();
```
rpdome: nit: final.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/UsbSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,149 @@
...
+                                                     @NonNull final DialogHolder dialogHolder,
+                                                     @NonNull final CertBasedAuthTelemetryHelper telemetryHelper) {
+        super(activity, usbSmartcardCertBasedAuthManager, dialogHolder, telemetryHelper, UsbSmartcardCertBasedAuthChallengeHandler.class.getSimpleName());
+        final String methodTag = TAG + ":UsbSmartcardCertBasedAuthChallengeHandler";
+        mSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
+            @Override
+            public void onCreateConnection() {
+                //Reset DialogHolder to null if necessary.
+                //In this case, DialogHolder would be an ErrorDialog if not null.
+                mDialogHolder.dismissDialog();
```
rpdome: would this code ever be triggered? 

If we ever land here, I wonder what would the user see (and later proceed).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/UsbSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            @Override
+            public void onClosedConnection() {
+                if (mDialogHolder.isDialogShowing()) {
```
rpdome: even if the dialog is not being displayed, should we just terminate the flow if the device is disconnected?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,170 @@
...
+            //If pin is successfully verified, we will use the certificate to perform the rest of the logic for authentication.
+            useSmartcardCertForAuth(certDetails, pin, session, request);
+            return;
+        }
+        final int attemptsRemaining = session.getPinAttemptsRemaining();
+        mSmartcardCertBasedAuthManager.stopDiscovery(mActivity);
+        // If the number of attempts is 0, no more attempts will be allowed.
+        if (attemptsRemaining == 0) {
+            //We must display a dialog informing the user that they have made too many incorrect attempts,
+            // and the user will need to figure out a way to reset their key outside of our library.
```
rpdome: nit: I think the comment 
// If the number of attempts is 0, no more attempts will be allowed.
and
//We must display a dialog informing the user that they have made too many incorrect attempts,
// and the user will need to figure out a way to reset their key outside of our library.
can be removed, as the code is clear enough().

(maybe rename indicateTooManyFailedAttempts to promptTooManyFailedPinAttempts()?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,170 @@
...
+                                            @NonNull final ISmartcardSession session)
+            throws Exception {
+        final String methodTag = TAG + ":tryUsingSmartcardWithPin";
+        if (session.verifyPin(pin)) {
+            //If pin is successfully verified, we will use the certificate to perform the rest of the logic for authentication.
+            useSmartcardCertForAuth(certDetails, pin, session, request);
+            return;
+        }
+        final int attemptsRemaining = session.getPinAttemptsRemaining();
+        mSmartcardCertBasedAuthManager.stopDiscovery(mActivity);
```
rpdome: actually.... maybe we can just rename onGetSessionFinished() to onPausedSmartCardDiscovery().

then... make that invoke stopDiscovery() in the NFC flow, and replace with none in USB flow (since the USB flow does NOT need to pause the discovery).

if we do this, I assume we should be able to pull this method up to the base class?

rpdome: (and we can add a comment in the NFC counterpart that we need to pause discovery because when the user enters pin, they will need to move the card away from the phone. keeping the discovery on might lead to unneccessary callback trigger)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbSmartcardCertBasedAuthManager.java
```diff
@@ -0,0 +1,204 @@
...
+        });
+        return true;
+    }
+
+    /**
+     * Ceases usb discovery for YubiKeys.
+     * Usually called when a host fragment is destroyed.
+     */
+    @Override
+    void stopDiscovery(@NonNull final Activity activity) {
```
rpdome: note: let's add a comment/details on how usb's stopDiscovery() is inconsistent (for future devs to understand why stopDiscovery()'s invoking pattern is not consistent between the two flows).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbSmartcardCertBasedAuthManager.java
```diff
@@ -0,0 +1,204 @@
...
+    void onDestroy(@NonNull final Activity activity) {
+        stopDiscovery(activity);
+    }
+
+    /**
+     * Used to provide PivProvider constructor a Callback that will establish a new PivSession when it is needed.
+     * @return A Callback which returns a Callback that will return a new PivSession instance.
+     */
+    @NonNull
+    Callback<Callback<Result<PivSession, Exception>>> getPivProviderCallback() {
```
rpdome: maybe this could be moved to the base class?

```Java
protected Callback<Callback<Result<PivSession, Exception>>> getPivProviderCallback(YubiKeyDevice device) {
...
     yubikeyDevice.requestConnection()
{
```

rpdome (approved):
I'm okay with the code in general. (I do think the code could be made cleaner. gut feeling. That said, I'm still not good enough to figure out how to improve it further tbh. Need to dig more. If I stumble on something I will let you know). 

Left some suggestions (mostly to improve readability).

Otherwise, thanks for bearing with my comments. Good job :)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthManager.java
melissaahn: I'll extract `IConnectionCallback` into its own Interface that only has onCreateConnection and create a subinterface called `IUsbConnectionCallback` which includes onClosedConnection.

melissaahn: I think I'll haveto also add generics to AbstractSmartcardCertBasedAuthManager for the usb subtype to take in IUsbConnectionCallback instead

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: I'm thinking `==` is supposed to be `!=` above?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: You're probably right. I'll make the change suggested in the constructor and add the necessary null checks for `mUsbSmartcardCertBasedAuthManager`.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: This nfc reminder dialog is only shown when the user has nfc turned off in settings. So once the user turns it on, they won't see this dialog anymore.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: created `clearConnectionCallback` method in `AbstractSmartcardCertBasedAuthManager` and added it under stopDiscovery for NFC.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/UsbSmartcardCertBasedAuthChallengeHandler.java
melissaahn: This will be triggered when a user either cancels or fails out of CBA and sees that "authentication failed" screen with an error dialog on top. If the user plugs in again here, without closing the dialog manually, then we'll close the dialog for them. 
This code doesn't do much, but I think it improves the user exp just a little bit

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/UsbSmartcardCertBasedAuthChallengeHandler.java
melissaahn: If the cba flow is active, a dialog should always be showing. So there would be no need to do anything upon closed connection in the situation described in the comment just above this one.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbSmartcardCertBasedAuthManager.java
melissaahn: Just for more context, since I didnt necessarily update the description, the abstract classes for manager are now based on usb vs nfc instead of by smartcard brand, so the abstract class for YubiKit in general doesn't exist anymore. This change was done after logic needed to be added for determining if a previousy connected device was the same as the current, only for nfc and not usb.
I could maybe move this to the 'YubiKeyPivProviderManager' class though, and change the parameters of `requestConnection` based on an `instanceof` check (they change for nfc vs usb)

melissaahn: That will basically just create a method that combines both `getPivProviderCalback` methods though, with an `if` statement. Esp since I would have to pass `YubiKeyDevice` as a parameter, not sure if it's an improvement? I think this is potentially what I had before I did much of the Usb vs NFC refactoring. Let me know if I'm misunderstanding though.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: I'll still do the renaming because I think that makes sense, but this method still can't be pulled up to the base class with the suggested change alone because of the extra showPinDialog call in the NFC flow vs USB. This extra call is necessary because the NFC flow would be currently showing a different dialog (either "tap smartcard" or "keep steady", whatever the phrasing is for those) and it must go back to the pin dialog BEFORE setPinDialogErrorMode is called on the dialog holder. USB stays on the PIN dialog until the pin is correct or the flow cancels out. 

I think what I'll do is create some sort of similar extra logic method to hold the extra showPinDialog for NFC, maybe called "setPinDialogForIncorrectAttempt", and include the setPinDialogErrorMode method at the end in both (for completion, since it seems ... too custom to just call that method in the middle).  Both of the changes combined show allow the tryUsingSmartcardWithPin method to be brought up.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: I think this could be done for NFC, but not needed for USB. So a method to clear itself would only be applied to NFC classes and not base ones. That's something I'll look into.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbSmartcardCertBasedAuthManager.java
```diff
@@ -0,0 +1,208 @@
...
+import com.yubico.yubikit.core.util.Callback;
+import com.yubico.yubikit.core.util.Result;
+import com.yubico.yubikit.piv.PivSession;
+
+import java.io.IOException;
+import java.security.PrivateKey;
+import java.security.cert.X509Certificate;
+import java.util.concurrent.Callable;
+
+public class YubiKitUsbSmartcardCertBasedAuthManager extends AbstractUsbSmartcardCertBasedAuthManager {
```
fadidurah, line 49: java doc?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardPromptDialog.java
```diff
@@ -0,0 +1,102 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.ui.webview.challengehandlers;
```
fadidurah: Would it be good to create a separate package for smart card (or CBA in general) classes?

fadidurah: We can extend this idea to some other CBA classes that are added in this PR, I think

fadidurah:
@melissaahn Have you already added testing for this PR's additions?

melissaahn:
@fadidurah I'll revisit creating unit tests for the usb and nfc challenge handlers, and maybe other classes that would make sense. For manual, hopefully I can limit it to one or two tests for NFC.
I'll put the link to a PR for the unit tests here once it's ready.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardPromptDialog.java
melissaahn: Good point. See this draft PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1931/files 
I have it pointed towards the feature branch for readability, but I'd prefer to merge the draft PR into dev after this main feature PR gets moved into dev, since it's humongous as it is.

fadidurah: Sounds good

fadidurah (approved):
Signing off as testing can be included in separate PR, this one is quite big as it is. Please link the testing PR to this one once it is created.

melissaahn:
Unit tests PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1933
Two manual tests were added to the master test plan specifically testing NFC logic.

PR #1898, opened 2022-11-09
Added android device type method to get the specific device type

somalaya:
Added helper method to get android device type from package metadata
Related PR : https://github.com/AzureAD/ad-accounts-for-android/pull/2051

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -22,8 +22,13 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.internal.platform;
 
+import android.content.Context;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
 import android.os.Build;
+import android.os.Bundle;
 
+import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.platform.AbstractDeviceMetadata;
 
 import lombok.NonNull;
@@ -34,6 +39,8 @@ import lombok.NonNull;
 public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
 
     private static final String ANDROID_DEVICE_TYPE = "Android";
+    private static final String DEVICE_TYPE = "DeviceType";
+    private static final String TAG = AndroidDeviceMetadata.class.getSimpleName();
 
     @Override
     @NonNull
@@ -63,7 +70,9 @@ public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
     }
 
     @Override
-    public @NonNull String getOsForMats() { return android.os.Build.VERSION.RELEASE; }
+    public @NonNull String getOsForMats() {
+        return android.os.Build.VERSION.RELEASE;
+    }
 
     @Override
     public @NonNull String getOsForDrs() {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -81,5 +90,28 @@ public String getDeviceModel() {
...
+
+    /**
+     * Get the android device type, i.e; if it is an nGMS teams device or a mobile device
+     *
+     * @param context {@link Context}
+     * @return device type
+     */
+    public static String getAndroidDeviceTypeFromMetadata(@NonNull final Context context) {
+        final String methodTag = TAG + " :getDeviceType";
+        String defaultDeviceType = "mobileDevice";
```
shahzaibj: Can we define this as a constant at the top of the file?

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -81,5 +90,28 @@ public String getDeviceModel() {
...
+        try {
+            final PackageManager packageManager = context.getPackageManager();
+            final ApplicationInfo appInfo = packageManager.getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+            final Bundle metaDataBundle = appInfo.metaData;
+            final String deviceType = metaDataBundle.getString(DEVICE_TYPE, defaultDeviceType);
+            Logger.verbose(methodTag, "Setting the deviceType as " + deviceType);
+            return deviceType;
+        } catch (final PackageManager.NameNotFoundException e) {
+            // Do not throw the exception to break the auth request when getting the app's telemetry
+            Logger.warn(methodTag, "Unable to find the app's package name from PackageManager.");
```
shahzaibj, line 114: Should we return unknown in this case?

shahzaibj (approved):
Left a couple of comments, but overall LGTM

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
somalaya: If I return unknown here, I should also change the default deviceType above to unknown instead of mobileDevice. I am thinking I should change it to unknown in both the places

shahzaibj: I mean we still need a way to handle regular modileDevice. So you basically need 3 possible outcomes from this API

- TeamsDevice
- RegularMobileDevice
- Unknown

As long as that can be achieved then we should be fine

somalaya: We can't get the value as regular mobile device. Because in non-nGMS teams device builds.. authenticator has a null value set for this metadata.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -81,5 +92,28 @@ public String getDeviceModel() {
...
+     */
+    public static String getAndroidDeviceTypeFromMetadata(@NonNull final Context context) {
+        final String methodTag = TAG + " :getDeviceType";
+        try {
+            final PackageManager packageManager = context.getPackageManager();
+            final ApplicationInfo appInfo = packageManager.getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+            final Bundle metaDataBundle = appInfo.metaData;
+            // If the deviceType property is not found, default it to mobile device
+            final String deviceType = metaDataBundle.getString(DEVICE_TYPE, MOBILE_DEVICE);
+            Logger.verbose(methodTag, "Setting the deviceType as " + deviceType);
```
fadidurah, line 110: Are you technically "setting" anything here?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
somalaya: No, it is just getting the value from metadata. If there is no value set, then the default value is mobileDevice

PR #1899, opened 2022-11-10
NFC Reminder Dialog (CBA)

melissaahn:
## Summary
This PR builds upon [the main NFC feature PR](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1896).
When a user is trying CBA for the first time, they may not know that they have NFC turned off, and thus won't be able to authenticate with their smartcard via NFC. 
I added logic to show a reminder dialog when a user is going forth with smartcard CBA, but it is detected that NFC is turned off. The dialog tells them to turn on NFC in settings, then retry the CBA flow. 
When the user dismisses the dialog by either clicking "got it" or pressing the back button, they will be brought to the regular smartcard prompt dialog. This ensures that users wishing to use USB only (and have NFC turned off) won't be blocked from authenticating via USB.

I tested with both msalTestApp and AzureSample by making sure the dialog came up when NFC was turned off and that the regular NFC CBA flow was successful once NFC was turned on again.


fadidurah (approved):
:shipit:

PR #1903, opened 2022-11-18
Otel for CertBasedAuth

melissaahn:
## Summary
This PR adds OTel to the latest CBA feature currently in review. 
Due to the asynchronous nature of CBA, `CertBasedAuthTelemetryHelper` was created to keep track of a Span object that can collect attributes for the duration of a CBA flow. This Span is never set as "current", as an instance of `CertBasedAuthTelemetryHelper` is passed along as a parameter instead.
Telemetry details:
Span name: CertBasedAuth
Attributes:
- `cert_based_auth_challenge_handler` (string):  Indicates which CertBasedAuthChallengeHandler was handling the CBA flow. ( can be null)
- `cert_based_auth_existing_piv_provider_present` (boolean): Indicates if PivProvider (part of YubiKit) was already present in the Security static list prior to adding a new PivProvider.
-  `cert_based_auth_user_choice` (string): Indicates which CBA flow the user intended to select. (smartcard, on-device, null)

An OK status will have no message attached, while an ERROR status will either have an exception or error message attached.
Note that because the CBA feature is housed in common, the attribute names are in `AttributeNames` located in common4j, but they also need to be in broker4j (this will be released as a PR once the updated CBA feature is ready to be merged into dev).

I've made sure that the spans and attributes are being propagated into Aria via the Aria live events as well as via Kustos queries.

[Link to NFC Feature draft PR (main feature branch)](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1896)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
@@ -41,6 +41,7 @@ public class CertBasedAuthFactory {
     private static final String USER_CANCEL_MESSAGE = "User canceled smartcard CBA flow.";
     private static final String ON_DEVICE_CHOICE = "on-device";
     private static final String SMARTCARD_CHOICE = "smartcard";
+    private static final String NON_APPLICABLE = "N/A";
     private final Activity mActivity;
     private final AbstractSmartcardCertBasedAuthManager mSmartcardCertBasedAuthManager;
     private final DialogHolder mDialogHolder;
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -73,13 +77,24 @@ public void onException(@NonNull final Exception exception) {
      * @param callback logic to run after a ICertBasedAuthChallengeHandler is chosen.
      */
     public void createCertBasedAuthChallengeHandler(@NonNull final CertBasedAuthChallengeHandlerCallback callback) {
+        final CertBasedAuthTelemetryHelper telemetryHelper = new CertBasedAuthTelemetryHelper();
+        telemetryHelper.setUserChoice(NON_APPLICABLE);
```
shahzaibj, line 81: Wouldn't this be considered `on-device`?

shahzaibj (approved):
lgtm

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: I don't think so, since the user will still be on the UserChoice dialog and have not signaled that they want to use on-device yet. Unless you're referring to the if-block starting on line 83, which I don't consider as a user choice since this means they're just not able to use smartcard CBA at all (so on-device is the default).

fadidurah (approved):
:shipit:

PR #1906, opened 2022-11-21
Support 2 new UI Test Cases, fix update scenario in Play Store

fadidurah:
**What:**
This common PR supports 2 new UI Test cases in MSAL and Broker. Also, fixes checks for the update scenario in Play Store, this was causing issues where the test would finish while the app was still installing, which would cause some unpredictable behavior during rules and specifically broker installation

MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1731
Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/2068

**Why:**
Increase UI Automation coverage, fix playstore check to make sure installation is complete before proceeding.

**How**
Update Ui in playstore to check for different buttons in install/update scenarios.

**Testing**
Test included in MSAL and Broker PRs

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
@@ -104,13 +104,6 @@ public class AzureSampleApp extends App {
                 new MicrosoftStsPromptHandler(promptHandlerParameters);
 
         microsoftStsPromptHandler.handlePrompt(username, password);
-
-        // sleep as it can take a bit for UPN to appear in Azure Sample app
-        try {
-            Thread.sleep(TimeUnit.SECONDS.toMillis(5));
-        } catch (InterruptedException e){
-            e.printStackTrace();
-        }
     }
 
     /**
@@ -132,13 +125,6 @@ public class AzureSampleApp extends App {
             // handle browser first run as applicable
             ((IApp) browser).handleFirstRun();
         }
-
-        // sleep as it can take a bit for UPN to appear in Azure Sample app
-        try {
-            Thread.sleep(TimeUnit.SECONDS.toMillis(5));
-        } catch (InterruptedException e){
-            e.printStackTrace();
-        }
     }
 
     /**
@@ -157,13 +143,16 @@ public class AzureSampleApp extends App {
      */
     public void confirmSignedIn(@NonNull final String username) {
         Logger.i(TAG, "Confirming account with supplied username is signed in..");
+
+        final UiObject signedInUser = UiAutomatorUtils.obtainUiObjectWithResourceId("com.azuresamples.msalandroidapp:id/current_user");
+
+        // Sometimes need some time for upn to appear
         try {
-            Thread.sleep(TimeUnit.SECONDS.toMillis(5));
+            Thread.sleep(TimeUnit.SECONDS.toMillis(10));
         } catch (InterruptedException e) {
             e.printStackTrace();
         }
 
-        final UiObject signedInUser = UiAutomatorUtils.obtainUiObjectWithResourceId("com.azuresamples.msalandroidapp:id/current_user");
         try {
             Assert.assertEquals("User is signed into Azure Sample App", username, signedInUser.getText());
         } catch (final UiObjectNotFoundException e) {
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
@@ -27,6 +27,7 @@ import androidx.annotation.Nullable;
 import androidx.test.uiautomator.UiObject;
 import androidx.test.uiautomator.UiObjectNotFoundException;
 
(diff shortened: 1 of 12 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/IApp.java
```diff
@@ -27,6 +27,11 @@
  */
 public interface IApp {
 
+    /**
+     * Returns the package name of the broker.
```
shahzaibj:
```suggestion
     * Returns the package name of the app.
```

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
```diff
@@ -112,6 +113,34 @@ public void signInWithSingleAccountFragment(@NonNull final String username,
...
+        Logger.i(TAG, "Signing in into Azure Sample App with Single Account Mode Fragment..");
+        // Click Sign In in Single Account Fragment
+        UiAutomatorUtils.handleButtonClick("com.azuresamples.msalandroidapp:id/btn_signIn");
+
+        if (broker == null && browser != null && shouldHandleBrowserFirstRun) {
+            // handle browser first run as applicable
+            ((IApp) browser).handleFirstRun();
+        }
+
+        // sleep as it can take a bit for UPN to appear in Azure Sample app
```
shahzaibj: We don't seem to be doing anything after the sleep...that's a little strange

shahzaibj (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
```diff
@@ -128,13 +143,16 @@ public void signOutFromSingleAccountFragment() {
      */
     public void confirmSignedIn(@NonNull final String username) {
         Logger.i(TAG, "Confirming account with supplied username is signed in..");
+
+        final UiObject signedInUser = UiAutomatorUtils.obtainUiObjectWithResourceId("com.azuresamples.msalandroidapp:id/current_user");
+
+        // Sometimes need some time for upn to appear
```
iamgusain: 10 secs seem high, does it happen in some specific scenarios?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/GoogleSettings.java
```diff
@@ -344,5 +345,22 @@ private void handleDoneButton() throws UiObjectNotFoundException {
...
     }
 
+    @Override
+    public void disableAppThroughSettings(@NonNull final String packageName) {
+        Logger.i(TAG, "Disabling app through settings: " + packageName);
+        launchAppInfoPage(packageName);
+        // This is the id for the disable button
+        handleButtonClick("com.android.settings:id/button2");
+        // Confirm disabling app
+        handleButtonClick("android:id/button1");
```
iamgusain, line 355: What does button1 do?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/SamsungSettings.java
```diff
@@ -217,4 +217,15 @@ public void setPinOnDevice(final String password) {
     public void removePinFromDevice(String pin) {
         //TODO: implement removing PIN for SAMSUNG device.
     }
+
+
+    @Override
+    public void disableAppThroughSettings(@NonNull final String packageName) {
+        //TODO: implement disabling app through settings
```
iamgusain, line 223: Should we throw UnSupportedException, so the test case will fail with this reason if run on Samsung Device.

iamgusain (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/GoogleSettings.java
fadidurah: When you click disable in settings, you get a pop up to confirm. button1 is to confirm the disable.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/device/settings/SamsungSettings.java
fadidurah: Yeah, we can do that

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/AzureSampleApp.java
fadidurah: No, it's basically to give azure sample app time to show upn as it's starting up. We can go about this a different way. Instead of finding the ui object and checking it's text, which leaves us open to checking it too quick before it loads. We can instead just search for the username text directly (or "none" if no account is logged in), and this will allow us to pass the test as soon as the username becomes apparent. I will do some testing for this.

fadidurah: Implemented these changes.

PR #1908, opened 2022-11-30
Moving ClearClientCertPreferences back to onCreate and handleBackButtonPressed

melissaahn:
## Context
The most recent PR I made regarding ClearClientCertPreferences, which has been included in common version 9.0.0, has been bringing up a spike of "ERR_CERT_DATABASE_CHANGED" errors in the OneAuth UI automation pipeline. 
This PR moves ClearClientCertPreferences back to OnCreate and handleBackButtonPressed so that ClearClientCertPreferences is only called once upon WebView creation and for every back button press, instead of multiple times per page reload, which I believe is the cause of the issue here. 

With this change, the user will not be able to do CBA via the "Other ways to sign in" link on the "Certificate validation failed" page. I've let one of the CBA PMs know that eSTS will need to change their UI so that the user is not prompted to try other ways to login when they see the "Certificate validation failed" page. 

## Related PR
- [Moving clearClientCertPreferences to onPageLoaded](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1855)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,9 @@
 V.Next
 ----------
 
-
+V.9.0.1
+----------
+- [MINOR] Moving ClearClientCertPreferences back to onCreate and handleBackButtonPressed (#1908)
 
 V.9.0.0
 ----------
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,7 @@ V.Next
 
 V.9.0.1
 ----------
-- [MINOR] Moving ClearClientCertPreferences back to onCreate and handleBackButtonPressed (#1908)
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,7 +1,9 @@
 V.Next
 ----------
 
-
+V.9.0.1
+----------
+- [MINOR] Moving ClearClientCertPreferences back to onCreate and handleBackButtonPressed (#1908)
```
shahzaibj:
```suggestion
- [PATCH] Moving ClearClientCertPreferences back to onCreate and handleBackButtonPressed (#1908)
```

fadidurah (approved):

shahzaibj (approved):

PR #1909, opened 2022-12-02
Convert crypto operation spans into metrics

shahzaibj:
This PR basically converts the instrumentation for crypto operations from Spans/Traces to Metrics. 

There is a Meter of type LongCounter that will be incremented every time an Exception occurs. 

Rationale is documented here: https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Telemetry/cost_optimization.md&version=GBshahzaibj/optimize-telemetry&_a=preview

Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2080

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
@@ -30,12 +30,19 @@ import com.microsoft.identity.common.java.crypto.ICryptoFactory;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.util.StringUtil;
 
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.context.Scope;
 import lombok.NonNull;
 
 public class CryptoFactoryTelemetryHelper {
+
+    private static final LongCounter sFailedCryptoOperationCount = OTelUtility.createLongCounter(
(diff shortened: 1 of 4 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
```diff
@@ -48,11 +51,37 @@ public static Span createSpan(@NonNull final String name) {
...
+
+        return meter
+                .counterBuilder(name)
+                .setDescription(description)
+                .setUnit("count")
+                .build();
+    }
+
+    /**
+     * Get name of the current span, if possible.
```
tanmaymanolkar1, line 69: is this supposed to mean 'for readable spans'?

tanmaymanolkar1 (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CryptoFactoryTelemetryHelper.java
```diff
@@ -22,59 +22,67 @@
...
-            span.setStatus(StatusCode.OK);
+        try {
             return cryptoOperation.perform();
         } catch (final Exception e) {
-            span.setStatus(StatusCode.ERROR);
-            span.recordException(e);
-            span.setAttribute(crypto_exception_stack_trace.name(),
-                    StringUtil.getStackTraceAsString(e));
+            final String parentSpanName = OTelUtility.getCurrentSpanName();
+            final Attributes attributes = Attributes.of(
```
tanmaymanolkar1, line 68: If there be a need to calculate the % number for threshold, we would need the total cases as well, my guess would be we are also storing that data, and not just the failed cases for cryptos?

moumighosh (approved):

PR #1913, opened 2022-12-07
Elly/edge prompthandler

ellymakuba:
Edge browser does not have resource id for email text field thus sign in with prompthandler does not work. This PR provides a custom fix for edge sign in without use of an email field resource id.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
@@ -27,16 +27,14 @@ import androidx.test.platform.app.InstrumentationRegistry;
 import androidx.test.uiautomator.UiDevice;
 import androidx.test.uiautomator.UiObject;
 import androidx.test.uiautomator.UiObjectNotFoundException;
+import androidx.test.uiautomator.UiSelector;
 
 import com.microsoft.identity.client.ui.automation.app.App;
 import com.microsoft.identity.client.ui.automation.interaction.FirstPartyAppPromptHandlerParameters;
 import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.AadPromptHandler;
 import com.microsoft.identity.client.ui.automation.logging.Logger;
+import com.microsoft.identity.client.ui.automation.utils.CommonUtils;
 import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
-
-import org.junit.Assert;
-
-import static org.junit.Assert.fail;
 
 /**
  * A model for interacting with the Microsoft Edge Browser App during UI Test.
@@ -55,17 +53,8 @@ public class BrowserEdge extends App implements IBrowser {
     public void handleFirstRun() {
         Logger.i(TAG, "Handle First Run of Browser..");
         // cancel sync in Edge
-        UiAutomatorUtils.handleButtonClick("com.microsoft.emmx:id/not_now");
+        UiAutomatorUtils.handleButtonClickForObjectWithText("Not now");
         sleep(); // need to use sleep due to Edge animations
-        // cancel sharing data
-        UiAutomatorUtils.handleButtonClick("com.microsoft.emmx:id/not_now");
-        sleep(); // need to use sleep due to Edge animations
-        // cancel personalization
-        UiAutomatorUtils.handleButtonClick("com.microsoft.emmx:id/fre_share_not_now");
-        sleep();// need to use sleep due to Edge animations
-        // avoid setting default
-        UiAutomatorUtils.handleButtonClick("com.microsoft.emmx:id/no");
-        sleep();// need to use sleep due to Edge animations
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
```diff
@@ -51,22 +53,10 @@ public BrowserEdge() {
...
-            sleep();// need to use sleep due to Edge animations
-            // avoid setting default
-            UiAutomatorUtils.handleButtonClick("com.microsoft.emmx:id/no");
-            sleep();// need to use sleep due to Edge animations
-            shouldHandleFirstRun = false;
-        }
+        Logger.i(TAG, "Handle First Run of Browser..");
+        // cancel sync in Edge
+        UiAutomatorUtils.handleButtonClickForObjectWithText("Not now");
+        sleep(); // need to use sleep due to Edge animations
```
fadidurah, line 59: Is this an accurate change? I feel like there may be some scenarios where we would get more than just 1 page with "Not now" button. Maybe just double check that other test cases that use Edge are still passing with this change, we don't want to break existing automation.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
```diff
@@ -155,17 +145,28 @@ private void signInWithWorkOrSchoolAccount(@NonNull final String username,
...
+        }catch(UiObjectNotFoundException ex){
+            throw new AssertionError(ex);
+        }
+
+        //handle password field
+        Logger.i(TAG, "Handle Aad Login Password UI..");
+        UiAutomatorUtils.handleInput("i0118", password);
+        UiAutomatorUtils.handleButtonClick("idSIButton9");
 
+        UiAutomatorUtils.handleButtonClickForObjectWithText("Confirm");
```
fadidurah, line 169: What is this confirming? Please add a comment to convey what page we are expecting here.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
```diff
@@ -155,17 +145,28 @@ private void signInWithWorkOrSchoolAccount(@NonNull final String username,
...
+            emailField.setText(username);
+            UiAutomatorUtils.handleButtonClickForObjectWithText("Next");
+        }catch(UiObjectNotFoundException ex){
+            throw new AssertionError(ex);
+        }
+
+        //handle password field
+        Logger.i(TAG, "Handle Aad Login Password UI..");
+        UiAutomatorUtils.handleInput("i0118", password);
+        UiAutomatorUtils.handleButtonClick("idSIButton9");
```
fadidurah: I believe this is a deviation from what we agreed on during our call. We should still use these lines:
        final AadPromptHandler aadPromptHandler = new AadPromptHandler(promptHandlerParameters);
        aadPromptHandler.handlePrompt(username, password);

This is so that we handle any other behavior that comes up for an account (something like "Stay Signed In" page or consent page would not be handled in this operation), all of which is contained in the AadPromptHandler call.

fadidurah: I see you've added the username as loginHint in your broker PR, so that is good (this means the AadPromptHandler will not try to handle email page, since it's already being handled earlier in this method.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
```diff
@@ -155,17 +145,28 @@ private void signInWithWorkOrSchoolAccount(@NonNull final String username,
...
 
         // click Sign In with work or school account btn
         signInWithWorkAccountBtn.click();
 
         Logger.i(TAG, "Handle Sign-In Prompt for Work or School account..");
-        // handle prompt
-        final AadPromptHandler aadPromptHandler = new AadPromptHandler(promptHandlerParameters);
-        aadPromptHandler.handlePrompt(username, password);
+        // handle email field
+        final UiObject emailField = UiAutomatorUtils.obtainUiObjectWithUiSelector(new UiSelector().className("android.widget.EditText"), CommonUtils.FIND_UI_ELEMENT_TIMEOUT);
```
fadidurah, line 156: Let's have a more descriptive comment here, explaining why we needed custom behavior for handling email field instead of just relying on AadPromptHandler (the email field in Edge UI is missing a resource id, so we find it with EditText class).

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
fadidurah: If anything, why are we removing the conditional check for shouldHandleFirstRun, and subsequent setting "Handle First Run" to false once we finish handling first run.

fadidurah: Let's make sure the existing cases using edge are still passing. If so, we can merge this change in to also allow this case to pass.

bktowett (approved):

fadidurah (approved):

iamgusain (approved):
LGTM,
If we need resource-ids, i can connect you to folks on the Edge team.

PR #1917, opened 2022-12-14
Format thread+correlationId metadata only once logging is clearly opted-in

nickbopp:
After previous performance improvements, this is now ~25-40% of cached token read time in local experiments, since this code is synchronizing and string formatting for all the Logger.verbose() calls even when verbose logging is turned off.

Synchronization requirements are also removed here, as the underlying getRequestContext() call is a ThreadLocal and there's no data race.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MINOR] Format thread+correlationId metadata only once logging is clearly opted-in (#1917)
 
 V.9.1.0
 ----------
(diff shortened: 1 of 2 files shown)


melissaahn (approved):
Looks good to me, but @iamgusain should probably take a look too.

common/src/main/java/com/microsoft/identity/common/logging/Logger.java
```diff
@@ -216,18 +216,6 @@ public void log(String tag,
...
 
-    /**
-     * Get only the required metadata from the DiagnosticContext
-     * to plug it in the log lines.
-     * Here we are considering the correlation_id and the thread_name.
-     * The need for this is because DiagnosticContext contains additional metadata which is not always required to be logged.
-     *
-     * @return String The concatenation of thread_name and correlation_id to serve as the required metadata in the log lines.
-     */
-    public static String getDiagnosticContextMetadata() {
```
iamgusain: I am not sure if any of our partner/consumer is using these or not, as these being public static. (Likely not).
To be on safer side, let's keep them and mark as deprecated.

iamgusain (approved):

PR #1920, opened 2022-12-20
Melissaahn/cert based auth manager refactor

rpdome:
Creating this to get a complete picture for reviewing PR.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
     public CertBasedAuthFactory(@NonNull final Activity activity) {
         final String methodTag = TAG + ":CertBasedAuthFactory";
         mActivity = activity;
-        mSmartcardCertBasedAuthManager = new YubiKitCertBasedAuthManager(mActivity.getApplicationContext());
-        mSmartcardCertBasedAuthManager.setDiscoveryExceptionCallback(new AbstractSmartcardCertBasedAuthManager.IDiscoveryExceptionCallback() {
+        mSmartcardUsbCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createSmartcardUsbCertBasedAuthManager(mActivity.getApplicationContext());
+        mSmartcardNfcCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createSmartcardNfcCertBasedAuthManager(mActivity.getApplicationContext());
+        mDialogHolder = new DialogHolder(mActivity);
+        if (mSmartcardUsbCertBasedAuthManager == null) {
```
rpdome: can mSmartcardNfcCertBasedAuthManager be null also?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
```diff
@@ -0,0 +1,107 @@
...
+        mSpan.recordException(exception);
+        mSpan.setStatus(StatusCode.ERROR);
+        mSpan.end();
+    }
+
+    /**
+     * Sets attribute that indicates the user's intended choice for CBA (smartcard or on-device).
+     * @param choice string indicating user's intended choice for CBA.
+     */
+    public void setUserChoice(@NonNull final String choice) {
```
rpdome: enum?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+        telemetryHelper.setUserChoice(NON_APPLICABLE);
+        telemetryHelper.setCertBasedAuthChallengeHandler(NON_APPLICABLE);
+        if (mSmartcardUsbCertBasedAuthManager == null) {
+            //Smartcard CBA is not available, so default to on-device.
+            callback.onReceived(new OnDeviceCertBasedAuthChallengeHandler(
+                    mActivity,
+                    telemetryHelper));
+            return;
+        }
+        else if (mSmartcardUsbCertBasedAuthManager.isDeviceConnected()) {
```
rpdome: nit: add extra line and remove "else"

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+            telemetryHelper.setUserChoice(SMARTCARD_CHOICE);
+            callback.onReceived(new SmartcardUsbCertBasedAuthChallengeHandler(
+                    mActivity,
+                    mSmartcardUsbCertBasedAuthManager,
+                    mDialogHolder,
+                    telemetryHelper));
+            return;
+        }
+        //Need input from user to determine which CertBasedAuthChallengeHandler to return.
```
rpdome, line 93: nit: add extra line above

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+            @Override
+            public void onClosedConnection() {
+                //ConnectionCallback will be changed before ever reaching this method.
+            }
+        });
+
+        if (mSmartcardNfcCertBasedAuthManager == null) {
+            return;
         }
+        mSmartcardNfcCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
```
rpdome: maybe we can add mSmartcardNfcCertBasedAuthManager.startDiscovery(mActivity)) here, and just use this whole method to setUpForSmartcardCertBasedAuth()?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+            @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+            @Override
+            public void onCreateConnection() {
+                mSmartcardUsbCertBasedAuthManager.setConnectionCallback(null);
```
rpdome: do we also need to invoke stopDiscovery on usb?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+        mSmartcardUsbCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createSmartcardUsbCertBasedAuthManager(mActivity.getApplicationContext());
+        mSmartcardNfcCertBasedAuthManager = SmartcardCertBasedAuthManagerFactory.createSmartcardNfcCertBasedAuthManager(mActivity.getApplicationContext());
+        mDialogHolder = new DialogHolder(mActivity);
+        if (mSmartcardUsbCertBasedAuthManager == null) {
+            //This means that a user cannot use a smartcard via USB.
+            //For consistency, we aren't going to allow the user to authenticate via smartcard CBA.
+            //In the rare case a user is able to use NFC but not USB, this will require some different design work.
+            return;
+        }
+        mSmartcardUsbCertBasedAuthManager.setDiscoveryExceptionCallback(new AbstractSmartcardCertBasedAuthManager.IDiscoveryExceptionCallback() {
```
rpdome: nit: maybe we could remove this unused method for now. (Logging should be done in the underlying layer anyway).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -139,13 +138,19 @@ public void onException(@NonNull final Exception e) {
         return null;
     }
 
+    /**
+     * Additional logic to run upon before the end of processChallenge.
+     */
+    protected abstract void onGetCertSessionExtendedLogic();
```
rpdome: nit: onGetSessionFinished();

what would happen if we don't invoke this? (as in - keep the logic similar to the usb one, we discover at the start, and stop discovering only at the very end.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -297,16 +261,24 @@ private void clearPin(@NonNull final char[] pin) {
...
     public void emitTelemetryForCertBasedAuthResults(@NonNull final RawAuthorizationResult response) {
-        if (mIsSmartcardCertBasedAuthProceeding) {
-            final CertBasedAuthResultEvent certBasedAuthResultEvent = new CertBasedAuthResultEvent(TelemetryEventStrings.Event.CERT_BASED_AUTH_RESULT_SMARTCARD_EVENT);
-            mIsSmartcardCertBasedAuthProceeding = false;
-            Telemetry.emit(certBasedAuthResultEvent.putResponseCode(response.getResultCode().toString()));
-            //If an exception was provided, emit an ErrorEvent.
+        if (!mIsSmartcardCertBasedAuthProceeding) {
+            return;
+        }
+        mIsSmartcardCertBasedAuthProceeding = false;
```
rpdome: nit: probably don't need this line.

rpdome: (This method is for emitting telemetry only - it shouldn't do any other work - including resetting an internal state.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/OnDeviceCertBasedAuthChallengeHandler.java
```diff
@@ -133,13 +145,19 @@ public void alias(String alias) {
...
     public void emitTelemetryForCertBasedAuthResults(@NonNull final RawAuthorizationResult response) {
         if (mIsOnDeviceCertBasedAuthProceeding) {
-            final CertBasedAuthResultEvent certBasedAuthResultEvent= new CertBasedAuthResultEvent(TelemetryEventStrings.Event.CERT_BASED_AUTH_RESULT_ON_DEVICE_EVENT);
-            mIsOnDeviceCertBasedAuthProceeding = false;
-            Telemetry.emit(certBasedAuthResultEvent.putResponseCode(response.getResultCode().toString()));
-            //If an exception was provided, emit an ErrorEvent.
-            final BaseException exception = response.getException();
-            if (exception != null) {
-                Telemetry.emit(new ErrorEvent().putException(exception));
+            final RawAuthorizationResult.ResultCode resultCode = response.getResultCode();
```
rpdome, line 148: nit: this whole block looks like a redundant logic - it's duplicated in every "CertBasedAuth" challenge handler. Look for a way to consolidate them (can be done separately, later. This PR is already too big.)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -233,46 +216,27 @@ public void onException(@NonNull final Exception e) {
...
-                indicateTooManyFailedAttempts(methodTag);
-                request.cancel();
-            } else {
-                //Update Dialog to indicate that an incorrect attempt was made.
-                mDialogHolder.setPinDialogErrorMode();
-            }
-
-        }
-    }
+    protected abstract void tryUsingSmartcardWithPin(@NonNull final char[] pin,
```
rpdome: This is not used in the abstract class - no need to add an abstract method.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardCertBasedAuthManagerFactory.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param context current Context.
+     * @return An AbstractSmartcardCertBasedAuthManager if NFC is supported on the device; null otherwise.
+     */
+    @Nullable
+    static AbstractSmartcardCertBasedAuthManager createSmartcardNfcCertBasedAuthManager(@NonNull final Context context) {
+        try {
+            return new YubiKitNfcCertBasedAuthManager(context);
+        } catch (final NfcNotAvailable e) {
+            //This means that the device does not have an NFC reader.
+            return null;
```
rpdome, line 68: Log?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,175 @@
...
+     * @param smartcardNfcCertBasedAuthManager SmartcardNfcCertBasedAuthManager instance.
+     * @param dialogHolder DialogHolder instance.
+     * @param telemetryHelper CertBasedAuthTelemetryHelder instance.
+     */
+    public SmartcardNfcCertBasedAuthChallengeHandler(@NonNull final Activity activity,
+                                                     @NonNull final AbstractSmartcardCertBasedAuthManager smartcardNfcCertBasedAuthManager,
+                                                     @NonNull final DialogHolder dialogHolder,
+                                                     @NonNull final CertBasedAuthTelemetryHelper telemetryHelper) {
+        super(activity, smartcardNfcCertBasedAuthManager, dialogHolder, telemetryHelper, SmartcardNfcCertBasedAuthChallengeHandler.class.getSimpleName());
+        mSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
```
rpdome: Do we need this? I would expect the callback to be set in CertBasedAuthFactory only.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardUsbCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,150 @@
...
+            @Override
+            public void onCreateConnection() {
+                //Reset DialogHolder to null if necessary.
+                //In this case, DialogHolder would be an ErrorDialog if not null.
+                mDialogHolder.dismissDialog();
+            }
+
+            @Override
+            public void onClosedConnection() {
+                if (mDialogHolder.isDialogShowing()) {
```
rpdome: do we need to show a similar dialog in the NFC case?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbCertBasedAuthManager.java
```diff
@@ -0,0 +1,198 @@
...
+
+                    mUsbDevice.setOnClosed(new Runnable() {
+                        @Override
+                        public void run() {
+                            Logger.verbose(TAG, "A YubiKey device was disconnected via usb.");
+                            synchronized (sDeviceLock) {
+                                mUsbDevice = null;
+                            }
+                            //Remove the YKPiv security provider if it was added.
+                            if (Security.getProvider(YUBIKEY_PROVIDER) != null) {
```
rpdome: Should this if-block move to AbstractYubiKitCertBasedAuthManager

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitNfcCertBasedAuthManager.java
```diff
@@ -0,0 +1,193 @@
...
+     */
+    @Override
+    void stopDiscovery(@NonNull Activity activity) {
+        synchronized (sDeviceLock) {
+            if (isDeviceConnected()) {
+                mNfcDevice.remove(new Runnable() {
+                    @Override
+                    public void run() {
+                        mNfcDevice = null;
+                        mNfcYubiKitManager.disable(activity);
```
rpdome: remove yubikit provider?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitNfcCertBasedAuthManager.java
```diff
@@ -0,0 +1,193 @@
...
+    @Override
+    boolean startDiscovery(@NonNull final Activity activity) {
+        try {
+            mNfcYubiKitManager.enable(
+                    activity,
+                    new NfcConfiguration().timeout(NFC_TIMEOUT),
+                    new Callback<NfcYubiKeyDevice>() {
+                        @Override
+                        public void invoke(@NonNull NfcYubiKeyDevice value) {
+                            mNfcDevice = value;
```
rpdome: How does this work? 

If I tap, untap and then tap again, would it get the same mNfcDevice? Is there any setOnClosed() method similar to the usb one?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardUsbCertBasedAuthChallengeHandler.java
rpdome: Also... should this reside in the CertBasedAuth factory?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,175 @@
...
+        mSmartcardCertBasedAuthManager.stopDiscovery(mActivity);
+        // If the number of attempts is 0, no more attempts will be allowed.
+        if (attemptsRemaining == 0) {
+            //We must display a dialog informing the user that they have made too many incorrect attempts,
+            // and the user will need to figure out a way to reset their key outside of our library.
+            indicateTooManyFailedAttempts(methodTag);
+            request.cancel();
+            return;
+        }
+        mDialogHolder.showPinDialog(
```
rpdome: do we need this? (since setPinDialogErrorMode will be invoked right below)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                                            @NonNull final ISmartcardSession session)
+            throws Exception {
+        final String methodTag = TAG + ":tryUsingSmartcardWithPin";
+        if (session.verifyPin(pin)) {
+            //If pin is successfully verified, we will use the certificate to perform the rest of the logic for authentication.
+            useSmartcardCertForAuth(certDetails, pin, session, request);
+            return;
+        }
+        final int attemptsRemaining = session.getPinAttemptsRemaining();
+        mSmartcardCertBasedAuthManager.stopDiscovery(mActivity);
```
rpdome: move the implementation of this method to the abstract class, and invoke onGetCertSessionExtendedLogic in the  attemptsRemaining==0 block instead?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -125,9 +121,12 @@ public void onGetSession(@NonNull final ISmartcardSession session) throws Except
                             @Override
                             public void onCancel() {
                                 mDialogHolder.dismissDialog();
+                                mTelemetryHelper.setResultFailure(USER_CANCEL_MESSAGE);
                                 request.cancel();
                             }
                         });
+
+                onGetCertSessionExtendedLogic();
```
rpdome: Do we need to invoke this here?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
rpdome: what would happen if stopDiscovery() is invoked in the same pattern as the usb one? (That way we don't need to add the logic, and the code would be more sharable between nfc/usb.

rpdome (commented):
Left some comments.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
```diff
@@ -0,0 +1,175 @@
...
+                    }
+                });
+                mSmartcardCertBasedAuthManager.setConnectionCallback(new AbstractSmartcardCertBasedAuthManager.IConnectionCallback() {
+                    @Override
+                    public void onCreateConnection() {
+                        mDialogHolder.showDialog(new SmartcardNfcLoadingDialog(mActivity));
+                        mSmartcardCertBasedAuthManager.requestDeviceSession(new AbstractSmartcardCertBasedAuthManager.ISessionCallback() {
+                            @Override
+                            public void onGetSession(@NonNull final ISmartcardSession session) throws Exception {
+                                tryUsingSmartcardWithPin(pin, certDetails, request, session);
```
rpdome: what would happen if I tap with a different yubikey?

rpdome: I'm wondering if it's possible to front-load pin prompt (even before entering NFC/USB flow)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
+                    telemetryHelper));
+            return;
+        }
+        //Need input from user to determine which CertBasedAuthChallengeHandler to return.
+        mDialogHolder.showUserChoiceDialog(new UserChoiceDialog.PositiveButtonListener() {
+            @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+            @Override
+            public void onClick(final int checkedPosition) {
+                //Position 0 -> On-device
+                //Position 1 -> Smartcard
```
rpdome, line 99: Can we explicitly ask if they are going to use NFC or usb?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
```diff
@@ -45,37 +54,209 @@ public class CertBasedAuthFactory {
...
      */
-    @NonNull
-    public ICertBasedAuthChallengeHandler createCertBasedAuthChallengeHandler() {
-        if (mSmartcardCertBasedAuthManager.isDeviceConnected()) {
-            return new SmartcardCertBasedAuthChallengeHandler(mSmartcardCertBasedAuthManager, new DialogHolder(mActivity));
-        } else {
-            return new OnDeviceCertBasedAuthChallengeHandler(mActivity);
+    private void showSmartcardPromptDialogAndSetConnectionCallback(@NonNull final CertBasedAuthChallengeHandlerCallback challengeHandlerCallback,
+                                                                   @NonNull final CertBasedAuthTelemetryHelper telemetryHelper) {
+        mDialogHolder.showSmartcardPromptDialog(new SmartcardPromptDialog.CancelCbaCallback() {
```
rpdome, line 181: Can we invoke showSmartcardPromptDialogAndSetConnectionCallback() right at the beginning of createCertBasedAuthChallengeHandler(), and proceed with corresponding flow if USB or NFC device is detected (via the callback).

Otherwise, display **_a_** dialog asking if they want to proceed with on-device cba, or plug/tap the key if they want to proceed with smartcard cba. (I don't think we need 2-3 separate dialogs here)

I'm imagining something like this
https://blogger.googleusercontent.com/img/b/R29vZ2xl/AVvXsEiwAdVPzjoZ5m_Ob3DRmOXgLlUBKWxYorXerbVyxQB2DZOrPZU-f43aEJx5iUCcBVFarZvo8WTNuSxcjeVaTPfF46o1hztbbAlHrKriJ8ESEnUPV9Vt3coCKPNqcp0Vkv6w6HK0xNIh95jRh3rYklORnZ8UNgfw3v7KOVnAY_R1py_U521JHJqVGuhk/s1600/image3.png

(change the fingerprint into some sort of image that guides user to tap/plug yubikey/smartcard, and "use pin" can be changed to "use On device cert". Sizing of each element could be changed.)

rpdome: Also, query https://stackoverflow.com/questions/7618657/how-to-check-whether-nfc-is-enabled-or-not-in-android and display some sort of "hint" to show if NFC is enabled? (refresh upon UI relaunch - e.g. if the customer minimize the dialog and come back, we should reload this)

rpdome: These are all UX (designer) pending... i guess

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
melissaahn: yes, at least, in the current implementation, where it only keeps NFC discovery on when necessary, Though I think this will probably be discussed more in the comments under this one.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/AbstractSmartcardCertBasedAuthChallengeHandler.java
melissaahn: Renamed.

So this is feasible, but the connection callback then needs to change to null to avoid going back in the dialog flow, or something unexpected happening. For example, a user is on the pin dialog page, but they accidently tap their smartcard, which will bring them back to the initial cert picker dialog. I think something does need to be called at the end of a session in the case of NFC, whether that's stopping discovery or setting the callback to null.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: > Also, query https://stackoverflow.com/questions/7618657/how-to-check-whether-nfc-is-enabled-or-not-in-android and display some sort of "hint" to show if NFC is enabled? (refresh upon UI relaunch - e.g. if the customer minimize the dialog and come back, we should reload this)

If NFC is not enabled, then a dialog is shown that tells the user they need to turn on NFC in settings to use NFC. This doesn't cause the CBA flow to fail though, as clicking "got it" on the ack dialog just takes them to the smartcard prompt dialog.
I didn't write logic to refresh upon UI relaunch, so that will most likely be a separate PR.

melissaahn: We talked about the UI things offline.

> Can we invoke showSmartcardPromptDialogAndSetConnectionCallback() right at the beginning of createCertBasedAuthChallengeHandler(), and proceed with corresponding flow if USB or NFC device is detected (via the callback).

If I understand the point you're making here, I did include a check for if a usb smartcard is already connected at the beginning of createCertBasedAuthChallengeHandler. I'm not sure if showSmartcardPromptDialogAndSetConnectionCallback() is the right method to call here since no smartcard prompt is needed (can go directly to cert picker). I can't really think of a scenario where an NFC connection is already made and active at the start of createCertBasedAuthChallengeHandler(), since an NFC connection is not constantly active like a usb one would be.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: Yes, and this means that the device doesn't have an NFC reader. I didn't think there was a need to check that in the constructor since NFC discovery isn't turned on yet. The proper null checks are made whenever this object is referenced though.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: Is your suggestion to have three radio buttons?
- On-device
- USB
- NFC

Since this would be a UI change, I would need to bring this up with the designer.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: I extracted some lines into showSmartcardPromptDialogAndSetConnectioNCallback() because of the extra dialog that appears when a device doesn't have NFC capabilities, and I was trying to avoid some repeated logic, so I'm not sure if the suggestion would be a full replacement for setUpForSmartcardCertBasedAuth()? I'm wondering if I'm just not understanding your comment correctly though.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CertBasedAuthFactory.java
melissaahn: If stopping and starting Usb discovery was more consistent, I would probably go ahead and stop it here. However, I've noticed some irregularities in the past with calling startDiscovery/stopDiscovery for USB only when necessary, mostly not detecting plugged in smartcards consistently. I also think that while the stopping/starting is not as consistent across USB and NFC, the use cases are kinda different.
I don't think there's as much of a benefit from the user's perspective for having NFC discovery on all the time (if the user taps before they get to the custom dialogs, nothing can really happen, since the NFC connection is designed to survive for a finite amount aof time, and I'm not sure if there are many use cases when a user is constantly holding their smartcard to their device instead of tapping when prompted?). But I think there is for USB, where a user can just plug in their smartcard whenever, and it'll just stay there, stuck in the phone, ready to make a connection when needed.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/OnDeviceCertBasedAuthChallengeHandler.java
melissaahn: Will do in a separate, smaller PR.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
melissaahn: Yeah, I think this call is redundant, so I'll delete it. 
But setting the callback is also done below in getSmartcardPinDialogPositiveButtonListener, and that one is important because different logic needs to occur upon a tap (ex. reading certs off smartcard -> confirming PIN). So I wouldn't necessarily say that all calls to set the callback can be confined to CertBasedAuthFactory.createCertBasedAuthChallengeHandler.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardUsbCertBasedAuthChallengeHandler.java
melissaahn: > do we need to show a similar dialog in the NFC case?

I was thinking about including a dialog that specifically tells the user that they didn't hold the NFC card long enough, but the exceptions returned in this case were inconsistent (and not exclusive to NFC issues), so I need to do some more testing to see if I can narrow down the types of exceptions returned. Right now, there is a generic error dialog being displayed and the request is handled accordingly (cancelled), and there is a dialog while an NFC connection is active that instructs the user to keep holding the tap. 
So a PR with a custom NFC error message will most likely come later with other UI changes.


> Also... should this reside in the CertBasedAuth factory?

There wouldn't be any issues with moving it there, but in a previous comment I address that the connection callback does need to be altered within the nfc challenge handler, so there would be some issues with making setting the callback exclusive in CertBasedAuthFactory.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
melissaahn: Yes, since unlike the usb implementation, the nfc flow includes an extra two dialogs after the pin dialog (one for prompting user to tap and another telling them to hold their smartcard) , so the displayed dialog is no longer the pin dialog. The pin dialog needs to be displayed again to the user, and then setPinDialogErrorMode can be called (note that setPinDialogErrorMode does not create and display the pin dialog, it just sets some ui properties).

melissaahn (commented):
Addressing around half of comments. Working on getting to the rest of them.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
melissaahn: enum CertBasedAuthChoice has been added.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitNfcCertBasedAuthManager.java
melissaahn: Added in the destroy method of YubiKitNfcSmartcardCertBasedAuthManager

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitUsbCertBasedAuthManager.java
melissaahn: Ended up adding a removeProvider static method to YubiKeyPivProviderManager class and call that method here and also within the corresponding nfc manager.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/YubiKitNfcCertBasedAuthManager.java
melissaahn: It doesn't seem to get the same mNfcDevice, but it does get the same nfc tag object, so I use that to make a comparison instead.
I can't seem to find a setOnClosed() async method, but the alternative is a remove() method, which has to be called manually.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SmartcardNfcCertBasedAuthChallengeHandler.java
melissaahn: :^) good point. added a method isCurrDiffFromPrevConnectedDevice() to abstract class AbstractNfcSmartcardCertBasedAuthManager, and this method is called before  tryUsingSmartcardWithPin() is called.
If it returns true, then a generic error dialog will show and the flow will cancel. Showing a generic error dialog for now since creating a more specific one requires the designer and localization, so that will be replaced in a later PR.

PR #1921, opened 2022-12-21
Add support for new test case, fetch lab account with upn, annotate brokerhost tests, support edge local installation

fadidurah:
PR to support Decemeber ui automation maintenance PRs:
(MSAL) https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1741
(Broker) https://github.com/AzureAD/ad-accounts-for-android/pull/2092

- Add support for new test case (device registration through authenticator settings page)
- add support to fetch lab accounts by passing in the account's upn
- add annotation for tests that should be ran with LocalBrokerHost build variant for msaltestapp
- Support local installation for Edge app


melissaahn (approved):

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -77,6 +77,13 @@ public ILabAccount getLabAccount(@NonNull final LabQuery labQuery) throws LabApi
         return getLabAccountObject(configInfo);
     }
 
+    public ILabAccount getLabAccount(@NonNull final String upn) throws LabApiException {
+        final List<ConfigInfo> configInfos = fetchConfigsFromLab(upn);
+        // We still get a list of configs despite passing in a single upn, so we select the account from the list.
+        final ConfigInfo configInfo = configInfos.get(0);
+        return getLabAccountObject(configInfo);
```
somalaya, line 85: can configInfo be null? seeing if we have to add a null check here

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
```diff
@@ -45,15 +45,15 @@
 public class BrokerAuthenticatorUpdatedVersionImpl extends BrokerMicrosoftAuthenticator {
 
     private static final String TAG = BrokerAuthenticatorUpdatedVersionImpl.class.getSimpleName();
-    private static final boolean isExpectingMFA = true;
+    public static boolean shouldUseDeviceSettingsPage = true;
```
somalaya, line 48: isExpectingMFA flag was initially added because there was a bug in Authenticator app during device registration. It was asking for MFA. If it is resolved now, we can simply remove the flag right?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
```diff
@@ -342,6 +342,11 @@ protected void performDeviceRegistrationHelper(@NonNull final String username,
         }
     }
 
+    public void shouldUseDeviceSettingsPage(final boolean shouldUseDeviceSettingsPage) {
```
somalaya: Are we running our tests on previous versions of authenticator today? The reason I added the prev/updated variants for Authenticator impl is because the UI changed for authenticator and we wanted the test cases to succeed on both previous and updated versions of Auth app. Now that this bug has been fixed for a long time.. I think we should just let it fail on previous versions instead of adding these checks again. LMK if we still want the test cases to run on prev versions.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
```diff
@@ -321,6 +321,22 @@ public static void handleButtonClickForObjectWithText(@NonNull final String text
...
+     *
+     * @param text the text on the button to click
+     */
+    public static void handleButtonClickForObjectWithTextSafely(@NonNull final String text) {
+        final UiObject button = obtainUiObjectWithText(text);
+
+        try {
+            button.click();
+        } catch (final UiObjectNotFoundException e) {
+            Logger.i(TAG, "Button with text \"" + text + "\" was not found: " + e.getMessage());
```
somalaya: Why do we need this safe functionality? I see you are using this during launching of edge app. Should we not simply fail if the "No, Thanks" button is not found?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
fadidurah: We're still using older versions of authenticator when running the update scenario tests

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
fadidurah: From my perspective, I don't think so. I say that because our test is not concerned with these popups that edge shows the user, we should not assert them as they do not interact with broker and they change depending on updates from Edge (they can't truely be predicted from our end, we basically change the check depending on whether or not they appear). I think it's fine to skip asserting some ui aspects of these tests if they are not tied to anything in our library or the overall token flows.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
fadidurah: We could, I figured we could keep it for now so as not to mess with any of the other test cases. I don't think the MFA prompt is an issue anymore. I'm hesitant to do that as part of this Pr as many changes are already coming in. I will probably investigate in january maintenance effort

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -77,6 +77,13 @@ public ILabAccount getLabAccount(@NonNull final LabQuery labQuery) throws LabApi
         return getLabAccountObject(configInfo);
     }
 
+    public ILabAccount getLabAccount(@NonNull final String upn) throws LabApiException {
```
shahzaibj, line 81: I suggest we also add the method to `ILabClient` and then add the `overrides` annotation here

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
```diff
@@ -85,15 +85,10 @@ public void performDeviceRegistration(@NonNull final String username,
...
-                Assert.assertTrue(
-                        "Microsoft Authenticator - Unregister Button is clickable.",
-                        unRegisterBtn.isClickable()
-                );
-
                 // after device registration, make sure that the current registration upn matches with
                 // with what was passed in
                 final UiObject currentRegistration = UiAutomatorUtils.obtainUiObjectWithResourceId(
-                        "com.azure.authenticator:id/current_registered_email"
+                        "workPlaceRegisteredAccountEmailText"
```
shahzaibj, line 91: So this resource id got changed?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
shahzaibj: The name of this method sounds like it is supposed to return a boolean, however, I see that the method is `void`

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
```diff
@@ -45,12 +46,30 @@ public class BrowserEdge extends App implements IBrowser {
     private final static String TAG = BrowserEdge.class.getSimpleName();
     private static final String EDGE_PACKAGE_NAME = "com.microsoft.emmx";
     private static final String EDGE_APP_NAME = "Microsoft Edge";
-    private boolean shouldHandleFirstRun = true;
+    private static final String EDGE_APK = "Edge.apk";
+    private boolean shouldHandleAutoFill = true;
```
shahzaibj, line 50: What's the difference between auto fill and first run?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
shahzaibj: the log level here should probably be at least `warn`

shahzaibj (commented):
Left some comments

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
shahzaibj: Okay, it seems this method is actually a setter. Should the method name be `setShouldUseDeviceSettingsPage` ?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/utils/UiAutomatorUtils.java
fadidurah: Sure, let's do that.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
fadidurah: From what I saw, yes. This code is only being ran in one test case though.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerMicrosoftAuthenticator.java
fadidurah: Ahh yes, it's not properly named. I'll fix that. Although I think I'll probably be getting rid of it fairly soon.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/browser/BrowserEdge.java
fadidurah: Some slight differences in UI elements that edge has based on a "first run" and the specific autofill pop-up that appears. I need to revisit this though, it seems it still not solving the issue.

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
fadidurah: I added an assertionError throw in getLabAccountObject. If no lab account is found, we still get a ConfigInfo object, but it has a null userinfo, so this is what I based the assertion on.

somalaya (approved):

shahzaibj (approved):

PR #1923, opened 2022-12-21
Only reset Mac after the operation is done

rpdome:
WolfCrypt's HMac Implementation will throw an exception if reset() is invoked.


mohitc1 (approved):

somalaya (approved):

titorodi:
👏🏻

PR #1926, opened 2023-01-11
Propagate span context from MSAL to Broker

shahzaibj:
This creates a Serializable variant of Span Context that is added to BrokerRequest.

This is done per the [spec for Span Context Propagation](https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?version=GBshahzaibj/context-propagation&path=/%5BAndroid%5D%20Telemetry/span_context_propagation.md&_a=preview) which is being done to allow our partners to onboard to Open Telemetry and allows us to start traces directly in client applications and allows broker to pick up that trace as part of the flow.

Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2101

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
@@ -27,6 +27,7 @@ import androidx.annotation.Nullable;
 
 import com.google.gson.annotations.SerializedName;
 import com.microsoft.identity.common.java.authscheme.AbstractAuthenticationScheme;
+import com.microsoft.identity.common.java.opentelemetry.SerializableSpanContext;
 import com.microsoft.identity.common.java.providers.oauth2.OpenIdConnectPromptParameter;
 import com.microsoft.identity.common.java.request.SdkType;
 
@@ -69,6 +70,7 @@ public class BrokerRequest implements Serializable {
         final static String AUTHORIZATION_AGENT = "authorization_agent";
         final static String AUTHENTICATION_SCHEME = "authentication_scheme";
         final static String POWER_OPT_CHECK_ENABLED = "power_opt_check_enabled";
+        final static String SPAN_CONTEXT = "span_context";
     }
 
     /**
@@ -218,5 +220,9 @@ public class BrokerRequest implements Serializable {
     @Nullable
     @SerializedName(SerializedNames.POWER_OPT_CHECK_ENABLED)
     private boolean mPowerOptCheckEnabled;
+
+    @Nullable
+    @SerializedName(SerializedNames.SPAN_CONTEXT)
+    private SerializableSpanContext mSpanContext;
 
 }
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
@@ -25,18 +25,20 @@ package com.microsoft.identity.common.internal.controllers;
 
 import android.os.Bundle;
 
-import com.microsoft.identity.common.java.marker.CodeMarkerManager;
-import com.microsoft.identity.common.java.marker.PerfConstants;
 import com.microsoft.identity.common.exception.BrokerCommunicationException;
 import com.microsoft.identity.common.internal.broker.ipc.BrokerOperationBundle;
 import com.microsoft.identity.common.internal.broker.ipc.IIpcStrategy;
-import com.microsoft.identity.common.java.commands.parameters.CommandParameters;
 import com.microsoft.identity.common.internal.telemetry.Telemetry;
 import com.microsoft.identity.common.internal.telemetry.events.ApiEndEvent;
 import com.microsoft.identity.common.internal.telemetry.events.ApiStartEvent;
+import com.microsoft.identity.common.java.commands.parameters.CommandParameters;
 import com.microsoft.identity.common.java.exception.BaseException;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.exception.ErrorStrings;
+import com.microsoft.identity.common.java.marker.CodeMarkerManager;
+import com.microsoft.identity.common.java.marker.PerfConstants;
+import com.microsoft.identity.common.java.opentelemetry.OTelUtility;
+import com.microsoft.identity.common.java.opentelemetry.SpanName;
 import com.microsoft.identity.common.java.util.StringUtil;
 import com.microsoft.identity.common.logging.Logger;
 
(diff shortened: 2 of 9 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -35,5 +35,8 @@ public enum SpanName {
     WorkplaceJoin,
     DoDiscovery,
     WorkplaceLeave,
-    DeviceState
+    DeviceState,
+    MSAL_AcquireTokenInteractive,
+    MSAL_AcquireTokenSilent,
```
shahzaibj: remove these

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -219,4 +221,8 @@ private static final class SerializedNames {
     @SerializedName(SerializedNames.POWER_OPT_CHECK_ENABLED)
     private boolean mPowerOptCheckEnabled;
 
+    @Nullable
+    @SerializedName(SerializedNames.SPAN_CONTEXT)
+    private SerializableSpanContext mSpanContext;
```
fadidurah, line 226: More so a question for my own knowledge: I understand the use case for serialization in something like PublicClientApplicationConfiguration in Msal, as it allows us to read a configuration json to create a PublicClientApplicationConfiguration object. What is the use case as far as the open telemetry/span implementation? Is serialization in this case used to write to a json to capture the span's telemetry?

fadidurah (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -214,11 +219,22 @@ private <T> T performStrategy(@NonNull final IIpcStrategy strategy,
...
-        operation.performPrerequisites(strategy);
-        final BrokerOperationBundle brokerOperationBundle = operation.getBundle();
-        final Bundle resultBundle = strategy.communicateToBroker(brokerOperationBundle);
-        return operation.extractResultBundle(resultBundle);
-
-        // TODO: Emit success rate and performance of each strategy to eSTS in a finally block.
+        final Span span = OTelUtility.createSpan(SpanName.MSAL_PerformIpcStrategy.name());
+
+        try (final Scope scope = span.makeCurrent()) {
+            span.setAttribute("strategy", strategy.getType().name());
```
somalaya: nit : can we move "strategy" to AttributeNames?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -113,6 +118,13 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Inte
                                 AuthorizationAgent.WEBVIEW.name()
                 ).authenticationScheme(parameters.getAuthenticationScheme())
                 .powerOptCheckEnabled(parameters.isPowerOptCheckEnabled())
+                .spanContext(SerializableSpanContext.builder()
+                        .traceId(SpanExtension.current().getSpanContext().getTraceId())
+                        .spanId(SpanExtension.current().getSpanContext().getSpanId())
+                        .traceFlags(SpanExtension.current().getSpanContext().getTraceFlags().asByte())
+                        .parentSpanName(OTelUtility.getCurrentSpanName())
+                        .build()
+                )
```
somalaya, line 127: so shahzaib, do we have to add this span context change to every type of request made to broker? Asking because Pallav is adding a new method to create a broker request for DCF.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
```diff
@@ -51,6 +55,36 @@ public static Span createSpan(@NonNull final String name) {
...
+            Logger.warn(methodTag, "parentSpanContext is INVALID. Creating span without parent.");
+            return createSpan(name);
+        }
+
+        final Tracer tracer = GlobalOpenTelemetry.getTracer(TAG);
+        final Span span = tracer.spanBuilder(name)
+                .setParent(Context.current().with(Span.wrap(parentSpanContext)))
+                .startSpan();
+
+        if (parentSpanContext instanceof SerializableSpanContext) {
```
somalaya, line 81: else, is it an unexpected scenario? Do we want to add a log for such a scenario?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
```diff
@@ -80,13 +114,13 @@ public static Attributes getCurrentSpanAttributes() {
...
      **/
-    @Nullable
+    @NonNull
     public static String getCurrentSpanName() {
         final Span span = SpanExtension.current();
         if (span instanceof ReadableSpan) {
             return ((ReadableSpan) span).getName();
         }
-        return null;
+        return "null";
```
somalaya: can we just make this an empty string instead of "null"? Was thinking if it may reduce size of a row if we leave it empty.  Although I believe span name is mostly set when we create a span and this is just an edge case. Correct me if I am wrong.

shahzaibj: Great point. I'll change it to empty string

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
shahzaibj: The use case here is about sending data over IPC from one app/process to another app/process. A complex object needs to be represented in a type such as String or byte to be able to be sent across processes. This is achieved through serialization and we are using that here to convert it to a String so we can send it from MSAL to Broker over IPC.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
shahzaibj: @somalaya span context is an optional field so it is not required to be present. If we want to capture telemetry for DCF then Pallav should add it in that method.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1926/commits/90b4d126ee672aba3e398e7efa58c755ef490d24

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1926/commits/b2095cb6271fdac7c4a150253e2ea36bdc661390

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1926/commits/9d5bb835341070a37807e077cbd8d4d631582efa

somalaya (approved):

PR #1927, opened 2023-01-12
Write back read successes to cache and stop extra lookups in getAll

nickbopp:
Nearly every storage operation involves `getAll` on all accounts or all credentials.  When an app restarts, the shared preferences storage has assets already sourced from disk.  Right now successful reads (and crucially, decrypts) do not store their value in `fileCache`, so they are decrypted repeatedly.  Only values actively written to storage by the application in any given boot were being placed into `fileCache`.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -154,26 +154,54 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
         }
     }
 
-    @Override
-    @Nullable
-    public final String getString(final String key) {
+    private String getStringInternal(final String key, String valueFromSharedPrefs)
+    {
+        // Check the cache for the (potentially decrypted) value from the cache, and return
         synchronized (cacheLock) {
             String memCache = fileCache.get(key);
             if (memCache != null) {
                 return memCache;
             }
-            String restoredValue = mSharedPreferences.getString(key, null);
-
-            if (null != mEncryptionManager && !StringUtil.isNullOrEmpty(restoredValue)) {
-                restoredValue = decrypt(restoredValue);
-
-                if (StringUtil.isNullOrEmpty(restoredValue)) {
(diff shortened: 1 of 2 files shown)


iamgusain (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -216,7 +244,7 @@ public final Map<String, String> getAll() {
                 final Map.Entry<String, String> entry = iterator.next();
                 //This is slightly wasteful, but we have no better key iterator and decryption
                 //is probably more painful than the additional file read when we miss in the cache.
-                String decryptedValue = getString(entry.getKey());
+                String decryptedValue = getStringInternal(entry.getKey(), entry.getValue());
```
rpdome, line 247: nit: im wondering if it would be cleaner to read if 

1. We created a "getValueFromSharedPrefEntry()" method, which performs both decryption and writeback.
2. Invoke the method here and in getString().

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
nickbopp: I tried to figure out a way to make this work, but I don't think I can structure it this way without adding the overall flow making extraneous calls to SharedPreferences.getString(), which has a larger perf cost that just using the values we already got from SharedPreferences.getAll().

We always want to prefer existing decrypted-cache values over either using the getAll() value or a new SharedPreferences.getString() and decrypting + write-backing it, and to avoid getString() calls if that value exists.

nickbopp: I could make the second parameter to getStringInternal a Callable<String> instead of a nullable String and pass `() -> mSharedPreferences.getString(key, null)` for the getString() implementation and `() -> entry.getValue()` for the getAll() implementation.  Do you think that would be an improvement?

PR #1929, opened 2023-01-19
Keep in-memory cache of all credentials and accountrecords

nickbopp:
High level token operations start with SharedPreferencesFileManager.getCredentails()/getAccountRecords().  Right now, this requires hitting the SharedPreferencesFileManager and at the very least deserializing all credential/account types every time.

This change maintains a layer of caching at the List<AccountRecord> and List<Credential> level. that will prevent any decryption/deserialization from taking place in the 'get' path.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.common.java.cache;
 
 import com.microsoft.identity.common.java.dto.AccessTokenRecord;
+import com.microsoft.identity.common.java.dto.AccountCredentialBase;
 import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.common.java.dto.Credential;
 import com.microsoft.identity.common.java.dto.CredentialType;
@@ -88,6 +89,12 @@ public class SharedPreferencesAccountCredentialCache extends AbstractAccountCred
 
     private final ICacheKeyValueDelegate mCacheValueDelegate;
 
+    private final Object mCacheLock = new Object();
+    private boolean mLoaded = false;
+
+    private Map<String, AccountRecord> mCachedAccountRecordsWithKeys = new HashMap<>();
+    private Map<String, Credential> mCachedCredentialsWithKeys = new HashMap<>();
+
     /**
      * Constructor of SharedPreferencesAccountCredentialCache.
      *
@@ -100,113 +107,153 @@ public class SharedPreferencesAccountCredentialCache extends AbstractAccountCred
         Logger.verbose(TAG, "Init: " + TAG);
         mSharedPreferencesFileManager = sharedPreferencesFileManager;
         mCacheValueDelegate = accountCacheValueDelegate;
+        new Thread(() -> load()).start();
+    }
+
+    private void load() {
+        final String methodTag = TAG + ":load";
+
+        synchronized (mCacheLock) {
+            try {
+                mCachedAccountRecordsWithKeys = loadAccountsWithKeys();
+                Logger.info(methodTag, "Loaded " + mCachedAccountRecordsWithKeys.size() + " AccountRecords");
+                mCachedCredentialsWithKeys = loadCredentialsWithKeys();
+                Logger.info(methodTag, "Loaded " + mCachedCredentialsWithKeys.size() + " Credentials");
+            } catch (final Throwable t) {
+                Logger.error(methodTag, "Failed to load initial accounts or credentials from SharedPreferences", t);
+            } finally {
+                mLoaded = true;
+                mCacheLock.notifyAll();
+            }
+        }
+    }
+
+    private void waitForInitialLoad() {
+        final String methodTag = TAG + ":waitForInitialLoad";
+
+        while (!mLoaded) {
+            try {
+                mCacheLock.wait();
+            } catch (final InterruptedException e) {
+                Logger.error(methodTag, "Caught InterruptedException while waiting", e);
+            }
+        }
     }
 
     @Override
-    public synchronized void saveAccount(@NonNull final AccountRecord accountToSave) {
-        Logger.verbose(TAG, "Saving Account...");
-        Logger.verbose(TAG, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
+    public void saveAccount(@NonNull final AccountRecord accountInput) {
+        final String methodTag = TAG + ":saveAccount";
+
+        AccountRecord accountToSave = null;
+        try {
+            accountToSave = (AccountRecord) accountInput.clone();
+        } catch (final CloneNotSupportedException e) {
+            Logger.error(methodTag, "Failed to clone AccountRecord", e);
+            return;
+        }
+
+        Logger.verbose(methodTag, "Saving Account...");
+        Logger.verbose(methodTag, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
         final String cacheKey = mCacheValueDelegate.generateCacheKey(accountToSave);
-        Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
+        Logger.verbosePII(methodTag, "Generated cache key: [" + cacheKey + "]");
 
-        // Perform any necessary field merging on the Account to save...
-        final AccountRecord existingAccount = getAccount(cacheKey);
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
 
-        if (null != existingAccount) {
-            accountToSave.mergeAdditionalFields(existingAccount);
+            // Perform any necessary field merging on the Account to save...
+            final AccountRecord existingAccount = getAccount(cacheKey);
+
+            if (null != existingAccount) {
+                accountToSave.mergeAdditionalFields(existingAccount);
+            }
+
+            final String cacheValue = mCacheValueDelegate.generateCacheValue(accountToSave);
+            mSharedPreferencesFileManager.put(cacheKey, cacheValue);
+            mCachedAccountRecordsWithKeys.put(cacheKey, accountToSave);
         }
-
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(accountToSave);
-        mSharedPreferencesFileManager.put(cacheKey, cacheValue);
     }
 
     @Override
-    public synchronized void saveCredential(@NonNull Credential credentialToSave) {
-        Logger.verbose(TAG, "Saving credential...");
-        final String cacheKey = mCacheValueDelegate.generateCacheKey(credentialToSave);
-        Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
+    public void saveCredential(@NonNull Credential credentialInput) {
+        final String methodTag = TAG + ":saveCredential";
 
-        // Perform any necessary field merging on the Credential to save...
-        final Credential existingCredential = getCredential(cacheKey);
-
-        if (null != existingCredential) {
-            credentialToSave.mergeAdditionalFields(existingCredential);
+        Credential credentialToSave = null;
+        try {
+            credentialToSave = (Credential) credentialInput.clone();
+        } catch (final CloneNotSupportedException e) {
+            Logger.error(methodTag, "Failed to clone Credential", e);
+            return;
         }
 
-        final String cacheValue = mCacheValueDelegate.generateCacheValue(credentialToSave);
-        mSharedPreferencesFileManager.put(cacheKey, cacheValue);
+        Logger.verbose(methodTag, "Saving credential...");
+        final String cacheKey = mCacheValueDelegate.generateCacheKey(credentialToSave);
+        Logger.verbosePII(methodTag, "Generated cache key: [" + cacheKey + "]");
+
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
+
+            // Perform any necessary field merging on the Credential to save...
+            final Credential existingCredential = getCredential(cacheKey);
+
+            if (null != existingCredential) {
+                credentialToSave.mergeAdditionalFields(existingCredential);
+            }
+
+            final String cacheValue = mCacheValueDelegate.generateCacheValue(credentialToSave);
+            mSharedPreferencesFileManager.put(cacheKey, cacheValue);
+            mCachedCredentialsWithKeys.put(cacheKey, credentialToSave);
+        }
     }
 
     @Override
     public AccountRecord getAccount(@NonNull final String cacheKey) {
-        Logger.verbose(TAG, "Loading Account by key...");
-        AccountRecord account = mCacheValueDelegate.fromCacheValue(
-                mSharedPreferencesFileManager.get(cacheKey),
-                AccountRecord.class
-        );
+        final String methodTag = TAG + ":getAccount";
 
-        if (null == account) {
-            // We could not deserialize the target AccountRecord...
-            // Maybe it was encrypted for another application?
-            Logger.warn(
-                    TAG,
-                    ACCOUNT_RECORD_DESERIALIZATION_FAILED
-            );
-        } else if (EMPTY_ACCOUNT.equals(account)) {
-            Logger.warn(TAG, "The returned Account was uninitialized. Removing...");
-            mSharedPreferencesFileManager.remove(cacheKey);
-            account = null;
+        AccountRecord foundValue;
+
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
+            foundValue = mCachedAccountRecordsWithKeys.get(cacheKey);
         }
 
-        return account;
+        try {
+            if (foundValue != null) {
+                foundValue = (AccountRecord) foundValue.clone();
+            }
+        } catch (final CloneNotSupportedException e) {
+            Logger.error(methodTag, "Failed to clone AccountRecord", e);
+        }
+        return foundValue;
     }
 
     @Override
     @Nullable
     public Credential getCredential(@NonNull final String cacheKey) {
-        // TODO add support for more Credential types...
-        Logger.verbose(TAG, "getCredential()");
-        Logger.verbosePII(TAG, "Using cache key: [" + cacheKey + "]");
+        final String methodTag = TAG + ":getCredential";
 
-        final CredentialType type = getCredentialTypeForCredentialCacheKey(cacheKey);
-        Class<? extends Credential> clazz = null;
+        Credential foundValue;
 
-        if (null != type) {
-            clazz = getTargetClassForCredentialType(cacheKey, type);
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
+            foundValue = mCachedCredentialsWithKeys.get(cacheKey);
         }
 
-        Credential credential = null;
-
-        if (null != clazz) {
-            credential = mCacheValueDelegate.fromCacheValue(
-                    mSharedPreferencesFileManager.get(cacheKey),
-                    clazz
-            );
+        try {
+            if (foundValue != null) {
+                foundValue = (Credential) foundValue.clone();
+            }
+        } catch (final CloneNotSupportedException e) {
+            Logger.error(methodTag, "Failed to clone Credential", e);
         }
 
-        if (null == credential) {
-            // We could not deserialize the target Credential...
-            // Maybe it was encrypted for another application?
-            Logger.warn(
-                    TAG,
-                    CREDENTIAL_DESERIALIZATION_FAILED
-            );
-        } else if ((AccessTokenRecord.class == clazz && EMPTY_AT.equals(credential))
-                || (RefreshTokenRecord.class == clazz && EMPTY_RT.equals(credential))
-                || (IdTokenRecord.class == clazz) && EMPTY_ID.equals(credential)) {
-            // The returned credential came back uninitialized...
-            // Remove the entry and return null...
-            Logger.warn(TAG, "The returned Credential was uninitialized. Removing...");
-            mSharedPreferencesFileManager.remove(cacheKey);
-            credential = null;
-        }
-
-        return credential;
+        return foundValue;
     }
 
     @NonNull
-    private Map<String, AccountRecord> getAccountsWithKeys() {
-        Logger.verbose(TAG, "Loading Accounts + keys...");
+    private Map<String, AccountRecord> loadAccountsWithKeys() {
+        final String methodTag = TAG + ":loadAccountsWithKeys";
+
+        Logger.verbose(methodTag, "Loading Accounts + keys...");
         final Iterator<Map.Entry<String, String>> cacheValues = mSharedPreferencesFileManager.getAllFilteredByKey(new Predicate<String>() {
             @Override
             public boolean test(String value) {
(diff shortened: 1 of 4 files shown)


nickbopp:
I need some feedback here about how to handle defensive copying of mutable objects.  This was previously not important since none of the objects were held on to beyond the lifetime of a call.  This is important for both accepting input (Save methods) as well as producing output (getAll*).  Extra defensive copies, especially in the 'getAll' case will require extra performance overhead, albeit much less than the JSON deserialization costs.

On the OneAuth side, we're not making edits on Credential/AccountRecord objects afer getting them from the SharedPreferencesAccountCredentialCache layer, so I'm not sure it's worth the penalty to try to copy.  It would be good to hear whether this would be a valuable tradeoff to make for MSAL.Android.

---
In reply to: [1396437318](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1929#issuecomment-1396437318) [](http://example.com/codeflow?ancestors=1396437318)

iamgusain:
> I need some feedback here about how to handle defensive copying of mutable objects. This was previously not important since none of the objects were held on to beyond the lifetime of a call. This is important for both accepting input (Save methods) as well as producing output (getAll*). Extra defensive copies, especially in the 'getAll' case will require extra performance overhead, albeit much less than the JSON deserialization costs.
> 
> On the OneAuth side, we're not making edits on Credential/AccountRecord objects afer getting them from the SharedPreferencesAccountCredentialCache layer, so I'm not sure it's worth the penalty to try to copy. It would be good to hear whether this would be a valuable tradeoff to make for MSAL.Android.

Can you elaborate on "defensive copying of mutable objects". Which objects are we talking about here?

nickbopp:
> > I need some feedback here about how to handle defensive copying of mutable objects. This was previously not important since none of the objects were held on to beyond the lifetime of a call. This is important for both accepting input (Save methods) as well as producing output (getAll*). Extra defensive copies, especially in the 'getAll' case will require extra performance overhead, albeit much less than the JSON deserialization costs.
> > On the OneAuth side, we're not making edits on Credential/AccountRecord objects afer getting them from the SharedPreferencesAccountCredentialCache layer, so I'm not sure it's worth the penalty to try to copy. It would be good to hear whether this would be a valuable tradeoff to make for MSAL.Android.
> 
> Can you elaborate on "defensive copying of mutable objects". Which objects are we talking about here?

@iamgusain 

The `List<Credential>` and `List<AccountRecord>` results of getAccounts() and getCredentials().  Those return a new collection, but still reference the same objects stored in our cache.  Since those objects are mutable, a caller could accidentally mutate the objects in the cache.  Previously, the List<Credential> returned by these functions was a new list of completely brand new objects, so whether the caller altered them in the future made no difference.

The same is true for the values saved in 'saveAccount' and 'saveCredential'.  Since they're just saving the reference to the passed in object, it's possible the caller could alter them afterwards.  That's why I needed to change a test above -- it was creating a Credential, calling saveCredential once, then altering a field and calling it a second time.  That actually wound up altering the original cached reference as well.

It really just comes down to expectations of the callers.  I don't expect anything malicious here, but it's possible to make mistakes.  I explored what it would take to account for this, and it's a non-trivial change.  There is the perf penalty of cloning every object on save* calls and getAll* calls, and the need to implement a new virtual Credential clone() method that knows how to copy each of the 5 subclasses of Credential correctly.

In the MsalCpp case, I can see all the paths where we use these things, and the need to defensively copy them is wasted, because the few paths we have don't mutate them, and we are about to translate them in to MsalCpp types anyway, which will be unique copies.  But I wasn't sure if those types got used in higher-level operations that were of concern in the MSAL/ADAL/Broker flows.

---
In reply to: [1404534860](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1929#issuecomment-1404534860) [](http://example.com/codeflow?ancestors=1404534860)

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,112 +107,102 @@ public SharedPreferencesAccountCredentialCache(
...
-    public synchronized void saveAccount(@NonNull final AccountRecord accountToSave) {
+    public void saveAccount(@NonNull final AccountRecord accountToSave) {
         Logger.verbose(TAG, "Saving Account...");
         Logger.verbose(TAG, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
         final String cacheKey = mCacheValueDelegate.generateCacheKey(accountToSave);
         Logger.verbosePII(TAG, "Generated cache key: [" + cacheKey + "]");
 
-        // Perform any necessary field merging on the Account to save...
-        final AccountRecord existingAccount = getAccount(cacheKey);
+        synchronized (mCacheLock) {
```
iamgusain, line 160: should we check for mLoaded outside of the `synchronized` block to avoid it,
I think that way we can avoid the overhead of  `synchronized` block from each method

Same in other places. #WontFix

nickbopp: I don't think it's safe to access mLoaded outside of the lock.  'waitForInitialLoad' is intended to only be called while holding the lock, because it will also '.wait()' on it if necessary, which requires ownership of the lock.  And we want to then hold the lock for the entirety of touching the cache variables afterwards.

Even if we could avoid taking the lock to wait, we need to immediately take the lock for the following operations that getAccount and then save in the cache.

nickbopp:
I now have the cloning/defensive copying in place.

---
In reply to: [1405909824](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1929#issuecomment-1405909824) [](http://example.com/codeflow?ancestors=1405909824)

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,112 +107,114 @@ public SharedPreferencesAccountCredentialCache(
...
+                mCacheLock.wait();
+            } catch (InterruptedException e) {}
+        }
     }
 
     @Override
-    public synchronized void saveAccount(@NonNull final AccountRecord accountToSave) {
+    public void saveAccount(@NonNull final AccountRecord accountInput) {
+        AccountRecord accountToSave = null;
+        try { accountToSave = (AccountRecord) accountInput.clone(); } catch (CloneNotSupportedException e) {}
```
matthewyu-ms: If accountInput.clone() fails and throws an exception, we still continue to proceed to saving account with accountToSave which can potentially be null right? #Resolved

matthewyu-ms: Same with other places where we call .clone()

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -285,15 +304,24 @@ public boolean test(String value) {
...
             if (null == credential) {
                 Logger.warn(methodTag, CREDENTIAL_DESERIALIZATION_FAILED);
-            } else {
-                credentials.put(cacheKey, credential);
+            } else if ((AccessTokenRecord.class == clazz && EMPTY_AT.equals(credential))
+                || (RefreshTokenRecord.class == clazz && EMPTY_RT.equals(credential))
+                || (IdTokenRecord.class == clazz) && EMPTY_ID.equals(credential)) {
+                // The returned credential came back uninitialized...
+                // Remove the entry and return null...
+                Logger.warn(TAG, "The returned Credential was uninitialized. Removing...");
```
matthewyu-ms: methodTag instead of TAG #Resolved

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: These clone()'s actually can't fail.  clone() is a bit weird in Java -- the default Object method defines clone() as protected, and if you want a type to actually be cloneable, you 'implements Cloneable' and then make the clone() method public.  The signature of clone() has the Java checked exception 'CloneNotSupportedException' in it, in case you wind up calling clone() on some object hierarchy that doesn't actually support it.

In this case, our AccountCredentialBase-derived objects do support it, and there's no case where this can fail.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: I'll get these fixed up.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: I'll add some logging and handling here to handle this, but if we land here, we have hit developer error.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,112 +107,114 @@ public SharedPreferencesAccountCredentialCache(
...
         mCacheValueDelegate = accountCacheValueDelegate;
+        new Thread(() -> load()).start();
+    }
+
+    private void load()
+    {
+        synchronized (mCacheLock) {
+            try {
+                mCachedAccountRecordsWithKeys = loadAccountsWithKeys();
+                Logger.info(TAG, "Loaded " + mCachedAccountRecordsWithKeys.size() + " AccountRecords");
```
iamgusain: nit: methodTag 
TAG + ":load" #Resolved

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
iamgusain: We should log the CloneNotSupportedException ? #Resolved

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,112 +107,114 @@ public SharedPreferencesAccountCredentialCache(
...
-            Logger.warn(TAG, "The returned Account was uninitialized. Removing...");
-            mSharedPreferencesFileManager.remove(cacheKey);
-            account = null;
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
+            foundValue = mCachedAccountRecordsWithKeys.get(cacheKey);
         }
 
-        return account;
+        try { if (foundValue != null) foundValue = (AccountRecord) foundValue.clone(); } catch (CloneNotSupportedException e) {}
```
iamgusain: nit: please expand this, and add log for exception. #Resolved

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,112 +107,114 @@ public SharedPreferencesAccountCredentialCache(
...
-                || (IdTokenRecord.class == clazz) && EMPTY_ID.equals(credential)) {
-            // The returned credential came back uninitialized...
-            // Remove the entry and return null...
-            Logger.warn(TAG, "The returned Credential was uninitialized. Removing...");
-            mSharedPreferencesFileManager.remove(cacheKey);
-            credential = null;
-        }
-
-        return credential;
+        try { if (foundValue != null) foundValue = (Credential) foundValue.clone(); } catch (CloneNotSupportedException e) {}
```
iamgusain: same here, expand and log #Resolved

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
matthewyu-ms: If we ever do hit a cloning error, it would be good in that (rare) scenario to return early instead of having an NPE. And yeah logging would help in this case too

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
         Logger.verbose(TAG, "Init: " + TAG);
         mSharedPreferencesFileManager = sharedPreferencesFileManager;
         mCacheValueDelegate = accountCacheValueDelegate;
+        new Thread(() -> load()).start();
```
shahzaibj, line 110: We should probably use an ExecutorService to manage these threads

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+    }
+
+    private void load()
+    {
```
shahzaibj:
```suggestion
    private void load() {
```

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+        final String methodTag = TAG + ":load";
+
+        synchronized (mCacheLock) {
+            try {
+                mCachedAccountRecordsWithKeys = loadAccountsWithKeys();
+                Logger.info(methodTag, "Loaded " + mCachedAccountRecordsWithKeys.size() + " AccountRecords");
+                mCachedCredentialsWithKeys = loadCredentialsWithKeys();
+                Logger.info(methodTag, "Loaded " + mCachedCredentialsWithKeys.size() + " Credentials");
+            } catch (Throwable t) {
```
shahzaibj:
```suggestion
            } catch (final Throwable t) {
```

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+                Logger.error(methodTag, "Failed to load initial credentials from SharedPreferences", t);
```
shahzaibj: I believe credentials here could mean either accounts records or credentials i.e. any of them could've failed to load to cause this log message

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+                mCacheLock.notifyAll();
+            }
+        }
+    }
+
+    private void waitForInitialLoad() {
+        while (!mLoaded) {
+            try {
+                mCacheLock.wait();
+            } catch (InterruptedException e) {}
```
shahzaibj:
```suggestion
            } catch (final InterruptedException e) {}
```

shahzaibj: Do we want to log a message in the catch block?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
-    public synchronized void saveAccount(@NonNull final AccountRecord accountToSave) {
-        Logger.verbose(TAG, "Saving Account...");
-        Logger.verbose(TAG, "Account type: [" + accountToSave.getClass().getSimpleName() + "]");
+    public void saveAccount(@NonNull final AccountRecord accountInput) {
+        final String methodTag = TAG + ":saveAccount";
+
+        AccountRecord accountToSave = null;
+        try {
+            accountToSave = (AccountRecord) accountInput.clone();
+        } catch (CloneNotSupportedException e) {
```
shahzaibj:
```suggestion
        } catch (final CloneNotSupportedException e) {
```

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+            Logger.error(methodTag, "Failed to clone AccountRecord", e);
```
shahzaibj, line 151: So we are not going to save the account in this case? Should we just throw an Exception rather than returning silently out of this method?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
     @Override
-    public synchronized void saveCredential(@NonNull Credential credentialToSave) {
-        Logger.verbose(TAG, "Saving credential...");
+    public void saveCredential(@NonNull Credential credentialInput) {
+        final String methodTag = TAG + ":saveCredential";
+
+        Credential credentialToSave = null;
+        try {
+            credentialToSave = (Credential) credentialInput.clone();
+        } catch (CloneNotSupportedException e) {
```
shahzaibj:
```suggestion
        } catch (final CloneNotSupportedException e) {
```

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,113 +107,150 @@ public SharedPreferencesAccountCredentialCache(
...
+            Logger.error(methodTag, "Failed to clone Credential", e);
```
shahzaibj, line 184: So we are not going to save the credential in this case? Should we just throw an Exception rather than returning silently out of this method?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -224,14 +268,17 @@ public boolean test(String value) {
                 );
 
                 if (null == account) {
-                    Logger.warn(TAG, ACCOUNT_RECORD_DESERIALIZATION_FAILED);
+                    Logger.warn(methodTag, ACCOUNT_RECORD_DESERIALIZATION_FAILED);
+                } else if (EMPTY_ACCOUNT.equals(account)) {
+                    Logger.warn(methodTag, "The returned Account was uninitialized. Removing...");
```
shahzaibj, line 276: I'm wondering why might this ever happen. Have you seen this happen in your testing?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -515,16 +592,20 @@ public boolean removeAccount(@NonNull final AccountRecord accountToRemove) {
...
-        }
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
+            boolean accountRemoved = false;
+            if (mSharedPreferencesFileManager.keySet().contains(cacheKey))
+            {
+                mSharedPreferencesFileManager.remove(cacheKey);
+                accountRemoved = true;
+            }
+            Logger.info(methodTag, "Account was removed? [" + accountRemoved + "]");
```
shahzaibj, line 630: Based on the code, I think this log line may say `false` even in the case where Account was actually not present in `mSharedPreferencesFileManager`. I'm not sure if this should mark as removed FALSE vs removal not actually possible.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: This is a single one-shot job at the load of the cache.  I modeled this largely after how Android source's SharedPreferences initially loads the XML file into cached memory.

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/SharedPreferencesImpl.java#134

Is there an existing Executor in use as a general thread pool that I could post this job to in AndroidCommon land without spinning up a new thread?  Priority #1 was not keeping a new thread worth of resources around for a long time to do this one job, but not spinning one up in the first place would be an improvement.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: This exception needing to be handled at all is an oddity of the Java implementation of `clone()`.  Because we've implemented it on this base type, we can't actually end up here.  This 'throws' is still part of the method signature though, in case you've managed to somehow call this on an object that doesn't implement Cloneable but still exposed a public method called `clone()` that called into the protected `super.clone()`:
https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/lang/Object.java;l=261?q=object.clone&sq=

I don't think we should give the higher level code any more conditions to handle -- logging this impossibility here seems enough.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: See above

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: There's nothing new here, this just moved around a bit with the caching changes. With that said, I think this looks correct.  The only case where something isn't removed is when it isn't present.  So I don't think there's any ambiguity in using a boolean. All the callers here also handle the boolean and expect these conditions.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: I don't know of any production cases, but there were test cases failing when I didn't migrate this logic from the existing code. I believe some of the tests were inserting intentionally broken objects and seeing how this layer handled them.

iamgusain (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
shahzaibj: We have some in [CommandDisptacher ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java#L91)that we use...these are private and specific to that class though.

You may want to just create an executor inside `SharedPreferencesAccountCredentialCache.java` and delegate your load operations to that executor.

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: I looked into this, and the thread from newSingleThreadExecutor will stick around forever, which we don't want for this single job.  For running a single job, without having a better long lived queue to share at a wider level, I think this is the best we can do.

PR #1931, opened 2023-01-20
Moving classes from challengehandlers to certbasedauth

melissaahn:
# Summary
Certificate Based Authentication classes have effectively taken over the challengehandlers package, so this PR moves the CertBasedAuth associated classes into their own package, certbasedauth. 
I made sure to rebuild and test with the msaltestapp to ensure that nothing breaks with this move.

If suggested, I can also break down into further packages within this one (challengehandlers, managers, dialogs, etc.), but let me know what you all think.


fadidurah (approved):
LGTM!

rpdome (approved):

PR #1937, opened 2023-01-27
Fixes for the new mutiple device registration API

p3dr0rv:
Related PR https://github.com/AzureAD/ad-accounts-for-android/pull/2122


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1544,8 +1544,11 @@ public enum API {
 
         /**
          * Broker api path constant for execute device registration protocols.
+         * Note: The path was updated because in release 9.0.1 a part of the NEW WPJ API was exposed
```
shahzaibj, line 1547: I'm not sure I fully understand this comment. Did we introduce a bug that we're trying to fix? Do we have a link to the bug that we can add here?

p3dr0rv: is not a bug, but it will become a bug if we do not change it.
Why. if we do not change the path in the future
an app using the new API may call a broker using the version 9.0.1 and this will execute the protocols, but the protocols here will work using the legacy WPJ controller, which can result in strange behavior as these API's are designed to work with the new WPJ controller.
with this change we will prevent the new API from being able to communicate with these protocols that work using the legacy WPJ controller

I also had a check to prevent this error.
see: https://github.com/AzureAD/ad-accounts-for-android/pull/2122/files#diff-8bef4ec73159bb746ab5fe18efb624925fcc12c192dd583103556e25e818b6c5

rpdome: @shahzaibj he accidentally exposed this path in previous releases - meaning that the new API in this release might be able to talk with the half-baked older broker.

This change is to make sure the new API are not compatible with older broker.

rpdome (approved):

shahzaibj (approved):

PR #1938, opened 2023-01-27
January UI maintenance (plus unit test fix)

fadidurah:
PR for this month's UI Automation maintenance work. Just making changes to make BrokerHost flighting UI tests more reliable, and updating as BrokerHost UI changed. also fixed some telemetry unit tests.

### DIFF
--- a/common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
+++ b/common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
@@ -179,7 +179,6 @@ public class TelemetryTest {
     @Test
     public void testTelemetryMap() {
         Telemetry.emit(new HttpStartEvent().putMethod("GET"));
-        Assert.fail("Intentional fail");
         testTelemetryMapHelper();
         testTelemetryMapHelper();
     }
--- a/common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
+++ b/common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
@@ -22,6 +22,8 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.telemetry;
 
+import com.microsoft.identity.common.java.logging.DiagnosticContext;
+import com.microsoft.identity.common.java.logging.RequestContext;
 import com.microsoft.identity.common.java.telemetry.events.HttpStartEvent;
 import com.microsoft.identity.common.java.telemetry.observers.ITelemetryAggregatedObserver;
 import com.microsoft.identity.common.java.telemetry.observers.ITelemetryDefaultObserver;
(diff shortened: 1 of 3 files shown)


melissaahn (approved):

common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
```diff
@@ -46,6 +48,15 @@ public void setup() {
                 .isDebugging(false)
                 .defaultConfiguration(new TelemetryConfiguration())
                 .build();
+
+        if (DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.CORRELATION_ID) == null) {
```
shahzaibj, line 52: Why exactly is this needed?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -357,8 +357,17 @@ public void overwriteFlights(@Nullable final String flightsJson) {
         UiAutomatorUtils.obtainChildInScrollable("Overwrite flights (When BrokerHost is the active broker)");
         // input flights string in flights input box
         UiAutomatorUtils.handleInput("com.microsoft.identity.testuserapp:id/editTextFlights", flightsJson);
+        // Set to use LocalStorage Flights
```
shahzaibj, line 360: What happens if we don't set this?

shahzaibj: What is the default flights provider in BrokerHost? Is there any default?

common4j/src/test/com/microsoft/identity/common/java/telemetry/TelemetryTest.java
fadidurah: Amit asked for help with some common unit tests. We realized that on one of the pipelines, instead of setting correlation id to "UNSET", it was actually showing up as null and causing an exception in 4 unit tests. This is happening on one particular pipeline but seems to work fine on others running the same test cases. Here's the pipeline experiencing this issue, you can see the 4 failed telemetry tests under the tests tab https://identitydivision.visualstudio.com/Engineering/_build?definitionId=1515&_a=summary

Ideally, I would spend more time diagnosing why this is happening, but we needed these passing to get code cov percentages and to have them working for release.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
fadidurah: Some recent UI updates in brokerhost made it so that the update flight and overwrite flight buttons were not enabled unless this setting was used. looks like there are no default flights set

shahzaibj (approved):

PR #1940, opened 2023-01-29
Capture span status and error codes in missing scenarios

shahzaibj:
Span Status and error codes were not being captured in some scenarios to addressed that. 

An exception is not always thrown during an error and sometimes the error is just present on the Acquire Token Result. So we will read it from there to set it on the span.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
@@ -25,6 +25,8 @@ package com.microsoft.identity.common.java.commands;
 import com.microsoft.identity.common.java.WarningType;
 import com.microsoft.identity.common.java.commands.parameters.InteractiveTokenCommandParameters;
 import com.microsoft.identity.common.java.controllers.BaseController;
+import com.microsoft.identity.common.java.controllers.ExceptionAdapter;
+import com.microsoft.identity.common.java.exception.BaseException;
 import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.opentelemetry.AttributeName;
 import com.microsoft.identity.common.java.opentelemetry.OTelUtility;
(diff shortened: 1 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
```diff
@@ -79,7 +81,23 @@ public AcquireTokenResult execute() throws Exception {
...
+
+                if (result == null) {
+                    span.setStatus(StatusCode.ERROR, "empty result");
+                } else if (result.getSucceeded()) {
+                    span.setStatus(StatusCode.OK);
+                } else {
+                    final BaseException exception = ExceptionAdapter.exceptionFromAcquireTokenResult(result, getParameters());
+                    if (exception != null) {
+                        span.recordException(exception);
+                        span.setStatus(StatusCode.ERROR);
```
fadidurah, line 94: I feel like setting statusCode to ERROR can be handled inside the recordException call, to avoid someone recording the exception without updating the status. Are there any scenarios where we record an exception but do not set the status to error?

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
```diff
@@ -79,7 +81,23 @@ public AcquireTokenResult execute() throws Exception {
 
                 span.setAttribute(AttributeName.controller_name.name(), controller.getClass().getSimpleName());
 
-                return controller.acquireToken((InteractiveTokenCommandParameters) getParameters());
+                final AcquireTokenResult result = controller.acquireToken((InteractiveTokenCommandParameters) getParameters());
+
+                if (result == null) {
+                    span.setStatus(StatusCode.ERROR, "empty result");
```
melissaahn, line 87: Could "empty result" and "empty exception" be set as static strings in BaseCommand and referenced from there in both classes?

melissaahn (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
shahzaibj: We don't own the `recordException` so we can't put code inside that method.

common4j/src/main/com/microsoft/identity/common/java/commands/InteractiveTokenCommand.java
shahzaibj: We can put it there, but for now I've merged the PR in as I have a few PRs to merge in. I can address this in the future.

PR #1942, opened 2023-02-01
Add application identifier during authorization request initialization

iamgusain:
### What
Add application identifier during authorization request initialization

### Why
This is later used to filter the credentials from the cache, if this is not set it results in cache miss, and silent requests might not be served from cache.

### How
Setting the application identifier in authorization request based on the application identifier passed in token parameters.

### Testing
Tested locally with unitTest in broker testATSWithPopAuthenticationSchemeWithClientKey, which was failing before this change. it passes after the change.
Also added change to compare authscheme to SCHEME_POP_WITH_CLIENT_KEY and SCHEME_POP in case access token type is "pop", as both of these scheme set the token type as "pop" in cache.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
@@ -24,6 +24,8 @@ package com.microsoft.identity.common.java.cache;
 
 import static com.microsoft.identity.common.java.AuthenticationConstants.DEFAULT_SCOPES;
 
+import com.microsoft.identity.common.java.authscheme.PopAuthenticationSchemeInternal;
+import com.microsoft.identity.common.java.authscheme.PopAuthenticationSchemeWithClientKeyInternal;
 import com.microsoft.identity.common.java.dto.AccessTokenRecord;
 import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.common.java.dto.Credential;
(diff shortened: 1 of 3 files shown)


shoatman (approved):
Thanks Amit.

shoatman (approved):

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -241,7 +244,14 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
...
                 }
 
-                matches = matches && authScheme.equalsIgnoreCase(atType);
+                if (TokenRequest.TokenType.POP.equalsIgnoreCase(atType)) {
+                    matches = matches && (
+                            authScheme.equalsIgnoreCase(PopAuthenticationSchemeWithClientKeyInternal.SCHEME_POP_WITH_CLIENT_KEY)
+                            || authScheme.equalsIgnoreCase(PopAuthenticationSchemeInternal.SCHEME_POP)
+                    );
+                } else {
+                    matches = matches && authScheme.equalsIgnoreCase(atType);
```
shahzaibj, line 253: What exactly are we matching against here? Bearer?

iamgusain: Bearer or Pop

shahzaibj: but PoP is in the `if` block....I believe only possibility for else if Bearer?

shahzaibj (approved):

PR #1944, opened 2023-02-02
Setting sub error codes to UiRequiredException for MsalUiRequiredException

melissaahn:
## Summary
This PR addresses a request from Teams to expose the server sub error code when we return a `MsalUiRequiredException`. This sub error code can help Teams provide the user with a more meaningful message when prompting for reauthentication. 

When generated due to an eSTS response, the only time `MsalUiRequiredException` is created is within `MsalExceptionAdapter`, where the `BaseException` transform into `MsalUiRequiredException` only if the `BaseException` is an instance of `UiRequiredException`. 
Fortunately, `UiRequiredException` already has a field for an OAuth sub error code via `ServiceException`, so the only work that needed to be done was:
- Make sure the OAuth sub error field is set properly within `UiRequiredException` when it is created due to an error received from the server (interaction_Required and invalid_grant error cases). 
- Add a corresponding OAuth sub error field to `MsalUiRequiredException` and set it based on `UiRequiredException` in the one place it is created. 

This PR addresses the former point, while [this PR](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1758) addresses the latter. 
Logic was added to set the sub error code field in `UiRequiredException` within the broker exception adapter . Note that the sub error code field was already being set in `ExceptionAdapter` within the `getExceptionFromTokenErrorResponse` method. 

Let me know if I seem to be missing anything else.

I tested by attempting to silently auth with an account with a changed password (got invalid_grant error code with bad_token subcode).

## Related PR to review
- https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1758


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -314,10 +314,16 @@ private BaseException getBaseExceptionFromExceptionType(@NonNull final String ex
...
+            final String errorCode = brokerResult.getErrorCode();
             baseException = new UiRequiredException(
                     brokerResult.getErrorCode(),
                     brokerResult.getErrorMessage()
             );
+            if (OAuth2ErrorCode.INTERACTION_REQUIRED.equalsIgnoreCase(errorCode) ||
+                    OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(errorCode)) {
+                ((UiRequiredException) baseException).setOauthSubErrorCode(brokerResult.getSubErrorCode());
+            }
+
```
shahzaibj: You may choose to create a helper method called `getUiRequiredException` or `createUiRequiredException` and move this code there and then just invoke that method here.

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
melissaahn: Done!

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -507,6 +504,25 @@ private ServiceException getServiceException(@NonNull final BrokerResult brokerR
...
+    /**
+     * Helper method to retrieve UiRequiredException from BrokerResult
+     *
+     * @return {@link com.microsoft.identity.common.java.exception.UiRequiredException}
+     */
+    @NonNull
+    private UiRequiredException getUiRequiredException(@NonNull final BrokerResult brokerResult) {
+        final String errorCode = brokerResult.getErrorCode();
+        final UiRequiredException exception = new UiRequiredException(
+                brokerResult.getErrorCode(),
```
mohitc1: >brokerResult.getErrorCode() [](http://example.com/codeflow?start=16&length=27)

Can be replaced with errorCode you just created.

mohitc1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
melissaahn: Done!

PR #1948, opened 2023-02-06
Remove OTel SDK from common and common4j

shahzaibj:
The otel sdk package is causing size increase for common and common4j so removing it from these libs. This means we cannot capture `parent_span_name` on the spans, but that's fine as we would still have `parent_span_id` and can look it up that way.


iamgusain:
Let's also add to changelog.

iamgusain (approved):
Try squash commit. :)

shahzaibj:
> Let's also add to changelog.

Done here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1948/commits/9652125148da00b47109036f58a4f6b0e62f1f8a

PR #1953, opened 2023-02-08
Add a Basic Encryptor Interface + Add support to Encrypt / Decrypt using AES GCM

shahzaibj:
Related: https://github.com/AzureAD/ad-accounts-for-android/pull/2147

I was primarily interested adding support for decrypting with AES GCM and in writing tests for BasicDecryptor here: https://github.com/AzureAD/ad-accounts-for-android/pull/2147
https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/2194698

The BasicDecryptor is used in a number of places so it would be good to get some tests for it. While writing tests I noticed the need to encrypt prior to trying out decrypt and noticed there was no BasicEncryptor. That seemed strange so I've also added a basic encryptor here.

**Testing**: Tests are in the associated broker repo and utilizing Broker Crypto Factory: https://github.com/AzureAD/ad-accounts-for-android/pull/2147/files

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
@@ -125,7 +125,7 @@ public class BasicDecryptor implements IDecryptor {
             throws ClientException {
         final Cipher cipher = mCryptoFactory.getCipher(decryptAlgorithm);
         try {
-            final GCMParameterSpec spec = new GCMParameterSpec(tagLength * 8, iv);
+            final GCMParameterSpec spec = new GCMParameterSpec(tagLength * Byte.SIZE, iv);
             cipher.init(Cipher.DECRYPT_MODE, key, spec);
 
             if (aad != null) {
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
@@ -62,7 +62,7 @@ public class BasicDecryptor implements IDecryptor {
                 new ICryptoOperation<byte[]>() {
                     @Override
                     public byte[] perform() throws ClientException {
-                        return decryptWithCipher(key, decryptAlgorithm, iv, dataToBeDecrypted);
+                        return decryptWithIv(key, decryptAlgorithm, iv, dataToBeDecrypted);
                     }
                 }
         );
@@ -82,7 +82,7 @@ public class BasicDecryptor implements IDecryptor {
                 new ICryptoOperation<byte[]>() {
                     @Override
                     public byte[] perform() throws ClientException {
-                        return decryptWithCipher(
+                        return decryptWithGcm(
                                 key, decryptAlgorithm, iv, dataToBeDecrypted, tagLength, aad
                         );
                     }
@@ -90,10 +90,10 @@ public class BasicDecryptor implements IDecryptor {
         );
     }
 
-    private byte[] decryptWithCipher(@NonNull final Key key,
-                                     @NonNull final String decryptAlgorithm,
-                                     final byte[] iv,
-                                     byte[] dataToBeDecrypted)
+    private byte[] decryptWithIv(@NonNull final Key key,
+                                 @NonNull final String decryptAlgorithm,
+                                 final byte[] iv,
+                                 byte[] dataToBeDecrypted)
             throws ClientException {
         final Cipher cipher = mCryptoFactory.getCipher(decryptAlgorithm);
         try {
@@ -116,12 +116,12 @@ public class BasicDecryptor implements IDecryptor {
     }
 
     @SuppressWarnings("NewApi")
-    private byte[] decryptWithCipher(@NonNull final Key key,
-                                     @NonNull final String decryptAlgorithm,
-                                     final byte[] iv,
-                                     byte[] dataToBeDecrypted,
-                                     final int tagLength,
-                                     @Nullable final byte[] aad)
+    private byte[] decryptWithGcm(@NonNull final Key key,
+                                  @NonNull final String decryptAlgorithm,
+                                  final byte[] iv,
+                                  byte[] dataToBeDecrypted,
+                                  final int tagLength,
+                                  @Nullable final byte[] aad)
             throws ClientException {
         final Cipher cipher = mCryptoFactory.getCipher(decryptAlgorithm);
         try {
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
@@ -132,9 +132,7 @@ public class BasicDecryptor implements IDecryptor {
                 cipher.updateAAD(aad);
             }
 
-            return cipher.doFinal(
-                    dataToBeDecrypted, iv.length, dataToBeDecrypted.length - iv.length
-            );
+            return cipher.doFinal(dataToBeDecrypted);
         } catch (final BadPaddingException e) {
             throw new ClientException(ClientException.BAD_PADDING, e.getMessage(), e);
         } catch (final IllegalBlockSizeException e) {
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
@@ -35,8 +35,10 @@ import java.security.Key;
 import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
 import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.IvParameterSpec;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.NonNull;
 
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
```diff
@@ -49,4 +49,20 @@ byte[] decrypt(@NonNull final Key key,
...
+     * @param key               the key to decrypt with.
+     * @param decryptAlgorithm  algorithm to decrypt with.
+     * @param iv                an initialization vector (IV).
+     * @param dataToBeDecrypted the data to be encrypted.
+     * @return a decrypted byte array.
+     */
+    byte[] decrypt(@NonNull final Key key,
+                   @NonNull final String decryptAlgorithm,
+                   final byte[] iv,
+                   final byte[] dataToBeDecrypted,
```
mohitc1: >final byte[] dataToBeDecrypted, [](http://example.com/codeflow?start=19&length=31)

why there's no tag parameter here?
Is it expected from caller to send dataToBeEncrypted with tag inside at?

shahzaibj: We have `tagLength` right below it. 

The actual tag is not required for decryption. We only care about the length of tag.

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicEncryptor.java
```diff
@@ -0,0 +1,148 @@
...
+            throw new ClientException(ClientException.INVALID_BLOCK_SIZE, e.getMessage(), e);
+        } catch (final InvalidKeyException e) {
+            throw new ClientException(ClientException.INVALID_KEY, e.getMessage(), e);
+        } catch (final InvalidAlgorithmParameterException e) {
+            throw new ClientException(ClientException.INVALID_ALG_PARAMETER, e.getMessage(), e);
+        }
+    }
+
+    @SuppressWarnings("NewApi")
+    private byte[] encryptWithCipher(@NonNull final Key key,
```
rpdome: encryptWithGcm?

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicEncryptor.java
```diff
@@ -0,0 +1,148 @@
...
+                    public byte[] perform() throws ClientException {
+                        return encryptWithCipher(
+                                key, encryptAlgorithm, iv, dataToBeEncrypted, tagLength, aad
+                        );
+                    }
+                }
+        );
+    }
+
+    private byte[] encryptWithCipher(@NonNull final Key key,
```
rpdome: encryptWithIv?

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
shahzaibj: I mean I guess the answer to your question would be that `dataToBeEncrypted` is meant to be exact encrypted data produced through an AES GCM encryption. While decryption we don't need to know the tag, we just need to know the length of the tag. I would think that eSTS will supply us the tag length. Let me know if I missing something.

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
```diff
@@ -66,20 +68,73 @@ public byte[] perform() throws ClientException {
...
+            throw new ClientException(ClientException.INVALID_BLOCK_SIZE, e.getMessage(), e);
+        } catch (final InvalidKeyException e) {
+            throw new ClientException(ClientException.INVALID_KEY, e.getMessage(), e);
+        } catch (final InvalidAlgorithmParameterException e) {
+            throw new ClientException(ClientException.INVALID_ALG_PARAMETER, e.getMessage(), e);
+        }
+    }
+
+    @SuppressWarnings("NewApi")
+    private byte[] decryptWithCipher(@NonNull final Key key,
```
rpdome: decryptWithGcm

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
```diff
@@ -66,20 +68,73 @@ public byte[] perform() throws ClientException {
...
+                    public byte[] perform() throws ClientException {
+                        return decryptWithCipher(
+                                key, decryptAlgorithm, iv, dataToBeDecrypted, tagLength, aad
+                        );
+                    }
+                }
+        );
+    }
+
     private byte[] decryptWithCipher(@NonNull final Key key,
```
rpdome: decryptWithIv

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
mohitc1: yes, we need tag as well to perform decryption. 
Response from ests has format
[Header].[EncryptedKey].[IV].[Payload].[Tag]

At the end data that needs to decrypted is payload+tag, so we need to concatenate them and tag length is used in setting up the spec.
If we pass payload and tag here, we would concatenate them and decrypt it. 
With the signature you have now, such concatenation would be exacted by caller.

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
```diff
@@ -49,4 +49,20 @@ byte[] decrypt(@NonNull final Key key,
...
+    /**
+     * Decrypt the given byte array.
+     *
+     * @param key               the key to decrypt with.
+     * @param decryptAlgorithm  algorithm to decrypt with.
+     * @param iv                an initialization vector (IV).
+     * @param dataToBeDecrypted the data to be encrypted.
+     * @return a decrypted byte array.
+     */
+    byte[] decrypt(@NonNull final Key key,
```
mohitc1: can name this differently to indicate it uses Aes Gcm.

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
shahzaibj: Okay, so if I understand correctly, the payload that eSTS sends us (that needs to be decrypted) doesn't actually include the tag? They seem to be sending us the raw tag separately per your response example. eSTS wants us to include the tag and then do the decryption?

I mean I can modify the interface, but the current interface is designed to be more generic encrypt / decrypt interface and NOT specific to the mentioned eSTS scenario. From what I saw in examples online, it seemed the cipherText given to the AES GCM decrypt logic would include the tag and thus while decryption the only thing needed to be known was the tag length and so I followed similar logic in my implementation.

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicEncryptor.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/8ef8a06a8f8c110c7e06918e087f97256b0ef69c

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicEncryptor.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/8ef8a06a8f8c110c7e06918e087f97256b0ef69c

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/8ef8a06a8f8c110c7e06918e087f97256b0ef69c

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/8ef8a06a8f8c110c7e06918e087f97256b0ef69c

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
shahzaibj: I've updated the name on the internal methods and kept the interface name unchanged.

shahzaibj: Alternatively, I can also refactor it to just take in parameter spec...that might actually be better.

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
shahzaibj: For now, I'm keeping this interface generic. We can build a wrapper around this for purposes of your specific scenario to perform the concatenation.

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
```diff
@@ -60,26 +62,79 @@ public byte[] decrypt(@NonNull final Key key,
...
+        try {
+            final GCMParameterSpec spec = new GCMParameterSpec(tagLength * 8, iv);
+            cipher.init(Cipher.DECRYPT_MODE, key, spec);
+
+            if (aad != null) {
+                cipher.updateAAD(aad);
+            }
+
+            return cipher.doFinal(
+                    dataToBeDecrypted, iv.length, dataToBeDecrypted.length - iv.length
```
mohitc1: >iv.length, dataToBeDecrypted.length - iv.length [](http://example.com/codeflow?start=39&length=47)

here also, do you expect iv to be part of dataToBeDecrypted?

mohitc1: It should just be cipher.doFinal(dataToBeDecrypted) since you taking iv separately

common4j/src/main/com/microsoft/identity/common/java/crypto/IDecryptor.java
mohitc1: Yes, they send tag separately. 
I get your point. I see that with Java Cipher/GCMParameterSpec the tag is generated internally and is part of final ciphertext you get after encryption. 
So, it seems fine to keep decrypt analogous. We combine text and tag in a layer above.

common4j/src/main/com/microsoft/identity/common/java/crypto/BasicDecryptor.java
shahzaibj: `iv` is being for making the parameter spec and not part of `cipher.doFinal` call. Here I'm just providing the length of the iv for offset.

If I make the change that you suggest then tests are failing. I have tests for these here: https://github.com/AzureAD/ad-accounts-for-android/pull/2147

mohitc1: That's because in encrypt you are adding iv to final ciphertext you return.
            encryptedText.put(iv);
            encryptedText.put(cipherText);
I don't think we should do that.

shahzaibj: Why shouldn't we do that? I looked at two solutions online that seemed to be doing this and I followed the same algorithm.

shahzaibj: Synced offline. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/bb1949549d262200094b11ee93c56ad7efec295e

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/crypto/IEncryptor.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Encrypt the given byte array.
+     *
+     * @param key               the key to encrypt with.
+     * @param encryptAlgorithm  algorithm to encrypt with.
+     * @param iv                an initialization vector (IV).
+     * @param dataToBeEncrypted the data to be encrypted.
+     * @return a decrypted byte array.
+     */
+    byte[] encrypt(@NonNull final Key key,
```
rpdome: we might want to rename (to Iv and Gcm) on the interface as well.

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/crypto/IEncryptor.java
rpdome: same for the decryptor?

rpdome: It's the mode name (CBC/GCM) and i think it's valid to make it explicit.

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1953/commits/48d6ab35db07766e6409149e68ff41f1cf25c08e

PR #1954, opened 2023-02-08
Add credential adapter methods for creating cache records

mohitc1:
Adding cache record methods in IAccountCredentialAdapter. The implementation is moved over from BrokerJoinedAccountController. These methods are used saving token and account to broker cache.

Moved out these implementations out of BrokerJoinedAccountController to re-used in new PrtV3 based token acquisition.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
@@ -24,15 +24,27 @@ package com.microsoft.identity.common.java.cache;
 
 import static com.microsoft.identity.common.java.AuthenticationConstants.DEFAULT_SCOPES;
 
+import com.microsoft.identity.common.java.authorities.Authority;
+import com.microsoft.identity.common.java.authscheme.AbstractAuthenticationScheme;
+import com.microsoft.identity.common.java.authscheme.PopAuthenticationSchemeInternal;
+import com.microsoft.identity.common.java.authscheme.PopAuthenticationSchemeWithClientKeyInternal;
+import com.microsoft.identity.common.java.commands.parameters.TokenCommandParameters;
+import com.microsoft.identity.common.java.crypto.IDevicePopManager;
 import com.microsoft.identity.common.java.dto.AccessTokenRecord;
 import com.microsoft.identity.common.java.dto.AccountRecord;
+import com.microsoft.identity.common.java.dto.Credential;
 import com.microsoft.identity.common.java.dto.CredentialType;
 import com.microsoft.identity.common.java.dto.IdTokenRecord;
 import com.microsoft.identity.common.java.dto.RefreshTokenRecord;
+import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.exception.ServiceException;
 import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftAccount;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftRefreshToken;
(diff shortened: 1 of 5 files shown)


shahzaibj (approved):
Just seems like code moved around. This looks fine to me.

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -268,6 +284,195 @@ public IdTokenRecord asIdToken(MicrosoftAccount msAccount, MicrosoftRefreshToken
...
+            accountRecord = new AccountRecord(microsoftStsAccount);
+        }
+
+        URL authority = parameters.getAuthority().getAuthorityURL();
+
+        // If token response has authority set it else fallback to request authority
+        if (!microsoftStsTokenResponse.getAuthority().isEmpty()) {
+            try {
+                authority = new URL(microsoftStsTokenResponse.getAuthority());
+            } catch (MalformedURLException e) {
```
somalaya: nit : final MalformedURLException e

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -268,6 +284,195 @@ public IdTokenRecord asIdToken(MicrosoftAccount msAccount, MicrosoftRefreshToken
...
+                Logger.error(TAG,
```
somalaya: nit : use methodTag instead of TAG

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -268,6 +284,195 @@ public IdTokenRecord asIdToken(MicrosoftAccount msAccount, MicrosoftRefreshToken
...
+            final MicrosoftStsAccount microsoftStsAccount = new MicrosoftStsAccount(
+                    new IDToken(microsoftStsTokenResponse.getIdToken()),
+                    clientInfo
+            );
+            accountRecord = new AccountRecord(microsoftStsAccount);
+        }
+
+        URL authority = parameters.getAuthority().getAuthorityURL();
+
+        // If token response has authority set it else fallback to request authority
```
somalaya: nit: If token response has authority set it, else fallback to request authority

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -268,6 +284,195 @@ public IdTokenRecord asIdToken(MicrosoftAccount msAccount, MicrosoftRefreshToken
...
+        }
+
+        final AzureActiveDirectoryCloud cloudEnv =
+                AzureActiveDirectory.getAzureActiveDirectoryCloud(authority);
+
+        if (cloudEnv != null) {
+            Logger.info(methodTag, "Using preferred cache host name...");
+            accountRecord.setEnvironment(cloudEnv.getPreferredCacheHostName());
+        } else {
+            accountRecord.setEnvironment(authority.getHost());
```
somalaya, line 332: is it possible to have authority as null here? I see that is being assigned above from parameters, but want to check if it can be null at this point?

somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
mohitc1: I don't think so. At least TokenCommandParameters always has it.

PR #1958, opened 2023-02-13
Update JWE Response 

mohitc1:
1. Added "Additional Authenticated Data" to JWE response.
2. Update the JweResponse APIs to return byte[]

These changes will be consumed by Session Key part of PRTv3.

Related PR https://github.com/AzureAD/ad-accounts-for-android/pull/2157

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
@@ -110,10 +110,10 @@ public class JweResponse {
         response.mAAD = header.getBytes(AuthenticationConstants.CHARSET_ASCII);
 
         if (split.length > 4) {
-            response.mAuthenticationTag = Base64.decode(split[4], Base64.URL_SAFE);
+            response.mAuthenticationTag = base64Decode(split[4], Base64.URL_SAFE, "Tag is not base64 url-encoded");
         }
 
-        final byte[] headerDecodedBytes = Base64.decode(header, Base64.URL_SAFE);
+        final byte[] headerDecodedBytes = base64Decode(header, Base64.URL_SAFE, "Header is not base url-encoded");
         final String decodedHeader = StringUtil.fromByteArray(headerDecodedBytes);
 
         final JSONObject jsonObject = new JSONObject(decodedHeader);
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
@@ -40,8 +40,7 @@ import lombok.Getter;
 import lombok.NonNull;
 import lombok.experimental.Accessors;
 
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
```diff
@@ -69,59 +74,76 @@ public static class JweHeader {
 
     JweHeader mJweHeader;
 
-    String mEncryptedKey;
+    private String mEncryptedKey;
```
rpdome, line 80: Can we list which fields are non-null, and which fields are nullable? (and in what case they're nullable)?

Alternatively, I would suggest branching this class out into V2 and V3. It's probably easier to detect that way if we're getting NOT a value that is expected in V3 response.

common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
```diff
@@ -138,4 +160,20 @@ public static JweResponse parseJwe(String jwe) throws JSONException {
...
         return response;
     }
+
+    /***
+     * Helper to perform base64 decoding with logging.
+     * @param input Input string
+     * @param flags
+     * @param failureMessage The message to log in case of failure.
+     */
+    public static byte[] base64Decode(@NonNull final String input, int flags, @NonNull final String failureMessage) {
```
rpdome: This one should probably go to common4j's StringUtil.

common4j/src/main/com/microsoft/identity/common/java/platform/JweResponse.java
mohitc1: Annotated the fields and added comment for nullable. I think the class is meant for JWE parsing and help layers above (PRTv2 and PRTv3) read the values. The calling flow would validate if particular field were required and is correct for them.

shahzaibj (approved):

common4j/src/test/com/microsoft/identity/common/java/platform/JweResponseTests.java
```diff
@@ -0,0 +1,70 @@
...
+import org.json.JSONException;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for {@link JweResponse}.
+ */
+@RunWith(JUnit4.class)
```
rpdome, line 38: Please add tests where the

1. nonnull/nullable values in the response is missing. Do we crash?
2. the response contains unexpected field. Do we crash?
3. Malformed response.

rpdome (approved):

common4j/src/test/com/microsoft/identity/common/java/platform/JweResponseTests.java
mohitc1: Added

PR #1961, opened 2023-02-14
Add flighting capability for common.

iamgusain:
### What
Adding flighting capability to common

### Why
To dynamically change the behavior of common code based on the flight value, which can be provided without code deployment

### How
Adding class CommonFlightManager which is used to set the FlightsProvider for CommonFlights.
This FlightsProvider is used to read the value of the flights. If no flightsProvider is set default Values will be used.

### Testing
Using the flighting capability to control whether or not to use NetworkCapability for checking network connectivity.

Corresponding Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2158

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
@@ -51,7 +51,7 @@ import static org.robolectric.Shadows.shadowOf;
 @RunWith(RobolectricTestRunner.class)
 @Config(shadows = {
         PowerManagerWrapperShadow.class
-})
+}, sdk = 28)
 public class DefaultConnectionServiceTest {
 
     protected Context mContext;
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [MINOR] Add flighting capability for common (#1961)
 - [MINOR] Add preferred browser support (#1957)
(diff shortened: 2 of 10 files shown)


common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
```diff
@@ -51,7 +51,7 @@
 @RunWith(RobolectricTestRunner.class)
```
rpdome, line 51: Add TODO: for API>28 tests.

rpdome (approved):
I assume the test for NetworkCapabilities will follow in a subsequent PR.

common/src/test/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionServiceTest.java
rpdome: actually, should this be 23? given that M = 23.

shahzaibj (approved):
It looks like code similar to what we have in broker repo. We should probably just port that code over to this repo, but it doesn't have to be done immediately.

iamgusain:
> It looks like code similar to what we have in broker repo. We should probably just port that code over to this repo, but it doesn't have to be done immediately.

https://github.com/AzureAD/ad-accounts-for-android/pull/2158

iamgusain:
> I assume the test for NetworkCapabilities will follow in a subsequent PR.

I will include telemetry for the new check in a separate PR.

PR #1962, opened 2023-02-14
Add metric to measure network check success, failure counts

iamgusain:
### What
Add metric to measure network check success, failure counts when using network capability for the check

### Why
To measure the success rate of the new check, as we gradually increase the flight exposure which controls it.

### How
Added two metrics for measuring success and failure counts.


shahzaibj:
I don't think we need this metric as these is already being captured as part of Spans because network check happens as part of ATS operation and we capture exception through recordException and that captures network not available.

shahzaibj:
> I don't think we need this metric as these is already being captured as part of Spans because network check happens as part of ATS operation and we capture exception through recordException and that captures network not available.

device network not available is our greatest error code today

image

iamgusain:
> I don't think we need this metric as these is already being captured as part of Spans because network check happens as part of ATS operation and we capture exception through recordException and that captures network not available.

I am specifically adding this to measure the success rate for the new network capability check, which I have put behind flight.
Once we know that it is behaving correctly, I plan to remove the flight and telemetry both.

shahzaibj:
> > I don't think we need this metric as these is already being captured as part of Spans because network check happens as part of ATS operation and we capture exception through recordException and that captures network not available.
> 
> I am specifically adding this to measure the success rate for the new network capability check, which I have put behind flight. Once we know that it is behaving correctly, I plan to remove the flight and telemetry both.

Ok, do we know on how many devices we will enable the flight? I'm just worried that since we are also capturing success in the counter, then if the flight is enabled on too many devices, then even with the aggregation we may end up getting too many of these generated. We may have to tweak how often OTel is aggregating because network check successes will happen too often.

iamgusain:
> > > I don't think we need this metric as these is already being captured as part of Spans because network check happens as part of ATS operation and we capture exception through recordException and that captures network not available.
> > 
> > 
> > I am specifically adding this to measure the success rate for the new network capability check, which I have put behind flight. Once we know that it is behaving correctly, I plan to remove the flight and telemetry both.
> 
> Ok, do we know on how many devices we will enable the flight? I'm just worried that since we are also capturing success in the counter, then if the flight is enabled on too many devices, then even with the aggregation we may end up getting too many of these generated. We may have to tweak how often OTel is aggregating because network check successes will happen too often.

We can start with a small number (1%), I just want to make sure that we don't get high % failure with the new check.

shahzaibj (approved):

somalaya (approved):

PR #1966, opened 2023-02-17
Adding Dialog Prompts to Remove Smartcard (at times that make sense)

melissaahn:
### Why the CBA dialog flow is being changed again
Feedback from testers of the NFC dialog flow showed that it would be beneficial to have dialog prompts that let the user know when they should remove their smartcard. For example, after the first NFC tap, once the client determines that there are/are not any errors and have read off the certificates, we should let the user know that they do not need to hold their smartcard to the device anymore. 

There are two main places that a user should see such a dialog for NFC:
- After initially tapping in order to retrieve certs from the smartcard (before the cert picker)
- After tapping in order to confirm PIN and sign with a private key

There is a potential use for the removal prompt dialog in the USB flow:
-After signing with the private key, and other processes involving the smartcard have completed.
    - UPDATE: Upon positive feedback, this is included in this PR.

Including this prompt in the USB flow could potentially decrease the amount of configuration change related errors, depending upon when most users will plug in their smartcard. If they plug in their smartcard only during the times our activity is running (when the authentication WebView is showing), then implementing such a change will keep the configuration changes occurring within our activities, which have the appropriate configChanges flags. The flip side of this is that a user who plugged in their smartcard within the host app activity will experience a configuration change within the host app activity upon our activities destroying. Logic has been added to detect if this is the case, and no removal prompt will be displayed if a smartcard is still detected by the end of the flow. 

### What main changes were made
- Added a new `ShowSmartcardRemovalPromptDialog` method to `DialogHolder` to be called when we want a user to remove their smartcard (after NFC taps, before sending auth results, for NFC and USB).
    - corresponding Dialog is `SmartcardRemovalPromptDialog`
- Made changes to `WebViewAuthorizationFragment` and `AzureActiveDirectoryWebViewClient` to set a callback that sends the auth result and finishes (in order to delay sending results until the user removes their smartcard). This callback will be invoked immediately if the cba challenge handler is null, and otherwise will be invoked based on the status of the challenge handler. A callback is needed here in order to make sure the user removes their smartcard before our library's activity destroys.
- `onPausedSmartcardDiscovery` renamed to `prepForNextUserInteraction`. Though we are stopping discovery in NFC's case, the emphasis is put on giving some time to prepare for an expected user interaction, which would be pausing for the user to remove their smartcard (in NFC case; USB implementation of this method just invokes callback).
- <strike>Got rid of the `onCancelCba()` method in smartcard dialogs. This was originally used to run logic when a smartcard was unexpectedly unplugged. To keep more consistent with the `connectionCallback` variable of the CBA managers, I created a `DisconnectedCallback` local variable that should be set for unexpected/expected unplugging on the USB side, or expected removal on the NFC side. </strike> Actually ended up keeping the `onCancelCba()` logic, as I was overthinking things. The method was renamed to be `onUnexpectedUnplug()` to be more specific and to avoid confusion with the cancelCallbacks.
    - Was able to get rid of callback generics for `AbstractSmartcardCertBasedAuthManager`.
- <strike>Consolidated the local `cancelCallback` interfaces attached to each dialog into a universal `ICancelCbaCallback`.</strike> Done in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1972#pullrequestreview-1316779182
- <strike>Consolidated the local `dismissCallback` interfaces attached to each dialog into a universal `IDismissCallback`</strike> Done in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1975
-Added a `disconnect` method to the NFC manager that should disconnect a smartcard and wait for the user to remove the smartcard.

### Testing
The Units tests associated with CBA have been updated to check for `SmartcardRemovalPromptDialog` being shown at correct times. 
I've manual tested different scenarios to make sure that the changes made to `AzureActiveDirectoryWebVIewClient` does not interfere with regular auth. I'll continue to manual test to make sure all bases that I can think of are covered.


common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/TestNfcSmartcardCertBasedAuthManager.java
```diff
@@ -101,5 +101,12 @@ public void mockConnect(final boolean isChanged) {
 
     public void mockDisconnect() {
         mIsConnected = false;
+        if (mDisconnectionCallback != null) {
```
fadidurah: what do we do if the callback is null? should we fail?

fadidurah: same question for the usb counterpart

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/UserChoiceDialog.java
```diff
@@ -94,25 +94,10 @@ public void onCancel(DialogInterface dialog) {
         mDialog = dialog;
     }
 
-    /**
-     * Should dismiss dialog and call the appropriate methods to help cancel the CBA flow.
-     */
-    @Override
-    void onCancelCba() {
```
fadidurah, line 101: Sorry, definitely something I missed, but how do we call this cancel method to dismiss the dialog, again?

fadidurah: Ahh, through SmartcardRemovalPromptDialog

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/UserChoiceDialog.java
```diff
@@ -36,15 +36,15 @@
...
+    private final ICancelCbaCallback mCancelCbaCallback;
 
     /**
      * Creates new instance of SmartcardDialog.
      *
      * @param activity Host activity.
      */
     public UserChoiceDialog(@NonNull final PositiveButtonListener positiveButtonListener,
-                            @NonNull final CancelCbaCallback cancelCbaCallback,
+                            @NonNull final ICancelCbaCallback cancelCbaCallback,
```
fadidurah: Since we don't cancel from this dialog anymore, do we still need to pass the cancel callback? I don't see it used in this class anymore.

fadidurah: I think same idea applies to SmartcardPromptDialog, SmartcardPinDialog, and others

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/YubiKitNfcSmartcardCertBasedAuthManager.java
```diff
@@ -104,18 +104,29 @@ public void invoke(@NonNull NfcYubiKeyDevice value) {
     @Override
     void stopDiscovery(@NonNull Activity activity) {
         synchronized (sDeviceLock) {
-            if (isDeviceConnected()) {
+            mNfcDevice = null;
```
fadidurah, line 112: We removed the check for isDeviceConnected?

fadidurah: Are we assuming that disconnect() from this class will be called before this method everytime?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -88,57 +89,84 @@ public Void processChallenge(ClientCertRequest request) {
...
+                        }
 
-                //Build and show Smartcard cert picker, which also handles the rest of the smartcard CBA flow.
-                mDialogHolder.showCertPickerDialog(
-                        certList,
-                        getSmartcardCertPickerDialogPositiveButtonListener(request),
-                        new SmartcardCertPickerDialog.CancelCbaCallback() {
-                            @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+                        final ICancelCbaCallback cancelCallback = getGeneralCancelCbaCallback(request);
+                        mDialogHolder.showCertPickerDialog(
```
fadidurah, line 116: I think some coments to explain why we are using showCertPickerDialog (I'm a little confused because we closed the connection already, right?)

fadidurah: I think some terminology is confusing (and this could very well be from CBA documentation side and out of your control), for example I was misinterpreting "CloseConnection" to mean something relating to closing a network connection, but it can refer to physically removing a smart card from the device. Just something to think about if we can create some terminology that would be more specific to what is happening, if not maybe we can increase documentation to make it easier for developers to understand what a method is doing.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -51,14 +52,39 @@ public NfcSmartcardCertBasedAuthChallengeHandler(@NonNull final Activity activit
...
+            public void onClosedConnection() {
+                mDialogHolder.dismissDialog();
+                mCbaManager.clearDisconnectionCallback();
+                //Helps prevent unnecessary callback trigger. Nfc discovery should only be active when
+                // the user is prompted to tap.
+                mCbaManager.stopDiscovery(mActivity);
+                callback.onClosedConnection();
+            }
+        });
+        mCbaManager.disconnect();
```
fadidurah: I see that we are setting the DisconnectCallback and then calling disconnect right after. Are there situations where CbaManager would call disconnect again even after the connection has been closed? If not, I think maybe we would not need to set the callback here and can just call the cancelation code:

                mCbaManager.clearDisconnectionCallback();
                //Helps prevent unnecessary callback trigger. Nfc discovery should only be active when
                // the user is prompted to tap.
                mCbaManager.stopDiscovery(mActivity);
                callback.onClosedConnection();

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -77,24 +103,27 @@ protected SmartcardPinDialog.PositiveButtonListener getSmartcardPinDialogPositiv
...
                     @Override
                     public void onCreateConnection() {
                         mDialogHolder.showSmartcardNfcLoadingDialog();
                         if (mCbaManager.isDeviceChanged()) {
-                            //In a future version, an error dialog with a custom message could be shown here instead of a general error.
-                            indicateGeneralException(methodTag, new Exception("Device connected via NFC is different from initially connected device."));
-                            request.cancel();
                             clearPin(pin);
-                            onPausedSmartcardDiscovery();
+                            request.cancel();
```
fadidurah, line 121: Does it matter if we clear pin or cancel request first?

fadidurah (commented):
Left some comments on this, will have another look soon 👍

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandler.java
melissaahn: `CloseConnection` refers to making sure the user removes their NFC smartcard before we proceed with the rest of the logic. For NFC, we want to only have the connection open when it is necessary (this is unlike USB), so I changed the logic to get the necessary data, then close the connection before proceeding with that data. So, before `pauseToCloseConnection` is called, we get the `pinAttemptsRemaining` and `certList`, then close the connection, since we don't need it to be open in order to show a certpicker dialog (using the certList).

I can add "smartcard" back into the method name, and maybe rename it to something like `pauseSmartcardConnection`, and maybe leave "close" out of it, if that's confusing. That would also make it a bit more applicable for the USB side (where the connection isn't being closed at all here, but we are pausing, in a sense, from reading off of the card).
I'll also add a comment before the call to explain that we're pausing the connection because we extracted all that we need for now, and don't need the connection at the moment.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: The callback is set right before disconnect() because disconnect() contains smartcard api specific logic that waits for the user to remove their smartcard, and then calls the disconnectCallback. So calling the cancellation code directly after won't really work here, but I think the equivalent of your point would be to pass the code into disconnect() as a callback parameter instead. I did that initially, but opted for setting the disconnectCallback instead, in order to keep the disconnect callback logic consistent. I could go back to passing as a callback parameter though if that would make more sense. That would essentially leave the disconnectcallback variable within the NFC manager useless, so I would have to rethink some of the shared logicof the abstract classes, and maybe add some generics back in.
I'll take some time to move stuff around again and get back to this comment.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: Doesn't matter. Tried to keep it consistent by clearing pin first then cancelling, but there isn't any meaning behind it.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/UserChoiceDialog.java
melissaahn: You can't see it in this PR file without expanding down a bit, but cancelCbaCallback is set to the class variable mCancelCbaCallback. The onCancel() method of it is called when the user presses the cancel button on these particular dialogs that have a cancel button.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/UserChoiceDialog.java
melissaahn: onCancelCba() was originally meant for calling some cleaning logic within the challenge handlers when the CBA flow was cancelled unexpectedly (unplugging the USB smartcard early). <strike>The name and purpose of this method is pretty confusing... so I instead decided to create the disconnection callback within the smartcard managers, and the challenge handlers could set and clear that callback as needed. Hoping this helps alleviate some of the confusion.</strike> see below comment

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/YubiKitNfcSmartcardCertBasedAuthManager.java
melissaahn: I left it up to the developer to make sure that the user has disconnected their smartcard beforehand via the disconnect() method. I don't think anything adverse happens if something is still connected when we disable discovery, and adding this check in would mean we would need to include some sort of logic as a parameter to show the removal prompt dialog, because we need to tell the user to disconnect. 
However, I'm willing to add that logic back in, if I end up making the change mentioned earlier above regarding disconnect().

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/TestNfcSmartcardCertBasedAuthManager.java
melissaahn: We do nothing. The disconnection callback doesn't necessarily need to be set. For example, discovery for USB smartcards is always running while the auth WebView is showing, so a user could potentially plug in a smartcard, then unplug the smartcard, all without actually starting the CBA flow, and that would be an instance where we don't need to do anything upon disconnection.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/UserChoiceDialog.java
melissaahn: I ended up renaming this method to `onUnexpectedUnplug()`. Hopefully this helps some of the confusion.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: Coming back to this comment... I did end up adding a disconnectCallback object as a parameter within disconnect(), and the class variable mDisconnectCallback no longer exists for the NFC manager.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandler.java
fadidurah: I think you can keep the existing method names, just adding comment would be helpful

melissaahn: See lines 95-97, 115-116

fadidurah (approved):

melissaahn:
I want to note that the removal prompt dialog is no longer being dismissed automatically in the NFC flow because of an issue I am investigating, where certain Android devices are getting blocked when the removal method is being called. 
A dismiss button was added to the dialog so that the user does not get blocked here. I'm going to continue to investigate this issue with YubiCo so that the remove() YubiKit method can be brought back in upon next release.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandlerTest.java
```diff
@@ -76,7 +76,7 @@ public void testCancelAtPinDialog() {
         setAndProcessChallengeHandler(getMockCertList());
         checkIfCorrectDialogIsShowing(TestDialog.cert_picker);
         goToPinDialog();
-        final ICancelCbaCallback callback = mDialogHolder.getCancelCbaCallback();
+        final ICancelCbaCallback callback =  mDialogHolder.getCancelCbaCallback();
```
p3dr0rv: nit: extra space
here and below

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/ISendResultCallback.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Should send a form of a result when invoked.
+ */
+public interface ISendResultCallback {
+
+    /**
+     * Runs logic that sends a form of a result.
+     */
+    void onResultReady();
```
shahzaibj, line 33: Should this accept an input result? What result does it send and where?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/DialogHolder.java
```diff
@@ -192,12 +209,25 @@ public synchronized boolean isDialogShowing() {
...
+     * Informs if a {@link SmartcardRemovalPromptDialog} is currently showing.
+     *
+     * @return True if the current dialog showing is an instance of SmartcardRemovalPromptDialog. False otherwise.
+     */
+    @Override
+    public synchronized boolean isSmartcardRemovalPromptDialogShowing() {
+        if (mCurrentDialog != null) {
+            return mCurrentDialog instanceof SmartcardRemovalPromptDialog;
+        }
+        return false;
```
shahzaibj: Instance Of checks will return false if null
```suggestion
        return mCurrentDialog instanceof SmartcardRemovalPromptDialog;
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -51,14 +53,42 @@ public NfcSmartcardCertBasedAuthChallengeHandler(@NonNull final Activity activit
...
+     * When a connection is no longer actively being used, the dialog flow should pause
+     * so the user can remove their smartcard before flow can continue.
+     * @param nextInteractionCallback the next logic to be run after a user removes their smartcard.
      */
     @Override
-    protected void onPausedSmartcardDiscovery() {
-        //Helps prevent unnecessary callback trigger. Nfc discovery should only be active when
-        // the user is prompted to tap.
-        mCbaManager.stopDiscovery(mActivity);
+    protected void prepForNextUserInteraction(@Nullable final IDisconnectionCallback nextInteractionCallback) {
```
shahzaibj: Should the callback rather be a NonNull?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
```diff
@@ -89,11 +119,15 @@ public void onCancel() {
...
-                            indicateGeneralException(methodTag, new Exception("Device connected via NFC is different from initially connected device."));
-                            request.cancel();
                             clearPin(pin);
-                            onPausedSmartcardDiscovery();
+                            request.cancel();
+                            prepForNextUserInteraction(new IDisconnectionCallback() {
+                                @Override
+                                public void onClosedConnection() {
+                                    //In a future version, an error dialog with a custom message could be shown here instead of a general error.
+                                    indicateGeneralException(methodTag, new Exception("Device connected via NFC is different from initially connected device."));
```
shahzaibj: What does indicate do? Do we throw these at some point?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/SmartcardNfcLoadingDialog.java
```diff
@@ -68,10 +68,11 @@ public void run() {
...
     /**
-     * Should dismiss dialog and call the appropriate methods to help cancel the CBA flow.
+     * Called when smartcard is unexpectedly disconnected via USB from device.
+     * Used to run any cancellation logic needed (without the cancel button needing to be pressed).
      */
     @Override
-    void onCancelCba() {
-        //No options to cancel here.
+    void onUnexpectedUnplug() {
+        //This method is for NFC, so it will never be called from here.
```
shahzaibj, line 76: Should this method not be here? Not sure I understand the comment here

shahzaibj: Same for the similar comments below

shahzaibj (approved):
Left some comments

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/SmartcardNfcLoadingDialog.java
melissaahn: The onUnexpectedUnplug() method is a member of the SmartcardDialog abstract class, so it needs to be implemented in all of the children. This dialog and a few others are meant to be shown in the NFC flow only, so there shouldn't be any scenario where this method is being called from this particular dialog.
I would have created an abstract subclass meant for USB dialogs and put this onUnexepectedUnplug() method there, but there aren't any USB specific dialogs (they're shared between USB and NFC). 
Some sort of refactoring for this is planned for a future PR.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: Yep, there's no reason for it to be Nullable. I think it was residual from a previous implementation. I'll change it to NonNull.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/ISendResultCallback.java
melissaahn: This callback is a part of this portion of my description: 
Made changes to WebViewAuthorizationFragment and AzureActiveDirectoryWebViewClient to set a callback that sends the auth result and finishes (in order to delay sending results until the user removes their smartcard). This callback will be invoked immediately if the cba challenge handler is null, and otherwise will be invoked based on the status of the challenge handler. A callback is needed here in order to make sure the user removes their smartcard before our library's activity destroys.

I needed a way to delay the sendResult() and finish() method calls within WebViewAuthorizationFragment. [This line](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1966/files#diff-82c6ecf5054ca1eddf5258b0521e7162c60d136c8b1d7dc1d0003658a74ee852R300) is where the callback is first created, and it gets passed down to the AzureActiveDirectoryWebViewClient instance, which passes it to the challenge handler for CBA.

melissaahn: (Any suggestions for method renames, preferably not tied to CBA, are appreciated)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/NfcSmartcardCertBasedAuthChallengeHandler.java
melissaahn: indicateGeneralException is a helper method that logs, shows general error dialog, and emits telemetry for when an unexpected exception is thrown.
This method is meant to be used when an exception that I don't have a ton of information about is thrown from YubiKit. My plan was to create more specific error dialogs once the telemetry from customers starts coming in.
For this instance though, I think I should get rid of the call here and instead manually log send telemetry etc.Will update.

PR #1968, opened 2023-02-22
Move JWT classes from broker to common

mohitc1:
Move JWT classes from broker to common.
This is to help using these functionalities in common. 

I intend to use that in Prtv3 implementation to add new OAuth2Strategy

Related broker PR https://github.com/AzureAD/ad-accounts-for-android/pull/2166

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/jwt/AbstractJwtRequest.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/jwt/AbstractJwtRequest.java
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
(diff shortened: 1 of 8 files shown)


rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/jwt/AbstractJwtRequest.java
```diff
@@ -0,0 +1,75 @@
...
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+@SuppressFBWarnings({"URF_UNREAD_FIELD", "URF_UNREAD_FIELD"})
+public abstract class AbstractJwtRequest {
+
+    @Setter
+    @Accessors(prefix = "m")
+    @SerializedName("ctx")
+    @SuppressFBWarnings(value = "URF_UNREAD_FIELD", justification = "This is output through serialization")
```
shahzaibj, line 38: You are probably getting these because you don't have a getter. Do you really not need a getter on these at all?

shahzaibj (approved):
Looks like it is just moving classes, but it would be good to add javadoc where missing

common4j/src/main/com/microsoft/identity/common/java/jwt/AbstractJwtRequest.java
mohitc1: It has been in use only for serialization, I guess that's why no getters. I will probably add them (separate change) to aid in UT.

PR #1972, opened 2023-02-24
ICancelCbaCallback

melissaahn:
## Summary
The main purpose of this PR is to consolidate the local `cancelCallback` interfaces attached to each dialog into a universal `ICancelCbaCallback`. Most of the changes here pertain to parameter types of methods that are within `SmartcardDialog` classes or create `SmartcardDialog`s. 
A method `getGeneralCancelCbaCallback` was created within the abstract smartcard challenge handler to return a callback that will dismiss the current dialog, finish the current telemetry span, and cancel the request. 
No new features were added to this PR, but some small fixes regarding stopping discovery after a user cancels out are added, hence the changelog.
Note that this PR overlaps with [the current CBA feature work PR](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1966). I created this PR to help reduce the size of the feature work PR.

## Related PRs
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1971
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1966

PR #1975, opened 2023-02-28
IDismissCallback Refactor

melissaahn:
## Summary
Similar to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1972#pullrequestreview-1316779182 , this PR consolidates the local `dismissCallback` interfaces attached to two types of `SmartcardDialog`s into `IDismissCallback` and updates the appropriate method parameters to use the new type.
This PR also cleans up some of the smartcard CBA test classes to use one instance of `IDismissCallback` and `ICancelCallback` (since no two dialogs will be shown at the same time).
All of the unit tests (order of operations untouched) passed, and I manually tested to confirm that the smartcard CBA flow still functioned, especially in the error case.

## Related PRs
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1972#pullrequestreview-1316779182
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1966

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
@@ -30,5 +30,5 @@ public interface IDismissCallback {
     /**
      * Action that prompts dismissal.
      */
-    void onAction();
+    void onDismiss();
 }
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Callback interface for when dialog is to be dismissed
+ * (usually by positive button click,back button, or by an expected user interaction).
+ */
+public interface IDismissCallback {
+    /**
+     * Action that prompts dismissal.
+     */
+    void onAction();
```
shahzaibj: onDismiss?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/IDismissCallback.java
fadidurah: +1

melissaahn: Does it still make sense to call the method `onDismiss` if the dismissal occurs in the method itself?

melissaahn: This is something I've been pondering about for a bit

fadidurah (approved):

PR #1977, opened 2023-03-02
Create a custom safe Global Open Telemetry that doesn't crash

shahzaibj:
A custom safe Open Telemetry Instance holder that doesn't crash on the call to "get" Open Telemetry. The motivation behind this is that all the Broker code has state that is per-process, whereas other libraries such as Open Telemetry can carry state across processes. The default Global Open Telemetry only allows initializing Open Telemetry once and subsequent calls to initialize it will throw an error. This is problematic for code such as ours that runs in multiple processes because we end up initializing OTel twice (once in app process and once in auth process).

For more details, see Javadoc

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OpenTelemetryHolder.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OpenTelemetryHolder.java
@@ -24,6 +24,8 @@ package com.microsoft.identity.common.java.opentelemetry;
 
 import io.opentelemetry.api.NoopOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.trace.Tracer;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.Setter;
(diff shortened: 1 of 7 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OpenTelemetryHolder.java
```diff
@@ -0,0 +1,84 @@
...
+ * initialize it will throw an error. This is problematic for code such as ours that runs in
+ * multiple processes because we end up initializing OTel twice (once in app process and once in
+ * auth process).
+ * <p>
+ * Open Telemetry itself recommends not using {@link io.opentelemetry.api.GlobalOpenTelemetry} and
+ * with this change we are now moving away from that model.
+ * TODO: We should make this even better by putting an instance of {@link OpenTelemetry} on the
+ * {@link com.microsoft.identity.common.java.interfaces.IPlatformComponents} and completely move
+ * away from static state, however, that requires us to then pass around either the components or
+ * the Open Telemetry instances in all places where we want to create Spans and that change is
```
melissaahn, line 50: For my understanding, is what's described here in the TODO similar to what is being done in CBA currently? (Passing a TelemetryHelper around as a parameter instead of using the static context)

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OpenTelemetryHolder.java
```diff
@@ -0,0 +1,84 @@
...
+     * See {@link io.opentelemetry.api.GlobalOpenTelemetry#getTracer(String)}.
+     */
+    public static Tracer getTracer(final String instrumentationScopeName) {
+        return sOpenTelemetry.getTracerProvider().get(instrumentationScopeName);
+    }
+
+    /**
+     * See {@link io.opentelemetry.api.GlobalOpenTelemetry#getMeter(String)}.
+     */
+    public static Meter getMeter(String instrumentationScopeName) {
```
melissaahn, line 76: final?

melissaahn (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OpenTelemetryHolder.java
shahzaibj: I think CBA just passes the Span around, here we are talking about passing Open Telemetry instance itself.

rpdome (approved):

PR #1988, opened 2023-03-14
Getting rid of account manager strategy in MSAL/OneAuth

rpdome:
Remove Account Manager strategy, both from MSAL/OneAuth, and from Broker side)
      - Nothing should break, as we never ship AccountManager as a standalone IPC mechanism. BoundService will always be there in any circumstance.
      - This code should've been removed long time ago, after ContentProvider is introduced.

this work is to prepare our stack for the upcoming LTW work.


shahzaibj (approved):

fadidurah (approved):
RIP account manager

PR #1992, opened 2023-03-21
CIAM Authority and Split Domain

fadidurah:
What:
This PR includes Common-side changes to support a new CIAM Authority type and Split Domain

Changes:
- Created `CIAMAuthority` class to support new CIAM type in JSON configuration, updated serializer to reflect this and updated `Authority` to support different CIAM authority formats
- Added check for `CIAMAuthority` in `SilentTokenCommandParameters` to skip check for account environment
- Added method in `MicrosoftStsOAuth2Strategy` to pass in Open ID Configuration, which is used to fetch the CIAM authorization end point (the recommended approach from eSTS team is to fetch the endpoints from open id configuration as CIAM tenants seem to not match authority with their endpoints (see appendix C in Split Domain Document below, not sure if this is changing in the future, so i put it under a flag))
- Added testing in `AuthorityDeserializerTest`
- Added `CIAMTestAuthorty` to use in network tests along with constants for test CIAM JSON Config files 

Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1783

Why:
Simply explained, CIAM is a new offering from Microsoft to provide a combination of B2C and AAD features without sacrificing aspects of or changing the AAD developer experience.

For a much more detailed breakdown, here are some relevant resources:
- [CIAM Introductory Presentation](https://microsoft.sharepoint.com/:p:/t/aad/devex/EWBXiSmdhMNKoFhZ0LiYDpcBySPB7zfl79TI9EkI8a3ZKQ?e=kk6tHW)
- [CIAM Split Domain Doc](https://microsoft-my.sharepoint-df.com/:w:/r/personal/jmprieur_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7BF07A764A-DA4A-4CE9-A2F2-D6671C174C47%7D&file=Impact%20of%20CIAM%20split%20domain%20on%20DevEx.docx&_DSL=1&action=default&mobileredirect=true)
- [CIAM Private Preview Repo](https://github.com/microsoft/entra-previews)

How:
My implementation focused on building off of existing AAD flow infrastructure to make sure we stay as close as possible to that developer experience.

Testing:
This PR includes 3 new classes in our network testing infrastructure to support three format of CIAM authorities passed in the JSON configuration file.
- https://tenantDomain.ciamlogin.com
- https://tenantDomain.ciamlogin.com/GUID
- https://tenantDomain.ciamlogin.com/tenantDomain.onmicrosoft.com

Also did manual testing for these three authority formats using MsalTestApp with the following test steps:
- AcquireToken
- AcquireTokenSilent
- AcquireTokenSilent after forwarding device time by 1 day


common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -134,7 +135,12 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
 
         final List<String> pathSegments = authorityCommonUriBuilder.getPathSegments();
 
-        if (pathSegments.size() == 0) {
+        // Adding check in case we have a trailing "/" at the end of the authority
+        if (pathSegments.size() == 0 || (pathSegments.size() == 1 && pathSegments.get(0).equals(""))) {
```
fadidurah, line 139: I'm not sure if this is really needed or preferred, even. Basically checking if they passed `https://tenant.ciamlogin.com/` instead of just `https://tenant.ciamlogin.com`. Maybe it's ok to simply expect devs to pass `https://tenant.ciamlogin.com`, and whatever behavior occurs with `https://tenant.ciamlogin.com/` is out of scope. Thoughts?

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+            return strategy;
+    }
+
+    /**
+     * This method takes a CIAM authority string of format "tenant.ciamlogin.com" or "https://tenant.ciamlogin.com"
+     * @param authorityNoPath authority to be transformed
+     * @return full CIAM authority with path
+     */
+    public static String getFullAuthorityUrlFromAuthorityWithoutPath(@NonNull String authorityNoPath){
```
SammyO, line 98: Does this account for the temporary CIAM domain `https://ciamlogin.com/tenant` (temporary for Build 2023)? Not sure if that's in scope anymore, but I see it mentioned in some documentation (e.g. https://microsoft-my.sharepoint-df.com/:w:/r/personal/jmprieur_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7BF07A764A-DA4A-4CE9-A2F2-D6671C174C47%7D&file=Impact%20of%20CIAM%20split%20domain%20on%20DevEx.docx&_DSL=1&action=default&mobileredirect=true&cid=2cedb99e-423c-463f-98ea-0f424b9ad51d)

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -146,6 +152,8 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
 
             if (B2C.equalsIgnoreCase(authorityTypeStr)) {
                 authority = new AzureActiveDirectoryB2CAuthority(authorityUrl);
+            } if (CIAM.equalsIgnoreCase(authorityTypeStr)) {
```
SammyO: Can't comment there, but should `authorityType` `CIAM` check also be added to else block (L160-L186)?

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: I did not test with that formatting. I haven't seen it recently in the updated documentation so i did not take it into account. Will confirm with Jean-Marc

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -134,7 +135,12 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
 
         final List<String> pathSegments = authorityCommonUriBuilder.getPathSegments();
 
-        if (pathSegments.size() == 0) {
+        // Adding check in case we have a trailing "/" at the end of the authority
+        if (pathSegments.size() == 0 || (pathSegments.size() == 1 && pathSegments.get(0).equals(""))) {
+            if (authorityUrl.contains("ciamlogin.com")){
```
shahzaibj: Let's define this in a constant somewhere

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+import com.microsoft.identity.common.java.logging.Logger;
+import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectorySlice;
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Configuration;
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Strategy;
+import com.microsoft.identity.common.java.providers.oauth2.OAuth2Strategy;
+import com.microsoft.identity.common.java.providers.oauth2.OAuth2StrategyParameters;
+
+import lombok.NonNull;
+
+public class CIAMAuthority extends Authority {
```
shahzaibj, line 41: Javadoc

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+    private static transient final String TAG = CIAMAuthority.class.getSimpleName();
+
+    // Adding this flag to indicate whether or not we should load the OpenId Configuration as part of handling
+    // CIAM flows. This is currently relevant for fetching the authorization endpoint from OpenId rather than
+    // using the authority itself + adding the default authorization endpoint.
+    private final boolean CIAM_USE_OPENID_CONFIGURATION = true;
```
shahzaibj, line 47: This sounds like the right approach. It is usually best to load them from Open Id.

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public CIAMAuthority(String authorityUrl) {
```
shahzaibj: final and nullability annotations on param

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+        mAuthorityTypeString = "CIAM";
+        mAuthorityUrlString = authorityUrl;
+    }
+
+    protected MicrosoftStsOAuth2Configuration createOAuth2Configuration() {
```
shahzaibj: Overide?

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+        final String methodName = ":createOAuth2Configuration";
+        Logger.verbose(
+                TAG + methodName,
+                "Creating OAuth2Configuration"
+        );
+        MicrosoftStsOAuth2Configuration config = new MicrosoftStsOAuth2Configuration();
```
shahzaibj:
```suggestion
        final MicrosoftStsOAuth2Configuration config = new MicrosoftStsOAuth2Configuration();
```

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+            slice.setDataCenter(mSlice.getDataCenter());
+            config.setSlice(slice);
+        }
+
+        return config;
+    }
+
+    @Override
+    public OAuth2Strategy createOAuth2Strategy(OAuth2StrategyParameters parameters) throws ClientException {
+            MicrosoftStsOAuth2Configuration config = createOAuth2Configuration();
```
shahzaibj:
```suggestion
            final MicrosoftStsOAuth2Configuration config = createOAuth2Configuration();
```

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,96 @@
...
+            final MicrosoftStsOAuth2Strategy strategy = new MicrosoftStsOAuth2Strategy(config, parameters);
+            if (CIAM_USE_OPENID_CONFIGURATION) {
+                strategy.loadOpenIdProviderConfiguration();
+            }
+            return strategy;
+    }
+
+    /**
+     * This method takes a CIAM authority string of format "tenant.ciamlogin.com" or "https://tenant.ciamlogin.com"
```
shahzaibj, line 93: This method takes a CIAM Authority and then what does it do?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -809,4 +811,21 @@ private boolean cachedAccessTokenKidMatchesKeystoreKid(@Nullable final String at
 
         return deviceKid.equals(atKid);
     }
+
+    /**
+     * Using this method to load the {@link OpenIdProviderConfiguration}
+     */
+    public void loadOpenIdProviderConfiguration() {
```
shahzaibj: Wouldn't it be better for this method to actually return the configuration? and then the caller uses the configuration

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -809,4 +811,21 @@ private boolean cachedAccessTokenKidMatchesKeystoreKid(@Nullable final String at
...
+        try {
+            final OpenIdProviderConfigurationClient client =
+                    new OpenIdProviderConfigurationClient();
+            mOpenIdProviderConfiguration = client.loadOpenIdProviderConfigurationFromAuthority(mConfig.getAuthorityUrl().toString());
+        } catch (ServiceException e) {
+            Logger.error(
```
shahzaibj, line 845: Do w need to throw in this case? What is the expected behavior for an Acquire Token call when this exception occurs?

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
shahzaibj: Why do we need to do this check? Shouldn't CIAM authorities by always known from configuration?

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: We do not currently do this for AAD, we can create some work in the future to update AAD to also use openid configuration?

shahzaibj: Hmm. Interesting. Perhaps there are reasons for not doing it historically. Probably a good idea to create a work item to revisit this at some point and see if we can leverage Open Id Configuration in all places.

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: This method isn't defined in `Authority.java`

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
fadidurah: In this use case, the caller (`CIAMAuthority`) doesn't need to access the configuration. I wanted to avoid the confusion of devs thinking they would need to utilize the Open ID Configuration themselves by simply not returning it, as it used internally by `MicrosoftStsAuthorizationRequest`

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
fadidurah: I think it's best to fallback to the default way we did it where we use the default authorization end point, as that may still work in some cases (although probably would still fail, if the authority url does not have an Open ID Configuration). I could go either way. Since it's probably going to fail anyway, maybe we can just throw a client exception. Thoughts?

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
fadidurah: We would have to move the path segment check to a point after the `authorityIsKnownFromConfiguration` call. I think this is fine, but didn't want to mess with the logic of this method too much as it's used in a majority of MSAL flows, and didn't want to introduce any breaking change so close to release. I think it's safer to keep it as is for now, and once we've validated the current changes, we can go back and re-order this method. Also I'd probably still keep this second condition incase the shortened CIAM Authority is not known to MSAL for some reason.

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: Sounds good!

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
fadidurah: @shahzaibj thoughts? Is it common practice/commonplace for Authorities to be input with a trailing "/"?

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
fadidurah: Added a check

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: Confirmed with Jean-marc and Ray, `https://ciamlogin.com/tenant` is not supported, so should be good on this front

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
shahzaibj: Okay, is there a reason for this to be `protected`? Should it be either `private` or perhaps `public`?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
shahzaibj: Okay, but why is the caller meant to invoke this method? If caller needs to invoke this i.e. `CIAMAuthority` then perhaps the method should move to `CIAMAuthority` and become a private method there?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
shahzaibj: Remind me, the default authorization endpoint here refers to creating the endpoint ourselves?

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
shahzaibj: I think this is a question about how is the feature supposed to work? Should developers be allowed to get a token from a CIAM Authority even if the authority is not declared in msal json configuration?

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: private is best, i think

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -170,12 +178,21 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
...
-                    authority = createAadAuthority(authorityCommonUriBuilder, pathSegments);
-                    break;
+                    if (authorityUrl.contains(CIAMAuthority.CIAM_LOGIN_URL_SEGMENT)) {
+                        Logger.verbose(
+                                TAG + methodName,
+                                "Authority type is CIAM"
+                        );
+                        authority = new CIAMAuthority(authorityUrl);
+                        break;
+                    } else {
```
mohitc1: Consider moving this out of switch case

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public OAuth2Strategy createOAuth2Strategy(OAuth2StrategyParameters parameters) throws ClientException {
+            final MicrosoftStsOAuth2Configuration config = createOAuth2Configuration();
+            final MicrosoftStsOAuth2Strategy strategy = new MicrosoftStsOAuth2Strategy(config, parameters);
+            if (CIAM_USE_OPENID_CONFIGURATION) {
+                strategy.loadOpenIdProviderConfiguration();
+            }
+            return strategy;
```
mohitc1: indent

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsAuthorizationRequest.java
```diff
@@ -148,6 +149,13 @@ public static final class Prompt {
      */
     public static final String HIDE_SWITCH_USER_QUERY_PARAMETER = "hsu";
 
+    /**
+     * Add this field to store the openIdConfiguration passed as part of the builder.
+     * This will be used to fetch the atuorization endpoint from openID Configutration if it was
```
mohitc1: nit: spelling

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
fadidurah: I tried to set it up in a way that any class that uses `MicrosoftStsOauth2Strategy` can decide whether or not to load this configuration. For example, if you want to change AAD Authorities to use OpenID for the authorization endpoint, they would just have to call this method in `MicrosoftStsOauth2Strategy`, without worrying about supplying the authority url as that will be provided from `MicrosoftStsOauth2Strategy`.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
fadidurah: yes,
`return UrlUtil.appendPathToURL(this.getAuthority(), AUTHORIZATION_ENDPOINT).toString();`
in `MicrosoftStsAuthorizationRequest`

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
jmprieur: let's be flexible on the trailing slash. It does not cost much to be.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
fadidurah: Pass flag to load open id configuration through `OAuth2StrategyParameters`

shahzaibj (approved):
Thanks Fadi!

common4j/src/main/com/microsoft/identity/common/java/authorities/CIAMAuthority.java
fadidurah: Not much bandwidth this quarter, lets rain check this

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
fadidurah: Done, this format is supported

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
fadidurah: Ended up getting rid of the switch statement

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -825,4 +830,43 @@ private boolean cachedAccessTokenKidMatchesKeystoreKid(@Nullable final String at
...
+        }
+    }
+
+    /**
+     * Using this method to load the {@link OpenIdProviderConfiguration} with extra parameters
+     * This will cause the strategy to fetch the authorization endpoint from OpenId Configuration rather
+     * than generating one with the default authorization endpoint
+     */
+    @SuppressFBWarnings
+    private void loadOpenIdProviderConfiguration(final String extraParams) {
```
mohitc1: >final String extraParams [](http://example.com/codeflow?start=49&length=24)

@ NonNull

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2StrategyParameters.java
```diff
@@ -42,4 +43,7 @@ public class OAuth2StrategyParameters {
 
     @Nullable
     private final transient AbstractAuthenticationScheme mAuthenticationScheme;
+
+    @Setter
+    private transient boolean mUsingOpenIdConfiguration;
```
mohitc1, line 50: This probably better belongs to MicrosoftStsOAuth2Configuration class and a public method isOpenIdConfigUsed() or getOpenIdConfig()
With that MicrosoftStsOAuth2Configuration .getAuthorizationEndpoint() method can be modified to consider this openid configuration before returning result (That method currently does not have any callers, if it had you would have had to make modifications there as well.).

The config object can also then be used to setup the authorization request builder in strategy.

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2StrategyParameters.java
fadidurah: I'll add a TODO to consider moving this to MicrosoftStsOAuth2Configuration. I think it would be better to make this change and update the endpoint methods in the same PR, but I'm going to save it for a later PR as this one is decently dense as is and is mainly focused on getting everything wired up.

fadidurah: On  a side note, why does the configuration class even have a method for the authorization endpoint, shouldn't that come from the strategy class?

PR #1993, opened 2023-03-21
Add method to clear receiver concurrentHashMap in LocalBroadcaster

sansrivastav:
This is continuation of changes done in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1895.
Even though we are now resetting the broadcast service between 2 tests, the cancel authorization request broadcast is triggered while the executor service is being reset causing a concurrent.RejectedExecutionException which kills the MSAL test app.
With current change a method to clear receiver concurrent hash map is introduced and this should be called just before calling resetBroadcast().
AB#2478513

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
@@ -68,18 +68,18 @@ public enum LocalBroadcaster {
 
     public void broadcast(@NonNull final String alias, @NonNull final PropertyBag propertyBag) {
         final String methodName = ":broadcast";
-//        sBroadcastExecutor.execute(new Runnable() {
-//            public void run() {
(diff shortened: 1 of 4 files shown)


nickbopp (approved):

common4j/src/main/com/microsoft/identity/common/java/util/ported/LocalBroadcaster.java
```diff
@@ -82,6 +82,13 @@ public void run() {
         });
     }
 
+    /**
+     * Clears the receivers associated with this instance.
+     */
+    public void clearReceivers() {
```
iamgusain, line 88: Can you please also add tests for this and the previously added methods.
Thanks

sansrivastav: done

iamgusain (approved):

PR #1996, opened 2023-03-24
Fix for ADAL W/Broker Silent call 

p3dr0rv:
related to https://github.com/AzureAD/ad-accounts-for-android/pull/2200

Description on https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=2483343

## Tested 

Install BrokerHost 
Install Adal Test App
Call AcquireToken (this should pass)
Call AcquireTokenSilent(this should pass too)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [PATCH] Send AT on KEY_AUTHTOKEN for ADAL Acquire token silently with Broker (#1996)
 - [PATCH] Expose Cached Credential Service request ID in tokenResponse (#1991)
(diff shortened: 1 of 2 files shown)


fadidurah (approved):
I also validated this and looks like it fixes the issue, thanks for picking this up Pedro! Does this require a new version being released in ADAL to make sure apps using adal are compatible with the new changes going into broker? I guess it depends on which version of common would be used (the one specified in broker-hosting app, or the one specified in the app making the ADAL call)

somalaya (approved):

PR #1998, opened 2023-03-28
Allow MsalCpp to specify an InMemoryCache implementation of SharedPreferencesAccountCredentialCache

nickbopp:
This is a reimplementation of #1929, where the calling app needs to choose to configure its token cache for this new in-memory caching behavior.  The goal here is that MSAL.CPP will specify this for now, and other usages (broker, MSAL) could be brought over in the future in a slower or optional rollout. 

After consulting with Amit, and because the implementations here are so different, it wasn't possible to reasonably maintain a single class with forked implementation, and so instead the class itself has been forked. Similarly the tests have been forked.  I think it would be possible to instead turn those into parameterized tests, and pull some of the setup implementation as a preable to all test methods that uses that parameter to choose whether to run against the memory-backed version or not.  I think that's still a fairly messy change, but I'm open to that option if reviewers prefer it (or have a better idea).

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
@@ -85,9 +85,7 @@ public class MsalCppOAuth2TokenCache
     }
 
     /**
-     * Factory method for creating an instance of MsalCppOAuth2TokenCache
-     * <p>
-     * NOTE: Currently this is configured for AAD v2 as the only IDP
+     * Factory method for creating an instance of MsalCppOAuth2TokenCache.
      *
      * @param platformComponents The Application Context
      * @return An instance of the MsalCppOAuth2TokenCache.
@@ -99,9 +97,7 @@ public class MsalCppOAuth2TokenCache
     }
 
     /**
-     * Factory method for creating an instance of MsalCppOAuth2TokenCache
-     * <p>
-     * NOTE: Currently this is configured for AAD v2 as the only IDP
+     * Factory method for creating an instance of MsalCppOAuth2TokenCache.
      *
      * @param platformComponents The Application Context
      * @param useInMemoryCache Opt-in to caching layer that holds account and credential objects in memory
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
@@ -95,7 +95,22 @@ public class MsalCppOAuth2TokenCache
     // Suppressing unchecked warning as the return type requiring generic parameter which is not provided
     @SuppressWarnings(WarningType.unchecked_warning)
     public static MsalCppOAuth2TokenCache create(@NonNull final IPlatformComponents platformComponents) {
-        final MsalOAuth2TokenCache msalOAuth2TokenCache = MsalOAuth2TokenCache.create(platformComponents);
+        return create(platformComponents, false);
+    }
+
+    /**
+     * Factory method for creating an instance of MsalCppOAuth2TokenCache
+     * <p>
+     * NOTE: Currently this is configured for AAD v2 as the only IDP
+     *
+     * @param platformComponents The Application Context
+     * @param useInMemoryCache Opt-in to caching layer that holds account and credential objects in memory
+     * @return An instance of the MsalCppOAuth2TokenCache.
+     */
+    // Suppressing unchecked warning as the return type requiring generic parameter which is not provided
+    @SuppressWarnings(WarningType.unchecked_warning)
+    public static MsalCppOAuth2TokenCache create(@NonNull final IPlatformComponents platformComponents, boolean useInMemoryCache) {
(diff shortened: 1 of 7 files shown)


rpdome:
@nickbopp Would it be possible to inherit from SharedPreferenceFileManager and override getString(key) Instead?

(I assume that #1927 is the only delta here)

nickbopp:
> @nickbopp Would it be possible to inherit from SharedPreferenceFileManager and override getString(key) Instead?
> 
> (I assume that #1927 is the only delta here)

Apologies, I linked to the wrong one.  It's #1929 .  I updated the initial comment.  It's a pretty large change to most methods in the class.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -0,0 +1,768 @@
...
+
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public class SharedPreferencesAccountCredentialCacheWithMemoryCache extends AbstractAccountCredentialCache {
+
+    private static final String TAG = SharedPreferencesAccountCredentialCacheWithMemoryCache.class.getSimpleName();
+
+    /**
+     * The name of the SharedPreferences file on disk.
+     */
+    public static final String DEFAULT_ACCOUNT_CREDENTIAL_SHARED_PREFERENCES =
```
iamgusain: We can get these public constants from one place.

iamgusain:
I haven.t taken a look at the new implementation closely, but otherwise lgtm

iamgusain (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -95,7 +95,22 @@ private MsalCppOAuth2TokenCache(final IPlatformComponents commonComponents,
...
     @SuppressWarnings(WarningType.unchecked_warning)
     public static MsalCppOAuth2TokenCache create(@NonNull final IPlatformComponents platformComponents) {
-        final MsalOAuth2TokenCache msalOAuth2TokenCache = MsalOAuth2TokenCache.create(platformComponents);
+        return create(platformComponents, false);
+    }
+
+    /**
+     * Factory method for creating an instance of MsalCppOAuth2TokenCache
+     * <p>
+     * NOTE: Currently this is configured for AAD v2 as the only IDP
```
shahzaibj:
```suggestion
     * Factory method for creating an instance of MsalCppOAuth2TokenCache.
     * <p>
     * NOTE: Currently this is configured for AAD v2 as the only IDP.
```

shahzaibj: What exactly is specific to AAD v2 here?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -0,0 +1,768 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+import lombok.NonNull;
+
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public class SharedPreferencesAccountCredentialCacheWithMemoryCache extends AbstractAccountCredentialCache {
```
shahzaibj, line 54: javadoc

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -0,0 +1,768 @@
...
+            DEFAULT_ACCOUNT_CREDENTIAL_SHARED_PREFERENCES
+                    + ".foci-1";
+
+    /**
+     * Returns the generated filename for UID-specific caches.
+     *
+     * @param uid The uid of the current (or targeted) application.
+     * @return The uid-based cache filename.
+     */
+    public static String getBrokerUidSequesteredFilename(final int uid) {
```
shahzaibj: I learned a new word today :D

common4j/src/main/com/microsoft/identity/common/java/dto/AccountCredentialBase.java
```diff
@@ -78,6 +78,13 @@ public void mergeAdditionalFields(@NonNull final AccountCredentialBase other) {
         }
     }
 
+    @Override
+    public AccountCredentialBase clone() throws CloneNotSupportedException {
+        AccountCredentialBase other = (AccountCredentialBase) super.clone();
```
shahzaibj:
```suggestion
        final AccountCredentialBase other = (AccountCredentialBase) super.clone();
```

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
nickbopp: I agree this is unneeded, copy-pasted from the original one but I've deleted it in both.

PR #2000, opened 2023-03-31
Add CommandDispatcher methods to stop and reset silent request executor.

iamgusain:
### What 
- Adding methods in CommandDispatcher class to stop and reset the silent request executor.

### Why
- In ShareDeviceMode when signout is triggered. It is possible that there are pending silent requests in the queue.
- If this happens there is a possibility that the silent requests are completed after signout operation is complete
- This results in the silent requests populating the account information back in broker storage. resulting in the previous user data to be still available in broker even after signout
- We need to make sure that all silent requests are either complete or cancelled before we process the signout request in SDM. And also that no new silent requests are processed until signout is complete.

### How
- Added method to stop and reset the silent request executor. These methods will be called from broker signout method. (change included in [broker PR)](https://github.com/AzureAD/ad-accounts-for-android/pull/2220)

### Testing
- Added UTs to verify the new methods
- Also added a E2E test to verify this scenario ([Included in MSAL PR](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1789))

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -22,23 +22,20 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common;
 
-import android.content.Intent;
+import static com.microsoft.identity.common.java.exception.ServiceException.SERVICE_NOT_AVAILABLE;
 
 import androidx.annotation.NonNull;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
-import com.microsoft.identity.common.java.result.ILocalAuthenticationResult;
-import com.microsoft.identity.common.java.result.LocalAuthenticationResult;
+import com.microsoft.identity.common.internal.commands.RefreshOnCommand;
 import com.microsoft.identity.common.java.cache.CacheRecord;
-import com.microsoft.identity.common.java.exception.TerminalException;
 import com.microsoft.identity.common.java.cache.ICacheRecord;
 import com.microsoft.identity.common.java.commands.BaseCommand;
 import com.microsoft.identity.common.java.commands.CommandCallback;
+import com.microsoft.identity.common.java.commands.EmptyCommandCallback;
+import com.microsoft.identity.common.java.commands.ICommandResult;
 import com.microsoft.identity.common.java.commands.parameters.CommandParameters;
-import com.microsoft.identity.common.internal.commands.RefreshOnCommand;
-import com.microsoft.identity.common.java.commands.parameters.DeviceCodeFlowCommandParameters;
-import com.microsoft.identity.common.java.commands.parameters.GenerateShrCommandParameters;
 import com.microsoft.identity.common.java.commands.parameters.DeviceCodeFlowCommandParameters;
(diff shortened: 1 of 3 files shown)


common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -367,6 +366,46 @@ public String execute() {
...
+        // Stop the silent executor
+        CommandDispatcher.stopSilentRequestExecutor();
+
+        // verify that the previous command results in error
+        CommandResult result = future1.get();
+        Assert.assertEquals(ICommandResult.ResultStatus.ERROR, result.getStatus());
+
+        // try scheduling a new command
+        try {
+            CommandDispatcher.submitSilentReturningFuture(new LongRunningTestCommand(getEmptyTestParams(), new EmptyCommandCallback()));
```
shahzaibj, line 384: So we cannot run any commands after stopping the executor?

shahzaibj: Ok, it needs to be reset it seems

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -754,7 +753,36 @@ private static String statusMsg (String status){
...
-
+    /***
+     * Resets the SilentRequestsExecutor.
+     * This creates a new Executor for the silent request.
+     * This is expected to be called after global signout is performed in Shared Device mode.
+     * This should be called if previously the Executor was stopped using 'stopSilentRequestExecutor'
+     */
+    public static void resetSilentRequestExecutor() {
+        Logger.info(TAG + ":resetSilentRequestExecutor", "Resetting silent Executor");
+        sSilentExecutor = Executors.newFixedThreadPool(SILENT_REQUEST_THREAD_POOL_SIZE);
```
shahzaibj, line 785: Do we have to necessarily create a new executor? Is there a way to just restart the one we stopped earlier?

iamgusain: Unfortunately no, there is no way currently to restart the existing one

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -754,7 +753,36 @@ private static String statusMsg (String status){
...
 
-
+    /***
+     * Resets the SilentRequestsExecutor.
+     * This creates a new Executor for the silent request.
+     * This is expected to be called after global signout is performed in Shared Device mode.
+     * This should be called if previously the Executor was stopped using 'stopSilentRequestExecutor'
+     */
+    public static void resetSilentRequestExecutor() {
+        Logger.info(TAG + ":resetSilentRequestExecutor", "Resetting silent Executor");
```
fadidurah, line 784: What if the executor was reset without calling stopSilentRequestExecutor prior?

fadidurah (approved):

somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
iamgusain: It will create a new pool to start executing silent requests.

PR #2001, opened 2023-03-31
Add more log statements to the getValidActiveBrokerPackageName and add more details on the client exceptions 

p3dr0rv:
ClientExceptions require an error code and error message, so we are adding the code, and improving the message, to identify root cause of MWPJ failing on initialization.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
@@ -24,6 +24,7 @@ package com.microsoft.identity.common.internal.broker;
 
 import static com.microsoft.identity.common.java.AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE;
 import static com.microsoft.identity.common.java.exception.ClientException.ACCOUNT_MANAGER_FAILED;
+import static com.microsoft.identity.common.java.exception.ClientException.BROKER_VERIFICATION_FAILED_ERROR;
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -247,22 +249,27 @@ public String getCurrentActiveBrokerPackageName() {
...
-            for (final AuthenticatorDescription authenticator : authenticators) {
-                if (BROKER_ACCOUNT_TYPE.equals(authenticator.type)) {
-                    verifySignatureAndThrow(authenticator.packageName);
-                    return authenticator.packageName;
-                }
-            }
+            authenticators = AccountManager.get(mContext).getAuthenticatorTypes();
         } catch (final Exception exception) {
-            final ClientException clientException = new ClientException(NOT_VALID_BROKER_FOUND, exception.getMessage());
+            final ClientException clientException = new ClientException(ACCOUNT_MANAGER_FAILED, exception.getMessage());
```
shahzaibj, line 256: So this is Account Manager being completely broken?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
p3dr0rv: this is account manager just failing unexpectedly

fadidurah (approved):

PR #2002, opened 2023-04-04
LTW support work

fadidurah:
LTW support work

Supported so far:
LTW class for handling installation, updating, etc.
Custom broker installation (broker shuffling)
Added support for flag `preInstallLtw` which installs LTW first in our existing test cases that use broker.
To use, pass `-PpreInstallLtw` as a command line parameter when building either of the automation apps.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
@@ -22,45 +22,89 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.ui.automation;
 
+import com.microsoft.identity.client.ui.automation.broker.BrokerCompanyPortal;
+import com.microsoft.identity.client.ui.automation.broker.BrokerHost;
+import com.microsoft.identity.client.ui.automation.broker.BrokerLTW;
+import com.microsoft.identity.client.ui.automation.broker.BrokerMicrosoftAuthenticator;
+
 /**
- * An interface to facilitate testing a very specific broker installation order with support for
- * old and current version of the broker.
+ * An interface to facilitate testing a very specific app installation order with support for
+ * old and current version of the app.
  */
 public interface ICustomBrokerInstallationTest {
 
     /**
      * Install old/legacy BrokerHost
      */
-    void installOldBrokerHost();
+    default BrokerHost installOldBrokerHost(){
+        final BrokerHost brokerHost = new BrokerHost(BrokerHost.OLD_BROKER_HOST_APK,
+                BrokerHost.BROKER_HOST_APK);
+        brokerHost.install();
+        return brokerHost;
+    }
     /**
      * Install updated BrokerHost
      */
-    void installBrokerHost();
+    default BrokerHost installBrokerHost(){
+        final BrokerHost brokerHost = new BrokerHost();
+        brokerHost.install();
+        return brokerHost;
+    }
 
     /**
      * Install old/legacy Authenticator
      */
-    void installOldAuthenticator();
+    default BrokerMicrosoftAuthenticator installOldAuthenticator(){
+        final BrokerMicrosoftAuthenticator authenticator = new BrokerMicrosoftAuthenticator(BrokerMicrosoftAuthenticator.OLD_AUTHENTICATOR_APK,
+                BrokerMicrosoftAuthenticator.AUTHENTICATOR_APK);
+        authenticator.install();
+        return authenticator;
+    }
     /**
      * Install updated Authenticator
      */
-    void installAuthenticator();
+    default BrokerMicrosoftAuthenticator installAuthenticator(){
+        final BrokerMicrosoftAuthenticator authenticator = new BrokerMicrosoftAuthenticator();
+        authenticator.install();
+        return authenticator;
+    }
 
     /**
      * Install old/legacy Company Portal
      */
-    void installOldCompanyPortal();
+    default BrokerCompanyPortal installOldCompanyPortal(){
+        final BrokerCompanyPortal companyPortal = new BrokerCompanyPortal(BrokerCompanyPortal.OLD_COMPANY_PORTAL_APK,
+                BrokerCompanyPortal.COMPANY_PORTAL_APK);
+        companyPortal.install();
+        return companyPortal;
+    }
     /**
      * Install updated Company Portal
      */
-    void installCompanyPortal();
+    default BrokerCompanyPortal installCompanyPortal(){
+        final BrokerCompanyPortal companyPortal = new BrokerCompanyPortal();
+        companyPortal.install();
+        return companyPortal;
+    }
 
     /**
      * Install old/legacy LTW
      */
-    void installOldLtw();
+    default BrokerLTW installOldLtw(){
+        final BrokerLTW ltw = new BrokerLTW(BrokerLTW.OLD_BROKER_LTW_APK,
+                BrokerLTW.BROKER_LTW_APK);
+        ltw.install();
+        return ltw;
+    }
     /**
      * Install updated LTW
      */
-    void installLtw();
+    default BrokerLTW installLtw(){
+        final BrokerLTW ltw = new BrokerLTW();
+        ltw.install();
+        return ltw;
+    }
+
+    // TODO: LTW folks, expand this to include msaltestapp, adaltestapp, oneuathtestapp
+    
 }
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
@@ -34,7 +34,7 @@ import com.microsoft.identity.client.ui.automation.broker.BrokerMicrosoftAuthent
 public interface ICustomBrokerInstallationTest {
 
     /**
-     * Install old/legacy BrokerHost
+     * Install old/legacy BrokerHost.
      */
     default BrokerHost installOldBrokerHost(){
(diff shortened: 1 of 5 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,66 @@
...
+public interface ICustomBrokerInstallationTest {
+
+    /**
+     * Install old/legacy BrokerHost
+     */
+    void installOldBrokerHost();
+    /**
+     * Install updated BrokerHost
+     */
+    void installBrokerHost();
```
somalaya: is this going to replace the update() method in App.java? 
calling installOldBrokerHost() and installBrokerHost() is going to update the broker host right?

fadidurah: Followed up online, Sowmya to investigate if the -r flad in updte() is significant.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerLTW.java
```diff
@@ -0,0 +1,108 @@
...
+
+import com.microsoft.identity.client.ui.automation.constants.DeviceAdmin;
+import com.microsoft.identity.client.ui.automation.installer.IAppInstaller;
+
+/**
+ * Handles app interaction with Broker-hosting app LTW
+ */
+public class BrokerLTW extends AbstractTestBroker {
+    private static final String TAG = BrokerLTW.class.getSimpleName();
+    public final static String BROKER_LTW_APP_PACKAGE_NAME = "LTW PACKAGE NAME HERE";
```
shahzaibj: We don't know this today?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
shahzaibj: If these methods here i.e. `installBrokerHost` , `installAuthenticator` etc. are updating the broker then I suggest we rename these accordingly.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerLTW.java
fadidurah: Nope

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
fadidurah: They don't update it per se. More so they install the latest apk, regardless of whether or not an existing one exists. The question is whether or not this is actually the same as updating the app.

somalaya: @fadidurah adb install -r test.apk reinstalls the app, keeping its data.
So the question here is, do we want to install OldBrokerHost and then the new one by keeping the data? We need to confirm this with Dome on whether we want to retain the data. If yes, we should ideally update the app with -r instead of just installing new one on top of the old one.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerLTW.java
somalaya: it is com.microsoft.appmanager

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,66 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.ui.automation;
+
+/**
+ * An interface to facilitate testing a very specific broker installation order with support for
+ * old and current version of the broker.
+ */
+public interface ICustomBrokerInstallationTest {
```
somalaya, line 34: Should we restrict this class for Broker app installations? Or can we add methods to installOld and new versions of MsalTestApp, OneAuthTestApp as well?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
fadidurah: In that case, I think that this approach will be ideal, because we will have access to both a new installation and an update method. Perhaps we can outline which to use as part of the test case itself?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/InstallBrokerTestRule.java
```diff
@@ -52,6 +54,13 @@ public Statement apply(final Statement base, final Description description) {
             public void evaluate() throws Throwable {
                 Logger.i(TAG, "Applying rule....");
                 Logger.i(TAG, "Installing broker: " + ((App) broker).getAppName());
+
+                if (BuildConfig.PRE_INSTALL_LTW) {
+                    final BrokerLTW brokerLTW = new BrokerLTW();
```
somalaya, line 64: can we move this RulesHelper.java like this? https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/a2b6d6423079de31f570c9db27bcfd8e0c8dd75f/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/RulesHelper.java#L88

somalaya: Also, do we know if PREFER_PRE_INSTALLED_APKS is being set currently in the pipelines?
Looks like we have to add BrokerLTW in the preinstalled apks rule as well

somalaya: Can you also update the PR description with instructions on how we can set this flag? for example : -PpreInstallLTW=true in the project to run locally.

fadidurah: LTW is covered already in the remove pre installed brokers rule
I'll mention the flag in the description

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
fadidurah: No, we can expand it to cover all of the classes, that should be a problem. I'm going to leave that for when the ui automation classes for MsalTestApp and OneAuthTestApp etc. are added by LTW folks. They can rename this class and add the methods to install old and new.

fadidurah: ~~Scratch that, I'll go ahead and rename it and add a couple TODOs~~

I'm not going to rename it since you already made the document for them, I'll just the TODO

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/ICustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.microsoft.identity.client.ui.automation.broker.BrokerMicrosoftAuthenticator;
+
+/**
+ * An interface to facilitate testing a very specific app installation order with support for
+ * old and current version of the app.
+ */
+public interface ICustomBrokerInstallationTest {
+
+    /**
+     * Install old/legacy BrokerHost
```
shahzaibj:
```suggestion
     * Install old/legacy BrokerHost.
```

Let's add periods at the end of sentences in javadoc; here and everywhere

shahzaibj (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/InstallBrokerTestRule.java
```diff
@@ -52,6 +54,13 @@ public Statement apply(final Statement base, final Description description) {
             public void evaluate() throws Throwable {
                 Logger.i(TAG, "Applying rule....");
                 Logger.i(TAG, "Installing broker: " + ((App) broker).getAppName());
+
+                if (BuildConfig.PRE_INSTALL_LTW) {
```
rpdome, line 63: could you please elaborate more/ add comments on what this is needed for?

If this is to cover "install LTW first and everything else must not break", we need to support both .. (where LTW is, or isn't installed).

Not sure how that would be executed. Execute the suite twice with buildconfig flag set and unset?

fadidurah: Yes, discussed this with Sowmya and most appropriate in my opinion is to have this flag be set based on a pipeline parameter/variable. That way you can choose to run this suite with the flag or not. you can run it twice to cover "Install LTW first and everything else must not break"

rpdome: Could you please add a comment in the code, above that line?

(For future engineers, so that they don't get confused)

rpdome (approved):
Left additional comments. Otherwise LGTM.

PR #2004, opened 2023-04-06
release 11.0.0

p3dr0rv:


iamgusain:
There should be build.gradle change too for common, consuming the new common4j.

p3dr0rv:
> There should be build.gradle change too for common, consuming the new common4j.

you are right, thanks!

fadidurah (approved):
LGTM

PR #2010, opened 2023-04-14
Pulling device cert issuer check to beginning of OnReceivedClientCertRequest

melissaahn:
## Summary
The C++ team let me know that some of their OneAuth UX automation tests were timing out due to the new user choice dialog I added as a part of the CBA support. These tests were ADFS related, but not CBA. The expected result was to not see any sort of CBA dialog at all.

The particular account that is being used here is related to an ADFS server that has its deviceAuthenticationMethod set to ClientTLS, so the ClientCertRequest coming in is for device authentication. A check was made a while back (when initial CBA support was first added) to account for this particular ClientCertRequest, and it immediately cancels the request once it determines it's for device auth, I mistakenly kept this check in the challenge handler for on-device CBA, when the check should actually be run before any sort of CBA logic can run. 

This PR pulls the cert issuer check to the very beginning of OnReceivedClientCertRequest, as there's no reason to create CBA challenge handlers for a request that isn't for CBA. I was able to first reproduce the issue with the test account OneAuth was using, and then I confirmed that the changes in this PR would suppress the user choice dialog, fixing the issue.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 V.NEXT
 ----------
 - [PATCH] Fix Error Type thrown for NO_ACCOUNT_FOUND (#2006)
+- [PATCH] Pulling device cert issuer check to beginning of OnReceivedClientCertRequest (#2010)
 
 V.11.0.0
 ----------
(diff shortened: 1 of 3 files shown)


nickbopp (approved):
Thanks for the quick investigation!

rpdome (approved):

PR #2012, opened 2023-04-17
Add telemetry attribute.

rpdome:
related: https://github.com/AzureAD/ad-accounts-for-android/pull/2238

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
@@ -124,5 +124,15 @@ public enum AttributeName {
     /**
      * The message accompanying an Exception.
      */
-    error_message;
+    error_message,
+
+    /**
+     * Name of the storage operation.
+     */
+    storage_operation_name,
+
+    /**
+     * Name of the storage being used..
+     */
+    storage_name
 }


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -124,5 +124,15 @@ public enum AttributeName {
...
+
+    /**
+     * Name of the storage operation.
+     */
+    storage_operation_name,
+
+    /**
+     * Name of the storage being used..
+     */
+    storage_name
```
shahzaibj, line 137: Can you ALSO add these in the broker repo as well? (There is another attribute name file in broker4j - we had to had these 2 files because the file in broker repo stores the data classifications as those classifications had to live in a private repo)

rpdome: oh ok. in that case i could **_just_** add these two in broker repo.

rpdome: (or do we also need them here as well?)

shahzaibj: If you only need to access these in Broker repo then better to put it in broker repo only

shahzaibj: Looks like you only access the constants in Broker repo (I don't see them being accessed in common) so let's put them in Broker repo ONLY.

rpdome:
Will put the attributes in broker repo.

PR #2015, opened 2023-04-18
Mapping ECDSA to EC in ClientCertRequest's keyTypes array

melissaahn:
### Problem
The CBA team brought me an issue they were seeing with on-device CBA, specifically on Pixels. Despite installing the user certificate in the user cert store, they couldn't get the system cert picker to pop up when selecting on-device CBA, and the request would immediately fail. I was able to reproduce the issue on my Pixel 6, but not my Samsung device. The CBA team mentioned that they weren't reproducing the error on their Samsung phones either. 

### Debugging
From my experience, there are two main ways in which the system cert picker will not show up:
- No certs are in the user store.
- An issuer and/or key types filter was provided, and none of the certs installed match those filters.

I played around with the keyTypes filter, and we were receiving the string array [RSA, ECDSA]. While RSA is [an expected constant in the keyTypes array](https://developer.android.com/reference/android/security/KeyChain#choosePrivateKeyAlias(android.app.Activity,%20android.security.KeyChainAliasCallback,%20java.lang.String[],%20java.security.Principal[],%20android.net.Uri,%20java.lang.String)), ECDSA is not. After learning that the certificate being installed was of an Elliptical Curve type, I tried adding the [EC constant](https://developer.android.com/reference/android/security/keystore/KeyProperties#KEY_ALGORITHM_EC) to the array, and the cert picker finally came up. 

For some reason, [Chromium doesn't map ECDSA to EC for WebViews](https://source.chromium.org/chromium/chromium/src/+/main:android_webview/browser/aw_contents_client_bridge.cc;l=184;bpv=1), but it does for [the chromium based browsers](https://source.chromium.org/chromium/chromium/src/+/main:components/browser_ui/client_certificate/android/ssl_client_certificate_request.cc;l=172?q=clientcertrequest&ss=chromium). I'm not certain why this is inconsistent, especially since ECDSA doesn't have a literal constant like RSA does. I'm even more confused on how we haven't had an issue with this for so long (as we haven't changed this logic in years), almost as if the filters weren't really working until now (and only on Pixels).

### Patch
The medium (?) term solution that we can make on our side is to directly map "ECDSA" in the keyTypes array to the EC constant, which is what this PR does (medium, because it's not really the root of the problem in my opinion, and previous versions of common not including this change will still have the issue). 
For the long term, it would be good to know the percentage of users intending to use RSA certs vs EC certs so that we can determine impact. Since we don't know the intentions of the user (and can't query the cert store) before selection, I added an Otel span named cert_based_auth_public_key_algo_type which is populated with the algorithm name of the cert a user picks. Hopefully this will give us a general idea of the types of certs being used in successful scenarios, which may help in further identifying the scope of a future issue. 

### Testing
I confirmed the cert picker was able to come up in the Pixel 6 for both RSA and EC certificates, and no behavior changed for the Samsung phone. 

### Related PRs
- Same changes need to be made in ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1726
- Adding new CBA Otel attribute name to broker file as well: https://github.com/AzureAD/ad-accounts-for-android/pull/2240

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@ V.NEXT
 - [PATCH] Fix Error Type thrown for NO_ACCOUNT_FOUND (#2006)
 - [PATCH] Pulling device cert issuer check to beginning of OnReceivedClientCertRequest (#2010)
 - [MINOR] Add method to clear receiver concurrentHashMap in LocalBroadcaster (#1993)
+- [PATCH] Mapping ECDSA to EC in ClientCertRequest's keyTypes array (#2015)
 
 V.11.0.0
 ----------
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandler.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/AbstractSmartcardCertBasedAuthChallengeHandler.java
@@ -287,6 +287,7 @@ public abstract class AbstractSmartcardCertBasedAuthChallengeHandler<T extends A
         final PrivateKey privateKey = session.getKeyForAuth(certDetails, pin);
         //Cert chain only needs the cert to be used for authentication.
         final X509Certificate[] chain = new X509Certificate[]{certDetails.getCertificate()};
+        mTelemetryHelper.setPublicKeyAlgoType(chain[0].getPublicKey().getAlgorithm());
         //Clear current dialog.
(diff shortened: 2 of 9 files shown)


rpdome (approved):

fadidurah (approved):
I think this may have been the same issue I ran into when trying to automate some of your manual tests. So we might be able to reduce your testing load a little each release! Let's follow up next month when I have some bandwidth

PR #2016, opened 2023-04-18
Move Broker encryption logic to broker code. Some more refactoring.

rpdome:
related: https://github.com/AzureAD/ad-accounts-for-android/pull/2241

## Changes
1. Move `AndroidBrokerStorageEncryptionManager` and its tests  to Broker (`AndroidBrokerStorageEncryptionManagerWithLegacyKeyTest`)

2. Remove ITelemetryCallback usage. We don't need it anymore.

3. Kill AndroidPlatformComponents. We don't use it anymore.


common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
```diff
@@ -109,8 +109,7 @@ public byte[] generate() {
     }
 
     @Override
-    @NonNull
-    public byte[] encrypt(@NonNull final byte[] plaintext)
```
iamgusain, line 113: Why are we removing the NonNull annotations?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
```diff
@@ -29,43 +29,51 @@
...
+ * and KeyStore-wrapped key.
  * */
 public class AndroidAuthSdkStorageEncryptionManager extends StorageEncryptionManager {
     private static final String TAG = AndroidAuthSdkStorageEncryptionManager.class.getSimpleName();
 
     /**
      * Alias persisting the keypair in AndroidKeyStore.
      */
-    /* package */ static final String KEY_STORE_ALIAS = "AdalKey";
+    public static final String KEY_STORE_ALIAS = "AdalKey";
```
iamgusain: Should we name this as "WRAPPING_KEY_ALIAS"

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
```diff
@@ -29,43 +29,51 @@
...
+
+    /**
+     * Name of the file contains the symmetric key used for encryption/decryption.
+     */
+    public static final String KEY_FILE_NAME = "adalks";
```
iamgusain: And this can be renamed as WRAPPED_KEY_FILE_NAME

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -87,22 +85,26 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
...
-     * Indicate that token item is encrypted with the key persisted in AndroidKeyStore.
+     * Indicate that token item is encrypted with the key loaded in this class.
      */
-    public static final String KEY_IDENTIFIER = "A001";
-
-    /**
-     * Name of the file contains the symmetric key used for encryption/decryption.
-     */
-    /* package */ static final String KEY_FILE_PATH = "adalks";
+    public static final String WRAPPED_KEY_KEY_IDENTIFIER = "A001";
```
iamgusain, line 90: Since the other one is named as USER_PROVIDED_KEY_IDENTIFIER, should we name this as 
AUTH_SDK_PROVIDED_KEY_IDENTIFIER?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -124,15 +126,16 @@ public SecretKey getData() {
      * Default constructor
      *
      * @param alias             Alias(name) of this key
```
iamgusain:
```suggestion
     * @param alias             Alias(name) of wrapping key
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -235,9 +233,7 @@ private AuthorizationResult performAuthorizationRequest(@NonNull final OAuth2Str
                 mAuthorizationStrategy
         );
 
-        final AuthorizationResult result = future.get();
-
-        return result;
+        return future.get();
```
iamgusain, line 236: Maybe we should add a timeout here?

iamgusain:
Let's also verify if this change would break oneauth

common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
rpdome: nonnull has no effect on primitives (byte array). I just removed as indicated by the warning.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
rpdome: Broker's going to use this as well.

rpdome:
> Let's also verify if this change would break oneauth

I don't think it would... the change regarding SDK are just renames and the removal of ITelemetryCallback.

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
rpdome: We should, but let's do that separately....

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
iamgusain:
````suggestion
    public static final String AUTH_LIB_PROVIDED_KEY_IDENTIFIER = "A001";
````
or maybe this could be
````suggestion
    public static final String DEFAULT_KEY_IDENTIFIER = "A001";
````
and the other one can be APPLICATION_PROVIDED_KEY_IDENTIFIER

Feel free to won't fix this.

iamgusain (approved):

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
rpdome: A001 is going to be used for both broker and auth libs though (and it's not configurable as of now).

i'll mark this as won't fix.

PR #2019, opened 2023-04-25
Replacing SHA-1 used in broker validation with SHA-512

melissaahn:
### Summary
This is the first of a few rounds of PRs that address the use of SHA-1 in our libraries. This set of PRs focuses on replacing SHA-1 usage in the broker verification logic with SHA-512. 

Main changes in Common:
- The AuthenticationConstants file containing hardcoded signature hashes (note: "signature" here refers to the [Android Signature Class](https://developer.android.com/reference/android/content/pm/Signature), which is an object representation of the signing certificate of an app... kinda confusing) now also contains SHA-512 hashes for brokerHost, Company Portal, and Authenticator. The BrokerData file has been updated with the SHA-512 static strings accordingly.
    - Teams IP Phone signature hash will be updated in a later PR. Its current use shouldn't be affected by the changes (its used with the original getCurrentSignatureForPackage method)
- The verifySignatureHash method now uses SHA-512 instead of SHA-1.
- The getCurrentSignatureForPackage static method now has a boolean parameter (which determines whether to use SHA-512 or not). This method is still used in many redirect URI and application identifier cases (which will be changes that will occur in the future), so two methods were created (getSha1SignatureForPackage and getSha512SignatureForPackage). The uses related to broker validation are using getSha512SignatureForPackage while redirect uri use cases will use getSha1SignatureForPackage. More uses cases will be ported over to the 512 method in the future.

### Testing
- General logging in and logging out with msalTestApp and Authenticator/Company Portal/BrokerHost as broker.
- Getting SSO Token (via BrokerHost broker api) 
- Device Token (via BrokerHost broker api) 
- Device signout unit tests passing

### Related PRs
- Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/2251
- ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1732
- MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1826


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1032,28 +1032,52 @@ public static String computeMaxHostBrokerProtocol() {
 
         /**
          * Signature info for Intune Company portal app that installs authenticator
-         * component.
+         * component. Generated with SHA-1.
          */
         public static final String COMPANY_PORTAL_APP_RELEASE_SIGNATURE = "1L4Z9FJCgn5c0VLhyAxC5O9LdlE=";
```
rpdome, line 1037: Do we still need these sha1 values?

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -79,7 +79,25 @@ public PackageHelper(final Context context) {
...
+    }
+
+    /**
+     * Reads first signature in the list for given package name.
+     *
+     * @param packageName name of the package for which signature should be returned
+     * @param useSha512 if true, uses SHA-512 to generate signature hash (should be used for verification purposes); if false, uses default SHA (redirect URI purposes)
+     * @return signature for package
+     */
+    public String getCurrentSignatureForPackage(final String packageName,
```
rpdome: nit - to make sure people do not use the wrong method, let's make 2 public method

"getSha1SignatureForPackage"
"getSha512SignatureForPackage"

and you can make this new method private.

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -134,15 +134,14 @@ public static List<X509Certificate> readCertDataForApp(final String packageName,
      * @throws CertificateEncodingException if a certificate was corrupt.
      * @throws ClientException if no valid hash was found in the list.
      */
-    public static final String verifySignatureHash(final @NonNull List<X509Certificate> certs,
+    public static String verifySignatureHash(final @NonNull List<X509Certificate> certs,
                                              final @NonNull Iterator<String> validHashes)
```
rpdome, line 138: nit: either rename the method or rename the variable (to make it explicit that we're validating againts sha 512 hashes only.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
melissaahn: Yeah, they're referenced in some redirect uri checks and tests. Though the brokerHost sha1 signature isn't currently being used, I don't think it would hurt to keep it until we actually change the redirect uri ourselves.

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
melissaahn: Renamed variable inside method: cddee2f39f960e1a1c9064564b56ab01962a30c7

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
melissaahn: Addressed here: 7c4e9582357f2cd85b9344f335d471303ad4aa5d

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -71,34 +70,52 @@ public PackageHelper(final Context context) {
...
+            return getCurrentSignatureForPackage(getPackageInfo(mPackageManager, packageName), true);
         } catch (NameNotFoundException e) {
             Logger.error(methodTag, "Calling App's package does not exist in PackageManager. ", "", e);
         }
         return null;
     }
 
     /**
-     * Reads first signature in the list for given package name.
+     * Reads first signature in the list for given package name.v
```
mohitc1: >v [](http://example.com/codeflow?start=64&length=1)

nit

mohitc1 (approved):

PR #2021, opened 2023-04-25
Updating different NFC smartcards error message

melissaahn:
### Summary
For the scenario when a user tries to tap a different smart card via NFC than the one initially used (for the first tap), a general error message dialog was shown. This PR changes the message to be more specific, which should help remind users that they should be using the same smartcard for both taps.

### Testing
Made sure that the specific error message was showing up correctly via manual testing. Switched the system to a different language to make sure localization was also taking effect.
Note: some of the strings.xml files have an extra string change because of an update I made to the strings last month, so it is expected.


fadidurah (approved):

iamgusain (approved):
LGTM

PR #2024, opened 2023-04-26
Consolidate IStorageSupplier logic

rpdome:
related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2256

The PR looks huge... but the actual changes happens in a few files only. The rest are due to the class/method signature change.

- Remove `getStorageEncryptionManager()` from IPlatformComponents.
- Have IPlatformComponents return `IStorageSupplier` directly.
- Consolidate `IStorageSupplier` methods - there is no "get encrypted" and "get unencrypted" storage. There's only "get storage" methods. It's up to the implementation if the data's going to be encrypted or not.
- Each storage layer will contain their own Encryption Manager. This will allow us to have a different encryption manager for each storages.
     - See `AndroidEncryptedStorageSupplier` (code moved from `AndroidStorageSupplier`)

Note: With this PR - We'll also start encrypting everything
Right now the data that are unencrypted are _**NOT**_ sensitive (app metadata, etc), but this will future-proof any accidents (or any legal changes) that might happen in the future.

Backward compatibility wise - our [StorageEncryptionManager can handle unencrypted data](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/60be84d8114c177c65fa8b63bd4d7c7b1f558bac/common4j/src/test/com/microsoft/identity/common/java/crypto/StorageEncryptionManagerTest.java#L188) It would just return the value unencrypted if that's the case.


common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -119,7 +119,6 @@ private static void fillBuilder(@NonNull final PlatformComponents.PlatformCompon
                                     @NonNull final Context context,
                                     @Nullable final Activity activity,
                                     @Nullable final Fragment fragment) {
-        builder.storageEncryptionManager(new AndroidAuthSdkStorageEncryptionManager(context));
         fillBuilderWithBasicImplementations(builder, context, activity, fragment);
```
melissaahn, line 110: Is it possible to replace the fillBuilder call in line 110 with fillBuilderWithBasicImplementation, and then delete fillBuilder?

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/components/AndroidEncryptedStorageSupplier.kt
```diff
@@ -0,0 +1,54 @@
...
+class AndroidEncryptedStorageSupplier(private val context: Context,
+                                      private val storageEncryptionManager: StorageEncryptionManager)
+    : IStorageSupplier {
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> getNameValueStore(storeName: String, clazz: Class<T>): INameValueStorage<T> {
+        val mgr: IMultiTypeNameValueStorage =
+            SharedPreferencesFileManager.getSharedPreferences(context, storeName, storageEncryptionManager)
+        if (Long::class.java.isAssignableFrom(clazz)) {
+            return (SharedPreferenceLongStorage(mgr) as INameValueStorage<T>)
```
iamgusain, line 43: Are we using any if these?
`SharedPreferenceLongStorage` or `SharedPrefStringNameValueStorage`
I think we only have the String values, if that is the case, we should remove the other one, to keep it simple

common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt
```diff
@@ -0,0 +1,46 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.interfaces
+
+import com.microsoft.identity.common.java.cache.IMultiTypeNameValueStorage
+
+/**
+ * An interface for loading [INameValueStorage]
+ */
+interface IStorageSupplier {
```
iamgusain, line 30: Can we simplify this by only exposing one method to getNameValueStore( storeIdentifier: String)
I don't think we are using the specific typed or multitype NameValue storage.

common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java
```diff
@@ -37,21 +37,14 @@
 
 /**
  * A platform agnostic implementation of {@link IPerSeparatorMultiTypeNameValueStorage} where the
- * separator is of type {@link String}. The internal storage mechanism here could be backed by
- * anything as determined by the {@link IPlatformComponents#getFileStore(String)} or even be an
- * encrypted variant as determined by {@link IPlatformComponents#getEncryptedFileStore(String, IKeyAccessor)}.
+ * separator is of type {@link String}.
  */
 @AllArgsConstructor
 public class StringSeparatedMultiTypeNameValueStorage extends AbstractPerSeparatorMultiTypeNameValueStorage<String> {
```
iamgusain, line 43: Is this <`StringSeparatedMultiTypeNameValueStorage`> being used anywhere?

common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
```diff
@@ -0,0 +1,63 @@
...
+package com.microsoft.identity.common.components
+
+import com.microsoft.identity.common.java.cache.IMultiTypeNameValueStorage
+import com.microsoft.identity.common.java.cache.MapBackedPreferencesManager
+import com.microsoft.identity.common.java.interfaces.INameValueStorage
+import com.microsoft.identity.common.java.interfaces.IStorageSupplier
+import com.microsoft.identity.common.java.util.ported.InMemoryStorage
+import java.util.concurrent.ConcurrentHashMap
+
+class InMemoryStorageSupplier : IStorageSupplier {
```
iamgusain: Please add javadocs or should I say kotlindocs :)

common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
```diff
@@ -0,0 +1,63 @@
...
+        ConcurrentHashMap()
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> getNameValueStore(storeName: String, clazz: Class<T>): INameValueStorage<T> {
+        val ret = mNameValueStores[storeName]
+        if (ret != null) {
+            return ret as INameValueStorage<T>
+        }
+
+        val nameValueStore = InMemoryStorage<Any>()
```
iamgusain: why "Any"??

common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val mNameValueStores: MutableMap<String, INameValueStorage<Any>> =
+        ConcurrentHashMap()
+
+
+    private val mFileStores: MutableMap<String, IMultiTypeNameValueStorage> =
+        ConcurrentHashMap()
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> getNameValueStore(storeName: String, clazz: Class<T>): INameValueStorage<T> {
+        val ret = mNameValueStores[storeName]
```
iamgusain:
````suggestion
        val nameValueStore = mNameValueStores[storeName]

File: common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt (hunk 6 above)
Comment (rpdome) [reply common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt]: to match the type in mNameValueStores. (getNameValueStore can be used to get any storage types).

File: common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java (hunk 4 above)
Comment (rpdome) [reply common4j/src/main/com/microsoft/identity/common/java/storage/StringSeparatedMultiTypeNameValueStorage.java:43]: WpjTelemetryMetadataStore

File: common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt (hunk 3 above)
Comment (rpdome) [reply common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt:30]: yes... make sense.

Comment (rpdome) [reply common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt:30]: We're using Long storage in linux.

File: common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt (hunk 7 above)
Comment (rpdome) [reply common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt]: the name is already used below. (I'm avoiding using var as much as possible as best practice)

renamed to existingStorage instead.

File: common/src/main/java/com/microsoft/identity/common/components/AndroidEncryptedStorageSupplier.kt (hunk 2 above)
Comment (rpdome) [reply common/src/main/java/com/microsoft/identity/common/components/AndroidEncryptedStorageSupplier.kt:43]: we use SharedPreferenceLongStorage, but we're not getting it from here...

It's probably okay to just keep it here to match with what we're supporting in linux...

Comment (iamgusain) [reply common/src/main/java/com/microsoft/identity/common/components/AndroidEncryptedStorageSupplier.kt:43]: If we are not using it, then let's remove it from here.

File: common4j/src/main/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt [hunk 8]
```diff
@@ -0,0 +1,67 @@
...
+import com.microsoft.identity.common.java.cache.IMultiTypeNameValueStorage
+import com.microsoft.identity.common.java.cache.MapBackedPreferencesManager
+import com.microsoft.identity.common.java.interfaces.INameValueStorage
+import com.microsoft.identity.common.java.interfaces.IStorageSupplier
+import com.microsoft.identity.common.java.util.ported.InMemoryStorage
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Storage provided by this class are stored in-memory.
+ * Used in tests only.
```
iamgusain: If this is only for tests, let's move it under testFixtures

common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt
iamgusain: What is the difference between. NameValue and File Store?
When do we use one vs the other.

rpdome: One is specific to a certain type, the other is more flexible.

Technically we should be able to convert all usage of nameValue to the latter, let's see how much of the change that is....

Edit: Not possible due to Linux.  (or rather.. not simple to do as part of this PR).

iamgusain: So, in Linux we are using both one generic one that can support more than 1 type and the other is specific typed. (Long) ?

iamgusain (approved):

common4j/src/main/com/microsoft/identity/common/java/interfaces/IStorageSupplier.kt
rpdome: @iamgusain yes

PR #2025, opened 2023-04-27
Moving ClearCertPref call to clean up method

melissaahn:
### Summary
The C++ team mentioned that they're still seeing some ERR_CERT_DATABASE_CHANGED errors pop up in a few apps, which we attributed last time to the [WebView.ClearClientCertPreferences](https://developer.android.com/reference/android/webkit/WebView#clearClientCertPreferences(java.lang.Runnable)) call that we have to make to allow CBA users to add more than one account without having to restart the app.
This PR moves the one ClearClientCertPreferences call we have to the cleanUp method of the OnDevice challenge handler. The good thing about this move is that it will only get called if on-device CBA was ever initiated, instead of every time the WebView gets initialized. (This option has more recently been easily accessible due to the multiple rounds of refactoring that have been done to CBA in the previous few months :) ). 
If this change doesn't improve the hit rate of the error the OneAuth folks are seeing, I'll talk to the CBA team regarding the customer impact of removing/keeping this method call.

### Testing
Tested logging in with CBA and making sure the clearClientCertPreferences call isn't causing any abnormal behavior. Was able to add another account without closing the app, and I was also able to cancel out of the CBA flow, but then successfully add an account upon restart of the WebView.  Made sure that the method call wasn't being made in a regular password login flow.
Also asked the C++ team if they can run this branch through their automated tests to catch any issues on their side early.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/OnDeviceCertBasedAuthChallengeHandler.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/OnDeviceCertBasedAuthChallengeHandler.java
@@ -129,7 +129,8 @@ public class OnDeviceCertBasedAuthChallengeHandler extends AbstractCertBasedAuth
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             WebView.clearClientCertPreferences(null);
         } else {
-            Logger.warn(methodTag, "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
+            Logger.warn(methodTag, "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP). " +
+                    "Subsequent CBA attempts will fail due to the cached action, so the user must restart the app before attempting to login with CBA again.");
         }
     }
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/OnDeviceCertBasedAuthChallengeHandler.java
```diff
@@ -122,7 +123,14 @@ public void alias(String alias) {
...
     @Override
     public void cleanUp() {
-        //Nothing needed at the moment.
+        final String methodTag = TAG + ":cleanUp";
+        //For on-device CBA, we need to clear the certificate choice cache here so that
+        // the user will be able to login with multiple accounts with CBA
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            WebView.clearClientCertPreferences(null);
+        } else {
+            Logger.warn(methodTag, "Client Cert Preferences cache not cleared due to SDK version < 21 (LOLLIPOP)");
```
iamgusain: Could you also add to this message, the possible effect of cleanup not happening.
(what issues/errors user might see)

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/OnDeviceCertBasedAuthChallengeHandler.java
melissaahn: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2025/commits/58df6b36b5bb66e3aa3bfb6efccd610679d51014

nickbopp (approved):
MSAL.CPP tests are all green.

PR #2026, opened 2023-04-28
Improve getCurrentActiveBrokerPackageName function to make it more reliable.

p3dr0rv:
## Why
Authenticator has reported cases were broker validation fails because none of the authenticator types is broker account type.
We cannot reproduce the issue and the conditions to repro the issue are unknown (it is happening across multiple OEMs and SDK versions)

we decided to reduce the variables that might be causing the issue, by:
1. removing the case sensitivity.
2. trimming the string


fadidurah (commented):
Do we need extra testing for this change?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
```diff
@@ -245,8 +216,7 @@ public String getCurrentActiveBrokerPackageName() {
      * @return PackageName of the broker
      */
     @NonNull
-    // TODO: Consolidate getCurrentActiveBrokerPackageName and getValidActiveBrokerPackageName into one.
-    public String getValidActiveBrokerPackageName() throws ClientException {
```
iamgusain, line 249: Is there any other team that depends on this method, Intune?, Worth checking, once before removing.
Maybe mark is as deprecated with a comment that this would be removed in next version. and call the new method from inside it.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -178,8 +170,12 @@ public HelloCache getHelloCache() {
     }
 
     @VisibleForTesting
-    public String getActiveBrokerPackageName() {
-        return new BrokerValidator(mApplicationContext).getCurrentActiveBrokerPackageName();
+    public String getActiveBrokerPackageName() throws IllegalStateException{
+        try {
+            return new BrokerValidator(mApplicationContext).getCurrentActiveBrokerPackageName();
+        } catch (final ClientException e) {
```
iamgusain, line 176: Is ClientException the only type of exception to catch here?

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java
p3dr0rv: this method is similar to getCurrentActiveBrokerPackageName, it was created with the intention to log more information.
it was only used in the controller of the new WPJ API.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
p3dr0rv: yes, we are catching the exception in getCurrentActiveBrokerPackageName

PR #2027, opened 2023-04-28
Fix Access token caching

mohitc1:
**Issue:**
Steps to repro the issue
1. Acquire token interactively (Broker) with user.read
This setups up access token like this in the cache
7df0e758-afda-447e-a298-232b5d5baa12.5e930fc0-2e40-43d7-88e3-375e7f0f9d8e-login.windows.net-accesstoken-4b0db8c2-9f26-4417-8bde-3f0e3656f8e0-5e930fc0-2e40-43d7-88e3-375e7f0f9d8e-**openid User.Read profile email**-com.msft.identity.client.sample.local/1wiqxsqbj7w+h11zifsnqwgykry=
2. Acquire token interactively (broker) with user.read scope and with device claim string {\"access_token\":{\"deviceid\":{\"essential\":true}}}"
This setups up new cache value as 7df0e758-afda-447e-a298-232b5d5baa12.5e930fc0-2e40-43d7-88e3-375e7f0f9d8e-login.windows.net-accesstoken-4b0db8c2-9f26-4417-8bde-3f0e3656f8e0-5e930fc0-2e40-43d7-88e3-375e7f0f9d8e-**user.read openid offline_access profile**-com.msft.identity.client.sample.local/1wiqxsqbj7w+h11zifsnqwgykry=

There's difference between cache key values to different target strings (target strings are built from scopes).
3. New ATS call comes with the same claim string. Even though this would request a new AT from eSTS it would end up updating the 2nd entry and not the first. And then in the result, we end up choosing value stored in step 1(without device id claim).

Note: Above steps assume that the corresponding broker fix has been made.

 **Fix target in token records**
1. Fixing target string using response scopes instead of request.
2. Fixing logic where we delete any existing AT which intersect with new AT returned in response. Improved comparison by doing case-insensitive matching of scopes, so that we don't end up duplicate ATs.
Related broker PR https://github.com/AzureAD/ad-accounts-for-android/pull/2259

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,6 +10,7 @@ V.NEXT
 - [PATCH] Fix NPE in OTEL code (#2018)
 - [MINOR] Updating different smartcards error message; updated string res files (#2021)
 - [PATCH] Moving ClearCertPref call to clean up method (#2025)
+- [PATCH] Fix target in token records to fix cache keys
 
 V.11.0.0
 ----------
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,7 +10,7 @@ V.NEXT
 - [PATCH] Fix NPE in OTEL code (#2018)
 - [MINOR] Updating different smartcards error message; updated string res files (#2021)
 - [PATCH] Moving ClearCertPref call to clean up method (#2025)
-- [PATCH] Fix target in token records to fix cache keys
+- [PATCH] Fix target in token records to fix cache keys (#2027)
 
 V.11.0.0
 ----------
--- a/common/src/test/java/com/microsoft/identity/common/MicrosoftStsAccountCredentialAdapterTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/MicrosoftStsAccountCredentialAdapterTest.java
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common;
 
+import com.microsoft.identity.common.java.AuthenticationConstants;
 import com.microsoft.identity.common.java.authorities.Authority;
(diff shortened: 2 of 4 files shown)


rpdome:
Is this a breaking change? as in, would I still be able to get a cache hit if the entry is written *before* this fix?

(i'm not as worried for AT because we can just get a new one. RT for non-joined flow is a different story).

mohitc1:
Yes, it would still get. The change is around on persisting response scopes (Querying part remains same and takes care of removing default scopes and matching in case-insensitive manner) and replacing existing AT if scopes intersect.
I tested following flows
Non-Joined
1. Acquire token interactively non-joined
2. Acquire token silently with and without force refresh
3. upgrade app which has this change
4. Acquire token silently with and without force refresh -> successful.

Joined
1. Acquire token interactively
2. Acquire token silently.
3. Acquire token with access token device id claim, perform inline WPJ. Result AT has device id claim
4. Upgrade app to this change.
5. Acquire token silently without claim -> Has device id claim (from cache)
6. Acquire token silently with device id claim -> Has device claim



---
In reply to: [1527854802](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2027#issuecomment-1527854802) [](http://example.com/codeflow?ancestors=1527854802,1527854802)

shahzaibj:
So querying part was already matching in case-insensitive manner?

rpdome (approved):

iamgusain (approved):

mohitc1:
> So querying part was already matching in case-insensitive manner?

Yes, referring to method AbstractAccountcredentialCache#targesIntersect

PR #2028, opened 2023-05-01
Rapong/prototype

rpdome:


common/src/main/kotlin/com/microsoft/identity/common/BrokerDiscoveryClient.kt
```diff
@@ -0,0 +1,166 @@
...
+                                    forceTrigger: Boolean): BrokerData? {
+            val methodTag = "$TAG:queryFromBroker"
+
+            val bundle = Bundle()
+            bundle.putBoolean(FORCE_DISCOVERY_BUNDLE_KEY, forceTrigger)
+
+            // ping all candidates to frontload bootstrapping!
+            val installedCandidates = brokerCandidates.filter(isPackageInstalled)
+
+            val list = mutableListOf<Deferred<Unit>>()
```
rpdome, line 73: Add tests where broker discovery on multiple broker happens multiple time.
How are we making sure the result is NOT corrupted?

(technically... they should yield the same result... but just in case... maybe use timestamp when the whole flow is initiated?)

rpdome: apart from that... maybe we still don't need to parallelize this .

PR #2030, opened 2023-05-02
add active broker cache

rpdome:
This is part of the LTW Broker support.

This cache will be used on both Broker and SDK side.

- Broker side = remember who active broker is, so that next time it gets a request from an SDK, it knows what it needs to do.
- SDK side = remember who active broker is, so that it can just forward the request directly.

To support broker hosting app (which dual stacks MSAL and Broker), I created 2 separate helper functions to avoid the same storage from being accessed from >1 processes.

Also moved BrokerData to kotlin...


common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.microsoft.identity.common.internal.broker.BrokerData
+import com.microsoft.identity.common.java.interfaces.INameValueStorage
+import com.microsoft.identity.common.java.interfaces.IStorageSupplier
+import net.jcip.annotations.ThreadSafe
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+@ThreadSafe
+class ActiveBrokerCache
```
fadidurah, line 37: kotlindoc

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -0,0 +1,118 @@
...
+            return false
+        }
+
+        return packageName.equals(other.packageName, ignoreCase = true) &&
+                signatureHash == other.signatureHash
+    }
+
+    override fun hashCode(): Int {
+        var result = packageName.hashCode()
+        result = 31 * result + signatureHash.hashCode()
```
p3dr0rv, line 51: why 31?

rpdome: This was actually autogenerated by the IDE.

As for your question - https://stackoverflow.com/questions/60406166/why-do-we-use-number-31-while-calculating-hashcode

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -0,0 +1,118 @@
...
+            AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME,
+            AuthenticationConstants.Broker.COMPANY_PORTAL_APP_RELEASE_SIGNATURE
+        )
+
+        val debugBrokerHost = BrokerData(
+            AuthenticationConstants.Broker.BROKER_HOST_APP_PACKAGE_NAME,
+            AuthenticationConstants.Broker.BROKER_HOST_APP_SIGNATURE
+        )
+
+        val debugBrokers: Set<BrokerData> =
```
p3dr0rv, line 81: nit: consider adding @JvmStatic here and below, this will prevent your Java code from polluting with the name Companion everywhere.

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
```diff
@@ -0,0 +1,137 @@
...
+            inMemoryCachedValue = BrokerData(packageName, signatureHash)
+            return inMemoryCachedValue
+        }
+    }
+
+    override fun setCachedActiveBroker(brokerData: BrokerData) {
+        lock.write {
+            storage.put(ACTIVE_BROKER_CACHE_PACKAGE_NAME_KEY, brokerData.packageName)
+            storage.put(ACTIVE_BROKER_CACHE_SIGHASH_KEY, brokerData.signatureHash)
+            inMemoryCachedValue = BrokerData(brokerData.packageName, brokerData.signatureHash)
```
p3dr0rv: BrokerData.copy() will also work? this is just the way you prefer to do it, right?

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
rpdome: lol no. You overestimated my skill 😋

PR #2033, opened 2023-05-02
Consolidate Storage Supplier (again)

rpdome:
pretty much https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2024, with the following change.

IStorageSupplier now returns getEncryptedStorage and getUnencryptedStorage.
(similar with what we currently have, but make the name obvious, and remove the encryption manager from the parameter list).

---

## Why do we have to do this?

With the original PR, the issue is that we have a handful of storages (files) which are shared by msal and broker. (in this case, EstsTelemetry). These storage are previously NOT encrypted... so no error has been thrown. 

As I enforce encryption, now it stops working in broker hosting apps (which dual stacks MSAL and Broker).... as broker and MSAL does NOT share keys. 

- From BrokerHost, we make a request with BrokerHost's MSAL, so BrokerHost's MSAL, writes its telemetry data encrypted with its key.
- As we switch to the MSAL test app, we invoke acquireToken. The request goes to broker. Broker tries to read the telemetry data... and found the data encrypted by its BrokerHost's MSAL.
- Broker does NOT have an access to BrokerHost's MSAL, so an exception is thrown.

Ultimately, it's ideal to have separate storages for broker and MSAL (e.g. because you don't want to mix up telemetry data between MSAL and Broker) - but that's quite costly so it's not going to be done soon.


common4j/src/testFixtures/java/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
```diff
@@ -0,0 +1,78 @@
...
+        val fileStore = MapBackedPreferencesManager()
+        mFileStores[storeName] = fileStore
+        return fileStore
+    }
+
+    override fun <T> getEncryptedNameValueStore(
+        storeName: String,
+        clazz: Class<T>
+    ): INameValueStorage<T> {
+        return getUnencryptedNameValueStore(storeName, clazz)
```
iamgusain, line 72: Should be throwing NotSupported Exception.
Else test might fail, trying to decrypt.

common4j/src/testFixtures/java/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    override fun getEncryptedFileStore(storeName: String): IMultiTypeNameValueStorage {
+        return getUnencryptedFileStore(storeName)
```
iamgusain, line 76: same here

iamgusain (approved):

iamgusain (approved):

melissaahn (approved):

common4j/src/testFixtures/java/com/microsoft/identity/common/components/InMemoryStorageSupplier.kt
rpdome: This one we're using in unit tests. We do not need to do actual encryption here.

PR #2034, opened 2023-05-02
Modify Broker Host to support e2e tests that require MWPJ enable

p3dr0rv:
Related PR https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1817


fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -44,10 +46,13 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 public class BrokerHost extends AbstractTestBroker {
     private final static String TAG = BrokerHost.class.getSimpleName();
     // tenant id where lab api and key vault api is registered
     private final static String LAB_API_TENANT_ID = "72f988bf-86f1-41af-91ab-2d7cd011db47";
+    // flight to enable/disable the multiple wpj feature
+    private final static String FLIGHT_FOR_WORKPLACE_JOIN_CONTROLLER = "ENABLE_MULTIPLE_WORKPLACE_JOIN_PP";
```
rpdome, line 61: can we use the enum directly?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -79,19 +85,32 @@ public BrokerHost(@NonNull final String brokerHostApkName,
         brokerFlightsFragment = new BrokerFlightsFragment();
         brokerApiFragment = new BrokerApiFragment();
         singleWpjApiFragment = new SingleWpjApiFragment();
+        multipleWpjApiFragment = new MultipleWpjApiFragment();
     }
 
     @Override
     public void performDeviceRegistration(@NonNull final String username,
```
rpdome, line 102: there probably should be 3 sets of methods.

1. Legacy WPJ operations
2. MWPJ Operations
3. override methods. This will either invoke 1 or 2 based on the flag.

You can put 1 and 2 into a separate class for readability.

If you're writing test specifically for MWPJ, invoke 2
If you're writing test that specifically requires legacy WPJ, invoke 1
Otherwise, invoke the override methods. (the rest of the test suites doesn't care if MWPJ or legacy WPJ is used).

rpdome (commented):
Left some comments.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
p3dr0rv: I am hesitant to do this change because it will require making uiautomationutilities dependent on broker4j, if you think this is fine I can do the change and update the pipelines that build uiautomationutilities

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
p3dr0rv: Legacy operations are contained in the Single WPJ API fragment, 
and MWPJ operations in the Multiple WPJ API fragment,
by default the broker interface uses the legacy API.

rpdome: Please add that as a comment somewhere.

rpdome (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
p3dr0rv: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2034/commits/76ecbc02d47877ec42b45ae63513912e32b63e74

PR #2036, opened 2023-05-04
Add Broker Metadata Retrieval operation, Tweaks IPCStrategy

rpdome:
- Add Broker Metadata Retrieval IPC Operation For https://github.com/AzureAD/ad-accounts-for-android/pull/2268
- Make changes to existing IPC strategies to double check that the service is available before making a request. (see `AbstractIpcStrategyWithServiceValidation `)
- Removed the unused Bound Service Client.


common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -87,13 +88,13 @@ public enum Operation {
     }
 
     @Getter
-    @NonNull final private Operation operation;
```
rpdome, line 90: make this change because getter somehow doesn't play nice with lombok here (And we shouldn't need it to begin with)

Intentionally keep the change minimum so that i don't bloat up the PR and scare people off.

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1572,6 +1573,11 @@ public enum API {
          */
         public static final String BROKER_API_GET_FLIGHTS_PATH = "/brokerApi/getFlights";
 
+        /**
+         * ContentProvider path for retrieving Broker Discovery Metadata.
+         */
+        public static final String BROKER_DISCOVERY_METADATA_RETRIEVAL_PATH = "/brokerElection/brokerDiscoveryMetadataRetrieval";
```
iamgusain:
````suggestion
        public static final String RETRIEVE_BROKER_DISCOVERY_METADATA_PATH = "/brokerApi/retreiveBrokerDiscoveryMetadata";

Comment (rpdome) [reply common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java]: will rename the variable, but i'll keep the path the same (since this is not part of the broker hosting app API (getAccounts, getFlags/setFlags, etc)

File: common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStrategy.java [hunk 3]
```diff
@@ -140,7 +149,8 @@ private String getContentProviderAuthority(final @NonNull String targetedBrokerP
     /**
      * Returns true if the target package name supports this content provider strategy.
      */
-    public boolean isBrokerContentProviderAvailable(final @NonNull String targetedBrokerPackageName) {
+    @Override
+    public boolean isSupportedByTargetedApp(final @NonNull String targetedBrokerPackageName) {
         final String methodTag = TAG + ":isBrokerContentProviderAvailable";
```
mohitc1: >isBrokerContentProviderAvailable [](http://example.com/codeflow?start=41&length=32)

Update tag.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/AbstractIpcStrategyWithServiceValidation.kt
```diff
@@ -0,0 +1,71 @@
...
+     * @param bundle a [BrokerOperationBundle] object.
+     * @return a response bundle (returned from the active broker).
+     */
+    @Throws(BrokerCommunicationException::class)
+    protected abstract fun communicateToBrokerAfterValidation(bundle: BrokerOperationBundle): Bundle?
+
+    /**
+     * Returns true if the target package name supports this strategy.
+     */
+    abstract fun isSupportedByTargetedApp(targetedBrokerPackageName: String): Boolean
```
mohitc1: >isSupportedByTargetedApp [](http://example.com/codeflow?start=17&length=24)

nit: isSupportedByTargetedBroker?

mohitc1 (approved):

PR #2039, opened 2023-05-11
Add BrokerDiscoveryClient + Tests

rpdome:
In order for MSAL/OneAuth to figure out who the active broker is (under the new broker election mechanism), we'll be using `BrokerDiscoveryClient` to do the job.

This class will
1. Return the cached value, if there is any (so we do not have to invoke IPC every time we're making a request).
2. If there is no cache value (or shouldSkipCache = true), trigger IPC requests to the list of (hardcoded) known broker apps - one by one.
3. If all IPC request fails, fall back to AccountManager.

cc: @nickbopp - in case you have any feedback, especially on performance angle.


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -52,7 +54,58 @@ data class BrokerData(val packageName : String,
...
+         **/
+        fun getKnownBrokerApps() : Set<BrokerData> {
+            return if (sShouldTrustDebugBrokers) allBrokers else prodBrokers
+        }
+
+        /**
+         * Validate that the installed counterpart of the given [BrokerData] is signed by known keys.
+         * @return true if the validation succeeds. False otherwise.
+         **/
+        fun isSignedByKnownKeys(brokerData: BrokerData, context: Context): Boolean {
```
rpdome, line 82: this is taken from BrokerValidator.verifySignatureAndThrow

common/src/main/java/com/microsoft/identity/common/internal/util/AccountManagerUtil.java
```diff
@@ -1,99 +0,0 @@
...
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-
-package com.microsoft.identity.common.internal.util;
```
rpdome, line 24: No longer used anywhere.

common/src/test/java/com/microsoft/identity/common/internal/activebrokerdiscovery/InMemoryActiveBrokerCache.kt
```diff
@@ -0,0 +1,49 @@
...
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.activebrokerdiscovery
+
+import com.microsoft.identity.common.internal.broker.BrokerData
+import com.microsoft.identity.common.internal.cache.IActiveBrokerCache
+
+/**
+ * An [IActiveBrokerCache] which stores values in-memory.
+ **/
+class InMemoryActiveBrokerCache: IActiveBrokerCache {
```
fadidurah, line 31: is this supposed to be in test package?

fadidurah (approved):

common/src/test/java/com/microsoft/identity/common/internal/activebrokerdiscovery/InMemoryActiveBrokerCache.kt
rpdome: This is in the test directory.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -0,0 +1,198 @@
...
+                isPackageInstalledAndEnabled(brokerData.packageName)
+        })
+
+    override fun getActiveBroker(shouldSkipCache: Boolean): BrokerData? {
+        val methodTag = "$TAG:getActiveBroker"
+
+        lock.lock()
+        try {
+            if (!shouldSkipCache) {
+                val cachedData = cache.getCachedActiveBroker()
```
p3dr0rv, line 158: nit: cachedBroker

p3dr0rv (approved):
LGTM

PR #2044, opened 2023-05-18
Added common method to check if a current package name is a valid teams app

Amritkrsaini1:
Added utility to check if current package name is from valid installed teams app.


somalaya:
Can you provide a description on why this change is required?

somalaya (approved):

mohitc1 (approved):

PR #2050, opened 2023-05-24
Stop caching account manager values. Make BrokerDiscoveryClient coroutine-safe

rpdome:
1. Parallelize IPC requests in BrokerDiscoveryRequests.

2. Stop caching account manager values.... because if broker switch happens, then the cached value here could be blocking. we will always get a fresh value.

3. Introduce `IClientActiveBrokerCache.shouldUseAccountManager()`, which, if returns true, will just fall back to account manager. This is an optimization for the phase where we're flighting the new mechanism. Without it, the SDK might end up making unnecessary IPC requests to unsupported broker. 

4. Make the class coroutine safe. 

Given the code below, each coroutines would be triggered in the same thread (In this case, main thread).
```
runBlocking {
            launch {
                Assert.assertEquals(prodCompanyPortal, client1.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client2.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client3.getActiveBroker())
                countDownLatch.countDown()
            }
            launch {
                Assert.assertEquals(prodCompanyPortal, client2.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client3.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client1.getActiveBroker())
                countDownLatch.countDown()
            }
            launch {
                Assert.assertEquals(prodCompanyPortal, client3.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client1.getActiveBroker())
                Assert.assertEquals(prodCompanyPortal, client2.getActiveBroker())
                countDownLatch.countDown()
            }
        }
```
Therefore, we cannot use Java's Lock (which only protects in thread level). We need to use coroutine's lock.

coroutine's lock needs to be in... a coroutine, so i'm wrapping them with runBlocking.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -36,6 +36,7 @@ import com.microsoft.identity.common.java.interfaces.IPlatformComponents
 import com.microsoft.identity.common.java.logging.Logger
 import kotlinx.coroutines.*
 import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 
 
 /**
@@ -167,8 +168,7 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
         val methodTag = "$TAG:getActiveBroker"
 
         return runBlocking {
-            classLevelLock.lock()
-            try {
+            classLevelLock.withLock {
                 if (!shouldSkipCache) {
                     val cachedData = cache.getCachedActiveBroker()
                     cachedData?.let {
@@ -203,8 +203,6 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
                 )
 
                 return@runBlocking accountManagerResult
-            } finally {
-                classLevelLock.unlock()
             }
         }
     }
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -72,11 +72,6 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
         const val ERROR_BUNDLE_KEY = "ERROR_BUNDLE_KEY"
 
         /**
-         * Error code to be returned when the broker determines that only account manager can be used.
-         **/
-        const val ONLY_SUPPORTS_ACCOUNT_MANAGER_ERROR_CODE = "ONLY_SUPPORTS_ACCOUNT_MANAGER_ERROR_CODE"
-
-        /**
          * Per-process Thread-safe, coroutine-safe Mutex of this class.
          * This is to prevent the IPC mechanism from being unnecessarily triggered due to race condition.
          *
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -206,7 +206,8 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
                     isPackageInstalled = isPackageInstalled,
                     shouldStopQueryForAWhile = {
                         Logger.info(
-                            methodTag,"Will skip broker discovery for the next 60 minutes."
+                            methodTag,"Will skip broker discovery via IPC and fall back to AccountManager " +
+                                    "for the next 60 minutes."
                         )
                         cache.setShouldUseAccountManagerForTheNextMilliseconds(TimeUnit.MINUTES.toMillis(60))
                     }
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -30,13 +30,16 @@ import com.microsoft.identity.common.internal.broker.PackageHelper
 import com.microsoft.identity.common.internal.broker.ipc.BrokerOperationBundle
 import com.microsoft.identity.common.internal.broker.ipc.ContentProviderStrategy
 import com.microsoft.identity.common.internal.broker.ipc.IIpcStrategy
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
```diff
@@ -60,14 +62,20 @@ class ActiveBrokerCache
...
 
         /**
-         * The Lock for all [ActiveBrokerCache] instances used by the broker code.
+         * The Mutex for all [ActiveBrokerCache] instances used by the broker code.
+         * (As of May 24, 2023... Kotlin has yet to officially support ReadWriteMutex.
+         *  I don't think it's worth implementing our own (for now).
+         *  If we eventually are seeing a perf hit, sure...)
          **/
-        private val sBrokerSideLock = ReentrantReadWriteLock()
+        private val sBrokerSideLock = Mutex()
```
p3dr0rv, line 66: where is this used? same with sSdkSideLock
I believe we are using the provided Mutex

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
```diff
@@ -106,35 +114,50 @@ class ActiveBrokerCache
     internal var inMemoryCachedValue: BrokerData? = null
 
     override fun getCachedActiveBroker(): BrokerData? {
-        lock.read {
-            if (inMemoryCachedValue != null) {
-                return inMemoryCachedValue
+        return runBlocking {
+            lock.lock()
```
p3dr0rv: just a question, any reason you preferred to try catch vs mutex.withLock ?

rpdome: I wasn't aware of that one 😆.

That seems to be cleaner, so i'll switch.

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
rpdome: there's getBrokerMetadataStoreOnBrokerSide() and getBrokerMetadataStoreOnSdkSide() - both static methods.

We want to make sure a different set of underlying database, and therefore locks, are used for broker code and sdk code.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -86,31 +90,42 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
-                try {
-                    val result = ipcStrategy.communicateToBroker(operationBundle)
-                    return extractResult(result)
-                } catch (t: Throwable) {
-                    if (t is BrokerCommunicationException &&
-                        BrokerCommunicationException.Category.OPERATION_NOT_SUPPORTED_ON_SERVER_SIDE == t.category) {
-                        Logger.info(methodTag, "Tried broker discovery on ${candidate}. It doesn't support the operation")
-                    } else {
-                        Logger.error(methodTag, "Tried broker discovery on ${candidate}, get an error", t)
+            return runBlocking(dispatcher) {
```
p3dr0rv, line 102: is this runBlocking necessary?

In the caller function (getActiveBroker) we use runBlocking on the main thread which blocks the main thread and starts a coroutine in the main thread, now in this internal function we block again the main thread and start a IO coroutine.

SO my point is that if we stop the main thread from the upper function, we do not need to stop it here, right.
we are already inside of a coroutine scope so we ca just launch the IO coroutine. 
launch(dispatcher) { //my code } and probably annotate the function suspended.

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
p3dr0rv: Ok but if one is in the Sdkprocess and the other is in the auth process, Each process is in a separate address space. 
So one lock will be ok, they will never collide, right? because they exist in different processes

 this is more for readability and precautions, or there is something I am missing

p3dr0rv (approved):
LGTM, let a few question to test my recently acquired knowledge on coroutines XD

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCache.kt
rpdome: technically yes... but nothing really guarantees that the sdk storage will never be used in broker process (and vice versa). 

therefore, having separate lock for each storage space is safer - in the future.

so yeah... precaution.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
rpdome: coroutine's mutex has to be invoked in a suspend function. that's why i wrapped it with runblocking.

rpdome: what I wanted to achieve is to provide synchronization across threads AND coroutines - and I picked coroutine's Mutex for that. 

Coroutine's mutex needs to be invoked _inside_ a suspend function, so i turned the whole block into one by wrapping it with runBlocking().

Not sure if that clarifies. Please let me know @p3dr0rv

rpdome: @p3dr0rv In this code, I have a method that do something like this.
```
A() {
    readCache()
    runBlocking {
         doWorkAsync()
    }
}
```

I originally put a java-thread-level-lock over A(). 

but let's say i invoke something like this (which is the new test case I added).

```
runBlocking {
       launch {
             A(1)
       }
       launch {
             A(2)
       }
}
```

now both launch blocks will be executed on the main thread. let's say A(1) is invoked first. readCache() is done... then once it hits its runBlocking() - what I observed is is that A(1) will put doWorkAsync() at the end of the coroutine queue (can't find the official name here), and temporarily suspends itself.

then A(2) will be executed "by a different coroutine, from the very same thread". 

If the lock is re-entrant (read: if you're in the same thread as the lock holder, feel free to go through), then A(2) would just breeze through readCache() like nothing happened....which is not an intended behavior.

If I changed the lock to a non-reentrant one (e.g. semaphore) it seems like doWorkAsync() will never be triggered -- or resume after runBlocking() is done if you happen to use a different dispatcher for runBlocking itself. (since the lock is not yet released).

melissaahn (approved):

PR #2053, opened 2023-05-30
Add android 14 check for skipping strong box isolation in pop token

somalaya:
Created the PR on Fadi's behalf

Issue description  :  When a device is newly Intune enrolled in A14, the user is able to login to Defender, but unable to retrieve the pop access token. As a fallback, Defender allows onboarding without pop token (which is why you were able to onboard fully to Defender). However, in absence of the correct token Defender backend does not acknowledge heartbeat calls we send every few hours to maintain device compliance. Heartbeat failure subsequently fails the signals being sent to Intune backend marking the device non-compliant (Device_Threat_Health_Not_Reported reported by Intune).

This is an issue especially when policies are setup to make the device noncompliant if Defender is not setup (this is what Microsoft has setup internally).

Bug link : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2574078

Fix : There is  new IAR requirement introduced for Android 14 that is failing for Strong box isolation to work as expected. Skipping strong box until a fix for this is made. We are still investigating on what could be the long term fix. But until then, we will go ahead with this temporary fix.

Testing : Shared the new broker build with CP team and they tested that defender is working as expected after taking this fix.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
@@ -82,6 +82,13 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
     public static final String STRONG_BOX_UNAVAILABLE_EXCEPTION = "StrongBoxUnavailableException";
 
     /**
+     * Seeing this on android 14, we think it's being caused by the new IAR requirement
+     *
+     // https://android.googlesource.com/platform/compatibility/cdd/+/e2fee2f/9_security-model/9_11_keys-and-credentials.md
+     */
+    public static final String NEGATIVE_THOUSAND_INTERNAL_ERROR = "internal Keystore code: -1000";
+
+    /**
      * Error message from underlying KeyStore that an attestation certificate could not be
      * generated, typically due to lack of API support via {@link KeyGenParameterSpec.Builder#setAttestationChallenge(byte[])}.
      */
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -203,9 +210,15 @@ private KeyPair generateNewRsaKeyPair(@androidx.annotation.NonNull final Context
...
                         trySetAttestationChallenge = false;
 
+                        continue;
+                    } else if (tryStrongBox && Build.VERSION.RELEASE_OR_CODENAME.equals("UpsideDownCake") && isNegativeInternalError(e.getCause())) {
+                        // Android 14 specific error where strong box is failing, most likely because of IAR requirement in android 14
+                        // https://android.googlesource.com/platform/compatibility/cdd/+/e2fee2f/9_security-model/9_11_keys-and-credentials.md
+                        // Had to check code name, as android 14 device in beta seems to still show 33 as SDK int
+                        Logger.error(TAG, "Android 14 Internal Key store error with strongbox. Skipping strongbox", e);
+                        tryStrongBox = false;
                         continue;
```
iamgusain, line 221: Let's add work-item link, to investigate the root cause.

iamgusain (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
somalaya: Sure, done

rpdome (approved):

rpdome (approved):

fadidurah:
Thanks for following up on this Swomya!!!

PR #2057, opened 2023-06-01
Wire new Broker Discovery Client into MSAL - still disabled by default

rpdome:
MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1835
Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2315

## Changes

1. Created `BrokerDiscoveryClientFactory`, which provides either 
     - BrokerDiscoveryClient : The new Broker Discovery Mechanism
     - LegacyBrokerDiscoveryClient: The current mechanism based on AccountManager.
2. By default, `BrokerDiscoveryClientFactory` would return `LegacyBrokerDiscoveryClient`. If anyone wants to test the new mechanism (not yet wired on broker as of now), they could invoke `BrokerDiscoveryClientFactory.setNewBrokerDiscoveryEnabled(true)`
3. Make changes to the `BrokerMsalController`'s constructor.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
@@ -1,0 +1,79 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.internal.activebrokerdiscovery
+
+import android.content.Context
+import com.microsoft.identity.common.java.interfaces.IPlatformComponents
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+/**
+ * A class for initializing a new [IBrokerDiscoveryClient] object.
+ **/
+class BrokerDiscoveryClientFactory {
+    companion object {
+        @Volatile
+        private var IS_NEW_DISCOVERY_ENABLED = false
+
+        @Volatile
+        private var instance: IBrokerDiscoveryClient? = null
+
+        private val lock = Mutex()
+
+        /**
+         * If set to true, the new Broker discovery mechanism will be enabled.
+         * This is currently turned off by default - until we're ready to ship the feature.
+         **/
+        @JvmStatic
+        fun setNewBrokerDiscoveryEnabled(isEnabled: Boolean){
+            if (isEnabled != IS_NEW_DISCOVERY_ENABLED) {
+                instance = null
+                IS_NEW_DISCOVERY_ENABLED = isEnabled
+            }
+        }
+
+        /**
+         * Initializes a new [IBrokerDiscoveryClient] object.
+         **/
+        @JvmStatic
+        fun getInstance(context: Context,
+                        platformComponents: IPlatformComponents) : IBrokerDiscoveryClient{
+            if (instance == null){
+                runBlocking {
+                    lock.withLock {
+                        if (instance == null) {
+                            instance = if (IS_NEW_DISCOVERY_ENABLED) {
+                                BrokerDiscoveryClient(context, platformComponents)
+                            } else {
+                                LegacyBrokerDiscoveryClient(context)
+                            }
+                        }
+                    }
+                }
+            }
+            return instance!!
+        }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/LegacyBrokerDiscoveryClient.kt
```diff
@@ -0,0 +1,91 @@
...
+         * Known issue: When we're in an AccountManager callback (Especially on older Android devices, i.e. Android 10)
+         * Android For Work throws a SecurityException when we're calling AccountManager.getAuthenticatorTypes()
+         *
+         * E.g. Company Portal main process can call this freely, broker process can call this freely,
+         * but once running in AccountManager on the work profile (user != 0),
+         * apparently sometimes it tries to get accounts from user 0 (personal profile) and fails.
+         *
+         * @return Package Name of the broker
+         */
+        private fun getActiveBrokerPackageName(context: Context): String? {
```
rpdome, line 52: Taken from BrokerValidator... because i'm planning to kill the method there. (All the method there should be for "validating" only).

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
```diff
@@ -0,0 +1,79 @@
...
+        /**
+         * Initializes a new [IBrokerDiscoveryClient] object.
+         **/
+        @JvmStatic
+        fun getInstance(context: Context,
+                        platformComponents: IPlatformComponents) : IBrokerDiscoveryClient{
+            if (instance == null){
+                runBlocking {
+                    lock.withLock {
+                        if (instance == null) {
```
melissaahn, line 69: why do we need to null check instance both before runBlocking and inside?

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
rpdome: https://www.baeldung.com/java-singleton-double-checked-locking

shahzaibj: Does `runBlocking` work the same as `synchronized`?

rpdome: synchronize only protects in thread level. kotlin's mutex protects at coroutine level, but can only be invoked inside a coroutine (which is why i had to wrap it with a runBlocking)

It's possible for the code to be triggered twice by 2 coroutine belonging to the same thread. 

```
launch {
      taskA()
      taskA()
}
```

and
```
fun taskA(){
      runBlocking() {
            doB()
      }
}
```

in this case, we're invoking taskA() twice in the same coroutine-  i'll call it taskA1 and taskA2
however, when taskA1 is invoked, it will try to invoke its own runBlocking.
what would happen here is that taskA1's runBlocking will be scheduled, and then taskA2 will be invoked.

the order of operation will be like
taskA1() -> taskA2 -> taskA1's runBlocking -> taskA2's runBlocking

meaning that if i put a synchronize over taskA()'s runblocking - it will be walked pass right through by taskA2 (because taskA1 and taskA2 are invoked on the same thread).

so here i pretty much assume that i don't know what the underlying code of both constructor looks like, and assume that they might contain one of the runBlocking.

I might be overengineering it in this case - and this might never happen here. let me know if you think i should revert back to synchronized.

rpdome: no.

runBlocking links between the synchronous world to the coroutine world.

synchronized = java's thread level lock. the equivalent of that would be kotlin's mutex (which needs to be used in a coroutine only... this is why i wrapped it with a runBlocking())

rpdome: The reason i'm doing is that... consider this scnenario

```Kotlin
   fun testCoroutine(){
        runBlocking {
            launch {
                taskA(1)
            }
            launch {
                taskA(2)
            }
        }
    }

    val lock = ReentrantLock() // equivalent to synchronized in java.

    fun taskA(i: Int){
        lock.withLock {
            runBlocking {
                doB(i)
            }
        }
    }

    fun doB(i: Int){
        val x = i
    }

```

In this case, both taskA() will be executed on the same thread.

The order of execution will look like this

taskA1 -> taskA1's runBlocking (stop) -> taskA2 -> taskA2's runBlocking (stop) -> doB(1) -> doB(2)

doB() will be executed twice regardless of the lock - this is because reentrant lock (or synchronized) allows same thread to enter.

I wire up the lock this way to prevent this edge case (pretty much does NOT assume what's the calling pattern of the upper and the layer look like).

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
```diff
@@ -0,0 +1,79 @@
...
+ **/
+class BrokerDiscoveryClientFactory {
+    companion object {
+        @Volatile
+        private var IS_NEW_DISCOVERY_ENABLED = false
+
+        @Volatile
+        private var instance: IBrokerDiscoveryClient? = null
+
+        private val lock = Mutex()
```
shahzaibj: Comments/javadoc on these would be helpful

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
```diff
@@ -0,0 +1,79 @@
...
+
+        /**
+         * If set to true, the new Broker discovery mechanism will be enabled.
+         * This is currently turned off by default - until we're ready to ship the feature.
+         **/
+        @JvmStatic
+        fun setNewBrokerDiscoveryEnabled(isEnabled: Boolean){
+            if (isEnabled != IS_NEW_DISCOVERY_ENABLED) {
```
shahzaibj, line 54: I'm not sure I understand what we're trying to do here. Why can't we just set the flag?

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/LegacyBrokerDiscoveryClient.kt
```diff
@@ -0,0 +1,91 @@
...
+            BrokerData.getKnownBrokerApps().firstOrNull {
+                it.packageName == activeBrokerPkgName
+            }
+        } catch (e: Exception) {
+            Logger.error(methodTag, "Failed to get active broker", e)
+            null
+        }
+    }
+
+}
```
shahzaibj: nit: new line ending

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
rpdome: if the flag is changed, then i would need to wipe the existing singleton (which would belong to a different type)

shahzaibj (approved):

PR #2058, opened 2023-06-01
Clear cache of access tokens with an old application identifier field

melissaahn:
### Summary
_Note: A few unit tests are expected to fail until another PR which updates variables (https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2059) is merged into this one. I separated it out in order to improve the readability of this PR._

When created from the broker, access token records have an Application Identifier string field attached to them which consists of the calling app's package name and signing certificate hash:
`<Package name>/<signing certificate hash>`

The base-64 signing certificate hash is currently generated using SHA-1. As we're currently making the effort to migrate our uses of SHA-1 to SHA-512, I created a PR for the broker to now attach a SHA-512 hash to the broker parameter Signature field, replacing the SHA-1 hash, for both interactively and silently acquiring a token.

### Why this PR is also needed
Fortunately, the user shouldn't have to login again upon updating their broker app; there will be an access token cache miss, but the unaffected refresh token records will be able to reacquire an access token with the updated SHA-512 application identifier field. 
However, the access token with a SHA-1 application identifier will remain in the cache indefinitely, as it won't be explicitly removed if the broker is only searching for app identifiers including 512 hashes. 

This PR addresses this issue by removing access tokens with a SHA-1 application identifier upon instantiation of the SharedPreferencesAccountCredentialCache object (and the corresponding MemoryCache version). This ensures that the cache is cleaned of access tokens with an "old" identifier before any getCredentials() method is called. 

This logic should (in theory) run only once upon a user updating to the version which includes said logic. I added a unit test which mimics this scenario.

Related PRs:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2059 : Will be merged into this one.
- https://github.com/AzureAD/ad-accounts-for-android/pull/2317 : The mentioned broker PR, which will be merged _after_ this one.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
@@ -49,6 +49,7 @@ import lombok.NonNull;
 public abstract class AbstractAccountCredentialCache implements IAccountCredentialCache {
 
     private static final String TAG = AbstractAccountCredentialCache.class.getSimpleName();
+    public static final String SHA1_APPLICATION_IDENTIFIER_ACCESS_TOKEN_CLEARED = "sha1-cleared";
     private static final String NEW_LINE = "\n";
 
     @Nullable
@@ -274,25 +275,6 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
         }
 
         return matchingCredentials;
-    }
-
-    /**
-     * The application identifier field for access token records previously included a SHA-1 signing certificate hash.
-     * This method clears such tokens, as the application identifier should now contain a SHA-512 signing certificate hash.
-     */
-    public void clearSha1ApplicationIdentifierAccessTokens() {
-        final String methodTag = TAG + ":clearSha1ApplicationIdentifierAccessTokens";
-        for (final Credential credential : getCredentials()) {
-            if (credential instanceof AccessTokenRecord) {
-                final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
-                if (tokenAppIdentifier != null
-                        && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
-                    removeCredential(accessToken);
-                    Logger.info(methodTag, "Removed old access token with app identifier containing SHA-1. A new access token should be re-acquired with a SHA-512 app identifier.");
-                }
-            }
-        }
     }
 
     /**
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
@@ -279,22 +279,23 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
 
     /**
      * The application identifier field for access token records previously included a SHA-1 signing certificate hash.
-     * This method identifies such tokens, as the application identifier should now contain a SHA-512 signing certificate hash.
-     * @param credential token record.
-     * @return true if an access token containing a SHA-1 app identifier hash; false otherwise.
+     * This method first checks if it has been run before, and if not, removes such tokens, as application identifiers should now contain SHA-512 signing certificate hashes.
      */
-    public static boolean isSha1ApplicationIdentifierAccessToken(@NonNull final Credential credential) {
-        final String methodTag = TAG + ":isSha1ApplicationIdentifierAccessToken";
-        if (credential instanceof AccessTokenRecord) {
-            final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-            final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
-            if (tokenAppIdentifier != null
-                    && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
-                Logger.info(methodTag, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier.");
-                return true;
+    protected void removeSha1ApplicationIdentifierAccessTokensIfNeeded() {
+        if (!isSha1Cleared()) {
+            for (final Credential credential : getCredentials()) {
+                if (credential instanceof AccessTokenRecord) {
+                    final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
+                    final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
+                    if (tokenAppIdentifier != null
+                            && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
+                        Logger.info(TAG, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier.");
+                        removeCredential(credential);
+                    }
+                }
             }
+            saveSha1ClearedFlag();
         }
-        return false;
     }
 
     /**
@@ -306,6 +307,17 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
         final String[] appIdentifierArr = applicationIdentifier.split("/", 2);
         return (appIdentifierArr.length > 1 && appIdentifierArr[1].length() == 28);
     }
+
+    /**
+     * Tells if access tokens with SHA-1 app identifiers have been cleared yet.
+     * @return true if cleared; false otherwise.
+     */
+    abstract protected boolean isSha1Cleared();
+
+    /**
+     * Saves the flag indicating that access tokens with SHA-1 app identifiers have been cleared.
+     */
+    abstract protected void saveSha1ClearedFlag();
 
     /**
      * Examines the intersections of the provided targets (scopes).
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
@@ -294,6 +294,7 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
      * This method first checks if it has been run before, and if not, removes such tokens, as application identifiers should now contain SHA-512 signing certificate hashes.
      */
     public void removeSha1ApplicationIdentifierAccessTokensIfNeeded() {
+        final String methodTag = TAG + ":removeSha1ApplicationIdentifierAccessTokensIfNeeded";
         if (!isSha1Cleared()) {
             for (final Credential credential : getCredentials()) {
                 if (credential instanceof AccessTokenRecord) {
@@ -301,7 +302,7 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
                     final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
                     if (tokenAppIdentifier != null
                             && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
-                        Logger.info(TAG, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier.");
+                        Logger.info(methodTag, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier.");
                         removeCredential(credential);
                     }
                 }
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
@@ -199,14 +199,7 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
             if (mustMatchOnApplicationIdentifier) {
                 if (credential instanceof AccessTokenRecord) {
                     final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                    //Check here for an application identifier with a SHA-1 hash.
-                    final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
-                    final String[] appIdentifierArr = tokenAppIdentifier.split("/", 2);
-                    if (appIdentifierArr.length > 1
-                            && appIdentifierArr[1].length() == 28) {
-                        removeCredential(accessToken);
-                    }
-                    matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(applicationIdentifier, tokenAppIdentifier);
+                    matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(applicationIdentifier, accessToken.getApplicationIdentifier());
                 } else {
                     Logger.verbose(TAG, "Query specified applicationIdentifier match, but credential type does not have application identifier");
                 }
@@ -281,6 +274,35 @@ public abstract class AbstractAccountCredentialCache implements IAccountCredenti
         }
 
         return matchingCredentials;
+    }
+
+    /**
+     * The application identifier field for access token records previously included a SHA-1 signing certificate hash.
+     * This method clears such tokens, as the application identifier should now contain a SHA-512 signing certificate hash.
+     */
+    public void clearSha1ApplicationIdentifierAccessTokens() {
+        final String methodTag = TAG + ":clearSha1ApplicationIdentifierAccessTokens";
+        for (final Credential credential : getCredentials()) {
+            if (credential instanceof AccessTokenRecord) {
+                final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
+                final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
+                if (tokenAppIdentifier != null
+                        && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
+                    removeCredential(accessToken);
+                    Logger.info(methodTag, "Removed old access token with app identifier containing SHA-1. A new access token should be re-acquired with a SHA-512 app identifier.");
+                }
+            }
+        }
+    }
+
+    /**
+     * Can tell if an application identifier string (<package name>/<signing certificate hash>) contains a SHA-1 hash.
+     * @param applicationIdentifier application identifier field string
+     * @return true if contains SHA-1 hash; false otherwise.
+     */
+    static boolean applicationIdentifierContainsSha1(@NonNull final String applicationIdentifier) {
+        final String[] appIdentifierArr = applicationIdentifier.split("/", 2);
+        return (appIdentifierArr.length > 1 && appIdentifierArr[1].length() == 28);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -276,6 +276,35 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
         return matchingCredentials;
     }
 
+    /**
+     * The application identifier field for access token records previously included a SHA-1 signing certificate hash.
+     * This method clears such tokens, as the application identifier should now contain a SHA-512 signing certificate hash.
+     */
+    public void clearSha1ApplicationIdentifierAccessTokens() {
+        final String methodTag = TAG + ":clearSha1ApplicationIdentifierAccessTokens";
+        for (final Credential credential : getCredentials()) {
```
rpdome: how much perf does this cost (given getCredentials is an I/O operation)

OneAuth is pretty sensitive about it and i'm wondering if we should add a flag (to the storage) to mark if the "clean sha1" operation is already done.

melissaahn: Gave your suggested implementation a go. let me know what you think

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -100,6 +100,14 @@ public SharedPreferencesAccountCredentialCache(
...
         mSharedPreferencesFileManager = sharedPreferencesFileManager;
         mCacheValueDelegate = accountCacheValueDelegate;
+        if (mSharedPreferencesFileManager.get(SHA1_APPLICATION_IDENTIFIER_ACCESS_TOKEN_CLEARED) == null) {
+            for (final Credential credential : getCredentials()) {
+                if (isSha1ApplicationIdentifierAccessToken(credential)) {
+                    removeCredential(credential);
+                }
+            }
+            mSharedPreferencesFileManager.put(SHA1_APPLICATION_IDENTIFIER_ACCESS_TOKEN_CLEARED, String.valueOf(true));
+        }
```
fadidurah: We don't get another token here, we just remove the existing one, right? So I'm assuming the new token will be acquired through a silent request after the user updates the broker-hosting app on their device?

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -276,6 +277,36 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
...
+     * @return true if an access token containing a SHA-1 app identifier hash; false otherwise.
+     */
+    public static boolean isSha1ApplicationIdentifierAccessToken(@NonNull final Credential credential) {
+        final String methodTag = TAG + ":isSha1ApplicationIdentifierAccessToken";
+        if (credential instanceof AccessTokenRecord) {
+            final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
+            final String tokenAppIdentifier = accessToken.getApplicationIdentifier();
+            if (tokenAppIdentifier != null
+                    && applicationIdentifierContainsSha1(tokenAppIdentifier)) {
+                Logger.info(methodTag, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier.");
```
nickbopp: >Logger.info(methodTag, "Identified old access token with app identifier containing SHA-1. This token shall be removed, and a new access token should be re-acquired with a SHA-512 app identifier."); [](http://example.com/codeflow?start=16&length=197)

This log line might be better one stack frame further up, where the decision is actually made to do the delete.  It probably doesn't matter much, but this function is really just about determining if something is sha1, while the caller is the one that decides sha1 == delete #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
nickbopp: construction of a sharedpreferences kicks off a bakcground thread to load the xml file into memory.  once it's loaded, any get() call is just reading from an in-memory Map<String, String>, but while that async operation is still going on, a call to get() will block waiting for it to finish.  It needs to load the whole file (every account/credential) into memory before any of the values are accessible.

I think as written, this could add some significant delays to construction path, where construction here now requires that async load to complete in order to check the flag.  For OneAuth, since we're moving to the separate implementation, I'm less concerned here. Is this path performance critical for the broker or MSAL? If a token request is going to happen very soon after this, then I don't think it makes a difference, since you'll just pay for the wait microseconds later.  In Oneauth, there may be no token request coming for many minutes after initialization, but we might be unique in that. #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -736,4 +747,11 @@ private static boolean isCredential(@NonNull String cacheKey) {
         return isCredential;
     }
 
+    private boolean loadSha1ClearedFlag() {
```
nickbopp: >loadSha1ClearedFlag [](http://example.com/codeflow?start=20&length=19)

isSha1Cleared() is more clear I think.  load is a bit ambiguous in that the load result could succeed but the value is false. #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -332,6 +339,10 @@ public boolean test(String value) {
                 // Remove the entry and return null...
                 Logger.warn(methodTag, "The returned Credential was uninitialized. Removing...");
                 mSharedPreferencesFileManager.remove(cacheKey);
+            } else if (AccessTokenRecord.class == clazz
+                    && !mSha1Cleared
+                    && isSha1ApplicationIdentifierAccessToken(credential)) {
+                    mSharedPreferencesFileManager.remove(cacheKey);
```
nickbopp: >mSharedPreferencesFileManager.remove(cacheKey); [](http://example.com/codeflow?start=20&length=47)

I don't remember the iterator rules for safety in java.  Is it safe to remove an item from a collection while iterating over it -- or does it need to be a separate step? #Closed

nickbopp: I guess we're doing it a few lines above so it can't be that unsafe :) Consider it closed.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -332,6 +339,10 @@ public boolean test(String value) {
                 // Remove the entry and return null...
                 Logger.warn(methodTag, "The returned Credential was uninitialized. Removing...");
                 mSharedPreferencesFileManager.remove(cacheKey);
+            } else if (AccessTokenRecord.class == clazz
+                    && !mSha1Cleared
```
nickbopp: >mSha1Cleared [](http://example.com/codeflow?start=24&length=12)

Put this condition first since a bool check will be faster (and more often false) than the class equality check. #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -86,6 +88,7 @@ private void load() {
 
         synchronized (mCacheLock) {
             try {
+                mSha1Cleared = loadSha1ClearedFlag();
```
nickbopp: >mSha1Cleared = loadSha1ClearedFlag(); [](http://example.com/codeflow?start=16&length=37)

to try to prevent keeping more state around in this class that needs reasoning about, could this move to a local boolean, and a parameter 'boolean cleanUpSha1AccessTokens' on `loadCredentialWithKeys`? #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -94,6 +97,10 @@ private void load() {
                 Logger.error(methodTag, "Failed to load initial accounts or credentials from SharedPreferences", t);
             } finally {
                 mLoaded = true;
+                if (!mSha1Cleared) {
+                    saveSha1ClearedFlag();
+                    mSha1Cleared = true;
+                }
```
nickbopp: Originally I was going to suggest removing the conditional and just saving() every time, but it does force a full I/O write of the whole sharedpreferences to do that, so better to keep this conditional.  With that said, with the advice above, you won't need to update mSha1Cleared anymore afterwards.

I do think this block should only be inside the try.  If we Throw for some reason during the load, I don't expect the write to be any more successful.  And if we're throwing on the save() operation, we have no handler here and likely crash because it will get to the top level of our thread. #Closed

nickbopp (changes requested):
:clock1:

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -332,6 +339,10 @@ public boolean test(String value) {
                 // Remove the entry and return null...
                 Logger.warn(methodTag, "The returned Credential was uninitialized. Removing...");
                 mSharedPreferencesFileManager.remove(cacheKey);
+            } else if (AccessTokenRecord.class == clazz
+                    && !mSha1Cleared
+                    && isSha1ApplicationIdentifierAccessToken(credential)) {
+                    mSharedPreferencesFileManager.remove(cacheKey);
             }
```
nickbopp: For the sake of simplicity, I don't think it would be a perf problem to implement this load->cleanup->save logic onc as something like `cleanUpSha1CredentialsIfNeeded` in AbstractAccountCredentialCache using the available getCredentials() abstraction.  SPACC could call that function in its constructor, basically where you have it in the block now (assuming we decide its okay to do it then for perf).  And SPACCWMC could call it as the last line of its load() method that's running synchronously on a background thread. #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: Going to resolve this comment because it's referring to a different conditional that I had before I did some more refactoring, and that might be confusing to read in the future.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
melissaahn: Yep!

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: Looking at how `mSharedPreferencesFileManager.getAllFilteredByKey` is implemented, the iterator does refer to a Map that's directly returned from shared preferences, so I'm not sure either if this is actually safe. Though it doesn't really apply to this particular PR anymore because I removed the referenced line. 
I suppose if the existing remove() call hasn't caused any issues so far, seems fine for now?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: I added a method `removeSha1ApplicationIdentifierAccessTokensIfNeeded()` in AbstractAccountCredentialCache as well as two abstract methods, `isSha1Cleared` and `saveSha1ClearedFlag`. 
For SPACCWMC, I did initially try to put the `removeSha1ApplicationIdentifierAccessTokensIfNeeded()` call in its own thread at the end of the try block in load(), but this actually caused my unit test to fail. It turns out that the `getCredential()` call in the unit test, which was made after the cache initialization, was running before the `remove ...` call could be made. The `remove ...` call does need to be made before any other getCredential() call (excluding the call within `remove ... ` of course). So I copied over `mLoaded = true` and `mCacheLock.notifyAll()` to run right before `remove ...`, and encased those two lines in the finally block in `if (!mLoaded)`. 
lemme know what you think.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
melissaahn: I'll consult my team and come back to this comment. It doesn't seem to me that there will always be a token request after the cache is initialized, so this is definitely worth considering.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
nickbopp: This is a great catch.  I forgot that implementing this in terms of getCredentials() at the high level would have these locking problems since they're inherently public methods that need to guard to wait for the load.  This solution works for me.  One more small suggestion now that we're doing things this way: can you get rid of the finally block and instead just have the duplicate mLoaded=true/mCacheLock.notifyall() inside the catch block? That should cover both the success and exception case.  Now that we need to handle success specially, the finally block with an if makes less sense than just putting it in the catch.

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -736,4 +740,26 @@ private static boolean isCredential(@NonNull String cacheKey) {
...
     }
 
+    /**
+     * Tells if access tokens with SHA-1 app identifiers have been cleared yet.
+     * @return true if cleared; false otherwise.
+     */
+    @Override
+    protected boolean isSha1Cleared() {
+        synchronized (mCacheLock) {
+            waitForInitialLoad();
```
nickbopp: These don't actually need to be guarded in this way -- the mCacheLock is protecting specifically the 
mCachedAccountRecordsWithKeys and mCachedCredentialsWithKeys structures.  It's safe to go touch this mSharedPreferencesFileManager without them.  I think this guarding difference was probably the cause for this becoming an abstract method, so it can probable have a single implementation back in AbstractAccount* now as well. #Closed

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
melissaahn: Coming back to this... I got some advice from @rpdome and learned two things:

1. It really doesn't matter when the SHA1 app identifier access token is removed, because no method other than mine is going to query for it. And as far as I know, outside of unit tests, there aren't cases where we rely on the size of getCredentials() directly. So, your original suggestion of running the `removeSha1ApplicationIdentifierAccessTokensIfNeeded()` in a separate thread is actually totally fine, and I updated both SPACC and SPACCWMC to now do this. The unit test was updated to just test on `removeSha1ApplicationIdentifierAccessTokensIfNeeded()` now, (since I wasn't able to find a good way to test the method in the constructor due to the threads) but I think they're just as effective.
2. Dome let me know that there's a `OnUpgradeReceiver` class in broker, which, as the name suggests, is meant to run logic upon the app package being updated. So this gives the opportunity to just run `removeSha1ApplicationIdentifierAccessTokensIfNeeded()` once per update, and the calls in the cache constructors can be removed. This all being said, I think I still want to merge in this PR for the sake of getting rid of the SHA1 usage here, and then create a fresh PR with this new implementation in the next week or so (I promise!).

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: welp... ended up keeping the finally block due to going with your original suggestion, as mentioned in my other comment. 😅
commit with new changes: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/4b24755cb9168aa44753751b7b993bf65e2f585f

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/4b24755cb9168aa44753751b7b993bf65e2f585f

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -90,6 +86,7 @@ private void load() {
                 Logger.info(methodTag, "Loaded " + mCachedAccountRecordsWithKeys.size() + " AccountRecords");
                 mCachedCredentialsWithKeys = loadCredentialsWithKeys();
                 Logger.info(methodTag, "Loaded " + mCachedCredentialsWithKeys.size() + " Credentials");
+                new Thread(() -> removeSha1ApplicationIdentifierAccessTokensIfNeeded()).start();
```
nickbopp: >new Thread(() -> removeSha1ApplicationIdentifierAccessTokensIfNeeded()).start(); [](http://example.com/codeflow?start=16&length=80)

Since this is already its own thread, there's no need to take the perf penalty of spinning up another single-use thread.  We can just call this method here.

I would probably take the Thread function on line 77 and instead turn that into a method called 'initialize()' that does load() and then removeSha1*().  That should have a general catch around it so we don't risk remove throwing and crashing us with an uncaught top level exception.

I suggest doing it that way because the load() isn't dependent on the cleanup to finish given the requirements you've set forth, and we can let other apps come in and make reads even while we're doing this work of iterating and removing. #Closed

nickbopp (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
melissaahn: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2058/commits/3e147324a108f039db78dfbdbbc2581f87b44919
Thanks Nick!

PR #2060, opened 2023-06-06
make change to support mock broker election app

rpdome:
to enable https://github.com/AzureAD/ad-accounts-for-android/pull/2322

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
@@ -28,6 +28,7 @@ import android.os.Build;
 
 import androidx.annotation.Nullable;
 
+import com.microsoft.identity.common.BuildConfig;
 import com.microsoft.identity.common.java.broker.IBrokerAccount;
 import com.microsoft.identity.common.logging.Logger;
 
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
```diff
@@ -123,6 +126,17 @@ public static AndroidBrokerAccount create(@NonNull final AccountManager accountM
...
+            accountManager.setAccountVisibility(
+                    account,
+                    MOCK_AUTH_APP_PACKAGE_NAME,
+                    AccountManager.VISIBILITY_VISIBLE
+            );
+            accountManager.setAccountVisibility(
+                    account,
+                    MOCK_CP_PACKAGE_NAME,
+                    AccountManager.VISIBILITY_VISIBLE
+            );
```
fadidurah: Do we not need LTW here as well?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/AndroidBrokerAccount.java
rpdome: No. LTW won't be shipping with account manager.

In fact... this is probably not needed. Not even broker host. let me remove it.

melissaahn (approved):

PR #2064, opened 2023-06-07
Send client id as part of request bundle for getSsoToken Api

iamgusain:
### What
Adding Client_ID parameter to the request bundle for getSsoTokenAPI

### Why 
client_id is a required parameter for getSsoToken API for cases where account_name is not passed/known.

### How
Added Client_ID parameter to the request bundle for in the MsalBrokerRequestAdapter class

### Testing
Added a UT to verify that required parameter are passed to the broker.

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
+++ b/common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
@@ -1,0 +1,78 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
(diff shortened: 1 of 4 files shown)


shahzaibj:
How was this working earlier? Is this a recent regression?

iamgusain:
> How was this working earlier? Is this a recent regression?

The Api is not called by any partner yet. The integration work started now, where the issue is caught.

mohitc1 (approved):

common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class MsalBrokerRequestAdapterTests {
+
+    @Test
+    public void test_getRequestBundleForSsoToken() {
```
fadidurah, line 44: Do we have a null/empty check for clientId? Might be good to also have a test to show expected failure behaviour if we don't pass clientId.

fadidurah (approved):

common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
iamgusain: The check is on the Broker side, I will add test on the Broker side.

PR #2065, opened 2023-06-07
perform sighash validation when the value is read from cache

rpdome:
If the active broker is read from cache - we want to double check if the installed package name still has a valid signature hash before making a request.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -57,13 +57,14 @@ import java.util.concurrent.TimeUnit
  * @param ipcStrategy                       An [IIpcStrategy] to aggregate data with.
  * @param cache                             A local cache for storing active broker discovery results.
  * @param isPackageInstalled                a function to determine if any given broker app is installed.
- * @param lock                              a lock for preventing race condition of the operations in this class.
+ * @param isValidBroker                     a function to determine if the installed broker app contains a matching signature hash.
  **/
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -174,6 +175,16 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
             PackageHelper(context).
                 isPackageInstalledAndEnabled(brokerData.packageName)
+        },
+        isValidBroker = { brokerData ->
+            val methodTag = "$TAG:isValidBroker"
+            val installedHash = PackageHelper(context).getSha512SignatureForPackage(brokerData.packageName)
+            val isHashMatch = installedHash == brokerData.signatureHash
+            if (!isHashMatch) {
+                Logger.warn(methodTag, "Hash does not match for app ${brokerData.packageName}. " +
+                        "Expected: ${brokerData.signatureHash} but get: $installedHash")
```
somalaya: nit : "Expected: ${brokerData.signatureHash} but got: $installedHash"

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -235,4 +255,5 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
             return accountManagerResult
         }
     }
+
```
somalaya: nit : can remove this new line

somalaya (approved):

melissaahn (approved):

PR #2072, opened 2023-06-13
Forward request headers on redirects

matthewyu-ms:
What
Catch request headers in the original loadUrl request, and forward them to handleUrl so request headers are not dropped.

Why
On API31+ devices, OneAuth's MSA interactive SSO tests started failing consistently. Upon further investigation, this is because on API31+ (loaded with WebView 88+), Webview shipped a security fix that prevents request headers from being forwarded if the domain doesn't match the original URL. OneAuth-MSAL appends the refresh token in the request header, but in Fiddler traces, this is dropped when browser redirects from login.microsoftonline.com to login.live.com. This results in other telemetry information like correlation ID being dropped.

Reference: https://infosecwriteups.com/leakage-of-sensitive-data-through-android-webviews-3b0b86486a28

How
Capture request header in original loadURL request, and then in handleUrl, override the current request and append original request headers there.

Testing
Confirmed OneAuth's MSA interactive SSO tests are working again with this fix.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -258,7 +258,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
         // Old Chromium versions <88 did this behavior by default, but it was removed in more recent versions.
         // For now, reproduce this only for MSA, and consider adding more trusted ESTS endpoints in the future.
         final boolean urlIsTrustedToReceiveHeaders =  url.startsWith("https://login.live.com/");
-        boolean originalRequestHasHeaders = mRequestHeaders != null && !mRequestHeaders.isEmpty();
+        final boolean originalRequestHasHeaders = mRequestHeaders != null && !mRequestHeaders.isEmpty();
         return urlIsTrustedToReceiveHeaders && originalRequestHasHeaders;
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -257,7 +257,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
         // investigations.
         // Old Chromium versions <88 did this behavior by default, but it was removed in more recent versions.
         // For now, reproduce this only for MSA, and consider adding more trusted ESTS endpoints in the future.
-        boolean urlIsTrustedToReceiveHeaders =  url.startsWith("https://login.live.com/");
+        final boolean urlIsTrustedToReceiveHeaders =  url.startsWith("https://login.live.com/");
         boolean originalRequestHasHeaders = mRequestHeaders != null && !mRequestHeaders.isEmpty();
         return urlIsTrustedToReceiveHeaders && originalRequestHasHeaders;
     }
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -56,6 +56,7 @@ import com.microsoft.identity.common.logging.Logger;
 
 import java.net.URISyntaxException;
 import java.security.Principal;
+import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 
@@ -83,6 +84,8 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     private final String mRedirectUrl;
     private final CertBasedAuthFactory mCertBasedAuthFactory;
     private AbstractCertBasedAuthChallengeHandler mCertBasedAuthChallengeHandler;
+
+    private HashMap<String, String> mRequestHeaders;
 
     public AzureActiveDirectoryWebViewClient(@NonNull final Activity activity,
                                              @NonNull final IAuthorizationCompletionCallback completionCallback,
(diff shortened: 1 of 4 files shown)


nickbopp (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -182,7 +189,15 @@ private boolean handleUrl(final WebView view, final String url) {
             } else {
                 Logger.info(methodTag,"This maybe a valid URI, but no special handling for this mentioned URI, hence deferring to WebView for loading.");
                 processInvalidUrl(url);
-                return false;
+
+                // Request headers submitted in the initial loadUrl will no longer be propagated for redirects on API31+.
+                // Override redirect and pass original request headers here.
+                if (mRequestHeaders != null && !mRequestHeaders.isEmpty()) {
+                    view.loadUrl(url, mRequestHeaders);
```
iamgusain: do we need some additional checks to verify if it is safe to send the headers to the redirected url?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -182,7 +189,15 @@ private boolean handleUrl(final WebView view, final String url) {
             } else {
                 Logger.info(methodTag,"This maybe a valid URI, but no special handling for this mentioned URI, hence deferring to WebView for loading.");
                 processInvalidUrl(url);
-                return false;
+
+                // Request headers submitted in the initial loadUrl will no longer be propagated for redirects on API31+.
```
iamgusain: Since this is webview version related and not API31+, should we call that out in the comment.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
shahzaibj: We should definitely add some checks given that these headers contain RT. We can create an allow list of domains (Microsoft Identity URLs) for which headers should be appended

shahzaibj:
Curious, why in your testing was this specific to MSA? The same should be applicable to AAD too, right?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
shahzaibj: The allowed list can also be made such that it is only applicable to certain headers. For instance, the RT header is only appended if domains is part of allowed list, whereas the other headers can be appended regardless of the domain.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
nickbopp: Yes, I will change this.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
nickbopp: I'm validating behaviors on other platforms to understand what they're doing.

nickbopp:
> Curious, why in your testing was this specific to MSA? The same should be applicable to AAD too, right?

AcquireCredentialInteractively MSA is the only thing where SSO relies on the RT in a header being carried to a cross-host redirect.  This general problem does show up everywhere (e.g. correlationid getting dropped).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
nickbopp: On Windows, 302's are issued for Identity->Identity endpoints (e.g. Global ESTS -> MSA or UsGov ESTS -> Global ESTS).  Those are carrying all headers.

200's with a body of `document.location.replace` are being issued for ADFS cases, and headers are not being sent there.  I'm validating the behavior on Android now.

nickbopp: Actually, swapping between using WAM and using in-proc network calls is changing the UsGov -> Global /authorize redirect between a 200 with WAM and a 302 in-proc.  So I don't think that's a good gate.  I was trying to avoid a whitelist, but maybe that's the way.  For speed of this change, maybe let's just gate to `login.live.com` for now to scope things down.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -241,6 +251,17 @@ private boolean isWebCpUrl(@NonNull final String url) {
         return url.startsWith(AuthenticationConstants.Broker.BROWSER_EXT_WEB_CP);
     }
 
+    private boolean isHeaderForwardingRequiredUri(@NonNull final String url) {
+        // MSAL makes MSA requests first to login.microsoftonline.com, and then gets redirected to login.live.com.
+        // This drops all the headers, which can have credentials useful for SSO and correlationIds useful for
+        // investigations.
+        // Old Chromium versions <88 did this behavior by default, but it was removed in more recent versions.
+        // For now, reproduce this only for MSA, and consider adding more trusted ESTS endpoints in the future.
+        boolean urlIsTrustedToReceiveHeaders =  url.startsWith("https://login.live.com/");
```
shahzaibj:
```suggestion
        final boolean urlIsTrustedToReceiveHeaders =  url.startsWith("https://login.live.com/");
```

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -241,6 +251,17 @@ private boolean isWebCpUrl(@NonNull final String url) {
...
+        boolean originalRequestHasHeaders = mRequestHeaders != null && !mRequestHeaders.isEmpty();
```
shahzaibj:
```suggestion
        final boolean originalRequestHasHeaders = mRequestHeaders != null && !mRequestHeaders.isEmpty();
```

shahzaibj (approved):

iamgusain (approved):

PR #2073, opened 2023-06-13
Add method to return list of broker that supports account manager

rpdome:
Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2339

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
@@ -106,6 +106,15 @@ data class BrokerData(val packageName : String,
             }
         }
 
+        /**
+         * Returns true if the owner of the [Context] is a broker app
+         * which relies on AccountManager as a broker discovery mechanism.
+         * */
+        @JvmStatic
+        fun isAccountManagerSupported(context: Context): Boolean {
+            return accountManagerBrokers.contains(context.packageName)
+        }
+
         @JvmStatic
         val debugMicrosoftAuthenticator = BrokerData(
             AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME,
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -160,6 +169,18 @@ data class BrokerData(val packageName : String,
             AuthenticationConstants.Broker.LTW_APP_SHA512_DEBUG_SIGNATURE
         )
 
+        @JvmStatic
+        val accountManagerBrokers: Set<String> =
+            Collections.unmodifiableSet(object : HashSet<String>() {
```
shahzaibj, line 126: I believe this creates a new object on each invokation of the method? Could the (unmodifiable) list be declared as a constant in this file and then just return that constant?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
rpdome: This is a static constant (val), not a method.

const in kotlin means compile-time constant. val = runtime constant.

shahzaibj: Can we move it and all constants to the top of the file like we do in Java files (for easy discoverability)?

shahzaibj:
Add changelog

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -67,45 +67,6 @@ data class BrokerData(val packageName : String,
          */
         val sShouldTrustDebugBrokers = BuildConfig.DEBUG
 
-        /**
-         * Returns the list of known broker apps (which SDK should make requests to).
-         * see [sShouldTrustDebugBrokers] for more info regarding testing.
-         **/
-        fun getKnownBrokerApps() : Set<BrokerData> {
```
rpdome, line 74: move methods below constants/variables as suggested by shahzaib.
no other changes.

iamgusain (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -160,6 +121,18 @@ data class BrokerData(val packageName : String,
...
 
+        @JvmStatic
+        val accountManagerBrokers: Set<String> =
+            Collections.unmodifiableSet(object : HashSet<String>() {
+                init {
+                    add(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
+                    add(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)
+                    add(AuthenticationConstants.Broker.BROKER_HOST_APP_PACKAGE_NAME)
+                    add(AuthenticationConstants.Broker.MOCK_AUTH_APP_PACKAGE_NAME)
+                    add(AuthenticationConstants.Broker.MOCK_CP_PACKAGE_NAME)
```
iamgusain, line 132: Should these be added behind the check sShouldTrustDebugBrokers

rpdome: It already is (but as a set of BrokerData, not just names like this one object).

PR #2074, opened 2023-06-14
Cache handshake error and add expirty time to hello cache.

mohitc1:
This PR addresses requirements in https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2582901
1. Caching handshake error.
If the min required version from MSAL/MSALCPP is > Broker max supported version, then broker returns unsupported_broker_version in result bundle, this is thrown as exception from MSAL. This failure is not cached in hello cache. So, when MSAL makes call again, the call again goes to broker and fails. So, every time caller attempts to use broker, there's handshake failure after roundtrip to Broker, making flow inefficient for caller. So, we are caching the handhsake error now, so that once roundtrip to broker fails, next calls can quickly return.
2. Expiry time with cache entry.
In addition to above changes, also adding expiration time to cache entry. This helps in MSAL refresh handshake details periodically. With PRTv3, max protocol version is kept behind feature flag and can change dynamically as we rollout the feature. So, it's possible that negotiated broker protocol version changes from the previous value (which could be handshake_error or different version)

**Code changes**
1. The cache entry now looks like
```<string name="msal.to.broker[3.0,14.0]:com.azure.authenticator[202334773]">S,12.0,1686746861108</string>```
```<string name="msal.to.broker[14.0,14.0]:com.azure.authenticator[202334773]">E,handshake_error,1686746861108</string>```
Prefix S -> indicates an entry with valid negotiated protocol version value.
Prefix E -> indicates and entry with an error. Only handshake_error is supported.
2. Return HelloCacheResult object instead of from HelloCache. The result either contains negotiated broker protocol version or isHandshakeError boolean indicator. Provides type safety over passing just String as result.

### CODE REVIEW

File: common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java [hunk 1]
```diff
@@ -70,7 +80,29 @@ public static void setIsEnabled(final boolean value) {
     }
 
     /**
-     * Default constructor.
+     * Holds result of querying cache result.
+     */
+    @RequiredArgsConstructor
+    public static class HelloCacheResult {
```
rpdome: create a class with private constructor, but expose 
1. static HelloCacheResult createWithError()
2. static HelloCacheResult createWithProtocolVersion()
3. String getProtocolVersionOrThrow()

Alternatively... use Kotlin's Result<> class. not sure if it can be used in java. (should be)

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
```diff
@@ -168,4 +284,51 @@ public void clearCache() {
...
+    @Accessors(prefix = "m")
+    @Getter
+    private static class HelloCacheValue {
+        private static final String TAG = HelloCacheValue.class.getSimpleName();
+
+        private static final String SEPARATOR = ",";
+        /**
+         * Stores either negotiated protocol value or handshake error.
+         */
+        private final String mNegotiatedValue;
```
rpdome: can we keep the negotiated value and the error separately?

rpdome: (maybe we don't need 2 separate classes for this and the Result one.)

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.java
mohitc1: Java does not work well with kotlin's Result object; can't reference methods on kotlin's result class.

mohitc1: Made suggested changes except that we don't throw from Hello* classes and only send error in result. Caller can decide based on result contents. BrokerMsalController creates UnsupportedBrokerException and returns.

Also moved entire code to kotlin.

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+                packageInfo.versionCode.toString()
+            }
+        }
+
+    /**
+     * Internal helper class to store and retrieve cache value in safe manner.
+     * @param negotiatedProtocolVersion Stores either negotiated protocol value or handshake error.
+     * @param timeStamp Time stamp of entry.
+     */
+    class HelloCacheResult (
```
rpdome: nit: move this to its own file.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -246,17 +259,54 @@ String hello(final @NonNull IIpcStrategy strategy,
...
+            );
+            throw e;
+        }
+    }
+
+    /**
+     * Tries reading negotiated protocol version from hello cache and returns it.
+     * @throws UnsupportedBrokerException when there's handshake error present in hello cache.
+     */
+    private String tryGetNegotiatedProtocolVersionFromHelloCache(
```
rpdome, line 289: nit: add nullable for the function return value.

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+        private val TAG = HelloCache::class.java.simpleName
+        private const val SHARED_PREFERENCE_NAME = "com.microsoft.common.ipc.hello.cache"
+        private var sIsEnabled = true
+        private val sTimeoutInMs = TimeUnit.HOURS.toMillis(4)
+
+        /**
+         * If set to false, Hello cache will be disabled.
+         * When you're developing protocol change, you might not want the cache to be enabled.
+         *
+         *
```
rpdome: nit: remove extra line

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+            getNegotiatedProtocolVersionCacheKey(
+                clientMinimumProtocolVersion,
+                clientMaximumProtocolVersion
+            )
+        } catch (e: PackageManager.NameNotFoundException) {
+            Logger.error(methodTag, "Failed to retrieve key", e)
+            return null
+        }
+        val negotiationValue = fileManager[key]
+        if (StringUtil.isNullOrEmpty(negotiationValue)) {
```
rpdome: nit: negotiationValue.isNullOrEmpty()

(and you shouldn't have to use the !! next line)

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+    private val protocolName: String,
+    private val targetAppPackageName: String,
+    components: IPlatformComponents,
+    private val timeoutInMs: Long = sTimeoutInMs
+) {
+    companion object {
+        private val TAG = HelloCache::class.java.simpleName
+        private const val SHARED_PREFERENCE_NAME = "com.microsoft.common.ipc.hello.cache"
+        private var sIsEnabled = true
+        private val sTimeoutInMs = TimeUnit.HOURS.toMillis(4)
```
rpdome: nit: add comments on this (and maybe rename to DEFAULT_TIMEOUT_MS or something of sorts)

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+            private const val SUCCESS_PREFIX = "S"
+            private const val SUCCESS_CACHE_VALUE_FORMAT = "$SUCCESS_PREFIX$SEPARATOR%s$SEPARATOR%d"
+            private const val ERROR_CACHE_VALUE_FORMAT = "$ERROR_PREFIX$SEPARATOR%s$SEPARATOR%d"
+            private const val HANDSHAKE_ERROR = "handshake_error"
+
+            /**
+             * Reads raw entry from cache into HelloCacheValue object.
+             * @param value cache entry read from file expected to be in format <negotiated protocol version>,<timestamp>
+             * @return null if value is not valid.
+            </timestamp></negotiated> */
```
rpdome: nit: typo? wrong line order?

common/src/main/java/com/microsoft/identity/common/internal/cache/HelloCache.kt
```diff
@@ -0,0 +1,352 @@
...
+            private const val SEPARATOR = ","
+            private const val ERROR_PREFIX = "E"
+            private const val SUCCESS_PREFIX = "S"
+            private const val SUCCESS_CACHE_VALUE_FORMAT = "$SUCCESS_PREFIX$SEPARATOR%s$SEPARATOR%d"
+            private const val ERROR_CACHE_VALUE_FORMAT = "$ERROR_PREFIX$SEPARATOR%s$SEPARATOR%d"
+            private const val HANDSHAKE_ERROR = "handshake_error"
+
+            /**
+             * Reads raw entry from cache into HelloCacheValue object.
+             * @param value cache entry read from file expected to be in format <negotiated protocol version>,<timestamp>
```
rpdome: not [PREFIX],[PROTOCOL_VERSION/ERROR],[TIMESTAMP]?

rpdome (approved):
left a lot of nit, but overall lgtm.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -220,14 +227,21 @@ String hello(final @NonNull IIpcStrategy strategy,
...
 
-        final String cachedProtocolVersion = mHelloCache.tryGetNegotiatedProtocolVersion(
-                minRequestedVersion, clientMaxProtocolVersion);
+        final String cachedProtocolVersion = tryGetNegotiatedProtocolVersionFromHelloCache(
+                minRequestedVersion,
+                clientMaxProtocolVersion
+        );
 
-        if (!StringUtil.isEmpty(cachedProtocolVersion)) {
+        if (cachedProtocolVersion != null) {
```
nickbopp: >if (cachedProtocolVersion != null) { [](http://example.com/codeflow?start=8&length=36)

From this it seems like empty string == no negotiated version.  Is that right?

Down below, could we make the catch block return an empty string instead of an exception then?  It seems like the first time will throw, and the second time will return an empty string.  If I'm reading that right, I would prefer they were consistent.

Let me also make a note to go edit our inMimiumBrokerVerisonSupported* function because right now it has this block, which I think would be wrong:
String version = mController.hello(mStrategy, brokerVersion);
                if (version != null) {
                    return true;
                }

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -246,17 +260,56 @@ String hello(final @NonNull IIpcStrategy strategy,
...
-                clientMaxProtocolVersion,
-                negotiatedProtocolVersion);
+            mHelloCache.saveNegotiatedProtocolVersion(
+                    minRequestedVersion,
+                    clientMaxProtocolVersion,
+                    negotiatedProtocolVersion);
+
+            return negotiatedProtocolVersion;
+        } catch (final UnsupportedBrokerException e) {
+            mHelloCache.saveHandShakeError(
```
nickbopp: >S [](http://example.com/codeflow?start=32&length=1)

Handshake one word

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
nickbopp: Nevermind, I see now this is still a throw.  Would it be breaking to swap this to a sentinel empty string value instead? I'm not sure what the general Java opinion is on throwing in 'expected' failure cases like the broker version not being new enough that we'll hit on every MSA request.

nickbopp (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: Thanks for catching. Switched back to isEmpty.

In theory, it can be breaking since for a handshake failure we have always returned this exception.
I kept the same contract with cache also.

PR #2085, opened 2023-06-20
Fix possible NPE in isFullBrowser()

zeeshanjamal:


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
```diff
@@ -233,7 +233,8 @@ private static boolean isCustomTabsServiceSupported(@NonNull final Context conte
 
     private static boolean isFullBrowser(final ResolveInfo resolveInfo) {
         // The filter must match ACTION_VIEW, CATEGORY_BROWSEABLE, and at least one scheme,
-        if (!resolveInfo.filter.hasAction(Intent.ACTION_VIEW)
+        if (resolveInfo.filter == null
```
shahzaibj, line 236: Should we also check for `resolveInfo` itself being `null` as well? (before checking the resolve.filter)

shahzaibj (dismissed):

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserSelector.java
zeeshanjamal: I don't think so. `resolveInfo` is already checked by the caller

shahzaibj: Ok, in that case let's add the `NonNull` annotation to the `resolveInfo` parameter to this method

zeeshanjamal: Added.

zeeshanjamal:

PR #2086, opened 2023-06-20
Use System.Accesstoken for vsts-publish task

iamgusain:
Using [system.accesstoken](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken) for the vsts-publish task. This will remove the dependency on the updating the PAT when it expires.

### DIFF
--- a/azure-pipelines/templates/steps/vsts-release/vsts-release-template.yml
+++ b/azure-pipelines/templates/steps/vsts-release/vsts-release-template.yml
@@ -30,7 +30,7 @@ jobs:
     displayName: Set MVN AccessToken in Environment
     inputs:
       filename: echo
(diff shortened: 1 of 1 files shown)


mohitc1 (approved):

shahzaibj:
Can you add some documentation or link to some documentation that explains how it avoids dependency on updating an expired PAT?

shahzaibj (approved):

somalaya (approved):

iamgusain:
> Can you add some documentation or link to some documentation that explains how it avoids dependency on updating an expired PAT?

[system.accesstoken](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken) is a predefined variable that cotains the system token the build is running with and has the necessary scope to access the packages and builds. With this we don't need to use personal tokens,

PR #2092, opened 2023-06-22
June Ui Maintenance

fadidurah:
Some Ui Maintenance Work

Bypass creating powerlift incidents for failed tests by default and add a build parameter to enable them `-PsendPowerLiftLogs` (I don't find these particularly useful, and last time checked they were visible in powerlift, but this could have changed. Saves a little bit of cost in case of retries/failures.)

I feel that we've had excesive waiting in our UI automation. We originally did this to improve reliability of the test cases, but i've attempted to run it for 2 days with these changes and didn't see much impact on relibaility, so i'm reducing the amount of time we wait for certain things (temp user creation, timeout until UI element is found, Password reset wait, etc.)

Also removed step when we were turning off battery optimization in CP, added extra time cost but didn't seem to give much benefit overall.

### DIFF
--- a/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
+++ b/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
@@ -57,9 +57,8 @@ import lombok.NonNull;
 public class LabClient implements ILabClient {
 
     private final LabApiAuthenticationClient mLabApiAuthenticationClient;
-    private final long PASSWORD_RESET_WAIT_DURATION = TimeUnit.SECONDS.toMillis(90);
-    private final long LAB_API_RETRY_WAIT = TimeUnit.SECONDS.toMillis(5);
-    private final long TEMP_USER_CREATION_WAIT = TimeUnit.SECONDS.toMillis(30);
+    private final long PASSWORD_RESET_WAIT_DURATION = TimeUnit.SECONDS.toMillis(65);
+    private final long LAB_API_RETRY_WAIT = TimeUnit.SECONDS.toMillis(2);
 
     /**
      * Temp users API provided by Lab team can often take more than 10 seconds to return...hence, we
@@ -212,13 +211,6 @@ public class LabClient implements ILabClient {
         }
 
         final String password = getPassword(tempUser);
-
-        // Adding a wait to finish temp user creation
-        try {
-            Thread.sleep(TEMP_USER_CREATION_WAIT);
-        } catch (final InterruptedException e) {
-            e.printStackTrace();
-        }
 
         return new LabAccount.LabAccountBuilder()
                 .username(tempUser.getUpn())
--- a/uiautomationutilities/build.gradle
+++ b/uiautomationutilities/build.gradle
@@ -22,6 +22,7 @@ android {
         dependencies {
             coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:$rootProject.ext.coreLibraryDesugaringVersion"
         }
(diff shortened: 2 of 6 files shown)


LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -213,13 +212,6 @@ private ILabAccount createTempAccountInternal(@NonNull final TempUserType tempUs
 
         final String password = getPassword(tempUser);
 
-        // Adding a wait to finish temp user creation
-        try {
-            Thread.sleep(TEMP_USER_CREATION_WAIT);
-        } catch (final InterruptedException e) {
-            e.printStackTrace();
-        }
-
```
shahzaibj, line 222: So this is no longer needed?

uiautomationutilities/build.gradle
```diff
@@ -105,6 +106,7 @@ android {
         buildConfigField("String", "UPDATE_SOURCE_LOCAL_APK", "\"$appSourceLocalApk\"")
         buildConfigField("boolean", "PREFER_PRE_INSTALLED_APKS", "$usePreInstalledApks")
         buildConfigField("boolean", "PRE_INSTALL_LTW", "$preInstallLtw")
+        buildConfigField("boolean", "SEND_POWERLIFT_FAILURES", "$sendPowerLiftLogs")
```
shahzaibj: Perhaps rename the field to `SEND_POWERLIFT_LOGS`

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/PowerLiftIncidentRule.java
```diff
@@ -32,30 +33,39 @@ public void evaluate() throws Throwable {
...
-                    try {
-                        Logger.e(
-                                TAG,
-                                "Encountered error during test....creating PowerLift incident.",
-                                originalThrowable
-                        );
-                        powerLiftIncidentDetails = powerLiftIntegratedApp.createPowerLiftIncident();
-                    } catch (final Throwable powerLiftError) {
-                        Logger.e(
+                    if (BuildConfig.SEND_POWERLIFT_FAILURES) {
```
shahzaibj: Rather than putting this condition here, the better way to do would be to just not add the rule to RuleChain by default and move the condition in the RulesHelper and only add the powerlift incident rule if  `if (BuildConfig.SEND_POWERLIFT_FAILURES)`

shahzaibj (changes requested):
Left some suggestions

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
fadidurah: There's a different wait already happening within the code, in the Lab account rule. This change essentially cuts the wait time in half

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/PowerLiftIncidentRule.java
fadidurah: Makes sense!

uiautomationutilities/build.gradle
fadidurah: Sounds good, I'm not sure why I named it that honestly, considering the build param is already named with "logs"

shahzaibj (approved):

somalaya (approved):

PR #2097, opened 2023-06-27
Rocognize LTW as broker host app

cikelengfeng:
Add LTW's package name to queries tag in manifest.xml so that we can know if LTW app is installed on Android API level 30+.


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationSettings.java
```diff
@@ -153,8 +153,8 @@ public void setBrokerSecretKeys(final Map<String, byte[]> secretKeys) {
             throw new IllegalArgumentException("The passed in secret key map is null.");
         }
 
-        if (secretKeys.size() != 2) {
```
rpdome: This change should not be required.

cikelengfeng: Yes, I removed this change.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -61,12 +59,6 @@ data class BrokerData(val packageName : String,
     companion object {
         val TAG = BrokerData::class.simpleName
 
-        /**
-         * Determines if the debug brokers should be trusted or not.
-         * This should only be set to true only during testing.
-         */
-        val sShouldTrustDebugBrokers = BuildConfig.DEBUG
```
rpdome: we're moving away bit-by-bit from broker validator, so the variable here should be kept instead.

rpdome: (the way to think of it is that BrokerValidator should only contain "operations". BrokerData provide the source of data, and therefore we should be controlling what data to operate on at the source class).

cikelengfeng: I found it's not that easy to remove the variable from BrokerValidator, I reverted changes around "sShouldTrustDebugBrokers".

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -152,6 +152,7 @@ data class BrokerData(val packageName : String,
                 init {
                     add(prodMicrosoftAuthenticator)
                     add(prodCompanyPortal)
+                    add(prodLTW)
```
cikelengfeng: For these two lines of code, I am not sure if we missed them or it's not the right time to add them.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
cikelengfeng: Had a offline conversation with Dome, refrain the 2nd one for now (as it might affect the existing logic - just want to be sure)

somalaya (approved):

PR #2101, opened 2023-06-28
[Common] Invalid broker protocol version check in AT/ATS

mohitc1:
Refer 
https://github.com/AzureAD/ad-accounts-for-android/pull/2384

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -404,7 +404,7 @@ public class BrokerMsalController extends BaseController {
                 saveMsaAccountToCache(resultBundle, (MsalOAuth2TokenCache) parameters.getOAuth2TokenCache());
             }
 
-            result = new MsalBrokerResultAdapter().getAcquireTokenResultFromResultBundle(resultBundle);
+            result = this.getAcquireTokenResultFromResultBundle(resultBundle, parameters.getRequiredBrokerProtocolVersion());
         } catch (final BaseException | ExecutionException e) {
             Telemetry.emit(
                     new ApiEndEvent()
@@ -649,7 +649,8 @@ public class BrokerMsalController extends BaseController {
                         if (resultBundle == null) {
                             throw mResultAdapter.getExceptionForEmptyResultBundle();
                         }
-                        return mResultAdapter.getAcquireTokenResultFromResultBundle(resultBundle);
+
+                        return getAcquireTokenResultFromResultBundle(resultBundle, parameters.getRequiredBrokerProtocolVersion());
                     }
 
                     @Override
@@ -1153,6 +1154,21 @@ public class BrokerMsalController extends BaseController {
             throw new ClientException(ClientException.AUTH_SCHEME_NOT_SUPPORTED,
                     "The min broker protocol version for PopAuthenticationSchemeWithClientKey should be equal or more than 11.0."
                             + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
+        }
+    }
+
+    private AcquireTokenResult getAcquireTokenResultFromResultBundle(
+            @NonNull final Bundle resultBundle,
+            @Nullable final String requiredBrokerProtocolVersion
+    ) throws BaseException {
+        try {
+            return mResultAdapter.getAcquireTokenResultFromResultBundle(resultBundle);
+        } catch (final BaseException e) {
+            if (ErrorStrings.UNSUPPORTED_BROKER_VERSION_ERROR_CODE.equals(e.getErrorCode())) {
+                mHelloCache.saveHandshakeError(requiredBrokerProtocolVersion, CLIENT_MAX_PROTOCOL_VERSION);
+                throw new UnsupportedBrokerException(mActiveBrokerPackageName);
+            }
+            throw e;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1155,4 +1166,30 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
                             + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
         }
     }
+
+    /**
+     * Check if have received broker version not supported error.
```
somalaya, line 1171: nit : Check if we have received

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1155,4 +1166,30 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+        final String methodTag = TAG + ":verifyBrokerVersionIsSupported";
+        if (resultBundle == null) {
+            Logger.info(methodTag, "result bundle is null");
+            return;
+        }
+
+        // check if result bundle contains unsupported broker version exception
+        try {
+            final BrokerResult brokerResult = mResultAdapter.brokerResultFromBundle(resultBundle);
+            if (!brokerResult.isSuccess()
```
somalaya, line 1186: can brokerResult be null?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1155,4 +1166,30 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+    /**
+     * Check if have received broker version not supported error.
+     * @param resultBundle Result bundle from a broker operation.
+     * @param requiredBrokerProtocolVersion Required broker protocol version sent in request.
+     * @throws UnsupportedBrokerException if result contains broker not supported error.
+     */
+    private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle, @Nullable final String requiredBrokerProtocolVersion) throws UnsupportedBrokerException {
+        final String methodTag = TAG + ":verifyBrokerVersionIsSupported";
+        if (resultBundle == null) {
+            Logger.info(methodTag, "result bundle is null");
```
shahzaibj, line 1179: Is this an error case? Do we need to throw?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1155,4 +1166,30 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+        // check if result bundle contains unsupported broker version exception
+        try {
+            final BrokerResult brokerResult = mResultAdapter.brokerResultFromBundle(resultBundle);
+            if (!brokerResult.isSuccess()
+                    && ErrorStrings.UNSUPPORTED_BROKER_VERSION_ERROR_CODE.equals(brokerResult.getErrorCode())) {
+                mHelloCache.saveHandshakeError(requiredBrokerProtocolVersion, CLIENT_MAX_PROTOCOL_VERSION);
+                throw new UnsupportedBrokerException(mActiveBrokerPackageName);
+            }
+        } catch (final ClientException e) {
+            Logger.error(methodTag, "Unable to read broker result from result bundle", e);
```
shahzaibj, line 1192: Do we need to throw the exception back?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: No that method is marked non-null

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: This method is just to check if version check succeeds. Each operation later has validation to create success or failure results.  I added broker protocol version failure in same manner across all operation, so it can be verified at common place here.
So, essentially handling that error here while letting the existing code take care of per operation validation and result creation.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: Like I said in other comment, each operation has different format on result.

Only handling version check here

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
shahzaibj: Ok, "later" seems strange. To me it seems like the check for `resuldBundle == null` needs to be done outside this method (prior to calling this method), and then this method can make the param NonNull and not have to worry about the nullability of the param because that this is then already validated.

I'm not sure how big of a change that is, so I'm fine with what you have for now.

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: Actually null check is there for each operation before this method is called except removeAccount (which is apparently treating null as success.). To not break that I handled it this way.

PR #2107, opened 2023-07-07
Tidy up BrokerValidator

rpdome:
Break down the logic(s) in BrokerValidator - put them into where they actually belong.

Each class should have  "single responsibility" 
1. BrokerData - containing package name and hash of Broker Apps. sShouldTrustDebugBrokers is moved here.
2. BrokerValidator - Validate if a given app is a broker.
3. AccountManagerBrokerDiscoveryUtil - determine who's the active broker "based on the legacy account manager mechanism".
4. getBrokerRedirectUri is now moved to PackageHelper

Also 
- add unit tests to 1-3.
- rename signatureHash -> signingCertificateHash as suggested by the crypto board.
- remove redundant code.


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -131,7 +134,34 @@ public void removeCookiesFromWebView() {
...
     @Override
     public boolean isValidCallingApp(@NonNull String redirectUri, @NonNull String packageName) {
-        return BrokerValidator.isValidBrokerRedirect(redirectUri, mContext, packageName);
+        final String methodTag = TAG + ":isValidCallingApp";
+        final String expectedBrokerRedirectUri = PackageHelper.getBrokerRedirectUri(mContext, packageName);
+        boolean isValidBrokerRedirect = StringUtil.equalsIgnoreCase(redirectUri, expectedBrokerRedirectUri);
+        if (packageName.equals(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
+            final PackageHelper info = new PackageHelper(mContext.getPackageManager());
+            //For merely verifying that the app is AuthApp, use a 512 hash.
+            final String signatureDigest = info.getSha512SignatureForPackage(packageName);
```
melissaahn, line 143: I should probably rename the getSha1/512SignatureForPackage methods to getSha512HashForPackage, or getSha512ThumbprintForPackage, to match with the naming changes in this PR.

common/src/test/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientTests.kt
```diff
@@ -70,7 +70,7 @@ class BrokerDiscoveryClientTests {
                         )
                         returnBundle.putString(
                             BrokerDiscoveryClient.ACTIVE_BROKER_SIGNATURE_HASH_BUNDLE_KEY,
```
melissaahn: This constant should probably be renamed, but that can be done in a diff PR (by either you or me)

melissaahn (approved):
Thank you for doing some renaming. One thing I would mention is that "thumbprint" could be used instead of "hash" (so the variable names could be "signingCertificateThumbprint"), and I think that would be more consistent with how I see the hash of a signing certificate referred to. But it's not something that I feel super strongly about, so I'm good with either one.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
rpdome: I'll just do additional renaming in these PRs.

rpdome: decide to keep getSha512SignatureForPackage() and getSha1SignatureForPackage() the same for now, as i'm avoiding making breaking change to ADAL.

p3dr0rv (approved):

PR #2111, opened 2023-07-14
change domain for tls handshake test

rpdome:
It seems like something is going on with the microsoft.com host, as the handshake now returns 1.2

I looked into our stack, it seems like this layer has been untouched for two years, and it was still passing yesterday.

change the URL to Akamai's test server to unblock tests.


shahzaibj (approved):

mohitc1 (approved):
:shipit:

PR #2114, opened 2023-07-19
Add method to enable PRTv3 feature flag in UI automation

mohitc1:
Adding method to enable PRTv3 while running UI automation


shahzaibj:
Can we and should we just turn the flight on from ECS for Broker versions less than 1.0.0? This would avoid going through the UI screens to do it (which is expensive in a UI Automation test)

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -321,4 +323,19 @@ public void disableMultipleWpj() {
         ThreadUtils.sleepSafely(500, TAG, "Wait before launch.");
         launch();
     }
+
+    /**
+     * Changes flight provider to local flights provider and sets
+     * PRTv3 flight flag to true.
+     */
+    public void enablePrtV3() {
```
iamgusain, line 319: can we make it generic. 
public void setLocalFlightConfigurations(Map<string, string> flightConfigurations)

mohitc1:
> Can we and should we just turn the flight on from ECS for Broker versions less than 1.0.0? This would avoid going through the UI screens to do it (which is expensive in a UI Automation test)

That will enable PRTv3 by default, and we would lose on testing current stack. 

But eventually, we would move to PRTv3 stack, so step of enabling the feature is transient and eventually be gone.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
mohitc1: Added generic method in addtion to this.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerHost.java
```diff
@@ -302,23 +304,35 @@ public String dismissDialog() {
...
+     */
+    public void setLocalFlight(@NonNull final String key, @NonNull final String value) {
         brokerFlightsFragment.launch();
         brokerFlightsFragment.selectLocalProvider();
-        brokerFlightsFragment.setLocalFlight(FLIGHT_FOR_WORKPLACE_JOIN_CONTROLLER, "false");
+        brokerFlightsFragment.setLocalFlight(key, value);
         ThreadUtils.sleepSafely(500, TAG, "Wait before force stop.");
         forceStop();
-        ThreadUtils.sleepSafely(500, TAG, "Wait before launch.");
+        ThreadUtils.sleepSafely(500, "TAG", "Wait before launch.");
```
fadidurah: Quotes around TAG needed?

fadidurah (approved):

iamgusain (approved):

PR #2121, opened 2023-07-24
Merging changes from release 14.0.1 into dev

melissaahn:
- [PATCH] Reverting token removal logic (#2117)
- Updates the maven release pipeline for common4j and common to reference system access token


fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -96,7 +96,6 @@ public SharedPreferencesAccountCredentialCache(
         super(sharedPreferencesFileManager);
         Logger.verbose(TAG, "Init: " + TAG);
         mCacheValueDelegate = accountCacheValueDelegate;
-        new Thread(this::removeSha1ApplicationIdentifierAccessTokensIfNeeded).start();
```
fadidurah, line 99: Is this part of reverting token removal logic?

melissaahn: yes, that thread ran the logic to remove the token if necessary.

mohitc1 (approved):

PR #2123, opened 2023-07-24
Move Broker side active broker cache to broker repo

rpdome:
see https://github.com/AzureAD/ad-accounts-for-android/pull/2421

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheTest.kt
+++ b/common/src/test/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheTest.kt
@@ -31,7 +31,7 @@ import org.junit.Assert
 import org.junit.Test
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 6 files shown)


common/src/test/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheTest.kt
```diff
@@ -35,42 +33,22 @@ import java.util.concurrent.TimeUnit
 
 class ActiveBrokerCacheTest {
```
p3dr0rv: rename file/class BaseActiveBrokerCacheTest

rpdome: There are test from the client cache here as well... although i can extract those out.

p3dr0rv (approved):

fadidurah (approved):

PR #2125, opened 2023-07-26
Set target and compile sdk to 34

somalaya:
- Increasing the targetSDK and compileSDK to 34
- After increasing the targetSDK, it gave some errors like exported attribute must be added when there is an intent filter. So I added that.
- PowerliftSDK version to the latest '1.0.3' as this one has fixes for target and compile sdk to be 34.


shahzaibj (approved):
You probably need to do this in other repos as well

p3dr0rv (approved):

PR #2127, opened 2023-07-28
Initial dependencies for passkey work

melissaahn:
## Summary
This brings in the necessary dependencies (that I know of so far) into the common build.gradle, along with a few other changes.
- `androidx.credentials:credentials` and `androidx.credentials:credentials-play-services-auth` are the [Credential Manager] (https://developer.android.com/jetpack/androidx/releases/credentials) dependencies needed to interact with Google Password Manager and other 3P passkey providers. The most stable version is a beta version, so I expect the version to change a few times in the next few months. 
    - I'll create actual variables to hold the dependency versions in a future PR (before the feature branch gets merged into dev). 
- `androidx.lifecycle:lifecycle-runtime-ktx` in order to make use of the lifecycle scope. The main methods from CredMan need to run in a coroutine, and the lifecycle scope fits what we need there.
- CredMan documentation suggests adding directives to the proguard-rules.pro file to not optimize `androidx.credentials.playservices` right away.
- Changes to the gradle, kotlin, and compileSdk versions are expected to be merged directly into dev with separate PRs. For now this means that all common checks are probably going to fail.

I was able to make some test calls with CredMan to create a demo passkey and sign a challenge with the passkey.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -145,6 +145,7 @@ dependencies {
     testImplementation "org.jetbrains.kotlin:kotlin-stdlib:${rootProject.ext.kotlinVersion}"
     androidTestImplementation "org.jetbrains.kotlin:kotlin-stdlib:${rootProject.ext.kotlinVersion}"
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.4.0"
 
     implementation "androidx.datastore:datastore-preferences:1.0.0"
 
@@ -159,6 +160,10 @@ dependencies {
     //For Executive Order work
     implementation "com.yubico.yubikit:android:$rootProject.ext.yubikitAndroidVersion"
     implementation "com.yubico.yubikit:piv:$rootProject.ext.yubikitPivVersion"
+    //For passkeys
+    implementation "androidx.credentials:credentials:1.2.0-beta01"
+    // Needed for credentials support from play services, for devices running Android 13 and below.
+    implementation "androidx.credentials:credentials-play-services-auth:1.2.0-beta01"
 
     compileOnly 'com.github.spotbugs:spotbugs-annotations:4.3.0'
 
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -6,12 +6,12 @@ ext {
     brokerProjectMinSdkVersion = 24
     automationAppMinSDKVersion = 21
     targetSdkVersion = 30
-    compileSdkVersion = 30
+    compileSdkVersion = 34
     buildToolsVersion = "28.0.3"
 
     // Plugins
-    gradleVersion = '4.2.2'
-    kotlinVersion = '1.7.21'
+    gradleVersion = '7.4.2'
+    kotlinVersion = '1.8.22'
     spotBugsGradlePluginVersion = '4.7.1'
     jupiterApiVersion = '5.6.0'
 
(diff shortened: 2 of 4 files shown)


common/build.gradle
```diff
@@ -159,6 +160,10 @@ dependencies {
     //For Executive Order work
     implementation "com.yubico.yubikit:android:$rootProject.ext.yubikitAndroidVersion"
     implementation "com.yubico.yubikit:piv:$rootProject.ext.yubikitPivVersion"
+    //For passkeys
+    implementation "androidx.credentials:credentials:1.2.0-beta01"
```
rpdome, line 164: are we planning to GA this feature with beta?

gradle/versions.gradle
```diff
@@ -6,12 +6,12 @@ ext {
...
     automationAppMinSDKVersion = 21
     targetSdkVersion = 30
-    compileSdkVersion = 30
+    compileSdkVersion = 34
     buildToolsVersion = "28.0.3"
 
     // Plugins
-    gradleVersion = '4.2.2'
-    kotlinVersion = '1.7.21'
+    gradleVersion = '7.4.2'
```
rpdome, line 13: might have to update the whole project (from android complete downward) as well.

rpdome (approved):
lgtm

note: I would suggest avoiding having a feature branch. (due to merge conflict nightmare)

rather, think of your work as a "module". 
- Check in individual pieces in the module + non breaking behavior (like this PR) first (into dev). 
    - These individual pieces should be verifiable via unit test + dependency injection (Move any android/webview dependency into a function,interface that you pass into your code as a parameter. In unit test, simply mock those function/interface). Something like [this](https://github.com/AzureAD/ad-accounts-for-android/blob/f7e02fea5def163004280d5c9d4aa7667f1e5562/AADAuthenticator/src/main/java/com/microsoft/identity/broker/activebrokerselection/BrokerDiscoveryManager.kt#L56), and [this](https://github.com/AzureAD/ad-accounts-for-android/blob/f7e02fea5def163004280d5c9d4aa7667f1e5562/AADAuthenticator/src/test/java/com/microsoft/identity/broker/activebrokerselection/BrokerDiscoveryManagerTest.kt#L400)
- wire the module up like [this](https://github.com/AzureAD/ad-accounts-for-android/blob/f7e02fea5def163004280d5c9d4aa7667f1e5562/AADAuthenticator/src/main/java/com/microsoft/identity/broker/activebrokerselection/BrokerSelectionInstanceFactory.kt#L57)
- wire your module into the code using a "static variable" switch.  
```
      if (MELISSA_FEATURE_ENABLED){
            return YourModule()
      }
      
      return existingModule()
```
(Even you're not flighting your feature - writing a flightable code is a good practice in general IMO)

4. In the test app, either hardcode or add a button to toggle this static switch.

common/build.gradle
melissaahn: We'll GA with the latest tested version that CredMan has available (at the time of public preview/GA). They seem to be updating on a monthly cadence: https://developer.android.com/jetpack/androidx/releases/credentials#version_12_2.

gradle/versions.gradle
melissaahn: In the upcoming month, gradle and kotlin versions are going to be updated in separate PRs for the whole project. And based on your module suggestion above, I will also update the whole project with the passkey dependency versions once the gradle and kotlin ones are merged in.

melissaahn:
Based on Dome's suggestion, I think it would be best to close this PR for now and wait for the kotlin and gradle PRs to be created/merged in first. I'll then create a set of PRs for all libraries + android-complete that updates with the passkey dependencies.

PR #2129, opened 2023-07-31
July UI, add support for a second speed bump, new annotation to avoid running some tests on daily

fadidurah:
July Ui maintentance, some cross cloud tests are experiencing a second speed bump page, also created an annotation to signify cases that would fail if ran with daily versions (due to the versions starting with 0.0).

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/annotations/FailsWithDailyVersions.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/annotations/FailsWithDailyVersions.java
@@ -1,0 +1,38 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AadLoginComponentHandler.java
```diff
@@ -212,6 +212,9 @@ public void handleVerifyYourIdentity() {
     public void handleHowWouldYouLikeToSignIn() {
         // Looks like we sometimes see this UI prompt asking "How would like to sign in?"
         // We press button1, which is "Ok" to confirm the default selection of using device certificate.
-        UiAutomatorUtils.handleButtonClick("android:id/button1", mFindLoginUiElementTimeout);
+        UiAutomatorUtils.handleButtonClickSafely("android:id/button1", mFindLoginUiElementTimeout);
+
+        // Sometimes need to also confirm the selection
```
shahzaibj: Is this going to attempt two times in every test case? So if there is not a need to confirm then code will still try to click this button?

fadidurah: Yes, this probably should be another parameter. I'll add that.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/annotations/FailsWithDailyVersions.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * An annotation indicating that an automated End-to-End test case is failing when used with Daily Pipeline versions,
+ * but should still be attempted on Monthly Release Pipeline for validation.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface FailsWithDailyVersions {
+    String value() default "";
```
shahzaibj, line 37: Why is a value needed?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/PromptHandlerParameters.java
```diff
@@ -133,6 +139,11 @@ public class PromptHandlerParameters {
      */
     private final boolean howWouldYouLikeToSignInExpected;
 
+    /**
+     * Denotes whether or not to except "Choose certificate" Prompt
```
shahzaibj:
```suggestion
     * Denotes whether or not to expect "Choose certificate" Prompt.
```

shahzaibj (approved):
Left some comments but otherwise lgtm

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/annotations/FailsWithDailyVersions.java
fadidurah: Just to give room to provide reasoning for adding the annotation

rpdome (approved):

PR #2131, opened 2023-08-01
Create auth FIDO challenge object and factory

melissaahn:
## Summary
This PR adds new classes that represent FIDO challenges (auth) and expected fields from the passkey server protocol. A factory is also added that will return an IFidoChallenge based on the redirect URI received. This factory will eventually include logic to return either an AuthFidoChallenge or a RegFidoChallenge based on the design of the passkey server protocol.
All of the query parameters of the auth passkey protocol except allowedCredentials and keyTypes are required, so those two parameters are nullable. 

Note that these classes are not currently used or called anywhere, but they will eventually be integrated in future PRs.
Unit tests were added for FidoChallengeFactory to test well-formed and malformed redirect URIs.


common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallenge.kt
```diff
@@ -0,0 +1,40 @@
...
+    override val userVerificationPolicy: String,
+    override val version: String,
+    override val submitUrl: String,
+    override val keyTypes: List<String>,
+    override val context: String,
+    /**
+     * List of credential IDs the client will use for filtering.
+     */
+    val allowedCredentials: List<String>
+) : IFidoChallenge
```
fadidurah: nit: new lines at the end of the new files

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        /**
+         * Validates that the given parameter is not null or empty.
+         * @param value list value for a passkey protocol parameter
+         * @param field passkey protocol field
+         * @throws ClientException if the parameter is null or empty
+         */
+        @Throws(ClientException::class)
+        private fun validateRequiredListParameter(field: String, value: String?): List<String> {
+            return validateRequiredParameter(field, value).split(DELIMITER).toList()
```
fadidurah: Not sure what we are doing here. Do we want to check each individual element of the list? Seems like we are only checking it as a full string then splitting it

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
```diff
@@ -0,0 +1,112 @@
...
+                throw ClientException(PASSKEY_PROTOCOL_REQUEST_INVALID, "$field not provided")
+            } else if (value.isBlank()) {
+                throw ClientException(PASSKEY_PROTOCOL_REQUEST_INVALID, "$field is empty")
+            }
+            return value;
+        }
+
+        /**
+         * Validates that the given parameter is not null or empty.
+         * @param value list value for a passkey protocol parameter
```
fadidurah:
```suggestion
         * @param list value for a passkey protocol parameter
```

fadidurah (approved):
Left some comments but LGTM

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallenge.kt
melissaahn: done

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
melissaahn: It was only meant to check as full string and split, but checking individual elements against known values when possible is something I'll do in a future PR as well (becaues I need the complete server side design to know what to check against)

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
melissaahn: "value" is the parameter name, but I did edit the kotlindocs to be better written,

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallenge.kt
melissaahn: Interesting, I did put an extra newline at the end of all the classes, but that doesn't seem to be reflected in the latest commit. Please take my word for it that I did it :)

common/src/test/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactoryTest.kt
```diff
@@ -0,0 +1,258 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+//Robolectric used for Uri dependency
+@RunWith(RobolectricTestRunner::class)
+class FidoChallengeFactoryTest {
```
rpdome, line 35: nit: I'd suggest adding test for the validation methods as well. This will help you identify where the code breaks easier (if there's any change in the future).

rpdome (approved):

common/src/test/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactoryTest.kt
melissaahn: Added more tests directly testing validateRequiredParameter and validateOptionalListParameter. Made these validation methods public as a result.

PR #2103, opened 2023-06-30
Task 2616363: MSAL test app infra extension: Add support for selecting Auth Scheme, Click Generate SHR & remove user button.

amoyixi:
Add support for:

1. select POP or BEARER from Auth Scheme spinner
2. click Generate SHR button and validate the showing token
3. click Remove User button and validate the message showing on toast

PR #2136, opened 2023-08-07
Refactored OneAuthTestApp infra in order to process LTW test cases

amoyixi:
This pr is about refactoring OneAuthTestApp infra in order to process LTW test cases
1. add method `handlePreferBrokerSwitchButton`
2. add method `selectFromAppConfiguration`
3. add method `handleSignInWithoutPrompt`


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/OneAuthTestApp.java
fadidurah, line 1: Seems handleUserNameInput was not added

fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/OneAuthTestApp.java
```diff
@@ -141,6 +141,11 @@ private void signIn(@NonNull final String username,
         new AadPromptHandler(promptHandlerParameters).handlePrompt(username, password);
     }
 
+    public String handleSignInWithoutPrompt() {
+        UiAutomatorUtils.handleButtonClick("com.microsoft.oneauth.testapp:id/sign_in_button");
+        return getTokenSecret();
+    }
+
```
somalaya: can you confirm if the indentation here is right? It seems off to me

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/OneAuthTestApp.java
```diff
@@ -141,6 +141,11 @@ private void signIn(@NonNull final String username,
         new AadPromptHandler(promptHandlerParameters).handlePrompt(username, password);
     }
 
+    public String handleSignInWithoutPrompt() {
```
somalaya: your method name says sign in without prompt. But you are returning a secret. These look a bit disconnected. I would suggest you move your assert statements to this method. Assert here itself if the token is valid. then you do not have to return the token to the test

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/OneAuthTestApp.java
amoyixi: My mistake, handleUserNameInput already got approved in commit #2103. Modified pr description by removing handleUserNameInput.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/OneAuthTestApp.java
amoyixi: I've checked and the indentation is correct.

somalaya (approved):

PR #2138, opened 2023-08-09
Fixed debug apps not recognized as active broker issue

somalaya:
Issue : Debug apps of CP and auth app were not being recognized as active broker due to a regression introduced in this PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2107/files
We were only taking the first app from the knowBrokerApps list for a given pkg name. So it was always returning the release broker apps. And when debug broker apps are installed, the check for isValidBrokerPackage fails.

Fix :  Made a change to look through all the matches for a given pkg name.  

Testing : Added UTs and also tested from MSALTestAPp if the debug auth app is being recognized as an active broker


common/src/test/java/com/microsoft/identity/common/internal/AccountManagerBrokerDiscoveryUtilTests.kt
```diff
@@ -58,6 +60,34 @@ class AccountManagerBrokerDiscoveryUtilTests {
...
+    @Test
+    fun testGetCurrentActiveBrokerWithReleaseAuthApp() {
+        val util = AccountManagerBrokerDiscoveryUtil(
+            knownBrokerApps = BrokerData.prodBrokers,
+            isSignedByKnownKeys = { brokerData ->
+                // Mock validation.
+                brokerData == BrokerData.prodMicrosoftAuthenticator
+            },
+            getAccountManagerApps = {
+                getMockedAccountManager()?.authenticatorTypes!!
```
rpdome, line 86: nit: can we return the list of AuthenticatorDescription directly? (as opposed to setting up mockito.

rpdome (approved):

common/src/test/java/com/microsoft/identity/common/internal/AccountManagerBrokerDiscoveryUtilTests.kt
somalaya: I tried doing that, if this is what you meant 
 val accountManager = AccountManager.get(ApplicationProvider.getApplicationContext())
        Shadows.shadowOf(accountManager).addAuthenticator("com.microsoft.workaccount")
   getAccountManagerApps = {
                accountManager.authenticatorTypes
            })

But, this was not setting the authenticator.pkgName field. And in the method to get the valid broker, we are checking if authenticator.pkgName is null. We simply return. So I had to mock it.

shahzaibj (approved):

PR #2140, opened 2023-08-10
Update active broker cache upon returned result

rpdome:
PBI: https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=2582815

With the new Broker selection mechanism - every response from broker will contain the updated active broker package name and signing certificate thumbprint.

(This is because the broker app that was made a request to might *not* be the broker app that process the request. This is rare, but technically can happen)

This change is to make sure that the active broker cache remains fresh if such edge case happens.

In `BrokerOperationExecutor.performStrategy()`, we'll always _try_ to update the cache (if the active broker info is returned along with the response bundle). 

NOTE: I also remove the in-memory cached value from BaseActiveBrokerCache.. simply because there could be multiple cache instances, and this (premature optimization) could lead to issues.

NOTE-2: this change will only be applicable to MSAL and Broker API. Extra work is needed for the new MWPJ API.

----

Validated by the added unit tests in BrokerOperationExecutorTests and ActiveBrokerCacheUpdaterUtilTest

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
@@ -28,6 +28,8 @@ import android.os.Bundle;
 import com.microsoft.identity.common.exception.BrokerCommunicationException;
 import com.microsoft.identity.common.internal.broker.ipc.BrokerOperationBundle;
 import com.microsoft.identity.common.internal.broker.ipc.IIpcStrategy;
+import com.microsoft.identity.common.internal.cache.IActiveBrokerCache;
+import com.microsoft.identity.common.internal.cache.ActiveBrokerCacheUpdaterUtil;
 import com.microsoft.identity.common.internal.telemetry.Telemetry;
 import com.microsoft.identity.common.internal.telemetry.events.ApiEndEvent;
 import com.microsoft.identity.common.internal.telemetry.events.ApiStartEvent;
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -109,13 +112,20 @@ public interface BrokerOperation<T> {
...
     /**
      * @param strategies list of IIpcStrategy to be invoked.
      */
-    public BrokerOperationExecutor(final @NonNull List<IIpcStrategy> strategies) {
+    public BrokerOperationExecutor(@NonNull final List<IIpcStrategy> strategies,
+                                   @NonNull final IActiveBrokerCache activeBrokerCache,
+                                   @NonNull final ActiveBrokerCacheUpdaterUtil cacheUpdaterUtil) {
         mStrategies = strategies;
+        mActiveBrokerCache = activeBrokerCache;
+        mCacheUpdaterUtil = cacheUpdaterUtil;
```
p3dr0rv: Why do you choose to provide the Util as a constructor parameter instead of just invoking the Util class?

p3dr0rv: I assume this is done for testing purposes.

p3dr0rv (approved):

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
rpdome: That's correct. I'm allowing the util class to take in an instance of "isValidBroker" implementation here. 

Perhaps I should rename it to ActiveBrokerCacheUpdater instead.

PR #2142, opened 2023-08-14
Target, compile SDK, AGP and gradle version increase

somalaya:
1. Target sdk version changed to 33
2. Compile sdk version changed to 34
3. AGP version changed to 7.4.2
4. Gradle distribution url changed to 7.5
**After changing above versions, I had to update other dependencies for compatibility**
1. Powerlift version changed to 1.0.3
2. Bouncy castle version changed to 1.69
3. robolectric version to 4.9
4. Also, with new gradle versions, some of the properties used in gradle were deprecated (naming conventions changed). So, made those changes
5. With gradle version changes, we should not be having a package name in AndroidManifest.xml.. instead, should add a namespace property in build.gradle. https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace
6. For apps targeting Android 14, Android restricts apps from sending implicit intents to internal app components.  So , added exported attribute. https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents
7. Also, fixed a lint issue
MicrosoftTeams-image (9)
8. Removed SonarQube Quality gate analysis because it was not being used - Was set to default. Also, we have CodeQL that is doing the Quality gate analysis.

9.  Modified the publish task so that maven release pipeline would find the artifact to be published in common-{$version} format! This leads to the daily pipeline to break because publish task in daily pipeline has a dependency task that looks for common-dist-release.aar. Because we renamed the artifact for the mvn release pipeline to work, daily pipeline breaks. For this, I gave a flag to let common-dist-release.aar be renamed ONLY in release pipeline and not in daily pipeline. In daily pipeline, the publish task's -PprojVersion would take care of changing common-dist-release.aar to common-{version}.aar file
The flag added is : -PshouldRenameBuildFile

10. ******** IMPORTANT ******** I have also removed common:build task from maven release pipeline. Instead, I am making the assembleDistRelease task depend on sourcesJar task to build the Jar file. This was necessary because, if we have common:build task, it would build localDebug, localRelease, distDebug and distRelease and these will be pushed as artifacts as well.


rpdome:
Have you synced with partner teams (LTW, AuthApp, CP) on the updated versions? (esp. BouncyCastle)

somalaya:
> Have you synced with partner teams (LTW, AuthApp, CP) on the updated versions? (esp. BouncyCastle)

No, I have not synced because I was able to generate LTW, AuthApp and CP with my changes and also the UI tests were passing. 
Please let me know if you still think we should ask them?

shahzaibj:
The pipelines are broken

melissaahn:
For Lint and CodeQL, looks like there might need to be a change to run those tasks with JDK 11 instead of 8?

gradle/versions.gradle
```diff
@@ -5,18 +5,18 @@ ext {
     minSdkVersion = 16
     brokerProjectMinSdkVersion = 24
     automationAppMinSDKVersion = 21
-    targetSdkVersion = 30
-    compileSdkVersion = 30
+    targetSdkVersion = 33
+    compileSdkVersion = 34
```
iamgusain, line 9: Why are we picking two different versions for target vs compile?

gradle/versions.gradle
```diff
@@ -5,18 +5,18 @@ ext {
...
     buildToolsVersion = "28.0.3"
 
     // Plugins
-    gradleVersion = '4.2.2'
+    gradleVersion = '7.4.2'
```
iamgusain, line 13: Can we rename this to  `gradlePluginVersion`

gradle/versions.gradle
```diff
@@ -5,18 +5,18 @@ ext {
     minSdkVersion = 16
```
iamgusain, line 5: Should we also update our minSdkVersion to a higher version?

gradle/versions.gradle
somalaya: Just matching these with auth app. We can increase the targetSDK version in the next release once Auth app increases it

somalaya:
> The pipelines are broken

They are broken because we did not checkin  my change in android-complete. Once that is checked in, the pipeline issues should be resolved

somalaya:
> For Lint and CodeQL, looks like there might need to be a change to run those tasks with JDK 11 instead of 8?

Yes, you are right. JDK 11 will resolve these issues. Gradle 7.5 needs java 11 for Lint and CodeQL to work

common/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.microsoft.identity.common">
+    xmlns:tools="http://schemas.android.com/tools">
```
iamgusain, line 2: I did not find this directly in the documentation you linked in the description.

somalaya: This is the documentation https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace
Updated description

somalaya:
> The pipelines are broken

Pipelines are broker because it uses the template in android complete. And the android complete branch being pulled in these pipelines is master branch. My fixes in android-complete are still on my branch

gradle/versions.gradle
somalaya: I am actually not aware of what is the process for increasing the minSDKVersion. I believe it should be >= minSDK of our consumers. 
minSDKVersion of CP is 26 and that of Auth app is also 26. 
Since this is common, OneAuth should also be considered. Its minSDK is 22. So, I think we can increase it to 22.

melissaahn (approved):
LGTM, and I had tested the layout suggestion by Lint (which had no detectable changes to the UI), so adding that in should hopefully make the Lint task pass

somalaya:
> > Have you synced with partner teams (LTW, AuthApp, CP) on the updated versions? (esp. BouncyCastle)
> 
> No, I have not synced because I was able to generate LTW, AuthApp and CP with my changes and also the UI tests were passing. Please let me know if you still think we should ask them?

Update : CP, Auth app and OneAuth teams have verified their apps based on the private build I shared and they signed off. I am waiting for LTW team to confirm.

gradle/versions.gradle
```diff
@@ -68,7 +68,7 @@ ext {
     javaAssistVersion = "3.27.0-GA"
     yubikitAndroidVersion = "2.3.0"
     yubikitPivVersion = "2.3.0"
-    powerliftAndroidVersion="1.0.0"
+    powerliftAndroidVersion="1.0.3"
```
rpdome, line 71: also something to cross check with partner teams.

rpdome (approved):

gradle/versions.gradle
rpdome: minSdkVersion update is a breaking change (For consuming apps)

https://developer.android.com/build/manage-manifests#override_uses-sdk_for_imported_libraries

shahzaibj: If we want to increase it, then I suggest we increase to 24 which is what the Broker's MIN SDK is. 

Min SDK should be increased with a MAJOR version release of MSAL. MAJOR in the semver world means that client should expect it to be a breaking change.

shahzaibj: TBH, I would really prefer that we increase it and increase to 24 and mark in changelog as MAJOR. I think this has been long overdue.

shahzaibj: That said, we may want to run this by @negoe

gradle/versions.gradle
somalaya: This is the change in the payload for powerlift https://office.visualstudio.com/PowerLift/_git/powerlift-android-sdk?path=/MIGRATION_GUIDE.md&version=GBmaster&_a=preview
This does not have any breaking change in the API that Auth app, LTW and CP are using. 
I also generated CP and Auth app apks and checked if we are able to upload an incident and it is working as expected. I will also keep the partner teams informed about this change

gradle/versions.gradle
somalaya: We cannot increase the version to 24 as OneAuth is still on 22. I would want to take this up separately as I do not want to increase many versions in one go. I have created a PBI to track this item. When OneAuth moves to minSDK 24, I will make a change to directly move to 24 https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2667751

gradle/versions.gradle
shahzaibj: Just curious, if you build AuthApp / CP, then which version is resolved through dependency resolution?

somalaya: It is 1.0.3, it resolves to the dependency we have adedd

azure-pipelines/maven-release/common.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     project: common
     projectVersion: $(commonVersion)
     envVstsMvnAndroidAccessTokenVar: ENV_VSTS_MVN_ANDROIDCOMMON_USERNAME
-    gradleAssembleReleaseTask: common:clean common:build common:assembleDistRelease
+    gradleAssembleReleaseTask: common:clean common:assembleDistRelease -PshouldRenameBuildFile
```
iamgusain: Add comments describing what "shouldRenameBuildFile" does.

somalaya: done

iamgusain (approved):

PR #2143, opened 2023-08-14
Update TokenRequest.java with NAA params

somalaya:
Updating TokenRequest to take child clientId and child redirectUri which are required for NAA flow.
These 2 additional parameters will be set by OneAuth when requesting for tokens for a nested app

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/TokenRequest.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/TokenRequest.java
@@ -170,7 +170,7 @@ public class TokenRequest implements IHasExtraParameters {
         mClientId = clientId;
     }
 
-        /**
+    /**
      * @return mChildRedirectUri of the token request.
      */
     public String getChildRedirectUri() {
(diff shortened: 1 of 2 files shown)


common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/TokenRequest.java
```diff
@@ -162,6 +170,34 @@ public void setClientId(final String clientId) {
...
+
+    /**
+     * @param redirectUri redirect URI of the token request.
+     */
+    public void setChildRedirectUri(final String childRedirectUri) {
+        mChildRedirectUri = childRedirectUri;
+    }
+
+    /**
+     * @return mClientId of the token request.
```
nickbopp: Need some comment fixes on these from their original non-child forms

nickbopp (approved):

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/TokenRequest.java
```diff
@@ -162,6 +170,34 @@ public void setClientId(final String clientId) {
         mClientId = clientId;
     }
 
+    /**
+     * @return mChildRedirectUri of the token request.
+     */
+    public String getChildRedirectUri() {
```
shahzaibj, line 176: At some point we should bring lombok into this class

shahzaibj (approved):

PR #2144, opened 2023-08-14
Add a label to skip common consumer validation

fadidurah:
Add a label to skip consumer validation.

The label is named `Skip-Consumers-Check`, and should only be used when creating PRs that purposefully break common consumers, and must be accompanied by Respective PRs in the other repos to address the breaks. It would be impossible to make this check required without a mechanism like this to enable bypassing it in the scenario outlined above.

### DIFF
--- a/azure-pipelines/pull-request-validation/build-consumers.yml
+++ b/azure-pipelines/pull-request-validation/build-consumers.yml
@@ -92,7 +92,7 @@ stages:
     steps:
     - bash: |
         echo "Looking for label at https://api.github.com/repos/$BUILD_REPOSITORY_ID/issues/$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER/labels"
-        if curl -s "https://api.github.com/repos/$BUILD_REPOSITORY_ID/issues/$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER/labels" | grep '"name": "Skip-Consumers-check"'
+        if curl -s "https://api.github.com/repos/$BUILD_REPOSITORY_ID/issues/$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER/labels" | grep '"name": "Skip-Consumers-Check"'
         then
           echo "##vso[task.setvariable variable=prWithCILabel;isOutput=true]true"
           echo "fullci label found!"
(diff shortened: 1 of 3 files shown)


rpdome (approved):
perhaps rename the variable name (?) to "skipConsumerValidation" (instead of just skipValidation)

azure-pipelines/pull-request-validation/build-consumers.yml
```diff
@@ -85,12 +85,30 @@ stages:
         }
       name: echovar
       displayName: Echo branch name
+  # Check PR Label
```
iamgusain: suggestion:
We can make this a template, which enumerates all the PR labels and stores them in an output variable.
That way we can extend this functionality for other purposes if needed.

azure-pipelines/pull-request-validation/build-consumers.yml
```diff
@@ -85,12 +85,30 @@ stages:
...
+          echo "##vso[task.setvariable variable=skipValidation;isOutput=true]true"
+          echo "Skip-Consumers-Check label found, skip validation"
+        fi
+      condition: eq(variables['Build.Reason'], 'PullRequest')
+      name: checkPrCILabel
   # msal
   - job: msalValidation
     displayName: MSAL
-    dependsOn: setupBranch
+    dependsOn:
```
iamgusain, line 104: suggestion:
we can have two stages.. 1. to check PR label, 2. to build consumers..
that way we can apply the condition at stage level instead of each job level.

iamgusain (approved):

azure-pipelines/pull-request-validation/build-consumers.yml
fadidurah: Oh, that's a useful suggestion!

azure-pipelines/pull-request-validation/build-consumers.yml
fadidurah: Yeah that makes more sense

fadidurah:
> perhaps rename the variable name (?) to "skipConsumerValidation" (instead of just skipValidation)

Sure I can do that

azure-pipelines/pull-request-validation/build-consumers.yml
fadidurah: Seems the output variable is not set when setting condition at stage level, was only able to access it when setting condition at job level

PR #2145, opened 2023-08-15
handle broker discovery disabled on SDK side

rpdome:
In the scenario where we have to disable the broker discovery feature (e.g. via flighting)

We want to make sure that the SDK stops pinging the cached active broker, and go back to using AccountManager to discovery the active broker.

`ActiveBrokerCacheUpdater.appendBrokerDiscoveryDisabledToResultBundle()` will be invoked on the broker side whenever 
1. the feature is disabled on the broker side.
2. The pinged active broker is NOT the same broker that was elected by Account Manager.

Generally, this wouldn't happen (as our new broker selection logic will pick the broker with existing credentials, so it would be the same with the AccountManager-picked one).

We would only see this flag being returned to the SDK when Google regresses the broker switch bug only.

Validated with the added test.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
@@ -27,6 +27,7 @@ import android.os.Bundle
 import com.microsoft.identity.common.internal.broker.BrokerData
 import com.microsoft.identity.common.internal.broker.BrokerValidator
 import com.microsoft.identity.common.logging.Logger
+import java.util.concurrent.TimeUnit
 
 /**
  * With the new broker selection mechanism, it is possible that the active broker that actually
@@ -43,9 +44,9 @@ import com.microsoft.identity.common.logging.Logger
  * */
 class ActiveBrokerCacheUpdater(
     private val isValidBroker: (BrokerData) -> Boolean,
-    private val cache: IActiveBrokerCache) {
+    private val cache: IClientActiveBrokerCache) {
 
-    constructor(context: Context, cache: IActiveBrokerCache): this(
+    constructor(context: Context, cache: IClientActiveBrokerCache): this(
         isValidBroker = { brokerData: BrokerData ->
             BrokerValidator(context).isSignedByKnownKeys(brokerData)
         },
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
```diff
@@ -76,20 +85,42 @@ class ActiveBrokerCacheUpdater(
...
                 activeBroker.signingCertificateThumbprint)
         }
+
+        /**
+         * Adds the "broker discovery is now disabled" to the result bundle.
+         *
+         * @param bundle       The result bundle.
+         */
+        @JvmStatic
+        fun appendBrokerDiscoveryDisabledToResultBundle(bundle: Bundle) {
```
fadidurah, line 95: Currently we seem to only use this in testing, do we need to also call it in other places we call ActiveBrokerCacheUpdater?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
```diff
@@ -76,20 +85,42 @@ class ActiveBrokerCacheUpdater(
...
         val methodTag = "$TAG:updateCachedActiveBrokerFromResultBundle"
         if (bundle == null) {
             return
         }
 
+        val shouldWipeCachedActiveBroker = bundle.getBoolean(BROKER_DISCOVERY_DISABLED_KEY, false)
+        if (shouldWipeCachedActiveBroker) {
+            Logger.info(methodTag, "Got a response indicating that the broker discovery is disabled." +
+                    "Will also wipe the local active broker cache," +
+                    "and skip broker discovery via IPC (only fall back to AccountManager) for the next 60 minutes.")
```
shahzaibj, line 116: We expect it to work again after 60 minutes?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
rpdome: @shahzaibj The feature might be turned on again in the next 60 mins (meaning, try the broker discovery again when that time comes. if it's still disabled, then wait for another 60 mins before the next try)

In the meantime - don't bother trying. just use accountmanager.

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
rpdome: yes. in this PR https://github.com/AzureAD/ad-accounts-for-android/pull/2485

PR #2146, opened 2023-08-15
PasskeyFidoChallengeHandler and associated classes

melissaahn:
## Summary
The majority of this PR consists of new classes and their associated unit tests. The new classes are wired up, but are guarded by a flag until the passkey feature release.
It also includes a new dependency which allows us to use lifecycle coroutine scopes, as well as a test dependency ([JSONAssert](https://jsonassert.skyscreamer.org/cookbook.html)) which allows for cleaner comparisions between JSON objects in Junit tests.

The general steps after user clicks "use a passkey/security key" (or some string like that) in the WebView:
1. Server sends a redirect uri, which our AzureActiveDirectoryWebViewClient will identify as a passkey protocol redirect uri.
2. We first try to create the challenge handler. Then we try to create the challenge from the query string parameters in the uri. 
3. We will process the challenge by creating a Json string request in a particular WebAuthn format and invoking the Fido API (which will be Credential Manager, but that implementation will come in the last PR for this feature), which will guide the user through choosing a passkey and providing their biometrics/system PIN. 
4. If successful, we will receive the results back as a JSON string response, which we will parse out to get the object that we need to return to the server. If not successful (exception thrown), we will get the error message details. (Future work will involve putting the exception in a certain format that the server expects; this format is in discussion).
5. The results will be sent back to the server via an Assertion header.
6. User will either be logged in, or they will see an error message and will be asked to try again.

### AuthFidoChallengeHandler
`AuthFidoChallengeHandler` takes in `FidoChallenge` into `processChallenge` and handles getting the assertion from the `FidoManager`, which is run in a lifecycle coroutine scope. The class also has a `respondToChallenge` method which handles responding back to the server with the expected header values. 
Unit tests were created for `AuthFidoChallengeHandler` using test implementations of `IFidoManager` and `WebView`. Additional unit tests were created for `WebAuthnJsonUtil`.

### LifecycleScope
We'll need to run the CredMan API logic in a coroutine scope, so I chose to use a Lifecycle scope so that the passkey logic is tied to the lifecycle of the WebView. For more information: https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope
The dependencies added are `androidx.lifecycle:lifecycle-runtime-ktx` and `androidx.lifecycle:lifecycle-runtime-testing`.
The addition of the lifecycle dependency will not bring us above the current limit we've set for dependency size for common (10.57 MB -> 10.66 MB, from dependenciesSizeCheck).

### Wiring up in AzureActiveDirectoryWebViewClient
The passkey protocol request will come in as a redirect uri. We'll identify the uri in the handleUrl method and create the appropriate classes (challenge handler, challenge, etc.) to handle the challenge and respond back to the server. 

### Upcoming Work
- Actual implementation of `CredManFidoManager` will be done in a future PR, as we can't add in the Credential Manager API until our main consumer apps have bumped their compileSdk versions to 34 (a requirement of Credential Manager).
- Discussion is still occurring as to in what format we should be sending exceptions/errors back to the server. Currently I am just sending the error string/message in the Assertion header, which will not crash anything, but result in a "invalid assertion" message from the server. 
- Very basic telemetry has been added to the challenge handler, and more investigation will be needed to see if it is possible to determine other information from the WebAuthn response (current investigations on my part have determined that this might not be possible from the response alone).


common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoRequestField.kt
```diff
@@ -25,7 +25,9 @@ package com.microsoft.identity.common.internal.fido
 /**
  * Request fields specific to an Auth FIDO request.
  */
-enum class AuthFidoRequestField {
-    AllowedCredentials,
-    KeyTypes
+class AuthFidoRequestField {
```
melissaahn: Changing class type, since enums are capitalized by convention. (Had to change the values to match what the server side implemented for the protocol)

common/src/main/java/com/microsoft/identity/common/internal/fido/IFidoManager.kt
```diff
@@ -22,16 +22,15 @@
...
 package com.microsoft.identity.common.internal.fido
 
-import android.webkit.WebView
-import com.microsoft.identity.common.internal.ui.webview.challengehandlers.IChallengeHandler
-
 /**
- * Abstract class that handles a FidoChallenge.
+ * Representation of a manager that handles interactions with a passkey provider (usually through an API).
  */
-abstract class AbstractFidoChallengeHandler
```
melissaahn: This is an odd way that Git is depicting the change here, but I deleted the AbstractFIdoChallengeHandler class (don't need to shadow ChallengeHandler and we will be using the passkey challenge handler for security key support as well) and I'm creating a new IFidoManager interface,

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -0,0 +1,42 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.fido
+
+import android.content.Context
+
+/**
+ * Makes calls to the Android Credential Manager API in order to return an attestation.
+ */
+class CredManFidoManager (val context: Context) : IFidoManager {
```
melissaahn, line 30: This is a placeholder for the actual implementation of the CredManFidoManager class, which will be adding in a future PR.

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoRequestField.kt
```diff
@@ -25,11 +25,13 @@ package com.microsoft.identity.common.internal.fido
...
  * Represents FIDO request fields shared among registration and authentication requests.
  */
-enum class FidoRequestField {
-    Challenge,
-    RelyingPartyIdentifier,
-    UserVerificationPolicy,
-    Version,
-    SubmitUrl,
-    Context
+class FidoRequestField {
```
melissaahn: Changing class type, since enums are capitalized by convention. (Had to change the values to match what the server side implemented for the protocol)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -160,6 +171,36 @@ private boolean handleUrl(final WebView view, final String url) {
                 final PKeyAuthChallenge pKeyAuthChallenge = factory.getPKeyAuthChallengeFromWebViewRedirect(url);
                 final PKeyAuthChallengeHandler pKeyAuthChallengeHandler = new PKeyAuthChallengeHandler(view, getCompletionCallback());
                 pKeyAuthChallengeHandler.processChallenge(pKeyAuthChallenge);
+            } else if (FidoConstants.IS_PASSKEY_SUPPORT_READY && isPasskeyUrl(formattedURL)) {
```
melissaahn, line 171: The IS_PASSKEY_SUPPORT_READY constant is currently set to false, so we won't hit this logic until the feature is ready for prod.

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoRequestField.kt
rpdome: you can actually do something like this

```kotlin
enum class Color(val rgb: Int) {
    RED(0xFF0000),
    GREEN(0x00FF00),
    BLUE(0x0000FF)
}
```
(from https://kotlinlang.org/docs/enum-classes.html)

rpdome: (but this is ok too).

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+                    assertion = e.message.toString()
+                    Logger.error(methodTag, assertion, e)
+                    span.setStatus(StatusCode.ERROR)
+                }
+                respondToChallenge(fidoChallenge.submitUrl, assertion, fidoChallenge.context)
+            }
+            return null
+        }
+        var errorMessage = "Failed based on parameter values."
+        if (lifecycleOwner == null) {
```
rpdome: nit: early return

```kotlin
   if (lifecycleOwner == null) {
       respondToChallenge(error)
       return
   }

   lifecycleOwner.lifecycleScope.launch {
        ...
   }

```

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+     * Makes a post request in the WebView with the submitUrl and headers.
+     * @param submitUrl The url to which the client submits the response to the server's challenge.
+     * @param context Server state that needs to be maintained between challenge and response.
+     * @param assertion string representing response with signed challenge.
+     */
+    fun respondToChallenge(submitUrl: String,
+                           assertion: String,
+                           context: String) {
+        val methodTag = "$TAG:respondToChallenge"
+        span.end()
```
rpdome: would it be possible that the provided parameters are malformed, and an error is thrown in this method?

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
```diff
@@ -0,0 +1,129 @@
...
+                }
+                respondToChallenge(fidoChallenge.submitUrl, assertion, fidoChallenge.context)
+            }
+            return null
+        }
+        var errorMessage = "Failed based on parameter values."
+        if (lifecycleOwner == null) {
+            errorMessage += " Cannot get lifecycle owner needed for FIDO API calls."
+        }
+        if (fidoChallenge !is AuthFidoChallenge) {
```
rpdome: I'm wondering if we need this check.

If this class only supports AuthFidoChallenge, then maybe we should just pass AuthFidoChallenge here in the first place (not IFidoChallenge)

essentially turning this class into IChallengeHandler<AuthFidoChallenge, Void>

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -160,6 +171,36 @@ private boolean handleUrl(final WebView view, final String url) {
...
+                final SpanContext spanContext = getActivity() instanceof AuthorizationActivity ? ((AuthorizationActivity)getActivity()).getSpanContext() : null;
+                final PasskeyFidoChallengeHandler challengeHandler = new PasskeyFidoChallengeHandler(
+                        new CredManFidoManager(view.getContext()),
+                        view,
+                        spanContext,
+                        ViewTreeLifecycleOwner.get(view));
+                try {
+                    final IFidoChallenge challenge = FidoChallengeFactory.createFidoChallengeFromRedirect(url);
+                    challengeHandler.processChallenge(challenge);
+                } catch (final ClientException e) {
```
rpdome: Should we log any error?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -160,6 +171,36 @@ private boolean handleUrl(final WebView view, final String url) {
...
+                    // Server wants to know if there's an incorrect query string parameter field/name sent by them,
+                    // so we need to validate and get submitUrl and context first in order to send those related exceptions to them.
+                    // If the submitUrl or context field themselves are incorrect... another ClientException will be thrown and the WebView will close.
+                    final Map<String, String> parameters = UrlUtil.getParameters(url);
```
rpdome: I'm wondering if these check should be in PasskeyFidoChallengeHandler instead.

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
rpdome: and if you get something else in the caller layer - you don't even need to create this object in the first place, just fail.

common/src/main/java/com/microsoft/identity/common/internal/fido/IFidoManager.kt
```diff
@@ -22,16 +22,15 @@
...
-    val webView: WebView
-) : IChallengeHandler<IFidoChallenge, Void>
+interface IFidoManager {
+    /**
+     * Interacts with the FIDO credential provider and returns an assertion.
+     * @param challenge AuthFidoChallenge received from the server.
+     * @return assertion
+     * @throws Exception
+     */
+    suspend fun authenticate(challenge: AuthFidoChallenge): String
```
rpdome: IFidoChallenge?

melissaahn: IFidoChallenge currently only has AuthFidoChallenge implementing it, but RegFidoChallenge could be an object in the future (original plan was to implement both registration and authentication, but the plans for native registration have been dropped for the time being). So my thinking was to have an authenticate method which only takes AuthFidoChallenges, and a register method which only takes RegFidoChallenges.

Unless I'm missing some fundamental design principle here.

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoRequestField.kt
melissaahn: Going to look into changing these back into enums...

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
melissaahn: This class currently only supports AuthFidoChallenge, but the intent was to have it support both authentication and registration challenges (something like RegFidoChallenge) in the future, whenever we're asked to implement native registration support. I think I put this weird check because I was getting a warning about it somewhere...

Should I be implementing this class as if we don't know about registration yet? Technically I don't know when we would be asked to take on that work, anyway.

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
melissaahn: I should probably account for the submitUrl being malformed. I think additional verification of the assertion and context can/should be done on the server side.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: That would be a lot cleaner, and this might be indicative of an earlier design issue...
We would get to this catch block upon an exception getting thrown in the creation of a FidoChallenge (line 183; and see this method: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/2bbd690317190157e980049bc95ee5bf2dc665d3/common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt#L45) 
All of the verification of the challenge's fields are done upon challenge creation. So with the current factory logic, any verification issues with the challenge gets thrown outside of the challenge handler. 

Do you think it would be better to just create the challenge with the redirect uri parameters as-is, and then do the verification of those parameters in the beginning of the processChallenge method of the challenge handler?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: Will add a log statement here.

common/src/main/java/com/microsoft/identity/common/internal/fido/IFidoManager.kt
rpdome: Is there any scenario where AuthFidoChallenges and RegFidoChallenges would be used interchangably?

if not, maybe we don't need IFidoChallenge.

(Interface = abstraction = meant to be used interchangably regardless of implementation details)

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
rpdome: I would expect the Registration flow to be in a separate challenge handler.
(Their algorithms are different, right?)

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoRequestField.kt
melissaahn: Changes made here: 2112dd72efe46ae748ab0b2525b21810e46df28a

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoRequestField.kt
melissaahn: Changes made here: 2112dd72efe46ae748ab0b2525b21810e46df28a

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
rpdome: If I assume that the challenge content might greatly vary (a required field in Auth flow might not be required in Reg flow, and vice versa), having its own flow validate their own values (URI) might make more sense.

If that's the case, I could even have a challenge class that simply takes in the url as-is. URI can be broken down and scrutinized separately in each flow.

You would need a factory for PasskeyFidoChallengeHandler instead (to determine if it's auth or reg, etc. would there ever be a hint for that?).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: Changes made here: 2112dd72efe46ae748ab0b2525b21810e46df28a

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
melissaahn: Changes made here: 2112dd72efe46ae748ab0b2525b21810e46df28a
Added a separate method to verify a malformed submitUrl. (this method, along with other verification methods, may be moved around depending on some other comments)

common/src/main/java/com/microsoft/identity/common/internal/fido/IFidoManager.kt
melissaahn: They wouldn't be used interchangeably, but the two of them do share many fields (such as the ones defined in IFidoChallenge). This was the reason why I created an interface.

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
melissaahn: Changes made here: 2112dd72efe46ae748ab0b2525b21810e46df28a

common/src/main/java/com/microsoft/identity/common/internal/fido/PasskeyFidoChallengeHandler.kt
melissaahn: The parameters and method calls are different for reg and auth. Makes sense to have different challenge handlers. 

( for some context: The original design considered both auth and reg, and at the time, we didn't know when CredMan would start supporting security keys. So my initial idea was that:
- challenges would have auth and reg types
- challenge handlers would have a passkey and a securitykey type (each one would handle reg/auth), and the associated factory would figure out which one we need there.
Things have changed, and now we know that CredMan will eventually support both passkeys and security keys and will provide dialogs to choose between passkey and security key itself (for auth at least, not sure about reg). Regardless of the changes, I'm realizing that I still made it more complicated then it maybe needed to be  )

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: So a general FidoChallenge class that has all the parameters combined from reg and auth (maybe just auth for now, since reg will come much later)? 
Then the challenge handler factory will determine which handler to return (this would be tricky without some hint from the protocol, as some fields can be malformed, so that's something I would ask the server side to provide. for now, it'll just return auth without checks. )
Then the challenge handler would do verification (not null, not empty, valid submiturl)

Iiuc, then I'll move the verification stuff from the challenge factory to the challenge handler itself.

melissaahn: Ended up doing some refactoring:
- Created FidoChallenge, which breaks down the uri but does no validation
- moved validation to AuthFidoChallengeHandler
- Moved and rewrote a lot of tests

I'm not sure if a factory for FidoChallengeHandler is needed right now, but if it would be good practice to have one, I can add it in a different PR (to avoid further increasing the size of this one)

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
```diff
@@ -0,0 +1,272 @@
...
+import io.opentelemetry.api.trace.StatusCode
+import kotlinx.coroutines.launch
+import java.net.MalformedURLException
+import java.net.URL
+
+/**
+ * Handles a FidoChallenge by either creating or authenticating with a passkey.
+ */
+class AuthFidoChallengeHandler
+/**
```
rpdome: nit: put param list in the comment section above line 44.

rpdome: also add tabs between the param name and description.

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
```diff
@@ -0,0 +1,272 @@
...
+ * @param spanContext Current spanContext, if present.
+ * @param lifecycleOwner instance to get coroutine scope from.
+ */(
+    private val fidoManager: IFidoManager,
+    private val webView: WebView,
+    private val spanContext : SpanContext?,
+    private val lifecycleOwner: LifecycleOwner?
+) : IChallengeHandler<FidoChallenge, Void> {
+    val TAG = AuthFidoChallengeHandler::class.simpleName.toString()
+    val span = if (spanContext != null) {
```
rpdome: does this need to be in processChallenge?

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * Validates that the given required parameter is not null or empty.
+     * @param field passkey protocol field
+     * @param value value for a passkey protocol parameter
+     * @return validated parameter value
+     * @throws ClientException if the parameter is null or empty.
+     */
+    @Throws(ClientException::class)
+    fun validateRequiredParameter(field: String, value: String?): String {
```
rpdome: you can do something like this

```Kotlin
        data class FidoChallengeField<K>(private val field: FidoRequestField,
                                         private val value: K?,
                                         private val throwIfValueIsInvalid: (FidoRequestField, K) -> Void) {
            
            @Throws(ClientException::class)
            fun getOrThrow(): K {
                if (value == null){
                    throw ClientException("$field is null")
                }

                throwIfValueIsInvalid(value)
                return value
            }
            
            companion object {
                @Throws(ClientException::class)
                fun throwIfInvalidUrl(field: FidoRequestField,
                                      url: String) {
                    try {
                        URL(url)
                    } catch (e : MalformedURLException) {
                        throw ClientException(
                            ClientException.PASSKEY_PROTOCOL_REQUEST_PARSING_ERROR,
                            "$field value is not a valid URL.")
                    }
                }
            }
        }

        data class FidoChallenge(
            val submitUrl: FidoChallengeField<String>,
            ...) {

            constructor(redirectUri: String) : this(
                 submitUrl = FidoChallengeField<String>(
                    FidoRequestField.SUBMIT_URL,
                    parameters[FidoRequestField.SUBMIT_URL.fieldName],
                    ::throwIfInvalidUrl
                ),
                ...
            )
        }
```
then
``` Kotlin
    fidoChallenge.submitUrl.getOrThrow()
```

I'll let you decide if you want to go that route.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -228,6 +246,10 @@ private boolean isPkeyAuthUrl(@NonNull final String url) {
         return url.startsWith(AuthenticationConstants.Broker.PKEYAUTH_REDIRECT.toLowerCase(Locale.ROOT));
     }
 
+    private boolean isPasskeyUrl(@NonNull final String url) {
+        return url.startsWith(FidoConstants.PASSKEY_PROTOCOL_REDIRECT.toLowerCase(Locale.ROOT));
```
rpdome, line 250: You mentioned the registration flow, would it use the same uri format?

(I assume we should never hit the scenario where old common get a registration uri - as that should only be triggered when the client says that it can handle registration).

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
melissaahn: Addressed here: 40f33a68e5d35e99684e0f93679acb9135abe05b

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
melissaahn: Resolved here: 40f33a68e5d35e99684e0f93679acb9135abe05b

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
melissaahn: Took your suggestion: 40f33a68e5d35e99684e0f93679acb9135abe05b

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: It most likely would, though I would ask the server team to update the version parameter to "2.0". With the current logic, if the version is anything other than "1.0", then we will send back an exception that tells them we don't support the protocol version they are using.

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
iamgusain: nit: Add a TODO and ADO item link for future work.

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallenge.kt
```diff
@@ -0,0 +1,74 @@
...
+    val keyTypes: FidoChallengeField<List<String>?>,
+    val allowedCredentials: FidoChallengeField<List<String>?>
+) {
+    companion object {
+        const val DELIMITER = ","
+        const val DEFAULT_USER_VERIFICATION_POLICY = "required"
+
+        @JvmStatic
+        operator fun invoke(redirectUri: String): FidoChallenge {
+            val parameters = UrlUtil.getParameters(redirectUri)
```
iamgusain: nit: Can we add the type of the parameters, it makes is easier to read through code.
val parameters: <"type"> = UrltUtils.getP....

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallenge.kt
```diff
@@ -0,0 +1,74 @@
...
+    val context: FidoChallengeField<String>,
+    val keyTypes: FidoChallengeField<List<String>?>,
+    val allowedCredentials: FidoChallengeField<List<String>?>
+) {
+    companion object {
+        const val DELIMITER = ","
+        const val DEFAULT_USER_VERIFICATION_POLICY = "required"
+
+        @JvmStatic
+        operator fun invoke(redirectUri: String): FidoChallenge {
```
iamgusain: Does this method invoke the uri as well, or just parsing and getting the FidoChallenge components?
It looks like latter, should update the method name.

iamgusain (approved):
Left some minor comments

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: Done: a376703a0ab288a577766be373400e1e0628536f

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallenge.kt
melissaahn: Done: a376703a0ab288a577766be373400e1e0628536f

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallenge.kt
melissaahn: It's the latter, but invoke in this case refers to this Kotlin operator: https://kotlinlang.org/docs/operator-overloading.html#invoke-operator. I was trying to create another constructor that could do some preprocessing (which a secondary operator would not allow), but unfortunately the invoke() needs to be mentioned when calling from java.
Changed from an operator to a static fun createFromRedirectUri, for better clarity: a376703a0ab288a577766be373400e1e0628536f

PR #2151, opened 2023-08-21
FidoConstants

melissaahn:
### Summary
This PR creates the `FidoConstants` class, which currently has two constants related to the WebAuthn extra query parameter. This query parameter will be passed into MSAL by host apps when they want to declare that they are WebAuthn capable. 
The constants in this class will be used by msalTestApp, which will have a switch to add the query parameter.


common/src/main/java/com/microsoft/identity/common/internal/fido/FidoConstants.kt
```diff
@@ -0,0 +1,40 @@
...
+         * Extra query parameter field to declare WebAuthn capability for a host app.
+         */
+        const val WEBAUTHN_QUERY_PARAMETER_FIELD = "webauthn"
+
+        /**
+         * Extra query parameter value to declare WebAuthn capability for a host app.
+         */
+        const val WEBAUTHN_QUERY_PARAMETER_VALUE = "1"
+    }
+}
```
shahzaibj: nit: new line ending

shahzaibj (approved):

tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoConstants.kt
melissaahn: Done

PR #2153, opened 2023-08-23
Add new constant for new authapp redirect uri

fadidurah:
Need to add this to support safer broker uri for Authenticator.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -721,7 +721,7 @@ public final class AuthenticationConstants {
         public static final String BROKER_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob";
 
         /**
-         * String for new broker redirect URI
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -720,6 +720,11 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String BROKER_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob";
 
+        /**
+         * String for new broker redirect URI
```
shahzaibj:
```suggestion
         * String for new broker redirect URI.
```

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
shahzaibj: We also have such constant in broker repo. Perhaps that one should just reference this one now.

rpdome (approved):

PR #2157, opened 2023-08-28
Deprecate logger wrappers and update comments

fadidurah:
Deprecate logger wrappers in common to notify partners to use `com.microsoft.identity.common.java.logging.Logger` instead.

Not sure if this needs a changelog entry...

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
@@ -31,11 +31,12 @@ import com.microsoft.identity.common.java.telemetry.events.DeprecatedApiUsageEve
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
- * Deprecated.
- * <p>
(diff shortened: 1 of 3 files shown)


rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/logging/Logger.java
```diff
@@ -31,11 +31,12 @@
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
- * Deprecated.
- * <p>
  * This is now acting as an adapter for {@link com.microsoft.identity.common.java.logging.Logger}.
+ * <p>
+ * Deprecated: Broker partners should use {@link com.microsoft.identity.common.java.logging.Logger}
```
rpdome: duplicate

shahzaibj (approved):
Let's put a changelog entry

fadidurah:
Seems we pushed some changelog entries incorrectly, moved everything that wasn't include in 15.0.0 into V.NEXT

PR #2158, opened 2023-08-28
Catch strongbox exception on Android 14

somalaya:
**Issue** : This is related to the fix we did in the PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2053

The fix we did previously worked and Defender team confirmed the same. They were unblocked on Andorid 14. But, I started seeing the same issue again on Android 14 because the "Build.VERSION.RELEASE_OR_CODENAME" is different now. This is happening only on some devices (I think.. because, Defender team is not seeing this issue). Build.VERSION.RELEASE_OR_CODENAME was previously UpsideDownCake. Now it is coming as "14". 

**Fix** : 

- I have added a new condition to check for Build.VERSION.RELEASE_OR_CODENAME as 14 as well. 
- Build.VERSION.RELEASE_OR_CODENAME was only available on Android API level 30, so added a check for this.
- Also added some null checks.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
@@ -217,8 +217,7 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
                         trySetAttestationChallenge = false;
 
                         continue;
-                    } else if (tryStrongBox && (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
-                            && (Build.VERSION.RELEASE_OR_CODENAME.equals("UpsideDownCake") || Build.VERSION.RELEASE_OR_CODENAME.equals("14"))
+                    } else if (tryStrongBox && (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
                             && (null != e.getCause()) && isNegativeInternalError(e.getCause())) {
                         // Android 14 specific error where strong box is failing, most likely because of IAR requirement in android 14
                         // https://android.googlesource.com/platform/compatibility/cdd/+/e2fee2f/9_security-model/9_11_keys-and-credentials.md
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
@@ -83,8 +83,7 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
 
     /**
      * Seeing this on android 14, we think it's being caused by the new IAR requirement
-     *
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -217,7 +217,9 @@ private KeyPair generateNewRsaKeyPair(@androidx.annotation.NonNull final Context
                         trySetAttestationChallenge = false;
 
                         continue;
-                    } else if (tryStrongBox && Build.VERSION.RELEASE_OR_CODENAME.equals("UpsideDownCake") && isNegativeInternalError(e.getCause())) {
+                    } else if (tryStrongBox && (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
+                            && (Build.VERSION.RELEASE_OR_CODENAME.equals("UpsideDownCake") || Build.VERSION.RELEASE_OR_CODENAME.equals("14"))
```
shahzaibj: Why do we need RELEASE_OR_CODENAME at all? Can't we just rely on VERSION.SDK_INT? I think that should cover all cases, no?

somalaya: I think it should be sufficient to use VERSION.SDK_INT unless I am missing something. @fadidurah  do we have any reason to use Build.VERSION.RELEASE_OR_CODENAME? I think why this was added previously was because we wanted to restrict this to the beta/development versions.. and find the right fix. But now that this is going to be applied to a release code name as well, we should use VERSION.SDK_INT instead.

fadidurah: I believe in my initial investigation, VERSION.SDK_INT would return 33, most likely that was related to the Beta of android 14. I'm guessing this would change in Android 14 proper release. If we wanted to be safe, we can include that as well in the or statement.

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -205,7 +205,7 @@ private KeyPair generateNewRsaKeyPair(@androidx.annotation.NonNull final Context
                         Logger.error(TAG, "Import unsupported. Skipping import flag then retry.", e);
                         tryImport = false;
 
-                        if (tryStrongBox && null != e.getCause() && isStrongBoxUnavailableException(e.getCause())) {
+                        if (tryStrongBox && null != e.getCause() && (isStrongBoxUnavailableException(e.getCause()) || isNegativeInternalError(e.getCause()))) {
```
fadidurah, line 208: Is this negative internal error check necessary?

somalaya: Yes, I cannot reproduce the issue now but I saw that when I removed this, i started seeing some errors. I believe there is no harm in keeping it as it is only adding a condition to skip strongbox when an unknown error occurs

shahzaibj (approved):

PR #2159, opened 2023-08-28
Add key to request the broker data

p3dr0rv:
Add a key to request broker data

We are setting this key in ContentProviderStrategy so all MSAL, MWPJ and Broker API request, can update the broker information.

related PR https://github.com/AzureAD/ad-accounts-for-android/pull/2507

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
@@ -71,6 +71,12 @@ class ActiveBrokerCacheUpdater(
          */
         const val BROKER_DISCOVERY_DISABLED_KEY = "broker.discovery.disabled"
 
+        /**
+         * If this key is present in the request bundle,
+         * The broker will return the active broker information in the result bundle.
+         * Note: This is only valid for requests that are processed by the BrokerOperationRequestDispatcher.
+         */
+        const val REQUEST_ACTIVE_BROKER_DATA_KEY = "com.microsoft.identity.request.broker.data"
 
         /**
          * Adds the active broker information to the result bundle.
@@ -125,6 +131,7 @@ class ActiveBrokerCacheUpdater(
         val signingCertThumbprint = bundle.getString(ACTIVE_BROKER_SIGNING_CERTIFICATE_THUMBPRINT_KEY)
 
         if (packageName.isNullOrEmpty() || signingCertThumbprint.isNullOrEmpty()) {
+            Logger.info(methodTag, "A response was received without active broker information.")
             return
         }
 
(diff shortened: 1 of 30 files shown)


p3dr0rv:
@rpdome, I wonder if we should also include the legacy WPJ? or just wait for multiple-wpj to be adopted, anyway this is just an optimization for edge cases

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
```diff
@@ -232,6 +234,9 @@ private <T> T performStrategy(@NonNull final IIpcStrategy strategy,
             span.setAttribute(AttributeName.ipc_strategy.name(), strategy.getType().name());
             operation.performPrerequisites(strategy);
             final BrokerOperationBundle brokerOperationBundle = operation.getBundle();
+            if (brokerOperationBundle.getBundle() != null) {
```
rpdome: if bundle is null, create an empty bundle with the flag?

rpdome (approved):

rpdome:
> @rpdome, I wonder if we should also include the legacy WPJ? or just wait for multiple-wpj to be adopted, anyway this is just an optimization for edge cases

no need to do for legacy wpj - as it doesn't use the new selection mechanism to begin with.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerOperationExecutor.java
p3dr0rv: bundle is final so I can move it inside brokerOperationBundle

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
```diff
@@ -71,6 +71,12 @@ class ActiveBrokerCacheUpdater(
          */
         const val BROKER_DISCOVERY_DISABLED_KEY = "broker.discovery.disabled"
 
+        /**
+         * If this key is present in the request bundle,
+         * The broker will return the active broker information in the result bundle.
+         * Note: This is only valid for requests that are processed by the BrokerOperationRequestDispatcher.
+         */
+        const val REQUEST_ACTIVE_BROKER_DATA_KEY = "com.microsoft.identity.request.broker.data"
```
shahzaibj: For all such keys, I think renaming them such that name starts with the word "KEY" might be better for readability. For example: `KEY_REQUEST_ACTIVE_BROKER_DATA`

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
```diff
@@ -125,6 +131,7 @@ class ActiveBrokerCacheUpdater(
         val signingCertThumbprint = bundle.getString(ACTIVE_BROKER_SIGNING_CERTIFICATE_THUMBPRINT_KEY)
 
         if (packageName.isNullOrEmpty() || signingCertThumbprint.isNullOrEmpty()) {
+            Logger.info(methodTag, "A response was received without active broker information.")
```
shahzaibj, line 134: Should we also log if it was requested to begin with?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/ActiveBrokerCacheUpdater.kt
p3dr0rv: done

PR #2163, opened 2023-08-30
Undeprecate Logger wrapper

fadidurah:
We didn't really need to deprecate this in the first place, as we'd already asked CP to use the correct logger. This is causing a warning (treated as error) in oneauth test app build, so i'm un-deprecating it.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 V.Next
 ----------
-- [MINOR] Deprecate Common logger wrappers (#2157)
+- [MINOR] Deprecate Common logger wrapper (#2157)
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,6 +1,6 @@
 V.Next
 ----------
-- [MINOR] Deprecate Common logger wrappers (#2157)
+- [MINOR] Deprecate Common logger wrapper (#2157)
```
somalaya, line 3: do we still need this changeLog?

fadidurah: Yeah, technically one other wrapper was deprecated

somalaya (approved):

rpdome (approved):

PR #2164, opened 2023-08-31
Separate Broker API and Client SDK Active Broker caches

rpdome:
In broker apps - MSAL and Broker might use different encryption/decryption keys.

This means the same active broker cache cannot be shared by MSAL, and the Broker API/WPJ API **_under the same app_**.

as each library will write data encrypted by its own key, and the other party will not be able to decrypt.

As a solution - we're having them write to different storage files.

--

The issue was discovered by @somalaya (thanks!) and the fix is already validated.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 V.Next
 ----------
+- [MAJOR] Separate Broker API and Client SDK Active Broker caches (#2164)
 - [MINOR] Deprecate Common logger wrapper (#2157)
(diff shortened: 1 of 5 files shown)


somalaya (approved):

somalaya:
you may have to make Assemble consumers pipeline run as optional. I think it is expected to fail

mohitc1 (approved):
:shipit:

PR #2168, opened 2023-09-06
Nested app auth implementation

somalaya:
- Added parameters for brkRedirectId and brkClientId to pass in /auth and /token requests for NAA request
- Min required protocol version for NAA is 15
- Made changes to NOT save tokens when the flow is an NAA request. We have designed NAA flow to not save tokens as OneAuth is anyway saving the AT acquired and requests broker only when AT is expired. When AT is expired, we anyway have to make a fetch AT with PRT or  RT request. So, saving the tokens in the cache would not be useful

Design PR link : https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/7876


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -54,6 +54,8 @@ private static final class SerializedNames {
         final static String SCOPE = "scopes";
         final static String REDIRECT = "redirect_uri";
         final static String CLIENT_ID = "client_id";
+        final static String CHILD_REDIRECT = "child_redirect_uri";
```
mohitc1: nit: CHILD_REDIRECT_URI

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,6 +1170,14 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
                     "The min broker protocol version for PopAuthenticationSchemeWithClientKey should be equal or more than 11.0."
                             + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
         }
+
+        // Nested app auth is only supported when required protocol version is >=15
+        if ((parameters.getChildClientId() != null || parameters.getChildRedirectUri() != null)
```
mohitc1: >parameters.getChildClientId() != null || parameters.getChildRedirectUri() != null) [](http://example.com/codeflow?start=13&length=82)

isNullOEmpty check?

mohitc1: >(parameters.getChildClientId() != null || parameters.getChildRedirectUri() != null) [](http://example.com/codeflow?start=12&length=83)

Should we fail the request if only one of them is set (invalid state)?

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/TokenRequest.java
```diff
@@ -173,29 +174,29 @@ public void setClientId(final String clientId) {
...
      */
-    public String getChildClientId() {
-        return mChildClientId;
+    public String getBrkClientId() {
+        return mBrkClientId;
     }
 
     /**
-     * @param childClientId child client ID of the token request.
+     * @param brkClientId child client ID of the token request.
```
mohitc1: >child client ID of the token request. [](http://example.com/codeflow?start=26&length=37)

Update doc here and other places.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -155,6 +157,18 @@ private static final class SerializedNames {
...
+     * The redirect uri for the request.
+     */
+    @SerializedName(SerializedNames.CHILD_REDIRECT)
+    private String mChildRedirect;
+
+    /**
+     * The client id of the application.
+     */
+    @SerializedName(SerializedNames.CHILD_CLIENT_ID)
+    private String mChildClientId;
```
mohitc1, line 172: here too.

mohitc1:
Also consider UTs for these changes.

somalaya:
> Also consider UTs for these changes.

UTs for this flow are covered in the broker PR https://github.com/AzureAD/ad-accounts-for-android/pull/2523/files

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
somalaya: yes, I think it makes sense to fail the request immediately here if only 1 of them is set.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -262,9 +265,17 @@ protected final AuthorizationRequest.Builder initializeAuthorizationRequestBuild
...
 
-        builder.setClientId(parameters.getClientId())
-                .setRedirectUri(parameters.getRedirectUri());
 
+        if (parameters.getChildRedirectUri() != null && parameters.getChildClientId() != null) {
+            // is NAA scenario, set all NAA params for auth flow
+            builder.setBrkRedirectUri(parameters.getRedirectUri());
+            builder.setClientId(parameters.getChildClientId())
+                    .setRedirectUri(parameters.getChildRedirectUri());
+            builder.setBrkClientId(parameters.getClientId());
```
mohitc1: builder.setBrkRedirectUri(parameters.getRedirectUri())
                    .setBrkClientId(parameters.getClientId());
                    .setClientId(parameters.getChildClientId())
                    .setRedirectUri(parameters.getChildRedirectUri())

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -604,6 +621,48 @@ public TokenResult renewAccessToken(@NonNull final SilentTokenCommandParameters
...
+                                                            @NonNull final MicrosoftStsAuthorizationRequest authorizationRequest) {
+        final MicrosoftStsAccountCredentialAdapter credentialAdapter = new MicrosoftStsAccountCredentialAdapter();
+        final AccountRecord accountRecord = credentialAdapter.createAccount(
+                oAuth2Strategy,
+                authorizationRequest,
+                microsoftStsTokenResponse
+        );
+
+        final AccessTokenRecord accessTokenRecord;
+        accessTokenRecord = credentialAdapter.createAccessToken(
```
mohitc1: nit: assign directly

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -734,6 +792,12 @@ protected TokenResult performSilentTokenRequest(
 
         final TokenRequest refreshTokenRequest = strategy.createRefreshTokenRequest(parameters.getAuthenticationScheme());
         refreshTokenRequest.setClientId(parameters.getClientId());
+        if (parameters.getChildClientId() != null) {
+            refreshTokenRequest.setClientId(parameters.getChildClientId());
+            refreshTokenRequest.setBrkClientId(parameters.getClientId());
+            refreshTokenRequest.setRedirectUri(parameters.getChildRedirectUri());
+            refreshTokenRequest.setBrkRedirectUri(parameters.getRedirectUri());
+        }
```
mohitc1: if else ?

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -734,6 +792,12 @@ protected TokenResult performSilentTokenRequest(
 
         final TokenRequest refreshTokenRequest = strategy.createRefreshTokenRequest(parameters.getAuthenticationScheme());
         refreshTokenRequest.setClientId(parameters.getClientId());
+        if (parameters.getChildClientId() != null) {
```
mohitc1: Add null or empty check and consistent with authorization builder, where you check for both child redirect uri and child client id?

mohitc1 (approved):
left some comments

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: Added a method in ComandParameters to check for these parameters so that it is consistent across all classes

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -733,7 +790,17 @@ protected TokenResult performSilentTokenRequest(
         }
 
         final TokenRequest refreshTokenRequest = strategy.createRefreshTokenRequest(parameters.getAuthenticationScheme());
-        refreshTokenRequest.setClientId(parameters.getClientId());
+
+        if (parameters.hasNestedAppParameters()) {
+            // isNAA request, set hub/brk and nested app parameters
+            refreshTokenRequest.setClientId(parameters.getChildClientId());
+            refreshTokenRequest.setBrkClientId(parameters.getClientId());
```
shahzaibj, line 798: I believe Brkr here is the hub like Teams Native app? and client id (child) is the embedded web app?

somalaya: yes child app can be any Single page app for now

shahzaibj: Not a being fan of the parameter naming here. The `Brkr` to me means the Android Broker, but over here I believe these are the hub apps. Should these be renamed to `HubClientId`?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,11 +1173,29 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
                             + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
         }
+
+        if (parameters.hasNestedAppParameters()) {
+            // Nested app auth is only supported when required protocol version is >= 15
+            if (!StringUtil.isNullOrEmpty(parameters.getChildClientId()) && !StringUtil.isNullOrEmpty(parameters.getChildRedirectUri())
+                    && !BrokerProtocolVersionUtil.canSupportNestedAppAuthentication(requiredProtocolVersion)) {
+                throw new ClientException(ClientException.NESTED_APP_AUTH_NOT_SUPPORTED,
+                        "The min broker protocol version for Nested app auth should be equal or more than 15.0."
+                                + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
```
rpdome: nit: use requiredProtocolVersion variable?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -239,7 +241,7 @@ String hello(final @NonNull IIpcStrategy strategy,
                 clientMaxProtocolVersion
         );
 
-        if (!StringUtil.isEmpty(cachedProtocolVersion)) {
+        if (!isEmpty(cachedProtocolVersion)) {
```
rpdome: nit: use common's StringUtil.isNullOrEmpty() instead for consistency throughout the class?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -155,6 +157,18 @@ private static final class SerializedNames {
     @SerializedName(SerializedNames.PROMPT)
     private String mPrompt;
 
+    /**
+     * The child redirect uri for the request.
```
rpdome, line 161: nit: probably good to leave a pointer to your NAA Design doc (or leave a brief description here).

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -206,6 +210,12 @@ public String getPackageNameFromUid(int uid) {
         return mContext.getPackageManager().getNameForUid(uid);
     }
 
+    private boolean isValidRedirectURIForNAA(String redirectUri) {
```
rpdome: nit: rename to indicate that the method is specifically used for testing.

you can even put BuildConfig.DEBUG here.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -437,20 +448,26 @@ protected void renewAccessToken(@NonNull final SilentTokenCommandParameters para
...
-            @SuppressWarnings(WarningType.unchecked_warning) final List<ICacheRecord> savedRecords = tokenCache.saveAndLoadAggregatedAccountData(
-                    strategy,
-                    getAuthorizationRequest(strategy, parameters),
-                    tokenResult.getTokenResponse()
-            );
-
-            final ICacheRecord savedRecord = savedRecords.get(0);
+            List<ICacheRecord> acquireTokenResultRecords = null;
+            if (parameters.hasNestedAppParameters()) {
+                // Do not save the token in naa flow
```
rpdome: nit: leave a brief reason on this design decision, so that the next engineer who's refactoring this knows what (not) to do.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -262,9 +265,17 @@ protected final AuthorizationRequest.Builder initializeAuthorizationRequestBuild
             Logger.error(TAG, "correlation id from diagnostic context is not a UUID", ex);
         }
 
-        builder.setClientId(parameters.getClientId())
-                .setRedirectUri(parameters.getRedirectUri());
 
+        if (parameters.hasNestedAppParameters()) {
```
rpdome, line 269: Does NAA require PRT?

I'm wondering if there should be a subclass specifically for NAA.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
rpdome: My understanding is that brkr is what ests uses... but I do agree that it's confusing given our context.

somalaya: I could change this to HubClientId but it would be inconsistent with ESTS's convention. They call it the brkClientId because they consider NAA requests coming from broker as double broker protocol requests. which means the hub is acting as the first broker.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: I tried adding a subclass for NAA. But it needs a lot of refactoring of classes than I expected. I added details on the broker PR (on your comment) on what is blocking me. I will leave this as it is for now as the changes under the if block for NAA are very small. I will take this up refactoring work as a different task parallelly so that the review is blocked.
JFYI, These are the changes I am planning to do to address this comment : 
common : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/new/somalaya/NAAInterface
broker : https://github.com/AzureAD/ad-accounts-for-android/compare/somalaya/NAAInterface?expand=1

somalaya: To answer your first question, yes nested app tokens are also acquired using PRT, NAA work is particularly dependent on PRTV3. Hence you would see changes in this PR are dependent on PRTV3 flight

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -155,6 +157,18 @@ private static final class SerializedNames {
...
+     */
+    @SerializedName(SerializedNames.CHILD_REDIRECT_URI)
+    private String mChildRedirectUri;
```
shahzaibj, line 165: I thought the child one was being called as the normal client id/ redirect uri and the parent one gets the name of broker client id/redirect uri

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,11 +1173,29 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
                     "The min broker protocol version for PopAuthenticationSchemeWithClientKey should be equal or more than 11.0."
                             + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
         }
+
+        if (parameters.hasNestedAppParameters()) {
+            // Nested app auth is only supported when required protocol version is >= 15
+            if (!StringUtil.isNullOrEmpty(parameters.getChildClientId()) && !StringUtil.isNullOrEmpty(parameters.getChildRedirectUri())
```
shahzaibj, line 1182: Should we validate these parameters at a different layer and here just do the broker protocol version check? For  most things, the parameter validation is happening inside the command parameters

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,11 +1173,29 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+                    && !BrokerProtocolVersionUtil.canSupportNestedAppAuthentication(requiredProtocolVersion)) {
+                throw new ClientException(ClientException.NESTED_APP_AUTH_NOT_SUPPORTED,
+                        "The min broker protocol version for Nested app auth should be equal or more than 15.0."
+                                + " Current required version is set to: " + parameters.getRequiredBrokerProtocolVersion());
+            } else if ((StringUtil.isNullOrEmpty(parameters.getChildClientId()) || StringUtil.isNullOrEmpty(parameters.getChildRedirectUri()))) {
+                // If only one of the nested app params are sent
+                throw new ClientException(ClientException.NESTED_APP_INVALID_PARAMETERS, "One of the required parameters for nested app is null or empty");
```
shahzaibj: Can we be specific and tell the customers which one was actually null? We clearly know it

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,11 +1173,29 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+            }
+
+            if (parameters.getSdkType() != SdkType.MSAL_CPP) {
+                throw new ClientException(ClientException.NESTED_APP_INVALID_PARAMETERS, "Nested app auth is only supported for request originating from OneAuth");
```
shahzaibj, line 1197: Why? I think comments here would be helpful for future readers of the code

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1170,11 +1173,29 @@ private void verifyTokenParametersAreSupported(@NonNull final TokenCommandParame
...
+            }
```
shahzaibj: This whole thing can be its own method. `validateNestedAppAuthParameters`

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
shahzaibj: isValid**Hub**RedirectUriForNaa?

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -206,6 +210,12 @@ public String getPackageNameFromUid(int uid) {
...
+        // The only allow-listed hub app on ESTS is Teams app. We cannot use our test app's clientId/redirecrURI for testing NAA scenarios
+        // Below redirectURI is being used in our automation tests and also by OneAuth tests for NAA
+        return redirectUri.equals("msauth://com.microsoft.teams/VCpKgbYCXucoq1mZ4BZPsh5taNE=") || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno=");
```
shahzaibj: I guess one of them is prod signed and other debug signed?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -100,6 +100,8 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Inte
                 .scope(TextUtils.join(" ", parameters.getScopes()))
                 .redirect(parameters.getRedirectUri())
                 .clientId(parameters.getClientId())
+                .childRedirectUri(parameters.getChildRedirectUri())
```
shahzaibj, line 103: Not of a fan of the naming. Some places you've chosen a new name for the child i.e. the child and use existing parameter for the hub. Whereas other places you are using existing parameter for the child and use the new naming for the hub.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
```diff
@@ -116,6 +117,10 @@ public String getApplicationIdentifier(){
         return String.format(APPLICATION_IDENTIFIER_FORMAT, this.callerPackageName, this.callerSignature);
     }
 
+    public boolean hasNestedAppParameters() {
```
shahzaibj, line 121: javadoc on all public methods

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
```diff
@@ -116,6 +117,10 @@ public String getApplicationIdentifier(){
...
+        // return true even if one of the nested app params is present
```
shahzaibj, line 122: Will the request succeed with even one of these?

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -262,9 +265,17 @@ protected final AuthorizationRequest.Builder initializeAuthorizationRequestBuild
...
+            // is NAA scenario, set all NAA params for auth flow
+            builder.setBrkRedirectUri(parameters.getRedirectUri())
```
shahzaibj, line 271: See my comments in other places about naming

shahzaibj: My overall recommendation, let's pick one "new" parameter name. So if you want to use new name "child" for the child app then let's use that in all places. Or if you want to rather use new word "brkr" for the hub then let's do that in all places. Rather than introducing multiple new names.

My preference is the introduction of "child" to denote child apps. I suggest we use that everywhere. 

I know you may have to use different naming for what you send to eSTS. My suggestion is that you achieve that via SerializedNames

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
shahzaibj: Can you add more comments here indicating the rationale behind this?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
somalaya: This is similar to other validations happening in this method. We want to verify here if the child params are sent, then the protocol has to be > 15.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/CommandParameters.java
somalaya: it should not succeed. ESTS should throw an error. However, the reason why I am returning true even if one of the params are present is to specify that we found a nested app paramater and not to miss appending this parameter when doing the token request.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
somalaya: [Combining response to this comment and another one below]
From OneAuth's perspective or client's perspective, they send 
child parameters as childClientId and chileRedirectURI  and hub app's parameters as clientId and redirectURI.

We read these params in broker
1. While doing a silent call, we would not have to make any changes to do a cache lookup -- the lookup still happens with clientId and redirectURI to get RT (hub app's RT)
2. While doing an interactive call, we would have to send clientId,  redirectURI as child params and brkClientId, brkRedirectURI as hub parameters. 

So, in the changes I made, you would see only TokenRequest and JwtRequestBody would be having new brkClientId, brkRedirect parameters added.

If we instead change to the way you are suggesting - to only use childClientId and childRedirectURi and use serialzed to change the names, we would have to change TokenRequest to below

  @SerializedName(SerializedNames.REDIRECT_URI)
    private String mChildRedirectUri;

  @SerializedName(SerializedNames.BRK_REDIRECT_URI)
    private String mRedirectUri;

This is even more confusing.. and this would mean, even in existing requests (without NAA), wherever we are setting mRedirectURI, we would be sending brk_redirect_uri to the service.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
somalaya: yes

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
somalaya: Added reply to a similar comment above

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: responded to this in prev similar comment above

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
shahzaibj: Could we add a new method callled `createNaaTokenRequest` that can achieve this?

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
somalaya: Taking this up in the next PR I am creating for adding flighting for NAA

PR #2172, opened 2023-09-11
WebAuthnJsonUtil; giving default value for UserVerificationPolicy

melissaahn:
### Summary
There are three main parts to this PR:
- `WebAuthnJsonUtil` class: helps convert a `FidoAuthChallenge` to a JSON string which will be used as an input to the CredMan API in the future.
- New dependency and plugin, [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization): This was a recommended dependency for converting Kotlin classes into JSON. Gson is known to be slow, and Moshi has some extra dependencies needed specifically for Kotlin that include reflections, which will most likely increase size and decrease speed at runtime. Kotlinx.serialization is created for kotlin and (from the .jar size) is relatively lighter. The versions selected correspond to our use of kotlin 1.7.
- UserVerificationPolicy default value of "required": the server side let me know that even though UserVerificationPolicy has multiple possible values by WebAuthn standards, we should always expect it to be "required" for AAD and MSA, and thus, the server side will never send it to us in the protocol. I do want to still have the option available to set that property in case the server side changes their mind, or "required" doesn't end up being the only option in the future. So I added a helper method to set the property as "required" by default, but if the query parameter is present in the url, set it to whatever value that may be. 
### Related PRs
- https://github.com/AzureAD/android-complete/pull/251
- https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1899
- https://github.com/AzureAD/ad-accounts-for-android/pull/2532
- https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1757


common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -0,0 +1,78 @@
...
+            try {
+                return Json.encodeToString(request)
+            } catch (e: Exception) {
+                //This would be either SerializationException or IllegalArgumentException,
+                // and both are very unlikely to be thrown since we're only working with one class,
+                // but this catch block is here in case I'm wrong about that, or somethings changes
+                // in the future.
+                val errorMessage = "Error while encoding PublicKeyCredentialRequestOptions to string."
+                Logger.error(methodTag, errorMessage, e)
+                throw Exception(errorMessage + " " + e::class.simpleName + ": " + e.message)
```
shahzaibj: Can we just not throw the original exception back?

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -0,0 +1,78 @@
...
+
+            try {
+                return Json.encodeToString(request)
+            } catch (e: Exception) {
+                //This would be either SerializationException or IllegalArgumentException,
+                // and both are very unlikely to be thrown since we're only working with one class,
+                // but this catch block is here in case I'm wrong about that, or somethings changes
+                // in the future.
+                val errorMessage = "Error while encoding PublicKeyCredentialRequestOptions to string."
+                Logger.error(methodTag, errorMessage, e)
```
shahzaibj: Do you also want to capture telemetry? Or perhaps that is part of a separate PR? Or perhaps already being captured by a top-level span?

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialRequestOptions.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Representation of WebAuthn's PublicKeyCredentialRequestOptions.
+ * https://w3c.github.io/webauthn/#dictdef-publickeycredentialrequestoptions
+ */
+@kotlinx.serialization.Serializable
+data class PublicKeyCredentialRequestOptions(
+    val challenge: String,
+    val rpId: String,
+    val allowCredentials: List<PublicKeyCredentialDescriptor>,
+    val userVerification: String
+)
```
fadidurah, line 35: Nit: Empty line

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Representation of WebAuthn's PublicKeyCredentialDescriptor.
+ * https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor
+ */
+@kotlinx.serialization.Serializable
+data class PublicKeyCredentialDescriptor(
+    val type: String,
+    val id: String
+)
```
fadidurah, line 33: nit: empty line

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -0,0 +1,78 @@
...
+                throw Exception(errorMessage + " " + e::class.simpleName + ": " + e.message)
+            }
+        }
+    }
+}
```
fadidurah, line 70: Nit: empty line

common/src/test/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtilTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        )
+        val result = WebAuthnJsonUtil.createJsonAuthRequestFromChallengeObject(authChallenge)
+        assertEquals(result, expectedJsonOnlyRequiredFields)
+    }
+
+    //No tests created for missing required fields because
+    // the AuthFidoChallenge's required fields are non-null.
+    //Note that there's also a catch block to handle any exceptions thrown by kotlinx.serialization,
+    // but it is very unlikely that such an exception would be thrown given the tests above pass.
+}
```
fadidurah, line 78: one more

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import com.microsoft.identity.common.java.logging.Logger
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+
+/**
+ * A utility class to help convert to and from strings in WebAuthn json format.
+ */
+class WebAuthnJsonUtil {
```
fadidurah, line 33: I guess we're using this in future PRs? Or did i miss a usage in this one?

fadidurah (approved):
Just Nits

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: Looking at this part again, I think I will change this to throw the original exception back, at the very least for the reason that it's confusing without the rest of the PRs out yet. 
(Long answer for a short question, to explain thinking)
My current plan is to handle the exceptions, telemetry, and final result to the PasskeyChallengeHandler (not out for review yet), but I wanted to avoid the challenge handler from "knowing too much". Particularly that applies to CredMan, which is going to be housed in a CredManManager object and can throw various dependency-specific exceptions. I don't want the challenge handler to directly catch certain Exception types, but many of these exceptions don't have messages (not sure if that's because CredMan is in beta?), and the challenge handler wouldn't know what the exception was for without checking the exception's type. 
So, I was thinking that for dependency specific exceptions, I would catch them in their manager/util class, wrap them in a generic Exception with an actual message written by us, and throw that exception. That Exception would be caught by challenge handler, which will record telemetry including the message and then send that message to ESTS as an error response to the protocol.
I think a lot of this makes little sense without the PRs to demonstrate what I'm thinking :) so will remove the catch block of this method for now.

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: Going to be part of a separate PR.

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: It'll be used in future PRs, particularly ones that have to do with CredMan later.

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.kt
melissaahn: Checking all the new files in this PR, I think I do already have an empty line at the end of each. Unless I'm misunderstanding the comment here.

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
```diff
@@ -127,5 +129,23 @@ class FidoChallengeFactory {
...
+        /**
+         * If parameter is empty, replace parameter value with given default value.
+         * @param field passkey protocol field
+         * @param value value for a passkey protocol parameter
+         * @param defaultValue value to be used as default
+         * @return validated parameter value, or default value if initial value is null.
+         * @throws ClientException if the parameter is empty
+         */
+        @Throws(ClientException::class)
+        fun validateParameterOrReturnDefault(field: String, value: String?, defaultValue: String): String {
```
mohitc1: private or internal?

mohitc1: And other validate*methods

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2172/commits/1916a14746ab487e73be1feb62b47af64d40ee7c

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoChallengeFactory.kt
melissaahn: Added internal to all validate methods: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2172/commits/ada80ee79a65f8d71d54e39ce234c9ee3f7358b1

mohitc1 (approved):
:shipit:

PR #2178, opened 2023-09-14
Capture broker_operation_name on MSAL_PerformIpcStrategy span

p3dr0rv:
### Why

We have multiple entries of this error:
 "Attempt to invoke virtual method 'boolean com.microsoft.identity.broker4j.workplacejoin.data.WorkplaceJoinData.isSharedDevice()' on a null object reference"

android_spans
| where PipelineInfo_IngestionTime between (ago(7d) .. now())
| where error_message == "Attempt to invoke virtual method 'boolean com.microsoft.identity.broker4j.workplacejoin.data.WorkplaceJoinData.isSharedDevice()' on a null object reference"
| where broker_version == "11.2.1"

to identify the operation that is failing we start capturing the broker_operation_name 

Note:
In addition we also add 2 const used int this PR

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
@@ -394,7 +394,7 @@ public class ClientException extends BaseException {
     public static final String UNSUPPORTED_ANDROID_API_VERSION = "unsupported_android_api_version";
 
     /**
-     * The android version used does not support the operation.
+     * The workplacejoin data is null.
      */
     public static final String WORKPLACE_JOIN_DATA_NULL = "workplace_join_data_null";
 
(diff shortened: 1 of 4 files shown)


fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
```diff
@@ -393,6 +393,11 @@ public class ClientException extends BaseException {
      */
     public static final String UNSUPPORTED_ANDROID_API_VERSION = "unsupported_android_api_version";
 
+    /**
+     * The android version used does not support the operation.
```
shahzaibj: The javadoc doesn't seem to match with the actual String here. The actual string seems to indicate that something was null

p3dr0rv: update common

shahzaibj (approved):

PR #2179, opened 2023-09-14
Add LTW as prod broker app

rpdome:
Add LTW into the prod broker list

Also make a change in AccountManagerBrokerDiscoveryUtil such that LTW is not included there.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/AccountManagerBrokerDiscoveryUtil.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/AccountManagerBrokerDiscoveryUtil.kt
@@ -50,7 +50,9 @@ class AccountManagerBrokerDiscoveryUtil(
     }
 
     constructor(context: Context): this(
-        knownBrokerApps = BrokerData.getKnownBrokerApps(),
+        knownBrokerApps = BrokerData.getKnownBrokerApps().filter {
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/AccountManagerBrokerDiscoveryUtil.kt
```diff
@@ -50,7 +50,9 @@ class AccountManagerBrokerDiscoveryUtil(
     }
 
     constructor(context: Context): this(
-        knownBrokerApps = BrokerData.getKnownBrokerApps(),
+        knownBrokerApps = BrokerData.getKnownBrokerApps().filter {
+            BrokerData.isAccountManagerSupported(it.packageName)
+        }.toSet(),
```
somalaya, line 55: I think we are making this change to unblock Yixi for an issue with identifiying PROD LTW? But, with this line, it means that we are excluding LTW from knownBrokerApps right? 
How is this going to help with her testing?

rpdome: we don't want LTW to be included in the accountmanager based discovery.

this should not block her. (as she will be using the new discovery method).

fadidurah (approved):

somalaya (approved):

PR #2181, opened 2023-09-18
Add passkey protocol header

melissaahn:
### Summary
To enable the passkey protocol, we need to send a specific header (`x-ms-PassKeyAuth`) to the server. We only want to enable the protocol for WebView, so conditionals check that before adding the header. The protocol only applies to interactive auth scenarios, so it isn't going to be added in any silent auth requests. Also, since the protocol and the feature as a whole is not ready yet, the logic is blocked behind a boolean flag for now.
Once the protocol is ready for testing, I'll reevaluate and make sure that the header is being added to all places needed and is working as expected.

### Testing
Enabled the passkey protocol by setting the constant to true and tested with msalTestApp. Put breakpoints at the loadUrl call to make sure that the passkey header was present. Tested with aad cloud account and msa account (set authorization agent to "WEBVIEW" in msal_config.msa; saw that passkey header was included in the forwarding method created for msa). Tested that the results were the same with brokered auth as well.


common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
```diff
@@ -217,17 +218,23 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
...
             );
-            if (((MicrosoftTokenRequest) request).isPKeyAuthHeaderAllowed()) {
+            if (microsoftTokenRequest.isPKeyAuthHeaderAllowed()) {
                 headers.put(PKEYAUTH_HEADER, PKEYAUTH_VERSION);
             }
+            // Not going to add passkey protocol header until full feature is ready.
+            if (FidoConstants.IS_PASSKEY_SUPPORT_READY
+                    && microsoftTokenRequest.isPasskeyAuthHeaderAllowed()) {
+                headers.put(FidoConstants.PASSKEY_PROTOCOL_HEADER, FidoConstants.PASSKEY_PROTOCOL_HEADER_VALUE);
+            }
```
mohitc1: We need the header for /token endpoint as well?

melissaahn: I did ask if the header was needed for endpoints other than /authorize, and the response was that the header should be added for other calls involving interactive parameters as well... my plan is to test with the protocol once it's available in a test slice, and if the result is the same regardless of the header being present in /token calls, I'll ask the server team again about this.

common4j/src/main/com/microsoft/identity/common/java/constants/FidoConstants.kt
```diff
@@ -36,5 +36,33 @@ class FidoConstants {
          * Extra query parameter value to declare WebAuthn capability for a host app.
          */
         const val WEBAUTHN_QUERY_PARAMETER_VALUE = "1"
+
+        /**
+         * Header name to signal that the custom passkey protocol should be used.
+         */
+        const val PASSKEY_PROTOCOL_HEADER = "x-ms-PassKeyAuth"
```
shahzaibj:
```suggestion
        const val PASSKEY_PROTOCOL_HEADER_NAME = "x-ms-PassKeyAuth"
```

shahzaibj: I guess the way you have it is fine as well. But I generally would xx_NAME and xx_VALUE to denote consts for each of those

common4j/src/main/com/microsoft/identity/common/java/constants/FidoConstants.kt
```diff
@@ -36,5 +36,33 @@ class FidoConstants {
...
+
+        /**
+         * Version of the passkey protocol that we want to use.
+         */
+        const val PASSKEY_PROTOCOL_VERSION = "1.0"
+
+        /**
+         * Key types that we support with the passkey protocol.
+         * Possible values: securitykey, passkey, ngc
```
shahzaibj: Are these values defined as consts somewhere?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftTokenRequest.java
```diff
@@ -106,6 +106,12 @@ public MicrosoftTokenRequest() {
     @Accessors(prefix = "m")
     private boolean mPKeyAuthHeaderAllowed;
 
+    // Send Passkey Auth header for interactive calls with WebView as authorization agent.
+    @Getter
+    @Setter
+    @Accessors(prefix = "m")
+    private boolean mPasskeyAuthHeaderAllowed;
```
shahzaibj: You want this being sent to eSTS in https request body? Probably not. If not then you should mark this as `transient`

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
```diff
@@ -217,17 +218,23 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         headers.put(HttpConstants.HeaderField.CONTENT_TYPE, TOKEN_REQUEST_CONTENT_TYPE);
 
         if (request instanceof MicrosoftTokenRequest) {
+            MicrosoftTokenRequest microsoftTokenRequest = (MicrosoftTokenRequest) request;
```
shahzaibj:
```suggestion
            final MicrosoftTokenRequest microsoftTokenRequest = (MicrosoftTokenRequest) request;
```

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
shahzaibj: I'd think we shouldn't need this for /token endpoint since we should be using the RT for auth.

melissaahn: Since both of you mentioned this, I'm removing this logic: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2181/commits/8a90970d55796d97a386628c3bd025d82e46f23f

common4j/src/main/com/microsoft/identity/common/java/constants/FidoConstants.kt
melissaahn: Done: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2181/commits/8a90970d55796d97a386628c3bd025d82e46f23f

common4j/src/main/com/microsoft/identity/common/java/constants/FidoConstants.kt
melissaahn: Added them to the same file: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2181/commits/8a90970d55796d97a386628c3bd025d82e46f23f

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftTokenRequest.java
melissaahn: Removed these lines as a result of one of the previous comments above.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
melissaahn: Done

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -312,6 +312,12 @@ protected final AuthorizationRequest.Builder initializeAuthorizationRequestBuild
             );
             completeRequestHeaders.put(PKEYAUTH_HEADER, PKEYAUTH_VERSION);
 
+            // Not going to add passkey protocol header until full feature is ready.
+            if (FidoConstants.IS_PASSKEY_SUPPORT_READY
+                    && interactiveTokenCommandParameters.getAuthorizationAgent() == AuthorizationAgent.WEBVIEW) {
+                completeRequestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_VALUE);
+            }
```
mohitc1, line 332: we need to make similar change in PrtV3AuthorizationStrategy as well, as it creates auth request for acquire PRTv3 (sorry to point this late.)

I see potential to refactor this code out to OAuthStrategy and it's derived classes and then won't have that redundancy. We can do that as separate change.

melissaahn: See this PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2540

iamgusain (approved):

mohitc1 (approved):

melissaahn:
Added SkipConsumers check label due to some errors occurring in the Linux Broker assemble consumers task which are unrelated to this PR. All the other assemble consumer tasks passed,

PR #2183, opened 2023-09-20
AbstractFidoChallengeHandler and telemetry names

melissaahn:
### Summary
This PR adds the classes of AbstractFidoChallengeHandler, ~~ITelemetryHelper, IFidoTelemetryHelper, and FidoTelemetryHelper~~ and span and attribute names for future telemetry calls.
The eventual PasskeyFidoChallengeHandler, which will take care of getting the FIDO assertion and sending it to the server, will extend AbstractFidoChallengeHandler.
~~ITelemetryHelper is an interface which helps consolidate the setStatus methods of the CBA and FIDO telemetry helpers. IFidoTelemetryHelper has methods to set the Challenge and ChallengeHandler names on the future Fido span. Noting these object names will help determine the type of FIDO request (authentication vs registration) and the type of passkey used (regular passkey on device, or security key). The main implementation of IFidoTelemetryHelper is FidoTelemetryHelper, and a future implementation  will be done for unit testing purposes.~~
None of these classes are currently being used in logic, but reviewing them now helps cut down the future PR which will contain the PasskeyFidoChallengeHandler implementation.

### Related PRs
- https://github.com/AzureAD/ad-accounts-for-android/pull/2547


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/IFidoTelemetryHelper.java
```diff
@@ -0,0 +1,39 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.opentelemetry;
+
+import lombok.NonNull;
+
+public interface IFidoTelemetryHelper extends ITelemetryHelper {
```
shahzaibj: javadoc

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/IFidoTelemetryHelper.java
```diff
@@ -0,0 +1,39 @@
...
+     * Sets attribute indicating the type of FIDO challenge received from the server.
+     * @param challengeName name of the FidoChallenge class.
+     */
+    void setFidoChallenge(@NonNull final String challengeName);
+
+    /**
+     * Sets attribute indicating the type of FIDO challenge handler handling the current FIDO operation.
+     * @param challengeHandlerName name of the FidoChallengeHandler class.
+     */
+    void setFidoChallengeHandler(@NonNull final String challengeHandlerName);
```
shahzaibj: Usually interfaces are not used for declaring setter methods. Is there a specific reason these setters cannot be declared directly on the class? Do you ever imagine having different implementations that would need to set these values _differently_?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/ITelemetryHelper.java
```diff
@@ -0,0 +1,51 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.opentelemetry;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import lombok.NonNull;
+
+public interface ITelemetryHelper {
```
shahzaibj: javadoc

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/ITelemetryHelper.java
```diff
@@ -0,0 +1,51 @@
...
+    @SuppressFBWarnings
+    void setResultFailure(@NonNull final String message);
+
+    /**
+     * Indicates on the Span that the operation failed and then ends current Span.
+     * @param exception exception thrown upon error.
+     */
+    //Suppressing warnings for RETURN_VALUE_IGNORED_NO_SIDE_EFFECT
+    @SuppressFBWarnings
+    void setResultFailure(@NonNull final Exception exception);
```
shahzaibj: All of these kinda sound like callback to me than setters. Should the class be called `ICbaCompletionCallback` and the methods be renamed like `onSuccess` and `onFailure`?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/FidoTelemetryHelper.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Main implementation of IFidoTelemetryHelper.
+ */
+public class FidoTelemetryHelper implements IFidoTelemetryHelper {
+
+    private final Span mSpan;
+
+    public FidoTelemetryHelper() {
+        mSpan = OTelUtility.createSpan(SpanName.CertBasedAuth.name());
```
shahzaibj: Is this the right name for this span? I believe this span name is being used already in `CertBasedAuthTelemetryHelper`. Do you want the same span name in Fido as well?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/FidoTelemetryHelper.java
```diff
@@ -0,0 +1,78 @@
...
+package com.microsoft.identity.common.java.opentelemetry;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+
+/**
+ * Main implementation of IFidoTelemetryHelper.
+ */
+public class FidoTelemetryHelper implements IFidoTelemetryHelper {
```
mohitc1: Instead of a helper class, why not directly populate span (current span or new span per your case) in the at the place you would call these methods. Common attributes can be emitted in parent abstract class?

melissaahn: I could do that. This feature relies on a lot of abstraction since it pulls in the credman api and potentially others later, so I got into the habit of creating helpers when using other apis. But given my current plan is to only emit telemetry in one class (the ChallengeHandler class), there isn't a need here for a helper.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/FidoTelemetryHelper.java
melissaahn: No I do not :^) It will be `Fido`. Though I am most likely going to remove this class from this PR to directly create the span in a future challenge handler class.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/IFidoTelemetryHelper.java
melissaahn: For the Fido/Passkey feature, not really (see response to Mohit's comment). For CBA, a helper was needed because there were multiple classes/times where attributes were being set. Though maybe it didn't have to be an interface. That's something I can address in a separate PR for CBA, but for this one, based on the comments received, I'm planning on removing these helper and directly call methods on the Otel span objects.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/ITelemetryHelper.java
melissaahn: CBA has sets of callbacks for different scenarios, so I'll need to put more thought into the naming in a separate PR.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/FidoTelemetryHelper.java
melissaahn: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2183/commits/7aedde14b813a93bf9a198865da859b68af12806

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/IFidoTelemetryHelper.java
melissaahn: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2183/commits/7aedde14b813a93bf9a198865da859b68af12806

mohitc1 (approved):

shahzaibj (approved):

melissaahn:
Added SkipConsumers check label due to some errors occurring in the Linux Broker assemble consumers task which are unrelated to this PR. All the other assemble consumer tasks passed.

PR #2186, opened 2023-09-21
[E2E] Handle browser first run in MsalTestApp.acquireToken

cikelengfeng:
In certain E2E test cases, such as 2517374, a browser is used for login. However, MsalTestApp's acquireToken method doesn't handle the browser's first run, this can cause failures of test cases. 
Add browser and necessary parameters to acquireToken so that MsalTestApp can handle browser's first run as applicable.


somalaya (approved):

fadidurah (approved):
LGTM, please add a description to the PR, though

PR #2189, opened 2023-09-22
Adding Moshi to common; rewriting WebAuthnJsonUtil classes to Java

melissaahn:
### Summary
This PR:
- removes the kotlinx.serialization dependencies and plugins
- replaces it with Moshi
- rewrites the ~~WebAuthnJsonUtil and other~~ related data classes to Java, so we don't have to take in additional dependencies for Kotlin.
- updates WebAuthnJsonUtilTest slightly to match with how Moshi writes to Json.

### Related PRs
- complete: https://github.com/AzureAD/android-complete/pull/256
- broker: https://github.com/AzureAD/ad-accounts-for-android/pull/2552
- msal: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1909
- adal: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1761


mohitc1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
```diff
@@ -20,14 +20,23 @@
...
+
+import androidx.annotation.NonNull;
+
+import lombok.Getter;
 
 /**
  * Representation of WebAuthn's PublicKeyCredentialDescriptor.
  * https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor
  */
-@kotlinx.serialization.Serializable
```
rpdome, line 29: Do we need to convert all classes into java? would the non-data kotlin class work?

asking because their repo says that they support both. https://github.com/square/moshi

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
```diff
@@ -20,14 +20,23 @@
...
-data class PublicKeyCredentialDescriptor(
-    val type: String,
-    val id: String
-)
+@Getter
+public class PublicKeyCredentialDescriptor {
+    public final String type;
+    public final String id;
```
shahzaibj: Would be a good practice to put Serialized Name annotation on these

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialRequestOptions.java
```diff
@@ -20,16 +20,32 @@
...
-    val rpId: String,
-    val allowCredentials: List<PublicKeyCredentialDescriptor>,
-    val userVerification: String
-)
+@Getter
+public class PublicKeyCredentialRequestOptions {
+    public final String challenge;
+    public final String rpId;
+    public final List<PublicKeyCredentialDescriptor> allowCredentials;
+    public final String userVerification;
```
shahzaibj: Would be a good practice to put Serialized Name annotation on these

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
melissaahn: Non-data Kotlin classes should still work. The unit tests are written in Kotlin and can run, and I did test calling WebAuthnJsonUtil from some other Kotlin classes (like the PasskeyFidoChallengeHandler, which will come out in a future PR) and it did what it needed to do. The only issue I was thinking about was the challenge object being a kotlin class, but it doesn't throw any errors while referenced in WebAuthnJsonUtil.

Moshi supports both, but for Kotlin, an extra dependency needs to be added: https://github.com/square/moshi#kotlin
Reflection adds a lot in size, and while the codegen option is smaller, from my googling, it slows down build speed.
I was thinking that if I just wrote this part of the feature (WebAuthnJsonUtil and associated data classes) in Java, it could help minimize size increase (only one dependency, no plugin, and it's already in broker) with no known potential impact to build time.

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
melissaahn: Added the equivalent Moshi Json annotation: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2189/commits/2e2c99898db30576eca9f758a4fbf12395573103

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialRequestOptions.java
melissaahn: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2189/commits/2e2c99898db30576eca9f758a4fbf12395573103

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
rpdome: i see. surprising that even WebAuthnJsonUtil itself needs to be in java.

got it. thanks for the info.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/PublicKeyCredentialDescriptor.java
melissaahn: Actually you're right... WebAuthnJsonUtil does not need to be written in Java, as the extra dependencies are specifically for converting Kotlin classes to and from JSON. I think I misunderstood your initial comment. 
Addressed here: 3723fbe9dc03bac93ad9e3a785ec01697d73a1b4

rpdome (approved):

mohitc1 (approved):

mohitc1 (approved):

PR #2190, opened 2023-09-22
handle crypto error gracefully

rpdome:
We've discovered that Android KeyStore could throw an unexpected error due to a bug in their stack.

This is to make sure the error is handled gracefully without crashing the app.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
@@ -33,6 +33,7 @@ import static com.microsoft.identity.common.java.exception.ClientException.INVAL
 import static com.microsoft.identity.common.java.exception.ClientException.NO_SUCH_ALGORITHM;
 import static com.microsoft.identity.common.java.exception.ClientException.NO_SUCH_PADDING;
 import static com.microsoft.identity.common.java.exception.ClientException.UNEXPECTED_HMAC_LENGTH;
+import static com.microsoft.identity.common.java.exception.ClientException.UNKNOWN_ERROR;
 
 import com.microsoft.identity.common.java.crypto.key.AbstractSecretKeyLoader;
 import com.microsoft.identity.common.java.crypto.key.KeyUtil;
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
```diff
@@ -175,6 +181,9 @@ public byte[] encrypt(final byte[] plaintext)
         } catch (final InvalidAlgorithmParameterException e) {
             errCode = INVALID_ALG_PARAMETER;
             exception = e;
+        } catch (final Throwable e) {
+            errCode = UNKNOWN_ERROR;
```
shahzaibj: Can we find a more meaningful error code?

shahzaibj (approved):
My only comment is about the error code.

common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
rpdome: This one is unknown though... as in, we don't know what Android throw out.

shahzaibj: But we at least know this is failing in encryption layer right? Can we create a new error code? Something like `ENCRYPTION_FAILED`?

rpdome: ENCRYPTION_FAILED might not be accurate in some cases- since the exception we got is in the key generation.

I guess i could create UNKNOWN_CRYPTO_ERROR?

shahzaibj: The method in which this is thrown seems to be called `encrypt` so I suspected in the encryption layer.

Anyway, any thing more specific than UNKNOWN_ERROR would be good so if you want `UNKNOWN_CRYPTO_ERROR` then that's fine.

shahzaibj (approved):

fadidurah (approved):

PR #2192, opened 2023-09-25
Add AzureEnvironment to ILabAccount

rpdome:
Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2553


LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/LabConstants.java
```diff
@@ -94,7 +94,7 @@ static final class FederationProvider {
         public static final String CIAM = "ciam";
     }
 
-    static final class AzureEnvironment {
+    public static final class AzureEnvironment {
```
shahzaibj, line 97: Was this change actually needed?

shahzaibj (approved):

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/constants/LabConstants.java
rpdome: Yes. In the other PR, I used the AzurePPE constant to check (against AzureEnvironment of the account) if we should be making a drs discovery request to prod or ppe.

p3dr0rv (approved):

PR #1310, opened 2021-04-13
Refresh_in init (eager)

aherciya:
These are the initial changes needed for enabling refreshes on silent flows for MSAL. 
Added a callback to be triggered once a refresh need is realized.

PR #2195, opened 2023-09-27
Add a hub app's redirectURI to an Allowlist of valid redirectURIs

somalaya:
- Added Teams app's redirectURL that is registered in US Gov cloud. Related broker PR that is using this : https://github.com/AzureAD/ad-accounts-for-android/pull/2555
- Also, in one of the renewAccessToken methods (that is not currently being used when PRTV3 is enabled), we did not make changes required for NAA (skipping caching the tokens in NAA flow). Although this is not required now, just adding the changes to maintain consistency.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
@@ -215,7 +215,7 @@ public class AndroidPlatformUtil implements IPlatformUtil {
         // Below redirectURI is being used in our automation tests and also by OneAuth tests for NAA
         return BuildConfig.DEBUG && (redirectUri.equals("msauth://com.microsoft.teams/VCpKgbYCXucoq1mZ4BZPsh5taNE=")
                         || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno=")
-                        || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno%3D"));
+                        || redirectUri.equals("https://login.microsoftonline.com/common/oauth2/nativeclient"));
     }
 
     /**
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -448,9 +448,9 @@ public abstract class BaseController {
                     methodTag,
                     "Token request was successful"
             );
-            List<ICacheRecord> acquireTokenResultRecords = null;
+            List<ICacheRecord> acquireTokenResultRecords;
             if (parameters.hasNestedAppParameters()) {
-                // Do not save the token in naa flow. This is because, NAA is curerntly only supported with OneAuth and OneAuth already caches AT until it is expired.
+                // Do not save the token in naa flow. This is because, NAA is currently only supported with OneAuth and OneAuth already caches AT until it is expired.
                 // Design doc : https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/7876
                 acquireTokenResultRecords = getAcquireTokenResultRecords(
                         (MicrosoftStsTokenResponse) tokenResult.getTokenResponse(),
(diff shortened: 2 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -564,20 +564,25 @@ public TokenResult renewAccessToken(@NonNull final SilentTokenCommandParameters
...
             );
 
-            // Remove old Access Token
-            Logger.info(
-                    methodTag,
-                    "Access token is refresh-expired. Removing from cache..."
-            );
-            final AccessTokenRecord accessTokenRecord = cacheRecord.getAccessToken();
-            cache.removeCredential(accessTokenRecord);
+            List<ICacheRecord> savedRecords;
```
rpdome, line 567: I'm not sure if this code path is even exercised at all.

This one seems to be invoked by the code path that relies on LibraryConfiguration.refreshInEnabled, which seems to be set to OFF by default.

Probably worth pursuing if this one is in use. Shane or Olga might be able to give you the answer. @shahzaibj do you remember? My memory around this feature is very very faint.

If not (and it's never going to be used) - maybe removing the code might be a better way to go. (so that we don't have to maintain / add test coverage for it... which I think we currently don't have one for this code path)

rpdome (commented):
left some comment. otherwise lgtm.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -215,7 +215,7 @@ private boolean isValidHubRedirectURIForNAATests(String redirectUri) {
         // Below redirectURI is being used in our automation tests and also by OneAuth tests for NAA
         return BuildConfig.DEBUG && (redirectUri.equals("msauth://com.microsoft.teams/VCpKgbYCXucoq1mZ4BZPsh5taNE=")
                         || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno=")
-                        || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno%3D"));
```
fadidurah, line 218: I see the description says you are adding a new redirect uri. Is this old one being deleted no longer in use?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
shahzaibj: This one seems the same as the one on the previous line

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -215,7 +215,7 @@ private boolean isValidHubRedirectURIForNAATests(String redirectUri) {
...
+                        || redirectUri.equals("https://login.microsoftonline.com/common/oauth2/nativeclient"));
```
shahzaibj, line 218: Generally apps use the same redirect URI across clouds. Is there a reason Teams has adopted a different one for USGov? Also strange that this has `.com` in it and not `.us`.

Also good but interesting that this is an app link.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
shahzaibj: I remember reviewing those PRs long time back along with Shane. The memory is faint but the parts I remember is that the feature was sort of opt-in and like you said it is OFF by default. Possible that there are customers who have enabled it and are getting benefits already. 

This was the PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1310

The feature is focused on resiliency so I think we certainly need it. Interestingly I had someone reach out today asking for something similar to this feature so I definitely think there is partner and customer desire for it as well. Let's keep the code as we likely need now and in the future to continue to build resiliency in our products.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
somalaya: yes, this one was duplicated. I have replaced it with USGov redirectURI

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
somalaya: You are right, they do have same redirectURI for USGov and WW clouds. But, we could not use the common ones in our testing with our lab accounts. It says the redirectURI is not recognized in msaidlab4 tenant. So, I believe Nick has resolved this issue by getting the other redirectURI I added above. I am not aware of what was done to make this usable. But, this is mostly used by us and OneAuth team for testing only NAA scenarios.  So, I did not check more on this.

shahzaibj (approved):

PR #2202, opened 2023-10-03
Fix crash during onCreate()

sansrivastav:
AB#2690156

LTW reported following crashes and this PR addresses the first 2 crashes:

Stack Trace
1.
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.os.Parcelable android.os.Bundle.getParcelable(java.lang.String)' on a null object reference
at com.microsoft.identity.common.internal.broker.InstallCertActivityLauncher.onCreate(InstallCertActivityLauncher.java:16)

2.
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.os.Parcelable android.os.Bundle.getParcelable(java.lang.String)' on a null object reference
at com.microsoft.identity.common.internal.broker.BrokerActivity.onCreate(BrokerActivity.java:16)

3.
Caused by: java.lang.IllegalStateException: Unexpected fragment type at com.microsoft.identity.common.internal.providers.oauth2.CurrentTaskAuthorizationActivity.onCreate(CurrentTaskAuthorizationActivity.java:142)

See this bug for additional details: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2690156

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
@@ -62,7 +62,10 @@ public final class BrokerActivity extends Activity {
         super.onCreate(savedInstanceState);
 
         if (savedInstanceState == null) {
-            mBrokerInteractiveRequestIntent = getIntent().getExtras().getParcelable(BROKER_INTENT);
+            Bundle extras = getIntent().getExtras();
+            if(extras != null) {
+                mBrokerInteractiveRequestIntent = extras.getParcelable(BROKER_INTENT);
+            }
         } else {
             mBrokerInteractiveRequestIntent = savedInstanceState.getParcelable(BROKER_INTENT);
             mBrokerIntentStarted = savedInstanceState.getBoolean(BROKER_INTENT_STARTED);
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
@@ -60,11 +60,14 @@ public final class BrokerActivity extends Activity {
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        final String methodTag = TAG + ":onCreate";
 
         if (savedInstanceState == null) {
             Bundle extras = getIntent().getExtras();
             if(extras != null) {
                 mBrokerInteractiveRequestIntent = extras.getParcelable(BROKER_INTENT);
+            } else {
+                Logger.warn(methodTag, "Extras is null.");
             }
         } else {
             mBrokerInteractiveRequestIntent = savedInstanceState.getParcelable(BROKER_INTENT);
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/InstallCertActivityLauncher.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/InstallCertActivityLauncher.java
@@ -82,10 +82,13 @@ public final class InstallCertActivityLauncher extends AppCompatActivity {
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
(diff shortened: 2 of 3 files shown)


shahzaibj:
> AB#2690156

Please add a proper description. PR descriptions should include a summary of What, Why, and How i.e. what the PR wants to accomplish, why the change is needed, and small summary of the changes made to accomplish the task.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
```diff
@@ -62,7 +62,10 @@ protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         if (savedInstanceState == null) {
-            mBrokerInteractiveRequestIntent = getIntent().getExtras().getParcelable(BROKER_INTENT);
+            Bundle extras = getIntent().getExtras();
+            if(extras != null) {
+                mBrokerInteractiveRequestIntent = extras.getParcelable(BROKER_INTENT);
```
iamgusain, line 68: Do we understand in what scenarios the BROKER_INTENT is not being passed?
Should we return an error or ignore silently?

common/src/main/java/com/microsoft/identity/common/internal/broker/InstallCertActivityLauncher.java
```diff
@@ -83,7 +83,10 @@ public final class InstallCertActivityLauncher extends AppCompatActivity {
     protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         if (savedInstanceState == null) {
-            mInstallCertificateIntent = getIntent().getExtras().getParcelable(INSTALL_CERT_INTENT);
+            Bundle extras = getIntent().getExtras();
+            if(extras != null) {
+                mInstallCertificateIntent = extras.getParcelable(INSTALL_CERT_INTENT);
```
iamgusain, line 89: Same question here, in what scenarios InstallCertActivity is being launched without the INSTALL_CERT_INTENT, and should we throw an error back?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
sansrivastav: There is not much information available except for the crash stack. The partner also has closed the investigation due to low hits.
Even though it is good to understand the reason, we should prevent crashes by adding this check for sure.
I do not think we can return error here, but logging may be helpful.

common/src/main/java/com/microsoft/identity/common/internal/broker/InstallCertActivityLauncher.java
sansrivastav: There is not much information available except for the crash stack. The partner also has closed the investigation due to low hits.
Even though it is good to understand the reason, we should prevent crashes by adding this check for sure.
I do not think we can return error here, but logging may be helpful.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerActivity.java
shahzaibj: If we don't return error here then what happens? What is the end user experience in this case?

common/src/main/java/com/microsoft/identity/common/internal/broker/InstallCertActivityLauncher.java
shahzaibj: If we don't return error here then what happens? What is the end user experience in this case?

shahzaibj (approved):

iamgusain (approved):

sansrivastav:
Only Linux broker tests on consumers pipeline is failing 

image

tanmaymanolkar1:
have mentioned @sansrivastav to add the skip consumer check label, since this is a known issue and the only tests which are failing are for Linux.
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1191967&view=results
image

PR #2206, opened 2023-10-10
[E2E Test] Fix OneAuthTestApp.handleFirstRun

cikelengfeng:
The current implementation logic of OneAuthTestApp.handleFirstRun is: switch the config first, then turn on "Prefer AAD Broker" flag. However, OneAuthTestApp decides whether to truly enable the Broker when switching the config. Thus, the current order doesn't activate the Broker as expected. 

The solution is to turn on "Prefer AAD Broker" before switching the config.


somalaya:
Do you mean after switching the toggle "Prefer AAD Broker", OneAuthTestApp does not route the requests through broker in the current order?

cikelengfeng:
> Do you mean after switching the toggle "Prefer AAD Broker", OneAuthTestApp does not route the requests through broker in the current order?

Yes, in my testing, switching app configuration after turning on "Prefer AAD Broker" can truly route requests to broker.

Switching app configuration is the key action which actually take effect according to the "Prefer AAD Broker" flag, so it should be done as last step, I guess there must be some bugs in OneAuthTestApp, but I don't have time to look into their code.

somalaya (approved):

somalaya:
> > Do you mean after switching the toggle "Prefer AAD Broker", OneAuthTestApp does not route the requests through broker in the current order?
> 
> Yes, in my testing, switching app configuration after turning on "Prefer AAD Broker" can truly route requests to broker.
> 
> Switching app configuration is the key action which actually take effect according to the "Prefer AAD Broker" flag, so it should be done as last step, I guess there must be some bugs in OneAuthTestApp, but I don't have time to look into their code.

Sure, we can go with this solution for now. Adding @matthewyu-ms to see if OneAuth team needs to correct this.

fadidurah (approved):

PR #2209, opened 2023-10-16
Passing span context to AuthorizationActivity

melissaahn:
### Summary
Currently, CBA spans are not correlating correctly with their parent span (supposed to be AcquireTokenInteractive). A new trace_id is being created for CBA spans upon instantiation, when we want the trace_id from the AcquireTokenInteractive span to be populated there instead. 
This is because the span context doesn't get inherited when we start the AuthorizationActivity, the main activity where the CBA code (and much logic running behind the embedded WebView) resides. The workaround for this is to put a SerializableSpanContext instance as an extra into the intent that starts AuthorizationActivity. The context is then retrieved and saved as a class variable within the Activity. The CBA classes can reference the span context as a parent span when it creates its own span (which means they'll share the same trace_id).


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
```diff
@@ -28,19 +28,27 @@
...
+    @Accessors(prefix = "m")
+    private SpanContext mSpanContext;
     private AuthorizationFragment mFragment;
 
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         final String methodTag = TAG + ":onCreate";
+        mSpanContext = (SpanContext) getIntent().getExtras().getSerializable(SerializableSpanContext.SERIALIZABLE_SPAN_CONTEXT);
```
p3dr0rv: NPE if getIntent().getExtras() is null? add a check to be on the safe side

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/CertBasedAuthFactory.java
```diff
@@ -90,7 +91,12 @@ protected CertBasedAuthFactory(@NonNull final Activity activity,
      * @param callback logic to run after an AbstractCertBasedAuthChallengeHandler is chosen.
      */
     public void createCertBasedAuthChallengeHandler(@NonNull final CertBasedAuthChallengeHandlerCallback callback) {
-        final ICertBasedAuthTelemetryHelper telemetryHelper = new CertBasedAuthTelemetryHelper();
+        final ICertBasedAuthTelemetryHelper telemetryHelper;
+        if (mActivity instanceof AuthorizationActivity) {
+           telemetryHelper = new CertBasedAuthTelemetryHelper(((AuthorizationActivity) mActivity).getSpanContext());
+        } else {
+            telemetryHelper = new CertBasedAuthTelemetryHelper(null);
```
p3dr0rv, line 98: why null, why no empty constructor?

p3dr0rv (approved):
left a comment, but otherwise LGTM

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/CertBasedAuthFactory.java
```diff
@@ -90,7 +91,12 @@ protected CertBasedAuthFactory(@NonNull final Activity activity,
      * @param callback logic to run after an AbstractCertBasedAuthChallengeHandler is chosen.
      */
     public void createCertBasedAuthChallengeHandler(@NonNull final CertBasedAuthChallengeHandlerCallback callback) {
-        final ICertBasedAuthTelemetryHelper telemetryHelper = new CertBasedAuthTelemetryHelper();
+        final ICertBasedAuthTelemetryHelper telemetryHelper;
+        if (mActivity instanceof AuthorizationActivity) {
```
shahzaibj: What is the other possible type for activity here? and in what cases do we hit that route?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
```diff
@@ -35,8 +38,16 @@ public class CertBasedAuthTelemetryHelper implements ICertBasedAuthTelemetryHelp
...
+    /**
+     * Creates an instance of CertBasedAuthTelemetryHelper.
+     * @param spanContext current span context.
+     */
+    public CertBasedAuthTelemetryHelper(@Nullable final SpanContext spanContext) {
+        if (spanContext != null) {
+            mSpan = OTelUtility.createSpanFromParent(SpanName.CertBasedAuth.name(), spanContext);
+        } else {
+            mSpan = OTelUtility.createSpan(SpanName.CertBasedAuth.name());
+        }
```
shahzaibj: Alternatively, you can provide two separate constructors: one that takes a NonNull SpanContext and the other that doesn't even take an argument

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -96,6 +98,12 @@ public static Intent getAuthorizationActivityIntent(final Context context,
         intent.putExtra(WEB_VIEW_ZOOM_CONTROLS_ENABLED, webViewZoomControlsEnabled);
         intent.putExtra(WEB_VIEW_ZOOM_ENABLED, webViewZoomEnabled);
         intent.putExtra(CORRELATION_ID, DiagnosticContext.INSTANCE.getRequestContext().get(DiagnosticContext.CORRELATION_ID));
+        intent.putExtra(SerializableSpanContext.SERIALIZABLE_SPAN_CONTEXT, SerializableSpanContext.builder()
```
shahzaibj: I'd suggest that you use JSON based serialization here...at least that's what the class was built with in mind.

Java serialization / de-serialization can be unsafe

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/CertBasedAuthFactory.java
melissaahn: This class gets the activity instance from AzureActiveDirectoryWebViewClient, which takes in any Activity as a parameter. This makes it possible for an activity other than AuthorizationActivity to be passed here, so in the (albeit unlikely) event that this does happen in the future, I'd rather handle this silently than assume and risk a crash. I will put a comment here to explain.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivity.java
melissaahn: Addressed here: 51e4f91049ddfd1979ec06cc9c4292590937d561

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/CertBasedAuthTelemetryHelper.java
melissaahn: Addressed here: 51e4f91049ddfd1979ec06cc9c4292590937d561

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
melissaahn: I created another class to help convert to and from JSON. 51e4f91049ddfd1979ec06cc9c4292590937d561 What do you think?

shahzaibj: Is there a reason we need a JsonAdapter class specifically for SpanContext?

Can't we write a generic JsonAdapter class that can adapt convert to/from JSON for any Java class? IIRC, I wrote something similar in Linux called as `MoshiJsonAdapter.java`

melissaahn: Created CommonMoshiJsonAdapter here: ae60a8e875510a2fb14275ef96f637bfcf3c6083 
I copied the toJson and fromJson methods from your MoshiJsonAdapter.java class. I do think that IJsonAdapter.java should probably be moved to common now, but moving it seemed to affect many LinuxBroker classes, so that's something I can save for a separate PR.

common/src/main/java/com/microsoft/identity/common/internal/util/CommonMoshiJsonAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param <T>      the type of the JAVA object that will be returned after deserialization
+     * @return a JAVA object
+     */
+    public <T> T fromJson(@NonNull final String json, @NonNull final Class<T> classOfT)
+            throws TerminalException {
+        final JsonAdapter<T> jsonAdapter = mMoshi.adapter(classOfT);
+        try {
+            return jsonAdapter.fromJson(json);
+        } catch (final IOException e) {
+            throw new TerminalException(e.getMessage(), e, ErrorStrings.IO_ERROR);
```
shahzaibj: Can we use more specific error code? Something that signals that deserialization failed? (I don't want this to be another thing that falls into the `io_error` bucket - that makes it very hard to draw conclusions from telemetry)

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/util/CommonMoshiJsonAdapter.java
melissaahn: Created json_deserialization_failure error string: 67dd754a61e537ed13468e183f1c5d0843da498a

PR #2211, opened 2023-10-18
Fix for ADO 2696517

shahzaibj:
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2568

This makes the following change:

- Always validate calling app's redirect URI (as opposed to only doing it for MSAL and OneAuth).
- If we don't validate redirect URI then that will break MAM as it will allows callers to impersonate app identity


somalaya (approved):

fadidurah (approved):

mohitc1 (approved):
:shipit:

PR #2212, opened 2023-10-18
add removeUserLegacy method in MsalTestApp infra

amoyixi:
This PR is 
1. adding removeUserLegacy method in MsalTestApp infra in order to get text after clicking removeUser button on legacy MsalTestApp
2. adding some comments for each method


somalaya (approved):

fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/MsalTestApp.java
```diff
@@ -172,6 +186,7 @@ public String getActiveBrokerPackageName() throws UiObjectNotFoundException {
         return text;
     }
 
+    // check MsalTestApp mode
     public String checkMode() throws UiObjectNotFoundException {
```
somalaya, line 190: I noticed that some of the tests in which checkMode() is getting called are failing. This is because, it takes some time for the screen to load for checkMode to be reflected. Could you please add a sleep (3secs should be enough) everywhere this method is being called?

amoyixi: this method is only used in TestCase2582292. I added wait time after hanleFirstRun here:
image

PR #2219, opened 2023-10-27
isQrPinAvailable API

p3dr0rv:
## Why?
To determine if QR + PIN authorization is available for MDM devices, Teams app will call this API.
if the result is true: Teams will show a button to login using QR + PIN, 
if the result is false: Teams will not show this button.

## How does this API works?
The QR + PIN feature will be configured by the administrator through app configuration policies for managed Android Enterprise devices. https://learn.microsoft.com/en-us/mem/intune/apps/app-configuration-policies-use-android
Where the administrator can enable/disable this type of authorization for all managed devices, by seating this key "isQrPinAvailable" to true in the Authenticator app. 

On Android we have 3 apps which can be the broker, Authenticator, CP (Company Portal) and LTW (Link to Windows). which makes things extremely complex.  
To keep the configuration simple for the tenant's administrator, we propose to use the managed configurations of the Authenticator app to read/set the admin flag. Meaning that authenticator app must be installed on Android devices to use the QR + PIN feature. 

qrpinapi drawio2

[Client design doc](https://microsoft-my.sharepoint-df.com/:w:/p/mipetriu/EWbcvu99HRxEgy-BC9jilKoBnqIr3yCJaNPZMlGSG5c3ow?e=aP0ey1)

## Changes

- Add constants for API
- Add BaseCommand for isQrPinAvailable, used by BrokerMsal API
- Add behavior for LocalMsalController (not supported) BrokerMsalController  and (supported, call the broker)

## Related PR
https://github.com/AzureAD/ad-accounts-for-android/pull/2587
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1931


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1679,6 +1687,16 @@ public enum API {
          */
         public static final String GET_SSO_TOKEN_PATH = "/ssoToken";
 
+        /**
+         * Broker api path to check if QR + PIN should de available.
+         */
+        public static final String IS_QR_PIN_AVAILABLE_PATH = "/isQrPinAvailable";
+
+        /**
+         * Broker api path to check if QR + PIN should de available.
```
rpdome: nit: please make it clear on the differences between these two operations.

common/src/main/java/com/microsoft/identity/common/internal/commands/IsQrPinAvailableCommand.kt
```diff
@@ -0,0 +1,69 @@
...
+    publicApiId: String
+) : BaseCommand<Boolean?>(parameters, controllers, callback, publicApiId) {
+
+    companion object {
+        private val TAG = IsQrPinAvailableCommand::class.java.simpleName
+    }
+
+    override fun execute(): Boolean {
+        val methodTag = "$TAG:execute"
+        // Is important that we check the controllers in reverse order, as the first controller in the list is LocalMSALController,
```
rpdome: We actually don't need to execute this on LocalMSALController at all, right?

You might want to proceed this the same way as AcquireToken (if broker installed, use broker) , not AcquireTokenSilent (try both)

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -800,6 +803,74 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                    @Override
+                    public void performPrerequisites(final @NonNull IIpcStrategy strategy) throws BrokerCommunicationException {
+                        // AccountManagerAddAccountStrategy and BoundServiceStrategy are not supported for this operation.
+                        // We are failing fast here to avoid unnecessary IPC calls.
+                        // IPC calls through BoundServiceStrategy takes approximate 25s to timeout.
+                        if (strategy instanceof BoundServiceStrategy || strategy instanceof AccountManagerAddAccountStrategy) {
+                            final String errorMessage = strategy + " is not supported for isQrPinAvailable operation";
+                            Logger.warn(methodTag,  errorMessage);
+                            throw new BrokerCommunicationException(
+                                    BrokerCommunicationException.Category.OPERATION_NOT_SUPPORTED_ON_CLIENT_SIDE,
```
rpdome: OPERATION_NOT_SUPPORTED_ON_SERVER_SIDE?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -800,6 +803,74 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                                    strategy.getType(),
+                                    errorMessage,
+                                    new ClientException(ErrorStrings.BROKER_BIND_SERVICE_FAILED, errorMessage)
```
rpdome: would BROKER_BIND_SERVICE_FAILED be misleading? perhaps we can keep it null.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -800,6 +803,74 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                            );
+                        }
+                    }
+
+                    @Override
+                    public @NonNull
```
rpdome: nit:

```
@Override 
@NonNull
public RETURN_VALUE methodName(){
}1
```

here and everywhere.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -800,6 +803,74 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                                mActiveBrokerPackageName,
+                                null
+                        );
+                    }
+
+                    @Override
+                    public @NonNull
+                    Boolean extractResultBundle(final @Nullable Bundle resultBundle) throws BaseException {
+                        if (resultBundle == null) {
+                            throw mResultAdapter.getExceptionForEmptyResultBundle();
```
rpdome: nit: this null check probably go inside getIfQrPinIsAvailableFromResultBundle.

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -509,6 +509,20 @@ public boolean removeAccount(
...
+     * Checks if QR + PIN auth is available.
+     * LocalMSALController is not eligible to use the broker.
+     * Do not check if QR + PIN is available and return false immediately.
+     *
+     * @return true false
+     */
+    @Override
+    public boolean isQrPinAvailable() {
+        final String methodTag = TAG + ":isQrPinAvailable";
+        Logger.warn(methodTag, "LocalMSALController is not eligible to use the broker.");
```
rpdome: nit: probably add something along the line of "QR+Pin requires broker"

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -816,4 +817,18 @@ public GenerateShrResult getGenerateShrResultFromResultBundle(@NonNull final Bun
...
+     *
+     * @param bundle The {@link Bundle} to check.
+     * @return True if the broker is capable of QR code + PIN auth, false otherwise.
+     * @throws BaseException If the bundle does not contain the key.
+     */
+    public boolean getIfQrPinIsAvailableFromResultBundle(@NonNull final Bundle bundle) throws BaseException {
+        if (!bundle.containsKey(IS_QR_PIN_AVAILABLE)) {
+            throw new MsalBrokerResultAdapter().getBaseExceptionFromBundle(bundle);
+        }
+        return bundle.getBoolean(IS_QR_PIN_AVAILABLE);
```
rpdome, line 835: nit: default = false?

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1550,7 +1555,10 @@ public enum API {
             BROKER_DISCOVERY_METADATA_RETRIEVAL(RETRIEVE_BROKER_DISCOVERY_METADATA_PATH, null, null),
             BROKER_DISCOVERY_FROM_SDK(BROKER_DISCOVERY_FROM_SDK_PATH, null, null),
             BROKER_DISCOVERY_SET_ACTIVE_BROKER(BROKER_DISCOVERY_SET_ACTIVE_BROKER_PATH, null, null),
-            PASSTHROUGH(PASSTHROUGH_PATH, null, null),;
+            PASSTHROUGH(PASSTHROUGH_PATH, null, null),
+
```
shahzaibj: I guess remove this extra line

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1679,6 +1687,16 @@ public enum API {
          */
         public static final String GET_SSO_TOKEN_PATH = "/ssoToken";
 
+        /**
+         * Broker api path to check if QR + PIN should de available.
```
shahzaibj: Content Provider path?

Perhaps that should be reflected in the javadoc

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -144,6 +144,8 @@ public abstract boolean removeAccount(
     public abstract boolean getDeviceMode(final CommandParameters parameters)
             throws Exception;
 
+    public abstract boolean isQrPinAvailable() throws Exception;
```
shahzaibj, line 152: javadoc

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
shahzaibj: Rather than throwing exceptions, we can just not execute these strategies altogether? I recall some place we loop through them...so if we can continue without executing these then you don't even have to figure out what exception to throw

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
p3dr0rv: addeds

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
p3dr0rv: updated

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
p3dr0rv: there is no possibility for a default because we check for they key in the statement above

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
p3dr0rv: we could call directly the content provider strategy and ignore the rest, but I prefer to leave it this way, in case we adopt a new strategy on the future, also in this way we can have all the span data from the BrokerOperationExecutor,

SaurabhMSFT (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -800,6 +803,71 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+
+    /**
+     * Checks if QR + PIN authorization is available.
+     *
+     * @return true if if QR + PIN authorization is available. False otherwise.
+     */
+    @Override
+    public boolean isQrPinAvailable() throws BaseException {
+        final String methodTag = TAG + ":isQrPinAvailable";
+        return mBrokerOperationExecutor.execute(
```
iamgusain, line 815: Should we make this call to always go to Authenticator from the Msal side, instead of the active Broker?
and do the checks for Authenticator being installed on the Msal layer, instead of Broker layer?

Since this code is in common, OneAuthMsal should also work.

PR #2220, opened 2023-10-28
Add JWT header field for KDF version

iamgusain:
### What
Adding new jwt header field to enable sending Key Derivation Function (KDF) version

### Why
With kdf_ver=2, the ctx value used in the KDF is derived from JWT payload which includes a nonce.  This ensures that once a key is derived from a ctx value it can not be used indefinitely and is only valid till the nonce in JWT payload is valid.
More details in below spec.
https://msazure.visualstudio.com/DefaultCollection/One/_git/ESTS-Docs?path=/Protocols/Windows/DerivedKeyExport.md&_a=preview&anchor=sample-assertion

### How
Adding the header field in the JwtRequestHeader class

### Testing
Verified sending request with kdf_ver=2 locally 

### related 
Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2584


common4j/src/main/com/microsoft/identity/common/java/jwt/JwtRequestHeader.java
```diff
@@ -58,6 +58,9 @@ public final class JwtRequestHeader extends AbstractJwtRequest {
     @SerializedName(ClaimNames.KID)
     private String mKId;
 
+    @SerializedName("kdf_ver")
```
mohitc1: nit: Define as Constant?

mohitc1 (approved):

fadidurah (approved):

PR #2229, opened 2023-11-08
Signup for Native Auth

SaurabhMSFT:
This PR includes the work for Signup related features of Native Auth. No new dependencies have been added as part of this PR.

A technical overview of Native Auth is available at https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Native%20authentication/technical_overview.md&version=GBsammy/native-authentication&_a=preview

A companion PR in MSAL is https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1949

PR #2207, opened 2023-10-10
SignIn features for Native Auth

SaurabhMSFT:
This PR includes the work for SignIn relates features of Native Auth. 
Mockito-kotlin has been added as dependency for test projects in common and testutils project. Powermock has been added as a dependency to testutils project.

Companion PR in MSAL repo is https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1919

PR #2166, opened 2023-09-01
Base work for the CIAM features

SaurabhMSFT:
The code in this pull request is for the base work needed to support Signup, SignIn and SSPR features for Native Auth. Any class that is used by at least two of the features has been included in this pull request. 

This pull request introduces `mockk` and `mockito` libraries as dependencies for test projects. The `common` library test project has a new dependency on `testutils` library project.

A companion PR in the MSAL repository is https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1897

PR #2233, opened 2023-11-09
Native Auth SignIn, Signup and SSPR changes

SaurabhMSFT:
This PR is to merge the SignIn, Signup and other common code into dev branch. This code was previously reviewed in #2166, #2207 and #2229

A technical overview of Native Auth is available at https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Native%20authentication/technical_overview.md&version=GBsammy/native-authentication&_a=preview

A companion PR in common repo is https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1947

_________________________
### NativeAuth
Native Authentication helps to create authentication experiences that are natively integrated into developer's applications without delegating to browsers. This means that the login interface is hosted on the client application and the application look and feel is controlled in app code, using the same technologies used to create the app. This ensures developers have full control of the signup-signin experiences.

### Size Increase for common library.
This change brings additional ~400KB in Native Auth code to common library.


common4j/src/main/com/microsoft/identity/common/java/commands/parameters/nativeauth/AcquireTokenNoFixedScopesCommandParameters.java
```diff
@@ -0,0 +1,147 @@
...
+
+        Logger.verbose(
+                TAG + methodName,
+                "Validating operation params..."
+        );
+
+        // AuthenticationScheme is present...
+        if (null == authenticationScheme) {
+            throw new ArgumentException(
+                    ArgumentException.ACQUIRE_TOKEN_SILENT_OPERATION_NAME, // TODO update argument name
```
fadidurah: This TODO still relevant?

common/src/test/java/com/microsoft/identity/common/MicrosoftStsAccountCredentialAdapterTest.java
```diff
@@ -71,8 +71,6 @@
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.when;
 
-@PowerMockIgnore("javax.crypto.*")
-@RunWith(PowerMockRunner.class)
```
fadidurah, line 75: Did Native Auth add this change?

SaurabhMSFT: yes, NativeAuth modified it and it was done as part of SignIn PR. We had discussed this change in that PR too. https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2207#discussion_r1385249284

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/nativeauth/AcquireTokenNoFixedScopesCommandParameters.java
SammyO: @SaurabhMSFT yeah I think we should update it - the argument name isn't in line with the command type.

SaurabhMSFT: I have changed it

fadidurah (approved):

common/build.gradle
```diff
@@ -179,13 +179,15 @@ dependencies {
...
     testImplementation "org.mockito:mockito-core:$rootProject.ext.mockitoCoreVersion"
+    testImplementation "org.mockito.kotlin:mockito-kotlin:3.6.28"
     testImplementation "org.powermock:powermock-module-junit4:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-module-junit4-rule:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-api-mockito2:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-classloading-xstream:$rootProject.ext.powerMockVersion"
     testImplementation "org.robolectric:robolectric:$rootProject.ext.robolectricVersion"
     testImplementation "androidx.test:core:$rootProject.ext.androidxTestCoreVersion"
     testImplementation project(":LabApiUtilities")
+    testImplementation project(":testutils")
```
shahzaibj: `testutils` internally depends on common...so if you use it in common then this would cause circular dependency. Please remove it

common4j/build.gradle
```diff
@@ -219,7 +227,9 @@ dependencies {
     testAnnotationProcessor "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
     testImplementation "junit:junit:$rootProject.ext.junitVersion"
     testImplementation "org.mockito:mockito-core:$rootProject.ext.mockitoCoreVersion"
-
+    testImplementation "io.mockk:mockk:1.11.0"
+    testImplementation "org.mockito:mockito-inline:$rootProject.ext.mockitoCoreVersion"
+    testImplementation "org.mockito.kotlin:mockito-kotlin:4.1.0"
```
shahzaibj: Please define these versions in the version file

common4j/src/main/com/microsoft/identity/common/java/BuildValues.java
```diff
@@ -0,0 +1,56 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.common.java;
+
+import javax.annotation.Nonnull;
+
+//Used as a wrapper for setting and accessing values through either the generated BuildConfig.java class,
+//or from parameters set via the NativeAuthPublicClientApplicationConfiguration.kt file.
+public class BuildValues {
```
shahzaibj, line 29: I remember mentioning that you should remove this class and either use BuildConfig directly or leverage the AadSliceInfo

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
```diff
@@ -67,6 +68,21 @@ public IRequestContext getRequestContext() {
...
+     * @return Correlation id or "UNSET"
+     */
+    public String getThreadCorrelationId() {
+        IRequestContext context = getRequestContext();
+        String threadId = context.get(DiagnosticContext.CORRELATION_ID);
+        if (threadId == null)
+        {
+            threadId = UNSET;
+        }
+        return threadId;
```
shahzaibj:
```suggestion
        final String correlationId = context.get(DiagnosticContext.CORRELATION_ID);
        if (correlationId == null)
        {
           return UNSET;
        }
        return correlationId;
```

shahzaibj (changes requested):
Left some comments. There is a circular dependency being introduced in the code.

common4j/src/main/com/microsoft/identity/common/java/BuildValues.java
SammyO: I'll let Saurabh run this conversation further, but AFAIK this was not the decision. We can't just rely on AadSliceInfo, as we also need a flag for useRealAuthority. And that flag needs to be controllable by a calling app (e.g. sample app), so BuildConfig won't be suited for that.
@SaurabhMSFT please let me know if I've missed anything. This is an important topic to align on.

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
SaurabhMSFT: I have changed the variable name but kept the single return statement. I prefer single return statement in short methods. I also find that the `final` qualifier doesn't add to more readability in short methods.

common/build.gradle
SaurabhMSFT: Done

common4j/build.gradle
SaurabhMSFT: Done

common4j/src/main/com/microsoft/identity/common/java/BuildValues.java
SaurabhMSFT: I have created a PBI to track this work item. We will be doing it in next quarter.

https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2707234

common4j/build.gradle
SaurabhMSFT: I will have to use the version numbers instead of variables otherwise the build for adal and broker will fail. I don't have privileges to modify the gradle version files for those two repositories. I can add the variables in android-complete but not the other two repositories.

common/build.gradle
shahzaibj: Looks like this isn't actually addressed yet. I still see it here in the code.

SaurabhMSFT: I have removed it now. The file got left out during commit.

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/BuildValues.java
SammyO: @shahzaibj are you sure the field you mean is called AadSliceInfo? A ctrl-F for that brings up 0 results in the project.

PR #2237, opened 2023-11-15
Move addition of passkey header from BaseController to WebView fragment

melissaahn:
### Summary
The initial work I did to add the passkey protocol header occurs in the BaseController and the PrtV3 authorization strategy classes. Adding the header in these two places covers the scenarios that MSAL/Broker needs.
However, in their non-brokered auth flow, OneAuth utilizes the `AuthorizationActivityFactory.getAuthorizationActivityIntent` directly, so they are not making use of any controllers in this case (though they do use BrokerMsalController for broker scenarios). (For context, `getAuthorizationActivityIntent` is called down the line by our strategies).  Thus, with the header logic the way it is, OneAuth is going to need to add the passkey header themselves to their `getAuthorizationActivityIntent` call.

They brought up the idea that if I were to move the addition of the passkey header to a class that's covered by `getAuthorizationActivityIntent`, then they would be able to get the logic automatically. 
If I move the passkey header logic to `WebViewAuthorizationFragment`, my understanding is that it would be inherited by both MSAL and broker (so I can also remove the logic from PrtV3 authorization strategy, iiuc) as well as OneAuth. And I also wouldn't have to check for authorization agent == WEBVIEW anymore. But this change would not be correlating to where we typically add headers (which is in BaseController and the PrtV3 auth strategy). 

~~I want to get some input from the team about this before commiting to making such a change.~~ Decided to make the change.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -47,6 +47,7 @@ import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.internal.ui.webview.AzureActiveDirectoryWebViewClient;
 import com.microsoft.identity.common.internal.ui.webview.OnPageLoadedCallback;
 import com.microsoft.identity.common.internal.ui.webview.WebViewUtil;
+import com.microsoft.identity.common.java.constants.FidoConstants;
 import com.microsoft.identity.common.java.ui.webview.authorization.IAuthorizationCompletionCallback;
 import com.microsoft.identity.common.java.providers.RawAuthorizationResult;
 import com.microsoft.identity.common.logging.Logger;
@@ -126,6 +127,14 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
         mAuthorizationRequestUrl = state.getString(REQUEST_URL);
         mRedirectUri = state.getString(REDIRECT_URI);
         mRequestHeaders = getRequestHeaders(state);
+        // In cases of WebView as an auth agent, we want to always add the passkey protocol header.
+        // (Not going to add passkey protocol header until full feature is ready.)
+        if (FidoConstants.IS_PASSKEY_SUPPORT_READY) {
+            if (mRequestHeaders == null) {
+                mRequestHeaders = new HashMap<>();
+            }
+            mRequestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_VALUE);
+        }
         mPostPageLoadedJavascript = state.getString(POST_PAGE_LOADED_URL);
         webViewZoomEnabled = state.getBoolean(WEB_VIEW_ZOOM_ENABLED, true);
         webViewZoomControlsEnabled = state.getBoolean(WEB_VIEW_ZOOM_CONTROLS_ENABLED, true);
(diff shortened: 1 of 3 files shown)


rpdome:
1) Do we plan to support passkey in both standalone MSAL and OneAuth? 

2) Will this be configurable flow by flow? 

--

The original design is letting the method that constructs the request to decide if passkey should be supported. This change means 2) is no longer doable. (e.g. you don't want concurrent requests to modify a global static variable)

If 1) and 2) is not a concern (e.g. every request header from MSAL/OneAuth/Broker would **_always_** have this value), then this seems OK.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -126,6 +127,14 @@ void extractState(@NonNull final Bundle state) {
         mAuthorizationRequestUrl = state.getString(REQUEST_URL);
         mRedirectUri = state.getString(REDIRECT_URI);
         mRequestHeaders = getRequestHeaders(state);
+        // In cases of WebView as an auth agent, we want to always add the passkey protocol header.
+        // (Not going to add passkey protocol header until full feature is ready.)
+        if (FidoConstants.IS_PASSKEY_SUPPORT_READY) {
```
rpdome: put in getRequestHeaders?

melissaahn:
> 1. Do we plan to support passkey in both standalone MSAL and OneAuth?
> 2. Will this be configurable flow by flow?
> 
> --
> 
> The original design is letting the method that constructs the request to decide if passkey should be supported. This change means 2) is no longer doable. (e.g. you don't want concurrent requests to modify a global static variable)
> 
> If 1) and 2) is not a concern (e.g. every request header from MSAL/OneAuth/Broker would **_always_** have this value), then this seems OK.

1. We do plan to support this in standalone MSAL and OneAuth (for 1P apps only). Noting that this header controls whether or not we want to use the custom passkey protocol, which is necessary for cases where WebView is the authorization agent. It does not change the UX of the login page, which is handled by a different query string parameter that is configurable flow by flow.
2. No, it shouldn't be. Android's WebView does not currently have the capability to support WebAuthn calls automatically like CCT does, which is why the custom protocol was built. So anytime a WebView is being used as the auth agent (whether it's for MSAL, OneAuth, or Broker), we want to state to the server that the passkey protocol should be used, should any passkey request come in. 
Another thing about this solution is that if in the future, Android does implement WebAuthn support for WebView, we can turn off this custom protocol in one shot, instead of making the same changes in broker and OneAuth individually.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: Done here: 5bc0a09b6ff9d082411ada6f9b99d9f50e249ecc

rpdome:
> > 1. Do we plan to support passkey in both standalone MSAL and OneAuth?
> > 2. Will this be configurable flow by flow?
> > 
> > --
> > The original design is letting the method that constructs the request to decide if passkey should be supported. This change means 2) is no longer doable. (e.g. you don't want concurrent requests to modify a global static variable)
> > If 1) and 2) is not a concern (e.g. every request header from MSAL/OneAuth/Broker would **_always_** have this value), then this seems OK.
> 
> 1. We do plan to support this in standalone MSAL and OneAuth (for 1P apps only). Noting that this header controls whether or not we want to use the custom passkey protocol, which is necessary for cases where WebView is the authorization agent. It does not change the UX of the login page, which is handled by a different query string parameter that is configurable flow by flow.
> 2. No, it shouldn't be. Android's WebView does not currently have the capability to support WebAuthn calls automatically like CCT does, which is why the custom protocol was built. So anytime a WebView is being used as the auth agent (whether it's for MSAL, OneAuth, or Broker), we want to state to the server that the passkey protocol should be used, should any passkey request come in.
>    Another thing about this solution is that if in the future, Android does implement WebAuthn support for WebView, we can turn off this custom protocol in one shot, instead of making the same changes in broker and OneAuth individually.

you mentioned "for 1p apps only" - what would happen in the case of 3p MSAL app?

shahzaibj (approved):

rpdome (approved):

PR #2239, opened 2023-11-16
always make a copy of bundle before modifying & sending out via ipc

rpdome:
Accidentally discovered this issue on "some" unit test runs. (V1ProtocolPerformPostElectionTaskTest)

As we're broadcasting broker selection result (via IPC) - the broker makes a request bundle, and it would try to invoke IPC to every other broker apps at the same time (concurrent).

The very same bundle is used to create BrokerOperationBundle of every request.

In, BrokerOperationBundle's constructor, every coroutine tries to write a value (KEY_REQUEST_ACTIVE_BROKER_DATA) to the bundle - and it seems like this is NOT thread safe.

As a result, the source bundle is messed up, causing some operation to fail.

------

Therefore, before modifying the source bundle - it is probably safer to make a copy of it first. (do not assume that the source will not be shared across concurrent operations).


tanmaymanolkar1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
```diff
@@ -97,11 +97,25 @@ public BrokerOperationBundle(@NonNull final Operation operation,
...
         this.targetBrokerAppPackageName = targetBrokerAppPackageName;
-        this.bundle = bundle != null ? bundle : new Bundle();
+        this.bundle = deepCopy(bundle);
         this.bundle.putBoolean(KEY_REQUEST_ACTIVE_BROKER_DATA, true);
         Logger.info(TAG, "Requested Active Broker Data");
     }
 
+    // To be replaced by bundle.deepCopy() (requires API 26)
+    @NonNull
+    Bundle deepCopy(@Nullable final Bundle source) {
```
somalaya: can this be a private method?

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/BrokerOperationBundle.java
rpdome: I'll move this out to a separate util class + add test.

rpdome: there's actually a better way to do this (sourceBundle.clone(), or new Bundle(sourceBundle))

PR #2243, opened 2023-11-27
Select MSA config in MsalTestApp automation

somalaya:
LTW tests need to be run with MSA accounts as well, so this PR has changes to support selecting the MSA configuration in MSALTestAPp as part of first run experience handling.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/MsalTestApp.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/MsalTestApp.java
@@ -39,6 +39,8 @@ import com.microsoft.identity.client.ui.automation.interaction.PromptHandlerPara
 import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.MicrosoftStsPromptHandler;
 import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.MicrosoftStsPromptHandlerParameters;
 import com.microsoft.identity.client.ui.automation.utils.UiAutomatorUtils;
+import com.microsoft.identity.labapi.utilities.constants.UserType;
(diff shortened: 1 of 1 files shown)


fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/app/MsalTestApp.java
```diff
@@ -221,4 +223,12 @@ protected void initialiseAppImpl() {}
...
         UiAutomatorUtils.handleButtonClick("com.msft.identity.client.sample.local:id/btnStartTask");
     }
+
+    // Handles first run of the app based on the user account type to be used.
+    public void handleFirstRunBasedOnUserType(UserType userType) throws UiObjectNotFoundException {
+        handleFirstRun();
+        if (userType == UserType.MSA) {
+            selectFromConfigFile("MSA");
+        }
+    }
```
fadidurah, line 233: Is this method needed? Seems more clear to use `handleFirstRun()` then ` selectFromConfigFile("MSA")` in the test case itself, that way devs can see exactly what is happening.

somalaya: IMO, selecting a config based on account type should be part of handling first run experience rather than explicitly selecting it in every test case.

fadidurah: Hmmm, I think that's fair. I also realize now that you are moving towards parameterized tests, so this approach will help keep tests uncluttered. Let's go with your existing implementation.

rpdome (approved):

PR #2246, opened 2023-11-30
Add tests for Native Auth

SaurabhMSFT:
This PR includes the unit and integration tests for Native Auth. This PR also includes changes done for telemetry and expose annotation for few response types.
mockk has been added as a new test dependency


azure-pipelines/pull-request-validation/common4j.yml
```diff
@@ -38,5 +38,5 @@ jobs:
     displayName: Run the unit tests.
     inputs:
       cwd: $(Build.SourcesDirectory)/common-java-root
-      tasks: common4j:test -Psugar=true
+      tasks: common4j:test -Psugar=true -PuseMockApiForNativeAuth=true
```
fadidurah, line 41: Perhaps we can set the default value for this to true, so that we don't have to explicitly pass it. We can change the default to false once prod api integration testing is available.

fadidurah: I imagine we will also need this added to the common consumers pipeline as well

fadidurah (approved):

azure-pipelines/pull-request-validation/common4j.yml
SaurabhMSFT: Boolean flags should all have a common default value of false as it allows a casual user of the library to use it without having to understand the intricacies of those flags. For this reason I have set the default value of this flag to false.

common/src/main/java/com/microsoft/identity/common/internal/commands/BaseNativeAuthCommand.kt
```diff
@@ -64,8 +64,7 @@ abstract class BaseNativeAuthCommand<T>(
         return false
     }
 
-    // TODO update this once telemetry is done
```
shahzaibj: Are you integrating any telemetry in this PR?

SaurabhMSFT: yes. we had made some changes for telemetry after our last merge and this PR includes those changes.

SammyO (approved):

PR #2255, opened 2023-12-08
Catch unknown Keystore errors during decryption

somalaya:
Issue : https://portal.microsofticm.com/imp/v3/incidents/incident/443725136/summary
In the above IcM, we are seeing some unknown keystore errors during keypair generation. This is only happening when KS is getting accessed during app boot up.  we do not have a repro for the issue and it is happening on 0.01% of devices. We have considered 3 ways of fixing this.

1. Find the actual root cause of the issue. For this, we have created a bug on google. This may take some time and we would need to come up with a short term solution
2. Since the KS errors are only happening on Android 10 and 11, we can setBrokerDiscovery as false on these versions in OneAuth. Howver, this would need to be reverted when we need to enable broker discovery for LTW.
3. Catch the KS errors and throw a client exception like we did in this PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2190.


rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
```diff
@@ -333,6 +333,9 @@ private byte[] decryptWithSecretKey(final byte[] encryptedBlobWithoutEncodeVersi
         } catch (final IllegalArgumentException e) {
             errCode = DATA_MALFORMED;
             exception = e;
+        } catch (final Throwable e) {
```
shahzaibj: Do we not know the specific type of Exception we are getting in the issue?

shahzaibj: I'm asking because not a fan of errors that say unknown

somalaya: This one is a provider exception. So let me just catch provider exception here. But that means that we may not be catching any new keystore exceptions that may pop up in the future and are android OS bugs.

shahzaibj: The error code you've picked though still seems to indicate something unknown happened. Can we use a more specific error code such that when we look at it in telemetry then we would immediately know what happened?

somalaya: This is in common code and is happening when OneAuth consumes common. That is why I could not get telemetry on the issue

somalaya: As discussed with you offline, this issue has a very low impact. Also, the exception is getting suppressed after being thrown so there is no point in throwing a more specific exception. I have moved the catch block to a different place though, to specifically handle this for Android 10 and 11 (Previously I added catch block in StorageEncryptionManager class which is in common4j)

shahzaibj (approved):

PR #2258, opened 2023-12-12
Request camera permission on the WebViewAuthorizationFragment

p3dr0rv:
Why? 
When a request is made to ESTS with Preferred authorization method "QR code + PIN", ESTS UX will return a page were the user needs to scan a QR code

What is in this PR? 
Added handler to the WebView for when a camera request is detected.

related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2673


common/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,8 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <!-- Camera permission is required for QR + PIN authorization method -->
+    <uses-permission android:name="android.permission.CAMERA" />
```
rpdome: Is QR + PIN feature exclusive to broker, or is this also applicable to oneauth + MSAL?

I'm asking because this would be displayed on all consuming app's permission that they now require access to camera. (for those who are not aware of this QR+Pin feature... it could sound scary)

If it's broker exclusive, let's put the camera code in broker and wire the code here (should be easier to talk to broker apps about this).

If not, this is (kind of) a breaking change. we can discuss this in the standup whether calling out this new permission in the changelog is enough, or do we expect any pushback from customers.

p3dr0rv: this is exclusive to broker so I would need to move this permission to the broker my b.
I can communicate this to them over email.
Also, the camera permission will not be requested at the first time the app is used like other permissions.
The camera permission is a [Runtime permission](https://developer.android.com/guide/topics/permissions/overview#runtime), therefore the permission must be requested at runtime once the permission is required. and because right now there is no eSTS UX that requires the camera I don't expect this permission request to show up.

rpdome: I mean it would still show up on the Play Store page 
i.e.
Screenshot 2024-01-25 at 12 42 22 PM

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +237,113 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+    }
+
+    /**
+     * Shows a dialog to the user explaining why the camera permission is required.
+     * If the user accepts the dialog, the camera permission request will be launched.
+     * If the user denies the dialog, the camera permission request will be denied.
+     */
+    private void showCameraRationale() {
+        final AlertDialog.Builder builder = new AlertDialog.Builder(getContext());
+        builder.setMessage("The app requires the camera permission to scan the QR Code, " +
```
rpdome: Every displayed strings would need to be localized.
(I assume the message is already green lit by PM?)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +237,113 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.getSettings().setBuiltInZoomControls(webViewZoomControlsEnabled);
         mWebView.getSettings().setSupportZoom(webViewZoomEnabled);
         mWebView.setVisibility(View.INVISIBLE);
+        mWebView.getSettings().setPluginState(WebSettings.PluginState.ON);
         mWebView.setWebViewClient(webViewClient);
+        mWebView.setWebChromeClient(new WebChromeClient() {
```
rpdome, line 247: nit: You might want to add log so that you know if the feature doesn't work due to any permission issue (e.g. denied, etc)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +237,113 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+            @Override
+            public void onPermissionRequest(final PermissionRequest request) {
+                // We can only grant or deny permissions for video capture/camera.
+                // To avoid unintentionally granting requests for not defined permissions.
+                if (iSPermissionRequestForCamera(request)) {
+                    mCameraPermissionRequest = request;
+                    if (isCameraPermissionGranted()) {
+                        acceptCameraRequest();
+                    } else if (shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)) {
+                        showCameraRationale();
```
rpdome, line 273: after requestCameraPermission(), do we have to invoke acceptCameraRequest()?

p3dr0rv: no, the user can deny or allow the permission. this is handled by the cameraRequestActivity

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: no, checking this right now.

common/src/main/AndroidManifest.xml
p3dr0rv: both teams are informed about this change.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2258/commits/3c1aad63823f6e287848a3972d69b76a3f833e1b
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2258/commits/26649bba7a24df0159f9bdc8d0291944b5352971

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,117 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+                        requestCameraPermissionFromUser();
+                    }
+                }
+            }
+        });
+    }
+
+    /**
+     * Call this method to grant the permission to access the camera resource.
+     * The granted permission is only valid for the current WebView.
```
shahzaibj, line 284: What does `current` mean here? On next launch of WebView will the user need to grant permission again?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,117 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
     }
 
+    /**
+     * Determines whatever if the given permission request is for the camera resource.
+     *
+     * @param request The permission request.
+     * @return true if the given permission request is for camera, false otherwise.
+     */
+    private boolean iSPermissionRequestForCamera(final PermissionRequest request) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
shahzaibj, line 332: Can you add some comments about OS versions checks?

It seems if the OS is not from among that range then we are assuming permission request isn't for camera by hard code returning false

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: That will depend on the behavior the user has chosen; the user can approve the camera permission just for this request or for this and all future requests.
image

p3dr0rv: If the user decided to be asked every time, every time the user launches the WebView will be prompted for camera permission, if the user chooses the option 1, the WebView automatically will grant this permission. see L252 to 260

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: I added a comment with more context about this version check. 
```
     * Note: This method is only available on API level 21 or higher.
     * Devices running on lower API levels will not be able to grant or deny camera permission requests.
     * getResources() method is only available on API level 21 or higher.
```
and a warning message
```
        Logger.warn(TAG, "PermissionRequest.getResources() method is not available on API:"
                + Build.VERSION.SDK_INT + ". We cannot determine if the request is for camera.");
```

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,127 @@ public boolean onTouch(final View view, final MotionEvent event) {
         mWebView.getSettings().setBuiltInZoomControls(webViewZoomControlsEnabled);
         mWebView.getSettings().setSupportZoom(webViewZoomEnabled);
         mWebView.setVisibility(View.INVISIBLE);
+        mWebView.getSettings().setPluginState(WebSettings.PluginState.ON);
```
rpdome, line 242: This one is deprecated right?
https://developer.android.com/reference/android/webkit/WebSettings.html#setPluginState%28android.webkit.WebSettings.PluginState%29

rpdome: This one is deprecated right?
https://developer.android.com/reference/android/webkit/WebSettings.html#setPluginState%28android.webkit.WebSettings.PluginState%29

do we need this line?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,127 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+     * Determines whatever if the given permission request is for the camera resource.
+     * <p>
+     * Note: This method is only available on API level 21 or higher.
+     * Devices running on lower API levels will not be able to grant or deny camera permission requests.
+     * getResources() method is only available on API level 21 or higher.
+     *
+     * @param request The permission request.
+     * @return true if the given permission request is for camera, false otherwise.
+     */
+    private boolean iSPermissionRequestForCamera(final PermissionRequest request) {
```
rpdome: nit: i"s"

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,127 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+        mWebView.getSettings().setPluginState(WebSettings.PluginState.ON);
         mWebView.setWebViewClient(webViewClient);
+        mWebView.setWebChromeClient(new WebChromeClient() {
+            @Override
+            public void onPermissionRequest(final PermissionRequest request) {
+                // We can only grant or deny permissions for video capture/camera.
+                // To avoid unintentionally granting requests for not defined permissions.
+                if (iSPermissionRequestForCamera(request)) {
+                    Logger.info(methodTag, "Camera request received.");
+                    mCameraPermissionRequest = request;
```
rpdome, line 267: nit: this doesn't need to be a class variable.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -227,9 +239,127 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+     * Note: This method is only available on API level 21 or higher.
+     * Devices running on lower API levels will not be able to grant or deny camera permission requests.
+     * getResources() method is only available on API level 21 or higher.
+     *
+     * @param request The permission request.
+     * @return true if the given permission request is for camera, false otherwise.
+     */
+    private boolean iSPermissionRequestForCamera(final PermissionRequest request) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            return request.getResources().length == 1 &&
```
rpdome, line 333: would it be safer to use request.getResources().contains(RESOURCE_VIDEO_CAPTURE) (or something along the line?)

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: not removed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: this is used on cameraRequestActivity, so I cannot move it.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: this is a primitive array so I cannot use contains, I would need to convert to listArray.
Also, in this way we can check the request is only for one permission.

shahzaibj (approved):

PR #2260, opened 2023-12-15
Run mock API tests without cmd line param 

SaurabhMSFT:
MSAL counterpart https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1980

### DIFF
--- a/azure-pipelines/pull-request-validation/common.yml
+++ b/azure-pipelines/pull-request-validation/common.yml
@@ -50,7 +50,7 @@ jobs:
     inputs:
       javaHomeSelection: JDKVersion
       jdkVersionOption: "1.11"
-      tasks: common:testLocalDebugUnitTest -PuseMockApiForNativeAuth=true
+      tasks: common:testLocalDebugUnitTest
   - task: Gradle@3
     displayName: Check Dependencies size
     condition: eq(variables['system.pullRequest.targetBranch'], 'dev')
--- a/azure-pipelines/pull-request-validation/common4j.yml
+++ b/azure-pipelines/pull-request-validation/common4j.yml
@@ -38,5 +38,5 @@ jobs:
     displayName: Run the unit tests.
     inputs:
       cwd: $(Build.SourcesDirectory)/common-java-root
-      tasks: common4j:test -Psugar=true -PuseMockApiForNativeAuth=true
+      tasks: common4j:test -Psugar=true
 ...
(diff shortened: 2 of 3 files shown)


common/src/test/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthControllerTest.kt
```diff
@@ -130,6 +134,20 @@ class NativeAuthControllerTest {
     @get:Rule
     var rule: MockitoRule = MockitoJUnit.rule()
 
+    companion object {
+        @BeforeClass
+        @JvmStatic
+        fun setupClass() {
+            BuildValues.setUseRealAuthority(true)
```
SaurabhMSFT: Will use new method signature for `setUseRealAuthority` when its refactoring PR gets merged.

azure-pipelines/pull-request-validation/common.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     inputs:
       javaHomeSelection: JDKVersion
       jdkVersionOption: "1.11"
-      tasks: common:testLocalDebugUnitTest -PuseMockApiForNativeAuth=true
+      tasks: common:testLocalDebugUnitTest
```
SammyO, line 53: Why are we disabling this flag from unit tests? We definitely need this for (some) of our unit tests.

On what build does `testLocalDebugUnitTest` run? If it's a release build, then you're right that setting `PuseMockApiForNativeAuth` would cause it to be set in the BuildConfig, which indeed we don't want. But if that's the case, then the solution isn't to disable the mock API altogether, but to create a test build. As we need to be able to continue to run the tests that need the mock API.

SaurabhMSFT: This flag is not needed as we are enabling the mockAPIs for our test when needed. `NativeAuthControllerTest` is the only class in this test project that requires the mockAPI value to be true in BuildValues class. For `NativeAuthControllerTest` we are enabling the value before running any test and disabling once all tests have run.

azure-pipelines/pull-request-validation/common4j.yml
```diff
@@ -38,5 +38,5 @@ jobs:
     displayName: Run the unit tests.
     inputs:
       cwd: $(Build.SourcesDirectory)/common-java-root
-      tasks: common4j:test -Psugar=true -PuseMockApiForNativeAuth=true
```
SaurabhMSFT, line 41: No existing tests in common4j project were dependent on this flag.

azure-pipelines/pull-request-validation/common.yml
SammyO: I'm not following. "When needed", is when, where? 

"For NativeAuthControllerTest we are enabling the value before running any test and disabling once all tests have run." - where is this done?

SaurabhMSFT: Line 137-150 in `NativeAuthControllerTest.kt` file we are enabling the mock API flag for the duration of test run and then disabling it.

SammyO (approved):

Yuki-YuXin (approved):

PR #2264, opened 2023-12-27
Changes to launch AEA app from Broker's WebView

shahzaibj:
This is needed to unblock our partner. Broker will launch the AEA app when it receives the above redirect with scheme aea://  (essentially by launching an Intent for this URI). 

At this stage, the AEA app will come to foreground and user can complete their action in the app.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -359,6 +367,14 @@ private void launchCompanyPortal() {
         returnResult(RawAuthorizationResult.ResultCode.MDM_FLOW);
     }
 
+    private void processAmazonAppUri(@NonNull final String url) {
```
iamgusain, line 392: Do we need to explicitly stop/close the webview after this, or that is something that is already happening.

shahzaibj: No. The webview needs to keep running. The user returns back to the flow after completing action in AEA app.

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
iamgusain: I see, so the web view is showing a page that indirectly waits for user performing an action in AEA app.

shahzaibj: Yes, correct.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -178,6 +179,9 @@ private boolean handleUrl(final WebView view, final String url) {
             } else if (isAuthAppMFAUrl(formattedURL)) {
                 Logger.info(methodTag,"Request to link account with Authenticator.");
                 processAuthAppMFAUrl(url);
+            } else if (isAmazonAppRedirect(url)) {
```
melissaahn: isAmazonAppRedirect(formattedURL)?

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2264/commits/b7d387cb12b8a6255f8c95d15d282cd8f0be98a4

PR #2266, opened 2023-12-28
Setting theme for Ntlm dialog

melissaahn:
### Summary
We've received various issues regarding the text in the Ntlm dialog being invisible on certain Teams devices.
From some of my debugging, it seems that the dialog will inherit the styling of the host app. In an effort to help resolve these issues and to help keep the styling consistent, this PR creates and enforces a theme upon the Ntlm dialog that hardcodes the text to be black and the background of the dialog window to be white.
I verified that the dialog theme is consistent among three of our test apps, and loaded a site that invokes a test http auth request (https://www.httpwatch.com/httpgallery/authentication/) for testing.


rpdome:
Are we currently forcing only text or background color?

Dialog theme should be determined by the app, but it's weird for the text to be white-on-white (or black-on-black).

melissaahn:
> Are we currently forcing only text or background color?
> 
> Dialog theme should be determined by the app, but it's weird for the text to be white-on-white (or black-on-black).

Forcing text to be black and window background color to be white (updated description). We've gotten at least two incidents in the past regarding a white-on-white issue in the edittext, both on Teams devices iirc. I can't say I'm totally sure that this theme will fix the issue, but it should at least bring some consistency to the dialog.

shahzaibj (approved):

PR #2267, opened 2023-12-28
CredManFidoManager

melissaahn:
### Summary
This PR adds the logic for CredManFidoManager, which makes use of a new dependency [Credential Manager](https://developer.android.com/training/sign-in/passkeys#:~:text=Credential%20Manager%20is%20a%20Jetpack,simplifying%20the%20integration%20for%20developers.) in order to get an assertion response from Google password manager and other passkey providers (such as Authenticator).  It also enables the existing passkey logic that has been added over the past few months.

This is considered a major change due to Credential Manager enforcing a min Compile sdk level of 34, which is transitive to us and all of our consumers. This API is also contributing to a larger size increase, and the max dep size check limit has been updated accordingly (more information in the size increase section below). I have been in communication with broker apps, OneAuth, and 1P apps regarding these changes.
This PR also contains some smaller updates to accommodate some observed bugs that need to be investigated further with the server side. 
Future PRs will (hopefully) include UI automated tests to test Credential Manager, but until then, I will conduct manual testing for the e2e passkey feature.

#### Size Increase
The Android Credential Manager API is going to contribute to a noticeable size increase in our libraries and consumers. After some initial generation of test apps and one of the broker apps with the dependency and associated passkey logic, the current estimated size increase in release apps will be around 300 - 550 KB.

The reason why we're including this new dependency is because it will provide passkey support to our 1P native apps currently using MSAL, OneAuth-MSAL, or relying on broker support. CredMan handles the interactions with the device authenticator as well as allows us to interact with 3P passkey providers (such as Microsoft Authenticator, in the near future). Upon GA, passkeys will widely be pushed as an alternative auth method to passwords, for both Entra ID and MSA users, across nearly all platforms. 

We do not currently have suggestions to avoid this size increase for our consumers; release test builds with just the inclusion of the dependency itself (no logic utilizing it, and no proguard rules) resulted in a size increase very near to the range mentioned earlier. But more suggestions/ideas are certainly welcome.


rpdome:
Proguard doesn't work I assume?

rpdome:
Does proguard work for lite apps?

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -67,9 +68,13 @@ class WebAuthnJsonUtil {
          * @param fullResponseJson AuthenticationResponse Json string.
          */
         fun extractAuthenticatorAssertionResponseJson(fullResponseJson : String): String {
-            return JSONObject(fullResponseJson)
+            val fullResponseJsonObject = JSONObject(fullResponseJson);
+            val authResponseJsonObject = fullResponseJsonObject
                 .getJSONObject(FidoConstants.WEBAUTHN_AUTHENTICATION_ASSERTION_RESPONSE_JSON_KEY)
-                .toString()
+                // Making sure that Id is here because ESTS expects it.
+                // I've noticed that GPM will sometimes not include the id in the response object.
```
rpdome: google's bug? should we file it?

gradle/versions.gradle
```diff
@@ -78,6 +78,8 @@ ext {
     openTelemetryVersion = "1.18.0"
     jetpackDataStoreVersion = "1.0.0"
     lifecycleKtxVersion="2.5.1"
+    AndroidCredentialsVersion="1.2.0"
```
rpdome, line 81: would these always be the same version?

rpdome (approved):

iamgusain:
Can you create a separate section in the PR description for size increase, containing the details on 
- how much the size is going to increase, 
- why we must take this size increase change (add brief detail and importance of the feature it is used for), 
- is it possible to avoid the size increase for our consumers.

So that we can include the information in release notes as well.
Thanks

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -38,15 +45,23 @@ class CredManFidoManager (val context: Context) : IFidoManager {
...
             challenge,
             relyingPartyIdentifier,
             allowedCredentials,
             userVerificationPolicy
         )
-        // TODO: Implementation will include Credential Manager calls; we aren't able to add the dependency yet,
-        //  so this method will only be filled once we can finish up the passkey work.
-        //  ADO Item: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2630696
-        return WebAuthnJsonUtil.extractAuthenticatorAssertionResponseJson("")
+        val publicKeyCredentialOption = GetPublicKeyCredentialOption(
```
iamgusain, line 54: nit: can you add the type information for each variable. I know we can get it by looking at the return type of the method but would be nice if I can see it in place as well.

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: Once folks get back from OOF, I'm going to ask internally if this is a bug or not. I'm looking through official Webauthn documentation, and I'm not finding concrete info on id as a field in authentication assertion responses (though it is clearly documented as a field in the outer JSON object). Our server side definitely expects it though, so I'm trying to a workaround until I can get some answers.

gradle/versions.gradle
melissaahn: I had separate defined versions in case they ended up being different, but seems they intend to keep the versions the same, as of the stable release. Will use AndroidCredentialsVersion for both dependencies: 8c9f0190b49ae2fe1d2a3907c016bbce6186ce73

melissaahn:
> Does proguard work for lite apps?

Most likely not, based on testing with generating our release test apps and one broker app with just the dependency added (no proguard rules to keep things, and no logic that used the dependency). But I also confirmed with passkey folks that it is expected for 1P lite apps to take in the passkey feature anyway, since passkeys are going to be pushed as a password alternative for both Entra and MSA.


> Can you create a separate section in the PR description for size increase, containing the details on
> 
> * how much the size is going to increase,
> * why we must take this size increase change (add brief detail and importance of the feature it is used for),
> * is it possible to avoid the size increase for our consumers.
> 
> So that we can include the information in release notes as well. Thanks

Updated the description with a more detailed sections pertaining to the size increase and reasoning.

PR #2281, opened 2024-01-15
Update Native Auth password reset API response

rmccahill:
This PR updates the relevant response classes to include the continuation token returned from the updated poll completion response, and updates constructors in test classes to reflect the change. 

Note: I've also updated the test runner to Robolectric from PowerMock for the reset password test class to be in line with other test cases, as it was causing test failures locally.


common/src/test/java/com/microsoft/identity/common/internal/providers/microsoft/nativeauth/integration/ResetPasswordOAuth2StrategyTest.kt
```diff
@@ -81,7 +81,7 @@ import java.util.UUID
 
 
 @RunWith(
-    PowerMockRunner::class
+    RobolectricTestRunner::class
```
SammyO, line 84: Why this change?

rmccahill: From the description above: 

> I've also updated the test runner to Robolectric from PowerMock for the reset password test class to be in line with other test cases, as it was causing test failures locally.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 V.Next
 ---------
 - [MINOR] Implement updates of the native auth web API (#2261)
+- [MINOR] Implement update of the native auth web password reset API response (#2281)
```
Yuki-YuXin: I personally feel that this description is part of the previous one. Could we use more specific words like
"Update ResetPasswordPollCompletion related command, results, response to supports automatic login after password reset."

rmccahill: Good point - I've updated the description to better reflect the changes

Yuki-YuXin (approved):

SammyO (approved):

PR #2282, opened 2024-01-17
Added support for setting no log level

somalaya:
**Issue** : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2642644

**Fix** : Adding support for NO_LOG level. 

**Testing** : Added UTs and also shared the common build with OneAuth team to test their requirement.


common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -378,7 +383,7 @@ private static void log(final String tag,
                             final String message,
                             final Throwable throwable,
                             final boolean containsPII) {
-        if (logLevel.compareTo(sLogLevel) > 0 || (!sAllowPii && containsPII)) {
+        if (logLevel.compareTo(sLogLevel) > 0 || (!sAllowPii && containsPII)  || (sLogLevel == LogLevel.NO_LOG)) {
```
mohitc1: >(sLogLevel == LogLevel.NO_LOG) [](http://example.com/codeflow?start=81&length=30)

Can be optimized by having this as first check

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -108,6 +108,11 @@ public enum LogLevel {
          * Undefined. Should be used in test only.
          */
         UNDEFINED,
+
+        /**
+         * No logs should be emitted
+         */
+        NO_LOG
```
mohitc1: >NO_LOG [](http://example.com/codeflow?start=8&length=6)

should this be first entry in the enum to ensure e1.compareTo.e2() checks work as expected?

somalaya: The comparison would not fail. But I moved it to first entry as I need to order it from most restrictive to least restrictive log level.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
somalaya: done

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
mohitc1: on both files?

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
shahzaibj: Well, since you moved it this means that each value in this file now has got a new ordinal value. I suggest checking with OneAuth that they are not relying on the ordinal value.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -88,6 +88,10 @@ public static void setPlatformString(String platformString) {
      * Enum class for LogLevel that the sdk recognizes.
      */
     public enum LogLevel {
+        /**
+         * No logs should be emitted
```
shahzaibj:
```suggestion
         * No logs should be emitted.
```

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
matthewyu-ms: Should be fine. These are the uses of Common's Logger values here:
https://office.visualstudio.com/OneAuth/_git/OneAuth?path=/sources/public/android/java/com/microsoft/authentication/OneAuth.java&version=GBdev&line=647&lineEnd=648&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

and

https://office.visualstudio.com/OneAuth/_git/OneAuth?path=/sources/public/android/java/com/microsoft/authentication/OneAuth.java&version=GBdev&line=698&lineEnd=699&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Both of these locations don't use the ordinal values of Common's LogLevel enums

shahzaibj (approved):

PR #2283, opened 2024-01-17
Consistent correlation ID

SammyO:
Make correlation ID consistent throughout a flow, meaning all API request in a scenario should use the same correlation ID
- Correlation ID is taken from API request headers, rather than request body (as it's only present in the body in case of errors, whereas we need it always).
- Add it to logging 
- Add it to CommandParameters

Note:
This PR doesn't contain tests that validate the correlation ID using the mock API. When we introduce automated tests on production environment, tests for correlation ID should be added.

PR for MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2007


common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignInCommandResult.kt
```diff
@@ -39,27 +39,50 @@ sealed interface SignInSubmitPasswordCommandResult: INativeAuthCommandResult
  * @see com.microsoft.identity.common.java.nativeauth.providers.responses.signin.SignInTokenApiResult
  */
 interface SignInCommandResult {
-    data class Complete(val authenticationResult: ILocalAuthenticationResult) :
+    data class Complete(
+        override val correlationId: String? = null,
```
SaurabhMSFT: Why does the `correlationId` variable has a default value in this class but not other classes in this file?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/IApiResponse.kt
```diff
@@ -22,10 +22,19 @@
...
-    @Transient open val statusCode: Int
-)
+    @Transient open val statusCode: Int,
+    correlationId: String?
+) {
+
+    // Custom setter. Limit access to this method, to avoid accidental overwrite.
+    var correlationId = correlationId
+        internal set
+}
```
SaurabhMSFT, line 40: missing newline at the end of the file.

common/src/test/java/com/microsoft/identity/common/internal/providers/microsoft/nativeauth/integration/scenario/SignUpScenarioTest.kt
```diff
@@ -115,49 +113,55 @@ class SignUpScenarioTest {
...
     fun testSignUpScenarioEmailPasswordWithOOBVerification() {
         var continuationToken: String
-        val correlationId = UUID.randomUUID().toString()
+        var correlationId: String? = UUID.randomUUID().toString()
+        val mockApiClientRequestId = UUID.randomUUID().toString()
 
         configureMockApi(
             endpointType = MockApiEndpoint.SignUpStart,
-            correlationId = correlationId,
+            correlationId = mockApiClientRequestId,
```
Yuki-YuXin: Why do we have both correlationId and mockApiClientReuestId?

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
```diff
@@ -71,6 +71,7 @@ class NativeAuthRequestHandlerTest {
     private val emptyString = ""
     private val emptyPassword = "".toCharArray()
     private val continuationToken = "uY29tL2F1dGhlbnRpY"
+    private val correlationId = "jsdfo4nslkjsrg"
```
Yuki-YuXin, line 73: If we have extra time, maybe we can adopt uppercase norms.

common/src/test/java/com/microsoft/identity/common/internal/providers/microsoft/nativeauth/integration/scenario/ResetPasswordScenarioTest.kt
```diff
@@ -156,11 +160,13 @@ class ResetPasswordScenarioTest {
         val mockResetPasswordSubmitCodeCommandParameters = mockk<ResetPasswordSubmitCodeCommandParameters>()
         every { mockResetPasswordSubmitCodeCommandParameters.getContinuationToken() } returns continuationToken
         every { mockResetPasswordSubmitCodeCommandParameters.getCode() } returns oobCode
+        every { mockResetPasswordSubmitCodeCommandParameters.getCorrelationId() } returns correlationId
         val ssprContinueResult = nativeAuthOAuth2Strategy.performResetPasswordContinue(
             mockResetPasswordSubmitCodeCommandParameters
         )
         Assert.assertTrue(ssprContinueResult is ResetPasswordContinueApiResult.PasswordRequired)
         continuationToken = (ssprContinueResult as ResetPasswordContinueApiResult.PasswordRequired).continuationToken
+        correlationId = ssprStartResult.correlationId ?: UUID.randomUUID().toString()
```
GBakolas: why are we generating a new correlation id before each request?

SammyO: we're not, only if it's null. Do you think we shouldn't?
@diegojerezba  how are you doing this on iOS?

diegojerezba: When creating requests: 

On iOS we have nullable correlationIds only in the public interface. If the users don't set them, we generate them.
After that, our correlationId fields that we pass through functions cannot be null.

When decoding the correlationId from the headers:

Our API models allow for nullable correlationIds. If the correlationId is null, we reuse the same correlationId that was sent during the request (ie: the correlationId that was set in the public interface, either by the user or generated by us)

When dealing with States:

We don't allow users to set a new correlationId in the intermediate states (ie: submitPassword()). We pass a non-null correlationId from the previous request (that we got either from the headers or was set in the public interface)

SammyO: > Our API models allow for nullable correlationIds. If the correlationId is null, we reuse the same correlationId that was sent during the request (ie: the correlationId that was set in the public interface, either by the user or generated by us)

this is a good point, which we're not doing right now. if the API returns null, it stays null. I'll fix that.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordChallengeApiResponse.kt
```diff
@@ -39,6 +38,7 @@ import java.net.HttpURLConnection
  */
 class ResetPasswordChallengeApiResponse(
     @Expose override var statusCode: Int,
+    correlationId: String?,
```
SaurabhMSFT: `@Expose` annotation can be added to `correlationId` as it does not contain any PII. This comment applies to all instances of correlation id in response classes.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordChallengeApiResult.kt
```diff
@@ -23,41 +23,51 @@
...
 
     data class CodeRequired(
+        override val correlationId: String?,
         val continuationToken: String,
         val challengeTargetLabel: String,
         val challengeChannel: String,
         val codeLength: Int) : ResetPasswordChallengeApiResult
 
     data class UnsupportedChallengeType(
+        override val correlationId: String?,
```
SaurabhMSFT: Can `correlationId` be the last (third) parameter as it will then align with the sequence in the base class?

SaurabhMSFT (changes requested):

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordChallengeApiResponse.kt
SammyO: `@Expose` marks a field as required for JSON serialization or deserialization. Since this field is not an API response field (but rather, the value is taken from the response header, and manually set into this class), it doesn't need this annotation.
Coincidentally MSAL uses this annotation to determine whether a field should be logged or not, which in theory it can. But IMO that's not the right approach (a different annotation should be used for that).

SammyO: Ah, that feature is not enabled by default in GSON and MSAL never sets it. So this annotation is actually _only_ used to PII logging. I'll update it 👍

SammyO: Ah, GSON doesn't ignore fields with this annotation by default, it's a setting that has to be enabled. And MSAL uses this _only_ during logging. So we can indeed mark this as `@Expose`.

An improvement to our code would be to include `excludeFieldsWithoutExposeAnnotation()` when constructing our native auth responses, so fields that are actually not present in the API response such as `statusCode` and `correlationId` are ignored.

SammyO: But - I can't add `@Expose`, because of:
image

This happens because of `IApiResponse`, which hides the field and exposes only internally a setter. This is to avoid classes overwriting this value. Ideally it would be a final value, but this is not possible due to how `NativeAuthResponseHandler` is designed, which creates objects through `ObjectMapper.deserializeJsonStringToObject`. This leaves us with no change to set the correlationId param if it was a val.

common/src/test/java/com/microsoft/identity/common/internal/providers/microsoft/nativeauth/integration/scenario/SignUpScenarioTest.kt
SammyO: Good question. These tests are failing for me locally, I don't understand why they're not failing in the pipeline.

SammyO: They are failing, didn't see that. Thanks for pointing out.

rmccahill (approved):

SaurabhMSFT (approved):

fadidurah (approved):

PR #2288, opened 2024-01-18
 Rename "is QR + PIN available" API to "get preferred auth method".

p3dr0rv:
After a discussion with the IOs team and OneAuth team we decided to change:
1. The API to return a more general PreferredAuthMethod enum rather than QRPINAvailability. This is to attempt to be extensible in case some other similar method is created as part of current discussions, such as NFC.
2. The admin config will reflect this - we'd ask for "preferred_auth_config= QRPIN" rather than "qrPinAvailability = true".


common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -145,11 +144,11 @@ public abstract boolean getDeviceMode(final CommandParameters parameters)
...
 
     /**
-     * This method is used to determine if the QR + PIN auth flow is available on the device.
+     * This method is used to read the preferred authentication method from the broker.
      *
-     * @return true if if QR + PIN authorization is available. False otherwise.
+     * @return The preferred authentication method.
      */
-    public abstract boolean isQrPinAvailable() throws Exception;
+    public abstract PreferredAuthMethod getPreferredAuthMethod() throws Exception;
```
tongyuze, line 151: Do we want to use "preferredAuthMethod" for here or using a different "preferredAuthConfiguration" to separate these two concepts (configuration from the device info and auth method sent from the parameters)

p3dr0rv: @tongyuze 
I think those as one, if we want to separate them in to 2 concepts that's mean to have 2 types of objects to maintain with the same information.


thepreferred auth method is the desire auth method to use by the user.
and we can operate with this to:
1 specify the "preferred auth method" when we make a token request.

2 obtain the "preferred auth method" configured by the MDM.

I think I can rename the method on point 2 to be more explicit.

thoughts?

tongyuze: I tend to agree. Let's sync with Mihai to see if he has any concern since I saw his change separated them. https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/1929/files#diff-74d19a29e3fdb6d3f69d35ea7972e9c6e2f5f45bd255e38b66213063579e6ba3

mipetriu: I separated them in my PR because to me they are different concepts - preferredAuthConfiguration is what was configured by the admin and preferredAuthMethod is what the app decides for its token request. You are suggesting that the same enum represents both cases, right?

tongyuze: @mipetriu 
Could these two concepts differ in the future (have different values)? The configuration is also used to setting the preferred auth method, I don't see there are big differences between them. But I do understand what you thought. I'm Ok with either option.

p3dr0rv: Final note. We agree to have just one object.

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/ui/PreferredAuthMethod.kt
```diff
@@ -23,13 +23,24 @@
...
+    NONE(0, null),
+
     /**
      * QR code + PIN authentication.
      */
-    QR(18)
+    QR(18, "qrpin");
+    companion object {
+        @JvmStatic
+        fun fromCode(code: Int) = PreferredAuthMethod.values().first { it.code == code }
```
rpdome, line 46: If a new one is introduced in the future, 
(let's say SOME_NEW_METHOD(30,"newmethod"))

the change is shipped via broker, but the MSAL side is still not updated.

Casting here would fail given that code "30" doesn't exist. How would we gracefully handle such scenario?

rpdome: I'm wondering if it would make sense to return a prioritized list instead of just one value, and the SDK side can pick the highest priority one that it supports.

(disclaimer: i don't have insight knowledge on how "PreferredAuthMethod" really works)

p3dr0rv: I think it makes sense to return (1) a list, or at least (2) return an exception indicating the preferred auth method is not supported on this version and that an update is required.

For point 1 I will need to discuss this with IOS and one Auth 
Point 2 nevertheless I can add it.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1281,9 +1281,9 @@ public static String computeMaxHostBrokerProtocol() {
         public static final String BROKER_GENERATE_SHR_RESULT = "broker_generate_shr_result";
 
         /**
-         * String for QR + PIN result.
+         * String for the preferred auth method code
```
shahzaibj:
```suggestion
         * String for the preferred auth method code.
```

common4j/src/main/com/microsoft/identity/common/java/ui/PreferredAuthMethod.kt
p3dr0rv: for 1 we decide to keep it as it is for this PP
for 2 I ipdate code to send an exception indicating an update is required.

p3dr0rv:
Tested with 
Authenticator and MSAL test app in this pipeline
Artifacts: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1259288&view=results
test plan: https://identitydivision.visualstudio.com/Engineering/_testPlans/define?planId=2827166&opId=1986&suiteId=2827225

iamgusain:
> After a discussion with the IOs team and OneAuth team we decided to change:
> 
> 1. The API to return a more general PreferredAuthMethod enum rather than QRPINAvailability. This is to attempt to be extensible in case some other similar method is created as part of current discussions, such as NFC.
> 2. The admin config will reflect this - we'd ask for "preferred_auth_config= QRPIN" rather than "qrPinAvailability = true".

should it return a list of allowed AuthMethod in some order of preference determined by Admin?

changelog.txt
```diff
@@ -1,5 +1,7 @@
 V.Next
 ---------
+- [MINOR] Rename "is QR + PIN available" API to "get preferred auth method". (#2288)
+- [PATCH] Handle Receiver Callback Exception in CommandDispatcher interactive command (#2315)
```
iamgusain: Is this change also part of the PR? (2315), or just a missed changelog entry.
If it is maybe we can create two separate PRs.

common4j/src/main/com/microsoft/identity/common/java/ui/PreferredAuthMethod.kt
```diff
@@ -22,14 +22,29 @@
...
+    /**
+     * No preferred authentication method.
+     */
+    NONE(0, null),
+
     /**
      * QR code + PIN authentication.
      */
-    QR(18)
+    QR(18, "qrpin");
```
iamgusain, line 42: Let's add comment where the magic number 18 comes from.

iamgusain (approved):

changelog.txt
p3dr0rv: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2288/commits/4d1c46ae147bf561b8d642aeaa7f9e485931f139

common4j/src/main/com/microsoft/identity/common/java/ui/PreferredAuthMethod.kt
p3dr0rv: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2288/commits/4d1c46ae147bf561b8d642aeaa7f9e485931f139

PR #2290, opened 2024-01-19
Redirect to hub app in case of NAA flow

somalaya:
**Issue** : Found a bug while ACW team was testing NAA flow. In case of an interrupt flow, the redirectURI of calling apps would be passed to eSTS as the redirectURI. In NAA flow, redirectURI parameter is child app's redirectURI (which could just be a single page app's redirectURI). eSTS always returns the auth code back to hub app's redirectURI.  We would not be able to read/extract the auth code received because the redirectURI eSTS sends code to and the redirectURI we have passed to WebviewAuthorizationStrategy are different. 

**Fix** : Whenever it is a nested app flow (brkRedirectURI!=null), I am passing hub app's redirectURI to authorization strategy to pass the comparison that is made at this line https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/4900edb7f9db5952474586f83efd37b90995cde8/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java#L182

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
@@ -126,6 +126,7 @@ public abstract class BrowserAuthorizationStrategy<
     // Suppressing unchecked warnings during casting to HashMap<String,String> due to no generic type with mAuthorizationRequest
     @SuppressWarnings(WarningType.unchecked_warning)
     private Intent buildAuthorizationActivityStartIntent(Intent authIntent, URI requestUrl) {
+         // RedirectURI used to get the auth code in nested app auth is that of a hub app (brkRedirectURI)   
         final String redirectUri = mAuthorizationRequest.getBrkRedirectUri() != null ? mAuthorizationRequest.getBrkRedirectUri() : mAuthorizationRequest.getRedirectUri();
         final Intent intent = AuthorizationActivityFactory.getAuthorizationActivityIntent(
                 getApplicationContext(),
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -126,11 +126,12 @@ public Future<AuthorizationResult> requestAuthorization(
     // Suppressing unchecked warnings during casting to HashMap<String,String> due to no generic type with mAuthorizationRequest
     @SuppressWarnings(WarningType.unchecked_warning)
     private Intent buildAuthorizationActivityStartIntent(Intent authIntent, URI requestUrl) {
+        final String redirectUri = mAuthorizationRequest.getBrkRedirectUri() != null ? mAuthorizationRequest.getBrkRedirectUri() : mAuthorizationRequest.getRedirectUri();
```
mohitc1, line 130: >final String redirectUri = mAuthorizationRequest.getBrkRedirectUri() != null ? mAuthorizationRequest.getBrkRedirectUri() : mAuthorizationRequest.getRedirectUri(); [](http://example.com/codeflow?start=8&length=162)

Could you add a comment explaining this?

somalaya: done

mohitc1 (approved):
Consider adding UTs

somalaya:
> Consider adding UTs

Yes, I am working on adding the UT in broker PR. It would not be part of this PR

p3dr0rv (approved):

PR #2294, opened 2024-01-23
Add AccountManagerBackupIpcStrategyTargetingSpecificBrokerApp

rpdome:
First part of https://microsoft-my.sharepoint-df.com/:w:/p/rapong/EUJd76fanC9Hgpic0UbLjvUBsFRv6jKt_uTz888YNUj07w?e=cJHbNz

Implement an IPC strategy which utilizes AccountManager.

added tests, and also validated e2e (with my prototype)
- with broker discovery
- with passthrough scenario 
- with passthrough scenario via Legacy WPJ API (chained AccountManager requests)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1532,8 +1532,6 @@ public final class AuthenticationConstants {
          * place, so that we only need to make updates in one location, and it's clearer what we need
          * to do when adding new APIs.
          */
-        @Getter
-        @Accessors(prefix = "m")
         @AllArgsConstructor
         public enum API {
             MSAL_HELLO(MSAL_HELLO_PATH, null, VERSION_3),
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1568,15 +1566,27 @@ public enum API {
...
             /**
              * The broker-host-to-broker protocol version that the API requires.
              */
-            private String mBrokerVersion;
+            private final String mBrokerVersion;
             /**
              * The msal-to-broker version that the API requires.
              */
-            private String mMsalVersion;
+            private final String mMsalVersion;
```
rpdome, line 1577: had to break it apart because lombok doesn't play nicely with kotlin.

common/src/main/java/com/microsoft/identity/common/internal/util/AccountManagerUtil.java
```diff
@@ -46,7 +48,8 @@ private AccountManagerUtil() {
     /**
      * To verify if the caller can use to AccountManager to use broker.
      */
-    public static boolean canUseAccountManagerOperation(final Context context) {
+    public static boolean canUseAccountManagerOperation(final Context context,
```
p3dr0rv, line 52: Update KDoc, 
maybe change this to set instead of list?

p3dr0rv (approved):

fadidurah (approved):

PR #2303, opened 2024-01-29
Adding base64Url encoding method

melissaahn:
### Summary
The challenge field of the `PublicKeyCredentialRequestOptionsJSON` class is meant to have a value that is base64Url encoded (see https://w3c.github.io/webauthn/#dictdef-publickeycredentialrequestoptionsjson). This PR adds a method to encode a String to be base64Url encoded, adds unit tests, and updates associated existing unit tests.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
@@ -88,7 +88,6 @@ class WebAuthnJsonUtil {
             val data: ByteArray = this.toByteArray(Charsets.UTF_8)
             val base64: String = Base64.encodeToString(data, (Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING))
             return base64.replace("=", "")
-            //return base64.replace("=", "").replace("+", "-").replace("/", "_")
         }
     }
 }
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -78,5 +79,15 @@ class WebAuthnJsonUtil {
...
         }
+
+        /**
+         * Returns a base64URL encoding of the string.
+         * @return String
+         */
+        fun String.base64UrlEncoded(): String {
+            val data: ByteArray = this.toByteArray(Charsets.UTF_8)
+            val base64: String = Base64.encodeToString(data, (Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING))
+            return base64.replace("=", "")
```
mohitc1: >base64.replace("=", "") [](http://example.com/codeflow?start=19&length=23)

Do you need this? Doesn't NO_PADDING take care of it?

melissaahn: Yep, thanks for the catch. Addressed here: 9c29abbc6ca0a117eb353731e292e170fe9c7db4

mohitc1 (approved):
:shipit:

fadidurah (approved):

PR #2305, opened 2024-01-31
Handle Receiver Callback Exception in CommandDispatcher interactive command

fadidurah:
Related Bug https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2502607

Related PR https://github.com/AzureAD/ad-accounts-for-android/pull/2690

Teams is seeing a crash during `LocalMSALController.onFinishAuthorizationSession()`, where we occasionally get a `NullPointerException` on `mAuthorizationStrategy.completeAuthorization()`. It's unclear why this happens, but the root issue that such an exception should be routed as a BaseException and sent back as a command result. This PR handles this by routing the exception through the command result if it happens during the callback

Also add a check to see if the authorization future is initialized and waiting, provide an exception for it if we run into the NPE.

E2E testing
I tested this with MSALTestApp by throwing a NPE explicitly in complereAuthorization() and caught the exception, routing through to a command result.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
@@ -94,7 +94,7 @@ public class LocalMSALController extends BaseController {
     private IAuthorizationStrategy mAuthorizationStrategy = null;
 
     @SuppressWarnings({WarningType.rawtype_warning, WarningType.unchecked_warning})
-    ResultFuture<AuthorizationResult> mAuthorizationResultFuture = null;
+    private ResultFuture<AuthorizationResult> mAuthorizationResultFuture = null;
 
     @SuppressWarnings(WarningType.rawtype_warning)
     private AuthorizationRequest mAuthorizationRequest = null;
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
@@ -261,8 +261,6 @@ public class LocalMSALController extends BaseController {
                         .put(TelemetryEventStrings.Key.REQUEST_CODE, String.valueOf(requestCode))
         );
 
-
-
         try {
             mAuthorizationStrategy.completeAuthorization(requestCode, RawAuthorizationResult.fromPropertyBag(data));
         } catch (Exception e){
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
@@ -55,6 +55,7 @@ import com.microsoft.identity.common.java.exception.ErrorStrings;
 import com.microsoft.identity.common.java.exception.ServiceException;
 import com.microsoft.identity.common.java.exception.UiRequiredException;
 import com.microsoft.identity.common.java.platform.DevicePoPUtils;
+import com.microsoft.identity.common.java.providers.RawAuthorizationResult;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftAuthorizationErrorResponse;
 import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsAuthorizationErrorResponse;
 import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsAuthorizationRequest;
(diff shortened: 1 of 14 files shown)


shahzaibj:
Can you try reproducing the issue by just hard-code throwing an NPE from that place in the code? I think we need at least some manual validation of this change

fadidurah:
> Can you try reproducing the issue by just hard-code throwing an NPE from that place in the code? I think we need at least some manual validation of this change

I tried this, the application does not crash with my fix. Unfortunately in my case it hangs forever, because the result future that would be completed in completeAuthorization is not getting a result sent to it, so i'm guessing some background thread is hanging forever so nothing actually happens in my application. On the plus side i don't think this scenario would happen in the user scenario because if the strategy is never initialized, then the future would not be either. But i added a check for both anyway.

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -255,7 +261,19 @@ public void onFinishAuthorizationSession(int requestCode,
                         .put(TelemetryEventStrings.Key.REQUEST_CODE, String.valueOf(requestCode))
         );
 
-        mAuthorizationStrategy.completeAuthorization(requestCode, RawAuthorizationResult.fromPropertyBag(data));
+
```
shahzaibj: nit: extra line

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -755,7 +754,11 @@ public void run() {
                             final LocalBroadcaster.IReceiverCallback resultReceiver = new LocalBroadcaster.IReceiverCallback() {
                                 @Override
                                 public void onReceive(@NonNull PropertyBag dataBag) {
-                                    completeInteractive(dataBag);
+                                    try {
+                                        completeInteractive(dataBag);
+                                    } catch (final Exception e) {
+                                        LocalBroadcaster.INSTANCE.registerExceptionDuringCallback(RETURN_AUTHORIZATION_REQUEST_RESULT, e);
```
shahzaibj: I'm actually not sure if we really need LocalBroadcaster to hold this exception for us. Can we just not hold it locally here in CommandDispatcher beginInteractive method and then throw it later?

fadidurah: Removed, went with a final array locally in CommandDispatcher

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -95,6 +96,9 @@ public class LocalMSALController extends BaseController {
     @SuppressWarnings(WarningType.rawtype_warning)
     private IAuthorizationStrategy mAuthorizationStrategy = null;
 
+    @SuppressWarnings({WarningType.rawtype_warning, WarningType.unchecked_warning})
+    ResultFuture<AuthorizationResult> mAuthorizationFuture = null;
```
shahzaibj: probably rename to `mAuthorizationResultFuture`

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -752,10 +751,16 @@ public void run() {
...
 
+                            final BaseException[] receiverException = new BaseException[1];
+
                             final LocalBroadcaster.IReceiverCallback resultReceiver = new LocalBroadcaster.IReceiverCallback() {
                                 @Override
                                 public void onReceive(@NonNull PropertyBag dataBag) {
-                                    completeInteractive(dataBag);
+                                    try {
+                                        completeInteractive(dataBag);
+                                    } catch (Exception e) {
```
shahzaibj:
```suggestion
                                    } catch (final Exception e) {
```

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -772,6 +777,12 @@ public void onReceive(@NonNull PropertyBag dataBag) {
 
                             LocalBroadcaster.INSTANCE.unregisterCallback(RETURN_AUTHORIZATION_REQUEST_RESULT);
 
+                            // If we received an exception during the receiver callback execution,
+                            // we should set that as the command result
+                            if (receiverException[0] != null) {
+                                commandResult = CommandResult.of(CommandResult.ResultStatus.ERROR, receiverException, correlationId);
```
shahzaibj: I think we should pass in `receiverException[0]` to this method

```suggestion
                                commandResult = CommandResult.of(CommandResult.ResultStatus.ERROR, receiverException[0], correlationId);
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/LocalMSALController.java
```diff
@@ -95,6 +93,9 @@ public class LocalMSALController extends BaseController {
     @SuppressWarnings(WarningType.rawtype_warning)
     private IAuthorizationStrategy mAuthorizationStrategy = null;
 
+    @SuppressWarnings({WarningType.rawtype_warning, WarningType.unchecked_warning})
+    ResultFuture<AuthorizationResult> mAuthorizationResultFuture = null;
```
shahzaibj: private?

shahzaibj (approved):

melissaahn (approved):

PR #2309, opened 2024-02-06
Revert changes made to method signature for futures

fadidurah:
Revert changes made to method signature to return back a ResultFuture instead of a Future. It seems to have introduced a breaking change in MSAL, instead of debugging this during release time, I rolled back part of the solution for the original google crash. In this PR i'm rolling back the whole fix for better tracking, will have a following PR to reintroduce the smaller fix into common to address the issue without adjusting method signatures that could affect msal or broker.

Original PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2305

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,5 @@
 V.Next
 ---------
-- [PATCH] Handle Receiver Callback Exception in CommandDispatcher interactive command (#2305)
 - [MINOR] Add IpcStrategyWithBackup (#2301)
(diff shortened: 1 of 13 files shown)


rpdome:
I would suggest having 2 separate PRs - one with a full revert, and another with the change you want to keep.

This will help us to easier keep track of the change (will be helpful if something goes wrong down the line)

rpdome (approved):

somalaya (approved):

PR #2311, opened 2024-02-07
Add flight to control silent token timeout

iamgusain:
Adding a flight to control the timeout duration for Silent Token Calls.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -28,7 +28,6 @@ import static com.microsoft.identity.common.java.AuthenticationConstants.LocalBr
 import static com.microsoft.identity.common.java.AuthenticationConstants.LocalBroadcasterFields.RESULT_CODE;
 import static com.microsoft.identity.common.java.AuthenticationConstants.SdkPlatformFields.PRODUCT;
 import static com.microsoft.identity.common.java.AuthenticationConstants.SdkPlatformFields.VERSION;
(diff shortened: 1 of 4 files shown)


shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -251,7 +252,8 @@ public static ILocalAuthenticationResult submitAcquireTokenSilentSync(@NonNull f
             if (BuildConfig.DISABLE_ACQUIRE_TOKEN_SILENT_TIMEOUT){
                 commandResult = submitSilentReturningFuture(command).get();
             } else {
-                commandResult = submitSilentReturningFuture(command).get(ACQUIRE_TOKEN_SILENT_DEFAULT_TIMEOUT_MILLISECONDS, TimeUnit.MILLISECONDS);
+                final int silentTokenTimeOutMs = CommonFlightManager.getIntValue(CommonFlight.ACQUIRE_TOKEN_SILENT_TIMEOUT_MILLISECONDS);
```
somalaya, line 255: How are we planning to target teams devices for this flight? Is there any specific version of teams app that we could use?

iamgusain: We have a Manufacturer Filter which we can use for request from TeamsDevices.

somalaya: Do we have a list of all manufacturers that are facing this issue? I think Teams devices folks have a big list of all manufacturers.

somalaya (approved):

PR #2318, opened 2024-02-13
Gracefully handles unexpected error in encryption layer

rpdome:
## Why?

We're getting unexpected error from KeyStore error from 14k Office app users. (~0.02%).
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2835735

The stack trace implies that this is a OS-level issue.  Somehow keystore is not working properly if it's invoked when the app is being booted up.

We've filed a fix to Google, but it hasn't been picked up yet.

On our side, we should be handle these errors gracefully, not crashing the hosting apps.

## Changes
1. Capture all unexpected throwable as ClientException with UNKNOWN_CRYPTO_ERROR. 
ClientException is not a RuntimeException and will be handled by the caller.

2. Catch the unexpected exception in AndroidKeyStoreUtil, as seen in the stack trace.

3. In decrypt(), only wraps error when there is more than one exception. (multiple key is used to decrypt, which is not the case for ADAL/MSAL/OneAuth)

I also removed isEncryptedByThisKeyIdentifier() because it's redundant with getKeyIdentifierFromCipherText(). 
This method is being consumed by broker, therefore it's a breaking change. 
Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2710

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -153,10 +153,15 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
                 return;
             }
 
-            // NOTE: Maybe the right behavior is that we shouldn't write NULL whenever encryption fails.
-            // However, it's possible that we're relying on this behavior in PROD (due to KeyStore failure)
-            // Not persisting the key in memory means getAll() or getAllFilteredByKey() will not work.
-            // Those operations load keys from the storage layer, and invoke getString() - which could return the in-memory cached value.
+            // What this does is that if the encryption fails, we would still write "null" to the storage.
+            // This might not be the right behavior, but changing this could break stuff.
+            // e.g.
+            //      1. In putString(), we would store data in the in-memory cache first, then try encrypting data.
+            //      2. Assuming the encryption fails, this will persist the key.
+            //      3. the getAll() and getAllFilteredByKey() below relies on the key "in the storage".
+            //         If we don't persist the key to the storage, getAll() will not have any key to pull data from in-memory storage.
+            //
+            // Therefore. i'm leaving this untouched.
             String encryptedValue = null;
             try {
                 encryptedValue = mEncryptionManager.encrypt(value);
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -176,7 +176,6 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
             final String storedValue = mSharedPreferences.getString(key, null);
             if (StringUtil.isNullOrEmpty(storedValue)) {
                 Logger.info(methodTag, "Data associated to the given key is null or empty", null);
-                remove(key);
                 return null;
             }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -137,44 +137,59 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
     public final void putString(
             final String key,
             final String value) {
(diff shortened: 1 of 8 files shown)


somalaya (approved):
LGTM. I would suggest we should add some telemetry to capture these errors on our side as well. Currently, we are having to rely on the data provided by Office team to push Android folks to fix the issue I have created on them.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -137,44 +137,64 @@ public final String getSharedPreferencesFileName() {
...
-                final String encryptedValue = encrypt(value);
-                editor.putString(key, encryptedValue);
+                editor.putString(key, value).apply();
+                return;
             }
 
-            editor.apply();
+            // NOTE: Maybe the right behavior is that we shouldn't write NULL whenever encryption fails.
+            // However, it's possible that we're relying on this behavior in PROD (due to KeyStore failure)
+            // Not persisting the key in memory means getAll() or getAllFilteredByKey() will not work.
```
somalaya: Just want to double check if I understand correctly, getAll() would not work correctly (with your proposed change) because, if we do not override it with null, the old (non null) value would be persisted and it would end up returning the old value.. whereas the encryption failed and it should not have returned the old value?

rpdome: GetAll() relies on the key in storage.

Let's say if i have an empty storage. I try to write something, but encryption fails. 
The current logic would have value in the in-memory cache, and persist (key, null) to the storage.

If i call getAll(), i would still be able to get the key (from storage) and use that to retrieve value (in-memory).


If I don't persist (key, null), then there would be no persisted key to even load the value from in-memory cache.

rpdome:
> LGTM. I would suggest we should add some telemetry to capture these errors on our side as well. Currently, we are having to rely on the data provided by Office team to push Android folks to fix the issue I have created on them.

Will do that in a separate PR :)
(want shahzaib's full attention so that I don't flood the telemetry quota with those)

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
rpdome: I'll reword the comment to make it clearer

shahzaibj: To clarify, the word `key` here means a `name` in the map? (rather than an encryption key)

common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
```diff
@@ -189,44 +193,43 @@ public byte[] encrypt(final byte[] plaintext)
...
-            Logger.verbose(TAG + methodName,
+            Logger.verbose(methodTag,
                     "Failed to strip encode version from cipherText, string might not be encrypted. Exception: ", e.getMessage());
             return cipherText;
         }
 
         final List<AbstractSecretKeyLoader> keysForDecryption = getKeyLoaderForDecryption(cipherText);
         if (keysForDecryption.size() == 0) {
-            throw new IllegalStateException("KeyLoader list must not be null or empty.");
+            // Developer error. Throw.
```
shahzaibj, line 217: How is this surfaced back to the developer? I see IllegalStateException being thrown but I believe this is runtime...are we converting to something else down the line?

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
rpdome: yeah

common4j/src/main/com/microsoft/identity/common/java/crypto/StorageEncryptionManager.java
rpdome: In broker, yes. 
In MSAL/OneAuth, It would just crash the app.

The idea here is that this error should be noticeable by us before it's even released to prod, so we want to make it very loud here.

(unlike other error that could be caused by bugs in OS. this is 100% dev error. either we encrypt data with a wrong key, or we didn't provide expected decryption key.)

shahzaibj (approved):

PR #2321, opened 2024-02-16
Return status code in errorResponse when server response is not in expected Json format.

iamgusain:
### What
Return status code in errorResponse when server response is not in expected Json format.

### Why
If the response is not in expected json format, today we don't catch JsonParsing exception and it bubbles up to the caller.
So the caller get the JsonException instead of the actual server error response.

### How
Catching the JSON parsing exception and returning the TokenErrorResponse with error and description set to the statuscode

### Testing
Tested with an error scenario, where the response was not in json format.


somalaya (approved):

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -584,10 +585,17 @@ protected TokenResult getTokenResultFromHttpResponse(@NonNull final HttpResponse
...
-            );
+            try {
+                tokenErrorResponse = ObjectMapper.deserializeJsonStringToObject(
+                        getBodyFromUnsuccessfulResponse(response.getBody()),
+                        MicrosoftTokenErrorResponse.class
+                );
+            } catch (final JsonParseException ex) {
+                tokenErrorResponse = new MicrosoftTokenErrorResponse();
+                final String statusCode = String.valueOf(response.getStatusCode());
+                tokenErrorResponse.setError(statusCode);
```
shahzaibj, line 619: Looks like this is set same as status code on line 597. Do we have a more specific error code (rather than a status code)?

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
iamgusain: Not sure if there is anything more specific. open to suggestion.

PR #2331, opened 2024-02-26
Clear active broker cache if the (cached) active broker app doesn't support ipc mechanism

rpdome:
## Scenario
1. Install LTW broker with feature flag enabled.
2. Launch MSAL Test App, trigger broker discovery (now it thinks that LTW is the active broker)
3. Uninstall LTW, and reinstall LTW without the broker code. 

In this scenario, MSALTestApp's active broker cache should be wiped.

## How?
I moved `isSupportedByTargetedBroker()` from  `AbstractIpcStrategyWithServiceValidation` to `IIpcStrategy`

In `BrokerDiscoveryClient`, invoke that method as we read value from cache.

This is a breaking change (to broker) because IIpcStrategy is being used in multiple places - especially in tests.

## Validation
1. Added a unit test
2. Manual test with the following step.
     1. Install MockCP
     2. Go to Broker API page, force set Prod LTW as the active broker
     3. Install Prod LTW
     4. Trigger Broker API's getAccounts(). No error should be returned.

Related PR:
MSAL (update tests) https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2047


common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IIpcStrategy.kt
```diff
@@ -0,0 +1,42 @@
+package com.microsoft.identity.common.internal.broker.ipc
```
fadidurah, line 23: License

fadidurah (approved):

melissaahn (approved):

melissaahn (approved):

PR #2334, opened 2024-02-29
Minor updates to passkey response logic

melissaahn:
### Summary
After talking with the server team, we figured out that the assertion value they were expecting in our response is based on a custom JSON object. When ESTS UX gets a WebAuthN standard response, they condense it to 5 fields and no inner objects. While I was under the impression that a WebAuthN standard object was expected, the inner object of the standard object was similar enough that ESTS accepted it, so it still functioned. But this PR includes changes to purposely construct the custom object, which makes sure that the required fields are present and that any extra fields are cut out. ESTS said that they will document the custom object on their side, and I'll link that in my own documentation.

This PR also makes small changes to the error/exception strings we send; the exception name and a prefix (for ESTS to more easily identify an error vs an actual assertion) are now added to the string.


common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
```diff
@@ -190,9 +190,10 @@ class AuthFidoChallengeHandler (
         if (exception != null) {
             span.recordException(exception)
             span.setStatus(StatusCode.ERROR)
+            respondToChallenge(submitUrl, exception.toString(), context, span)
```
rpdome: nit: exception.toString() -> e.getMessage()?

Not sure if the first one prints out a hash code. might want to double check.

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -67,17 +71,25 @@ class WebAuthnJsonUtil {
...
                 .getJSONObject(FidoConstants.WEBAUTHN_AUTHENTICATION_ASSERTION_RESPONSE_JSON_KEY)
-            // Making sure that Id is here because ESTS expects it.
-            // I've noticed that GPM will sometimes not include the id in the response object.
-            if (!authResponseJsonObject.has(WEBAUTHN_RESPONSE_ID_JSON_KEY)) {
-                authResponseJsonObject = authResponseJsonObject.put(WEBAUTHN_RESPONSE_ID_JSON_KEY, fullResponseJsonObject.get(WEBAUTHN_RESPONSE_ID_JSON_KEY))
-            }
-            return authResponseJsonObject.toString()
+            // ESTS expects a custom object with clientDataJSON, authenticatorData, signature, userHandle, and id.
+            val assertionResult = JSONObject();
+            assertionResult.put(WEBAUTHN_RESPONSE_ID_JSON_KEY, fullResponseJsonObject.get(
```
rpdome, line 82: if some of the fields here are null (or empty), what would happen? Would it return null or throw an exception?

do we still send this response to ESTS? or do we send an error response, or do we fail here?

melissaahn: It will throw a JSONException, which will be caught by a catch clause in AuthFidoChallengeHandler, and that exception will be sent back to ESTS via the protocol.

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
melissaahn: e.getMessage() doesn't include the actual name of the exception... so I changed this to exception.javaclass.name() + ": " + exception.getMessage(): 6a7adc04646c80b529cbbcf03721528ce47aef93

fadidurah (approved):

rpdome (approved):

melissaahn:
Putting the "skip-consumers-check" tag, not because it's a breaking change, but because all of the checks are passing except for the MSAL build and test, which is currently blocked for MSAL checks in general due to an issue with one of the lab accounts. 
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1262143&view=results

PR #2338, opened 2024-03-08
Add cache for ContentProviderStrategy.isSupportedByTargetedBroker()

rpdome:
packageManager.queryContentProviders involves an ipc call (we've seen DeadObjectException from it before. This is thrown when the IPC Buffer's is used up)
 
We'll avoid making unnecessary ipc request by caching them.
The content provider support status should remain the same for a given broker version.


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -162,12 +163,13 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
 
     constructor(context: Context,
+                components: IPlatformComponents,
                 cache: IClientActiveBrokerCache): this(
         brokerCandidates = BrokerData.getKnownBrokerApps(),
         getActiveBrokerFromAccountManager = {
             AccountManagerBrokerDiscoveryUtil(context).getActiveBrokerFromAccountManager()
         },
-        ipcStrategy = ContentProviderStrategy(context),
+        ipcStrategy = ContentProviderStrategy(context, components),
```
iamgusain, line 172: What is the component object used for?
Do we need to pass or take dependency on the entire component object?

iamgusain: We can just pass the name value storage object here, instead of the full components object.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoader.kt
```diff
@@ -0,0 +1,102 @@
...
+        getVersionCode = { pkgName -> getVersionCode(context, pkgName) },
+        supportedByContentProvider = { pkgName -> supportedByContentProvider(context, pkgName) }
+    )
+
+    override fun getResult(packageName: String) : Boolean {
+        val methodTag = "$TAG:getResult"
+        try {
+            val key = packageName + ":" + getVersionCode(packageName)
+
+            fileManager.get(key)?.let {
```
iamgusain: Is this doing an actual file read?
there would be some io cost associated with file reads, which may add to the overall perf.

rpdome: File read. 

It's a perf hit either way though, given that the actual operation seems to involve some sort of ipc call.

The pros of this is that we would decrease the load on the shared ipc buffer. (1mb per process, which could add up)

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
rpdome: I can make it pass just the storage supplier.... but IMO it's not that different than passing the entire object. 
(if anything, i wouldn't need to make another breaking change if i need some other stuffs in the components object in the future)

I would prefer having this class loads the storage itself, though. No outer layer should need to know how the data is stored here.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IQueryContentProviderLoader.kt
```diff
@@ -0,0 +1,28 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.broker.ipc
+
+interface IQueryContentProviderLoader {
+
+    fun getResult(packageName: String) : Boolean
```
shahzaibj: javadoc on interface and methods

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoader.kt
shahzaibj: This IPC buffer is shared across all apps on the device?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoader.kt
```diff
@@ -0,0 +1,102 @@
...
+        fileManager = components.storageSupplier.getUnencryptedNameValueStore(
+            SHARED_PREFERENCE_NAME, String::class.java),
+        getVersionCode = { pkgName -> getVersionCode(context, pkgName) },
+        supportedByContentProvider = { pkgName -> supportedByContentProvider(context, pkgName) }
+    )
+
+    override fun getResult(packageName: String) : Boolean {
+        val methodTag = "$TAG:getResult"
+        try {
+            val key = packageName + ":" + getVersionCode(packageName)
```
shahzaibj: Can you put some comments here to describe what's we are doing. It looks like we are using the concatenation of package name and package version to create a cache key for caching whether that given package/version pair supports content provider or not, but I think comments should make it more clear. Overall, I think this class can benefit from some comments. 

Also not a huge fan of `getResult` method name. And I'm also wondering if version code should also become a param to this method. So the caller has to supply both package name and version code to avoid any assumptions about what the method does internally. Alternatively sufficient javadoc can also help avoid these confusions.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IQueryContentProviderLoader.kt
```diff
@@ -0,0 +1,28 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.broker.ipc
+
+interface IQueryContentProviderLoader {
```
shahzaibj: may be rename to `IContentProviderStatusLoader`

shahzaibj: At the moment, I can't seem to find any meaning to the word `query` here

common/src/test/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoaderTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            },
+            supportedByContentProvider = { false },
+            fileManager = cache
+        )
+
+        Assert.assertFalse(loader.getResult("com.microsoft.test"))
+        Assert.assertFalse(cache.get("com.microsoft.test:$newVersionKey").toBoolean())
+        Assert.assertEquals(2, cache.size())
+    }
+}
```
shahzaibj: nit: new line ending

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IQueryContentProviderLoader.kt
rpdome: It's from the API we're wrapping around. 

context.packageManager.queryContentProviders

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoader.kt
rpdome: Version code is an internal implementation of this class.

This class is a wrapper over context.packageManager.queryContentProviders, and should take only the package name of the targeted app. caching logic (how key is constructed, how data is persisted) is internal.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/QueryContentProviderLoader.kt
rpdome: 1MB per process. (not much official docs on this here. This is based on the [ContentProvider design docs we have)](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/BrokerContract/ContentProviderAndroidBroker.md&_a=preview)

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStatusLoader.kt
```diff
@@ -0,0 +1,104 @@
...
+            // Construct a key for the cache.
+            val key = packageName + ":" + getVersionCode(packageName)
+
+            // Try loading from cache, return if the value is found.
+            fileManager.get(key)?.let {
+                return it.toBoolean()
+            }
+
+            val queriedResult = supportedByContentProvider(packageName)
+            fileManager.put(key, queriedResult.toString())
```
iamgusain, line 97: What's the expiry/TTL on this entry?

rpdome: No expiry (could set one, but the value here shall remain unchanged since Content Provider stuffs are defined in the manifest level)

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IContentProviderStatusLoader.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A wrapper interface around packageManager.queryContentProviders().
+ * */
+interface IContentProviderStatusLoader {
+
+    /**
+     * Determine if the targeted app supports (Broker) Content Provider.
+     **/
+    fun getStatus(packageName: String) : Boolean
```
iamgusain: nit: supportsContentProviderAuthority (packagename, authority)

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStatusLoader.kt
iamgusain: so next time the app upgrades, we should check it again, instead of reading from cache

rpdome: if the app upgrades, the app version should change -> the read should be a cache miss.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IContentProviderStatusLoader.kt
rpdome: Do you prefer me moving the authority logic out of this class?

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStatusLoader.kt
shahzaibj: Once the app upgrades, I think it is worth removing old entries, otherwise this map/list will keep on growing over time with entries that are totally useless.

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IContentProviderStatusLoader.kt
iamgusain: it can be in same place, just the method name can be clearer of what it is returning.
getStatus is not doing that clearly.
Basically, what you have in the comment description, can also reflect from the name as well.

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/ContentProviderStatusLoader.kt
rpdome: @shahzaibj let me do that in a separate pr (applicable to April release onward)

PR #2339, opened 2024-03-08
getIpcStrategies should only be invoked in bg thread

rpdome:
OneAuthSharedFunctions.getIpcStrategies, which invokes ContentProviderStrategy.isSupportedByTargetedBroker, which invokes IPC operation, should be invoked in background thread only.

The issue is that we're invoking it in the constructor today - and that constructor is invoked in main thread.

by making this lazy-load, it would only be invoked when the ipc request is actually happening - inside a command, which sits inside a commandDispatcher.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -142,6 +142,8 @@ public class BrokerMsalController extends BaseController {
     @Nullable
     private final List<IIpcStrategy> ipcStrategies;
 
+    private BrokerOperationExecutor mOperationExecutor;
+
     @VisibleForTesting
     public BrokerMsalController(@NonNull final Context applicationContext,
                                 @NonNull final IPlatformComponents components,
@@ -165,12 +167,15 @@ public class BrokerMsalController extends BaseController {
 
     /** Should only be invoked in Background thread, given that getIpcStrategies could be a long running operation. */
     @WorkerThread
-    private BrokerOperationExecutor getBrokerOperationExecutor(){
-        return new BrokerOperationExecutor(
-                ipcStrategies != null ? ipcStrategies :
-                        OneAuthSharedFunctions.getIpcStrategies(mApplicationContext, mActiveBrokerPackageName),
-                new ActiveBrokerCacheUpdater(mApplicationContext,
-                        ClientActiveBrokerCache.getClientSdkCache(mComponents.getStorageSupplier())));
+    private synchronized BrokerOperationExecutor getBrokerOperationExecutor(){
+        if (mOperationExecutor == null) {
+            mOperationExecutor = new BrokerOperationExecutor(
+                    ipcStrategies != null ? ipcStrategies :
+                            OneAuthSharedFunctions.getIpcStrategies(mApplicationContext, mActiveBrokerPackageName),
+                    new ActiveBrokerCacheUpdater(mApplicationContext,
+                            ClientActiveBrokerCache.getClientSdkCache(mComponents.getStorageSupplier())));
+        }
+        return mOperationExecutor;
     }
 
     @VisibleForTesting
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -133,22 +134,23 @@ public class BrokerMsalController extends BaseController {
...
     private final HelloCache mHelloCache;
     private final IPlatformComponents mComponents;
     private final Context mApplicationContext;
 
+    // Can be set in unit tests, otherwise this will be null.
+    @Nullable
+    private final List<IIpcStrategy> ipcStrategies;
+
+    @VisibleForTesting
     public BrokerMsalController(@NonNull final Context applicationContext,
```
iamgusain, line 148: probably we can create a separate constructor which does not take the strategy, instead of making it nullable.

common/src/main/java/com/microsoft/identity/common/sharedwithoneauth/OneAuthSharedFunctions.kt
```diff
@@ -45,6 +46,7 @@ class OneAuthSharedFunctions {
          * @param activeBrokerPackageName name of the app hosting the broker process to communicate to.
          **/
         @JvmStatic
+        @WorkerThread
```
iamgusain: This is a breaking change for oneauth

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -158,7 +160,17 @@ public BrokerMsalController(@NonNull final Context applicationContext,
...
                 components,
                 activeBrokerPackageName,
-                OneAuthSharedFunctions.getIpcStrategies(applicationContext, activeBrokerPackageName));
+                null);
+    }
+
+    /** Should only be invoked in Background thread, given that getIpcStrategies could be a long running operation. */
+    @WorkerThread
+    private BrokerOperationExecutor getBrokerOperationExecutor(){
+        return new BrokerOperationExecutor(
```
iamgusain: Should we create a new instance everytime, or store and reuse?
current behavior seem to store and reuse it. we should leave it the same ..
Only the initialization can be lazy. but we dont need to create the instacne everytime, right?

iamgusain (approved):
Left a suggestion

common/src/main/java/com/microsoft/identity/common/sharedwithoneauth/OneAuthSharedFunctions.kt
rpdome: I can remove this. Meant to be a reminder only (to make sure we're slowly moving this away from main thread)

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
rpdome: Yeah we can do that

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
rpdome: there's already one (the one that's not marked with VisibleForTesting)

This one (ipcStrategies) is added solely for unit testing purpose.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -158,7 +162,20 @@ public BrokerMsalController(@NonNull final Context applicationContext,
         this(applicationContext,
                 components,
                 activeBrokerPackageName,
-                OneAuthSharedFunctions.getIpcStrategies(applicationContext, activeBrokerPackageName));
+                null);
+    }
+
+    /** Should only be invoked in Background thread, given that getIpcStrategies could be a long running operation. */
+    @WorkerThread
+    private synchronized BrokerOperationExecutor getBrokerOperationExecutor(){
```
shahzaibj, line 171: probably mark method with `NonNull` annotation

shahzaibj (approved):

PR #2356, opened 2024-03-20
Fixing generateSHR command controller logic

somalaya:
Issue : AT PoP generate SHR tests are failing in the daily pipeline

Cause : In this PR, we changed the order of the controller - https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2040. We are first adding BrokerMsalController if broker is installed. After this change, GenerateSHRCommand was first trying to generateSHR with BrokerMsalController and it would succeed. But the result is not returned. It would then try with LocalMsalController and fails because the account is added to broker. This is an existing bug which got surfaced with recent changes.

Fix : Return from GenerateSHRCommand with the result if a controller returns a successful response.


common/src/main/java/com/microsoft/identity/common/internal/commands/GenerateShrCommand.java
```diff
@@ -106,6 +107,15 @@ public GenerateShrResult execute() throws Exception {
                 } else {
                     throw new ClientException(errorCode, errorMessage);
                 }
+            } else {
```
mohitc1, line 109: nit: comment on line 77-81 should be updated.

mohitc1 (approved):

fadidurah (approved):

PR #2363, opened 2024-03-27
Add parameter to show if request is ATv2

fadidurah:
Tiny PR to add a field in BrokerInteractiveTokenCommandParameters to denote ATv2 request.

Also add error string for SLK.

Had to skip cnsumers because of a failing msal test for DCF

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1282759&view=results

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
@@ -77,7 +77,7 @@ public class BrokerInteractiveTokenCommandParameters extends InteractiveTokenCom
     @Expose
     private final String homeTenantId;
 
-    // Parameter representing if this broker request is an ATv2 (Account Transfer) request
+    // Parameter representing if this broker request is an ATv2 request
     private final boolean isATv2Request;
 
     @Override
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
@@ -77,6 +77,9 @@ public class BrokerInteractiveTokenCommandParameters extends InteractiveTokenCom
     @Expose
     private final String homeTenantId;
 
+    // Parameter representing if this broker request is an ATv2 (Account Transfer) request
+    private final boolean isATv2Request;
+
     @Override
     public void validate() throws ArgumentException {
         super.validate();
--- a/common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
@@ -188,7 +188,7 @@ public class ClientException extends BaseException {
     public static final String DUPLICATE_QUERY_PARAMETER = "duplicate_query_parameter";
 
     /**
-     * An Account Transfer V2 request do not contain an Slk.
+     * An Account Transfer V2 request did not contain an Slk.
      */
     public static final String MISSING_SLK = "missing_slk";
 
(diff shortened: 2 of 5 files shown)


rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
```diff
@@ -77,6 +77,9 @@ public class BrokerInteractiveTokenCommandParameters extends InteractiveTokenCom
     @Expose
     private final String homeTenantId;
 
+    // Parameter representing if this broker request is an ATv2 request
```
rpdome: nit: use full name (Account Transfer instead of AT. AT could also mean Access Token)

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
```diff
@@ -77,6 +77,9 @@ public class BrokerInteractiveTokenCommandParameters extends InteractiveTokenCom
...
+    private final boolean isATv2Request;
```
mohitc1, line 81: >isATv2Request [](http://example.com/codeflow?start=26&length=13)

how and where do we populate this?

fadidurah: This will come in BrokerRequest, the change for that is in the API-side PR (On the way)

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
fadidurah: Sounds good

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/exception/ClientException.java
```diff
@@ -187,6 +187,11 @@ public class ClientException extends BaseException {
      */
     public static final String DUPLICATE_QUERY_PARAMETER = "duplicate_query_parameter";
 
+    /**
+     * An Account Transfer V2 request do not contain an Slk.
```
shahzaibj:
```suggestion
     * An Account Transfer V2 request did not contain an Slk.
```

common4j/src/main/com/microsoft/identity/common/java/exception/ErrorStrings.java
```diff
@@ -457,6 +457,9 @@ private ErrorStrings() {
                     "Please make sure to use your organizational account. " +
                     "If that doesn’t help, please return the device to your administrator.";
 
+    public static final String MISSING_SLK_ERROR_MESSAGE =
```
shahzaibj, line 460: If SLK was never passed then I believe it simply means this was not an Account Transfer request and rather was an AcquireTokenInteractive request? (because IIRC, you were using the present of slk to determine that we are in AT flow)

fadidurah: This is no longer the case with the new api path, there will be a new command parameters class for oneauth to use specifically for ATv2

fadidurah: So it's technically possible for them to call to ATv2 in broker without sending an SLK

shahzaibj: > This is no longer the case with the new api path, there will be a new command parameters class for oneauth to use specifically for ATv2

Interesting. So sounds like we ultimately decided to create a new API for this as opposed to using ATI. Can you point me to this new API? I didn't see it in this PR

fadidurah: It's coming in another PR, I did not combine them because the change list would be pretty large

fadidurah: The API PRs are still in draft, i'm waiting to merge these in to then publish the other ones, as the API PRs are dependent on some things i added in these 2 PRs

PR #2367, opened 2024-03-29
Added telemetry for cross cloud and MSA passthrough requests

somalaya:
Capturing telemetry to get the number of cross cloud requests. 
Related broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/2759

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
@@ -192,6 +192,7 @@ public class AzureActiveDirectoryAuthority extends Authority {
         }
 
         if (cloudOfThisAuthority != null && cloudOfAuthorityToCheck != null) {
+            // Depending upon the caller of this method, home_cloud_name may or may not be a PRT's home cloud.
             SpanExtension.current().setAttribute(AttributeName.home_cloud_name.name(), cloudOfAuthorityToCheck.getPreferredCacheHostName());
             SpanExtension.current().setAttribute(AttributeName.requested_cloud_name.name(), cloudOfThisAuthority.getPreferredCacheHostName());
         }
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
@@ -26,6 +26,8 @@ import com.google.gson.annotations.SerializedName;
 import com.microsoft.identity.common.java.WarningType;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.logging.Logger;
+import com.microsoft.identity.common.java.opentelemetry.AttributeName;
+import com.microsoft.identity.common.java.opentelemetry.SpanExtension;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectory;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectoryCloud;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectorySlice;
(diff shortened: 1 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -189,6 +191,11 @@ public synchronized boolean isSameCloudAsAuthority(@NonNull final AzureActiveDir
             return false;
         }
 
+        if (cloudOfThisAuthority != null && cloudOfAuthorityToCheck != null) {
+            SpanExtension.current().setAttribute(AttributeName.home_cloud_name.name(), cloudOfAuthorityToCheck.getPreferredCacheHostName());
```
iamgusain, line 196: This is utility method to compare two authorities and can potentially be used to compare authorities other than Prts home cloud. Which may give wrong information in our telemetry.
Should we add this telemetry at the source where we are calling this method.

iamgusain (approved):
Left a suggestion.

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
somalaya: It is only getting called using PRT's authorities as of now in the code. But yes, I will move it to the calling method to avoid any confusion

somalaya: After looking into this further, I noticed that we have to make a network call to get the getPreferredCacheHostName. I don't want to do this network call outside of this method again just for telemetry purposes.
So, I will let it be in this method only.

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
shahzaibj: I think the method today is also being invoked in BrokerControllerFactory

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
```diff
@@ -189,6 +191,12 @@ public synchronized boolean isSameCloudAsAuthority(@NonNull final AzureActiveDir
             return false;
         }
 
+        if (cloudOfThisAuthority != null && cloudOfAuthorityToCheck != null) {
+            // Depending upon the caller of this method, home_cloud_name may or may not be a PRT's home cloud.
+            SpanExtension.current().setAttribute(AttributeName.home_cloud_name.name(), cloudOfAuthorityToCheck.getPreferredCacheHostName());
+            SpanExtension.current().setAttribute(AttributeName.requested_cloud_name.name(), cloudOfThisAuthority.getPreferredCacheHostName());
```
shahzaibj, line 197: How do we know this is request authority? I don't see this being passed to the method. I would think that request authority should've been an input to this method but there is just one input `authorityToCheck` which it seems you are treating as home. But how do you know the other one is request authority because it doesn't seem to come from caller or parameters?

somalaya: cloudOfAuthorityToCheck is coming from requested parameter authorityToCheck.  
cloudOfThisAuthority is coming from current object. The method is invoked on the current object which is the requestAuthority in https://github.com/AzureAD/ad-accounts-for-android/blob/367491459f7fafa542b73b4f9032ee323a05e5d2/broker4j/src/main/com/microsoft/identity/broker4j/broker/BrokerControllerFactory.java#L248 .

shahzaibj: Ok I see, this is an instance method. Probably worth using `this` keyword when accessing current object

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAuthority.java
somalaya: Yes, the purpose of adding this telemetry is to see if BrokerControllerFactory's methods to get controllers for silent and interactive calls return BrokerLocalController or BrokerSSOController. 
In BrokerControllerFactory, we are invoking this method with a PRT's authority only.

PR #2369, opened 2024-04-02
Return BrokerData from getFromContext if invoked from unit test

rpdome:
Either we do this, or we have to shadow every tests that might touch this code path.

I was thinking that it's okay to add the logic here - given that this is to get the data for the calling broker process itself.  No validation is really done here.

(can't add BuildConfig.DEBUG because the test pipeline will always use the RELEASE build)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
@@ -227,17 +227,20 @@ data class BrokerData(val packageName : String,
          * Returns a [BrokerData] object matching the owner of the [Context].
          **/
         @JvmStatic
-        fun getFromContext(context: Context): BrokerData {
+        fun getFromContext(context: Context): BrokerData? {
             val signingCertificateThumbprint = PackageHelper(context).getSha512SignatureForPackage(context.packageName)
 
-            // If it's in the list of known brokers, return the one in the list since it has a log-friendly "nickName"
+            if (BuildConfig.DEBUG &&
+                context.packageName == debugBrokerHost.packageName &&
+                signingCertificateThumbprint.isNullOrEmpty()) {
+                // If invoked by unit test, signingCertificateThumbprint would be null.
+                return debugBrokerHost
+            }
+
             return allBrokers.firstOrNull {
                 it.packageName == context.packageName &&
                         it.signingCertificateThumbprint == signingCertificateThumbprint
-            } ?: BrokerData(
-                context.packageName,
-                signingCertificateThumbprint
-            )
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


melissaahn (approved):

mohitc1 (approved):

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -227,8 +227,15 @@ data class BrokerData(val packageName : String,
          * Returns a [BrokerData] object matching the owner of the [Context].
          **/
         @JvmStatic
-        fun getFromContext(context: Context): BrokerData?{
+        fun getFromContext(context: Context): BrokerData? {
             val signingCertificateThumbprint = PackageHelper(context).getSha512SignatureForPackage(context.packageName)
+
+            if (context.packageName == debugBrokerHost.packageName &&
```
shahzaibj, line 237: Will a malicious app be able to exploit this? I'm thinking a malicious app that's not published to PlayStore so it can easily emulate packageName for BrokerHost and then find a way to also go with an empty thumbprint (assuming that is allowed for a debug/local builds not published to play store). In that case I guess we would end up treating that as a legit Broker?

rpdome: @shahzaibj This is not a function we use to validate if the broker is legit. That logic is in BrokerValidator.

this is just to get BrokerData for the broker who invokes this method. (we read the package name from the context object)
(no need to self-validate)

shahzaibj: I believe this is during Broker Discovery right? So I guess the question is do we want malicious broker to invoke this?

rpdome: This method is for getting the data of the broker which is processing the request itself.

Not sure what you meant by "malicious broker invoking this". 
Let's say if they have an access to this method (in their apk), they could invoke it, they could get BrokerData of the brokerhost app, but that's it. it's not really going to do anything.

They might as well just invoke BrokerData.debugBrokerHost directly.

No validation is done here.

----

Before the broker is going to receive/send request to other broker, it would use BrokerValidator to do the actual validation.

shahzaibj: Ok I'm not familiar with all usages of `getFromContext` but if we can be certain that this is never used for getting data for a caller then I'm fine with that.

Technically nothing stops us from using this method in other places in the future. Imagine if we use use this in the future to identify calling broker and pass the caller activity then it can return a malicious broker. Probably good to point out that this API shouldn't be used for such purposes in the future.

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
rpdome: added comment.

PR #2370, opened 2024-04-03
Fix mAuthorizationStrategy is null in LocalMsalController

rpdome:
onFinishAuthorizationSession depends on an internal state of the controller, which is set as part of acquireToken()

therefore, we should be using the same controller object throughout the flow (not generating the new one from the factory).


melissaahn (approved):
When you get the chance, maybe add some details in the description.

shahzaibj (approved):

somalaya (approved):

PR #2374, opened 2024-04-08
Clean up native auth common logging

Yuki-YuXin:
### Goal:
- Review our log statements and reduce to only those that are relevant and useful. Especially the logMethodCall should be much reduced. Not every method call is relevant to know about when debugging.
- Double check that the existing logs use string interpolation correctly to avoid leaking sensitive information in the logs.

### Changes summary:
- Every log item has correlation argument. (make sure use getCorrelationId() )
- The [NativeAuthMsalController.kt](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374/files#diff-304daa7418210ef9cd11ee385364a4a6f74256dc627f404e53422ec50f75b282) layer acts as the main log layer. Instead, remove logSessions under OAuth2Configuration, OAuth2Strategy, RequestProvider since the users care more about the outcome of the request than the construction and sending of the request. Apply the rule "only logger unknown error" to NativeAuthMsalController.
- [Log challengeTypesWithDefault rather than passed challengeTypes](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374/commits/be777c5813c6c15518b358a718b863562d92f5b3)
- Complete // TODO add safe logging using string interpolation with curly brackets and check the rest according to MSAL practice under MicrosoftStsTokenResponse.
- Add request and result Logger.verbose in interactors.

### TODO

- Test PII
- infoWithObject(final String tag, final String correlationID, final String message, final ILoggable object) like warnWithObject
- every log has correlationID

### Company PR:
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2071

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
@@ -24,8 +24,6 @@ package com.microsoft.identity.common.nativeauth.internal.commands
 
 import com.microsoft.identity.common.nativeauth.internal.controllers.NativeAuthMsalController
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.AcquireTokenNoFixedScopesCommandParameters
-import com.microsoft.identity.common.java.logging.LogSession
-import com.microsoft.identity.common.java.logging.Logger
 import com.microsoft.identity.common.java.result.AcquireTokenResult
 
 /**
@@ -47,19 +45,8 @@ class AcquireTokenNoFixedScopesCommand(
     }
 
     override fun execute(): AcquireTokenResult {
-        LogSession.logMethodCall(
-            tag = TAG,
-            correlationId = parameters.getCorrelationId(),
-            methodName = "${TAG}.execute"
-        )
-
         val result = controller.acquireTokenSilent(
             parameters
-        )
-
-        Logger.info(
-            TAG,
-            "Returning result: $result"
         )
         return result
     }
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
@@ -24,6 +24,8 @@ package com.microsoft.identity.common.nativeauth.internal.commands
 
 import com.microsoft.identity.common.nativeauth.internal.controllers.NativeAuthMsalController
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.AcquireTokenNoFixedScopesCommandParameters
+import com.microsoft.identity.common.java.logging.LogSession
+import com.microsoft.identity.common.java.logging.Logger
 import com.microsoft.identity.common.java.result.AcquireTokenResult
 
 /**
@@ -45,8 +47,19 @@ class AcquireTokenNoFixedScopesCommand(
     }
 
     override fun execute(): AcquireTokenResult {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = parameters.getCorrelationId(),
+            methodName = "${TAG}.execute"
+        )
+
         val result = controller.acquireTokenSilent(
             parameters
+        )
+
+        Logger.info(
+            TAG,
+            "Returning result: $result"
         )
         return result
     }
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
@@ -59,6 +59,7 @@ class AcquireTokenNoFixedScopesCommand(
 
         Logger.info(
             TAG,
+            parameters.getCorrelationId(),
             "Returning result: $result"
         )
         return result
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
@@ -515,7 +515,7 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         } catch (e: Exception) {
             Logger.error(
                 TAG,
-                parameters.correlationId,
+                parameters.getCorrelationId(),
                 "Exception thrown in signUpStart",
                 e
             )
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
@@ -515,7 +515,7 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         } catch (e: Exception) {
             Logger.error(
                 TAG,
-                parameters.getCorrelationId(),
+                parameters.correlationId,
                 "Exception thrown in signUpStart",
                 e
             )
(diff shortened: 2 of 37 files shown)


common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
```diff
@@ -47,20 +45,9 @@ class AcquireTokenNoFixedScopesCommand(
...
-            tag = TAG,
-            correlationId = parameters.getCorrelationId(),
-            methodName = "${TAG}.execute"
-        )
-
         val result = controller.acquireTokenSilent(
             parameters
         )
-
-        Logger.info(
```
SammyO, line 60: Is the result logged in MSAL's SDK interface (when receiving the result) or in the Controller (when returning the result)? In my opinion this is valuable information. it's the summary of all the internal processing the SDK does.

SammyO: If we decide to keep this, we should probably add the correlation ID.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -126,7 +126,7 @@ class NativeAuthMsalController : BaseNativeAuthController() {
     fun signInStart(parameters: SignInStartCommandParameters): SignInStartCommandResult {
         LogSession.logMethodCall(
             tag = TAG,
-            correlationId = parameters.getCorrelationId(),
+            correlationId = parameters.correlationId,
```
SammyO: I don't think this compiles/works. See comment above `public String getCorrelationId()` in `CommandParameters.java`

SammyO: Which raises the question: did you test this? I.e. validate that correlation IDs are correctly logged?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthResponseHandler.kt
```diff
@@ -385,7 +385,12 @@ class NativeAuthResponseHandler {
                 response.body,
                 MicrosoftStsTokenResponse::class.java
             )
-            // TODO add safe logging
+
+            Logger.info(TAG, "MicrosoftStsTokenResponse authority:${apiResponse.authority} " +
```
SammyO: Why are we logging this at all? Does MSAL log it?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/ResetPasswordInteractor.kt
```diff
@@ -62,12 +61,6 @@ class ResetPasswordInteractor(
     fun performResetPasswordStart(
         parameters: ResetPasswordStartCommandParameters
     ): ResetPasswordStartApiResult {
-        LogSession.logMethodCall(
```
SammyO, line 65: I agree to remove the logs from OAuth2Strategy, RequestProvider and RequestHandler, but wouldn't it be useful to have some logging in the API layer? Since in the controller we're only logging method calls (not the paths through the business logic that's taken), having API-layer logging will help us figure out what API calls were made.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/SignUpInteractor.kt
```diff
@@ -60,18 +59,12 @@ class SignUpInteractor(
...
-            methodName = "${TAG}.performSignUpStart"
-        )
-
         val request = nativeAuthRequestProvider.createSignUpStartRequest(
             commandParameters = commandParameters
         )
         try {
             return performSignUpStart(
-                requestCorrelationId = commandParameters.getCorrelationId(),
+                requestCorrelationId = commandParameters.correlationId,
```
SammyO: Same here. Please check throughout the PR.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordContinueApiResponse.kt
```diff
@@ -57,11 +56,6 @@ class ResetPasswordContinueApiResponse(
      * @see com.microsoft.identity.common.java.nativeauth.providers.responses.resetpassword.ResetPasswordContinueApiResult
      */
     fun toResult(): ResetPasswordContinueApiResult {
-        LogSession.logMethodCall(
```
SammyO, line 60: Since we're doing quite some API response validation in these `toResult()` methods, I'm wondering whether it's useful to log the result of that. Perhaps that logging can be done in the Interactor; a log through which we can determine how the SDK has parsed the API response and what ApiResult was generated accordingly.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
Yuki-YuXin: The command layer is ignored in the first version. Put them back and add correlation ID for the Logger.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/ResetPasswordInteractor.kt
Yuki-YuXin: Agree that API-layer need some logging for us to debug. Put logSession back to the interactor and plan to add request and response logging of the API-layer.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordContinueApiResponse.kt
Yuki-YuXin: Put rawResponse -> apiResult logging in interactors.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthResponseHandler.kt
Yuki-YuXin: Here's our previous log and TODO. In my opinion, we don't have to log them. MSAL has authority and (request/granted)  scopes hint but in different place. 

In terms of the TokenResponse of MSAL:
MicrosoftStsOAuth2Strategy:TokenResult:MicrosoftStsTokenResponse INFO {"ext_expires_in":4165,"expires_in":4165,"mResponseReceivedTime":0,"scope":"email openid profile Application.Read.All User.Read","token_type":"Bearer"}

BaseController:TokenResult:MicrosoftStsTokenResponse INFO {"ext_expires_in":4165,"mRefreshTokenAge":"","mSpeRing":"","expires_in":4165,"mResponseReceivedTime":0,"scope":"email openid profile Application.Read.All User.Read","token_type":"Bearer"}

Thus update the contents of the log to include what MSAL displays.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
Yuki-YuXin: It somehow works in my local.
image

Just inc case, have reverted the changes to use getCorrelationId() but haven't check the consistent of correlation IDs of the flow.

Yuki-YuXin: Has validated that correlation IDs are correctly logged.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/SignInInteractor.kt
```diff
@@ -61,12 +62,20 @@ class SignInInteractor(
...
         )
+
         val request = nativeAuthRequestProvider.createSignInInitiateRequest(
             commandParameters = parameters
         )
+
+        Logger.verbose(
+            "${TAG}.performSignInInitiate",
+            parameters.getCorrelationId(),
+            "performSignInInitiate: request = $request"
```
rmccahill, line 76: Should we be logging the full requests + responses? Could they contain potentially sensitive information like sent passwords, codes, etc?

SammyO: Excellent point @rmccahill, and coincidentally: https://identitydivision.visualstudio.com/Engineering/_boards/board/t/DevExDub%20-B2C/Backlog%20items/?workitem=2900824

@Yuki-YuXin let's hold off with this PR for now. I need to work together with you to ensure we're being even more careful with logging.

PR #2375, opened 2024-04-09
Account Transfer Skips Account Chooser

fadidurah:
In this PR, i'm implementing the skipping of account chooser during account transfer requests.

changes:
- Pass account transfer token explicitly in InteractiveTokenCommandParameters
- Pass the transfer token as a separate field as part of the broker request
- Add a boolean to the broker request bundle to denote account transfer request
- Add boolean to BrokerInteractiveTokenCommandParameters to denote account transfer request


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1599,6 +1600,12 @@ public String getMsalVersion(){
          */
         public static final String MSAL_ACQUIRE_TOKEN_INTERACTIVE_PATH = "/acquireTokenInteractive";
 
+        /**
+         * URI Path constant for MSAL-to-Broker acquireTokenInteractive (for ATv2) request using ContentProvider.
+         *  TODO: Not really sure how this works, this seems right
```
fadidurah: Ping

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -415,6 +421,109 @@ public void onReceive(@NonNull PropertyBag propertyBag) {
...
+        mApplicationContext.startActivity(brokerActivityIntent);
+
+        final AcquireTokenResult result;
+        try {
+            // Wait to be notified of the result being returned... we could add a timeout here if we want to
+            final Bundle resultBundle = mBrokerResultFuture.get();
+
+            final String negotiatedBrokerProtocolVersion = interactiveATv2RequestIntent.getStringExtra(NEGOTIATED_BP_VERSION_KEY);
+
+            // TODO: Is this accurate? BrokerAccountDataManager contains MSA accounts
```
fadidurah: Ping

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -484,6 +593,75 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                    BrokerOperationBundle getBundle() {
+                        return new BrokerOperationBundle(
+                                MSAL_GET_INTENT_FOR_ATV2_INTERACTIVE_REQUEST,
+                                mActiveBrokerPackageName,
+                                null);
+                    }
+
+                    // TODO: Confused, why is this called "extractResultBundle", but we're actually
+                    //  returning an intent where we create the bundle for the request?
+                    //  Is this the result of performPrerequisites with broker?
```
fadidurah: Ping

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1536,6 +1536,7 @@ public static final class BrokerContentProvider {
         public enum API {
             MSAL_HELLO(MSAL_HELLO_PATH, null, VERSION_3),
             ACQUIRE_TOKEN_INTERACTIVE(MSAL_ACQUIRE_TOKEN_INTERACTIVE_PATH, null, VERSION_3),
+            ACQUIRE_TOKEN_ATV2_INTERACTIVE(MSAL_ACQUIRE_TOKEN_ATV2_INTERACTIVE_PATH, null, VERSION_3),
```
mohitc1: >VERSION_3 [](http://example.com/codeflow?start=91&length=9)

Do we need to update this?

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
```diff
@@ -38,6 +38,8 @@ interface IMicrosoftAuthService {
 
     Intent getIntentForInteractiveRequest();
 
+    Intent getIntentForATv2InteractiveRequest();
```
mohitc1: >ATv2 [](http://example.com/codeflow?start=23&length=4)

AccountTransferV2 instead of ATv2?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/ATv2TokenCommandParameters.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Command parameters class for ATv2 requests. Extends {@link InteractiveTokenCommandParameters} since ATv2
+ * can use the same set of parameters as a regular acquire token request. No additional parameters for ATv2 are
+ * required at this time, however.
+ */
+@Getter
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder(toBuilder = true)
+public class ATv2TokenCommandParameters extends InteractiveTokenCommandParameters {
```
mohitc1: Why not we define slk as a required parameter here? You'd get parameter validation while request creation itself and forces a technical constraint on Atv2 parameters

fadidurah: SLK is already being passed as an extra query parameter in the base token command parameters. I'd rather not extract it into its own field because that would mean we'd need to add another code path in acquirePrt to manually pass in slk in ATv2 scenarios to the authorize endpoint. This functionality is already present if SLK is in the extra query parameters field, I don't think we should ask oneauth to add it in 2 places within the same parameters object.

We could do it in a way where they only pass it in as a separate field in ATv2TokenCommandParameters, and we do some check in ATv2Activity to add it into the extra query parameter if it's not already there before sending the request through PrtController, but i feel this is adding more complexity that we probably don't need and is more or less still doing the parameter validation anyway.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: I'm honestly not sure, was hoping someone more familiar with this part of the code would give a definitive answer 😅

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
fadidurah: Sure thing

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
shahzaibj: I'm pretty sure this needs to be updated because older brokers are not going to understand ATv2 API.

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
fadidurah: renamed all references to ATv2 in both PRs

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/ATv2TokenCommandParameters.java
fadidurah: After some design changes, we've updated this to include slk as a direct parameter in this class.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -415,6 +417,109 @@ public void onReceive(@NonNull PropertyBag propertyBag) {
...
+            final String negotiatedBrokerProtocolVersion = interactiveAccountTransferV2RequestIntent.getStringExtra(NEGOTIATED_BP_VERSION_KEY);
+
+            // TODO: Is this accurate? BrokerAccountDataManager contains MSA accounts
+            // For MSA Accounts Broker doesn't save the accounts, instead it just passes the result along,
+            // MSAL needs to save this account locally for future token calls.
+            // parameters.getOAuth2TokenCache() will be non-null only in case of MSAL native
+            // If the request is from MSALCPP , OAuth2TokenCache will be null.
+            if (parameters.getOAuth2TokenCache() != null && !BrokerProtocolVersionUtil.canSupportMsaAccountsInBroker(negotiatedBrokerProtocolVersion)) {
+                saveMsaAccountToCache(resultBundle, (MsalOAuth2TokenCache) parameters.getOAuth2TokenCache());
+            }
```
mohitc1: This should not be needed anymore. This was before PRTv3 when MSA accounts where not supported in broker.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -260,6 +281,23 @@ public Bundle getRequestBundleForAcquireTokenInteractive(@NonNull final Interact
...
+                                                             @Nullable final String negotiatedBrokerProtocolVersion) {
+        final BrokerRequest brokerRequest = brokerRequestFromAccountTransferV2Parameters(parameters);
+        return getRequestBundleFromBrokerRequest(
+                brokerRequest,
+                negotiatedBrokerProtocolVersion,
+                parameters.getRequiredBrokerProtocolVersion()
+        );
+    }
+
     /**
```
mohitc1: Add UTs for these

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: This should not be needed anymore. This was before PRTv3 when MSA accounts where not supported in broker.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: This for extract the result obtained from broker (resultBundle) into object you want. In this case it's Intent object (BrokerOperation<Intent>()  {} above)

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
mohitc1: yes, just that this parameter is not in used today but can be put to use when making call from BrokerMsalController

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: What does "this" refer to, the whole if statement?

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: Is this not already covered by bumping the broker protocol version to 16?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
mohitc1: saveMsaAccountToCache part

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
mohitc1: You may update here as well (don't see any effect of it though)

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -90,12 +91,33 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Inte
...
+
+    private @NonNull BrokerRequest.BrokerRequestBuilder getBrokerRequestBuilderForInteractiveTokenCommandParameters(@NonNull final InteractiveTokenCommandParameters parameters) {
         final String extraQueryStringParameter = parameters.getExtraQueryStringParameters() != null ?
                 QueryParamsAdapter._toJson(parameters.getExtraQueryStringParameters())
                 : null;
         final String extraOptions = parameters.getExtraOptions() != null ?
                 QueryParamsAdapter._toJson(parameters.getExtraOptions()) : null;
-        final BrokerRequest brokerRequest = BrokerRequest.builder()
+
+        final BrokerRequest.BrokerRequestBuilder brokerRequest = BrokerRequest.builder()
```
mohitc1: >brokerRequest [](http://example.com/codeflow?start=49&length=13)

nit: rename to brokerRequestBuilder

mohitc1 (approved):
Approved with suggestions

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -123,6 +124,10 @@ public static String getDelimitedDefaultScopeString() {
     public abstract AcquireTokenResult acquireToken(final InteractiveTokenCommandParameters request)
             throws Exception;
 
+    public AcquireTokenResult acquireTokenForAccountTransferV2(final AccountTransferV2TokenCommandParameters request) throws Exception {
```
shahzaibj: Why not declare it abstract here like we are doing with the rest of the methods?

fadidurah: If i declare it abstract here, i'd need to define it in every class that extends BaseController, and have them all (except BrokerMsalController) throw an unsupported method exception. This seemed cleaner to do it this way, and just override it in BrokerMsalController.

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
```diff
@@ -38,6 +38,8 @@ interface IMicrosoftAuthService {
 
     Intent getIntentForInteractiveRequest();
 
+    Intent getIntentForAccountTransferInteractiveRequest();
```
rpdome: add it at the bottom. 

there might be some weird side effects if the order of aidl on broker and MSAL/OneAuth side mismatches.

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
```diff
@@ -98,6 +98,20 @@ public MicrosoftAuthClient(@NonNull final Context context,
 
                 return bundle;
 
+            case MSAL_GET_INTENT_FOR_ACCOUNT_TRANSFER_INTERACTIVE_REQUEST:
+                final Intent accountTransferIntent = microsoftAuthService.getIntentForAccountTransferInteractiveRequest();
+                final Bundle accountTransferBundle = accountTransferIntent.getExtras();
+
+                //older brokers (pre-ContentProvider) are ONLY sending these values in the intent itself.
```
rpdome: this msg could be confusing - since this operation is just added (it's not pre-ContentProvider)

you actually can just expose 
GetBundleForAccountTransferInteractiveRequestMsalBrokerOperation
on the Broker side, and return the resulted bundle as is.

The code will be cleaner on the Broker side, and you wouldn't need to do extra conversion here.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -484,6 +575,70 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+    private Intent getBrokerAuthorizationIntentForAccountTransfer(
+            final @NonNull AccountTransferTokenCommandParameters parameters) throws BaseException {
+        return getBrokerOperationExecutor().execute(parameters,
+                new BrokerOperation<Intent>() {
+                    private String negotiatedBrokerProtocolVersion;
+
+                    @Override
+                    public void performPrerequisites(final @NonNull IIpcStrategy strategy) throws BaseException {
+                        verifyTokenParametersAreSupported(parameters);
+                        negotiatedBrokerProtocolVersion = hello(strategy, parameters.getRequiredBrokerProtocolVersion());
```
rpdome: if the negotiatedBrokerProtocolVersion is lower than the new version you just added, it means the broker does NOT support this operation.

In that case, you might want to fail the request early. (no need to make the request to the broker because it'll just throw "not supported" back)

common/src/main/aidl/com/microsoft/identity/client/IMicrosoftAuthService.aidl
fadidurah: Done

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: So, in the parameters for ATv2, OneAuth will send in 16.0 protocol version. wouldn't that already take care of this issue? We'd get an error result from the hello operation in that case, right?

rpdome: Short answer: you're right.

Long answer: 
1. OneAuth would send a hello request with 16.0 as the *minimum* protocol version
2. If the broker doesn't support 16.0, the following block will be hit
```Java
        // Broker is older than the minimum version that the client requires.
        Logger.warn(
                methodTag,
                "Connection DISALLOWED"
                        + "\n"
                        + "Broker only supports protocol version up to: " + brokerMaxProtocolVersion
                        + " but client requires: " + clientMinProtocolVersion
        );

        final Bundle resultBundle = new Bundle();
        resultBundle.putString(HELLO_ERROR_CODE, ErrorStrings.UNSUPPORTED_BROKER_VERSION_ERROR_CODE);
        resultBundle.putString(HELLO_ERROR_MESSAGE, ErrorStrings.UNSUPPORTED_BROKER_VERSION_ERROR_MESSAGE);
        return resultBundle;
```
3. BrokerMsalController.hello() will invoke verifyHelloFromResultBundle(), and that method will convert an error into an exception.

common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
```diff
@@ -75,4 +85,42 @@ public void test_getRequestBundleForSsoToken() {
...
+                .redirectUri(aRedirect)
+                .scopes(scopes)
+                .applicationName(appName)
+                .applicationVersion(appVersion)
+                .sdkVersion(appVersion)
+                .sdkType(SdkType.MSAL)
+                .prompt(OpenIdConnectPromptParameter.SELECT_ACCOUNT)
+                .build();
+
+        MsalBrokerRequestAdapter msalBrokerRequestAdapter = new MsalBrokerRequestAdapter();
```
rpdome: nit: final

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1536,6 +1537,7 @@ public static final class BrokerContentProvider {
         public enum API {
             MSAL_HELLO(MSAL_HELLO_PATH, null, VERSION_3),
             ACQUIRE_TOKEN_INTERACTIVE(MSAL_ACQUIRE_TOKEN_INTERACTIVE_PATH, null, VERSION_3),
+            ACQUIRE_TOKEN_ACCOUNT_TRANSFER_INTERACTIVE(MSAL_ACQUIRE_TOKEN_ACCOUNT_TRANSFER_INTERACTIVE_PATH, null, VERSION_8),
```
rpdome: protocol version 14.0?

https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Broker%20API/broker_protocol_versions.md&_a=preview

(not sure if this docs is updated btw)

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -415,6 +414,98 @@ public void onReceive(@NonNull PropertyBag propertyBag) {
...
 
+    /**
+     * Performs Account Transfer request with Broker.
+     *
+     * @param parameters a {@link AccountTransferTokenCommandParameters}
+     * @return an {@link AcquireTokenResult}.
+     */
+    @Override
+    public AcquireTokenResult acquireTokenForAccountTransfer(final @NonNull AccountTransferTokenCommandParameters parameters) throws BaseException, ExecutionException, InterruptedException {
+        final String methodTag = TAG + ":acquireTokenForAccountTransfer";
```
rpdome: is there any way to avoid duplicating the code in this method?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -484,6 +575,70 @@ public void putValueInSuccessEvent(final @NonNull ApiEndEvent event, final @NonN
...
+                        negotiatedBrokerProtocolVersion = hello(strategy, parameters.getRequiredBrokerProtocolVersion());
+                    }
+
+                    @Override
+                    public @NonNull
+                    BrokerOperationBundle getBundle() {
+                        return new BrokerOperationBundle(
+                                MSAL_GET_INTENT_FOR_ACCOUNT_TRANSFER_INTERACTIVE_REQUEST,
+                                mActiveBrokerPackageName,
+                                null);
```
rpdome: alternatively, you can create a bundle with a flag here.

on the broker side, if the flag exists, return the AccountTransfer activity class. 
otherwise, return AccountChooser.

If you do it this way, then the change on msal side would be minimal. you'll get a lot of things for free.

(you can pretty much reuse the existing acquire token interactive code in this class. The only difference would be in this getBundle() method)

rpdome: could you please remind me again what's the reason of opening a new ipc interface/new activity?

rpdome:
Feel free to ping me / set up a meeting to expedite the review process.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: No longer applicable

common/src/main/java/com/microsoft/identity/common/internal/broker/MicrosoftAuthClient.java
fadidurah: No longer applicable

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: No longer applicable

common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapterTests.java
fadidurah: No longer applicable

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: No longer needed

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: No longer needed

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: No longer needed

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -310,10 +310,13 @@ private String tryGetNegotiatedProtocolVersionFromHelloCache(
      * @return an {@link AcquireTokenResult}.
      */
     @Override
-    public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandParameters parameters)
+    public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandParameters oldParameters)
             throws BaseException, InterruptedException, ExecutionException {
         final String methodTag = TAG + ":acquireToken";
 
+        // This is for testing only, will remove once validation is complete. Same parameter renaming
```
rpdome: remove this.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -387,10 +390,7 @@ public void onReceive(@NonNull PropertyBag propertyBag) {
             final Bundle resultBundle = mBrokerResultFuture.get();
 
             final String negotiatedBrokerProtocolVersion = interactiveRequestIntent.getStringExtra(NEGOTIATED_BP_VERSION_KEY);
-            // For MSA Accounts Broker doesn't save the accounts, instead it just passes the result along,
```
rpdome: nit: why removing this?

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -253,4 +253,8 @@ private static final class SerializedNames {
     @Nullable
     @SerializedName(SerializedNames.PREFERRED_AUTH_METHOD)
     private PreferredAuthMethod mPreferredAuthMethod;
+
+    @Nullable
+    @SerializedName(SerializedNames.ACCOUNT_TRANSFER_TOKEN)
+    private String mAccountTransferToken;
```
rpdome, line 259: sync with @mohitc1.

If he's planning to deserialize the BrokerRequest in AccountChooser.shouldDisplayAccountList(), then you might not need this. you can use BrokerInteractiveTokenCommandParameters.isAccountTransferRequest() instead.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: I think this is an old comment, because Broker does save the MSA accounts. This detail is actually important to my implementation because we query broker storage for msa accounts.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
fadidurah: My change was to allow the check without deserializing, Is it possible to do something similar in your implementation @mohitc1 ? Or do we feel we should pivot to deserializing the brokerRequest inside the account chooser.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: @mohitc1 Let me know if i'm understanding this wrong. I can add bnack the comment if it's still relevant.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
fadidurah: I pulled mohit's changes and used the earlier serializing infra

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: I've returned the comment, we can remove it later if necessary.

PR #2377, opened 2024-04-11
Return exception from first controller in SilentTokenCommand in case all controllers throw exceptions.

iamgusain:
Returning exception from first controller in SilentTokenCommand in case all controllers throw exceptions.
This is to ensure that for cases where both Broker and Msal Local controller fail, we return the error from Broker Controller (which is the first controller), so that client tries to handle/recover from the error on broker side.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
@@ -105,26 +105,28 @@ public class SilentTokenCommand extends TokenCommand {
                         span.setStatus(StatusCode.OK);
                         return result;
                     }
-                } catch (UiRequiredException | ClientException e) {
-                    if (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT) // was invalid_grant
-                            && controllers.size() > ii + 1) { // isn't the last controller we can try
-                        exceptionFromFirstController = exceptionFromFirstController == null ? e : exceptionFromFirstController;
-                        continue;
-                    } else if ((e.getErrorCode().equals(ErrorStrings.NO_TOKENS_FOUND)
-                            || e.getErrorCode().equals(ErrorStrings.NO_ACCOUNT_FOUND))
-                            && controllers.size() > ii + 1) {
-                        exceptionFromFirstController = exceptionFromFirstController == null ? e : exceptionFromFirstController;
-                        //if no token or account for this silent call, we should continue to the next silent call.
-                        continue;
+                } catch (final UiRequiredException | ClientException e) {
+                    // if this isn't the last controller and
+                    // error code was invalid_grant or if no token or account for this silent call
+                    // continue with next controller.
+                    if (controllers.size() > ii + 1
+                            && (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT)
+                            || e.getErrorCode().equals(ErrorStrings.NO_TOKENS_FOUND)
+                            || e.getErrorCode().equals(ErrorStrings.NO_ACCOUNT_FOUND))) {
+                        if (ii == 0) {
+                            exceptionFromFirstController = e;
+                        }
                     } else {
                         if (exceptionFromFirstController != null) {
+                            // throw exception from the first controller
                             throw exceptionFromFirstController;
                         } else {
                             throw e;
                         }
                     }
-                } catch (final Exception e) {
+                } catch (final Throwable e) {
                     if (exceptionFromFirstController != null) {
+                        // throw exception from the first controller
                         throw exceptionFromFirstController;
                     } else {
                         throw e;
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
@@ -123,6 +123,12 @@ public class SilentTokenCommand extends TokenCommand {
                             throw e;
                         }
                     }
+                } catch (final Exception e) {
+                    if (exceptionFromFirstController != null) {
+                        throw exceptionFromFirstController;
+                    } else {
+                        throw e;
+                    }
                 }
             }
 
--- a/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
@@ -106,27 +106,19 @@ public class SilentTokenCommand extends TokenCommand {
                         return result;
                     }
                 } catch (final UiRequiredException | ClientException e) {
+                    if (ii == 0 ) {
+                        exceptionFromFirstController = e;
+                    }
                     // if this isn't the last controller and
                     // error code was invalid_grant or if no token or account for this silent call
                     // continue with next controller.
-                    if (controllers.size() > ii + 1
-                            && (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT)
-                            || e.getErrorCode().equals(ErrorStrings.NO_TOKENS_FOUND)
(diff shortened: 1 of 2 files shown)


somalaya (approved):

somalaya:
Is SilentTokenCommand the only place where this fix needs to be made? Are there any other token commands that need this fix?

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -74,6 +72,7 @@ public AcquireTokenResult execute() throws Exception {
         final List<BaseController> controllers = getControllerFactory().getAllControllers();
 
         try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
+            Exception exceptionFromFirstController = null;
```
mohitc1, line 75: >exceptionFromFirstController [](http://example.com/codeflow?start=22&length=28)

If the intent is to prioritize broker's exception for client, then I think it's better to capture that exception e.g. Exception exceptionFromBrokerMsalController;

then below, 
expcliit check on if (controller instanceof BrokerMsalController)


Instead of using controller ordering.

shahzaibj: +1

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -109,14 +108,20 @@ public AcquireTokenResult execute() throws Exception {
                 } catch (UiRequiredException | ClientException e) {
```
shahzaibj: What about other types of Exceptions? I believe you want those to be returned from first controller as well? If yes, then you need a catch block for generic `Exception` and add the same logic there

iamgusain: Those are getting thrown already from first controller.
As we are only trying 2nd controller in cases of specific errors.

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
iamgusain: The idea is to return error from first/primary controller so that client prioritize to handle that.
In case all controllers fail, we want the client to handle/recover from the error from primary/first controller if possible.

iamgusain:
> Is SilentTokenCommand the only place where this fix needs to be made? Are there any other token commands that need this fix?

Checked offline with Sowmya, we dont need to make this specific change in other token commands/places

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
iamgusain: I see what you mean. let me check

iamgusain: done

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -109,12 +108,24 @@ public AcquireTokenResult execute() throws Exception {
...
                         //if no token or account for this silent call, we should continue to the next silent call.
                         continue;
+                    } else {
+                        if (exceptionFromFirstController != null) {
+                            throw exceptionFromFirstController;
+                        } else {
+                            throw e;
+                        }
+                    }
+                } catch (final Exception e) {
```
shahzaibj:
```suggestion
                } catch (final Throwable e) {
```

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -109,12 +108,24 @@ public AcquireTokenResult execute() throws Exception {
                 } catch (UiRequiredException | ClientException e) {
                     if (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT) // was invalid_grant
                             && controllers.size() > ii + 1) { // isn't the last controller we can try
+                        exceptionFromFirstController = exceptionFromFirstController == null ? e : exceptionFromFirstController;
```
shahzaibj: It seems the logic here is such that "_If not already set then this must be first controller exception else this was not first_" but I am thinking that code might be easier to understand if you just do the following:

```java
if (ii == 0) {
   exceptionFromFirstController = e;
}
```

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -105,26 +105,28 @@ public AcquireTokenResult execute() throws Exception {
...
-                    if (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT) // was invalid_grant
-                            && controllers.size() > ii + 1) { // isn't the last controller we can try
-                        exceptionFromFirstController = exceptionFromFirstController == null ? e : exceptionFromFirstController;
-                        continue;
-                    } else if ((e.getErrorCode().equals(ErrorStrings.NO_TOKENS_FOUND)
-                            || e.getErrorCode().equals(ErrorStrings.NO_ACCOUNT_FOUND))
-                            && controllers.size() > ii + 1) {
-                        exceptionFromFirstController = exceptionFromFirstController == null ? e : exceptionFromFirstController;
-                        //if no token or account for this silent call, we should continue to the next silent call.
-                        continue;
```
shahzaibj, line 117: Seems like you removed the `continue` statement but this needs to be put back

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
```diff
@@ -105,26 +105,28 @@ public AcquireTokenResult execute() throws Exception {
...
+                    // if this isn't the last controller and
+                    // error code was invalid_grant or if no token or account for this silent call
+                    // continue with next controller.
+                    if (controllers.size() > ii + 1
+                            && (e.getErrorCode().equals(OAuth2ErrorCode.INVALID_GRANT)
+                            || e.getErrorCode().equals(ErrorStrings.NO_TOKENS_FOUND)
+                            || e.getErrorCode().equals(ErrorStrings.NO_ACCOUNT_FOUND))) {
+                        if (ii == 0) {
+                            exceptionFromFirstController = e;
+                        }
```
shahzaibj: Could consider removing this whole catch block and execute this logic inside the generic catch block for `Exception` below and do an instance check on Exception for `UiRequiredException` and `ClientException` to decide if we need to `continue`

iamgusain: We can, but that would require us to cast the exception again to get the error code.
That would be more complex.

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
iamgusain: Continue was redundant.

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
shahzaibj: Is this fully addressed? I think you need to check for ii == 0 in the generic Exception block but I don't see that happening

shahzaibj: Discussed offline; lgtm

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/SilentTokenCommand.java
iamgusain: yes, this is addressed

PR #2384, opened 2024-04-16
Safe logging

SammyO:
The default implementation of `toString()` of Kotlin's data classes uses all fields to compose the String. This unintentionally includes fields that may not be safe for logging. 

Summary of solution:
- Introduce a new `ILoggable` interface. Recommendation is that all native auth classes implement this interface.
- `Logging` contains 2 methods:
    - `fun toUnsanitizedString(): String` - This method produces a String that may contain PII (PII = Personally identifiable information). The value of `containsPii()` will indicate whether the value actually contains PII.
    - `fun containsPii(): Boolean` - This method indicates whether the implementing class contains data fields that are considered PII (Personally identifiable information). If this method returns true, then the value of `toSafeString(true)` will return a String that contains PII.
 - Extend `Logger` with a new methods `*withObject()` & parameter, e.g.:  `public static void infoWithObject(final String tag, final String message, final Logging object)`. This logs safe object contents, based on the value of Logger's `allowPii` and the object's `containsPii()` value.
   - If Logger's `allowPii` is false, then `object.toString()` is logged, with `containsPii` set to `false`.
   - If Logger's `allowPii` is true, then `object.toUnsanitised()` is logged, with `containsPii` set to the object's `containsPii()`.
 - Every class should implement `toString()` as well, in case an object is accidentally turned into a String without using `toSafeString()`. E.g. "SDK will return result $result" (which is short-hand for `result.toString()`).
- `LoggerTest` has been updated with 4 additional test cases.

Note 1: it would have been more efficient to have `Logging` define `override fun toString()`, meaning every class would be required to override `toString()`. However, this is not possible in Kotlin (see [link](https://stackoverflow.com/questions/53771394/override-and-implement-fn-from-class-in-interface)).

Note 2: the issue that this PR addresses applies to `data` classes. However, this PR also includes some updates to regular classes, and thereby partially addressing https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2811937

Accompanying MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2081


common4j/src/main/com/microsoft/identity/common/java/nativeauth/util/Logging.kt
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.identity.common.java.nativeauth.util
+
+interface Logging {
```
rmccahill: Suggestion - would `ILoggable` work better as a name here?

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -284,6 +289,24 @@ public static void info(final String tag,
...
+     * @param message The message to log.
+     * @param object The object to be printed.
+     */
+    public static void infoWithObject(final String tag,
+                            final String message,
+                            final Logging object) {
+        if (isAllowPii()) {
+            log(tag, Logger.LogLevel.INFO, null, message, object.toSafeString(true), null, object.containsPii());
+        } else {
+            log(tag, Logger.LogLevel.INFO, null, message, object.toSafeString(false), null, false);
```
Yuki-YuXin: Why don't `log(tag, Logger.LogLevel.INFO, null, message, object.toSafeString(false), null, object.containsPii());`

SammyO: Good question. In this case, the developer has set that they don't want PII to be logged. `object.containsPii()` returns `true` if it contains fields that are PII. So we don't want to use this method, as that would make the value `containsPii` true, which is a contradiction of the configuration the developer set. 

Maybe I should rethink this approach, but my goal was to be able to distinguish between 3 cases: 
1. developer doesn't want PII to be logged. No PII should be logged (`object.toSafeString(mayContainPii = false)` and `containsPii = false`).
2. developer wants PII to be logged (`object.toSafeString(mayContainPii = true)` - this may or may not log PII)
   2a. The log actually contains PII - `object.containsPii()` returns `true`, which sets `containsPii` to `true`.
   2b. The log doesn't contain PII - `object.containsPii()` returns `false`, which sets `containsPii` to `false`.
   
 Developers may want to do different things depending on the value of `containsPii`.

Yuki-YuXin: I guess `object.containsPii()` indicates whether the object contains PII information.  `mayContainPii` determines which version of logging may need to be used and always remains the same as` isAllowPii()`. 

1. Developer doesn't want PII to be logged (`isAllowPii()=false`) 
1a. Object has PII info(`containsPii=true`).  -> mayContainPii=false, choose the version without PII -> void log return. There is no logging.     **That's not the ideal outcome. Thus use false instead of object.containsPii()**
1b. Object doesn't have PII info(`containsPii=false`). ->  no matter the value of mayContainPii, the log should be the same.  -> Log without PII info.

2. Developer wants PII to be logged (`isAllowPii()=true`)
2a.  Object has PII info(`containsPii=true`). -> mayContainPii=true, choose the PII version -> Log with PII info.
2b. Object doesn't have PII info(`containsPii=false`) ->  no matter the value of mayContainPii, the log should be the same. -> Log without PII info.

I understand why the `'false'` is set up here.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -62,7 +70,19 @@ interface INativeAuthCommandResult {
         SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
         SignUpSubmitPasswordCommandResult,
         ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
-        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult
+        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toSafeString(mayContainPii: Boolean): String {
+            return if (mayContainPii) {
```
SaurabhMSFT: Should it not be the other way round? If the object may contain PII then we just print the `correlationId`. If the object definitley does not contain PII then we print other details as well.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
SammyO: Exactly. In the case of 1a, we can't print any PII if `isAllowPii` is set to `false`. Thus, we use the `toSafeString(mayContainPii=false)`, which should return a string that contains no PII. Thus, by default, `containsPii` will always be `false`.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -62,7 +70,19 @@ interface INativeAuthCommandResult {
...
+                "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes"
+            } else {
+                "UnknownError(correlationId=$correlationId)"
+            }
+        }
+
+        override fun containsPii(): Boolean = true
```
SaurabhMSFT: Is there a use case for this flag to be set after the object has been constructed? It would be better if this value is set in the contructor and read only after that.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -62,7 +70,19 @@ interface INativeAuthCommandResult {
...
+
+        override fun toString(): String = toSafeString(false)
```
SaurabhMSFT: The `toString()` method should use `toSafeString(true)` as that would mean that any accidental use of this method does not contain any PII data. If anyone wants a String object with PII then they explicitly call the _dangerous_ method.

SaurabhMSFT (changes requested):

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: `mayContainPii` doesn't state whether the object contains PII. Rather it states whether the string that `toSafeString()` returns may contain PII. In the case of `true`, we can use PII fields like `error`. In the case of `false`, only `correlationId` is safe to return.

I will add comments explaining the role of these methods and params.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: I'm not sure I'm following. This method isn't a setter, but a getter. It isn't set, really. It's a "constant" that will always return true or false.

Depending on which interface/abstract class hierarchy we choose, we can make it a field in the constructor as well.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: I believe I answered this above (i.e. a different meaning of `mayContainPii`). Let me know if it doesn't.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -62,7 +70,19 @@ interface INativeAuthCommandResult {
         SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
         SignUpSubmitPasswordCommandResult,
         ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
-        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult
+        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toSafeString(mayContainPii: Boolean): String {
+            return if (mayContainPii) {
+                "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes"
+            } else {
+                "UnknownError(correlationId=$correlationId)"
```
SaurabhMSFT: errorCodes are numbers and do not contain PII so they should be part of both outputs.

SaurabhMSFT (commented):
My preferred approach would have been the design suggested by Giannis in Teams chat where the toString method is implemented only once in the abstract Logging class. As multiple inheritance is not supported, we had to rule out that option. So, we are left with a situation where we have to override toString method in all classes. In this scenario I would prefer the toString method to be the "safe" option by default and create another method toUnsafeString() that would return a string containing PII values. This would mean the  caller of the classes by default get the benefit of "safety". If anyone wants a string with PII then they must call the toUnsafeString() method. 

The interface would look like

```
interface Logging {
    fun toUnsafeString(): String

    fun containsPii(): Boolean
}

data class UnknownError(
    override val error: String?,
    override val errorDescription: String?,
    override val details: List<Map<String, String>>? = null,
    override val correlationId: String,
    override val errorCodes: List<Int>? = null,
    val exception: Exception? = null
): Error(error, errorDescription, details, correlationId, errorCodes), INativeAuthCommandResult,
    SignInStartCommandResult, SignInWithContinuationTokenCommandResult, SignInSubmitCodeCommandResult,
    SignInResendCodeCommandResult, SignInSubmitPasswordCommandResult,
    SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult,
    SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
    SignUpSubmitPasswordCommandResult,
    ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
    ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {    
    
    override fun toUnsafeString(): String { return "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes" }

    override fun containsPii(): Boolean = true

    override fun toString(): String { return "UnknownError(correlationId=$correlationId)" }
}

```

This design has the same complexity but is more readable as the developer does not have to understand the meaning of the parameter

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: They're not logged by us currently (in the API layer). We based this on MSAL's approach, who don't log errors. I'll check with MSAL folks to get their input.

SammyO:
> My preferred approach would have been the design suggested by Giannis in Teams chat where the toString method is implemented only once in the abstract Logging class. As multiple inheritance is not supported, we had to rule out that option. So, we are left with a situation where we have to override toString method in all classes. In this scenario I would prefer the toString method to be the "safe" option by default and create another method toUnsafeString() that would return a string containing PII values. This would mean the caller of the classes by default get the benefit of "safety". If anyone wants a string with PII then they must call the toUnsafeString() method.
> 
> The interface would look like
> 
> ```
> interface Logging {
>     fun toUnsafeString(): String
> 
>     fun containsPii(): Boolean
> }
> 
> data class UnknownError(
>     override val error: String?,
>     override val errorDescription: String?,
>     override val details: List<Map<String, String>>? = null,
>     override val correlationId: String,
>     override val errorCodes: List<Int>? = null,
>     val exception: Exception? = null
> ): Error(error, errorDescription, details, correlationId, errorCodes), INativeAuthCommandResult,
>     SignInStartCommandResult, SignInWithContinuationTokenCommandResult, SignInSubmitCodeCommandResult,
>     SignInResendCodeCommandResult, SignInSubmitPasswordCommandResult,
>     SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult,
>     SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
>     SignUpSubmitPasswordCommandResult,
>     ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
>     ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {    
>     
>     override fun toUnsafeString(): String { return "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes" }
> 
>     override fun containsPii(): Boolean = true
> 
>     override fun toString(): String { return "UnknownError(correlationId=$correlationId)" }
> }
> ```
> 
> This design has the same complexity but is more readable as the developer does not have to understand the meaning of the parameter

@SaurabhMSFT  I see your point. I don't like this exact approach though, because:
- `toUnsafeString()` doesn't apply to all classes; not all classes have unsafe parameters. It depends on the value of `containsPii()`. The term `unsafe string` gives a false impression that the string is always unsafe, which it isn't.
- the current `toSafeString()` returns a string that given the context is considered safe (either by removing PII, or by including it if the developer wants to add it, in which case it's also safe).

can you think of a way to address the above? For example, by renaming `mayContainPii` and setting a default value?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Should we remove the fields altogether, or obfuscate them? 
So `"UnknownError(correlationId=$correlationId)"`, or `UnknownError(correlationId=***, error=***, errorDescription=***, details=***, errorCodes=***)`

SaurabhMSFT:
> > My preferred approach would have been the design suggested by Giannis in Teams chat where the toString method is implemented only once in the abstract Logging class. As multiple inheritance is not supported, we had to rule out that option. So, we are left with a situation where we have to override toString method in all classes. In this scenario I would prefer the toString method to be the "safe" option by default and create another method toUnsafeString() that would return a string containing PII values. This would mean the caller of the classes by default get the benefit of "safety". If anyone wants a string with PII then they must call the toUnsafeString() method.
> > The interface would look like
> > ```
> > interface Logging {
> >     fun toUnsafeString(): String
> > 
> >     fun containsPii(): Boolean
> > }
> > 
> > data class UnknownError(
> >     override val error: String?,
> >     override val errorDescription: String?,
> >     override val details: List<Map<String, String>>? = null,
> >     override val correlationId: String,
> >     override val errorCodes: List<Int>? = null,
> >     val exception: Exception? = null
> > ): Error(error, errorDescription, details, correlationId, errorCodes), INativeAuthCommandResult,
> >     SignInStartCommandResult, SignInWithContinuationTokenCommandResult, SignInSubmitCodeCommandResult,
> >     SignInResendCodeCommandResult, SignInSubmitPasswordCommandResult,
> >     SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult,
> >     SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
> >     SignUpSubmitPasswordCommandResult,
> >     ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
> >     ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {    
> >     
> >     override fun toUnsafeString(): String { return "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes" }
> > 
> >     override fun containsPii(): Boolean = true
> > 
> >     override fun toString(): String { return "UnknownError(correlationId=$correlationId)" }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This design has the same complexity but is more readable as the developer does not have to understand the meaning of the parameter
> 
> @SaurabhMSFT I see your point. I don't like this exact approach though, because:
> 
> * `toUnsafeString()` doesn't apply to all classes; not all classes have unsafe parameters. It depends on the value of `containsPii()`. The term `unsafe string` gives a false impression that the string is always unsafe, which it isn't.
> * the current `toSafeString()` returns a string that given the context is considered safe (either by removing PII, or by including it if the developer wants to add it, in which case it's also safe).
> 
> can you think of a way to address the above? For example, by renaming `mayContainPii` and setting a default value?

We can call the method `toUnsanitizedString()'. That would should the intent that the output of this method may contain PII.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
Yuki-YuXin: I would prefer obfuscation, giving the develop a hint that they can choose to change the config to have more debug info.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
```diff
@@ -59,7 +59,7 @@ class AcquireTokenNoFixedScopesCommand(
 
         Logger.info(
             TAG,
-            "Returning result: $result"
+            "Returning result: AcquireTokenResult()"
```
rmccahill, line 62: Is there a reason for returning a new result, instead of the one provided from the function above?

Additionally, should it follow the format used elsewhere of `Logger.infoWithObject()` instead of `Logger.info()`?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -220,10 +220,11 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                 is SignInTokenApiResult.MFARequired, is SignInTokenApiResult.CodeIncorrect,
                 is SignInTokenApiResult.UserNotFound, is SignInTokenApiResult.InvalidCredentials,
                 is SignInTokenApiResult.UnknownError -> {
-                    Logger.warn(
+                    Logger.warnWithObject(
                         TAG,
                         tokenApiResult.correlationId,
-                        "Unexpected result: $tokenApiResult"
+                        "Unexpected result: ",
+                        tokenApiResult
```
rmccahill, line 227: More of a question than a request: From what I can see, there is a correlation ID on most of the result objects, which will be logged from the `ILoggable` interface. With that, do we need the additional `tokenApiResult.correlationId` parameter explicitly passed on line 225 as well? Might make the logs cleaner. Seems to be the case for a few lines in this file.

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -417,6 +487,51 @@ public void run() {
         });
     }
 
+    /**
+     * Temporary method to allow the project to compile, without having to make changes everywhere
+     */
```
rmccahill: Is this comment / method still relevant with the complete changes?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
SammyO: There's not a new result being returned, it's just a string being logged. So it's not `AcquireTokenResult()` but `"AcquireTokenResult()"`.

`infoWithObject()` requires the actual object to implement `ILoggable`. I didn't want to make those changes to `AcquireTokenObject`, as it's owned by MSAL and out of scope of this refactoring. 

So with this code change I actually want to remove the `result` from being printed at all (as it contains data that shouldn't be logged). I wanted to keep something informative there though, so I opted for `"Returning result: AcquireTokenResult()"` to at least inform the developer that this result class is being returned.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -43,7 +44,14 @@ interface INativeAuthCommandResult {
...
                 private const val BROWSER_REQUIRED_ERROR_DESCRIPTION: String = "The client's authentication capabilities are insufficient. Please redirect to the browser to complete authentication"
             }
-        }
+
+
+        override fun toUnsanitizedString(): String = "Redirect(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "Redirect(correlationId=$correlationId"
```
rmccahill: nit: missing end bracket. same for `toUnsanitizedString()` below

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SammyO: Good question. I haven't made any changes to other parts of logging, as that would make this PR bigger than it is, and @Yuki-YuXin  is working on that. My proposal would be to proceed with the version of this PR, and take a look at what you're pointing out in Yu's "log cleanup" work. Wdyt?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -80,5 +94,11 @@ interface INativeAuthCommandResult {
         override val errorCodes: List<Int>? = null,
         val exception: Exception? = null
     ) : Error(error, errorDescription, details, correlationId, errorCodes),
-        INativeAuthCommandResult, SignInStartCommandResult, SignUpStartCommandResult, SignUpSubmitPasswordCommandResult, ResetPasswordStartCommandResult
+        INativeAuthCommandResult, SignInStartCommandResult, SignUpStartCommandResult, SignUpSubmitPasswordCommandResult, ResetPasswordStartCommandResult {
+        override fun toUnsanitizedString(): String = "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "UnknownError(correlationId=$correlationId)"
```
rmccahill: Class name should be `InvalidUsername`, same for line 98

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
```diff
@@ -46,60 +46,98 @@ interface ResetPasswordCommandResult {
...
 
     data class PasswordRequired(
         override val correlationId: String,
         val continuationToken: String
-    ) : ResetPasswordSubmitCodeCommandResult
+    ) : ResetPasswordSubmitCodeCommandResult {
+        override fun toUnsanitizedString(): String = "PasswordRequired(correlationId=$correlationId)"
+
+        override fun toString(): String = toUnsanitizedString()
+    }
```
rmccahill, line 90: nit: `fun containsPii()` is not overridden here, same with `PasswordNotSet`, `UserNotFound`. The base implementation seems to have it, but probably better to explicitly state in each class.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
```diff
@@ -46,60 +46,98 @@ interface ResetPasswordCommandResult {
...
         val error: String,
         val errorDescription: String,
-        override val correlationId: String,
         val subError: String
-    ) :
-        ResetPasswordSubmitNewPasswordCommandResult
+    ) : ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toUnsanitizedString(): String = "PasswordNotAccepted(correlationId=$correlationId, error=$error, errorDescription=$errorDescription, subError=$subError)"
+
+        override fun containsPii(): Boolean = false
```
rmccahill: Should be true?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
```diff
@@ -46,60 +46,98 @@ interface ResetPasswordCommandResult {
...
+        override val correlationId: String,
         val error: String,
         val errorDescription: String,
-        override val correlationId: String
-    ) :
-        ResetPasswordSubmitNewPasswordCommandResult
+    ) : ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toUnsanitizedString(): String = "PasswordResetFailed(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
+
+        override fun containsPii(): Boolean = false
```
rmccahill: Same as above, should be true

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
```diff
@@ -81,60 +98,95 @@ interface SignUpCommandResult {
...
         val error: String,
         val errorDescription: String,
-        override val correlationId: String,
-    ) : SignUpStartCommandResult
+    ) : SignUpStartCommandResult {
+        override fun toUnsanitizedString(): String = "AuthNotSupported(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "AuthNotSupported(correlationId=$correlationId"
```
rmccahill: nit: end bracket missing here

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
```diff
@@ -81,60 +98,95 @@ interface SignUpCommandResult {
...
         val errorDescription: String,
         val invalidAttributes: List<String>,
-        override val correlationId: String,
-    ) : SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult
+    ) : SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult {
+        override fun toUnsanitizedString(): String = "InvalidAttributes(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "InvalidAttributes(correlationId=$correlationId, invalidAttributes=$invalidAttributes)"
```
rmccahill: Question: do the invalid attributes returned from the server contain the values in those attributes? If they do, there's potentially sensitive information in there like names, addresses, etc. If it's just the names of the attributes it's probably fine.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/IApiResponse.kt
```diff
@@ -32,7 +31,7 @@ abstract class IApiResponse(
     //HTTP status code
     @Transient open val statusCode: Int,
     correlationId: String
-) {
+) : ILoggable {
```
rmccahill, line 34: Is there a reason for specifying `ILoggable` here, instead of in the subclasses?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordChallengeRequest.kt
```diff
@@ -71,9 +71,21 @@ class ResetPasswordChallengeRequest private constructor(
...
         @SerializedName("client_id") override val clientId: String,
         @SerializedName("continuation_token") val continuationToken: String,
         @SerializedName("challenge_type") val challengeType: String?
-    ) : NativeAuthRequestParameters()
+    ) : NativeAuthRequestParameters() {
+        override fun toUnsanitizedString(): String = "NativeAuthResetPasswordChallengeRequestParameters(clientId=$clientId, challengeType=$challengeType)"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "NativeAuthResetPasswordChallengeRequestParameters(clientId=$clientId)"
```
rmccahill, line 89: Should the client ID be present in the sanitised version of the string?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordPollCompletionRequest.kt
```diff
@@ -60,20 +60,32 @@ class ResetPasswordPollCompletionRequest private constructor(
...
      */
-    data class NativeAuthResetPasswordPollCompletionRequestBody(
+    data class NativeAuthResetPasswordPollCompletionRequestParameters(
         @SerializedName("client_id") override val clientId: String,
         @SerializedName("continuation_token") val continuationToken: String
-    ) : NativeAuthRequestParameters()
+    ) : NativeAuthRequestParameters() {
+        override fun toUnsanitizedString(): String = "NativeAuthResetPasswordPollCompletionRequestParameters(clientId=$clientId)"
+
+        override fun containsPii(): Boolean = true
```
rmccahill: Can probably be false here

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordSubmitRequest.kt
```diff
@@ -73,13 +73,25 @@ class ResetPasswordSubmitRequest private constructor(
...
-    data class NativeAuthResetPasswordSubmitRequestBody(
+    data class NativeAuthResetPasswordSubmitRequestParameters(
         @SerializedName("client_id") override val clientId: String,
         @SerializedName("continuation_token") val continuationToken: String,
         @JsonAdapter(CharArrayJsonAdapter::class) @SerializedName("new_password") val newPassword: CharArray
-    ) : NativeAuthRequestParameters()
+    ) : NativeAuthRequestParameters() {
+        override fun toUnsanitizedString(): String = "NativeAuthResetPasswordSubmitRequestParameters(clientId=$clientId)"
+
+        override fun containsPii(): Boolean = true
```
rmccahill: Can probably be false here

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/signin/SignInTokenRequest.kt
```diff
@@ -172,5 +178,11 @@ data class SignInTokenRequest private constructor(
         @SerializedName("continuation_token") val continuationToken: String? = null,
         @SerializedName("scope") val scope: String?,
         @SerializedName("challenge_type") val challengeType: String?
-    ) : NativeAuthRequestParameters()
+    ) : NativeAuthRequestParameters() {
+        override fun toUnsanitizedString(): String = "NativeAuthRequestSignInTokenRequestParameters(nca=$nca, clientInfo=$clientInfo, clientId=$clientId, grantType=$grantType, scope=$scope, challengeType=$challengeType)"
+
+        override fun containsPii(): Boolean = true
+
+        override fun toString(): String = "NativeAuthRequestSignInTokenRequestParameters((nca=$nca, clientInfo=$clientInfo, clientId=$clientId)"
```
rmccahill: nit: double bracket here

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
rmccahill: Ah, I misread. Makes sense 👍

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
rmccahill: Makes sense 👍

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: Explained over Teams. Let me know whether you think that's okay.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/util/ILoggable.kt
```diff
@@ -0,0 +1,23 @@
...
+     * The value of containsPii() will indicate whether the value actually contains PII.
+     */
+    fun toUnsanitizedString(): String
+
+    /**
+     * This method indicates whether the implementing class contains data fields that are considered
+     * PII (Personally identifiable information). If this method returns true, then the value of
+     * toSafeString(true) will return a String that contains PII.
+     */
+    fun containsPii(): Boolean = toString() != toUnsanitizedString()
```
SammyO, line 15: Add tests for this

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
```diff
@@ -284,6 +333,27 @@ public static void info(final String tag,
...
+     * string representation of the object will be used.
+     *
+     * @param tag     Used to identify the source of a log message. It usually identifies the class
+     *                or activity where the log call occurs.
+     * @param message The message to log.
+     * @param object The object to be printed.
+     */
+    public static void infoWithObject(final String tag,
+                            final String message,
+                            final ILoggable object) {
```
Yuki-YuXin, line 345: Do we need public static void infoWithObject(final String tag, **final String correlationID**, final String message, final ILoggable object) like warnWithObject

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -62,15 +70,21 @@ interface INativeAuthCommandResult {
         SignUpSubmitCodeCommandResult, SignUpResendCodeCommandResult,
         SignUpSubmitPasswordCommandResult,
         ResetPasswordStartCommandResult, ResetPasswordSubmitCodeCommandResult,
-        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult
+        ResetPasswordResendCodeCommandResult, ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toUnsanitizedString(): String = "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes"
```
Yuki-YuXin: nit: `UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription, details=$details, errorCodes=$errorCodes)` ?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
```diff
@@ -80,5 +94,11 @@ interface INativeAuthCommandResult {
         override val errorCodes: List<Int>? = null,
         val exception: Exception? = null
     ) : Error(error, errorDescription, details, correlationId, errorCodes),
-        INativeAuthCommandResult, SignInStartCommandResult, SignUpStartCommandResult, SignUpSubmitPasswordCommandResult, ResetPasswordStartCommandResult
+        INativeAuthCommandResult, SignInStartCommandResult, SignUpStartCommandResult, SignUpSubmitPasswordCommandResult, ResetPasswordStartCommandResult {
+        override fun toUnsanitizedString(): String = "UnknownError(correlationId=$correlationId, error=$error, errorDescription=$errorDescription), details=$details, errorCodes=$errorCodes"
```
Yuki-YuXin: Here might be` InvalidUsername(correlationId=$correlationId, error=$error, errorDescription=$errorDescription, details=$details, errorCodes=$errorCodes)`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
```diff
@@ -22,10 +22,10 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.java.nativeauth.controllers.results
 
-import com.microsoft.identity.common.java.logging.DiagnosticContext
+import com.microsoft.identity.common.java.nativeauth.util.ILoggable
 
 
-sealed interface ResetPasswordStartCommandResult: INativeAuthCommandResult
+sealed interface ResetPasswordStartCommandResult: INativeAuthCommandResult, ILoggable
```
Yuki-YuXin: Does ResetPasswordStartCommandResult needs to implement ILoggable here?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
```diff
@@ -46,60 +46,98 @@ interface ResetPasswordCommandResult {
...
         val errorDescription: String,
-        override val correlationId: String
-    ) :
-        ResetPasswordSubmitNewPasswordCommandResult
+    ) : ResetPasswordSubmitNewPasswordCommandResult {
+        override fun toUnsanitizedString(): String = "PasswordResetFailed(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
+
+        override fun containsPii(): Boolean = false
+
+        override fun toString(): String = "PasswordResetFailed(correlationId=$correlationId"
```
Yuki-YuXin: nit: missing bracket.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignInCommandResult.kt
```diff
@@ -42,47 +41,78 @@ interface SignInCommandResult {
...
         override val correlationId: String,
         val continuationToken: String
-    ) :
-        SignInStartCommandResult
+    ) : SignInStartCommandResult {
+        override fun toUnsanitizedString(): String = "PasswordRequired(correlationId=$correlationId)"
+
+        override fun containsPii(): Boolean = false
+
+        override fun toString(): String = "PasswordRequired(correlationId=$correlationId)"
```
Yuki-YuXin: How about following the practice `override fun toString(): String = toUnsanitizedString()`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
```diff
@@ -81,60 +98,95 @@ interface SignUpCommandResult {
...
      */
     data class InvalidAttributes(
+        override val correlationId: String,
         val error: String,
         val errorDescription: String,
         val invalidAttributes: List<String>,
-        override val correlationId: String,
-    ) : SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult
+    ) : SignUpStartCommandResult, SignUpSubmitUserAttributesCommandResult {
+        override fun toUnsanitizedString(): String = "InvalidAttributes(correlationId=$correlationId, error=$error, errorDescription=$errorDescription)"
```
Yuki-YuXin: nit: missing invalidAttributes here.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordChallengeRequest.kt
Yuki-YuXin, line 85: It might be helpful to record what attributes contain PII. ChallengeType, username, attributes, grantType, scope, status, pollInterval are new to me. oob, password,continuationToken should not be recorded in both logs.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/util/Logging.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Please resolve if addressed @SaurabhMSFT

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Please resolve if addressed @SaurabhMSFT

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordChallengeApiResponse.kt
```diff
@@ -51,6 +51,19 @@ class ResetPasswordChallengeApiResponse(
...
+                "correlationId=$correlationId, challengeType=$challengeType, " +
+                "bindingMethod=$bindingMethod, challengeTargetLabel=$challengeTargetLabel, " +
+                "challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval, " +
+                "error=$error, errorDescription=$errorDescription, errorUri=$errorUri)"
+    }
+
+    override fun containsPii(): Boolean = true
+
+    override fun toString(): String = "ResetPasswordChallengeApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 63: nit: Missing bracket and some parameters. ("challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval")

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
SammyO: Good catch. Updated.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: that's intentional; `ILoggable` defines this. I've removed all unnecessary overridings of `containsPII()`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
SammyO: A few points on this: 
- I believe the server actually doesn't return this response at all. It's a missing feature at the moment.
- Our code assumes that if/when it is returned, the API returns a `List<Map<String, String>>`, of pairs as such: `"name"` - `"<attribute_key>"`. See `internal fun List<Map<String, String>>.toAttributeList(): List<String> `.
- For now, I want to error on the side of caution and consider it PII.
- Which means I need to move it from `toString()` to `toUnsanitisedString()`.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordContinueApiResponse.kt
```diff
@@ -48,6 +48,17 @@ class ResetPasswordContinueApiResponse(
...
+    override fun toUnsanitizedString(): String {
+        return "ResetPasswordContinueApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, challengeType=$challengeType, expiresIn=$expiresIn " +
+                "error=$error, errorUri=$errorUri, errorDescription=$errorDescription, subError=$subError)"
+    }
+
+    override fun containsPii(): Boolean = true
+
+    override fun toString(): String = "ResetPasswordContinueApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 58: nit: Missing bracket.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/IApiResponse.kt
SammyO: The higher up the hierarchy we implement `ILoggable`, the less likely it is that future code changes will overlook this. The need to extend `ILoggable` is the only "manual" step in the process, and the most likely place where we can introduce oversights. Thus, by adding it to base classes, I'm hoping it's less likely it'll be overlooked.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordPollCompletionApiResponse.kt
```diff
@@ -46,15 +46,26 @@ import java.net.HttpURLConnection
...
+    override fun toUnsanitizedString(): String {
+        return "ResetPasswordPollCompletionApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, status=$status, expiresIn=$expiresIn " +
+                "error=$error, errorUri=$errorUri, errorDescription=$errorDescription, subError=$subError)"
+    }
+
+    override fun containsPii(): Boolean = true
+
+    override fun toString(): String = "ResetPasswordPollCompletionApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 65: nit: Missing bracket.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordPollCompletionRequest.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordSubmitRequest.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/util/ILoggable.kt
SammyO: done

common4j/src/main/com/microsoft/identity/common/java/logging/Logger.java
SammyO: There are no usages of this method yet, so I didn't add it. You may have to add it with your "clean up logging" PR - is that okay?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/INativeAuthCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/ResetPasswordCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordStartApiResponse.kt
```diff
@@ -45,6 +45,15 @@ class ResetPasswordStartApiResponse(
...
 ): IApiResponse(statusCode, correlationId) {
 
+    override fun toUnsanitizedString(): String {
+        return "ResetPasswordStartApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, challengeType=$challengeType, error=$error, " +
+                "errorUri=$errorUri, errorDescription=$errorDescription)"
+    }
+
+    override fun toString(): String = "ResetPasswordStartApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 55: nit: Missing bracket here.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignInCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordStartApiResult.kt
```diff
@@ -32,12 +32,20 @@ import com.microsoft.identity.common.java.nativeauth.providers.responses.ApiResu
 sealed interface ResetPasswordStartApiResult: ApiResult {
     data class Redirect(
         override val correlationId: String,
-    ): ResetPasswordStartApiResult
+    ): ResetPasswordStartApiResult {
+        override fun toUnsanitizedString() = "Redirect(correlationId=$correlationId)"
+
+        override fun toString(): String = "Redirect(correlationId=$correlationId)"
```
Yuki-YuXin, line 38: nit: Maybe toUnsanitizedString()

common4j/src/main/com/microsoft/identity/common/java/nativeauth/controllers/results/SignUpCommandResult.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordStartApiResult.kt
```diff
@@ -32,12 +32,20 @@ import com.microsoft.identity.common.java.nativeauth.providers.responses.ApiResu
...
+    }
 
     data class Success(
         val continuationToken: String,
         override val correlationId: String,
-    ) : ResetPasswordStartApiResult
+    ) : ResetPasswordStartApiResult {
+        override fun toUnsanitizedString() = "Success(correlationId=$correlationId)"
+
+        override fun toString(): String = "Success(correlationId=$correlationId)"
```
Yuki-YuXin, line 47: nit: Maybe toUnsanitizedString()

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/resetpassword/ResetPasswordSubmitApiResponse.kt
```diff
@@ -44,6 +44,15 @@ class ResetPasswordSubmitApiResponse(
...
 ): IApiResponse(statusCode, correlationId) {
 
+    override fun toUnsanitizedString(): String {
+        return "ResetPasswordSubmitApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, pollInterval=$pollInterval, error=$error, " +
+                "errorUri=$errorUri, errorDescription=$errorDescription, subError=$subError)"
+    }
+
+    override fun toString(): String = "ResetPasswordSubmitApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 54: nit: Missing bracket here.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordChallengeRequest.kt
SammyO: can you explain further @Yuki-YuXin ? I don't see grantType, scopes, etc. in this class.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/signin/SignInTokenRequest.kt
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInChallengeApiResponse.kt
```diff
@@ -52,6 +52,18 @@ class SignInChallengeApiResponse(
...
+        return "SignInChallengeApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, challengeType=$challengeType, " +
+                "bindingMethod=$bindingMethod, challengeTargetLabel=$challengeTargetLabel, " +
+                "challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval, " +
+                "error=$error, errorDescription=$errorDescription, errorCodes=$errorCodes, " +
+                "errorUri=$errorUri)"
+    }
+
+    override fun toString(): String = "SignInChallengeApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 65: nit: Missing bracket and some parameters. ("challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval")

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResponse.kt
```diff
@@ -43,14 +43,24 @@ import com.microsoft.identity.common.java.nativeauth.util.isUserNotFound
...
     ): IApiResponse(statusCode, correlationId) {
 
+    override fun toUnsanitizedString(): String {
+        return "SignInTokenApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, error=$error, errorDescription=$errorDescription, " +
+                "errorCodes=$errorCodes, errorUri=$errorUri, subError=$subError)"
+    }
+
+    override fun toString(): String = "SignInTokenApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 61: nit: Missing bracket.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResult.kt
```diff
@@ -22,19 +22,26 @@
...
 import com.microsoft.identity.common.java.nativeauth.providers.responses.ApiResult
+import com.microsoft.identity.common.java.nativeauth.util.ILoggable
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsTokenResponse
 
 /**
  * Represents the potential result types returned from the OAuth /token endpoint,
  * including a case for unexpected errors received from the server.
  */
-sealed interface SignInTokenApiResult: ApiResult {
+sealed interface SignInTokenApiResult: ApiResult, ILoggable {
```
Yuki-YuXin, line 34: Is ILoggable necessary here?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signup/SignUpChallengeApiResponse.kt
```diff
@@ -41,17 +41,28 @@ import java.net.HttpURLConnection
...
+    override fun toUnsanitizedString(): String {
+        return "SignInChallengeApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, challengeType=$challengeType, " +
+                "bindingMethod=$bindingMethod, challengeTargetLabel=$challengeTargetLabel, " +
+                "challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval, " +
+                "error=$error, errorDescription=$errorDescription)"
+    }
+
+    override fun toString(): String = "SignInChallengeApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 64: nit: Missing bracket and some parameters. ("challengeChannel=$challengeChannel, codeLength=$codeLength, interval=$interval")

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signup/SignUpContinueApiResponse.kt
```diff
@@ -63,6 +63,15 @@ class SignUpContinueApiResponse(
...
 ) : IApiResponse(statusCode, correlationId) {
 
+    override fun toUnsanitizedString(): String {
+        return "SignUpContinueApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, expiresIn=$expiresIn, requiredAttributes=$requiredAttributes, " +
+                "error=$error, errorCodes=$errorCodes, errorDescription=$errorDescription, subError=$subError)"
+    }
+
+    override fun toString(): String = "SignUpContinueApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 73: nit: Missing bracket

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signup/SignUpContinueApiResult.kt
```diff
@@ -61,7 +72,14 @@ sealed interface SignUpContinueApiResult: ApiResult {
...
         errorDescription = errorDescription,
         correlationId = correlationId
-    ), SignUpContinueApiResult
+    ), SignUpContinueApiResult {
+        override fun toUnsanitizedString(): String = "AttributesRequired(correlationId=$correlationId, " +
+                "error=$error, errorDescription=$errorDescription, " +
+                "requiredAttributes=$requiredAttributes)"
+
+        override fun toString(): String = "AttributesRequired(correlationId=$correlationId, " +
+                "requiredAttributes=$requiredAttributes)"
```
Yuki-YuXin, line 81: What's the rule applied to the attributes? Currently attributes submitted by developers should not exist in both logs, however, invalid attributes contain PII. Required attributes are not. May I ask the reasons behind?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signup/SignUpStartApiResponse.kt
```diff
@@ -60,6 +60,15 @@ class SignUpStartApiResponse(
...
 ) : IApiResponse(statusCode, correlationId) {
 
+    override fun toUnsanitizedString(): String {
+        return "SignUpStartApiResponse(statusCode=$statusCode, " +
+                "correlationId=$correlationId, challengeType=$challengeType" +
+                "error=$error, errorCodes=$errorCodes, errorDescription=$errorDescription, subError=$subError)"
+    }
+
+    override fun toString(): String = "SignUpContinueApiResponse(statusCode=$statusCode, " +
+            "correlationId=$correlationId"
```
Yuki-YuXin, line 70: nit: Missing bracket.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signup/SignUpStartApiResult.kt
```diff
@@ -45,38 +51,52 @@ sealed interface SignUpStartApiResult: ApiResult {
...
         val subError: String,
-        override val correlationId: String
     ): ApiErrorResult(
         error = error,
         errorDescription = errorDescription,
         correlationId = correlationId
-    ), SignUpStartApiResult
+    ), SignUpStartApiResult {
+        override fun toUnsanitizedString() = "InvalidPassword(correlationId=$correlationId, " +
+                "error=$error, errorDescription=$errorDescription, errorCodes=$errorCodes)"
```
Yuki-YuXin, line 74: nit: Missing subError

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordChallengeRequest.kt
Yuki-YuXin: The log here is perfect without issues. grantType, scopes, etc appear later in the PR, which are also logged correctly.

This comment is to raise a personal suggestion of extending the table in [MSAL Android SDK Logging issues (microsoft.com)](https://loop.microsoft.com/p/eyJwIjp7InUiOiJodHRwczovL21pY3Jvc29mdGV1ci1teS5zaGFyZXBvaW50LmNvbS9wZXJzb25hbC9zb2RlbmhvdmVuX21pY3Jvc29mdF9jb20%2FbmF2PWN6MGxNa1p3WlhKemIyNWhiQ1V5Um5OdlpHVnVhRzkyWlc1ZmJXbGpjbTl6YjJaMFgyTnZiU1prUFdJaGVWVlZabVI2UVhveVJUSmxUM0l5V1RneU1GbDVSRXBOYkZwUk9YRXhlRTl2VVdsR2JqbEhlVmt6ZWt4R05rbHJhRkJ2T1ZOaFptaG1Xa2xEVVRrMVVpWm1QVEF4U0RZMFNrODBTbE5EU2xOR1QwdENOVkkxU0V0T1RsRktWelZQUVVSSE1sRW1ZejBtWm14MWFXUTlNUSUzRCUzRCIsInIiOmZhbHNlfX0%3D) and maybe a wiki for future reference.

rmccahill (approved):

SaurabhMSFT (approved):

PR #2385, opened 2024-04-16
Username checking in native auth requests

Yuki-YuXin:
Put back ArgUtils.validateNonNullArg(username, "username") in all initial native auth requests.

Company PRs:
msal: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2080
native sample app: https://github.com/Azure-Samples/ms-identity-ciam-native-auth-android-sample/pull/24

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordStartRequest.kt
+++ b/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordStartRequest.kt
@@ -53,6 +53,7 @@ data class ResetPasswordStartRequest private constructor(
             headers: Map<String, String?>
         ): ResetPasswordStartRequest {
             // Check for empty Strings and empty Maps
+            ArgUtils.validateNonNullArg(username, "username")
             ArgUtils.validateNonNullArg(clientId, "clientId")
             ArgUtils.validateNonNullArg(challengeType, "challengeType")
             ArgUtils.validateNonNullArg(requestUrl, "requestUrl")
--- a/common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
+++ b/common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
@@ -96,8 +96,8 @@ class NativeAuthRequestHandlerTest {
         )
 
     // signup start tests
-    @Test
-    fun testSignUpStartWithEmptyUsernameShouldNotThrowException() {
+    @Test(expected = ClientException::class)
+    fun testSignUpStartWithEmptyUsernameShouldThrowException() {
         val commandParameters = SignUpStartCommandParameters.builder()
             .platformComponents(mock<PlatformComponents>())
             .username(emptyString)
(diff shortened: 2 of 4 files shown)


SaurabhMSFT (approved):

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordStartRequest.kt
```diff
@@ -53,6 +53,7 @@ data class ResetPasswordStartRequest private constructor(
             headers: Map<String, String?>
         ): ResetPasswordStartRequest {
             // Check for empty Strings and empty Maps
+            ArgUtils.validateNonNullArg(username, "username")
```
SammyO, line 56: why was this removed in the first place?

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
```diff
@@ -96,8 +96,8 @@ class NativeAuthRequestHandlerTest {
         )
 
     // signup start tests
-    @Test
-    fun testSignUpStartWithEmptyUsernameShouldNotThrowException() {
+    @Test(expected = ClientException::class)
+    fun testSignUpStartWithEmptyUsernameShouldThrowException() {
```
SammyO, line 100: Wasn't this changed (i.e. the exception removed) as part of last year's work to align some of our errors and return an InvalidUsernameError in the case of an empty string? Giannis worked on that I believe.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/requests/resetpassword/ResetPasswordStartRequest.kt
Yuki-YuXin: Not sure about the reason but I would prefer to keep the checking here.

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
Yuki-YuXin: Yes. I think so. But either `CommandResultUtil.kt`
```kotlin
if (this.status != ICommandResult.ResultStatus.COMPLETED) {
        var exception: Exception? = null
        var exceptionMessage: String? = ""

        if (this.result is Exception) {
            exception = this.result as Exception
            exceptionMessage = exception.message
        }

        return com.microsoft.identity.common.java.nativeauth.controllers.results.INativeAuthCommandResult.UnknownError(
            error = UNSUCCESSFUL_COMMAND_ERROR,
            errorDescription = exceptionMessage,
            exception = exception,
            correlationId = this.correlationId
        ) as ExpectedType
```

or` NativeAuthPublicClientApplication.kt`
```kotlin
return withContext(Dispatchers.IO) {
            try {
                verifyNoUserIsSignedIn()

                if (username.isBlank()) {
                    return@withContext SignUpError(
                        errorType = ErrorTypes.INVALID_USERNAME,
                        errorMessage = "Empty or blank username",
                        correlationId = "UNSET"
                    )
                }
```

can handle empty username. 

I think the latter one act the main role of returning InvalidUsernameError before sending out the request/validating request parameters. Thus, I would prefer to keep the ThrowException here.

SammyO: Some thoughts: 
- if we're already doing a username is empty check in the interface, we don't need to do it again in the API layer.
- we shouldn't just be throwing exceptions. An exception is a pretty heavy mechanism to give feedback to a developer. We should aim for something more graceful than that.
- the fact that this "throwing an exception" was removed from this API layer a few months ago, means we had a reason why. If that reason has changed, let's talk about it. We can just revert the work we did a few months ago without good reason.

Yuki-YuXin: Will abandon this PR.

PR #2387, opened 2024-04-18
Extract the mock api url from code to pipeline argument 

Yuki-YuXin:
### Summary
The Mock API domain is hardcoded in our test source files. This domain should not be publicly known. 
Please refer to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2387 for comments related with the PR.

Company MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2103

PR #2388, opened 2024-04-18
Extract the mock api url from code to pipeline env

Yuki-YuXin:
### Summary:
Please go to the PR for further workhttps://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2387
This PR is closed because of the messy state during test.


azure-pipelines/pull-request-validation/common4j.yml
```diff
@@ -1,8 +1,12 @@
 # File: azure-pipelines\pull-request-validation\common4j.yml
 # Description: Run test in common4j
+# Variable: 'MOCK_API_URL' was defined in the Variables tab
 # Uses Java 11
 name: $(date:yyyyMMdd)$(rev:.r)
 
+variables:
```
SammyO, line 7: For my own understanding, why is this needed?

SammyO:
@Yuki-YuXin I'd like to see this PR in action, i.e. run the tests using the new env var. Could you take a look at that?

SammyO:
@Yuki-YuXin there are more instances of the mock API URL in the code base, that haven't been replaced as part of this PR. Some of them might be in MSAL. Could you take a look at that?

azure-pipelines/pull-request-validation/common4j.yml
Yuki-YuXin: https://identitydivision.visualstudio.com/Engineering/_library?itemType=VariableGroups
We can create the variable groups here.

Yuki-YuXin:
Summarize the discussion with Silviu in terms of this work item:

IOS don’t run the tests in the yml file. They run them locally, have a stash that we apply and then discard. As in they don’t run as part of the pipeline. They have different suite for unit which run in the pipeline and integration with mock api which run only locally.

SammyO:
> Summarize the discussion with Silviu in terms of this work item:
> 
> IOS don’t run the tests in the yml file. They run them locally, have a stash that we apply and then discard. As in they don’t run as part of the pipeline. They have different suite for unit which run in the pipeline and integration with mock api which run only locally.

What's your point here @Yuki-YuXin ?

common4j/src/testFixtures/java/com/microsoft/identity/common/nativeauth/ApiConstants.kt
```diff
@@ -30,8 +30,8 @@ import java.net.URL
  */
 interface ApiConstants {
     companion object {
-        const val BASEPATH = "https://native-auth-mock-api.azurewebsites.net/"
-        private const val BASE_REQUEST_PATH = BASEPATH + "1234/"
+        val BASEPATH: String? by lazy { System.getenv("MOCK_API_URL") }
```
SammyO: This is not how we should be dealing with env vars. They should be set as part of the gradle script, in BuildConfig vars. E.g. look at how the Labs API secret is set in various build.gradle files: `buildConfigField("String", "LAB_CLIENT_SECRET", "\"$labClientSecret\"")`.

SammyO: Let me know if you need some help with this.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthOAuth2Configuration.kt
```diff
@@ -51,7 +50,7 @@ class NativeAuthOAuth2Configuration(
 
     companion object {
         //Base url for the mock API to make Native Auth calls. See the swagger at
-        // https://native-auth-mock-api.azurewebsites.net/doc#/ for all possible urls
+        // $(MOCK_API_URL)/doc#/ for all possible urls
         private const val MOCK_API_URL_WITH_NATIVE_AUTH_TENANT = "https://native-auth-mock-api.azurewebsites.net/lumonconvergedps.onmicrosoft.com"
```
SammyO: Mock API value is still set here. Can you please do a ctrl-F in the entire project for this API value string and remove all occurrences?

azure-pipelines/pull-request-validation/common4j.yml
```diff
@@ -17,6 +21,8 @@ pool:
 jobs:
 - job: build_test
   displayName: Build & Test
+  env:
+    MOCK_API_URL: https://native-auth-mock-api.azurewebsites.net/
```
SammyO: Why is this here? Shouldn't this PR be introducing the integration with the environment var, so we can test that?

Yuki-YuXin:
> > Summarize the discussion with Silviu in terms of this work item:
> > IOS don’t run the tests in the yml file. They run them locally, have a stash that we apply and then discard. As in they don’t run as part of the pipeline. They have different suite for unit which run in the pipeline and integration with mock api which run only locally.
> 
> What's your point here @Yuki-YuXin ?

They don't seem to intend to integrate the mock api into the pipeline in the center server and they can easily accept the changes of mock api because every time they use the url locally.

Yuki-YuXin:
> @Yuki-YuXin I'd like to see this PR in action, i.e. run the tests using the new env var. Could you take a look at that?

From the experience of cert PR, if we can pass the PR pipeline verification check, we prove that the new env var actually work

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthOAuth2Configuration.kt
```diff
@@ -51,8 +50,9 @@ class NativeAuthOAuth2Configuration(
 
     companion object {
         //Base url for the mock API to make Native Auth calls. See the swagger at
-        // https://native-auth-mock-api.azurewebsites.net/doc#/ for all possible urls
-        private const val MOCK_API_URL_WITH_NATIVE_AUTH_TENANT = "https://native-auth-mock-api.azurewebsites.net/lumonconvergedps.onmicrosoft.com"
+        // $(MOCK_API_URL)/doc#/ for all possible urls
+        private val BASE_PATH = BuildValues.getMockApiUrlForNativeAuthAuthority()
```
Yuki-YuXin, line 54: The same here. Frist use the concrete string to pass the pipeline since it's not a test file.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthOAuth2Configuration.kt
Yuki-YuXin: I want to first let the PR passed first using the concrete string since it's not a test file.

PR #2389, opened 2024-04-18
 Migrate Base64 away from Msebera (in Android), Fixes AB#3077784

rpdome:
[AB#3077784](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3077784) (Part 2)

## Background

Our entire code base is written to be compatible with Android.util.Base64, which is a fork of a legacy version of apache commons-codec.

This means Android OS ships with apache commons-codec, and we cannot have that as a dependency in common4j project (otherwise the classloader would throw an error).

The safest, previous workaround is to consume [msebera HttpClient](https://github.com/smarek/httpclient-android), which also contains commons-codec, but under a different namespace. [(This is officially suggested by Apache)
](https://hc.apache.org/httpcomponents-client-4.5.x/android.html)

## Issue 

msebera library is no longer maintained, and it was recently flagged with an apk tool.

Even if the flagged part is NOT the code path we use. It's better that we migrate away from it (so that our customer don't see the false alarm)

There are multiple ways to tackle this
1. (Least risky) Make the android project consume Android.util.base64, keep msebera for Linux (which is already in maintainance mode) and test classes
2. Migrate to other libraries. (e.g. guava)
   - NOTE: Java's Base64 is not an option because it requires min API 26

I decide to go with 1 because 
- the other option might increase the final apk/aar size by a lot. 
- It would also introduce a breaking change (e.g. The value in storage - before upgrade - was encoded with msebera, and might not be compatible with the new library. 
- The new library might also handle error cases/unexpected format differently.

## Changes

All the Base64 usage is now consolidated into the Base64Util class. (most of the changes are under this bucket).

This Base64Util class will use a classloader to pick the implementation to use.
- In Android modules, it would load `AndroidBase64`
- In Linux Broker, it would load `MseberaBase64ForLinux`
- In Common4j's own unit test, it would load  `MseberaBase64ForCommon4jTests` (from common4j's testImplementation)
- In Broker4j's own unit test, it would load  `MseberaBase64ForBroker4jTests` (from common4j's testImplementation)
- In other tests, it would load  `MseberaBase64ForTestUtils` (from :testutils)


common4j/src/main/com/microsoft/identity/common/java/base64/Base64.kt
```diff
@@ -0,0 +1,129 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.base64
+
+import com.microsoft.identity.common.java.logging.Logger
+import com.microsoft.identity.common.java.util.StringUtil
+import java.nio.charset.StandardCharsets
+
+class Base64Util {
```
fadidurah: JavaDoc

fadidurah: KotlinDoc*?

common4j/src/main/com/microsoft/identity/common/java/base64/Base64Flags.kt
```diff
@@ -0,0 +1,51 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.base64
+
+enum class Base64Flags {
```
fadidurah, line 28: Nit: KotlinDoc

common4j/src/main/com/microsoft/identity/common/java/base64/IBase64.kt
```diff
@@ -0,0 +1,40 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.base64
+
+interface IBase64 {
```
fadidurah, line 28: Nit: KotlinDoc

common/src/main/java/com/microsoft/identity/common/components/AndroidPlatformComponentsFactory.java
```diff
@@ -36,6 +37,7 @@
 import com.microsoft.identity.common.internal.providers.oauth2.AndroidTaskStateGenerator;
 import com.microsoft.identity.common.internal.ui.AndroidAuthorizationStrategyFactory;
 import com.microsoft.identity.common.java.WarningType;
+import com.microsoft.identity.common.java.base64.Base64Util;
```
fadidurah: These imports needed?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/base64/AndroidBase64.kt
```diff
@@ -0,0 +1,57 @@
...
+package com.microsoft.identity.common.base64
+
+import com.microsoft.identity.common.java.base64.Base64Flags
+import com.microsoft.identity.common.java.base64.IBase64
+
+/**
+ * Base64 implementation based on android.util.Base64
+ **/
+
+class AndroidBase64 : IBase64 {
```
p3dr0rv, line 32: why the wrapper?

common4j/src/test/com/microsoft/identity/common/java/MseberaBase64.kt
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The solution is to exclude Msebera Base64 in Android module (and have them consume the Android Base64 instead).
+ *
+ * NOTE: Base64 implementation in both Android OS and Msebera is a clone of an (old version of) apache commons-codec.
+ * (read: Msebera Base64 should produce the exact same result that Android Base64 returns),
+ *
+ * We would still use Msebera's Base64 in places where Android Base64 is not available.
+ * (e.g. 4j tests, android unit tests, Linux Broker)
+ **/
+class MseberaBase64ForCommon4jTests : IBase64 {
```
melissaahn: The name of this file is 'MsebereBase64", but shouldn't it match the class name? Or does it not matter cause it's a class for testing purposes?

common4j/src/main/com/microsoft/identity/common/java/base64/Base64.kt
```diff
@@ -0,0 +1,202 @@
...
+            } catch (e: ClassNotFoundException) {
+                null
+            } catch (e: RuntimeException){
+                null
+            }
+        }
+
+        private fun tryLoadMseberaBase64InLinux(): IBase64? {
+            return try {
+                Class.forName("com.microsoft.identity.broker.base64.MseberaBase64ForLinux")
```
melissaahn: I see the intent of this class (to be able to use an all-encompassing Base64 Util class across the libraries, without the developer needing to know about the differences in dependencies underneath). But the use of hardcoded class name strings here seems interesting to me, especially when a broker class is being referenced in a common4j class. Do hardcoded class name strings need to be used here because of how the Android tool flags the usage of Msebera?

If not, I'm wondering if it would be possible to not have this consolidated Util class but instead just have each library use the designated class directly:
- Make the common4j, broker4j, testutils Msebera classes have the 'default' access modifier so that they are only accessible from within common4j/broker4j, and common/msal/broker cannot access them
- Keep the AndroidBase64 one the way it is (someone contributing to the Linux broker shouldn't be using a class which has Android in the name)
- Keep the Linux class the way it is

I'm additionally wondering if we could still name each of these classes Base64Util if they reside in different packages.
Alternatively, maybe some sort of Factory + depedency injection solution could be used?

melissaahn (commented):
I have one main comment, plus a question. Also, I think most of the new classes will need a trailing newline.

common4j/src/main/com/microsoft/identity/common/java/base64/Base64.kt
rpdome: @melissaahn
- I removed the msebera library from the entire android dependency chain because it got flagged.
(So msebera is no longer shipped with common4j)

- We still need the Base64 class, so...

In the android route (broker/msal), I sourced it from android.util.base64 (which is part of the OS code)

In linux route, since we're retiring them and there is no report of it being flagged, I keep msebera there.

Let's say you boot up android broker -> The android broker invoke a 4j class -> that 4j class needs to do a Base64 operation -> but  the 4j modules no longer have the msebera base64 implementation -> we'll need to source the implementation from somewhere -> we know that AndroidBase64 is also shipped as part of the final apk (via android common) -> so we use the classloader to "look up" for the class and initialize the object.

So... the answer to your first question is 'yes'.

rpdome: Lmk if that answers the question. We can have a call if preferred.

common4j/src/test/com/microsoft/identity/common/java/MseberaBase64.kt
rpdome: I forgot to rename the file 😅

testutils/src/main/java/com/microsoft/identity/internal/testutils/MseberaBase64.kt
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * The solution is to exclude Msebera Base64 in Android module (and have them consume the Android Base64 instead).
+ *
+ * NOTE: Base64 implementation in both Android OS and Msebera is a clone of an (old version of) apache commons-codec.
+ * (read: Msebera Base64 should produce the exact same result that Android Base64 returns),
+ *
+ * We would still use Msebera's Base64 in places where Android Base64 is not available.
+ * (e.g. 4j tests, android unit tests, Linux Broker)
+ **/
+class MseberaBase64ForTestUtils : IBase64 {
```
melissaahn: This file name also needs to be changed, and newline at the bottom

common4j/src/main/com/microsoft/identity/common/java/base64/Base64.kt
melissaahn: Had an offline discussion; agreed to put these class strings in constants.

melissaahn (approved):
Talked with @rpdome , and reasoning behind base64Util makes sense (other things happening under the hood with other options I was asking about, and it would be better to keep one util class vs multiple we may need to change later).

PR #2390, opened 2024-04-19
Support for multiple tokens in native auth flows

SaurabhMSFT:
This PR and its companion PR (https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2082)  enables multiple access token support for Native Auth. A pair of new method have been added in the AccountState class to allow developer to retrieve different access tokens according to the requested scopes. The way these methods acquire tokens is exactly as same as the one in MSAL.  The old method 'getAccessToken''s functionality hasn't been changed. It still returns the access token matching the default OIDC scope. This method though has been deprecated.
Two classes that have become unused have been removed.


common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -704,8 +709,17 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         // other tenants. Those tokens will be 'sparse', meaning that their AT/RT will not be loaded
         val fullCacheRecord = cacheRecords[0]
 
-        if (accessTokenIsNull(fullCacheRecord)) {
-            throw ServiceException(ErrorStrings.NATIVE_AUTH_NO_ACCESS_TOKEN_FOUND, "No access token found during refresh - user must be signed out.", null)
+        if (!accessTokenIsNull(fullCacheRecord) && silentTokenCommandParameters.scopes == null) {
```
SammyO: Same as other comment, my view on how we can achieve re-usability is to use the existing `acquireTokenSilent()` and can remove this extra added logic.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -667,7 +668,7 @@ class NativeAuthMsalController : BaseNativeAuthController() {
 
         // Convert AcquireTokenNoScopesCommandParameters into SilentTokenCommandParameters,
         // so we can use it in BaseController.getCachedAccountRecord()
-        val silentTokenCommandParameters =
+        var silentTokenCommandParameters =
```
SammyO: Why do we still need this method?

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -930,7 +930,10 @@ protected boolean idTokenIsNull(@NonNull final ICacheRecord cacheRecord,
     }
 
     protected Set<String> addDefaultScopes(@NonNull final TokenCommandParameters commandParameters) {
-        final Set<String> requestScopes = commandParameters.getScopes();
+        return addDefaultScopes(commandParameters.getScopes());
```
SammyO: Can we follow MSAL practice and add the default scopes in the same place as they do? Which is in the controller, not in the interface.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: We don't need this method now. The method  `acquireTokenSilently` depends on `RefreshCommand` class that is present in common project. I cannot move the method `acquireTokenSilently` to `BaseController` in common4j as that will break our dependency pattern. I will see if we can solve this dependency problem.

SammyO: Got it, my bad. I don't think we should move `acquireTokenSilently` to `BaseController`. Let's not change such fundamental MSAL logic. 

At the very least we could try to have our `NativeAuthController` extend `LocalMsalController`, so we can use that `acquireTokenSilent()`, but I think that won't work (or we would've done it already).

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -704,8 +703,17 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         // other tenants. Those tokens will be 'sparse', meaning that their AT/RT will not be loaded
         val fullCacheRecord = cacheRecords[0]
 
-        if (accessTokenIsNull(fullCacheRecord)) {
-            throw ServiceException(ErrorStrings.NATIVE_AUTH_NO_ACCESS_TOKEN_FOUND, "No access token found during refresh - user must be signed out.", null)
+        if (!accessTokenIsNull(fullCacheRecord) && silentTokenCommandParameters.scopes == null) {
```
rmccahill: There may be an unhandled case here - if an access token is not returned, and there are scopes present in the `silentTokenCommandParameters`, we would no longer throw a service exception. Is this the expected behaviour?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -665,13 +665,8 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
-        val silentTokenCommandParameters =
-            CommandUtil.convertAcquireTokenNoFixedScopesCommandParameters(
-                parameters,
-                parameters.getCorrelationId()
-            )
+        //We need to reassign parameters later in the function
+        var silentTokenCommandParameters = parameters
 
         // We want to retrieve all tokens from the cache, regardless of their scopes. Since in the
         // native auth get token flow the developer doesn't have the ability to specify scopes,
```
rmccahill: Is this comment still correct?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
```diff
@@ -36,6 +36,7 @@
 import java.net.URISyntaxException;
```
rmccahill: Do we still need this file if `AcquireTokenNoFixedScopesCommand` is being removed?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -665,13 +665,8 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
         parameters.validate()
 
-        // Convert AcquireTokenNoScopesCommandParameters into SilentTokenCommandParameters,
-        // so we can use it in BaseController.getCachedAccountRecord()
-        val silentTokenCommandParameters =
-            CommandUtil.convertAcquireTokenNoFixedScopesCommandParameters(
-                parameters,
-                parameters.getCorrelationId()
-            )
+        //We need to reassign parameters later in the function
```
SammyO: I'm not sure I understand this comment.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: I had tried extending `NativeAuthController` from `LocalMsalController` during merge to dev branch last year and that didn't work out.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: If we cannot find an access token for the given scopes then we will request a new access token from the server.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: This has been removed. A parameter is always passed as `val` in Kotlin so reassignment is not possible hence one has to make a copy before performing reassignment.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
SaurabhMSFT: The file has been removed.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: No, it has been removed.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
SaurabhMSFT: Done.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/BaseNativeAuthController.kt
```diff
@@ -69,15 +69,6 @@ abstract class BaseNativeAuthController : BaseController() {
         throw ClientException("acquireToken() not supported in NativeAuthController")
     }
```
SammyO, line 71: could you update your PRs to include an overview of the changes?

SammyO (changes requested):

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/BaseNativeAuthController.kt
SaurabhMSFT: Done

rmccahill (approved):

SammyO:
@SaurabhMSFT  can you add a change log?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -670,35 +668,34 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
+            .platformComponents(parametersWithScopes.platformComponents)
             .authenticationScheme(authScheme)
             .build()
-        val strategy = silentTokenCommandParameters.authority.createOAuth2Strategy(strategyParameters)
+        val strategy = parametersWithScopes.authority.createOAuth2Strategy(strategyParameters)
 
-        val tokenCache = silentTokenCommandParameters.oAuth2TokenCache
+        val tokenCache = parametersWithScopes.oAuth2TokenCache
+        var target: String? = null
+        if (parametersWithScopes.scopes != null) {
```
SammyO: This should never happen, right? Looking at browser-delegated, they don't add check but always perform `TextUtils.join(" ", parametersWithScopes.scopes)`. Which I assume will produce an exception if scopes is null. Meaning they will never execute a cache operation with `null`, unlike our flow (if scopes are null, then target is null and cache retrieval will be done with target=null).
Can we change this please?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -726,9 +719,9 @@ class NativeAuthMsalController : BaseNativeAuthController() {
             fullCacheRecord.accessToken != null && fullCacheRecord.accessToken.shouldRefresh()
         ) {
             if (!fullCacheRecord.accessToken.isExpired) {
-                setAcquireTokenResult(acquireTokenSilentResult, silentTokenCommandParameters, cacheRecords)
+                setAcquireTokenResult(acquireTokenSilentResult, parametersWithScopes, cacheRecords)
                 val refreshOnCommand =
-                    RefreshOnCommand(parameters, this.asControllerFactory(), PublicApiId.MSAL_REFRESH_ON)
+                    RefreshOnCommand(parametersWithScopes, this.asControllerFactory(), PublicApiId.MSAL_REFRESH_ON)
```
SammyO, line 721: Have you tested this? I'm asking because this controller factory is new, and I don't believe we've tested our NativeAuthController with it.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -737,8 +730,8 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                     "Access token is expired. Removing from cache..."
                 )
 
-                renewAT(
-                    silentTokenCommandParameters,
+                renewAccessToken(
```
SammyO, line 731: We're not removing anything from the cache. That operation seems to be missing from your changes.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -784,8 +777,8 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                 "Access token is expired. Removing from cache..."
             )
 
-            renewAT(
-                silentTokenCommandParameters,
+            renewAccessToken(
```
SammyO, line 780: Again, the token removal is missing.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
SammyO: Is this still needed, now that default scopes are added in the controller?

SammyO (changes requested):

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
SaurabhMSFT: I have removed this change

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: I have removed it now as the scopes are never null or empty for NativeAuth at this point in code.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: yes, I tested by modifying the sample app and performing different calls to getAccessToken. What I was not able to test was the expiration of token.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: i have added it

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SaurabhMSFT: i have added expired token removal

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -670,35 +668,31 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
-        // so we can use it in BaseController.getCachedAccountRecord()
-        val silentTokenCommandParameters =
-            CommandUtil.convertAcquireTokenNoFixedScopesCommandParameters(
-                parameters,
-                parameters.getCorrelationId()
-            )
+        // Add default scopes
+        val mergedScopes = addDefaultScopes(parameters)
+
+        val parametersWithScopes = parameters
```
SammyO, line 674: SDK result object's correlation ID is null, both in local-validation error scenarios (e.g. passing an empty scope list), as well as API error scenarios (e.g. passing an invalid scope). With success responses the correlation ID seems to be set.

image

SaurabhMSFT: I have set the correlation id to the value from accountstate and added tests for it in msal repo.

SammyO: I don't think that's correct. getAccessToken() is a new flow, and for every new flow we should be setting a new correlation ID.

SaurabhMSFT: In the old code (https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2082/files#diff-1b65b25a41283439e1932bae0a4dc86a788588813709114e41087ffde8b4ca09 AccountState.kt file line no 270), the correlation id for the flow is set to the correlation id of the accountstate object. If we want a new correlation id then it would be change in behaviour for the old getAccessToken() method.

SaurabhMSFT: If we want to change the old behaviour and go with a new correlation id then the empty scope error should return "UNSET" as the correlation id and the invalid scope should return a proper UUID.

SammyO: I see. In that case let's leave it as it is right now. We can always change the behaviour at a later point in time.

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -857,6 +857,7 @@ protected TokenResult performSilentTokenRequest(
             ((MicrosoftTokenRequest) refreshTokenRequest).setClaims(parameters.getClaimsRequestJson());
             ((MicrosoftTokenRequest) refreshTokenRequest).setClientAppName(parameters.getApplicationName());
             ((MicrosoftTokenRequest) refreshTokenRequest).setClientAppVersion(parameters.getApplicationVersion());
+            ((MicrosoftTokenRequest) refreshTokenRequest).setCorrelationId(UUID.fromString(parameters.getCorrelationId()));
```
SammyO: @shahzaibj can you lean in here? @SaurabhMSFT  can provide context if needed.

SaurabhMSFT: The correlation id was set in the old AcquireTokenNoFixedScope Command.  I also see that the correlation id is being set in different subclasses of TokenRequest. This sets the correlation id we have from from account state class for new token retrieval or a token refresh.

SammyO: > This sets the correlation id we have from from account state class for new token retrieval or a token refresh.

True, but it does more than that. It sets the correlation ID for _all_ MicrosoftTokenRequests, including those going through browser-delegated flows. MSAL team might have different requirements. I want to avoid accidental side effects, so let's have it reviewed by their team.

SammyO (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
SaurabhMSFT: This discussion has been moved to PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2413/files

PR #2396, opened 2024-04-26
Clean up native auth common logging

Yuki-YuXin:
### Goal:
- Review our log statements and reduce to only those that are relevant and useful. Especially the logMethodCall should be much reduced. Not every method call is relevant to know about when debugging.
- Double check that the existing logs use string interpolation correctly to avoid leaking sensitive information in the logs.

### Changes summary:
- Every log item has correlation argument. (make sure use getCorrelationId() )
- The [NativeAuthMsalController.kt](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374/files#diff-304daa7418210ef9cd11ee385364a4a6f74256dc627f404e53422ec50f75b282) layer acts as the main log layer. Instead, remove logSessions under OAuth2Configuration, OAuth2Strategy, RequestProvider since the users care more about the outcome of the request than the construction and sending of the request. Apply the rule "only logger unknown error" to NativeAuthMsalController.
- Remove // TODO 
- Add request and result Logger.infoWithObject in interactors.
- infoWithObject(final String tag, final String correlationID, final String message, final ILoggable object) like warnWithObject

### Company PR:
Previous comments for reference: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374
msal: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2071


common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -1520,6 +1520,11 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         oAuth2Strategy: NativeAuthOAuth2Strategy,
         parameters: SignUpStartCommandParameters
     ): SignUpStartApiResult {
+        LogSession.logMethodCall(
```
SammyO, line 1480: @Yuki-YuXin  given that the aim of this PR was to clean up logging, what's the reason for introducing logs here? Did you find that these are helpful breadcrumbs to have when e.g. debugging issues?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/authorities/NativeAuthCIAMAuthority.kt
```diff
@@ -104,7 +104,6 @@ class NativeAuthCIAMAuthority (
             correlationId = null,
             "${TAG}.getChallengeTypesWithDefault"
         )
-        Logger.info(TAG, "Challenge Types passed = $challengeTypes")
```
SammyO, line 107: Why was this removed? I'd say this is helpful debugging information.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthResponseHandler.kt
```diff
@@ -385,7 +385,15 @@ class NativeAuthResponseHandler {
                 response.body,
                 MicrosoftStsTokenResponse::class.java
             )
-            // TODO add safe logging
+
+            Logger.info(TAG, "MicrosoftStsTokenResponse " +
```
SammyO: Why are we putting this back?

SammyO: Right, because there's a TODO there. IMO we don't need this log. Does the browser-delegated flow log something like this?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/ResetPasswordInteractor.kt
```diff
@@ -94,11 +103,29 @@ class ResetPasswordInteractor(
...
+        Logger.infoWithObject(
+            "${TAG}.rawResponseToResetPasswordStartApiResult",
+            apiResponse.correlationId,
+            "rawApiResponse = ",
+            apiResponse
+        )
+
+        val result = apiResponse.toResult()
+
+        Logger.infoWithObject(
```
SammyO, line 121: This is a good idea. Did you test this? Does this not produce too many logs? If you're fine with this then so am I.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/authorities/NativeAuthCIAMAuthority.kt
Yuki-YuXin:
```kotlin
val challengeTypesWithDefault = (challengeTypes ?: emptyList()).plus(listOf(NativeAuthConstants.GrantType.REDIRECT)).distinct().joinToString(" ")
Logger.info(TAG, "Challenge Types used = $challengeTypesWithDefault")
return challengeTypesWithDefault
```
Put it back and use challengeTypesWithDefault

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
Yuki-YuXin: Here are some comments and discussion before in a legacy PR. https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374
One of the comments was raising an idea that remove the logs from OAuth2Strategy, RequestProvider and RequestHandler, but have some logging in the API layer.

SammyO: Gotcha, thanks for reminder me. All good.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthResponseHandler.kt
Yuki-YuXin: The browser-delegated flow log something like this:

 In terms of the TokenResponse of MSAL:
MicrosoftStsOAuth2Strategy:TokenResult:MicrosoftStsTokenResponse INFO {"ext_expires_in":4165,"expires_in":4165,"mResponseReceivedTime":0,"scope":"email openid profile Application.Read.All User.Read","token_type":"Bearer"}

BaseController:TokenResult:MicrosoftStsTokenResponse INFO {"ext_expires_in":4165,"mRefreshTokenAge":"","mSpeRing":"","expires_in":4165,"mResponseReceivedTime":0,"scope":"email openid profile Application.Read.All User.Read","token_type":"Bearer"}

Thus update the contents of the log to include what MSAL displays. But we can remove this log as well because token seems very risky and sensitive to me.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/ResetPasswordInteractor.kt
Yuki-YuXin: I think this change also comes from the idea that remove the logs from OAuth2Strategy, RequestProvider and RequestHandler, but have some logging in the API layer. 
In my opinion and experience, they are pretty useful and direct. Meanwhile, I am wondering if this batch of logs would cause some side effect of the SDK, for example, size?
Does Logger.debug or Logger.verbose help? Currently, we have infoWithObject and warnWithObject.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthResponseHandler.kt
SammyO: Yeah, let's remove this for now until we have a reason to put it back.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/ResetPasswordInteractor.kt
SammyO: It won't affect the binary size, but might affect performance. I doubt it though, and I don't think we need to prematurely optimise for that.

Yuki-YuXin:
Better to merge after multi access token work.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/authorities/NativeAuthCIAMAuthority.kt
```diff
@@ -104,8 +104,9 @@ class NativeAuthCIAMAuthority (
...
-        return (challengeTypes ?: emptyList()).plus(listOf(NativeAuthConstants.GrantType.REDIRECT)).distinct().joinToString(" ")
+        val challengeTypesWithDefault = (challengeTypes ?: emptyList()).plus(listOf(NativeAuthConstants.GrantType.REDIRECT)).distinct().joinToString(" ")
```
GBakolas: Nit: is it possible to simplify this expression, its a little confusing

common4j/src/main/com/microsoft/identity/common/java/nativeauth/authorities/NativeAuthCIAMAuthority.kt
```diff
@@ -104,8 +104,9 @@ class NativeAuthCIAMAuthority (
...
+        Logger.info(TAG, "Challenge Types used = $challengeTypesWithDefault")
```
GBakolas: Missing brackets around challengeTypesWithDefault

common4j/src/main/com/microsoft/identity/common/java/nativeauth/authorities/NativeAuthCIAMAuthority.kt
Yuki-YuXin: Use `challengeTypes.orEmpty() ` instead of `challengeTypes ?: emptyList()`. They are functionally equivalent. They returns an empty list if the original list is null.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/AcquireTokenNoFixedScopesCommand.kt
GBakolas, line 1: why we have this empty file here? Probably this needs to be deleted

Yuki-YuXin: This file is deleted when multiple access token work. I think it's because of the merging update. Nice catch.

GBakolas (commented):
Nit: In the ApiResponse files where you have removed the calls to logMethodCall() there are the LogSession import statements left. These statements are unused now so they can be removed for code cleanup

rmccahill (approved):

GBakolas (approved):

PR #2398, opened 2024-04-29
Update logging of CommandParameters

SammyO:
- Use different logic to log CommandParameters in `CommandDispatcher` in the case of native auth command parameters.
- Update native auth CommandParameters to use the new safe-logging framework `ILoggable`.


Yuki-YuXin (approved):
If username under SignInStartCommandParameters,  userattributes under SignUpStartCommandParameters are not sensitive info under allowPii=true mode, and challengeTypes can be logged under allowPii=false mode, then I am fine with the PR and have double checked the logging.

SammyO:
> If username under SignInStartCommandParameters, userattributes under SignUpStartCommandParameters are not sensitive info under allowPii=true mode, and challengeTypes can be logged under allowPii=false mode, then I am fine with the PR and have double checked the logging.

Good points.
- MSAL logs loginHint when allowPII=true, which I would consider the same as username. So I'm following existing practices.
- Given the above, I don't think userAttributes is more PII than the previous. So I'm following existing practices here as well
- challengeType is a developer config, like client ID and authority. It's not PII. Similarly, existing MSAL logs the authority value, regardless of PII setting.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -472,10 +474,14 @@ private static void logParameters(@NonNull String tag, @NonNull String correlati
                         + DiagnosticContext.INSTANCE.getRequestContext().toJsonString()
                         + ", with PublicApiId: " + publicApiId);
 
-        if (Logger.isAllowPii()) {
-            Logger.infoPII(TAG, ObjectMapper.serializeObjectToJsonString(parameters));
+        if (parameters instanceof BaseNativeAuthCommandParameters) {
```
SaurabhMSFT: This check will introduce NativeAuth types to this class and all of its dependencies. All consumers of this library except NativeAuth have set up a proguard rule to exclude NativeAuth types. Due to the introduction of this type, all those consumers will now have to include NativeAuth byte code in their library. Their binary size will increase and they may not like that. If possible we should move this check to some other location where this problem will not occur.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/ResetPasswordStartCommandParameters.java
```diff
@@ -22,6 +22,8 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.java.nativeauth.commands.parameters;
 
+import org.jetbrains.annotations.NotNull;
```
SaurabhMSFT: use `NonNull` from  `lombok.NonNull;`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/ResetPasswordSubmitCodeCommandParameters.java
```diff
@@ -22,6 +22,8 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.java.nativeauth.commands.parameters;
 
+import org.jetbrains.annotations.NotNull;
```
SaurabhMSFT: use `NonNull` from `lombok.NonNull;`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/ResetPasswordSubmitNewPasswordCommandParameters.java
```diff
@@ -22,6 +22,8 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.java.nativeauth.commands.parameters;
 
+import org.jetbrains.annotations.NotNull;
```
SaurabhMSFT: use `NonNull` from `lombok.NonNull;`

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
SammyO: @SaurabhMSFT  good point, let me look into this.
@shahzaibj @fadidurah  what size checks can I run to show that Saurabh is right in this observation?

shahzaibj: @SammyO Can we update the check to be like the following:

```java
if (parameters instanceof ILoggable)
```

This should avoid anyone taking dependency on native auth

SammyO: Good point @SaurabhMSFT.

@fadidurah  @shahzaibj could you review the changes I made to address this? `logParameter` logic is now delegated to the CommandParameter classes.

SammyO: @shahzaibj sorry I missed your comment. I went in a different direction. Let me know if you're happy with this as well or whether I should revert to your suggestion.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
```diff
@@ -57,16 +59,27 @@ public class AcquireTokenNoFixedScopesCommandParameters extends BaseNativeAuthCo
...
+    @Override
+    public String toUnsanitizedString() {
+        return "AcquireTokenNoFixedScopesCommandParameters(account=" + account + ", authenticationScheme=" + getAuthenticationScheme() + ", forceRefresh=" + forceRefresh + ", authority=" + authority + ", challengeTypes=" + challengeType + ")";
+    }
 
-    private final List<Map.Entry<String, String>> extraOptions;
+    @Override
+    public boolean containsPii() {
+        return !toString().equals(toUnsanitizedString());
+    }
```
shahzaibj, line 74: This implementation can probably move to `ILoggable` itself and then you won't have to override it everywhere

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
shahzaibj: I think either approach is fine

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
shahzaibj: Looks you have defined it there already so not sure why there is a need to override in so many places

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
SammyO: Java doesn't play nice with Kotlin (because it's a Kotlin interface with a default function). 
1. not overriding the method at all gives compilation error `Class 'AcquireTokenNoFixedScopesCommandParameters' must either be declared abstract or implement abstract method 'containsPii()' in 'ILoggable'`
2. overriding it and performing `return super.containsPii();` gives build error `error: abstract method containsPii() in ILoggable cannot be accessed directly`.

Fixing this properly would require `-Xjvm-default=all` compilation flag, which I would like to leave out of scope given the urgency of this PR.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/ResetPasswordSubmitNewPasswordCommandParameters.java
SammyO: Updated (everywhere).

PR #2401, opened 2024-05-05
Update logic for matching requested claims for AT

sansrivastav:
Current logic of matching credentials only compares requested claims when we specify requested claims.
So, if there are 2 accesss tokens in storage:
AT1 - with no requested claims specified
AT2 - with a specific requested claims RC2

And then we query the storage without specifying any claims, current logic returns both AT1 and AT2.

The proposed change overrides _getCredentialsFilteredBy()_ adding an extra boolean _mustMatchExactClaims_. When this parameter is set, we match claims explicitly even when no claims are provided. In the case above, only AT1 will be returned.
We will call this new overloaded method from MSAL CPP so that the behavior of MSAL CPP is same across all platforms including Android.


common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -164,7 +164,7 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
                 && !StringUtil.isNullOrEmpty(authScheme)
                 && credentialType == CredentialType.AccessToken_With_AuthScheme;
         final boolean mustMatchOnKid = !StringUtil.isNullOrEmpty(kid);
-        final boolean mustMatchOnRequestedClaims = !StringUtil.isNullOrEmpty(requestedClaims);
+        final boolean mustMatchOnRequestedClaims = null != requestedClaims;
```
dorinung: So the problem was that ATs with "" no claims where not matched. Is null not also semantically empty? Are there any cases where we need to return all all ATs that differ only in claims? I don;t think MSAL needs this.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -151,7 +151,8 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
                                                                 @Nullable final String target,
                                                                 @Nullable final String authScheme,
                                                                 @Nullable final String requestedClaims,
-                                                                @Nullable final String kid) {
+                                                                @Nullable final String kid,
+                                                                @Nullable Boolean mustMatchExactClaims) {
```
nickbopp: >@Nullable Boolean [](http://example.com/codeflow?start=64&length=17)

I'm looking at this and the new line on 170 -- should this just be a primitive lowercase `boolean`? #Closed

sansrivastav: done in latest iteration

nickbopp (approved):
:shipit:

common/src/test/java/com/microsoft/identity/common/SharedPreferencesAccountCredentialCacheTest.java
```diff
@@ -1401,6 +1401,127 @@ public void getCorrectCredentialWhenRequestedClaimsAreSpecified() {
...
+                HOME_ACCOUNT_ID,
+                ENVIRONMENT,
+                CredentialType.AccessToken,
+                CLIENT_ID,
+                APPLICATION_IDENTIFIER_SHA512,
+                MAM_ENROLLMENT_IDENTIFIER,
+                REALM,
+                null,
+                BEARER_AUTHENTICATION_SCHEME.getName(),
+                null,
```
mohitc1, line 1556: >null [](http://example.com/codeflow?start=16&length=4)

What's expected if requestedClaims is "" instead null? Should it behave same as null? Seems currently they would behave differently.

sansrivastav: Thanks for pointing this out. Made changes to handle both null and empty as same and added test case for it.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -271,10 +272,12 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
                 matches = matches && kid.equalsIgnoreCase(accessToken.getKid());
             }
 
-            if (mustMatchOnRequestedClaims) {
+            if (mustMatchOnRequestedClaims || mustMatchExactClaims) {
```
mohitc1, line 275: >mustMatchOnRequestedClaims || mustMatchExactClaims [](http://example.com/codeflow?start=16&length=50)

I think it would be simpler code if we can have just one boolean defined outside loop
boolean mustMatchClaims = mustMatchExactClaims || mustMatchOnRequestedClaims;

and here
```
if (mustMatchClaims) {
    matches = matches && 
    ((StringUtil.isNullOrEmpty(requestedClaims) && StringUtil.isNullOrEmpty(accessToken.getRequestedClaims())) // match due to emptyness
     || (StringUtil.equalsIgnoreCaseTrimBoth(requestedClaims, accessToken.getRequestedClaims()))
     );
}
```

mohitc1 (approved):
Left a suggestion.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
sansrivastav: I am trying to keep the new changes independent of existing changes as much as possible. This logic is used in a lot of places and by keeping the existing logic as it is in the existing flows is one of the reasons I did not optimize the flow here. Let me reach out to you offline, if you still feel strongly about this.

sansrivastav: Closing after offline discussion.

PR #2406, opened 2024-05-14
UI Maintenance, May 2024

fadidurah:
- Removed UncaughtExceptionHandlerRule, seems to be broken and causing some tests to run forever. I tried some fixes, but no luck keeping the rule and also fixing tests.
- Fix parameter in lab api for ONPREM
- Fix Cert in Common Testing utilities pipeline
- Adjustments to Chrome HandleFirstRun
- Add fallback in prompt handling methods to check for object by class, if resource ids are not working (saw some issues on local device where chrome wouldn't export resource ids, also appeared to happen in webview sometimes. Seemed to be happening constantly in ADFS (although this seems like an issue from ests side)). Either side, not a blocker for us with this fix.


melissaahn (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AdfsLoginComponentHandler.java
```diff
@@ -38,15 +38,25 @@ public class AdfsLoginComponentHandler extends AadLoginComponentHandler {
 
     @Override
     public void handleEmailField(@NonNull final String username) {
-        final UiObject2 usernameObject = UiAutomatorUtils.obtainAllEditTextObjects(CommonUtils.FIND_UI_ELEMENT_TIMEOUT).get(0);
-        usernameObject.setText(username);
+        try {
+            UiAutomatorUtils.handleInput("userNameInput", username, CommonUtils.FIND_UI_ELEMENT_TIMEOUT_SHORT);
+        } catch (AssertionError e) {
+            final UiObject2 usernameObject = UiAutomatorUtils.obtainAllEditTextObjects(CommonUtils.FIND_UI_ELEMENT_TIMEOUT).get(0);
```
rpdome, line 45: nit: please add comment in all catch blocks - which scenario we're expecting this error.

rpdome (approved):

PR #2413, opened 2024-05-23
Setting the correlation id for token request command

SaurabhMSFT:
This change has been moved out of PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2390 . NativeAuth is now using the SilentTokenCommand for requesting new tokens and performing token refresh. The correlation id from the account state object is used for sending the request.


common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -857,6 +857,7 @@ protected TokenResult performSilentTokenRequest(
             ((MicrosoftTokenRequest) refreshTokenRequest).setClaims(parameters.getClaimsRequestJson());
             ((MicrosoftTokenRequest) refreshTokenRequest).setClientAppName(parameters.getApplicationName());
             ((MicrosoftTokenRequest) refreshTokenRequest).setClientAppVersion(parameters.getApplicationVersion());
+            ((MicrosoftTokenRequest) refreshTokenRequest).setClientAppVersion(parameters.getCorrelationId());
```
SaurabhMSFT: @SammyO @shahzaibj @fadidurah Discussion from the old PR in below screenshot.
image

shahzaibj: Why are we setting correlation id as client app version?

shahzaibj (changes requested):
See comments

common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
SammyO: That's a mistake, let me update this as Saurabh is OOO.

SammyO: Updated. @shahzaibj please re-review. Specifically: 
- if the correlation ID in the parameter object isn't set, this will explicitly set the correlation ID in the request to null. Is that a problem? Is there another place in the code that checks whether the correlation ID is null and if so sets it to something (e.g. the API layer)?
- are there other unwanted side-effects of this change?

SammyO: @SaurabhMSFT did you test this? In my tests the request is not an instance of `MicrosoftTokenRequest`, so this code is not executed.

PR #2419, opened 2024-05-30
Adding check for OS version for passkeys

melissaahn:
### Summary
According to the Credential Manager documentation, passkeys are only available for devices running on Android 9 (API 28) and above. I'm adding a check before most of the logic in AuthFidoChallengeHandler so that we can swiftly end the logic and return a message to the server.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.internal.fido
 
+import android.os.Build
 import android.webkit.WebView
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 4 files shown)


fadidurah (approved):

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
```diff
@@ -69,6 +70,18 @@ class AuthFidoChallengeHandler (
         // If either one of these are missing or malformed, throw an exception and let the main WebViewClient handle it.
         val submitUrl = fidoChallenge.submitUrl.getOrThrow()
         val context = fidoChallenge.context.getOrThrow()
+        // Check the OS version as well. As of the time this is written, passkeys are only supported on devices that run Android 9 (API 28) or higher.
+        // https://developer.android.com/identity/sign-in/credential-manager
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
+            respondToChallengeWithError(
```
shahzaibj, line 76: So this error is sent back to the server? or is this presented to the user instead?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/AuthFidoChallengeHandler.kt
melissaahn: The error gets sent back to the server, and it's logged on our side but the user doesn't see the message. On the user's end, they will just see the page redirecting back to the login or password page when we send any sort of exception. 
We are working with the server side wrt UI for error scenarios.

PR #2420, opened 2024-05-31
Transfer token generation changes

somalaya:
Related broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/2821


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1369,6 +1369,21 @@ public static String computeMaxHostBrokerProtocol() {
...
 
+        /**
+         * Boolean to return when a transfer token is successfully saved.
+         */
+        public static final String SAVE_TRANSFER_TOKEN_SUCCEEDED = "save_transfer_token_succeeded";
+
+        /**
+         * The Bundle key name of transfer token.
+         */
+        public static final String TRANSFER_TOKEN = "transfer_token_value";
```
shahzaibj: So this is the key or the value? The actual content seems to call it value but javadoc says key

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -588,6 +588,11 @@ public static final class OAuth2Scopes {
          */
         public static final String EMAIL_SCOPE = "email";
 
+        /**
+         * Scope to get transfer token
+         */
+        public static final String TRANSFER_TOKEN_SCOPE = "service::http://Passport.NET/purpose::PURPOSE_TRANSFER_TOKEN_BACKUP_RESTORE";
```
shahzaibj, line 594: What is Passport.NET?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -155,7 +156,7 @@ public MicrosoftStsOAuth2Strategy(@NonNull final MicrosoftStsOAuth2Configuration
                 } else {
                     setTokenEndpoint(config.getTokenEndpoint().toString());
                 }
-            }  catch (ServiceException e) {
+            } catch (ServiceException e) {
```
shahzaibj:
```suggestion
            } catch (final ServiceException e) {
```

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
somalaya: This is bundle key only.  We are passing transfer token value using this key. Let me just call it TRANSFER_TOKEN_BUNDLE_KEY to avoid any confusion.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
somalaya: Not sure what Passport.NET signifies. But this is the custom scope required for getting transfer token instead of an access token using PRT. As per MSA server side spec https://msazure.visualstudio.com/One/_git/ESTS-Docs/pullrequest/7258879?_a=files&path=/TransferToken/BackupAndRestore.md

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -58,5 +58,10 @@ public enum SpanName {
     BrokerDiscoveryV1ProtocolBroadcastResult,
     Fido,
     BrokerAccountServiceRemoveAccounts,
-    AcquirePRTUsingTransferToken
+
+    AcquirePRTUsingTransferToken,
+
```
shahzaibj: Is there a reason for leaving empty line here and below?

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
mohitc1, line 594: If this is only ever going to be used in broker, can this be moved in broker4j/broker?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -58,5 +58,10 @@ public enum SpanName {
     BrokerDiscoveryV1ProtocolBroadcastResult,
     Fido,
     BrokerAccountServiceRemoveAccounts,
-    AcquirePRTUsingTransferToken
+
+    AcquirePRTUsingTransferToken,
```
mohitc1: >PRT [](http://example.com/codeflow?start=11&length=3)

Nit: AcquirePrtUsingTransferToken

mohitc1 (approved):
Left minor suggestions

PR #2424, opened 2024-06-06
Provide empty image when no video is present on QR +PIN auth, Fixes AB#3019283

p3dr0rv:
When QR + PIN auth is enabled, for a glimpse of a second you can see the default image.
image

The preferred experience is not to show the default image at all, so we are providing an empty image instead

[AB#3019283](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3019283)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -281,6 +282,17 @@ public void onPermissionRequest(final PermissionRequest request) {
...
+
+            @Override
+            public Bitmap getDefaultVideoPoster() {
+                // When not playing, video elements are represented by a 'poster' image.
+                // The image to use can be specified by the poster attribute of the video tag in HTML.
+                // If the attribute is absent, then a default poster will be used.
+                // This method allows the ChromeClient to provide that default image.
+                // On our case when an invalid QR code is provided,  we will return a 10x10 empty image.
+                // Instead of the default grey playback image. #2424
+                return Bitmap.createBitmap(10, 10, Bitmap.Config.ARGB_8888);
```
shahzaibj, line 312: TBH, the empty whitespace as you've shown in the AFTER screenshot looks weird too.

shahzaibj: Rather than returning an empty image, can we work with PM to create a dedicated ERROR image that renders some meaningful error img/text such as `Oops something went wrong :(` and then just use that?

p3dr0rv: Waiting for PM to decide what they want to do here.

p3dr0rv: ESTS UX made the fix and now the error page changed.  but still, we need to remove this image because this image is shown for a second before loading the camera. which looks odd.

rpdome (approved):

fadidurah (approved):

PR #2426, opened 2024-06-07
Platform Specific Extra Query Parameters

melissaahn:
### Summary
Adding a method to IPlatformUtil `setPlatformSpecificExtraQueryParameters`. This method takes in an existing list of parameters or null and returns a list of parameters which has been edited based on the platform, or null (if the existing list was null).
An example of a query string parameter which benefits from this method is the passkey parameter, which only needs to be sent for Android broker and not Linux broker.

This is going to make the consumer checks fail at the broker stage, but I'll run them all once to make sure that the other libraries don't run into issues.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
@@ -45,17 +45,22 @@ import com.microsoft.identity.common.internal.ui.webview.WebViewUtil;
 import com.microsoft.identity.common.java.commands.ICommand;
 import com.microsoft.identity.common.java.commands.InteractiveTokenCommand;
 import com.microsoft.identity.common.java.commands.parameters.InteractiveTokenCommandParameters;
+import com.microsoft.identity.common.java.constants.FidoConstants;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.exception.ErrorStrings;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightManager;
 import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.ui.BrowserDescriptor;
 import com.microsoft.identity.common.java.util.IPlatformUtil;
 import com.microsoft.identity.common.java.util.StringUtil;
 
 import java.security.NoSuchAlgorithmException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 
 import javax.net.ssl.KeyManagerFactory;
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -213,6 +218,23 @@ public String getPackageNameFromUid(int uid) {
         return mContext.getPackageManager().getNameForUid(uid);
     }
 
+    @Override
+    public List<Map.Entry<String, String>> setPlatformSpecificExtraQueryParameters(@Nullable List<Map.Entry<String, String>> originalList) {
+        List<Map.Entry<String, String>> queryParams = originalList != null ?  new ArrayList<>(originalList) : new ArrayList<>();
+
+        // Passkey feature support is only for Android at the moment.
+        final  Map.Entry<String, String> webauthnParam = new AbstractMap.SimpleEntry<>(FidoConstants.WEBAUTHN_QUERY_PARAMETER_FIELD, FidoConstants.WEBAUTHN_QUERY_PARAMETER_VALUE);
+        if (CommonFlightManager.isFlightEnabled(CommonFlight.ENABLE_PASSKEY_FEATURE)) {
```
melissaahn, line 227: Note that this check is going to be removed once the broker flight is finished and the passkey feature is stable. The eventual intent is that this query string is always sent for Android broker.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -213,6 +218,23 @@ public String getPackageNameFromUid(int uid) {
...
+            if (!queryParams.contains(webauthnParam)) {
+                queryParams.add(webauthnParam);
+            }
+        } else {
+            // If we don't want to add this query string param, then we should also remove other instances of it that might be already present from MSAL/OneAuth-MSAL.
```
shahzaibj, line 232: Does this mean we are overriding flights set by client apps?

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -213,6 +218,23 @@ public String getPackageNameFromUid(int uid) {
         return mContext.getPackageManager().getNameForUid(uid);
     }
 
+    @Override
+    public List<Map.Entry<String, String>> setPlatformSpecificExtraQueryParameters(@Nullable List<Map.Entry<String, String>> originalList) {
```
shahzaibj: Generally the purpose of a `setter` is to set or modify a value within `this` class i.e. a setter that lives on class `AndroidPlatformUtil` may modify a member variable of that class. We probably need a better name for this method, Maybe call it `updateAndGetQueryParamsWithPlatformSpecificQp`

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
tanmaymanolkar1: this is something even I am confused about, for the Linux case of supporting FIDO I checked with MSAL.CPP team, they do not pass the webauthn=1 parameter

melissaahn: > Does this mean we are overriding flights set by client apps?

Yes. Sending webauthn=1 declares that your app supports/has done the necessary work to support passkeys. A brokered request should only send the parameter if the brokered apps themselves are ready (which we know well). So if a 1P app sends the parameter, which is declaring for itself, and it goes through broker, but broker isn't ready yet, we don't want to send that parameter to the server side.


> this is something even I am confused about, for the Linux case of supporting FIDO I checked with MSAL.CPP team, they do not pass the webauthn=1 parameter

OneAuth-MSAL Android has a configuration for sending the webauthn parameter, and the control is with the individual 1P apps. Linux on the MSAL.CPP side shouldn't have logic for sending this.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
melissaahn: Maybe updateWithAndGetPlatformSpecificExtraQueryParameters? 4e76e7f2bd318e2245fb36dd47b7494e23c4a845

tanmaymanolkar1 (approved):

shahzaibj (approved):

melissaahn:
Going to disable consumers check because a fail for broker is expected. See https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1312023&view=results for all other libraries passing.

PR #2427, opened 2024-06-10
Fix AccountManager Strategy

rpdome:
Add the check to make sure that this strategy is only used when the targeted broker and the "com.microsoft.workaccount"-associated broker are the same one.


p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/AccountManagerAddAccountStrategy.java
```diff
@@ -101,9 +105,16 @@ public Type getType() {
 
     @Override
     public boolean isSupportedByTargetedBroker(@NonNull final String targetedBrokerPackageName) {
```
moumighosh, line 104: What does this code intending to do? In what scenario would this get invoked?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/AccountManagerAddAccountStrategy.java
rpdome: This is a backup IPC mechanism between MSAL and Broker, which was introduced before we discovered ContentProvider.

Currently, it's the latest backup option (only used when everything else fails). From what we heard from OneAuth, this covers < 0.5% of the overall traffic

PR #2430, opened 2024-06-13
Support Passing x-client-xtra-sku header to ESTS only for brokered requests

fadidurah:
Feature Spec: https://microsoft-my.sharepoint-df.com/:w:/g/personal/oldalton_microsoft_com1/ET9igW1SaHJLmlNIBi_CbqoBuyAKPWIZ0BpzcxTUB8yeLw?e=xmY1xe&nav=eyJoIjoiMTk3NDU4ODU3MyJ9&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cfadidurah%40microsoft.com&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNDA1MzEwMTQwNyIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D

Changes to enable passing in the new x-client-xtra-sku header to ests when calling /authorize and /token endpoints. This field has a standard schema agreed upon by all teams JavaDoc of `ClientExtraSkuAdapter.java` explains the schema of this new field, most of the fields in this new file.

To facilitate this, I'll need to propogate Product and Porduct version through the authorization workflow, as it runs on a new thread, and the DiagnosticContext wasn't being populated. I've updated the bundle to pass in these values so we can use them in the telemetry field, also used them to populate the diagnostic context for good measure.

For /Token endpoint, I noticed that in some cases the broker version wasn't being passed, even if it was a brokered scenario. This didn't seem to be by design, so i updated it to add the broker version in these flows. Please reference Broker PR (https://github.com/AzureAD/ad-accounts-for-android/pull/2832). This is important as i'm using the presence of the broker version as a check for sending in the new telemetry field, as we only want to send it in brokered scenarios.

Validated this in interactive and silent requests for the following apps:
OneAuth (MSA account)
MsalTestApp
AdalTestApp


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -77,7 +79,35 @@ public static Intent getAuthorizationActivityIntent(final Context context,
...
+     * @param productVersion             Product version to be of library making the request
+     * @return An android Intent which will be used by Android to create an AuthorizationActivity
+     */
+    public static Intent getAuthorizationActivityIntent(final Context context,
+                                                        final Intent authIntent,
+                                                        final String requestUrl,
+                                                        final String redirectUri,
+                                                        final HashMap<String, String> requestHeaders,
+                                                        final AuthorizationAgent authorizationAgent,
+                                                        final boolean webViewZoomEnabled,
```
moumighosh, line 106: What is webView zoom?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -49,6 +49,8 @@
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.REQUEST_URL;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_CONTROLS_ENABLED;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.AuthorizationIntentKey.WEB_VIEW_ZOOM_ENABLED;
+import static com.microsoft.identity.common.java.AuthenticationConstants.SdkPlatformFields.PRODUCT;
```
moumighosh, line 52: What does PRODUCT imply here, the various OS platforms? Or the brokers on each platform?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -77,7 +79,35 @@ public static Intent getAuthorizationActivityIntent(final Context context,
...
                                                         final boolean webViewZoomControlsEnabled) {
-        Intent intent;
+        return getAuthorizationActivityIntent(context, authIntent, requestUrl, redirectUri, requestHeaders, authorizationAgent, webViewZoomEnabled, webViewZoomControlsEnabled, null, null);
+    }
+
+    /**
+     * Return the correct authorization activity based on library configuration.
+     *
+     * @param context                    Android application context
+     * @param authIntent                 Android intent used by the authorization activity to launch the specific implementation of authorization (BROWSER, EMBEDDED)
```
moumighosh, line 89: Does eSTS capture this information today, if I am using MSAL (no broker) and the authorization intent is BROWSER? Is that data present in eSTS telemetry today?

shahzaibj: ESTS knows about Broker or not today and also knowns about authorization agent.

common4j/src/main/com/microsoft/identity/common/java/logging/ProductHelper.java
```diff
@@ -0,0 +1,68 @@
...
+public class ProductHelper {
+    private static final String TAG = ProductHelper.class.getSimpleName();
+
+    /**
+     * The String to be returned if the value is not set.
+     */
+    protected static final String NOT_SET = "NOT_SET";
+
+    @NonNull
+    public static String getProduct() {
```
shahzaibj: Javadoc on public methods

common4j/src/main/com/microsoft/identity/common/java/util/ClientExtraSkuAdapter.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Builder.Default
+    private String mBrowserExtVer = "";
+
+    @Builder.Default
+    private String mBrowserCoreVer = "";
+
+    public String toString(){
+        final StringBuilder stringBuilder = new StringBuilder();
+        // Index 0 - Auth SDK
```
shahzaibj: I'm not sure I follow what Index means. Is index determined by pipe separation? So after every pipe index is incremented? or something else?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
shahzaibj: This was added long back to allow MSAL customers to toggle whether zoom controls should be enabled in webview or not.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
fadidurah: SKU Name of the library making the request

common4j/src/main/com/microsoft/identity/common/java/util/ClientExtraSkuAdapter.java
fadidurah: indexes are separated by comma, each element in the index are separated by pipes

fadidurah: I linked the feature spec in the PR description

common4j/src/main/com/microsoft/identity/common/java/util/ClientExtraSkuAdapter.java
```diff
@@ -0,0 +1,96 @@
...
+ *     <li>Index 2 - Browser extension name and version</li>
+ *     <ul><li>Example: Chrome|1.0.7</li></ul>
+ *     <li>Index 3 -  Browser core version. Browser core name is omitted as static</li>
+ *     <ul><li>Example: |2.5.7</li></ul>
+ * </ul>
+ * Example output where no MSAL.Runtime or Browser Core is used: MSAL.Android|5.4.0,|,Chrome|1.0.7,|
+ */
+@Builder
+@Accessors(prefix = "m")
+public class ClientExtraSkuAdapter {
```
shahzaibj: I'm not sure I understand why this is an adapter. I believe you can just call this class `ClientExtraSku` because that's what it represents and omit the word adapter

shahzaibj (commented):
Can we add some tests for all the new code added here?

common4j/src/main/com/microsoft/identity/common/java/util/ClientExtraSkuAdapter.java
fadidurah: Sure thing, i'll change the name. I just called it an adapted because i see it as adapting multiple fields into one schema, i guess the word adapter doesn't fit with our current usage of adapter today.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -456,12 +460,29 @@ private HashMap<String, String> getRequestHeaders(final Bundle state) {
                 }
                 requestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_VALUE);
             }
+
+            // Attach client extras header for ESTS telemetry. Only done for broker requests
+            if (checkIfBrokerRequest(this.mAuthorizationRequestUrl)) {
+                final ClientExtraSku clientExtraAdapter = ClientExtraSku.builder()
```
shahzaibj:
```suggestion
                final ClientExtraSku clientExtraSku = ClientExtraSku.builder()
```

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -610,6 +610,11 @@ public static final class SdkPlatformFields {
          */
         public static final String VERSION = "x-client-Ver";
 
+        /**
+         * The String representing client extra telemetry information to be sent to ESTS.
+         */
+        public static final String CLIENT_EXTRAS = "x-client-xtra-sku";
```
shahzaibj:
```suggestion
        public static final String CLIENT_EXTRA_SKU = "x-client-xtra-sku";
```

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
```diff
@@ -201,17 +203,26 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
...
+
         if (request instanceof MicrosoftTokenRequest &&
                 !StringUtil.isNullOrEmpty(((MicrosoftTokenRequest) request).getBrokerVersion())) {
             headers.put(
                     Device.PlatformIdParameters.BROKER_VERSION,
                     ((MicrosoftTokenRequest) request).getBrokerVersion()
             );
+
+            // Attach client extras header for ESTS telemetry. Only done for broker requests
+            final ClientExtraSku clientExtraAdapter = ClientExtraSku.builder()
```
shahzaibj:
```suggestion
            final ClientExtraSku clientExtraSku = ClientExtraSku.builder()
```

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
```diff
@@ -77,7 +79,35 @@ public static Intent getAuthorizationActivityIntent(final Context context,
...
+     * @param requestUrl                 The authorization request in URL format
+     * @param redirectUri                The expected redirect URI associated with the authorization request
+     * @param requestHeaders             Additional HTTP headers included with the authorization request
+     * @param authorizationAgent         The means by which authorization should be performed (EMBEDDED, WEBVIEW) NOTE: This should move to library configuration
+     * @param webViewZoomEnabled         This parameter is specific to embedded and controls whether webview zoom is enabled... NOTE: Needs refactoring
+     * @param webViewZoomControlsEnabled This parameter is specific to embedded and controls whether webview zoom controls are enabled... NOTE: Needs refactoring
+     * @param product                    Product name to be of library making the request
```
rpdome: nit: sdkName, sdkVersion?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java
```diff
@@ -155,17 +157,19 @@ user commit() rather than commitNow() that the fragment manager that we were rem
...
     /**
      * When authorization fragment is launched.  It will be launched on a new thread. (TODO: verify this)
      * Initialize based on value provided in intent.
      */
-    private static void setDiagnosticContextForNewThread(final String correlationId) {
+    private static void setDiagnosticContextForNewThread(final String correlationId, final String product, final String productVersion) {
         final String methodTag = TAG + ":setDiagnosticContextForAuthorizationActivity";
         final RequestContext rc = new RequestContext();
         rc.put(DiagnosticContext.CORRELATION_ID, correlationId);
+        rc.put(PRODUCT, product);
```
rpdome: do we need this?

rpdome: DiagnosticContext is a global(static) object **_which could be modified by concurrent threads_**, and we should avoid using it as much as possible.

IIRC, today we use it for thread id and correlation id to avoid passing those value as a parameter every time we invoke a log function. 

Given that we invoke log() everywhere, the original designer of this class was probably okay with the trade off (that we might print out wrong value due to race condition)

common4j/src/main/com/microsoft/identity/common/java/logging/ProductHelper.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Returns the product (library) name by accessing it from DiagnosticContext.
+     * @return product name from DiagnosticContext.
+     */
+    @NonNull
+    public static String getProduct() {
+        final String methodName = ":getProduct";
+
+        final String product = DiagnosticContext.INSTANCE.getRequestContext().get(AuthenticationConstants.SdkPlatformFields.PRODUCT);
```
rpdome: maybe do this like the device class?

force set it in the platform component class *once*, under the sGlobalStateInitalized block.

After that, the values would never be modified by any threads.

common4j/src/main/com/microsoft/identity/common/java/logging/ProductHelper.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.microsoft.identity.common.java.AuthenticationConstants;
+import com.microsoft.identity.common.java.platform.Device;
+import com.microsoft.identity.common.java.util.StringUtil;
+
+import cz.msebera.httpclient.android.BuildConfig;
+import lombok.NonNull;
+
+/**
+ * Simple helper class to pull product telemetry information from diagnostic context.
```
rpdome: give more details on what you mean by product.

Broker? MSAL?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
rpdome: If we're making a request from OneAuth -> Broker, what is the value we are sending here? OneAuth or Broker?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -84,15 +85,26 @@ public Future<AuthorizationResult> requestAuthorization(GenericAuthorizationRequ
...
         final URI requestUrl = authorizationRequest.getAuthorizationRequestAsHttpRequest();
-        final Intent authIntent = buildAuthorizationActivityStartIntent(requestUrl);
+
+        String product = null;
+        String productVersion = null;
+
+        // TODO: Pretty sure this type check wouldn't be necessary, since all authorization requests that go through this path will
+        //  extend MicrosoftAuthorizationRequest. Perhaps we can change the Generic type to extend MicrosoftAuthorizationRequest rather than just AuthorizationRequest.
+        if (mAuthorizationRequest instanceof MicrosoftAuthorizationRequest) {
+            product = ((MicrosoftAuthorizationRequest) mAuthorizationRequest).getLibraryName();
```
rpdome, line 95: where are we setting the "source" of these values?

Unless i'm mistaken, i'm seeing a loop
1. This method relies on the name/version value in MicrosoftAuthorizationRequest
2. MicrosoftAuthorizationRequest relies on ProductHelper
3. ProductHelper relies on DiagnosticContext
4. DiagnosticContext values are set in setDiagnosticContextForNewThread

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
rpdome: If it's the former - we upload the "origin" of the request, 
then such origin should be part of the Request object that we construct in the IPC layer. We should not need to use DiagnosticContext

if it's the latter - we upload the info of whoever makes the request,
then we could use the static "unmodifiable" model like the Device class.

rpdome: If it's the former, we could set these values when we construct TokenCommandParameters in Msal/AdalAndroidBrokerCommandParameters, then we propagate these values as we construct the request object.

If it's the latter, you could set it in the request class with the your Product class. (but modify the product class to be set "one time only over the process lifetime" - similar to the Device class)

common4j/src/main/com/microsoft/identity/common/java/logging/ProductHelper.java
fadidurah: It's going to be the calling client library, so MSAL. I can edit this comment to make that more clear.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
fadidurah: I've discussed with @shahzaibj , we agreed that using Diagnostic Context for now is fine to get the change in and keep track with other teams (this is a multi team effort, most teams are aiming to get it in this month), propagating throughout the entire code base is going to balloon these PRs quite a bit, especially since the fields already exist in diagnostic context for other purposes. Also it leaves more room for misses if we start editing every flow since we I might miss something, while Diagnostic context gets set at the command dispatcher level, so coverage seems good end on that front. We aren't adding any additional fields in Diagnostic Context, they are already there.

To answer your question though, it would be "OneAuth", not Broker. We're interested in the source library.

fadidurah: We wanted to avoid a massive refactoring effort, we can discuss during scrum

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationFragment.java
fadidurah: This isn't really needed IIRC. I made this change to bring it to parity with the rest of our code base. Basically, in the thread used for authorization request, the product (library name) and version were not set in Diagnostic context, despite being set in the thread that launched the acquire token request, because it's a thread local instance.

Yes the original use case was for correlation id and thread id, but the product name and version were already added there for other purposes (I think Linux related primarily)

fadidurah: I'll revert this addition since i'm not using it in my implementation

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
fadidurah: changed these to sourceLibraryName and sourceLibraryVersion for clarity

common4j/src/main/com/microsoft/identity/common/java/logging/ProductHelper.java
fadidurah: device class isn't currently doing this for library version and name in the platform class, only android metadata. we'll need to assess where those methods are called and propogate those fields to where we create platform components. I think this is better covered in a refactoring effort.

To me, it seems the larger issue is the usage of diagnostic context for anything besides corr id and thread id. We should probably allocate bandwidth to adress that throughout, i think it's gonna need some pretty heavy investment. Aside from that, I think we can extend this discussion to corr id as well (maybe not thread id though), if we're going to move library version and library name to be passed in the parameters only, then i think we can do the same with corr id, no? Regardless, I think it's part of a larger design discussion.

rpdome: IMO, it's acceptable for corr ID to be in the DiagnosticContext -  since we need to print that in every log line, and we're not passing the token parameter in every place we invoke logging functions.

The library name and library version doesn't belong there. We don't print it in every log line log. We're sending them to eSTS.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -456,12 +460,29 @@ private HashMap<String, String> getRequestHeaders(final Bundle state) {
...
         } catch (Exception e) {
             return null;
         }
     }
 
+    /**
+     * Helper method to check if the authorization request is being made through broker.
+     * Done by checking for broker version key in the url
+     */
+    private boolean checkIfBrokerRequest(final String authorizationUrl) {
```
rpdome: nit: isBrokerRequest

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/EmbeddedWebViewAuthorizationStrategy.java
```diff
@@ -84,15 +85,26 @@ public Future<AuthorizationResult> requestAuthorization(GenericAuthorizationRequ
...
+        final Intent authIntent = buildAuthorizationActivityStartIntent(requestUrl, sourceLibraryName, sourceLibraryVersion);
 
         launchIntent(authIntent);
         return mAuthorizationResultFuture;
     }
 
     // Suppressing unchecked warnings during casting to HashMap<String,String> due to no generic type with mAuthorizationRequest
     @SuppressWarnings(WarningType.unchecked_warning)
-    private Intent buildAuthorizationActivityStartIntent(URI requestUrl) {
+    private Intent buildAuthorizationActivityStartIntent(URI requestUrl, @Nullable final String sourceLibraryName, @Nullable final String sourceLibraryVersion) {
```
rpdome: nit: comment and new line for each params.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactory.java
rpdome: ack. please add comment in the the library helper class (or DiagnosticContext) that we should move these two values to the parameter class at some point - and we should NEVER add anything new to DiagnosticContext unless we **_really_** need it for logging purpose.

rpdome (approved):

PR #2435, opened 2024-06-20
Add check for unset correlation ID when configuring request headers

rmccahill:
If a correlation ID hasn't been configured for a request, the 'UNSET' value will be sent to the server, and the server will then return that 'UNSET' value from the response. To prevent this, the correlation ID should only be appended if it is a valid value.

Additionally, tests have been added to ensure that an 'UNSET' value isn't added to the request headers, and valid correlation IDs are being added to the request headers on a valid request.

Also, there are a few tweaks made to `NativeAuthRequestHandlerTest` - there are some tests where a request is configured and created, but there are no assertions made on those requests. I've added a few checks to tests that were missing validations, and some other small tweaks.

Update: 
“If a correlation ID hasn't been configured for a request, the 'UNSET' value will be sent to the server, and the server will then return that 'UNSET' value from the response.” -> After experiment, this assumption is wrong.  The header used on the SDK side is client-request-id not ms-client-request-id, these two values are different if they both attached in the response header. The interesting thing is that, if client sends invalid correlation id like "UNSET", the server would  not return the client-request-id to the client. On the contrary, if the correlation id is valid and client sent it in the client-request-id header, the server will return it as the same value in the client-request-id header. SDK made the assumption that if no client-request-id header is sent to the server from client, the server will generate one for the client seems untrue. The correlation id is dependant on client-request-id x-ms-request-id is the request specific, it targets the Service Request Id not the CorrelationId, while x-ms-request-id is returned from the server in all times.


common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
```diff
@@ -310,7 +311,11 @@ class NativeAuthRequestProvider(private val config: NativeAuthOAuth2Configuratio
     //region helpers
     private fun getRequestHeaders(correlationId: String): Map<String, String?> {
         val headers: MutableMap<String, String?> = TreeMap()
-        headers[AuthenticationConstants.AAD.CLIENT_REQUEST_ID] = correlationId
+
+        if (correlationId != "UNSET") {
```
SaurabhMSFT, line 313: The string 'UNSET' should be used from a constant file.

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
```diff
@@ -148,6 +151,22 @@ class NativeAuthRequestHandlerTest {
         )
     }
 
+    @Test
+    fun testSignUpStartWithUnsetCorrelationIdShouldHaveNilHeader() {
```
SaurabhMSFT: `testSignUpStartWithUnsetCorrelationIdShouldHaveNilHeader` -> 'testSignUpStartWithUnsetCorrelationIdShouldNotHaveHeader'

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
```diff
@@ -310,7 +311,11 @@ class NativeAuthRequestProvider(private val config: NativeAuthOAuth2Configuratio
...
+            headers[AuthenticationConstants.AAD.CLIENT_REQUEST_ID] = correlationId
+        }
+
```
Yuki-YuXin: What about correlationId == "UNSET"? Will the correlationId be set as null somewhere?
` val privateCorrelationId = if (correlationId == "UNSET") { UUID.randomUUID().toString() } else { correlationId }` only for AccountState `acquireTokenSilentParameters`. Do we have other cases that may have UNSET?

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
rmccahill: The title is accurate as it is - the test passes a "UNSET" value, and in this case the header should be nil.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
rmccahill: Having a look at the rest of the codebase and "UNSET" seems to be used in place instead of being used as a constant value. I'd rather stick with the convention here, given that it's a straightforward name and something that doesn't have a specific meaning or is used outside of the context of the SDK.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
rmccahill: If it's equal to UNSET, we don't want to attach that UNSET value to the header, because then the server will return that back to us. We want to send a nil header so that the server will generate a correlation ID for us if it isn't provided. This is the block of code that configures the request headers for every request so it's a good place to check it here.

In terms of scenarios where it's possible the value is "UNSET", the example you gave is one. Additionally, for every flow that's initiated (sign in, sign up, reset password), the correlation ID will also be initialised as "UNSET". The check is added here as a 'catch-all' for those flows.

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
SammyO: @rmccahill can you elaborate? Why should we be sending an empty header (i.e. the `REQUEST_ID` header is set, but it doesn't contain a value), rather than not set the header at all?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
SammyO: > Additionally, for every flow that's initiated (sign in, sign up, reset password), the correlation ID will also be initialised as "UNSET".

@rmccahill why is that the case? In practice the correlation ID should never be set to `UNSET`. At the start of the flow, i.e. in `CommandParametersAdapter`, the ID is generated for `signIn()`, `signUp()` and `resetPassword()` (i.e. methods that start the flow).

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
SammyO: Actually, what you're saying seems to contradict line 316 of `NativeAuthRequestProvider.kt`. Or maybe we're mis-communicating :D 

The way I see it there's a difference between "no header" vs. "header is set and value is `null`". Saurabh and I mean the first (and so I would agree that the test name should be "should not have header").

SammyO:
@rmccahill could you remove the conditional setting of the correlation ID in `AccountState` and see if the multiple tokens tests still pass? 

Context:
- this issue came up when I was writing the multiple token tests, and added that statement was a quick fix to make the tests pass
- why? in the context of tests the threading naming mechanism in `DiagnosticContext` doesn't work correctly, and so the correlation ID for API calls was set to `UNSET` by `DiagnosticContext`. The API returned this, and in subsequent refresh token logic an UUID was attempted to be composed based on the `UNSET` String. This caused an exception.
- with the changes of this PR, this crash shouldn't be happening anymore.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
rmccahill: You're correct - I misread the underlying code, the DiagnosticContext defaults the correlation ID to be "UNSET", but the CommandDispatcher updates this when it runs a command. So the AccountState example should be the only case where this happens (that we know of at least)

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
rmccahill: Ah I see what you mean - in my mind 'nil header' refers to the header not being present. But I can see how that's confusing. I've updated the naming on these tests to reflect the suggested name.

Yuki-YuXin (approved):

Yuki-YuXin:
The removal and fix of the conditional setting of the correlation ID in AccountState is here https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2135

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
```diff
@@ -76,8 +78,8 @@ public IRequestContext getRequestContext() {
     public String getThreadCorrelationId() {
         IRequestContext context = getRequestContext();
         String correlationId = context.get(DiagnosticContext.CORRELATION_ID);
-        if (correlationId == null) {
-            correlationId = UNSET;
+        if (correlationId == null || correlationId.equals(UNSET)) {
+            correlationId = UUID.randomUUID().toString();
```
Yuki-YuXin, line 82: We can also use `correlationId = "";` and let CommandDispatcher.initializeDiagnosticContext
`
final String correlationId = StringUtil.isNullOrEmpty(requestCorrelationId) ?
                    UUID.randomUUID().toString() :
                    requestCorrelationId;
`

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
```diff
@@ -310,8 +311,8 @@ class NativeAuthRequestProvider(private val config: NativeAuthOAuth2Configuratio
     private fun getRequestHeaders(correlationId: String): Map<String, String?> {
         val headers: MutableMap<String, String?> = TreeMap()
         headers[AuthenticationConstants.AAD.CLIENT_REQUEST_ID] = correlationId
```
Yuki-YuXin: Revert it back because we are using the local parameters.getCorrelationId() in the http/SDK response rather than the response header CLIENT_REQUEST_ID, which means even though we filter out the UNSET in the header, it will still exist in the response.

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
```diff
@@ -763,17 +948,37 @@ class NativeAuthRequestHandlerTest {
     }
 
     @Test
-    fun testResetPasswordContinueSucces() {
+    fun testResetPasswordContinueWithUnsetCorrelationIdShouldFilterOutToUUID() {
+        val commandParameters = ResetPasswordSubmitCodeCommandParameters.builder()
```
Yuki-YuXin, line 952: The SDK should never set the correlationId to "UNSET" here in the process of a flow. Should we keep this test?

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestHandlerTest.kt
```diff
@@ -355,6 +436,21 @@ class NativeAuthRequestHandlerTest {
         )
     }
 
+    @Test
+    fun testSignInInitiateWithUnsetCorrelationIdShouldFilterOutToUUID() {
+        val commandParameters = SignInStartCommandParameters.builder()
+            .platformComponents(mock<PlatformComponents>())
+            .username(emptyString)
+            .correlationId("UNSET")
```
Yuki-YuXin, line 444: In the real SDK, we are using DiagnosticContext.INSTANCE.getThreadCorrelationId(), which will catch the "UNSET". In this test case, we set the "UNSET" intentionally which skip that part. Same below.

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
SammyO: Am I right in saying: 
it's okay to create a UUID here, because:
1) we don't need to use/set the thread ID, because that's not what's being used in the request headers (i.e. when setting the request headers, we use map key `DiagnosticContext.CORRELATION_ID`, not `DiagnosticContext.THREAD_ID`
2) `CommandDispatcher.initializeDiagnosticContext()` does the same; generate a UUID.

SammyO, line 82: Am I right in saying: 
it's okay to create a UUID here, because:
1) we don't need to use/set the thread ID, because that's not what's being used in the request headers (i.e. when setting the request headers, we use map key `DiagnosticContext.CORRELATION_ID`, not `DiagnosticContext.THREAD_ID`
2) `CommandDispatcher.initializeDiagnosticContext()` does the same; generate a UUID.

SammyO: @Yuki-YuXin  if what you mention is also a solution, then why did you add this code change here? to rephrase: I agree that what `CommandDispatcher.initializeDiagnosticContext` does should be sufficient. So why is your code change needed?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
```diff
@@ -310,8 +311,8 @@ class NativeAuthRequestProvider(private val config: NativeAuthOAuth2Configuratio
...
-        headers[AuthenticationConstants.SdkPlatformFields.PRODUCT] = LibraryInfoHelper.getLibraryName()
-        headers[AuthenticationConstants.SdkPlatformFields.VERSION] = LibraryInfoHelper.getLibraryVersion()
+        headers[AuthenticationConstants.SdkPlatformFields.PRODUCT] = DiagnosticContext.INSTANCE.requestContext[AuthenticationConstants.SdkPlatformFields.PRODUCT]
```
SammyO: I think this code change is wrong. And the one on the line below.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/SignInInteractor.kt
```diff
@@ -103,7 +103,7 @@ class SignInInteractor(
         )
 
         val rawApiResponse = nativeAuthResponseHandler.getSignInInitiateResultFromHttpResponse(
-            requestCorrelationId = requestCorrelationId,
+            requestCorrelationId = requestCorrelationId, // Question: should we use the CLIENT_REQUEST_ID header from response here?
```
SammyO: I think you're misunderstanding what's happening here (as did I when we talked about it yesterday).
1) when the request is created (`nativeAuthRequestProvider.createSignInInitiateRequest`) `getRequestHeaders()` is used to compose the correlation ID, which in turn takes the correlation ID from the command parameters.
2) when the response object is created (by converting the raw API response into a local data model), we pass the original correlation ID to the request provider. We do this because the API might not return a correlation ID, in which case we set it manually using `requestCorrelationId`.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
SammyO: I'm not following this. This method isn't used in responses; it's used to create requests.

SammyO: Going back to the original goal of this ticket: I still think we need a catch-all fallback here that catches any "UNSET" values and ensures we don't send it to the API (because if we do, the API will return "UNSET" back to us, which we don't want). There are 2 ways I can think of:
1. catch "UNSET" and replace with UUID
2. catch "UNSET" and not set correlation ID at all. This is what MSAL does. The expectation is that the API should then create a correlation ID and send it back to us.

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
Yuki-YuXin: Because the current code snippet will set the correlationId to UNSET. If the correlationId passed into CommandDispatcher.initializeDiagnosticContext is UNSET, then after correlationId = StringUtil.isNullOrEmpty(requestCorrelationId) ? UUID.randomUUID().toString() : requestCorrelationId,  the correlationId would keep as "UNSET"

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
Yuki-YuXin: Yes, the reason we don't want UNSET to be sent in the http request header is because we don't want the server to response "UNSET" to us in the http response header. However, the SDK didn't use the http response header to construct the SDK response. It used the local correlation passed in or recorded in DiagnosticContext. Since we don't use the correlation id in the http response header then nothing sent in the request will affect the correlation ID of the SDK response

common4j/src/main/com/microsoft/identity/common/java/logging/DiagnosticContext.java
Yuki-YuXin: Yes, that's correct.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
Yuki-YuXin: I have no idea why the changes here. If that's wrong, I will revert it back.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/interactors/SignInInteractor.kt
Yuki-YuXin: Oh, my bad.
After checking inside `getSignInInitiateResultFromHttpResponse`
```
// If the API doesn't return a correlation ID header value, use the correlation ID of the original API request
        val correlationId: String = response.getHeaderValue(AuthenticationConstants.AAD.CLIENT_REQUEST_ID, 0).let {responseCorrelationId ->
```
Then we definitely should have the try catch mechanism in the header.

Yuki-YuXin: This may explain why SDK will get UNSET even when the header filter is applied,
image

Yuki-YuXin: And the client request id only exists when client request id is valid. The client request id is different from x-ms-request-id
image

Yuki-YuXin:
@SammyO  After talking to IOS, the UNSET correlation id seems to be only an issue on the Android side.

"In brief what native auth iOS SDK operates around the correlationID:
- every initial actions on the public SDK interface (like signIN, getAccessToken, signUp, resetPassword) accept an optional correlationId
- this correlation has UUID as type, so this means that when the user specifies a correlation Id, it has a valid format
when the user doesn't not specify a correlationID, the SDK create a new one internally (just a random UUID). This is the same behaviour of MSAL Objective-C
- this means that native auth iOS SDK **always send a valid UUID** string as correlation ID to eSTS."

SammyO:
> @SammyO After talking to IOS, the UNSET correlation id seems to be only an issue on the Android side.
> 
> "In brief what native auth iOS SDK operates around the correlationID:
> 
> * every initial actions on the public SDK interface (like signIN, getAccessToken, signUp, resetPassword) accept an optional correlationId
> * this correlation has UUID as type, so this means that when the user specifies a correlation Id, it has a valid format
>   when the user doesn't not specify a correlationID, the SDK create a new one internally (just a random UUID). This is the same behaviour of MSAL Objective-C
> * this means that native auth iOS SDK **always send a valid UUID** string as correlation ID to eSTS."

@Yuki-YuXin I don't think that's exactly the question we want to ask the iOS team. iOS' intended behaviour that you describe is also the intended behaviour on Android. However, we need to account for situations where that intended design has a bug. The question is: if the SDK (iOS or Android) doesn't send a correlation ID to the API, does the API generate one and send it back through headers? Our assumption is yes, but our findings from Friday state the opposite.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
Yuki-YuXin: Currently, I keep the changes to apply the second option: "catch "UNSET" and not set correlation ID at all". The expectation could not be achieved for the reason in the PR summary. I don't think we should make changes to the NativeAuthResponseHandler part about the header in this PR. If the pipeline passes, maybe we can have a discussion later on what would be best.

SammyO (approved):

PR #2437, opened 2024-06-25
Reduce the timeout of backup IPC between broker apps

rpdome:
Reduce the timeout of backup IPC between broker apps from 5 to 2

5 is too long. was considering 1, but the broker process boot up today could take up to 1 second.

so I settled with 2.


somalaya (approved):

shahzaibj:
Can you add a `WHY` section to the PR description so we know the rationale behind this change?

rpdome:
> Can you add a `WHY` section to the PR description so we know the rationale behind this change?

@shahzaibj added.

shahzaibj (approved):

PR #2441, opened 2024-07-01
Classifying TimeoutException as timed_out

melissaahn:
### Summary
TimeoutException, a very common exception seen in telemetry, is currently being classified under error code unknown_error. We've been asked to classify this exception as a more known, retriable error, so this PR creates the new error_code timed_out and adds in a check to ExceptionAdapter.


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -389,6 +390,14 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
             );
         }
 
+        if (e instanceof TimeoutException) {
+            return new ClientException(
+                    ClientException.TIMED_OUT,
```
ademoreyaldo, line 395: Is this code that runs on the broker? How will this be interpreted by the AndroidCommon code on the app side? (What will be the exception type?)

melissaahn: This method does run in the broker via shared common methods, I tested this out with a silent token request that times out and gets caught here. 
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java#L258
The exception that OneAuth will receive will be like the others listed in this method (ClientException) but with the error code timed_out instead of unknown_error.

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
ademoreyaldo: Ok, so we will need to add a new timed_out error code handling case as soon as we ingest this change. Thanks. I'll open a work item to do that now.

ademoreyaldo (approved):

melissaahn:
Note that I've added a small change to the build-consumers.yml to mitigate some failing native auth tests which are not related to this PR.

PR #2445, opened 2024-07-05
Enable the new Broker Discovery by default in MSAL

rpdome:
Now that we're enabling the feature on LTW to 100%, and oneauth has turned this on by default since January.
It should be safe to enable this for MSAL apps (mostly 3rd party)


somalaya:
LGTM. Let's monitor telemetry to keep an eye on MSAL apps specific issues after we ship this.

somalaya (approved):

p3dr0rv (approved):

moumighosh (commented):
Please hold off merging this in the July release. Create an impact doc, with data from telemetry to access the impact across AAD and MSA accounts

PR #2450, opened 2024-07-09
Moving OS version check for passkeys

melissaahn:
### Summary
If a device's OS does not support passkeys, then we shouldn't even reveal the passkey option. This means that the OS check should be done while we are deciding to add `webauthn=1` or not. 
As a result of needing to move the OS checks to earlier points (specifically when we add the extra query param `webauthn=1`),  we also saw that the placement of the passkey extra query param logic needed to be revised. The `updateWithAndGetPlatformSpecificExtraQueryParameters` implementation for `AndroidPlatformUtil` is cleared and is instead put in the implementation of `AndroidBrokerPlatformUtil`, a new class which can be found in this PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2858/files

### Related PRs
- https://github.com/AzureAD/ad-accounts-for-android/pull/2858
- https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2138


common4j/src/main/com/microsoft/identity/common/java/util/IPlatformUtil.java
```diff
@@ -135,5 +135,5 @@ public interface IPlatformUtil {
      * @return a list of extra query parameters, or null.
      */
     @Nullable
-    List<Map.Entry<String, String>> updateWithAndGetPlatformSpecificExtraQueryParameters(@Nullable List<Map.Entry<String, String>> originalList);
+    List<Map.Entry<String, String>> updateWithAndGetPlatformSpecificExtraQueryParametersForBroker(@Nullable List<Map.Entry<String, String>> originalList);
```
shahzaibj: If this is for Broker then I'm not sure this needs to go in common repo.

shahzaibj: Do we actually not ever envision having a need for platform specific query parameters in general (for both Broker and Auth libs)?

melissaahn: I would think that there could be a need in the future, but with the passkey feature, there's this additional distinction between Android broker and Android MSAL, where the broker needs to automatically add this qp, and for MSAL, the host apps are responsible for adding the qp/enabling passkeys.

melissaahn: It does make sense that some logic which is broker specific should not be in common. So there's definitely an existing issue with having the passkey qp logic in this method (because I don't want this logic to be called from a common or MSAL context)

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoUtil.kt
```diff
@@ -0,0 +1,70 @@
...
+import android.os.Build
+import com.microsoft.identity.common.java.constants.FidoConstants.Companion.WEBAUTHN_QUERY_PARAMETER_FIELD
+import com.microsoft.identity.common.java.constants.FidoConstants.Companion.WEBAUTHN_QUERY_PARAMETER_VALUE
+import com.microsoft.identity.common.java.logging.Logger
+import java.util.AbstractMap
+
+/**
+ * A utility class to help with FIDO related operations.
+ */
+class FidoUtil {
```
shahzaibj: Since this is a Util class, do we still need a class for this? Why not write a top level function?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/FidoUtil.kt
melissaahn: Put the method in AndroidPlatformUtil: 3869563cc6c9c5e871ab4cff27942a95bd667e98
(Hopefully that's the top level class you were thinking of)

shahzaibj: My comment was more around considering writing a class-less kotlin file with just functions but I guess that wasn't going to be possible anyway since we need interop with java

PR #2451, opened 2024-07-10
Added span name for refresh transfer token

somalaya:
- Related broker PR https://github.com/AzureAD/ad-accounts-for-android/pull/2859


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -61,5 +61,6 @@ public enum SpanName {
     AcquirePrtUsingTransferToken,
     AcquireTransferTokenUsingPrt,
     SaveTransferTokenToBlockstore,
-    RetrieveTransferTokenFromBlockstore
+    RetrieveTransferTokenFromBlockstore,
+    RefreshTransferToken
```
shahzaibj, line 64: Do we know how often this operation will execute?

somalaya: RefreshTransferToken span would be once in every 30 days when an acquireTokenSilent call is made.
RetrieveTransferTokenFromBlockstore on the other hand would be once per acquireTokenSilent call if it is for an MSA account and active broker is LTW (pre-installed).  Is there any restriction on creating spans that are emitted this frequently?

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
shahzaibj: Generally we don't want to double the amount of our spans i.e. if you add a span that runs with every ATS request then the number of spans emitted for an ATS will double essentially. If you have 10 billion ATS spans then you will have additional 10 billion now due to this new span. That said, we should check how much traffic (spans for ATS) we get today for MSA + LTW scenario that you described. I think that would give a better idea of how much we are increasing here. Storage / COGS reduction is certainly very important to the org.

somalaya: Removed this span. I will add all the telemetry in 1 PR with more information added like how many spans will be meitted and why it is required.

shahzaibj (approved):

PR #2453, opened 2024-07-12
Changes needed for token binding CA policy error handling

iamgusain:
Adding logic to handle token binding CA policy errors.

Corresponding STS side PR for iOS and mac (Android will come later)- https://msazure.visualstudio.com/DefaultCollection/One/_git/ESTS-Main/pullrequest/10445085?_a=files&path=/src/Product/Microsoft.AzureAD.Common/Configuration/ICloudStsSettings.cs

Description of errors/redirect uri returned from server for upgrading to or performing strong device registration
Silent Token Request (when device is registered)--> error: invalid_grant, suberror: insufficient_device_registration
Silent Token Request (when device is not registered)--> error: device_authentication_failed
 
Interactive Token Request (when device is registered) --> redirect uri- msauth://upgradeReg?username=<>&client_info=<>&token_protection_required=true
Interactive Token Request (when device is not registered) --> msauth://wpj?username=<>&client_info=<>&token_protection_required=true


common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -22,6 +22,11 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.net;
 
+import static com.microsoft.identity.common.java.AuthenticationConstants.AAD.CLIENT_REQUEST_ID;
+import static com.microsoft.identity.common.java.net.HttpConstants.HeaderField.CONTENT_TYPE;
+import static com.microsoft.identity.common.java.net.HttpConstants.HeaderField.XMS_CCS_REQUEST_ID;
+import static com.microsoft.identity.common.java.net.HttpConstants.HeaderField.XMS_CCS_REQUEST_SEQUENCE;
```
mohitc1, line 28: unwanted imports?

mohitc1:
Consider adding UTs for new code

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -270,7 +270,7 @@ private static HttpRequest constructHttpRequest(@NonNull HttpClient.HttpMethod h
                                                     byte[] requestContent) {
 
         // Apply special backcompat behaviors for PATCH, if reqd
-        if (HttpClient.HttpMethod.PATCH == httpMethod) {
+        if (HttpClient.HttpMethod.PATCH == httpMethod && !requestUrl.getPath().contains("upgradeRegistration")) {
```
mohitc1: Can we create overload if method which accepts let caller pass if PATCH -> POST with appropriate header added? UpgradeDeviceRegistrationRequest can call it.

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
iamgusain: These are used

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
iamgusain: I override the patch method (which is a separate method) to keep the implementation to not convert it to Post and use PATCH.

mohitc1 (approved):

p3dr0rv:
LGTM, just add more UT to validate the new cases

p3dr0rv (approved):

PR #2458, opened 2024-07-23
Disable build flag checks on Scheduled test

rpdome:
- Move EnvironmentVariable and BuildReason from broker4j
- add a check in BuildConfigTest

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
+++ b/common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
@@ -24,6 +24,7 @@ package com.microsoft.identity.common
 
 import com.microsoft.identity.common4j.env.BuildReason
 import com.microsoft.identity.common4j.env.BuildReason.Companion.isBuildReason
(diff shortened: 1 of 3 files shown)


fadidurah (approved):

common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
```diff
@@ -31,11 +33,21 @@ class BuildConfigTest {
 
     @Test
     fun failIfBypassRedirecUriCheckEnabled(){
+        // Do not run this on scheduled test.
+        if (isBuildReason(BuildReason.Schedule)) {
+            return
+        }
```
shahzaibj: This is not the right way to skip test. You should use the JUnit Assume API. 

https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeFalse(boolean)

Same comment for all other places such pattern is being followed.

shahzaibj (approved):

PR #2460, opened 2024-07-26
Classifying more UNKNOWN_ERRORs

melissaahn:
### Summary
The error code `UNKNOWN_ERROR` appears frequently in our telemetry. We have a large amount of client exceptions going through the `clientExceptionFromException` method with an `UNKNOWN_ERROR` code, and there are also many places where `ClientException`s are being explicitly created with the code. 
After looking at the telemetry and doing some code searches in our codebase, I added some more specific error codes which will be used in the common and broker libraries. Any suggestions on name changes are welcomed.

### Related PRs

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
@@ -52,6 +52,7 @@ import com.microsoft.identity.common.java.util.StringUtil;
 import org.json.JSONException;
 
 import java.io.IOException;
+import java.net.SocketTimeoutException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
@@ -340,7 +341,8 @@ public class ExceptionAdapter {
 
     public static BaseException baseExceptionFromException(final Throwable exception) {
         Throwable e = exception;
-        if (exception instanceof ExecutionException) {
+        if (exception instanceof ExecutionException
+                && exception.getCause() != null) {
             e = exception.getCause();
         }
         if (e instanceof BaseException) {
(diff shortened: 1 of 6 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -92,7 +92,7 @@ public static BaseException exceptionFromAuthorizationResult(@NonNull  final Aut
                     methodTag,
                     "AuthorizationErrorResponse is not set"
             );
-            return new ClientException(ClientException.UNKNOWN_ERROR, "Authorization failed with unknown error. Authorization Status: " +authorizationResult.getAuthorizationStatus());
+            return new ClientException(ClientException.AUTHORIZATION_RESULT_ERROR, "Authorization failed with unknown error. Authorization Status: " +authorizationResult.getAuthorizationStatus());
```
shahzaibj: We can probably rename the error code here to `AUTHORIZATION_RESULT_NULL_ERROR_RESPONSE`

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
rpdome: +1 (and also improve the error message accordingly?)

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
melissaahn: Addressed here: 96ab98a6d06c248fb699749acea717d36f61c015

melissaahn:
I've tagged this PR with "skip-consumers-check" because there is one MSAL unit test which is expected to fail. See this PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2147

PR #2462, opened 2024-07-30
Authority data is persisted across different SDK instances

Yuki-YuXin:
Company PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2159

## Investigation Summary:
- The first SDK instance: since there is no authorityIsKnownFromConfiguration, Authority class will return CIAMAuthority(authorityUrl) which has no client id inside. However, under NativeAuthPublicClientApplicationConfiguration.validateConfiguration, the CIAMAuthority is converted into NativeAuthCIAMAuthority with client id attached.
```kotlin
if (defaultAuthority is CIAMAuthority) {
    val nativeAuthAuthority = NativeAuthCIAMAuthority(
    authorityUrl = defaultAuthority.authorityUri.toString(),
    clientId = clientId
    )
    authorities.clear()
    authorities.add(nativeAuthAuthority)
```

- The second SDK instance: there is an authorityIsKnownFromConfiguration because authority url is the same. Authority class used the previous cliend id that wrapped inside the authority instance to generate a new NativeAuthCIAMAuthority. Thus the client id inside this one authority is conflicted with the current right client id, which cause the failure of tests.
```kotlin
authority = new NativeAuthCIAMAuthority(authorityUrl, ((NativeAuthCIAMAuthority) configuredAuthority).getClientId());
```

## Changes Summary:
```
public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
        return getAuthorityFromAuthorityUrl(authorityUrl, null);
    }
```

### The target of this PR is dev branch


SammyO:
@Yuki-YuXin  can you explain why this fixes the issue (i.e. the issue of having a singleton to manage "known authorities" which effectively shares data across multiple SDK instances)? This PR seems to fix the issue in a different place; I'm not sure that's correct.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -157,9 +157,7 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
             } else if (CIAM.equalsIgnoreCase(authorityTypeStr)) {
                 authority = new CIAMAuthority(authorityUrl);
             } else if (AAD_NA.equalsIgnoreCase(authorityTypeStr) && configuredAuthority instanceof NativeAuthCIAMAuthority) {
-                authority = new NativeAuthCIAMAuthority(
-                        authorityUrl,
-                        ((NativeAuthCIAMAuthority) configuredAuthority).getClientId());
+                authority = new CIAMAuthority(authorityUrl);
```
SammyO: Why do we want to return a CIAMAuthority in the case of native auth, rather than a NativeAuthAuthority? Don't our flows need a NativeAuthAuthority to complete the API flows?

Yuki-YuXin: The CIAMAuthority will be transformed into NativeAuthAuthority in later stage using the current client id. `((NativeAuthCIAMAuthority) configuredAuthority).getClientId());` this line is the reason that cased persisted data in different SDK instances.

Yuki-YuXin:
@SammyO When I was trying to understand the problem, I personally think singleton authority may not be the reason. The root cause is NativeAuthCIAMAuthority wraps a client id parameters and store it for the second time usage. Indeed, the code only uses the authority url to search for the known authority in the second time, which is the first time NativeAuthCIAMAuthority (authorityUrl, legacy client id of the first time). The conflicts between the client id and the one inside NativeAuthCIAMAuthority fail those tests.

Yuki-YuXin:
A hint to add changelog entry

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
GBakolas: Maybe I haven't understood the problem correctly. What exactly is the SDK behavior we want to achieve. Do we want to create a new authority object and add it to the known authorities list for each SDK instance? What happens if two instances use the same client id are we going to create two different objects and keep them in the list in this case?

Yuki-YuXin: @GBakolas The SDK behavior we want to achieve is to create a new authority object and add it to the known authorities list for each different authority instance (the current SDK standard is different authority url). Each time an authority is added, the native auth instance clears the authority list

```kotlin
PublicClientApplication.createNativeAuthApplication

config.getAuthorities().clear();
final Authority authorityObject = Authority.getAuthorityFromAuthorityUrl(authority, clientId);
authorityObject.setDefault(true);
config.getAuthorities().add(authorityObject);
```
And
```kotlin
NativeAuthPublicClientApplicationConfiguration.validateConfiguration

if (defaultAuthority is CIAMAuthority) {
    val nativeAuthAuthority = NativeAuthCIAMAuthority(
        authorityUrl = defaultAuthority.authorityUri.toString(),
        clientId = clientId
    )
   authorities.clear()
   authorities.add(nativeAuthAuthority)
}
```
So each native auth SDK instance will only has one authority in its authority list.

For each E2E test file, such as SignUpEmailPasswordTest and SignUpEmailOTPTest, their SDK instance (application) are different. Each instance has its own configuration file with only one authority url and one client id. One SDK instance has multiple authorities in its authority list only happens when there are multiple authority urls in the configuration files, which is out of native auth's scopes in my mind.

SammyO: @Yuki-YuXin  does `config.getAuthorities().clear();` clear the `knownAuthorities` list inside `Authority`?

Yuki-YuXin: @SammyO The answer is No I think.  For example, the code below. `config.getAuthorities().clear()` is before `getAuthorityFromAuthorityUrl` while `knownAuthorities` still exist when doing the search.

```kotlin
PublicClientApplication.createNativeAuthApplication

config.getAuthorities().clear();
final Authority authorityObject = Authority.getAuthorityFromAuthorityUrl(authority, clientId);
authorityObject.setDefault(true);
config.getAuthorities().add(authorityObject);
```
When `NativeAuthPublicClientApplication.initializeApplication` 
```kotlin
private fun initializeApplication() {
        val context = nativeAuthConfig.appContext

        AzureActiveDirectory.setEnvironment(nativeAuthConfig.environment)
        Authority.addKnownAuthorities(nativeAuthConfig.authorities)
```
The `knownAuthorities` is changed.

```kotlin
Authority.java

private static final List<Authority> knownAuthorities = new ArrayList<>();
...
public static void addKnownAuthorities(List<Authority> authorities) {
        synchronized (sLock) {
            knownAuthorities.addAll(authorities);
        }
}
```

SammyO: Then I don't think your reply to Giannis is correct. What we're trying to achieve is:
1. not change the fundamental MSAL code that retrieves the authority from `knownAuthorities`: we need to understand why this logic is here.
2. not change the fundamental MSAL code that determines whether an authority is known: it currently does that based on authority URL, and doesn't take into account the client ID for matching. Same as above; we need to understand this logic better, so we're not touching it for now.
3. the native auth flows can continue to look up authorities from the `knownAuthorities` list, but we can't rely on the client ID from previous instances. We're therefore going to change the way `Authority.getAuthorityFromAuthorityUrl` works: an authority URL should be passed as well as a client ID (optional generally, but expected for native auth), so that this method can return a fully formed `NativeAuthCIAMAuthority` with updated data.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -157,9 +168,7 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
             } else if (CIAM.equalsIgnoreCase(authorityTypeStr)) {
                 authority = new CIAMAuthority(authorityUrl);
             } else if (AAD_NA.equalsIgnoreCase(authorityTypeStr) && configuredAuthority instanceof NativeAuthCIAMAuthority) {
-                authority = new NativeAuthCIAMAuthority(
-                        authorityUrl,
-                        ((NativeAuthCIAMAuthority) configuredAuthority).getClientId());
+                authority = new NativeAuthCIAMAuthority(authorityUrl, clientId);
```
SammyO: Can we do a client ID null check here? If it's AAD_NA && it's a NativeAuthCIAMAuthority, we check if the client ID is null. If so, throw an exception (or do whatever else is appropriate, given the existing logic). We shouldn't be composing NativeAuthCIAMAuthority with a null client ID.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
Yuki-YuXin: In terms of the first and second point, let me ask MSAL side offline for better explanation and understanding.

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
Yuki-YuXin: Done.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -157,9 +168,11 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
             } else if (CIAM.equalsIgnoreCase(authorityTypeStr)) {
                 authority = new CIAMAuthority(authorityUrl);
             } else if (AAD_NA.equalsIgnoreCase(authorityTypeStr) && configuredAuthority instanceof NativeAuthCIAMAuthority) {
-                authority = new NativeAuthCIAMAuthority(
-                        authorityUrl,
-                        ((NativeAuthCIAMAuthority) configuredAuthority).getClientId());
+                if (clientId == null) {
+                    authority = new CIAMAuthority(authorityUrl);
```
SammyO: Can you add a comment explaining why it's not a problem in our flows that a CIAMAuthority is created here?

Yuki-YuXin: Done

changelog.txt
```diff
@@ -3,6 +3,7 @@ V.Next
 - [MINOR] Add null cursor BrokerCommunicationException error category (#2459)
 - [MINOR] Classifying more UNKNOWN_ERRORs (#2460)
 - [MINOR] Setting flights to control UrlConnection timeout values (#2473)
+- [PATCH] Fix authority data being persisted across different SDK instances (#2462)
```
SammyO: can we update this to reflect that we fixed it only in native auth flows?

Yuki-YuXin: Done

SammyO (approved):

rmccahill (approved):

PR #2469, opened 2024-08-08
Small logs improvements

p3dr0rv:


common/src/main/java/com/microsoft/identity/common/internal/broker/ipc/IpcStrategyWithBackup.kt
```diff
@@ -44,13 +44,16 @@ class IpcStrategyWithBackup (
         return try {
             primary.communicateToBroker(bundle)
         } catch (t: Throwable) {
+            Logger.info(methodTag, "Primary ipc failed : ${t.message}")
+            val usedStrategies = mutableListOf<String>()
```
rpdome, line 48: use stringbuilder instead?

rpdome (approved):

fadidurah (approved):

PR #2474, opened 2024-08-13
MFA business logic

SammyO:
MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2160


common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -1217,6 +1388,25 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
+    private fun performSubmitChallengeTokenRequest(
+        oAuth2Strategy: NativeAuthOAuth2Strategy,
+        parameters: MFASubmitChallengeCommandParameters
+    ): SignInTokenApiResult {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = parameters.getCorrelationId(),
+            methodName = "${TAG}.performOOBTokenRequest"
+        )
+        // Transform MFASubmitChallengeCommandParameters into SignInSubmitCodeCommandParameters,
```
SammyO, line 1442: @nilo-ms this is why I don't like submitChallenge(): the API call is the same for submit code. So in order to support challenges that are not `oob`, we'll need to make changes to the SDK anyway.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -319,6 +328,84 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         }
     }
 
+    fun signInSubmitChallenge(parameters: MFASubmitChallengeCommandParameters): MFASubmitChallengeCommandResult {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = parameters.getCorrelationId(),
+            methodName = "${TAG}.signInSubmitCode"
```
rmccahill: Logged method name should be `signInSubmitChallenge`

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -436,6 +521,142 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         }
     }
 
+    fun signInChallenge(parameters: MFAChallengeCommandParameters): MFAChallengeCommandResult {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = parameters.getCorrelationId(),
+            methodName = "${TAG}.challengeDefaultAuth()"
```
rmccahill: Logged method name should be `signInChallenge`

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -365,15 +452,13 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                         correlationId = result.correlationId
                     )
                 }
-                is SignInChallengeApiResult.IntrospectRequired -> {
-                    TODO()
-                }
-                is SignInChallengeApiResult.UnknownError -> {
+                is SignInChallengeApiResult.IntrospectRequired, is SignInChallengeApiResult.UnknownError -> {
```
rmccahill, line 446: Just double checking - we group this with the unknown error since we should never get an introspect required error when using the default challenge method?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthOAuth2Strategy.kt
```diff
@@ -185,11 +186,11 @@ class NativeAuthOAuth2Strategy(
     /**
      * Performs API call to /oauth/v2.0/introspect
      */
-    fun performSignInIntrospect(
+    fun performInIntrospect(
```
rmccahill: Nit: Should this be `performIntrospect`?

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
nilo-ms: I see. We will do something similar for iOS too. Although I'm not particularly fan of the term "challenge" either, we opted for it as a more general alternative to "code". If we feel strongly about it, we could reconsider and discuss this with Kaushik.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/commands/GetAuthMethodsCommand.kt
```diff
@@ -0,0 +1,72 @@
...
+    publicApiId
+) {
+
+    companion object {
+        private val TAG = GetAuthMethodsCommand::class.java.simpleName
+    }
+
+    /**
+     * The execution part of the command, to be run on the background thread.
+     * It calls the signInStart method of the native auth MSAL controller with the given parameters.
```
Yuki-YuXin: The comments here don't seem to be accurate. The same for other commands.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -1865,7 +2139,17 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                     tokenApiResult = this
                 )
             }
-            is SignInTokenApiResult.CodeIncorrect, is SignInTokenApiResult.MFARequired,
+            is SignInTokenApiResult.MFARequired -> {
+                SignInCommandResult.MFARequired(
```
nilo-ms, line 2132: I don't know if this the right place, but when the "getAccessToken()" function calls the /token endpoint and it returns an MFA required error, what message do we provide? I believe it would be helpful to include a custom message advising the external developer to sign out and sign in again.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -319,6 +328,84 @@ class NativeAuthMsalController : BaseNativeAuthController() {
...
+            return when (tokenApiResult) {
+                is SignInTokenApiResult.Success -> {
+                    saveAndReturnTokens(
+                        oAuth2Strategy = oAuth2Strategy,
+                        parametersWithScopes = parametersWithScopes,
+                        tokenApiResult = tokenApiResult
+                    )
+                }
+                is SignInTokenApiResult.MFARequired -> {
+                    SignInCommandResult.MFARequired(
```
nilo-ms: I'm wondering if the "MFARequired" status is implemented as intended here. I thought we would expect "MFA required" result only after entering a password, not a code. In this case the "SignInCodeRequiredState" needs to be updated too, and if I'm not wrong this wasn't part of the technical design.  Please, correct me if I'm wrong here

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SammyO: This is an open point, which is captured by https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3015026 and requires PM input.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SammyO: Yes. If the API returns introspect_required when the default challenge is called, the SDK should never get into a state where resendCode() can be called. Hence, if the SDK is in a state where resend code can be called, there should be a default auth method that was queried in the first place, hence introspect_required is an unexpected response.

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
SammyO: That limitation is currently handled in the interface layer, but indeed moving it here makes it cleaner.

rmccahill (approved):

PR #2477, opened 2024-08-15
Adjust JsonUtil

fadidurah:
Found a bug where broker4j and android seem to use different json versions, depending on what the json version shipped with android is. This change will make JsonUtil more resilient regardless of json version.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/util/JsonUtil.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/util/JsonUtil.java
@@ -50,8 +50,13 @@ public class JsonUtil {
         final Map<String, String> responseItems = new HashMap<>();
         while (keyIterator.hasNext()) {
             final String key = keyIterator.next();
-            final String value = jsonObject.get(key).toString();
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/util/JsonUtil.java
```diff
@@ -50,7 +50,8 @@ public static Map<String, String> extractJsonObjectIntoMap(final String jsonStri
         final Map<String, String> responseItems = new HashMap<>();
         while (keyIterator.hasNext()) {
             final String key = keyIterator.next();
-            responseItems.put(key, jsonObject.getString(key));
+            final String value = jsonObject.get(key).toString();
```
somalaya: If jsonObject.get(key) is null, toString()  could throw NPE

fadidurah: I'll add a check, good catch!

somalaya (approved):

rpdome (approved):

PR #2478, opened 2024-08-15
Custom Response handler API in MicrosoftStsOAuth2Strategy

mohitc1:
Changes:
1. Added requestToken(MicrosoftStsTokenRequest, AbstractMicrosoftStsTokenResponseHandler) to MicrosoftStsOAuth2Strategy. This would be used for PRT acquisition flows so supply handler based on protocol leg.
2. Added Util method to check response content media type in HttpResponse

Corresponding broker PR
Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2911


common4j/src/main/com/microsoft/identity/common/java/net/HttpResponse.java
```diff
@@ -114,6 +114,22 @@ public String getHeaderValue(@NonNull final String key, final int index){
...
+    public boolean isContentTypeMediaType(@NonNull final String expectedResponseType) {
+        final String responseContentType = this.getHeaderValue(HttpConstants.HeaderField.CONTENT_TYPE, 0);
+        if (StringUtil.isNullOrEmpty(responseContentType)) {
+            return false;
+        }
+        // Split the content type by semicolon to extract media type
+        // e.g. JWE response content type is expected to be application/jose and there can be extra parameter e.g
+        // eSTS is sending "application/jose; charset=utf-8". We are parsing the media type part and validating
+        // it.
+        String mediaType = responseContentType.split(";")[0].trim();
```
shahzaibj:
```suggestion
        final String mediaType = responseContentType.split(";")[0].trim();
```

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/microsoftsts/MicrosoftStsOAuth2Strategy.java
```diff
@@ -612,6 +603,39 @@ protected void validateTokenResponse(@NonNull final MicrosoftStsTokenRequest req
         validateTokensAreInResponse(request, response);
     }
 
+    public TokenResult requestToken(
```
shahzaibj: We have a method with same name but slightly different signature in the parent class `OAuth2Strategy`. Why not just refactor that method and change the signature (rather than creating an almost-duplicate here)?

mohitc1: Added requestToken() at OAuth2Strategy level itself with a new ITokenResponseHandler interface.

somalaya (approved):

rpdome (approved):

PR #2480, opened 2024-08-16
Disable Network check by default

rpdome:
This network check doesn't seem reliable.
(a handful of broker requests failed with this check, but then succeeded with OneAuth.)

This PR will disable the test by default - bringing the behavior in BOTH Broker and Common to match with OneAuth's.

---

NOTE: [Doze mode network check](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/3bd7337d4bc0df4974cc30701b22a43b3a4d8059/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java#L122) is not in scope of this work.
(that check is set by SDK. 
- In MSAL, that's true by default (via MSAL config).
- In OneAuth, it's always false. There's no way for an app to set, and OneAuth never touches it).

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
@@ -75,6 +75,11 @@ public class DefaultConnectionService implements IConnectionService {
         ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
                 .getSystemService(Context.CONNECTIVITY_SERVICE);
 
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_NETWORK_CONNECTIVITY_CHECK)){
+            // Skip the check.
+            return true;
+        }
+
         final boolean isConnectionAvailable;
         final boolean useNetworkCapabilityForNetworkCheck
                 = CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.USE_NETWORK_CAPABILITY_FOR_NETWORK_CHECK);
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/net/DefaultConnectionService.java
```diff
@@ -75,6 +75,11 @@ public boolean isConnectionAvailable() {
         ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
                 .getSystemService(Context.CONNECTIVITY_SERVICE);
 
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_NETWORK_CONNECTIVITY_CHECK)){
```
iamgusain: can be moved before L75

iamgusain (approved):

melissaahn (approved):

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -68,7 +68,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to control the timeout duration for UrlConnection read timeout.
      */
-    URL_CONNECTION_READ_TIME_OUT("UrlConnectionReadTimeOut", DEFAULT_READ_TIME_OUT_MS);
+    URL_CONNECTION_READ_TIME_OUT("UrlConnectionReadTimeOut", DEFAULT_READ_TIME_OUT_MS),
+
+    /**
+     * Flight to disable the network connectivity check.
+     */
+    DISABLE_NETWORK_CONNECTIVITY_CHECK("DisableNetworkConnectivityCheck", true);;
```
melissaahn: nit: double ; ?

shahzaibj (approved):

PR #2481, opened 2024-08-17
Adding GeneralSecurityException block to ExceptionAdapter

melissaahn:
### Summary
A couple of KeyStore classes in broker have some logic which converts a caught exception into a ClientException. By adding these specific exception filters to ExceptionAdapter, we can instead use ExceptionAdapter's clientExceptionFromException static method, which will cut down on lines and also help classify any other extraneous exceptions seen in the applicable KeyStore classes.

### Related
- https://github.com/AzureAD/ad-accounts-for-android/pull/2914

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
@@ -52,6 +52,13 @@ import com.microsoft.identity.common.java.util.StringUtil;
 import org.json.JSONException;
 
 import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.UnrecoverableEntryException;
+import java.security.cert.CertificateException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -416,6 +423,40 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
+                return new ClientException(
+                        ClientException.INVALID_PROTECTION_PARAMS,
+                        e.getMessage(),
+                        e);
+            } else if (e instanceof InvalidKeyException) {
+                return new ClientException(
+                        ClientException.INVALID_KEY,
+                        e.getMessage(),
+                        e);
+            }
```
fadidurah, line 457: I assume we can't move this to a switch statement. I think java 17 introduced a way to emulate instanceof in a switch statement, I'm not sure if that's supported in our default java version though.

fadidurah (approved):
LGTM

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
melissaahn: I think you're talking about this? https://docs.oracle.com/en/java/javase/17/language/pattern-matching-switch-expressions-and-statements.html
Looks like we would need to enable it explicitly as a preview (I did check to see if it would automatically work in our codebase). Would be cool if this made it to the next prod Java version.

shahzaibj (approved):

PR #2483, opened 2024-08-22
Remove RT from AAD token responses

rpdome:
This change would kick in only if the OneAuth/MSAL side has the maximum hello() protocol version >= 16.0 AND the flight is enabled.

Otherwise, AAD RT would still be returned.

(No change if the caller is ADAL, or older versions of MSAL/OneAuth)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -277,7 +277,7 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
                 // Cannot determine if this is AAD or MSA. Do not strip values.
                 SpanExtension.current().setAttribute(
                         AttributeName.stop_returning_rt_result.name(),
-                        "cannot_determine"
+                        "cannot_determine_account_type"
                 );
                 return false;
             } catch (final Throwable t) {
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -147,15 +187,115 @@ public Bundle bundleFromAuthenticationResult(@NonNull final ILocalAuthentication
...
+                            AttributeName.stop_returning_rt_result.name(),
+                            resultToReturn
+                    );
+                    return resultToReturn;
+                }
+
+                // Cannot determine if this is AAD or MSA. Do not strip values.
+                SpanExtension.current().setAttribute(
+                        AttributeName.stop_returning_rt_result.name(),
+                        "cannot_determine"
```
somalaya: nit : cannot_determine_account_type

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -147,15 +187,115 @@ public Bundle bundleFromAuthenticationResult(@NonNull final ILocalAuthentication
...
+            try {
+                // Remove RT if the result is not MSA... by checking the "local" TenantID (realm).
+                // (e.g. For the guest account scenario,
+                //       if the account's home tenant is A and it's guested in B...
+                //       this value will be B.
+                if (!StringUtil.isNullOrEmpty(result.getTenantId())){
+                    final boolean resultToReturn = !AzureActiveDirectoryAudience.MSA_MEGA_TENANT_ID.equalsIgnoreCase(result.getTenantId());
+                    SpanExtension.current().setAttribute(
+                            AttributeName.stop_returning_rt_result.name(),
+                            resultToReturn
```
shahzaibj: What is the data type of `stop_returning_rt_result`? It seems sometimes we are putting boolean but sometimes we are putting String? Let's be consistent and only put 1 type of data in it.

rpdome: will convert all of them to string.

shahzaibj (approved):

PR #2490, opened 2024-08-28
Allow OneAuth to set Max MSAL-Broker Protocol version

rpdome:
Previously, we've always bumped the max protocol version on the MSAL/OneAuth side by default.
(= declaring that OneAuth is aware and supports whatever change introduced by that protocol version in broker)

OneAuth should verify first that its code is compatible with the new protocol version before bumping it on their side.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
@@ -144,17 +144,7 @@ public class BrokerMsalController extends BaseController {
 
     private BrokerOperationExecutor mOperationExecutor;
 
-    @VisibleForTesting
-    public BrokerMsalController(@NonNull final Context applicationContext,
-                                @NonNull final IPlatformComponents components,
-                                @NonNull final String activeBrokerPackageName,
-                                @Nullable final List<IIpcStrategy> ipcStrategies) {
-        mComponents = components;
-        mApplicationContext = applicationContext;
-        mActiveBrokerPackageName = activeBrokerPackageName;
-        this.ipcStrategies = ipcStrategies;
-        mHelloCache = getHelloCache();
-    }
+    private String mMaxMsalBrokerProtocolVersion;
 
     public BrokerMsalController(@NonNull final Context applicationContext,
                                 @NonNull final IPlatformComponents components,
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -144,25 +144,54 @@ public class BrokerMsalController extends BaseController {
 
     private BrokerOperationExecutor mOperationExecutor;
 
-    @VisibleForTesting
+    private String mMaxMsalBrokerProtocolVersion;
+
+    public BrokerMsalController(@NonNull final Context applicationContext,
```
shahzaibj, line 149: How do we ensure that OneAuth doesn't use this constructor and uses the other one?

rpdome: I'm working with nick on this as part of the feature pre-mortem.

mohitc1:
>     String hello(final @NonNull IIpcStrategy strategy,

Can't OneAuth this overload of hello to pass clientMaxProtocolVersion?

---
Refers to: common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java:246 in 6dc6c7e. [](commit_id = 6dc6c7e4f3c17062eaf7cf0c5a1409510d77d6ae, deletion_comment = False)

mohitc1 (approved):

mohitc1:
Left comment around existing hello method. This change looks fine, though.

rpdome:
> > ```
> > String hello(final @NonNull IIpcStrategy strategy,
> > ```
> 
> Can't OneAuth this overload of hello to pass clientMaxProtocolVersion?
> 
> Refers to: common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java:246 in [6dc6c7e](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/6dc6c7e4f3c17062eaf7cf0c5a1409510d77d6ae). [](commit_id = [6dc6c7e](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/6dc6c7e4f3c17062eaf7cf0c5a1409510d77d6ae), deletion_comment = False)

Hello() is invoked as part of other operations. 

Do you mean they should inherit this class and override the method? 

They can do that, but I think the change in this PR is cleaner.

shahzaibj (approved):

PR #2493, opened 2024-09-02
MFA error handling

SammyO:
When MFA is required during a call to /token with grant_type=refresh_token, we need some native auth specific handling to inform developers on how to handle this error.

**Note**: Ideally I'd like to store the `50076` error code in a dedicated file, but I'm unsure what to call such a file, as `OAuth2ErrorCode` already exists. IMO this class name is wrong, as these are OAuth2 errors, not error codes.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
@@ -491,6 +491,7 @@ public class ExceptionAdapter {
         if (!(errorResponse instanceof MicrosoftTokenErrorResponse)) {
             return false;
         }
+
         MicrosoftTokenErrorResponse microsoftTokenErrorResponse = ((MicrosoftTokenErrorResponse) errorResponse);
         return !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getError()) &&
                 !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getSubError()) &&
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
@@ -40,6 +40,7 @@ import com.microsoft.identity.common.java.exception.UserCancelException;
 import com.microsoft.identity.common.java.logging.Logger;
 import com.microsoft.identity.common.java.net.HttpResponse;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftAuthorizationErrorResponse;
+import com.microsoft.identity.common.java.providers.microsoft.MicrosoftTokenErrorResponse;
 import com.microsoft.identity.common.java.providers.oauth2.AuthorizationErrorResponse;
 import com.microsoft.identity.common.java.providers.oauth2.AuthorizationResult;
 import com.microsoft.identity.common.java.providers.oauth2.TokenErrorResponse;
@@ -91,7 +92,7 @@ public class ExceptionAdapter {
         return exceptionFromTokenResult(result.getTokenResult(), commandParameters);
     }
 
-    public static BaseException exceptionFromAuthorizationResult(@NonNull  final AuthorizationResult authorizationResult, @Nullable final CommandParameters commandParameters) {
+    public static BaseException exceptionFromAuthorizationResult(@NonNull final AuthorizationResult authorizationResult, @Nullable final CommandParameters commandParameters) {
         final String methodTag = TAG + ":exceptionFromAuthorizationResult";
         final AuthorizationErrorResponse authorizationErrorResponse = authorizationResult.getAuthorizationErrorResponse();
         if (authorizationErrorResponse == null) {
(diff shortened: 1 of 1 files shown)


fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -471,4 +480,24 @@ private static boolean isIntunePolicyRequiredError(
...
+    private static boolean isNativeAuthenticationMFAError(
+            @NonNull final TokenErrorResponse errorResponse) {
+        if (!(errorResponse instanceof MicrosoftTokenErrorResponse)) {
+            return false;
+        }
+
+        MicrosoftTokenErrorResponse microsoftTokenErrorResponse = ((MicrosoftTokenErrorResponse) errorResponse);
+        return !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getError()) &&
+                !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getSubError()) &&
+                !microsoftTokenErrorResponse.getErrorCodes().isEmpty() &&
```
fadidurah, line 497: is it possible for getErrorCodes to return null?

fadidurah, line 497: is it possible for getErrorCodes to return null?

fadidurah (changes requested):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: Just looking out for  NPE, if this can't be null i will reapprove

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -206,8 +207,20 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
...
+                    null);
 
+            String developerDescription = "Multi-factor authentication is required, which can't be fulfilled as part of this flow. Please sign out and perform a new sign in operation. Please see exception details for more information.";
+            outErr = new ServiceException(
+                    errorResponse.getError(),
+                    developerDescription,
+                    apiError
+            );
+        }
+        else if (shouldBeConvertedToUiRequiredException(errorResponse.getError())) {
```
shahzaibj, line 223: I'm wondering if we should have a dedicated exception handler for Native Auth i.e. NativeAuthExceptionAdapter and offload this logic to that class.

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -471,4 +480,24 @@ private static boolean isIntunePolicyRequiredError(
...
+            return false;
+        }
+
+        MicrosoftTokenErrorResponse microsoftTokenErrorResponse = ((MicrosoftTokenErrorResponse) errorResponse);
+        return !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getError()) &&
+                !StringUtil.isNullOrEmpty(microsoftTokenErrorResponse.getSubError()) &&
+                !microsoftTokenErrorResponse.getErrorCodes().isEmpty() &&
+                errorResponse.getError().equalsIgnoreCase(OAuth2ErrorCode.INVALID_GRANT) &&
+                errorResponse.getSubError().equalsIgnoreCase(OAuth2SubErrorCode.BASIC_ACTION) &&
+                microsoftTokenErrorResponse.getErrorCodes().contains((long) 50076);
```
shahzaibj, line 498: Can we just check for 50076 (rather than checking for invalid_grant and basic_action)?

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
SammyO: We're (re-)using the SilentTokenCommand, so I'd have to make changes there too. So given that this is shared logic, I'd prefer to leave it as is.

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
nilo-ms, line 498: In iOS we check only that the first error code is 50076. We don't check error or suberror field

nilo-ms (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
SammyO: Updated

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
SammyO: Updated

SammyO: Updated

fadidurah (approved):

PR #2495, opened 2024-09-06
Add Audience as span attribute for AAD/MSA/Common, Fixes #2970888

tanmaymanolkar1:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2970888

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -160,18 +160,18 @@ public abstract class AzureActiveDirectoryAudience {
         mTenantId = tenantId;
     }
 
-    public static String getAudienceFromTenantId( final String tenantId) {
+    public static String getAudienceFromAuthority( final AzureActiveDirectoryAuthority authority) {
 
-        if (tenantId == null) {
+        if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
             return UNKNOWN_AUDIENCE;
         }
 
-        final String audienceToCheck = tenantId.toLowerCase(Locale.ROOT);
+        final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
 
-        if (audienceToCheck.contains(CONSUMERS) || audienceToCheck.contains(MSA_MEGA_TENANT_ID)) {
+        if (audienceToCheck.equals(CONSUMERS) || audienceToCheck.equals(MSA_MEGA_TENANT_ID)) {
             return MSA_AUDIENCE;
         }
-        else if (audienceToCheck.contains(ALL)) {
+        else if (audienceToCheck.equals(ALL)) {
             return COMMON_AUDIENCE;
         }
         else {
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -53,6 +53,12 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
+    public static final String MSA_AUDIENCE = "MSA";
+    public static final String COMMON_AUDIENCE = "COMMON";
+    public static final String AAD_AUDIENCE = "AAD";
+    public static final String UNKNOWN_AUDIENCE = "UNKNOWN";
+
+
     public String getCloudUrl() {
         if (mCloudUrl == null) {
             return AzureActiveDirectory.getDefaultCloudUrl();
@@ -152,6 +158,25 @@ public abstract class AzureActiveDirectoryAudience {
 
     public void setTenantId(String tenantId) {
         mTenantId = tenantId;
+    }
+
+    public static String getAudienceFromTenantId( final String tenantId) {
+
+        if (tenantId == null) {
+            return UNKNOWN_AUDIENCE;
+        }
+
+        final String audienceToCheck = tenantId.toLowerCase(Locale.ROOT);
+
+        if (audienceToCheck.contains(CONSUMERS) || audienceToCheck.contains(MSA_MEGA_TENANT_ID)) {
+            return MSA_AUDIENCE;
+        }
+        else if (audienceToCheck.contains(ALL)) {
+            return COMMON_AUDIENCE;
+        }
+        else {
+            return AAD_AUDIENCE;
+        }
     }
 
     public static AzureActiveDirectoryAudience getAzureActiveDirectoryAudience(final String cloudUrl,
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -56,10 +56,10 @@ public abstract class AzureActiveDirectoryAudience {
     // The values here are audience values reported in telemetry
     // rather than the values used to send to eSTS.
     public enum AudienceTelemetryConstant {
-        MSA_AUDIENCE_FOR_REPORTING,
-        COMMON_AUDIENCE_FOR_REPORTING,
-        AAD_AUDIENCE_FOR_REPORTING,
-        UNKNOWN_AUDIENCE_FOR_REPORTING;
+        MSA,
+        COMMON,
+        AAD,
+        UNKNOWN;
     }
 
 
@@ -171,10 +171,10 @@ public abstract class AzureActiveDirectoryAudience {
      * @param authority The authority from which the audience is derived.
      * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
      */
-    public static String getAudienceFromAuthority( final AzureActiveDirectoryAuthority authority) {
+    public static AudienceTelemetryConstant getAudienceFromAuthority(final AzureActiveDirectoryAuthority authority) {
 
         if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
-            return AudienceTelemetryConstant.UNKNOWN_AUDIENCE_FOR_REPORTING.name();
+            return AudienceTelemetryConstant.UNKNOWN;
         }
 
         final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
@@ -182,11 +182,11 @@ public abstract class AzureActiveDirectoryAudience {
         switch (audienceToCheck) {
             case CONSUMERS:
             case MSA_MEGA_TENANT_ID:
-                return AudienceTelemetryConstant.MSA_AUDIENCE_FOR_REPORTING.name();
+                return AudienceTelemetryConstant.MSA;
             case ALL:
-                return AudienceTelemetryConstant.COMMON_AUDIENCE_FOR_REPORTING.name();
+                return AudienceTelemetryConstant.COMMON;
             default:
-                return AudienceTelemetryConstant.AAD_AUDIENCE_FOR_REPORTING.name();
+                return AudienceTelemetryConstant.AAD;
         }
     }
 
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -177,7 +177,7 @@ public abstract class AzureActiveDirectoryAudience {
             return AudienceTelemetryConstant.UNKNOWN;
         }
 
-        final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
+        final String audienceToCheck = authority.getAudience().getTenantId().toLowerCase(Locale.ROOT);
 
         switch (audienceToCheck) {
             case CONSUMERS:
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -171,7 +171,7 @@ public abstract class AzureActiveDirectoryAudience {
      * @param authority The authority from which the audience is derived.
      * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
      */
-    public static AudienceTelemetryConstant getAudienceFromAuthority(final AzureActiveDirectoryAuthority authority) {
+    public static AudienceTelemetryConstant getAudienceForTelemetryFromAuthority(final AzureActiveDirectoryAuthority authority) {
 
         if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
             return AudienceTelemetryConstant.UNKNOWN;
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -53,15 +53,6 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
-    // The values here are audience values reported in telemetry
-    // rather than the values used to send to eSTS.
-    public enum AudienceTelemetryConstant {
-        MSA,
-        COMMON,
-        AAD,
-        UNKNOWN;
-    }
-
 
     public String getCloudUrl() {
         if (mCloudUrl == null) {
@@ -162,32 +153,6 @@ public abstract class AzureActiveDirectoryAudience {
 
     public void setTenantId(String tenantId) {
         mTenantId = tenantId;
-    }
-
-    /**
-     * Determines the audience type (MSA, Common, AAD, or Unknown) from the given
-     * {@link AzureActiveDirectoryAuthority}.
-     *
-     * @param authority The authority from which the audience is derived.
-     * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
-     */
-    public static AudienceTelemetryConstant getAudienceForTelemetryFromAuthority(final AzureActiveDirectoryAuthority authority) {
-
-        if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
-            return AudienceTelemetryConstant.UNKNOWN;
-        }
-
-        final String audienceToCheck = authority.getAudience().getTenantId().toLowerCase(Locale.ROOT);
-
-        switch (audienceToCheck) {
-            case CONSUMERS:
-            case MSA_MEGA_TENANT_ID:
-                return AudienceTelemetryConstant.MSA;
-            case ALL:
-                return AudienceTelemetryConstant.COMMON;
-            default:
-                return AudienceTelemetryConstant.AAD;
-        }
     }
 
     public static AzureActiveDirectoryAudience getAzureActiveDirectoryAudience(final String cloudUrl,
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
@@ -55,11 +55,11 @@ public abstract class AzureActiveDirectoryAudience {
 
     // The values here are audience values reported in telemetry
     // rather than the values used to send to eSTS.
-    public static class TelemetryConstants {
-        public static final String MSA_AUDIENCE_FOR_REPORTING = "MSA";
-        public static final String COMMON_AUDIENCE_FOR_REPORTING = "COMMON";
-        public static final String AAD_AUDIENCE_FOR_REPORTING = "AAD";
-        public static final String UNKNOWN_AUDIENCE_FOR_REPORTING = "UNKNOWN";
+    public enum AudienceTelemetryConstant {
+        MSA_AUDIENCE_FOR_REPORTING,
+        COMMON_AUDIENCE_FOR_REPORTING,
+        AAD_AUDIENCE_FOR_REPORTING,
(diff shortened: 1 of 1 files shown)


common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
         mTenantId = tenantId;
     }
 
+    public static String getAudienceFromTenantId( final String tenantId) {
+
+        if (tenantId == null) {
```
shahzaibj: Let's check for both null and empty string here

```suggestion
        if (StringUtil.isNullOrEmpty(tenantId)) {
```

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
...
+            return UNKNOWN_AUDIENCE;
+        }
+
+        final String audienceToCheck = tenantId.toLowerCase(Locale.ROOT);
+
+        if (audienceToCheck.contains(CONSUMERS) || audienceToCheck.contains(MSA_MEGA_TENANT_ID)) {
```
shahzaibj: Instead of  `contains` why not do an `equals` comparison?

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
         mTenantId = tenantId;
     }
 
+    public static String getAudienceFromTenantId( final String tenantId) {
```
shahzaibj: I'm wondering if we should have a helper method takes in whole authority as opposed to just tenant id and then the helper method returns the audience.

So `getAudienceFromAuthority(authority)`

I think this should simplify the invocation in the Broker PR.

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: [fbe8a84](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2495/commits/fbe8a841d3d18134489e28e519829cdbeb0fdc98)

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: [fbe8a84](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2495/commits/fbe8a841d3d18134489e28e519829cdbeb0fdc98)

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: [fbe8a84](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2495/commits/fbe8a841d3d18134489e28e519829cdbeb0fdc98)

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
         mTenantId = tenantId;
     }
 
+    public static String getAudienceFromAuthority( final AzureActiveDirectoryAuthority authority) {
+
+        if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
```
shahzaibj: Can `authority.getAudience` return null?

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
...
+            return UNKNOWN_AUDIENCE;
+        }
+
+        final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
```
shahzaibj: What if `mTenantId` is null?

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +160,25 @@ public void setTenantId(String tenantId) {
         mTenantId = tenantId;
     }
 
+    public static String getAudienceFromAuthority( final AzureActiveDirectoryAuthority authority) {
```
shahzaibj: Let's move the method to `Authority` class itself

shahzaibj: Javadoc on all public methods

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: I didnt come across such use case, but I figured its better to have, It was mentioned in todays scrum to make the Authotiry @NonNull, so I think there is a low probability that authority would be null

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: for that we have a check isNullOrEmpty on the line above

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
shahzaibj: Not the authority but I'm referring to the audience itself

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: mTenantId is a field supported in AzureActiveDirectoryAudience class hence I chose to keep the method in this class a helper method. If its fits better in authority class I can move it, if need be.

shahzaibj: I don't feel strongly about this. You could also consider writing a completely new util class in Kotlin and put this method there. There are many options. Your call.

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -53,6 +53,12 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
+    public static final String MSA_AUDIENCE = "MSA";
+    public static final String COMMON_AUDIENCE = "COMMON";
+    public static final String AAD_AUDIENCE = "AAD";
+    public static final String UNKNOWN_AUDIENCE = "UNKNOWN";
```
shahzaibj: These are basically the audience values reported in telemetry (rather than the values used to send to eSTS). I suggest we rename each of these constants appropriately such as::

`public static final String MSA_AUDIENCE_FOR_REPORTING = "MSA";`

Alternatively you can consider putting these constants in an inner class for further clarity

```java
public class TelemetryConstants { 
    public static final String MSA_AUDIENCE = "MSA";
    public static final String COMMON_AUDIENCE = "COMMON";
    public static final String AAD_AUDIENCE = "AAD";
    public static final String UNKNOWN_AUDIENCE = "UNKNOWN";
}```

File: common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java (hunk 4 above)
Comment (tanmaymanolkar1) [reply common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java]: No I didnt come across a case with Audience Null, while creating the Authority we do populate the audience field
getAzureActiveDirectoryAudience

File: common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java (hunk 6 above)
Comment (tanmaymanolkar1) [reply common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java]: for now I will let this helper method rest in the audience class for the ease of access, we can chose to move it if the need be

File: common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java (hunk 7 above)
Comment (tanmaymanolkar1) [reply common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java]: [7453c3b](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2495/commits/7453c3bf9cbe16bbc888a59094413c0fb2872d26)

File: common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java (hunk 6 above)
Comment (tanmaymanolkar1) [reply common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java]: [7453c3b](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2495/commits/7453c3bf9cbe16bbc888a59094413c0fb2872d26)

File: common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java [hunk 8]
```diff
@@ -154,6 +164,32 @@ public void setTenantId(String tenantId) {
...
+        final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
+
+        if (audienceToCheck.equals(CONSUMERS) || audienceToCheck.equals(MSA_MEGA_TENANT_ID)) {
+            return TelemetryConstants.MSA_AUDIENCE_FOR_REPORTING;
+        }
+        else if (audienceToCheck.equals(ALL)) {
+            return TelemetryConstants.COMMON_AUDIENCE_FOR_REPORTING;
+        }
+        else {
+            return TelemetryConstants.AAD_AUDIENCE_FOR_REPORTING;
```
shahzaibj: Could consider writing a switch statement

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -53,6 +53,16 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
+    // The values here are audience values reported in telemetry
+    // rather than the values used to send to eSTS.
+    public static class TelemetryConstants {
```
shahzaibj: I suggest making this an enum.

And then the method `getAudienceFromAuthority` can return the enum directly.

shahzaibj: Probably rename the enum class to `ReportedAudience` or `AudienceTelemetryConstant`

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +164,32 @@ public void setTenantId(String tenantId) {
...
     }
 
+    /**
+     * Determines the audience type (MSA, Common, AAD, or Unknown) from the given
+     * {@link AzureActiveDirectoryAuthority}.
+     *
+     * @param authority The authority from which the audience is derived.
+     * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
+     */
+    public static String getAudienceFromAuthority( final AzureActiveDirectoryAuthority authority) {
```
shahzaibj: Let's return the `AudienceTelemetryConstant` directly from the method (rather than calling `.name` ).

The callers of the method can invoke .name as needed

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -53,6 +53,16 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
+    // The values here are audience values reported in telemetry
+    // rather than the values used to send to eSTS.
+    public enum AudienceTelemetryConstant {
+        MSA_AUDIENCE_FOR_REPORTING,
+        COMMON_AUDIENCE_FOR_REPORTING,
+        AAD_AUDIENCE_FOR_REPORTING,
+        UNKNOWN_AUDIENCE_FOR_REPORTING;
```
shahzaibj: Since you are invoking `.name` you need to make sure you define the names here exactly how you'd like them in telemetry:

```suggestion
        MSA,
        COMMON,
        AAD,
        UNKNOWN_AUDIENCE;
```

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: added

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: added

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: added

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: added

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +164,32 @@ public void setTenantId(String tenantId) {
...
     }
 
+    /**
+     * Determines the audience type (MSA, Common, AAD, or Unknown) from the given
+     * {@link AzureActiveDirectoryAuthority}.
+     *
+     * @param authority The authority from which the audience is derived.
+     * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
+     */
+    public static AudienceTelemetryConstant getAudienceFromAuthority(final AzureActiveDirectoryAuthority authority) {
```
shahzaibj:
```suggestion
    public static AudienceTelemetryConstant getAudienceForTelemetryFromAuthority(final AzureActiveDirectoryAuthority authority) {
```

shahzaibj: Slight suggesttion above for method rename

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
tanmaymanolkar1: added

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +164,32 @@ public void setTenantId(String tenantId) {
...
     }
 
+    /**
+     * Determines the audience type (MSA, Common, AAD, or Unknown) from the given
+     * {@link AzureActiveDirectoryAuthority}.
+     *
+     * @param authority The authority from which the audience is derived.
+     * @return The audience type: "MSA", "Common", "AAD", or "Unknown".
+     */
+    public static AudienceTelemetryConstant getAudienceForTelemetryFromAuthority(final AzureActiveDirectoryAuthority authority) {
```
shahzaibj, line 174:
```suggestion
    public static AudienceTelemetryConstant getAudienceForTelemetryFromAuthority(final Authority authority) {
```

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -154,6 +164,32 @@ public void setTenantId(String tenantId) {
...
+
+        if (authority == null || StringUtil.isNullOrEmpty(authority.getAudience().getTenantId())) {
+            return AudienceTelemetryConstant.UNKNOWN;
+        }
+
+        final String audienceToCheck = authority.getAudience().mTenantId.toLowerCase(Locale.ROOT);
```
mohitc1: >mTenantId [](http://example.com/codeflow?start=63&length=9)

getTenantId()

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -53,6 +53,16 @@ public abstract class AzureActiveDirectoryAudience {
     public static final String ALL = "common";
     public static final String MSA_MEGA_TENANT_ID = "9188040d-6c67-4c5b-b112-36a304b66dad";
 
+    // The values here are audience values reported in telemetry
+    // rather than the values used to send to eSTS.
+    public enum AudienceTelemetryConstant {
+        MSA,
+        COMMON,
+        AAD,
+        UNKNOWN;
```
mohitc1: let's have one more: TENANTED, when there's target tenant in authority

So,

COMMON,
MSA,
ORGANIZATIONS,
TENANTED

shahzaibj: Wouldn't tenanted be AAD?

shahzaibj: I think the goal here is to identify which service we are talking to and be able to easily identify that in telemetry. With your suggestion @mohitc1 I think it is better engineering data but needs the reader to know how to interpret it i.e. reader needs to know that tenanted and organizations represent AAD

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
```diff
@@ -53,6 +53,16 @@ public abstract class AzureActiveDirectoryAudience {
...
+    }
```
mohitc1: I think we should move this and the method to broker itself. Telemetry is for broker only. 

At this layer it's extra stuff present in AzureActiveDirectoryAudience

shahzaibj: +1

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
mohitc1: We can choose another name like SINGLE_ORGANIZATION (similar to Audience class names). This for completeness. 
We don't need to comeback and update this later if we realize need of /organizations vs /<tenantid> in future.

shahzaibj: I think there's actually 3 different types of things can track and emit:

1. Service we are talking to. Possible values:
- MSA
- AAD

2. Audience type in authority. Possible values:
- COMMON
- MSA
- ORGANIZATIONS
- TENANTED

3. The actual authority (OII). Possible values:
- login.micosoftonline.com/common
- login.microsoftonine.us/common
- login.microsoftonine.cn/common
- login.microsoftonline.com/consumers
- login.microsoftonline.com/organizations
- login.microsoftonline.com/<MSA_MEGA_TENANT_ID>
- login.micorsoftonline.com<SOME_AAD_TENANT_ID>


My 2 cents, I think there is value in no.3 because this will help us know actually which endpoint we are talking to and could be super helpful for debugging issues. It can even help track if we are talking to a different cloud and which one specifically and also help track tenant id for guest scenarios.

Number 1 is more for tracking quickly which service we are talking to (AAD vs MSA) for things such as easily filtering out MSA no_account_found errors as expected.

The number 2 is somewhere between 1 and 3 i.e. it is somewhat product information but also somewhat debugging information. I'm not too sure if we need all 3 as some may be overkill. I'm thinking we should do no. 3 and then either one of no. 1 and no. 2. 

Thoughts? @mohitc1 @tanmaymanolkar1

mohitc1: Agree we need all to capture the request target completely. But we need not capture all these separately. 
We can optimize on what we capture.
Option 1:
If we capture just two fields
Cloud: login.microsoftonline.com, login.microsoftonine.cn, login.microsoftonine.us (or any shortened form to save data)
AND
(Audience type without OII: COMMON, ORGANIZATIONS, SINGLE_ORGANIZATION, MSA, MSA_MEGA_TENANT
OR audience type with OII: COMMON, ORGANIZATIONS, MSA, MSA_MEGA_TENANT, <Some TenantId>)

Option2:
We simply capture authority URL e.g. login.microsoftline.com/<value>.. everything else can be derived from this. This would be OII.

tanmaymanolkar1:
Closing this PR as I will move the changes to broker

common4j/src/main/com/microsoft/identity/common/java/authorities/AzureActiveDirectoryAudience.java
shahzaibj: My worry with Option 2 is just not knowing how much it will add to complexity in the telemetry backend. We can write helper functions to derive these things but need to make sure that those helper functions get integrated into any materialized views etc and not sure if there would be any perf impact. @tanmaymanolkar1 You may want to do a quick analysis about how easy this will be on the backend.

PR #2500, opened 2024-09-13
Release 17.7.2 hot fix

rpdome:
Backporting #2480


fadidurah (approved):
Should we provide some more detail in PR description about what the hot fix is doing?

PR #2501, opened 2024-09-13
Remove triggers from pipelines and execute ABID validator on label change

p3dr0rv:
Something changed on ADO and that caused the pipeline to stop running on PR, the solution was to remove the trigger tag.

If you didn't specify any triggers, and the [Disable implied YAML CI trigger](https://learn.microsoft.com/en-us/azure/devops/release-notes/2023/sprint-227-update#prevent-unintended-pipeline-runs) setting is not enabled, the default is as if you wrote:
````
trigger:
  branches:
    include:
    - '*'  # must quote since "*" is a YAML reserved character; we want a string

### DIFF
--- a/.github/workflows/validate-pr-ab-id.yml
+++ b/.github/workflows/validate-pr-ab-id.yml
@@ -7,7 +7,7 @@
 name: 'AB#ID Check'
 on: 
   pull_request:
-    types: [opened, reopened, edited]
+    types: [opened, reopened, edited, labeled]
 
 jobs:
 # This action checks your pull request to make sure it is linked to a work item using AB# before you can merge.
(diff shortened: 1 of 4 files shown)


SammyO:
@p3dr0rv do we know why this was introduced and thus if we're okay to remove these checks? To rephrase: are we sure we're not disabling some new (security?) initiative?

fadidurah (approved):

p3dr0rv:
> @p3dr0rv do we know why this was introduced and thus if we're okay to remove these checks? To rephrase: are we sure we're not disabling some new (security?) initiative?

I think it's ok to remove this section, I wrote these files, and they worked at the time. Now it seems the logic for triggering the pipes is different.

tanmaymanolkar1 (approved):

PR #2503, opened 2024-09-16
Update device registration helper methods for test fixes with local flight for keystore keys enabled, Fixes AB#3039504

iamgusain:
Changes to fix tests for local flight run when keystore key flight is enabled. (Install cert operation is not supported separately with keystore keys flight enabled)
- When Authenticator is the active broker perform device registration with Authenticator settings UI instead of Device settings 
- When adding account from Device settings page handle install cert UI only when it appears, (it doesn't appear when keystore keys flight is enabled, in that case just press ok to complete the action)

[AB#3039504](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3039504)


somalaya:
Why would the test fail specifically when Auth app is the active broker? I mean why skip device reg from settings page in this case?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
```diff
@@ -47,7 +46,7 @@
 public class BrokerAuthenticatorUpdatedVersionImpl extends BrokerMicrosoftAuthenticator {
 
     private static final String TAG = BrokerAuthenticatorUpdatedVersionImpl.class.getSimpleName();
-    public static boolean shouldUseDeviceSettingsPage = true;
+    public static boolean shouldUseDeviceSettingsPage = false;
```
fadidurah, line 49: Do we have some validation to make sure this doesn't affect other test cases?

fadidurah (approved):

fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerAuthenticatorUpdatedVersionImpl.java
iamgusain: I ran multiple tests locally with this change succesfully.

mohitc1 (approved):

PR #2505, opened 2024-09-18
Remove accounts from broker span

somalaya:


shahzaibj:
The title of the PR is interesting. It seems like you are removing accounts from inside of a span object.

PR #2508, opened 2024-09-20
Use legacy FIDO2 API to provide security key support to Android 13 devices, Fixes AB#3027082

melissaahn:
### Summary
Android's [Credential Manager API](https://developer.android.com/identity/sign-in/credential-manager) does not have FIDO2 security key support for Android versions 13 and below. They recommend using their [legacy FIDO2 API](https://developers.google.com/identity/fido/android/native-apps) instead. Thus, this PR includes and uses the legacy FIDO2 API in order to provide FIDO2 security key support for our customers using Android 13 and below.

As discussed with the PMs, we are going to use this API as a fallback from an attempt with CredMan in the case where a `NoCredentialException` is thrown AND the device OS is Android 13 or lower. 
As a supplement, we're setting `preferImmediatelyAvailableCredentials` in the CredMan `getCredentialRequest` object to true if the device OS is Android 13 or lower. This ensures the behavior where no dialog from CredMan is shown if no passkey cred is present. The end goal is for an Android <= 13 user who only has a security key to see one dialog which will allow them to authenticate.

This PR includes some new classes in the `fido` package. I attempted to draw a diagram to explain the relation:
image

The legacy FIDO2 API provides a `PendingIntent` which the host app needs to start, and when the resulting Activity closes is when we get our results. This is different from CredMan's exposed endpoints, and thus I tried to wrap the legacy API in such a way which mimics the behavior of CredMan.

Some unit tests have been added, and I am going to continue to add unit tests this month, but planning on splitting this up into another PR since this PR is already on the longer side. (Shadow and extended classes will be needed)

There is a flight variable guarding the feature, but since the logic is in common, the `CommonFlight` variable is set to true by default, while the `BrokerFlight` variable is set to false by default.

[AB#3027082](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3027082)


rpdome:
As discussed with the PMs, we are going to use this API as a fallback from an attempt with CredMan in the case where a NoCredentialException is thrown AND the device OS is Android 13 or lower.


<< (Assuming that Crewman on <=13 is always going to fail) why not just do the OS check?

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -23,28 +23,49 @@
...
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager
+import com.microsoft.identity.common.java.opentelemetry.AttributeName
+import com.microsoft.identity.common.logging.Logger
+import io.opentelemetry.api.trace.Span
 
 /**
  * Makes calls to the Android Credential Manager API in order to return an attestation.
  */
-class CredManFidoManager (val context: Context) : IFidoManager {
+class CredManFidoManager (val context: Context,
```
rpdome, line 42: can we use the OS version check to determine if CredManFidoManager or LegacyFido2ApiManager should be returned?

(no need to invoke CredManFidoManager and catch exception if we know for sure it's going to fail)

rpdome: (e.g. put it in a factory 


```kotlin
object FidoManagerFactory { // or put it in a companion object somewhere.
    fun getFidoManager() : IFidoManager {
      if (flightEnabled && os<=13){
            return LegacyFido2ApiManager()
      }

     return CredManFidoManager()
   }
}

```

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,109 @@
...
+     * @param userVerificationPolicy User verification policy string.
+     * @param span OpenTelemetry span.
+     * @return assertion
+     */
+    override suspend fun authenticate(challenge: String,
+                                      relyingPartyIdentifier: String,
+                                      allowedCredentials: List<String>?,
+                                      userVerificationPolicy: String,
+                                      span: Span):
+            String = suspendCancellableCoroutine { continuation ->
```
rpdome: nit: please rearrange indentation of each block level for readability.

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,109 @@
...
+        val methodTag = "$TAG:authenticate"
+        val canceller = CancellationSignal()
+        continuation.invokeOnCancellation { canceller.cancel() }
+        span.setAttribute(
+            AttributeName.fido_manager.name,
+            TAG
+        )
+        val publicKeyCredentialDescriptorList = ArrayList<PublicKeyCredentialDescriptor>()
```
rpdome: nit:  

```Kotlin
val publicKeyCredentialDescriptorList = allowedCredentials?.map {
     PublicKeyCredentialDescriptor("public-key", it.toByteArray(), ArrayList())
}
```

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+        val requestOptions = com.google.android.gms.fido.fido2.api.common.PublicKeyCredentialRequestOptions.Builder()
+            .setChallenge(challenge.toByteArray())
+            .setRpId(relyingPartyIdentifier)
+            .setAllowList(publicKeyCredentialDescriptorList)
+            .build()
+        Logger.info(methodTag, "Calling the legacy FIDO2 API with public key credential options to get a PendingIntent.")
+        val result = legacyApi.getSignPendingIntent(requestOptions)
+
+        result.addOnSuccessListener(OnSuccessListener { pendingIntent ->
```
rpdome, line 78: onError?

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,109 @@
...
+import io.opentelemetry.api.trace.Span
+import kotlinx.coroutines.suspendCancellableCoroutine
+import java.util.ArrayList
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+/**
+ * Utilizes the legacy Android FIDO2 API in order to return an attestation.
+ * Note that the legacy FIDO2 API should only be used for older Android versions which need it,
+ * such as Android 13 and below for FIDO2 security key support.
```
rpdome, line 42: Does the Legacy code have min OS version requirement?

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        final PublicKeyCredential credential = PublicKeyCredential.deserializeFromBytes(bytes);
+        final AuthenticatorResponse response = credential.getResponse();
+        if (response instanceof AuthenticatorErrorResponse) {
+            final String errorMessage = ((AuthenticatorErrorResponse) response).getErrorMessage();
+            final String errorCode = String.valueOf(((AuthenticatorErrorResponse) response).getErrorCode());
+            errorCallback.invoke(
+                    new LegacyFido2ApiException(
+                            errorCode,
+                            errorMessage != null ? errorMessage : "AuthenticatorResponse has a null error message."
```
rpdome: nit: StringUtil.isNullOrEmpty()

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
```diff
@@ -0,0 +1,151 @@
...
+        final byte[] bytes = intent.getByteArrayExtra(Fido.FIDO2_KEY_CREDENTIAL_EXTRA);
+        if (bytes == null) {
+            errorCallback.invoke(
+                    new LegacyFido2ApiException(
+                            LegacyFido2ApiException.NULL_OBJECT,
+                            "Credential result from Intent is null."
+                    ));
+            return null;
+        }
+        final PublicKeyCredential credential = PublicKeyCredential.deserializeFromBytes(bytes);
```
rpdome: nit: extra line after each "return null" block for readability.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -115,13 +117,24 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
...
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        final String methodTag = TAG + ":onCreate";
         final FragmentActivity activity = getActivity();
         if (activity != null) {
             WebViewUtil.setDataDirectorySuffix(activity.getApplicationContext());
         }
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_LEGACY_FIDO_SECURITY_KEY_LOGIC)) {
+            mFidoLauncher = registerForActivityResult(
```
rpdome, line 133: mFidoLauncher is technically needed only in <= 13, right?

common/build.gradle
```diff
@@ -169,6 +169,7 @@ dependencies {
     implementation "androidx.credentials:credentials:$rootProject.ext.AndroidCredentialsVersion"
     //needed for credentials support from play services, for devices running
```
rpdome, line 170: How would this behave on non-GMS devices?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -475,6 +491,10 @@ private HashMap<String, String> getRequestHeaders(final Bundle state) {
         }
     }
 
+    public ActivityResultLauncher<LegacyFido2ApiObject> getFidoLauncher() {
```
rpdome, line 501: where is this invoked?

rpdome (commented):
Left a few comments. Still trying to wrap my head around the flow.

melissaahn:
> As discussed with the PMs, we are going to use this API as a fallback from an attempt with CredMan in the case where a NoCredentialException is thrown AND the device OS is Android 13 or lower.
> 
> << (Assuming that Crewman on <=13 is always going to fail) why not just do the OS check?

@rpdome This is actually something I'm trying to ask Android about.
Context is that CredMan on <=13 does not always fail- it's supposed to work for GPM passkeys. And on the FIDO2 legacy API docs, they recommend using CredMan for passkey purposes. Thing is, I do think GPM passkeys appear as an option in the legacy API UI dialog. I don't know if it actually works, due to a lack of Android 13 devices and some server side bugs, but I agree that if it can be asserted that it works on Android 13 and below, it would be much simpler to do an OS check.

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: (Gave answer in one of your other comments)

common/build.gradle
melissaahn: (if you're asking about line 172, the legacy API), my best guess is that we're going to get an error response back from the API, which the client logic will turn into an exception to be returned back to the server side.

rpdome: We ship brokers on nGMS devices, so I want to make sure we're not breaking those devices.

I asked this question because "play services" is mentioned here.

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
melissaahn: Android's docs don't indicate one, but my guess would be Android 9, since that's CredMan's passkey min. We do have an OS check for this that won't even send the extra qs param if the version is less than 9: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java#L250

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
melissaahn: Whoops; added onFailure and onCancelled listeners: cef6e97bf8ae3da4badd6b59bcbfed4e63358aba

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
melissaahn: cef6e97bf8ae3da4badd6b59bcbfed4e63358aba

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: True; I anticipate I will need to add in some null checks, so will implement this tomorrow

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/cef6e97bf8ae3da4badd6b59bcbfed4e63358aba/common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt#L89

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
melissaahn: Hopefully the change in this commit helps: cef6e97bf8ae3da4badd6b59bcbfed4e63358aba

common/build.gradle
melissaahn: I'll ask Intune and some other internal folks about this. I do think some nGMS customers are also interested in this feature, so you're right, this is a pretty relevant question.

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiException.java
```diff
@@ -0,0 +1,53 @@
...
+package com.microsoft.identity.common.internal.fido;
+
+import com.microsoft.identity.common.java.exception.BaseException;
+
+/**
+ * An exception class which holds attributes from the legacy FIDO2 API error responses.
+ */
+public class LegacyFido2ApiException extends BaseException {
+
+    public static final String NULL_OBJECT = "null_object";
```
shahzaibj: I think this constant was already defined somewhere in the code

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiException.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static final String BAD_ACTIVITY_RESULT_CODE = "bad_activity_result_code";
+
+    public static final String UNKNOWN_ERROR = "unknown_error";
```
shahzaibj: I think this constant was already defined somewhere in the code

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiException.java
melissaahn: null_object is also defined in ClientException, but LegacyFido2ApiException is different exception class which extends BaseException, so I'm not sure that these two exceptions should share strings.

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiException.java
melissaahn: (Replied to other comment)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: 086326adf883f7e09057ea2de33b10caf9745529

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -55,13 +76,32 @@ class CredManFidoManager (val context: Context) : IFidoManager {
             requestJson = requestJson
         )
         val getCredRequest = GetCredentialRequest(
```
rpdome, line 80: nit: add comment here (e.g. "we're setting preferImmediatelyAvailableCredentials in the CredMan getCredentialRequest object to true if the device OS is Android 13 or lower. This ensures the behavior where no dialog from CredMan is shown if no passkey cred is present. The end goal is for an Android <= 13 user who only has a security key to see one dialog which will allow them to authenticate.")

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,142 @@
...
+            .setChallenge(challenge.toByteArray(Charsets.UTF_8))
+            .setRpId(relyingPartyIdentifier)
+            .build()
+        Logger.info(methodTag, "Calling the legacy FIDO2 API with public key credential options to get a PendingIntent.")
+        val result = legacyApi.getSignPendingIntent(requestOptions)
+
+        result.addOnSuccessListener(OnSuccessListener { pendingIntent ->
+            if (pendingIntent != null) {
+                Logger.info(methodTag, "Launching the legacy FIDO2 API PendingIntent.")
+                fragment.fidoLauncher?.launch(
```
rpdome: nit: should be able to do
``` Kotlin

if (fragment.fidoLauncher != null){
    // Kotlin's compiler should automatically cast nullable for you.
     fragment.fidoLauncher.launch{
     }
} else {
    
}
```

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
```diff
@@ -0,0 +1,142 @@
...
+                    "Returned PendingIntent from legacy API is null."
+                )
+                Logger.error(methodTag, e.message, e)
+                if (continuation.isActive) {
+                    continuation.resumeWithException(e)
+                }
+            }
+        })
+        result.addOnFailureListener(OnFailureListener { exception ->
+            Logger.error(methodTag, "Failed to get a PendingIntent from the legacy FIDO2 API.", exception)
```
rpdome: nit: maybe this can be extracted into a helper function

```Kotlin
fun logAndReturnError(val continuation, val error){
   Logger.error(methodTag, error.message, e)
   if (continuation.isActive) {
      continuation.resumeWithException(error)
   }
}
```

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
```diff
@@ -0,0 +1,157 @@
...
+import com.google.android.gms.fido.fido2.api.common.PublicKeyCredential;
+import com.microsoft.identity.common.java.util.StringUtil;
+
+import kotlin.Unit;
+import kotlin.jvm.functions.Function1;
+
+/**
+ * An implementation of an ActivityResultContract which helps run the PendingIntent from the legacy FIDO2 API and connect its results to the FIDO challenge handlers and managers.
+ */
+public class LegacyFidoActivityResultContract extends ActivityResultContract<LegacyFido2ApiObject, Void> {
```
rpdome: qq: any reason why this was written in java? (as opposed to kotlin like the rest of the flow)? any specific limitation?

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
rpdome: (also... for my knowledge. Is there any particular reason you surround every .resume() with isActive check? 

Did you run into any errors without it? or is this just added as a precaution?

I never use this class, and the sample doesn't seem to have that. (hence the Q)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -115,13 +117,25 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
 
     private PermissionRequest mCameraPermissionRequest;
 
+    private ActivityResultLauncher<LegacyFido2ApiObject> mFidoLauncher;
```
rpdome, line 121: nit: add a comment that this would be invoked (required) by the legacy fido class?

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
rpdome: (If possible, I'd prefer us writing new code in kotlin so that we could utilize benefits of modern languages. (e.g. NPE should no longer happen if the entire codebase is in kotlin))

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: be71af20c85ba1d393b2f85e273871c29b9bdcef

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: be71af20c85ba1d393b2f85e273871c29b9bdcef

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
melissaahn: I had to set fragment.fidoLauncher to a local variable because it's an open value, but after doing that this worked. (be71af20c85ba1d393b2f85e273871c29b9bdcef)

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiManager.kt
melissaahn: Extracted into a helper method as suggested.
The isActive checks are added as a precaution since the logic lies in callbacks. It hopefully will protect against any possibility in which the coroutine gets cancelled, but the Intent is active before then, and once it finishes it calls the callbacks.

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFidoActivityResultContract.java
melissaahn: Done: 5341b318edaa29bebb75fa11da54f76c21b2ba03

common/src/main/java/com/microsoft/identity/common/internal/fido/LegacyFido2ApiException.java
shahzaibj: I think there is a file called as `ErrorStrings.java` (or something like that). We should define all common consts in that file

melissaahn: Done: 6922a39600e0ade06e818ae2f3dbdf856258b473

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -23,28 +23,49 @@
...
+import androidx.credentials.exceptions.NoCredentialException
+import com.microsoft.identity.common.internal.providers.oauth2.WebViewAuthorizationFragment
+import com.microsoft.identity.common.java.flighting.CommonFlight
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager
+import com.microsoft.identity.common.java.opentelemetry.AttributeName
+import com.microsoft.identity.common.logging.Logger
+import io.opentelemetry.api.trace.Span
 
 /**
  * Makes calls to the Android Credential Manager API in order to return an attestation.
```
mohitc1, line 37: Update doc

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -23,28 +23,49 @@
...
  */
-class CredManFidoManager (val context: Context) : IFidoManager {
+class CredManFidoManager (val context: Context,
+                          val fragment: WebViewAuthorizationFragment) : IFidoManager {
```
mohitc1: >val fragment: WebViewAuthorizationFragment [](http://example.com/codeflow?start=26&length=42)

suggestion:
2nd argument should be val legacyFallbackManager: LegacyFido2ApiManager? and set it only if flight is on and build <= 13

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -23,28 +23,49 @@
...
+
+    val TAG = CredManFidoManager::class.simpleName.toString()
 
     val credentialManager = CredentialManager.create(context)
```
mohitc1, line 50: private

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
```diff
@@ -23,28 +23,49 @@
...
+import io.opentelemetry.api.trace.Span
 
 /**
  * Makes calls to the Android Credential Manager API in order to return an attestation.
  */
-class CredManFidoManager (val context: Context) : IFidoManager {
+class CredManFidoManager (val context: Context,
+                          val fragment: WebViewAuthorizationFragment) : IFidoManager {
+
+    val TAG = CredManFidoManager::class.simpleName.toString()
```
mohitc1: > val TAG = CredManFidoManager::class.simpleName.toString() [](http://example.com/codeflow?start=3&length=58)

TAG in companion object

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -115,13 +117,26 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
...
             WebViewUtil.setDataDirectorySuffix(activity.getApplicationContext());
         }
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_LEGACY_FIDO_SECURITY_KEY_LOGIC)
+                && Build.VERSION.SDK_INT < Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            mFidoLauncher = registerForActivityResult(
+                    new LegacyFidoActivityResultContract(),
+                    result -> {
+                        Logger.info(methodTag, "Legacy FIDO2 API result received.");
+                    }
+            );
```
mohitc1, line 138: Can you register and manage launcher entirely in LegacyFido2ApiManager? .e.g call in registerForActivityResult(
                    new LegacyFidoActivityResultContract(),
                    result -> {
                        Logger.info(methodTag, "Legacy FIDO2 API result received.");
                    }

in LegacyFido2ApiManager construction

mohitc1:
Left a few suggestions. None blocking

mohitc1 (approved):
Left a few comments. None blocking

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
melissaahn: Unfortunately no (I did try before, but it results in a crash). This is because registerForActivityResult needs to be called before the fragment is created: https://developer.android.com/training/basics/intents/result#register

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: 37b32956aa6aa6df1f728faa590b0f9672a908a8

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: 37b32956aa6aa6df1f728faa590b0f9672a908a8

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: makes sense. Made the changes: 37b32956aa6aa6df1f728faa590b0f9672a908a8

common/src/main/java/com/microsoft/identity/common/internal/fido/CredManFidoManager.kt
melissaahn: 9184ae56b89a37a1aca1e8aa1aad13551d3a29a1

PR #2516, opened 2024-10-04
Add Child Spans for Interactive Span

tanmaymanolkar1:
Helper PR for : https://github.com/AzureAD/ad-accounts-for-android/pull/2970


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -33,6 +33,8 @@ public enum SpanName {
     GetAccounts,
     RemoveAccount,
     WorkplaceJoin,
+    preInterruptInteractive,
+    preInterruptSilent,
```
mohitc1: These names don't necessarily reflect their what they are being used for.

We should have better names

tanmaymanolkar1: @mohitc1 any suggestions?

tanmaymanolkar1: renamed it to ATIInteractively and ATISilently,

mohitc1: It would be good to add comment describing what it's used for.

mohitc1 (approved):
:shipit:

shahzaibj (approved):
Not a fan of these names but I guess this is fine

PR #2521, opened 2024-10-10
Clean up logs

rpdome:
Remove unnecessary logs.

Related: [Remove unneeded logs by rpdome · Pull Request #2974 · AzureAD/ad-accounts-for-android (github.com)](https://github.com/AzureAD/ad-accounts-for-android/pull/2974)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
@@ -201,13 +201,13 @@ public class AzureActiveDirectory
             // until the app is killed.
             HttpCache.flush();
 
-            Logger.info(TAG + methodName, "Parsing response.");
-            AzureActiveDirectoryInstanceResponse instanceResponse =
+            final AzureActiveDirectoryInstanceResponse instanceResponse =
                     ObjectMapper.deserializeJsonStringToObject(
                             response.getBody(),
                             AzureActiveDirectoryInstanceResponse.class
                     );
-            Logger.info(TAG + methodName, "Discovered ["
+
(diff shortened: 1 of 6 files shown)


common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -201,13 +201,13 @@ public static synchronized void performCloudDiscovery()
...
-            Logger.info(TAG + methodName, "Parsing response.");
-            AzureActiveDirectoryInstanceResponse instanceResponse =
+            final AzureActiveDirectoryInstanceResponse instanceResponse =
                     ObjectMapper.deserializeJsonStringToObject(
                             response.getBody(),
                             AzureActiveDirectoryInstanceResponse.class
                     );
-            Logger.info(TAG + methodName, "Discovered ["
+
+            Logger.verbose(TAG + methodName, "Discovered ["
```
fadidurah, line 210: My only issue with making these verbose is that CP includes verbose logs, so we'd still get these if CP is the active broker. We should probably tackle that issue by asking CP to change their logging level though.

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
rpdome: My understanding is that CP already disables verbose log in PROD build by default, but i'll double check.

rpdome: Confirmed with CP. They now have INFO as the default.

iamgusain (approved):

PR #2524, opened 2024-10-16
Camera permission request behavior changes for QR + PIN Auth, Fixes AB#3032520

p3dr0rv:
There is a new CELA requirement to change the behavior of the camera permission request for QR + PIN Auth.

The current behavior is that once we get the camera permission for the app we never ask again.
The change is to ask every time we need to use the camera.

In addition to this change we move some functionality to the new class CameraPermissionRequest and did some update on the name of the functions.

[AB#3032520](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3032520)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequest.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        isGranted = true;
+    }
+
+    /**
+     * Call this method to deny the permission to access the camera resource.
+     * <p>
+     * Note: This method is only available on API level 21 or higher.
+     */
+    void denny() {
```
rpdome: nit: deny

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequest.java
```diff
@@ -0,0 +1,103 @@
...
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.common.logging.Logger;
+
+/**
+ * Class responsible for handling camera permission requests.
+ * Note: This class is compatible only with API level 21 and above;
+ * functionality will not behave as expected on lower versions.
+ */
+public class CameraPermissionRequest {
```
rpdome: nit: please consider writing new classes in kotlin.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -276,29 +276,26 @@ public void onPermissionRequest(final PermissionRequest request) {
...
                 Logger.info(methodTag, "New camera request.");
-                mCameraPermissionRequest = request;
-                if (isCameraPermissionGranted()) {
+                mCameraPermissionRequest = new CameraPermissionRequest(request);
+                if (isAppCameraPermissionGranted()) {
                     Logger.info(methodTag, "Camera permission already granted.");
-                    acceptCameraRequest();
-                } else if (shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)) {
-                    Logger.info(methodTag, "Show camera rationale.");
                     showCameraRationale();
```
rpdome, line 299: are we supposed to ask just for the permission (1 dialog) or both rationale + permission (2 dialogs) every time?

rpdome: also... if isCameraPermissionGranted == false, then we don't show the rationale? we just pop the camera request activity?

p3dr0rv: 1
The idea with this change, it is to prompt always and prompt once (either system prompt or rationale).
I believe is a better user experience just to jump through one prompt instead of two.

2
if the permission is not granted, we show the system prompt. (Following the same principle mentioned before, I do not see much value in showing both the rationale + the system prompt)

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequest.java
p3dr0rv: done

tanmaymanolkar1 (approved):

shahzaibj:
Didn't the Android OS have a way for the user to configure when they provide permission by saying `This time only`, `Always`, and `Never`? 

The work you are doing here, does that take away from user being able to choose `Always`?

p3dr0rv:
> Didn't the Android OS have a way for the user to configure when they provide permission by saying `This time only`, `Always`, and `Never`?
> 
> The work you are doing here, does that take away from user being able to choose `Always`?

1 Yes, the OS provide some functionality like this at app level. so, the first time the app needs to access the camera resource it will ask for permissions, then the user can decide if it grant, _while using the app_ , _just this time_ or _deny_ the request.

2 Yes, this change here removes this behavior. why? because this type of QR + PIN flows are for SDM (now). So, the first user can grant the camera permission `Always` , so when another user come it will not be asked. And CELA requested to all users been prompted.

Note:
iOS has the same behavior.
In the future we might add a flag controlled by the admin to remove this behavior and respect the first user choice, but nothing has been decided yet.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
rpdome: would you mind adding that as a comment here?

rpdome: would you mind adding that as a comment here?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -277,29 +277,30 @@ public void onPermissionRequest(final PermissionRequest request) {
...
                     return;
                 }
                 Logger.info(methodTag, "New camera request.");
-                mCameraPermissionRequest = request;
-                if (isCameraPermissionGranted()) {
+                mCameraPermissionRequest = new CameraPermissionRequest(request);
+                // The current camera behavior is based on the principle of only prompt once and prompt always.
+                // If the OS level permission is not granted, we show the system prompt.
+                // If the OS level permission was granted previously, we show the rationale to confirm the consent witth the current user.
+                // If the OS level permission was denied previously, we just denny the request.
```
rpdome: nit: denny -> deny

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -277,29 +277,30 @@ public void onPermissionRequest(final PermissionRequest request) {
...
+                    handleRepeatedCameraRequests(request);
                     return;
                 }
                 Logger.info(methodTag, "New camera request.");
-                mCameraPermissionRequest = request;
-                if (isCameraPermissionGranted()) {
+                mCameraPermissionRequest = new CameraPermissionRequest(request);
+                // The current camera behavior is based on the principle of only prompt once and prompt always.
+                // If the OS level permission is not granted, we show the system prompt.
+                // If the OS level permission was granted previously, we show the rationale to confirm the consent witth the current user.
```
rpdome: nit:

- If the OS level permission was granted previously, we show the rationale to confirm the consent witth the current user.
- Otherwise, show the system prompt.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
rpdome: also, is there any way to programmatically reset this? (let's say if the user accidentally denies. how do we help them fix that? do we show something to guide them?)

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: This is handled by ESTS UX, if the system permission has been permanent denied (Deny, don't ask gain). meaning the user has denied the system permission multiple times. 
The only way to restore the permission is through the app info. So, ESTS UX will notify the user that it cannot proceed and that he needs to enable the permission on the app. Notice that the message is generic because the behavior is the same in iOS, Android and Windows.

p3dr0rv: We could show a dialog box and send the user to the app info so he can unblock this, but the Android docs disencourage this behavior

https://developer.android.com/training/permissions/requesting
// Explain to the user that the feature is unavailable because
                // the feature requires a permission that the user has denied.
                // At the same time, respect the user's decision. Don't link to
                // system settings in an effort to convince the user to change
                // their decision.

PR #2509, opened 2024-09-24
Check inbox code validity in Native Auth E2E tests, Fixes AB#3037823

rmccahill:
This PR updates the logic for retrieving a code from the test inbox - the code retrieved will be checked in the calling test to ensure that the retrieved code is valid, since we occasionally get expired codes from the inbox. If a code is invalid, the test is retried, filtering out the emails previously retrieved to ensure we're getting the latest email. If we don't get a valid code after 3 tries, an IllegalStateException will be thrown. The E2E tests will use this new function in this PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/compare/robert/retry-submit-code-e2e-tests

Work Item: [AB#3037823](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3037823)

PR #2527, opened 2024-10-29
Check out from Robert because of github policy, Fixes AB#3037823

Yuki-YuXin:
The original PR is https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2509 , which is blocked by github policy because the author left.

Work Item: [AB#3037823](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3037823)


spetrescu84 (approved):

GBakolas (approved):

nilo-ms (approved):

testutils/src/main/java/com/microsoft/identity/internal/testutils/nativeauth/api/TemporaryEmailService.kt
```diff
@@ -57,32 +67,47 @@ class TemporaryEmailService {
...
+                    .filter { Duration.between(LocalDateTime.parse(it.date, dateFormatter), currentTime).seconds < newEmailCutoff }
+                    .map { it.id }
+                    .firstOrNull()
+
+                if (newEmailId != null) {
+                    latestEmailId = newEmailId
+
+                    val emailContent = api.retrieveEmail(emailAddress, latestEmailId)
+                    otpValue = retrieveOtpFromEmailBody(emailContent.textBody)
+                    validCodeRetrieved = true
```
SammyO, line 96: we don't know if the code is valid, right? Not until the API has accepted it.

testutils/src/main/java/com/microsoft/identity/internal/testutils/nativeauth/api/TemporaryEmailService.kt
```diff
@@ -57,32 +67,47 @@ class TemporaryEmailService {
...
+        while (count < numberOfRetries && !validCodeRetrieved) {
+            try {
+                //wait before calling the email endpoint
+                Thread.sleep(retryDelayMs)
+
+                //API returns dates in the UTC timezone, so system time should also be converted
+                val currentTime = ZonedDateTime.ofInstant(Instant.now(), ZoneOffset.UTC)
+
+                val inboxEmails = api.retrieveMailbox(emailAddress)
+                val newEmailId = inboxEmails
```
SammyO, line 86: You may want to consider checking if the new email ID is different than the previous one, as there is no purpose in retrying with the same email and thus same OTP.

testutils/src/main/java/com/microsoft/identity/internal/testutils/nativeauth/api/TemporaryEmailService.kt
Yuki-YuXin: Yes, you are right. The ideal definition of valid code is the code accepted by the API. The current definition of valid code here is if the service successfully filters out/finds the code.

The temp email service is a whole thing that we use to get otp.
```
val otp = tempEmailApi.retrieveCodeFromInbox(user)
val submitCodeResult = (signInResult as SignInResult.CodeRequired).nextState.submitCode(otp)
```
If we can't get past step one, there's no way to implement and validate step two.

Thus I personally find it hard to realize the former definition.

testutils/src/main/java/com/microsoft/identity/internal/testutils/nativeauth/api/TemporaryEmailService.kt
Yuki-YuXin: Done.

PR #2532, opened 2024-11-05
Add more details when reset password required error is received, Fixes AB#3084603

nilo-ms:
When reset password is required during a call to /token, we need some native auth specific handling to inform developers on how to handle this error.

[AB#3084603](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3084603)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
@@ -236,6 +236,19 @@ public class ExceptionAdapter {
                     developerDescription,
                     apiError
             );
+        } else if (isNativeAuthenticationResetPasswordRequiredError(errorResponse)) {
+            ServiceException apiError = new ServiceException(
+                    errorResponse.getError(),
+                    errorResponse.getErrorDescription(),
+                    null);
+
+            String developerDescription = "Password change is required, which can't be fulfilled as part of this flow."+
+                    "Please reset the password and perform a new sign in operation. Please see exception details for more information.";
+            outErr = new ServiceException(
+                    errorResponse.getError(),
+                    developerDescription,
+                    apiError
+            );
         }
         else if (shouldBeConvertedToUiRequiredException(errorResponse.getError())) {
             outErr = new UiRequiredException(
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -236,6 +236,19 @@ public static ServiceException getExceptionFromTokenErrorResponse(@NonNull final
                     developerDescription,
                     apiError
             );
+        } else if (isNativeAuthenticationResetPasswordRequiredError(errorResponse)) {
+            ServiceException apiError = new ServiceException(
+                    errorResponse.getError(),
+                    errorResponse.getErrorDescription(),
+                    null);
+
+            String developerDescription = "Password change is required, which can't be fulfilled as part of this flow."+
```
SammyO: nit: maybe mention the word "user" here, to make it clear this is about a user's password and an action the user has to complete.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResponse.kt
```diff
@@ -81,6 +82,16 @@ class SignInTokenApiResponse(
                             correlationId = correlationId
                         )
                     }
+                    errorCodes[0].isPasswordChangeRequired() -> {
+                        val customDescription = "Password change is required, which can't be fulfilled as part of this flow. " +
```
SammyO: Is this necessary? You're already adding this information in `ExceptionAdapter`. Also, I'd recommend keeping the original API error description and not altering them. Developer should have a way to access to the raw API response.

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
nilo-ms: Thanks. Should this message be "User password change is required, which can't be..."?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResponse.kt
nilo-ms: I think the "ExceptionAdapter" class handle the error when we're refreshing the access token, while in this case we handle the scenario of obtaining a new token. Correct me if I'm wrong.
About the error description, I'm adding the description from API at the end of the custom message. Do you suggest adding the custom string in a different place?

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
SammyO: Sounds good to me.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResponse.kt
SammyO: > Correct me if I'm wrong.

You're probably right, but I'm not 100% sure - you'll have to run tests.

GBakolas (approved):

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/signin/SignInTokenApiResponse.kt
nilo-ms: Ok, I did a test, and this code is executed during the signIn. So, my assumption was correct

SammyO (approved):

spetrescu84 (approved):

PR #2540, opened 2024-11-14
Translate MFA token error to UIRequiredException instead of ServiceException to release branch, Fixes AB#3090086

nilo-ms:
When MFA is required, the token error response was previously translated to a ServiceException instead of a UIRequiredException. With this update, the token error will be translated to a UIRequiredException when MFA is required.
Customization of native authentication error descriptions will only occur when the native authentication interface is being used.

[AB#3090086](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3090086)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,8 @@
+Version 18.2.1
+----------
+(common4j 15.2.1)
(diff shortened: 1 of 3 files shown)


mohitc1 (approved):
:shipit:

PR #2544, opened 2024-11-18
Debug errors in Keystore layer , Fixes AB#3092835

somalaya:
Issue : We got a couple of IcMs (https://portal.microsofticm.com/imp/v5/incidents/details/561690347/summary) in which we found that the keystore operation "unwrap" is throwing InvalidKey Exception.
This is only happening on Pixel 5 devices so far and we have not noticed any recent OS update or security patch released. One of the IcMs says that this has been happening for 6 months now. Although the impacted devices are very low, Auth app team has been receiving at-least one similar IcM every other week. 

Fix : 
- Synchronizing the unwrap method. I am not sure if this would work as I have not been able to reproduce the same error on the test devices i have. Hence based on some online resources where other type of keystore errors were observed due to multiple threads accessing the keystore at the same time, I have added synchronized keyword on "unwrap" and 2 other methods.
- Also added a telemetry metric to get the exceptions (if it is still happening even after my fix)

Fixes [AB#3092835](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3092835)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,11 @@
 vNext
 ----------
 
+Version 18.2.2
+----------
+(common4j 15.2.2)
+- [PATCH] Debug errors in Keystore layer (#2544)
+
 Version 18.2.0
 ----------
 (common4j 15.2.0)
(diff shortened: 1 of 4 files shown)


mohitc1:
>     public static byte[] wrap(@NonNull final SecretKey key,

This too?

---
Refers to: common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java:355 in d55d896. [](commit_id = d55d89656e499d8e8019a266983bc52fdec906a9, deletion_comment = False)

mohitc1 (approved):
Left a suggestion

shahzaibj (approved):

PR #2548, opened 2024-11-22
For MSAL CPP flows, match exact claims when deleting AT with intersecting scopes, Fixes AB#3095218

sansrivastav:
Related to changes in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2401
[AB#3095218](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3095218)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
@@ -128,6 +128,15 @@ public class MsalCppOAuth2TokenCache
      * @throws ClientException  If the supplied Account or Credential are null or schema invalid.
      */
     public synchronized void saveCredentials(@NonNull final Credential... credentials) throws ClientException {
+        saveCredentials(false, credentials);
+    }
+
+    /**
+     * @param credentials          list of Credential which can include AccessTokenRecord, IdTokenRecord and RefreshTokenRecord.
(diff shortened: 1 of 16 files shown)


common4j/src/main/com/microsoft/identity/common/java/cache/MsalCppOAuth2TokenCache.java
```diff
@@ -148,7 +148,7 @@ public synchronized void saveCredentials(@NonNull final Credential... credential
             }
         }
 
-        saveCredentialsInternal(credentials);
+        saveCredentialsInternal(true, credentials);
```
nickbopp: >saveCredentialsInternal(true [](http://example.com/codeflow?start=8&length=28)

could we bubble this true/false decision up to this method's parameters? This seems risky enough that we'll want to guard it with a flight (even if it starts default-enabled), and I think we'll need that so our layer that understands the flights can make that choice.

nickbopp: >saveCredentialsInternal(true, credentials); [](http://example.com/codeflow?start=8&length=43)

Can we add a test case to common that describes the difference in behavior we expected from true vs. false here?

nickbopp (approved):
:shipit:

mohitc1 (approved):
:shipit:

PR #2559, opened 2024-12-20
Add support for OneBox Environment, Fixes AB#3113751

shahzaibj:
Fixes [AB#3113751](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3113751)

Changes to allow MSAL/Broker to talk to eSTS running locally. This basically does the following:

- Add build config flag to bypass authority validation. This is required because local eSTS authorities are not discovered via cloud discovery. We can enable this flag locally in gradle project properties to target local eSTS. 
- Read the port number of the authority when constructing the cloud URL because targeting local eSTS requires supplying exact port. I've scoped the change to read port only when environment is specified as onebox. 

Related MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2231

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
@@ -250,10 +250,17 @@ public abstract class Authority {
         return null != getEquivalentConfiguredAuthority(authorityStr);
     }
 
-    private static Authority createAadAuthority(@NonNull final CommonURIBuilder authorityCommonUriBuilder,
+    private static Authority createAadAuthority(@NonNull final CommonURIBuilder uriBuilder,
                                                 @NonNull final List<String> pathSegments) {
+        final String cloudUrl;
+        if (uriBuilder.getPort() != -1) {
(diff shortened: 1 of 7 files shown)


melissaahn (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -344,6 +351,10 @@ public static boolean isKnownAuthority(Authority authority) {
             return false;
         }
 
+        if (BuildConfig.ALLOW_ONEBOX_AUTHORITIES) {
+            return true; // onebox authorities are always considered to be known.
```
mohitc1, line 355: >return true [](http://example.com/codeflow?start=12&length=11)

Shouldn't you still check if input authority is OneBox? 

Even if the build flag is enabled, you can use other public authorities and MSAL configs.

shahzaibj: Sure. 

Addressed here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2559/commits/58d3c2fb0877dc1854096892d267d30d477f7969

mohitc1 (approved):

PR #2560, opened 2024-12-20
Make userHandle response field optional

melissaahn:
### Summary
The Windows app team found a bug while testing cross-device passkey auth: https://microsoft.visualstudio.com/OS/_workitems/edit/55529057

In the logs, I see that the first attempt fails due to the userHandle attribute being missing in the response from CredMan. We currently have this attribute set as required, since the server side had mentioned that this attribute was required for them, but looking at the official WebAuthN spec, userHandle is only required if allowCredentials is NOT provided. If it is (like in the first attempt) then userHandle is optional: https://w3c.github.io/webauthn/#iface-authenticatorassertionresponse

I've confirmed with Authenticator that they always do send userHandle, and our current thinking is that some middle layer (since this is cross-device auth) might be removing that value from the final response. I'm not sure why this might be the case, but this is perfectly fine according to the WebAuthN spec. Therefore, we will follow the spec as well and will not block the response on userHandle.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
@@ -93,15 +93,15 @@ class WebAuthnJsonUtil {
                 WEBAUTHN_RESPONSE_CLIENT_DATA_JSON_KEY))
             assertionResult.put(WEBAUTHN_RESPONSE_SIGNATURE_JSON_KEY, authResponseJsonObject.get(
                 WEBAUTHN_RESPONSE_SIGNATURE_JSON_KEY))
-            try {
+            // UserHandle is optional if allowCredentials was provided in the request (username flow).
+            if (authResponseJsonObject.isNull(WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY)) {
+                Logger.info(methodTag, "UserHandle not found in assertion response.")
+            } else {
                 assertionResult.put(
                     WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY, authResponseJsonObject.get(
                         WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY
                     )
                 )
-            } catch (e: JSONException) {
-                // UserHandle is optional if allowCredentials was provided in the request (username flow).
-                Logger.info(methodTag, "UserHandle not found in assertion response.")
             }
             return assertionResult.toString()
         }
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -87,8 +93,16 @@ class WebAuthnJsonUtil {
...
                 WEBAUTHN_RESPONSE_SIGNATURE_JSON_KEY))
-            assertionResult.put(WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY, authResponseJsonObject.get(
-                WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY))
+            try {
+                assertionResult.put(
+                    WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY, authResponseJsonObject.get(
+                        WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY
+                    )
+                )
+            } catch (e: JSONException) {
```
mohitc1: >catch (e: JSONException) { [](http://example.com/codeflow?start=14&length=26)

Is there a reason why you're not checking user handle's presence and going through exception catching?

consider adding exception message or isNull(authResponseJsonObject.get(WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY))

melissaahn: Updated: aa0d05a3019d16a222957d9f45a0105002745b01

mohitc1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
```diff
@@ -87,8 +93,16 @@ class WebAuthnJsonUtil {
                 WEBAUTHN_RESPONSE_CLIENT_DATA_JSON_KEY))
             assertionResult.put(WEBAUTHN_RESPONSE_SIGNATURE_JSON_KEY, authResponseJsonObject.get(
                 WEBAUTHN_RESPONSE_SIGNATURE_JSON_KEY))
-            assertionResult.put(WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY, authResponseJsonObject.get(
-                WEBAUTHN_RESPONSE_USER_HANDLE_JSON_KEY))
+            // UserHandle is optional if allowCredentials was provided in the request (username flow).
```
iamgusain, line 96: should we also log if allowCeredentials was provided.

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/internal/fido/WebAuthnJsonUtil.kt
melissaahn: Makes sense: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2560/commits/a58e216fca74dcb351c4297516ff9b4aae5d5d2c

melissaahn:
I'm going to disable the assemble consumers part for the purpose of merging. Everything passed except for the MSAL stage, which seems to be failing due to a Native Auth unit test.

PR #2561, opened 2024-12-30
Managed profile Android util method

melissaahn:
### Summary
This PR creates an `AndroidPlatformUtil` method named `isInWorkProfile`, which returns a boolean based on if the host app is in a work profile.
For Android level 30 (Android 11/R) and above, we are able to use the simple `isManagedProfile` method of `UserManager`. For Android level 21 (Android 5/Lollipop) and above, we get the `DevicePolicyManager` instance and get a list of the active admins. If any of these active admins are found to be the profile owner app, then we know that the calling app is in a work profile. (We do something similar in `InstallCertActivity` of the WPJ feature). If the device is below Android level 21 (should be very rare), we always return false.
I tested this method out using testDPC and made sure both paths of the logic return the correct values.


common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -261,6 +263,31 @@ public static ArrayList<Map.Entry<String, String>> updateWithOrDeleteWebAuthnPar
...
 
+    /**
+     * Check if the host app is running within work profile.
+     * @param appContext current application context.
+     * @return true if app is in work profile, false if in personal profile or OS is below LOLLIPOP.
+     */
+    public static boolean isInWorkProfile(@NonNull final Context appContext) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            final UserManager um = (UserManager) appContext.getSystemService(Context.USER_SERVICE);
+            return um.isManagedProfile();
```
mohitc1, line 276: > return um.isManagedProfile(); [](http://example.com/codeflow?start=11&length=30)

if the device is fully managed, would this return false?

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -261,6 +263,31 @@ public static ArrayList<Map.Entry<String, String>> updateWithOrDeleteWebAuthnPar
         return result;
     }
 
+    /**
+     * Check if the host app is running within work profile.
+     * @param appContext current application context.
+     * @return true if app is in work profile, false if in personal profile or OS is below LOLLIPOP.
+     */
+    public static boolean isInWorkProfile(@NonNull final Context appContext) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
```
fadidurah, line 274: Can we add comments explaining why we have different behavior based on android version

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
melissaahn: Yes: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2561/commits/5d9292b6ff85353ec672098d89e81da16ff589bf

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
melissaahn: Talked to Mohit about this offline; the thinking is that for COBO devices, since they would constitute as a managed device (and not a managed profile), this method would return false. One Intune engineer also believes this is the case, but I will also check in with a different Intune team.
This question also prompted a renaming from isInWorkProfile to isInManagedProfile, to avoid any confusion as to what constitutes as a work profile.

melissaahn: (Link Mohit found: https://developers.google.com/android/work/play/emm-api/management-types#devices_and_management_types)

PR #2567, opened 2025-01-10
Add Telemetry for OOM Errors, Fixes AB#3124149

fadidurah:
Update our Out Of memory handling to include additional telemetry. This PR does not include any induced broker crashes. 

[AB#3124149](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3124149)


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,11 +440,27 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
+            // Attach the stack trace, to debug in telemetry later
+            currentSpan.setAttribute(AttributeName.out_of_memory_exception_stacktrace.name(),
+                    StringUtil.getStacktraceAsStringFromElementArray(e.getStackTrace()));
+
+            Logger.error(TAG, "Received an out of memory error, shutting broker process so a new one can be launched.", e);
+
+            // Shut down current process
+            // Calling client app will receive "killed unexpectedly" error from MSAL if broker is shut down with this statement
+            // Calling application should not crash, and should be able to make another call, which will result in a new broker process.
+            System.exit(0);
```
mohitc1: What error/response does calling app receive in when this happens?

mohitc1: There's also android.os.Process.killProcess(android.os.Process.myPid());  this might be better to use in case of OOM. I read a bit, you may also read once and we can finalize

fadidurah: Off of a quick glance, it seems using system.exit is preferable because it shuts down the JVM itself, while killprocess only kills the android process. I think from a memory management perspective, wouldn't shutting down the JVM be preferable?

fadidurah: To answer your question, we get an MSALException, Activity is killed unexpectedly

image

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,11 +440,27 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
-            );
+            // When receiving an out of memory error, instead of gracefully returning a failure result, we should shut down
+            // current broker process, to allow a new broker process to be launched at the next request from client app.
+            // This will result in a new broker process with fresh memory allocation, rather than repeating out of memory
+            // errors. In the case that this is an msal-only scenario, and this code is running inside client app, then
+            // client app will be shut down and will need to be launched again.
+
+            // Log status code and record exception in span
+            final Span currentSpan = SpanExtension.current();
+            currentSpan.setStatus(StatusCode.ERROR);
```
somalaya: Do these attributes actually get set? Asking because we are killing the process without calling span.end

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
somalaya: Would this impact OOM errors thrown from broker process during WPJ?

fadidurah: Yes, any operation going through broker that runs into OOM would end up shutting down broker process

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: Added currentSpan.end()

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
mohitc1: I read otherwise on the internet that, prefer killprocess to completely release all resources jvm and no-jvm, and all threads

mohitc1: Do you same error when the OOM happened in acquire token silent case?

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
somalaya: Ok, I would say that we should check in sandbox db if it is flowing correctly by explicitly throwing an exception.

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
somalaya: If you are sure that this would have minimal impact on non teams devices scenarios, I think we are good if you kill the process. But if this is also happening in large numbers, it would lead to crashes if OOM exception is thrown in broker during some broker specific tasks like WPJ or flights fetch or active broker selection process. And broker apps track crash analytics more seriously. 
You may also want to think if you need a flight.

fadidurah: The issue is not exclusive to teams devices. The particular icm we got that pointed it out was from teams devices, but it's a fairly standard exception to see on android from what i can tell. I'll start a teams chat to discuss with you and Mohit

fadidurah: One thing to note is that this does not cause the calling app to crash. MSAL will send an client exception gracefully to calling app, "Activity killed unexpectedly". Calling app is able to handle this exception, and either retry the request or proceed in some other way.

fadidurah: Same thing happens for Silent. Issue with using killprocess is that our exceptionAdapter is in common4j. I could look at creating a wrapper in android for exceptionAdapter specifically for this change but it seems like overkill, not sure the advantage of using killprocess justifies that kind of change.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -328,4 +333,34 @@ private boolean isValidHubRedirectURIForNAATests(String redirectUri) {
...
+        currentSpan.end();
+
+        Logger.error(tag, "Received an out of memory error, shutting broker process so a new one can be launched.", exception);
+
+        // Shut down current process
+        // Calling client app will receive an MsalClientException with "Activity killed unexpectedly" from MSAL if broker is shut down with this statement
+        // Calling application should not crash, and should be able to make another call, which will result in a new broker process.
+
+        // killProcess preferable to system.exit, as it cleans up resources more thoroughly
+        android.os.Process.killProcess(android.os.Process.myPid());
```
mohitc1: >  android.os.Process.killProcess(android.os.Process.myPid()); [](http://example.com/codeflow?start=6&length=61)

only this is platform specific logic.

common4j/src/main/com/microsoft/identity/common/java/util/IPlatformUtil.java
```diff
@@ -129,4 +130,12 @@ public interface IPlatformUtil {
      */
     @Nullable
     List<Map.Entry<String, String>> updateWithAndGetPlatformSpecificExtraQueryParameters(@Nullable List<Map.Entry<String, String>> originalList);
+
+    /**
+     * Handles the shutdown of the current process if we receive an out of memory error
+     * @param exception out of memory exception
+     * @param tag tag to be included in log statement
+     */
+    void handleShutdownForOutOfMemoryError(@NonNull final BaseException exception, @NonNull final String tag);
```
mohitc1: >handleShutdownForOutOfMemoryError [](http://example.com/codeflow?start=9&length=33)

We should only have method likd killProcess(), because that's platform dependent

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -550,6 +550,8 @@ private static CommandResult executeCommand(@SuppressWarnings(WarningType.rawtyp
             if (baseException instanceof UserCancelException) {
                 commandResult = CommandResult.ofNull(CommandResult.ResultStatus.CANCEL,
                         correlationId);
+            } else if (baseException.getErrorCode().equals(ClientException.OUT_OF_MEMORY)) { // If we receive an out of memory error
+                command.getParameters().getPlatformComponents().getPlatformUtil().handleShutdownForOutOfMemoryError(baseException, TAG);
```
mohitc1: >command.getParameters().getPlatformComponents().getPlatformUtil().handleShutdownForOutOfMemoryError(baseException, TAG); [](http://example.com/codeflow?start=16&length=120)

I think there could be OOM errors even before command dispatcher.

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: I've reworked to instead handle this in command dispatcher using IPlatformUtil

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
fadidurah: What is the earliest common point to do this? I'm not sure if we can do msalbroker result adapter, IDK if that code is ran as part of broker process or msal process.

common4j/src/main/com/microsoft/identity/common/java/util/IPlatformUtil.java
fadidurah: Sorry, not sure what you mean here. are you saying the method should be renamed?

fadidurah: Nevermind, I see what you mean

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
fadidurah: Moved the other code

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: We can discuss this point in scrum

fadidurah: Decided to validate this with broker hosting app teams, and will instead merge this fix in feb

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: Resolving as i'm no longer crashing broker

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: Resolving as i'm no longer crashing broker

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
fadidurah: Resolving as i'm no longer crashing broker

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,6 +439,18 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
 
 
         if (e instanceof OutOfMemoryError) {
+            // Log status code and record exception in span
+            final Span currentSpan = SpanExtension.current();
```
somalaya: What span names are capturing this telemetry?

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,6 +439,18 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
+
+            // Attach the stack trace, to debug in telemetry later
+            currentSpan.setAttribute(
+                    AttributeName.out_of_memory_exception_stacktrace.name(),
+                    StringUtil.getStacktraceAsStringFromElementArray(e.getStackTrace())
+            );
+
+            com.microsoft.identity.common.java.logging.Logger.warn(
+                    TAG, "Received an out of memory error, stacktrace attached to span.");
```
somalaya: Do you also want to add spanId in this log so that you can look for the exact device and what caused that particular device to throw OOM error?

common4j/src/main/com/microsoft/identity/common/java/util/StringUtil.java
```diff
@@ -463,4 +463,24 @@ public static void overwriteWithNull(final char[] chars) {
...
+    /**
+     * Converts an array of stack trace elements (one method invocation) into one concatenated string object.
+     *
+     * @param elements list of stacktrace elements
+     * @return concatenated string list
+     */
+    @NonNull
+    public static String getStacktraceAsStringFromElementArray(@NonNull final StackTraceElement[] elements) {
+        final StringBuilder builder = new StringBuilder();
+
```
somalaya: extra line

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: It would be whatever span name is currently executing the broker command

fadidurah: This wouldn't be limited to interactive or silent, but whatever span runs into out of memory exception

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
fadidurah: Yeah we can add that

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,6 +439,18 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
 
         if (e instanceof OutOfMemoryError) {
+            // Log status code and record exception in span
+            final Span currentSpan = SpanExtension.current();
+
+            // Attach the stack trace, to debug in telemetry later
+            currentSpan.setAttribute(
+                    AttributeName.out_of_memory_exception_stacktrace.name(),
+                    StringUtil.getStacktraceAsStringFromElementArray(e.getStackTrace())
+            );
```
mohitc1: given we reach here due to OOM, this (stack trace method does significant allocation) also would likely cause OOM, so I am not sure if we gain anything here.. Still if we do this, this statement should also be under try 
try () {
   span...
} catch (OutOfMemory) {
    /no-op or log
} 

I wonder, If we never had this if block, OOM exception would have bubbled to the top and caused broker process to crash (and perhaps msal side as well) but eventually fixing the issue.

mohitc1: This was added in last year July by Melissa, we can check with her if OOM was there for reason or in general to handle UNKNWON_ERRORs. As a fix we can consider moving back earlier state for OOM in march.

fadidurah: I think melissa's change was to make unknown errors less common and give us more detailed exceptions, reverting would lead to uptik in unknown errors but functionally the same behavior. I do think a try catch is needed here though, good catch on that

fadidurah: I've added the try-catch now

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,6 +439,27 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
+
+                // Attach the stack trace, to debug in telemetry later
+                currentSpan.setAttribute(
+                        AttributeName.out_of_memory_exception_stacktrace.name(),
+                        StringUtil.getStacktraceAsStringFromElementArray(e.getStackTrace())
+                );
+
+                Logger.warn(TAG,
+                        "Received an out of memory error, stacktrace attached to span with id: " + currentSpan.getSpanContext().getSpanId());
+            } catch (Exception secondException) {
```
mohitc1: >Exception [](http://example.com/codeflow?start=21&length=9)

Throwable

OutOfMemory is not Exception

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -436,6 +439,27 @@ public static ClientException clientExceptionFromException(@NonNull final Throwa
...
+                currentSpan.setAttribute(
+                        AttributeName.out_of_memory_exception_stacktrace.name(),
+                        StringUtil.getStacktraceAsStringFromElementArray(e.getStackTrace())
+                );
+
+                Logger.warn(TAG,
+                        "Received an out of memory error, stacktrace attached to span with id: " + currentSpan.getSpanContext().getSpanId());
+            } catch (Throwable throwable) {
+                Logger.error(TAG,
+                        "Failed to emit telemetry for out of memory exception.", throwable);
```
mohitc1: maybe Logger.warn or Logger.error without throwable? To not make Logger heavy on memory

mohitc1 (approved):
Left a suggestion on logger

somalaya (approved):

changelog.txt
```diff
@@ -34,6 +34,7 @@ Version 19.0.0
 - [MINOR] Managed profile Android util method (#2561)
 - [PATCH] Make userHandle response field optional (#2560)
 - [MINOR] Nonce redirect changes (#2552)
+- [PATCH] Fix OOM error handling (#2567)
```
mohitc1: Update change log as well.

PR #2573, opened 2025-01-21
[SIWG] Pass Sign In With Google credential to broker, Fixes AB#3100603

mohitc1:
Fixes [AB#3100603](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3100603)
The PR adds to allow passing google id token credential from app side (in OneAuth) to Android Broker via Android Common.

OneAuth can get the id token by using Sign In With Google component added earlier and pass it to Android Broker via BrokerRequest object.

Changes include:
1. Make SignInWithGoogleCredential object serializable to allow passing the object from app side to broker as part of BrokerRequest.
2. Locally the reference to object can be maintained in Android AndroidInteractiveTokenCommandParameters which is renamed from AndroidActivityInteractiveTokenCommandParameters to be more flexible to accommodate more fields.
3. On Broker received, BrokerRequest is deserialized. If it has google credentials, it will be attached in BrokerInteractiveTokenCommandParameters as headers (x-ms-fidp-token: <itoken calue>) and extra query parameters (id_provder=google.com). Refer Broker PR.
4. Updated latest broker version to 17. OneAuth should call broker for sign in with google flow only if the broker version supports this version. OneAuth does broker version check early in deciding if support broker is present or not.
5. Added Unit tests.

Related PRs:
Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/3038
MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2247


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -605,7 +605,7 @@ public static final class Broker {
          *
          * @see <a href="https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Broker%20API/broker_protocol_versions.md">Android Auth Broker Protocol Versions</a>
```
somalaya, line 606: Once you merge the PR, could you also update the version in the link above?

common/src/main/java/com/microsoft/identity/common/internal/msafederation/FederatedCredential.kt
```diff
@@ -22,8 +22,17 @@
...
  * (Google/Apple). It can contain id token and/or auth code. See implementations for more details.
  */
-abstract class FederatedCredential(val federatedSignInProviderName: FederatedSignInProviderName)
+abstract class FederatedCredential(@SerializedName("signInProviderName") val signInProviderName: FederatedSignInProviderName) {
+    fun getIdProviderExtraQueryParam(): Map.Entry<String, String> {
+        return AbstractMap.SimpleEntry(MsaFederationConstants.MSA_ID_PROVIDER_EXTRA_QUERY_PARAM_KEY, signInProviderName.getIdProviderName())
+    }
+}
+
+
```
somalaya: Can remove the extra new lines #Resolved

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleCredential.kt
```diff
@@ -22,21 +22,24 @@
...
-data class SignInWithGoogleCredential internal constructor(internal val idToken: String) : FederatedCredential(FederatedSignInProviderName.GOOGLE) {
+data class SignInWithGoogleCredential internal constructor(
+    @SerializedName("idToken") internal val idToken: String) : FederatedCredential(FederatedSignInProviderName.GOOGLE), Serializable {
 
     /**
      * Helper method to create header that can be used in MSA
      * authorization. (Move to better place.)
      */
-    fun asHeaders(): Map<String, String> {
+    fun getIdProviderHeaders(): Map<String, String> {
```
shahzaibj: I suggest sticking to either abbreviation or full form word. `IdProvider` is somwhere in between.

I suggest choosing either `IdP` (a true abbreviation) or rather just using full word `IdentityProvider`

shahzaibj: Also, tbh, I can't really tell what this function is supposed to do by just looking at the method name. I mean it's clear that it is going to return a map that contains some headers but not clear on who/where/when these headers should be used. 

For example, someone may just get these and start attaching these to DRS Discovery requests...which is not what they are for.

common/src/main/java/com/microsoft/identity/common/internal/msafederation/FederatedCredential.kt
```diff
@@ -22,8 +22,17 @@
...
 
+import com.google.gson.annotations.SerializedName
+import java.util.AbstractMap
+
 /**
  * Represents credential artifact as result of successful sign in into a federated sign in provider
  * (Google/Apple). It can contain id token and/or auth code. See implementations for more details.
  */
-abstract class FederatedCredential(val federatedSignInProviderName: FederatedSignInProviderName)
+abstract class FederatedCredential(@SerializedName("signInProviderName") val signInProviderName: FederatedSignInProviderName) {
```
shahzaibj:
```suggestion
abstract class MsaFederatedCredential(@SerializedName("signInProviderName") val signInProviderName: FederatedSignInProviderName) {
```

shahzaibj: Suggest rename to avoid any confusion about AAD which is where the word `federation` mostly gets used.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
mohitc1: yes.

common/src/main/java/com/microsoft/identity/common/internal/msafederation/FederatedCredential.kt
mohitc1: renamed all classes/interfaces

common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleCredential.kt
mohitc1: Refactored into extension methods.

named IdProvider keeping on authorization endpoint parameter "id_provider" as reference.

common/src/main/java/com/microsoft/identity/common/internal/msafederation/MsaFederationExtensions.kt
```diff
@@ -0,0 +1,41 @@
...
+fun SignInWithGoogleCredential.getIdProviderHeadersForAuthorization(): Map<String, String> {
+    return mapOf(MsaFederationConstants.MSA_ID_TOKEN_HEADER_KEY to this.idToken)
+}
+
+/**
+ * Helper/Extension method to create query parameter that can be used in MSA authorization
+ */
+fun SignInWithGoogleCredential.getIdProviderExtraQueryParamForAuthorization(): Map.Entry<String, String> {
+    return AbstractMap.SimpleEntry(MsaFederationConstants.MSA_ID_PROVIDER_EXTRA_QUERY_PARAM_KEY, signInProviderName.getIdProviderName())
+}
```
shahzaibj: nit: new line ending

shahzaibj (approved):

PR #2564, opened 2025-01-03
Organize browser selection classes and change signature for get AuthorizationStrategy, Fixes AB#3119409

p3dr0rv:
The changes here are the preparatory work for the DUNA feature. Splitting this PR from DUNA to simplify the review process.

- Update Browser Selector and move it to Platform Components
- Change getAuthorizationStrategy signature from (InteractiveTokenCommandParameters) to (AuthorizationAgent , Browser, isBrokerRequest)
- Move getBrowserSafeListForBroker from Platform Util to BrowserDescriptor
- Move Browser to common4j

[AB#3119409](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3119409)

PR #2580, opened 2025-01-29
On getBrowserSafeList return empty list if null

p3dr0rv:
with the changes introduced here #2564 getDefaultBrowser does not accept null


mohitc1 (approved):
Consider adding UT if possible.

fadidurah (approved):

PR #2581, opened 2025-01-30
Add server client id to SignInWithGoogleParameters

mohitc1:
This PR has changes to allow caller to pass server client id as part of SignInWithGoogleParameters.

Adding this change in case we end up using App's own server client id instead of MSA's. In either case MSA service would validate it. The default stays MSA's server client id.

This client id identifies the backend to which google id token is sent. "sub" claim in the id token is this client id.

Added tests.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleParameters.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleParameters.kt
@@ -25,29 +25,39 @@ package com.microsoft.identity.common.internal.msafederation.google
 import android.app.Activity
 import com.microsoft.identity.common.internal.msafederation.MsaFederatedSignInParameters
 import com.microsoft.identity.common.internal.msafederation.MsaFederatedSignInProviderName
+import com.microsoft.identity.common.internal.msafederation.MsaFederationConstants
 
 /**
  * SignInWithGoogleParameters holds the parameters required for signing in with Google.
  *
  * @property activity The Activity context used for the sign-in process.
+ * @property serverClientId The target server OAuth2.0 client (backend) ID generated in google console project.
+ *                          Authenticates backend server with Google's APIs. Default value MSA's server client id.
  * UI for the sign-in process.
  */
-data class SignInWithGoogleParameters(
-    internal val activity: Activity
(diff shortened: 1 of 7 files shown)


shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleParameters.kt
```diff
@@ -25,29 +25,39 @@ package com.microsoft.identity.common.internal.msafederation.google
...
  * @property activity The Activity context used for the sign-in process.
+ * @property serverClientId The target server OAuth2.0 client (backend) ID generated in google console project.
+ *                          Authenticates backend server with Google's APIs. Default value MSA's server client id.
  * UI for the sign-in process.
  */
-data class SignInWithGoogleParameters(
-    internal val activity: Activity
+data class SignInWithGoogleParameters @JvmOverloads constructor(
+    internal val activity: Activity,
+    internal val serverClientId: String = MsaFederationConstants.GOOGLE_MSA_SERVER_CLIENT_ID
```
iamgusain, line 40: Should the serverClientId value be a flighting config, so it can be changed without code change?

iamgusain (approved):

shahzaibj:
When you say it "identifies the backend" what you mean? Isn't the backend always MSA service?

mohitc1:
> When you say it "identifies the backend" what you mean? Isn't the backend always MSA service?

Here the backend refers to an OAuth2 client id assigned to web service. And this can be registered by Copilot in its own google console project (Branding, consent UX and ToS would be of Copilot) or MSA in its console project (MSA's branding etc)

There's no consensus yet which one to use (conversation is driven by MSA team. MSA wants to use their own id). in either case MSA should accept the id token sent from client side.

From client side, we use this server id to request for id token, hence we either use MSA (default case) or let app pass it (if they own the id).
Hence this change. Possible that we may delete one of the paths later, once this consensus reached.

common/src/main/java/com/microsoft/identity/common/internal/msafederation/google/SignInWithGoogleParameters.kt
mohitc1: This would be executed on app side not broker, so we can't.

PR #2585, opened 2025-02-04
Move generateSignatureHashes to PackageHelper

p3dr0rv:
Moving this method to PackageHelper makes easier the fix for the breaking change on OneAuth lib


mohitc1 (approved):
:shipit:

fadidurah (approved):

PR #2586, opened 2025-02-06
Add fallback mechanisms for keypair generation, Fixes AB#3146656

somalaya:
Issue : Some specific device models do not like the purpose "PURPOSE_WRAP_KEY" to be passed while doing a keypair generation.
Root cause : Changes made in this PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2558/files
IcM : https://portal.microsofticm.com/imp/v5/incidents/details/592800555/summary
Fix : 
1. Added fallback mechanism to skip using wrap_key purpose when the SecureKeyImportUnavailableException occurs.
2. If the fallback mentioned in step 1 fails, we fallback to legacy keygen spec method.
3. Added telemetry to capture which method was finally successful and also the time taken to generate the keypair
4. Also kept my changes behind a flight to turn it off in case of any failures

Manually tested below steps
1. Signed in with an AAD account with the flight for using the latest KeyGenParameterSpec as ON in MsalTestApp with BrokerHost as the broker
5. Tried  ATS ---> Succeeds
6. Tried getBrokerAccounts API from BrokerHost app ---> Succeeds
7. Did a device registration -- Succeeds
-----------------------------------------------------------------------------
8. Assuming some issue happened, I turned off the flight
9. Tried  ATS ---> Succeeds
10. Tried ATI with same UPN --> Succeeds
11. Tried getBrokerAccounts API from BrokerHost app ---> Succeeds
12. Tried to get WPJ entries --> Gets them correctly

Fixes [AB#3146656](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3146656)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
@@ -37,6 +37,10 @@ import com.microsoft.identity.common.java.crypto.key.KeyUtil;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.flighting.CommonFlight;
 import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
+import com.microsoft.identity.common.java.opentelemetry.AttributeName;
+import com.microsoft.identity.common.java.opentelemetry.OTelUtility;
+import com.microsoft.identity.common.java.opentelemetry.SpanExtension;
+import com.microsoft.identity.common.java.opentelemetry.SpanName;
 import com.microsoft.identity.common.java.util.CachedData;
 import com.microsoft.identity.common.java.util.FileUtil;
 import com.microsoft.identity.common.logging.Logger;
@@ -46,6 +50,7 @@ import java.math.BigInteger;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.KeyStore;
+import java.security.ProviderException;
 import java.security.spec.AlgorithmParameterSpec;
 import java.util.Calendar;
 import java.util.Locale;
@@ -55,6 +60,9 @@ import javax.security.auth.x500.X500Principal;
 
 import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Scope;
 import lombok.NonNull;
 
 /**
@@ -249,15 +257,62 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
          * stomping/overwriting one another's keypair.
          */
         KeyPair keyPair = AndroidKeyStoreUtil.readKey(mAlias);
-        if(keyPair == null){
+        if (keyPair == null) {
             Logger.info(methodTag, "No existing keypair. Generating a new one.");
-            keyPair = AndroidKeyStoreUtil.generateKeyPair(
-                    WRAP_KEY_ALGORITHM,
-                    getSpecForKeyStoreKey(mContext, mAlias));
+            final Span span = OTelUtility.createSpan(SpanName.KeyPairGeneration.name());
+            final long keypairGenStartTime = System.currentTimeMillis();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP)) {
+                try (Scope scope = SpanExtension.makeCurrentSpan(span)) {
+                    keyPair = attemptKeyPairGeneration(mAlias, true, keypairGenStartTime);
+                    Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_with_wrap");
+                    span.setStatus(StatusCode.OK);
+                } catch (final ProviderException e) {
+                    if ("SecureKeyImportUnavailableException".equals(e.getClass().getSimpleName())) {
+                        Logger.warn(methodTag, "Wrap purpose may not be supported. Retrying without wrap.");
+                        keyPair = attemptKeyPairGeneration(mAlias, false, keypairGenStartTime);
+                        Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_without_wrap");
+                        span.setStatus(StatusCode.OK);
+                    } else {
+                        // Some unknown exception occurred. Rethrow it so that legacy keygen spec logic can run.
+                        throw e;
+                    }
+                } catch (final Exception e) {
+                    Logger.warn(methodTag, "Unexpected error with new KeyPairGeneratorSpec. Falling back to legacy spec.");
+                    keyPair = generateKeyPairWithLegacySpec(mAlias, keypairGenStartTime);
+                    Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "legacy_key_gen_spec");
+                    span.setStatus(StatusCode.OK);
+                } finally {
+                    span.end();
+                }
+            }
+            else {
+                // If flight for using new keygen spec is not enabled, use the legacy spec.
+                Logger.info(methodTag, "Using legacy spec for keypair generation directly.");
+                keyPair = generateKeyPairWithLegacySpec(mAlias, keypairGenStartTime);
+            }
         }
-
         final byte[] keyWrapped = AndroidKeyStoreUtil.wrap(unencryptedKey, keyPair, WRAP_ALGORITHM);
         FileUtil.writeDataToFile(keyWrapped, getKeyFile());
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.P)
+    private KeyPair attemptKeyPairGeneration(@NonNull final String alias, boolean useWrapPurpose, long keypairGenStartTime) throws ClientException{
+        KeyPair keyPair = AndroidKeyStoreUtil.generateKeyPair(
+                WRAP_KEY_ALGORITHM, getSpecForKeyStoreKey(mContext, alias, useWrapPurpose));
+        recordKeyGenerationTime(keypairGenStartTime);
+        return keyPair;
+    }
+
+    private KeyPair generateKeyPairWithLegacySpec(@NonNull final String alias, long keypairGenStartTime) throws ClientException{
+        KeyPair keyPair = AndroidKeyStoreUtil.generateKeyPair(
+                WRAP_KEY_ALGORITHM, getLegacySpecForKeyStoreKey(mContext, alias));
+        recordKeyGenerationTime(keypairGenStartTime);
+        return keyPair;
+    }
+
+    private void recordKeyGenerationTime(long keypairGenStartTime) {
+        long elapsedTime = System.currentTimeMillis() - keypairGenStartTime;
+        SpanExtension.current().setAttribute(AttributeName.elapsed_time_keypair_generation.name(), elapsedTime);
     }
 
     /**
@@ -305,20 +360,21 @@ public class AndroidWrappedKeyLoader extends AES256KeyLoader {
      * This is for the key to be generated in {@link KeyStore} via {@link KeyPairGenerator}
      *
      * @param context an Android {@link Context} object.
+     * @param alias   the alias for the key.
+     * @param tryPurposeWrap whether to try to use the wrap purpose in the key generation spec.
      * @return a {@link AlgorithmParameterSpec} for the keystore key (that we'll use to wrap the secret key).
      */
-    private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias) {
-        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            int purposes =  KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
-            return new KeyGenParameterSpec.Builder(alias, purposes)
-                    .setKeySize(2048)
-                    .setDigests(KeyProperties.DIGEST_SHA256, KeyProperties.DIGEST_SHA512)
-                    .setBlockModes(KeyProperties.BLOCK_MODE_ECB) // Ensure compatibility with RSA
-                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)
-                    .build();
-        } else {
-            return getLegacySpecForKeyStoreKey(context, alias);
+    @RequiresApi(api = Build.VERSION_CODES.P)
+    private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias, boolean tryPurposeWrap) {
+        int purposes = KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
+        if (tryPurposeWrap) {
+            purposes |= KeyProperties.PURPOSE_WRAP_KEY;
         }
+        return new KeyGenParameterSpec.Builder(alias, purposes)
+                .setKeySize(2048)
+                .setDigests(KeyProperties.DIGEST_SHA256, KeyProperties.DIGEST_SHA512)
+                .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)
+                .build();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -306,16 +308,16 @@ private static AlgorithmParameterSpec getLegacySpecForKeyStoreKey(@NonNull final
      * @return a {@link AlgorithmParameterSpec} for the keystore key (that we'll use to wrap the secret key).
      */
     private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias) {
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
-            return getLegacySpecForKeyStoreKey(context, alias);
-        } else {
-            int purposes = KeyProperties.PURPOSE_WRAP_KEY  | KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            int purposes =  KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
```
shahzaibj: You still seem to be passing `purposes` below at line 313. Is this expected? I'm not sure if the PR is achieving what the title says

shahzaibj:
It seems there is a contradiction between PR title and changelog / implementation

shahzaibj:
> It seems there is a contradiction between PR title and changelog / implementation

Title says the PR is going to remove the purpose

changelog says the PR is just going to move new key spec behind a flight (this is also what's implemented in the code in the PR)

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -306,16 +308,16 @@ private static AlgorithmParameterSpec getLegacySpecForKeyStoreKey(@NonNull final
      * @return a {@link AlgorithmParameterSpec} for the keystore key (that we'll use to wrap the secret key).
      */
     private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias) {
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
-            return getLegacySpecForKeyStoreKey(context, alias);
-        } else {
-            int purposes = KeyProperties.PURPOSE_WRAP_KEY  | KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
iamgusain: If the issue only happens for certain devices, how do you plan to selectively turn ON and OFF the flight for specific devices.
One way would be to add legacy logic as fallback in case the new logic fails.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
somalaya: I meant the PURPOSE_WRAP_KEY will be removed as that is the one that is causing the problem on this specific device. 
Hence the PR title. Changelog is not update at all. If you meant the commits, I have 2 commits. 
1st for removing the wrap purpose and 2nd for putting my whole code behind a flight just in case it is causing any new unknown problems.

somalaya: On top of it, I am also planning to push 1 more commit. Which will keep the PURPOSE_WRAP_KEY  purpose as well and catches the SecureImportUnavailableException for these special devices. It would then try again to generate a keypair with just the other 2 purposes.

shahzaibj: Ok so purpose itself is fine but only specific purpose is the problem?

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -88,7 +88,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable the re-attachment of new PRT header logic. Default is true.
      */
-    ENABLE_ATTACH_NEW_PRT_HEADER_WHEN_NONCE_EXPIRED("EnableAttachNewPrtHeaderWhenNonceExpired", true);
+    ENABLE_ATTACH_NEW_PRT_HEADER_WHEN_NONCE_EXPIRED("EnableAttachNewPrtHeaderWhenNonceExpired", true),
+
+    /**
+     * Flight to enable the new key generation spec for wrap key. Default is true.
+     */
+    ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP("EnableNewKeyGenSpecForWrap", true);
```
shahzaibj, line 96: I actually think we need to also have a flight for the new change here (whether to use purpose for new key spec).

So a total of 2 flights:

- 1 for enabling new key spec
- 1 for setting purpose wrap or not

This is because new key spec is already on 25% devices with purpose wrap, right? Are we okay to move forward with this change that says either turn on new spec without wrap purpose or just turn off new spec? Why not also have the option to have wrap purpose just in case the new spec without wrap purpose causes problems?

somalaya: As we discussed offline, I am going to skip using 2 flights and rely on try catch blocks and just have 1 flight.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,64 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
-                    WRAP_KEY_ALGORITHM,
-                    getSpecForKeyStoreKey(mContext, mAlias));
+            final Span span = OTelUtility.createSpan(SpanName.KeyPairGeneration.name());
+            final long keypairGenStartTime = System.currentTimeMillis();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP)) {
+                try (Scope scope = SpanExtension.makeCurrentSpan(span)) {
+                    keyPair = attemptKeyPairGeneration(mAlias, true, keypairGenStartTime);
+                    Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_with_wrap");
+                    span.setStatus(StatusCode.OK);
+                } catch (final ProviderException e) {
```
iamgusain, line 270: You can directly catch. SecureKeyImportUnavailableException
catch (final SecureKeyImportUnavailableException e) {...

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,64 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
          * stomping/overwriting one another's keypair.
          */
         KeyPair keyPair = AndroidKeyStoreUtil.readKey(mAlias);
-        if(keyPair == null){
+        if (keyPair == null) {
             Logger.info(methodTag, "No existing keypair. Generating a new one.");
-            keyPair = AndroidKeyStoreUtil.generateKeyPair(
-                    WRAP_KEY_ALGORITHM,
-                    getSpecForKeyStoreKey(mContext, mAlias));
+            final Span span = OTelUtility.createSpan(SpanName.KeyPairGeneration.name());
```
iamgusain, line 262: Should we set a specific sampling rate for this span, or the default one should be fine.

iamgusain (approved):

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
somalaya: yes purpose_wrap key is the problem

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -303,20 +360,21 @@ private static AlgorithmParameterSpec getLegacySpecForKeyStoreKey(@NonNull final
...
-        } else {
-            int purposes = KeyProperties.PURPOSE_WRAP_KEY  | KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
-            return new KeyGenParameterSpec.Builder(alias, purposes)
-                    .setKeySize(2048)
-                    .setDigests(KeyProperties.DIGEST_SHA256, KeyProperties.DIGEST_SHA512)
-                    .setBlockModes(KeyProperties.BLOCK_MODE_ECB) // Ensure compatibility with RSA
-                    .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)
-                    .build();
+    @RequiresApi(api = Build.VERSION_CODES.P)
+    private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias, boolean tryPurposeWrap) {
```
shahzaibj:
```suggestion
    private static AlgorithmParameterSpec getSpecForKeyStoreKey(@NonNull final Context context, @NonNull final String alias, final boolean tryPurposeWrap) {
```

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,64 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
+                    span.setStatus(StatusCode.OK);
+                } catch (final ProviderException e) {
+                    if ("SecureKeyImportUnavailableException".equals(e.getClass().getSimpleName())) {
+                        Logger.warn(methodTag, "Wrap purpose may not be supported. Retrying without wrap.");
+                        keyPair = attemptKeyPairGeneration(mAlias, false, keypairGenStartTime);
+                        Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_without_wrap");
+                        span.setStatus(StatusCode.OK);
+                    } else {
+                        // Some unknown exception occurred. Rethrow it so that legacy keygen spec logic can run.
+                        throw e;
```
mohitc1: >throw e; [](http://example.com/codeflow?start=24&length=8)

Doesn't look like this will be caught. you'd need add legacy logic here too

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,64 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
+            final long keypairGenStartTime = System.currentTimeMillis();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP)) {
+                try (Scope scope = SpanExtension.makeCurrentSpan(span)) {
+                    keyPair = attemptKeyPairGeneration(mAlias, true, keypairGenStartTime);
+                    Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_with_wrap");
+                    span.setStatus(StatusCode.OK);
+                } catch (final ProviderException e) {
+                    if ("SecureKeyImportUnavailableException".equals(e.getClass().getSimpleName())) {
+                        Logger.warn(methodTag, "Wrap purpose may not be supported. Retrying without wrap.");
+                        keyPair = attemptKeyPairGeneration(mAlias, false, keypairGenStartTime);
```
mohitc1: > keyPair = attemptKeyPairGeneration(mAlias, false, keypairGenStartTime); [](http://example.com/codeflow?start=23&length=72)

Exception from here too.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
somalaya: Yes, I am testing these flows and caught this.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
somalaya: Yes i did this in this PR https://github.com/AzureAD/ad-accounts-for-android/pull/3048

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,76 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
-        if(keyPair == null){
+        if (keyPair == null) {
             Logger.info(methodTag, "No existing keypair. Generating a new one.");
-            keyPair = AndroidKeyStoreUtil.generateKeyPair(
-                    WRAP_KEY_ALGORITHM,
-                    getSpecForKeyStoreKey(mContext, mAlias));
+            final Span span = OTelUtility.createSpan(SpanName.KeyPairGeneration.name());
+            final long keypairGenStartTime = System.currentTimeMillis();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP)) {
+                try (Scope scope = SpanExtension.makeCurrentSpan(span)) {
```
shahzaibj:
```suggestion
                try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
```

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,76 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
+                    span.setStatus(StatusCode.OK);
+                } catch (final ProviderException e) {
+                    if ("SecureKeyImportUnavailableException".equals(e.getClass().getSimpleName())) {
+                        Logger.warn(methodTag, "Wrap purpose may not be supported. Retrying without wrap.");
+                        try {
+                            keyPair = attemptKeyPairGeneration(mAlias, false, keypairGenStartTime);
+                            Logger.info(methodTag, "Successfully generated keypair with new KeyPairGeneratorSpec without wrap purpose.");
+                            Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "new_key_gen_spec_without_wrap");
+                            span.setStatus(StatusCode.OK);
+                        } catch (Exception ex) {
```
shahzaibj, line 278:
```suggestion
                        } catch (final Exception ex) {
```

shahzaibj, line 278: Probably worth capturing the exception class name / error message in telemetry. We don't seem to be doing that

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,76 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
+                            Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "legacy_key_gen_spec");
+                            span.setStatus(StatusCode.OK);
+                        }
+                    } else {
+                        Logger.info(methodTag, "Some unknown exception occurred. Running legacy keygen spec logic.");
+                        keyPair = generateKeyPairWithLegacySpec(mAlias, keypairGenStartTime);
+                        Span.current().setAttribute(AttributeName.key_pair_gen_successful_method.name(), "legacy_key_gen_spec");
+                        span.setStatus(StatusCode.OK);
+                    }
+                } catch (final Exception e) {
```
shahzaibj, line 294: Probably worth capturing the exception class name / error message in telemetry. We don't seem to be doing that

shahzaibj (approved):
Approving with suggestions:

I think we may want to know why certain mechanism failed in telemetry.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -247,17 +257,76 @@ private void saveSecretKeyToStorage(@NonNull final SecretKey unencryptedKey) thr
...
+        recordKeyGenerationTime(keypairGenStartTime);
+        return keyPair;
+    }
+
+    private KeyPair generateKeyPairWithLegacySpec(@NonNull final String alias, long keypairGenStartTime) throws ClientException{
+        KeyPair keyPair = AndroidKeyStoreUtil.generateKeyPair(
+                WRAP_KEY_ALGORITHM, getLegacySpecForKeyStoreKey(mContext, alias));
+        recordKeyGenerationTime(keypairGenStartTime);
+        return keyPair;
+    }
```
shahzaibj, line 329: The way you are capturing the perf time here (and other places), it means this time is going to represent the total time for key generation (as opposed to just one mechanism). I think this is fine but just want us to be aware of it. On instance where we see poor perf here it could be because the first 2 mechanisms failed and it fell back to 3rd one and so that's a representation of a total time on 3 attempts (vs perhaps just 1 attempt on other devices)

PR #2595, opened 2025-02-26
Fix issue 1991: The acquireToken method doesn't work if the app uses "withFragmemt"

iamgusain:
Creating this PR to merge changes in the Original PR: #2274


VolodaUa:
Hi @iamgusain 
Could you please merge this pull request? I see it was created, but it has not been active for a long time. 

Thanks a lot!

PR #2599, opened 2025-03-11
Fix multiple prompts issue in cross cloud request, Fixes AB#3179147

somalaya:
**Issue** : https://portal.microsofticm.com/imp/v5/incidents/details/603998142/summary
When cross cloud authentication happens, webview could be redirected to a different cloud rather than the initial cloud loaded in the webview. When such a redirect happens, the request have no more PRT header attached that can be consumed by the new cloud. Therefore users will be asked to input credential when redirected to a new cloud.

**Fix** : In this PR, we are adding the support to attach cross domain PRT in embedded webview requests.
To improve the SSO experince, we are now making the change to detect redirects between clouds and attach available PRT for the new cloud if cross cloud authentication happens.

**Testing** : 
For testing the fix
1. Ran all UI tests to ensure existing tests are not breaking https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1442539&view=ms.vss-test-web.build-test-results-tab&runId=4591284&resultId=100022&paneView=debug
2. Ran below steps to repro the issue and fix it

- Install MsalTestApp and BrokrHost
- In MsalTestApp, use default config, loginhint [idlabxcg@msidlab4.onmicrosoft.com](mailto:idlabxcg@msidlab4.onmicrosoft.com) and deviceId claim attached. Complete the sign in.
- Next, change the config to Arlington, loginhint [idlabxcg@msidlab4.onmicrosoft.com](mailto:idlabxcg@msidlab4.onmicrosoft.com), deviceId claim attached, authority https://login.microsoftonline.us/45ff0c17-f8b5-489b-b7fd-2fedebbec0c4 and click on acquire token
- Without my fix, we would be prompted to enter password twice at this point.
- With my fix, we are no longer prompted to enter password at all.

Fixes [AB#3179147](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3179147)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -216,7 +216,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
                         ViewTreeLifecycleOwner.get(view));
                 challengeHandler.processChallenge(challenge);
             } else if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_ATTACH_NEW_PRT_HEADER_WHEN_NONCE_EXPIRED) && isNonceRedirect(formattedURL)) {
-                Logger.info(methodTag,"Navigation contains new nonce within the redirect uri. "+ url);
+                Logger.info(methodTag,"Navigation contains new nonce within the redirect uri.");
                 processNonceAndReAttachHeaders(view, url);
              }
              else if (isRedirectUrl(formattedURL)) {
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -51,10 +51,12 @@ import com.microsoft.identity.common.internal.providers.oauth2.WebViewAuthorizat
 import com.microsoft.identity.common.internal.ui.webview.certbasedauth.AbstractSmartcardCertBasedAuthChallengeHandler;
 import com.microsoft.identity.common.internal.ui.webview.certbasedauth.AbstractCertBasedAuthChallengeHandler;
 import com.microsoft.identity.common.internal.ui.webview.certbasedauth.CertBasedAuthFactory;
+import com.microsoft.identity.common.internal.ui.webview.challengehandlers.CrossCloudChallengeHandler;
 import com.microsoft.identity.common.internal.ui.webview.challengehandlers.SwitchBrowserChallenge;
 import com.microsoft.identity.common.internal.ui.webview.challengehandlers.SwitchBrowserHandler;
 import com.microsoft.identity.common.internal.ui.webview.challengehandlers.NonceRedirectHandler;
 import com.microsoft.identity.common.internal.ui.webview.challengehandlers.SwitchBrowserUriHelper;
+import com.microsoft.identity.common.java.authorities.Authority;
 import com.microsoft.identity.common.java.constants.FidoConstants;
 import com.microsoft.identity.common.java.exception.IErrorInformation;
 import com.microsoft.identity.common.java.flighting.CommonFlight;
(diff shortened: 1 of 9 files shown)


common4j/src/main/com/microsoft/identity/common/java/broker/CommonRefreshTokenCredentialProvider.kt
```diff
@@ -49,4 +49,13 @@ object CommonRefreshTokenCredentialProvider : IRefreshTokenCredentialProvider {
         Logger.warn(methodTag, "mRefreshTokenCredentialHolder is not initialized!")
         return null
     }
+
+    override fun getRefreshTokenCredential(inputUrl : String, username : String) : String? {
+        val methodTag = "$TAG:getRefreshTokenCredential";
```
p3dr0rv, line 54: nit: use safe call instead of not null assertion 
return mRefreshTokenCredentialProvider?.getRefreshTokenCredential(inputUrl, username)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CrossCloudChallengeHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    // Updates the headers by attaching a refresh token credential header.
+    private fun modifyHeadersWithRefreshTokenCredential(
+        url: String,
+    ) {
+        val methodTag = "$TAG:modifyHeadersWithRefreshTokenCredential"
+        val parameters: Map<String, String> = StringExtensions.getUrlParameters(url)
+        val username = parameters["login_hint"]
+        if (!StringUtils.isNullOrEmpty(username)) {
```
p3dr0rv: nit: use username.isNullOrEmpty(), this way you avoid null assertion, null assertion indicates a code smell in many cases

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CrossCloudChallengeHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+            val updatedRefreshTokenCredentialHeader =
+                CommonRefreshTokenCredentialProvider.getRefreshTokenCredential(
+                    url, username!!
+                )
+            if (!StringUtils.isNullOrEmpty(updatedRefreshTokenCredentialHeader)) {
```
p3dr0rv: same here updatedRefreshTokenCredentialHeader.isNullorEmpty()

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/CrossCloudChallengeHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Handler for attaching prt credential header on web view in cross cloud redirect scenarios.
+ */
+class CrossCloudChallengeHandler(
+    private val webView: WebView,
+    private val headers: HashMap<String, String>,
+    private val span : Span
+) : IChallengeHandler<URL, Void> {
+    private val TAG = CrossCloudChallengeHandler::class.java.simpleName
+
+    override fun processChallenge(input: URL): Void? {
```
p3dr0rv: if we are not using URL capabilities, why don't we use string directly and avoid .toString()?
rename inputl

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -300,6 +312,25 @@ private boolean isNonceRedirect(@NonNull final String url) {
...
+            final URL currentUrl = new URL(url);
+            final URL startUrl = new URL(mRequestUrl);
+            if (AzureActiveDirectory.isValidCloudHost(currentUrl) && AzureActiveDirectory.isValidCloudHost(startUrl)) {
+                final Authority startUrlAuthority = Authority.getAuthorityFromAuthorityUrl(mRequestUrl);
+                final Authority currentLoadedUrlAuthority = Authority.getAuthorityFromAuthorityUrl(url);
+                if (!startUrlAuthority.getAuthorityURL().getHost().equalsIgnoreCase(currentLoadedUrlAuthority.getAuthorityURL().getHost())) {
+                    Logger.info(TAG, "Detected a cross cloud redirect.");
+                    return true;
+                }
+            }
```
mohitc1, line 326: is this logic same as iOS's?

somalaya: Yes, iOS checks for 2 things
1. Is it a known cloud
2. If yes, then it checks if the host part of the startUrl and inputUrl are the same. 
image

mohitc1 (approved):
:shipit:

p3dr0rv (approved):

PR #2604, opened 2025-03-13
[WIP] Add a common implementation for the BrokerRestrictionsManager, Fixes AB#3181005

p3dr0rv:
## Context
Currently, for QR + PIN authentication, users are prompted for camera permission every time the camera is used, even if OS-level permissions are granted. This is a CELA requirement to ensure user consent. However, a new feature request aims to suppress this camera permission prompt if an admin enables a specific flag in the restriction manager of the authenticator app.

## Changes
* The current PR involves refactoring the RestrictionsManagerHelper class to make it reusable for reading the new flag.
* Create a new class, CommonBrokerRestrictionsManager, that implements the IBrokerRestrictionsManager interface.

Upcoming PRs will leverage this flag to implement functionality that suppresses the camera permission prompt when enabled by the admin.

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3066
[AB#3181005](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3181005)


fadidurah (dismissed):

common/src/main/java/com/microsoft/identity/common/internal/util/RestrictionsManagerHelper.kt
```diff
@@ -0,0 +1,258 @@
...
+        return restrictions?.getString(key) ?: defaultValue
+    }
+
+    /**
+     * Gets the [key] value from the [appPackageName] restrictions manager.
+     * Returns null if the key is not found or failed to read the value.
+     * The default [appPackageName] is the current app.
+     */
+    fun getBoolean(key: String, appPackageName: String = context.packageName, defaultValue: Boolean = false): Boolean {
+        val methodTag = "$TAG:getString"
```
mohitc1: >getString [](http://example.com/codeflow?start=30&length=9)

fix tag

common/src/main/java/com/microsoft/identity/common/internal/util/RestrictionsManagerHelper.kt
```diff
@@ -0,0 +1,258 @@
...
+
+import com.microsoft.identity.common.internal.broker.ipc.BrokerOperationBundle
+import com.microsoft.identity.common.internal.broker.PackageHelper
+import com.microsoft.identity.common.java.interfaces.IPlatformComponents
+import com.microsoft.identity.common.logging.Logger
+
+/**
+ * Helper class to read the app restrictions manager of the current broker app or another broker app.
+ */
+class RestrictionsManagerHelper(
```
mohitc1: >RestrictionsManagerHelper [](http://example.com/codeflow?start=6&length=25)

If you're going to use this in WebViewAuthorizationFragment, how will it work in OneAuth only flow?

p3dr0rv: If we are in a non-broker flow, we do not call RestrictionsManagerHelper for this QR PIN scenario, checking the Activity type that contains the fragment.

However, in a hypothetical scenario where RestrictionsManagerHelper is used in a non-broker flow:
1. Requesting data from a broker app: will make an IPC call to the broker app, read the Restrictions manager from there, and return the result.
2. Requesting data from the current app: will read the Restrictions manager of the app, which will likely be null.

mohitc1: I raised this because I feel like this is broker only component is being put in common and can potentially run in non-brokered flow. It would be better if could inject an object in common layer (or WebViewAuthorizationFragment in this case) to _customize_ behavior for broker vs other. It's general problem that is seen on other cases as well e.g. nonce redirect handling, flighting 

I guess, it might be too much work if you do that in this PR?

mohitc1: Q: For what would be logic for non-broker flow?

common/src/main/java/com/microsoft/identity/common/internal/util/RestrictionsManagerHelper.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    /**
+     * Fetches the restriction manager bundle from the [appPackageName],
+     * filtering the keys based on the [dataRequired] bundle.
+     */
+    private fun fetchRestrictionsFromApp(
+        appPackageName: String,
+        dataRequired: Bundle
+    ): Bundle? {
+        val methodTag = "$TAG:fetchRestrictionManagerBundleFromTargetApp"
```
mohitc1: >fetchRestrictionManagerBundleFromTargetApp [](http://example.com/codeflow?start=30&length=42)

here too

common/src/main/java/com/microsoft/identity/common/internal/util/RestrictionsManagerHelper.kt
p3dr0rv: for non-broker flow we do not check the restriction manager, and we just grant the camera permission if we have it.
But even if the component were called in a non-broker flow it will keep working as I change it to only read RestrictionsManager  from broker apps

PR #2608, opened 2025-03-18
Add a telemetry field to track whether or not we are creating an IntuneAppProtectionPolicyRequiredException, Fixes AB#3183025

fadidurah:
Add a telemetry field to track whether or not we are creating an IntuneAppProtectionPolicyRequiredException

[AB#3183025](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3183025)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -339,7 +339,7 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
 
         if (exception instanceof IntuneAppProtectionPolicyRequiredException) {
             // Record MAM flow in telemetry
-            SpanExtension.current().setAttribute(AttributeName.is_mam_flow.name(),true);
+            SpanExtension.current().setAttribute(AttributeName.is_mam_flow.name(), true);
 
             builder.userName(((IntuneAppProtectionPolicyRequiredException) exception).getAccountUpn())
                     .localAccountId(((IntuneAppProtectionPolicyRequiredException) exception).getAccountUserId())
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -338,11 +338,8 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
         }
 
         if (exception instanceof IntuneAppProtectionPolicyRequiredException) {
-            // Record presence of exception in telemetry
-            SpanExtension.current().setAttribute(
-                    AttributeName.is_mam_flow.name(),
-                    Boolean.toString(true)
-            );
+            // Record MAM flow in telemetry
+            SpanExtension.current().setAttribute(AttributeName.is_mam_flow.name(),true);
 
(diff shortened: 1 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -333,5 +333,10 @@ public enum AttributeName {
     /**
      * Records the stacktrace for an out-of-memory exception.
      */
-    out_of_memory_exception_stacktrace
+    out_of_memory_exception_stacktrace,
+
+    /**
+     * Records if an IntuneAppProtectionPolicyRequiredException is received as part of this request.
+     */
+    intune_protection_policy_exception_present
```
shahzaibj: I don't think we want a separate field for this and `enrollment_id_present`. We discussed in scrum that these 2 should essentially be a single field called as `is_mam_flow` and emit it as true if either we get an IntuneAppProptectionPolicyException OR we have an enrollment id.

shahzaibj (changes requested):
Left a comment

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
fadidurah: Gotcha, we can do that. I figured having the separation could be useful, and when we use an MV or query we can use an or expression to check for both fields, but i can create just one for simplicity

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -338,6 +338,12 @@ public Bundle bundleFromBaseException(@NonNull final BaseException exception,
         }
 
         if (exception instanceof IntuneAppProtectionPolicyRequiredException) {
+            // Record presence of exception in telemetry
+            SpanExtension.current().setAttribute(
+                    AttributeName.is_mam_flow.name(),
+                    Boolean.toString(true)
```
shahzaibj: Is there a reason we need to convert to String? What are we doing for all the existing boolean? I suggest we follow the same pattern

fadidurah: Honestly, I don't think it makes a difference. I was originally not converting, but i saw another field being set with string conversion and it made sense to do it. Let me check frequency of each

fadidurah: Seems like string conversion for booleans is mainly used outside of telemetry purposes, i've switched back to just passing the boolean itself

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -338,6 +338,9 @@ public Bundle bundleFromBaseException(@NonNull final BaseException exception,
         }
 
         if (exception instanceof IntuneAppProtectionPolicyRequiredException) {
+            // Record MAM flow in telemetry
+            SpanExtension.current().setAttribute(AttributeName.is_mam_flow.name(),true);
```
shahzaibj:
```suggestion
            SpanExtension.current().setAttribute(AttributeName.is_mam_flow.name(), true);
```

shahzaibj (approved):
Please also add a changelog entry

melissaahn (approved):

PR #2612, opened 2025-03-25
Add telemetry for the switch browser protocol

p3dr0rv:
# Adding telemetry for Switch browser protocol.

## How to read the data?

- ATIInteractively.is_switch_browser_protocol indicates that the client is sending a switch browser protocol.
- SwitchBrowserProtocol.is_switch_browser_request_handled indicates that the webview received a switch browser redirect URL and launched the browser with the following characteristics: browser_package_name, is_custom_tabs_supported.
- SwitchBrowserProtocol.is_switch_browser_resume_handled indicates that the broker redirect URL in the browser was received and handled by the webview.

In conclusion, in a successful scenario, we should have 2 SwitchBrowserProtocol entries: one for the request and another for the resume, and the parent span_id should be the ATIInteractively span_id.

## Example

android_spans
| where EventInfo_Time  between(ago(1d)..now()) 
| where DeviceInfo_Model == "SM-G988U"
| where DeviceInfo_Id == "3f0fa7b42bed152f"
| project span_name,trace_id, span_id, parent_span_id, is_switch_browser_protocol, browser_package_name, is_custom_tabs_supported, is_switch_browser_request_handled, is_switch_browser_resume_handled
| order by trace_id

span_name | trace_id | span_id | parent_span_id | is_switch_browser_protocol | browser_package_name | is_custom_tabs_supported | is_switch_browser_request_handled | is_switch_browser_resume_handled
-- | -- | -- | -- | -- | -- | -- | -- | --
ATIInteractively | 8a92ec54a756969a499d70d67901abb6 | bb2b33d4b17f8dae | 2dc57e2a0a44e41f | true |   |   |   |  
AcquireTokenInteractive | 8a92ec54a756969a499d70d67901abb6 | 2dc57e2a0a44e41f | 0000000000000000 |   |   |   |   |  
SwitchBrowserProtocol | 8a92ec54a756969a499d70d67901abb6 | 7f2edab5f5801692 | bb2b33d4b17f8dae |   | com.android.chrome | true | true |  
SwitchBrowserProtocol | 8a92ec54a756969a499d70d67901abb6 | 71f66086a4c6c15a | bb2b33d4b17f8dae |   |   |   |   | true
SwitchBrowserProtocol | 81f98ddfaf5dad5fa511d3afe4a7cb37 | e8275d7dab01427d | e5d0a07c8d826a5f |   | com.android.chrome | true | true |  
SwitchBrowserProtocol | 81f98ddfaf5dad5fa511d3afe4a7cb37 | 20e90241b19c1f1e | e5d0a07c8d826a5f |   |   |   |   | true
ATIInteractively | 81f98ddfaf5dad5fa511d3afe4a7cb37 | e5d0a07c8d826a5f | d771e73d1ddcfacf | true |   |   |   |  
AcquireTokenInteractive | 81f98ddfaf5dad5fa511d3afe4a7cb37 | d771e73d1ddcfacf | 0000000000000000 |   |   |   |   |  

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3069


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SwitchBrowserRequestHandler.kt
```diff
@@ -42,9 +49,13 @@ class SwitchBrowserRequestHandler(
     private val activity: Activity,
     private val context: Context,
     private val customTabsManager: CustomTabsManager,
-    private val browserSelector: IBrowserSelector
+    private val browserSelector: IBrowserSelector,
+    private val spanContext: SpanContext?
 ) : IChallengeHandler<SwitchBrowserChallenge, Unit> {
 
+    val span: Span by lazy {
+        OTelUtility.createSpanFromParent(SpanName.SwitchBrowserProtocol.name, spanContext)
```
fadidurah: Is there not an existing span at this point? Seems excessive to create a new span for this

shahzaibj:
Why parent span id is ATIInteractively? Why is it not AcquireTokenInteractive?

shahzaibj:
I think prior to releasing this we need to increase the sampling rate for interactive requests, otherwise we will lose important data

shahzaibj:
Since you are doing /process and /resume in separate span records, then why even have a generic `SwitchBrowserProtocol` span? Perhaps instead should have 2 distinct spans: `SwitchBrowserProcess` and `SwitchBrowserResume` with their own sets of required fields.

That said, +1 to Fadi's comment, we should see if we can put the information directly on AcquireTokenInteractive spans to save cost.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SwitchBrowserRequestHandler.kt
p3dr0rv: Once we launch the intent to WebView, there is no active span in this context. You can look at other features that capture telemetry at this point and then create a new span.


Example: in AzureActiveDirectoryWebViewClient
is_sso_nonce_found_in_ests_request is not captured as the span doesn't exist.
We need to create spans like we do for ProcessNonceFromEstsRedirect, ProcessCrossCloudRedirect, Fido.


https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2209

cc: @shahzaibj

p3dr0rv:
> Why parent span id is ATIInteractively? Why is it not AcquireTokenInteractive?

I think we can remove it, seems like an unnecessary span all this data can be in AcquireTokenInteractive span.
All the places where ATIInteractively is called come from a AcquireTokenInteractive span

But removing this can be dangerous and affect dashboards or materialized views, so we might want to do this in a different PR

p3dr0rv:
> en have a generic `SwitchBrowserProtocol` span? Perhaps instead should have 2 distinct spans: `SwitchBrowse`

Yes, I am fine with that.

p3dr0rv:
> > en have a generic `SwitchBrowserProtocol` span? Perhaps instead should have 2 distinct spans: `SwitchBrowse`
> 
> Yeah, I am fine with that, I just did it this way to have more granularity

Regarding the cost of this span, this is lazy span so it will only be generated if is a DUNA flow is invoked.

fadidurah:
> > Why parent span id is ATIInteractively? Why is it not AcquireTokenInteractive?
> 
> I think we can remove it, seems like an unnecessary span all this data can be in AcquireTokenInteractive span. All the places where ATIInteractively is called come from a AcquireTokenInteractive span
> 
> But removing this can be dangerous and affect dashboards or materialized views, so we might want to do this in a different PR

Maybe we should flight it?

fadidurah (approved):

shahzaibj (approved):

shahzaibj (approved):

PR #2613, opened 2025-03-27
Only show QR +PIN rationale when the request origin is from Microsoft

p3dr0rv:
This change here is a temporary fix for QR + PIN.
The final solution will be incorporated in the next month April 2025.
On the final solution we only allow QR + PIN prompt when the Authenticator RestrictionsManager contains QR PIN key.

This fix will limit the QR +PIN camera consent prompt to requests coming from login.microsoftonline.com only.
We also reworked the CameraPermissionRequest to avoid NPR when PermissionRequest is null, by calling only the CameraPermissionRequestHandler that is null safe.


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -318,36 +317,36 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+                    );
+                    // We can only grant or deny permissions for video capture/camera.
+                    // To avoid unintentionally granting requests for not defined permissions
+                    // we check if the request is for camera.
+                    if (!mCameraPermissionRequestHandler.setIfValid(request)) {
+                        Logger.warn(methodTag, "Permission request is not valid, returning.");
+                        return;
+                    }
+
+                    // If the OS level permission was granted previously,
```
mohitc1, line 335: I think all logic can go inside the CameraPermissionHandler

if (isCameraRequest(request)) {
     val ret = cameraPermissionHandler.handle(request)
}

since this is temporary fix, you may do that along with next month changes.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequestHandler.kt
```diff
@@ -0,0 +1,163 @@
...
+     * If valid, set the current permission request and return true.
+     * If not valid, deny the request and return false.
+     * <p>
+     * A request is considered valid if:
+     * - The request is for the camera resource only.
+     * - The request is not a repeated request (i.e., not for the same origin and resources).
+     *
+     */
+    fun setIfValid(request: PermissionRequest): Boolean {
+        val methodTag = "$TAG:isValid"
```
mohitc1: >"$TAG:isValid" [](http://example.com/codeflow?start=24&length=14)

nit: tag value

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -318,36 +317,36 @@ public boolean onTouch(final View view, final MotionEvent event) {
...
+                    // we show the rationale to confirm the consent with the current user.
+                    // Otherwise, show the system prompt.
+                    if (isAppCameraPermissionGranted()) {
+                        Logger.info(methodTag, "App level camera permission already granted.");
+                        if(request.getOrigin() != null &&  "https://login.microsoftonline.com/".equalsIgnoreCase(request.getOrigin().toString())) {
```
fadidurah: Nit: I think we can put "https://login.microsoftonline.com/" in a constant.

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: Good idea, but yes, due to time constraints I will fix this on next release

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: sure

PR #2616, opened 2025-03-31
Improve logs and error reporting SIWG flow

mohitc1:
Changes:
1) Improve logs method tag + messages
2) For GetCredentialCustomCredential, adding type as sub error code.
3) If bottom sheet UX (GetGoogleIdOption) is used, and NoCredentialException is received, this indicates no valid credential is present (could be either no account present at all or account is there but credential is invalid and need re-auth). In such case, start GetSignInWithGoogleOption based flow to allow re-auth or adding google account
4) added UTs

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [PATCH] Improve logs and error reporting SIWG flow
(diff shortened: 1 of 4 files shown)


fadidurah (approved):
LGTM

somalaya (approved):

PR #2617, opened 2025-04-02
Expose a JavaScript API in brokered Webviews to facilitate Improved Same Device NumberMatch , Fixes AB#3203956

fadidurah:
SPEC: https://microsoft-my.sharepoint-df.com/:w:/r/personal/siddhijain_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7BD1D944D5-2047-40AB-B8F9-98506BF212A3%7D&file=Engineering%20design%20-%20Number%20matching%20on%20same%20device.docx&action=default&mobileredirect=true&share=IQHVRNnRRyCrQLj5mFBr8hKjAbj81fEnpO6X-99laqs2j_M&wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1743094076241&web=1

Word Doc: for JavaScript Api: https://microsoft-my.sharepoint-df.com/:w:/p/veenasoman/EY1AZIeT8X5KrXVz97Vx520B3Jj0fBLSPlklnoRvcmbh0Q?e=ZVVUrw&nav=eyJoIjoiMjEzMzE1Mzg5NSJ9

Structure has changed a bit for this. To facilitate future work, we will have a generalized JavaScript API that takes in a json string payload. This is used to parse out a function name, and data field, both of which are used to call a specific function in broker code. This same functionality will be used next month for CA Block improvment work (I don't have a spec to this one yet).

Expected method call in JavaScript is now something like this, we are working on finalizing json schema:
`BrokerJS.postToBroker('{function: NUMBER_MATCH,data: {sessionID: id, numberMatch: number}}')`

I added some unit tests in the broker PR, but primary validation will be when ests exposes a test slice that calls the JavaScript API. Did some testing in our webview class to call javascript code, and was able to prompt the numberMatch method.

Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3073

[AB#3203956](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3203956)


common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,73 @@
...
+ * supply the number match value and store it in ephemeral storage (kept as long as current broker
+ * process is alive), where Authenticator can call a broker API to fetch the number match, and immediately
+ * prompt user for consent, rather than first asking them to check the number match.
+ */
+class NumberMatchJavaScriptInterface {
+
+    // Store number matches in a static hash map
+    // No need to persist this storage beyond the current broker process, but we need to keep them
+    // long enough for AuthApp to call the broker api to fetch the number match
+    // TODO: Is this a security concern?
```
shahzaibj: As a starter, we should ensure that only AuthApp can call this API

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,73 @@
...
+        numberMatchMap[sessionID] = numberMatch
+    }
+
+    /**
+     * Clear existing number match key:value pairs
+     */
+    fun clearNumberMatchMap() {
+        numberMatchMap.clear()
+    }
+}
```
shahzaibj: nit: new line ending

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,73 @@
...
+    companion object {
+        val TAG = NumberMatchJavaScriptInterface::class.java.simpleName
+        val numberMatchMap: HashMap<String?, String?> = HashMap()
+    }
+
+    /**
+     * Method to add a key:value pair of sessionID:numberMatch to static hashmap. This hashmap will be accessed
+     * by broker api to get the number match for a particular sessionID.
+     *
+     * TODO: Check, I feel params should be nullable, and should fallback to null check in AuthApp code.
```
shahzaibj: If you are marking them as nullable then you should also check for them being non-null before putting them in the map

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -214,6 +217,9 @@ void extractState(@NonNull final Bundle state) {
         mAuthIntent = state.getParcelable(AUTH_INTENT);
         mPkeyAuthStatus = state.getBoolean(PKEYAUTH_STATUS, false);
         mAuthorizationRequestUrl = state.getString(REQUEST_URL);
+        if (mAuthorizationRequestUrl != null) {
+            isBrokerRequest = checkBrokerRequest(mAuthorizationRequestUrl);
```
shahzaibj: Can we instead check if we are inside :auth process? I think we do that in some parts of the code

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: I meant is storing them in a static hashmap a security concern. I don't think it would be, but wanted to check.

fadidurah: This api is called by AuthUX, the api called by authenticator is in the broker PR, i can add the check there 😄

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: Yeah, this makes sense to me. I don't want to have be non-null to avoid a crash, so we can fall back to existing behavior. Having them nullable with a check makes sense

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
fadidurah: I'll look into this, that would be better

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,74 @@
...
+        val TAG = NumberMatchJavaScriptInterface::class.java.simpleName
+        val numberMatchMap: HashMap<String?, String?> = HashMap()
+    }
+
+    /**
+     * Method to add a key:value pair of sessionID:numberMatch to static hashmap. This hashmap will be accessed
+     * by broker api to get the number match for a particular sessionID.
+     */
+    @JavascriptInterface
+    fun postCodeMatch(sessionID : String?, numberMatch : String?) {
```
somalaya: can we make SessionId  non null?

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,74 @@
...
+ * prompt user for consent, rather than first asking them to check the number match.
+ */
+class NumberMatchJavaScriptInterface {
+
+    // Store number matches in a static hash map
+    // No need to persist this storage beyond the current broker process, but we need to keep them
+    // long enough for AuthApp to call the broker api to fetch the number match
+    companion object {
+        val TAG = NumberMatchJavaScriptInterface::class.java.simpleName
+        val numberMatchMap: HashMap<String?, String?> = HashMap()
```
shahzaibj: Why do the key, value in map have to be nullable? I think you are already checking for nullability prior to putting them in the map, right? So this here is redundant

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,74 @@
...
+ * matching challenge is issued, where used is given a number and asked to open authenticator and check
+ * for the same number in authenticator UI. This feature cuts out one UI step, where this API is used to
+ * supply the number match value and store it in ephemeral storage (kept as long as current broker
+ * process is alive), where Authenticator can call a broker API to fetch the number match, and immediately
+ * prompt user for consent, rather than first asking them to check the number match.
+ */
+class NumberMatchJavaScriptInterface {
+
+    // Store number matches in a static hash map
+    // No need to persist this storage beyond the current broker process, but we need to keep them
```
shahzaibj: If the broker process dies for some reason then what happens?

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Java Script API to be exposed as part of brokered WebView requests.
+ * Will be used by AuthUX as part of number match feature.
+ * When authenticator is installed, and phone uses MFA or PSI in an interactive flow, a number
+ * matching challenge is issued, where used is given a number and asked to open authenticator and check
+ * for the same number in authenticator UI. This feature cuts out one UI step, where this API is used to
+ * supply the number match value and store it in ephemeral storage (kept as long as current broker
+ * process is alive), where Authenticator can call a broker API to fetch the number match, and immediately
+ * prompt user for consent, rather than first asking them to check the number match.
```
shahzaibj: So with this user doesn't have to select the matching number?

Or rather is the benefit is that the UX to select matching number can be done inline?

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: If the broker process dies then this number match data is lost, but the interactive flow is killed as well anyway. When user starts another request, a new number match will be saved that authenticator can look at. Regardless, nothing breaks if this data is not available, Authenticator would just fall back to existing behavior of asking user to manually enter the number match.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: Auth UX will still show the number match to the user as well as sending it through the API, authenticator will use sessionID to query for the number match and try to bypass the manual entry step, and only ask user to authorize the sign in.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: They don't, missed changing this since the first revision

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchJavaScriptInterface.kt
fadidurah: There's no guarantee that authux will not send a null value, from what i understand. We have a meeting later today to discuss with sidhi

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterface.kt
```diff
@@ -0,0 +1,120 @@
...
+            val correlationID = parsedJson["correlationID"]
+            Logger.info(methodTag, "Correlation ID during JavaScript Call: [$correlationID]")
+
+            // TODO: Leaving these here, as these will be relevant for next WebCP feature
+            // val actionName = parsedJson["action_name"]
+            // val actionComponent = parsedJson["action_component"]
+
+            val parameters = JsonUtil.extractJsonObjectIntoMap(parsedJson["params"])
+            val function = parameters["function"]
+            val data = JsonUtil.extractJsonObjectIntoMap(parameters["data"])
```
shahzaibj: Let's define a DTO representing the json schema and then you can just use the kotlinx serialization libraries (or even GSON or Moshi if you prefer that) to deserialize the raw json directly to a custom dto.

For instance:

```java
@Serializable
data class AuthUxApiResponse(
    val correlationId: String,
    val actionName: String,
    val actionComponent: String,
    val params: Params; // define this as a DTO as well
)
```

and then you can deserialize as follows using kotlinx serialization framework

```java
val authUxResponse: AuthUxApiResponse = Json.decodeFromString(jsonString)
```

and then you can access individual properties as follows:

```java
   val numberMatch = authUxResponse.params.data.numberMatch;
```

fadidurah: Neat! I can adjust to this, this current Json parsing implementation is a bit janky

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -211,6 +217,12 @@ void extractState(@NonNull final Bundle state) {
         mAuthIntent = state.getParcelable(AUTH_INTENT);
         mPkeyAuthStatus = state.getBoolean(PKEYAUTH_STATUS, false);
         mAuthorizationRequestUrl = state.getString(REQUEST_URL);
+        if (mAuthorizationRequestUrl != null) {
+            isEstsRequest = mAuthorizationRequestUrl.startsWith("https://login.microsoftonline.com");
```
shahzaibj: What about sovereign clouds?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -122,6 +125,9 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
     // This is used by the switch browser protocol to handle the resume of the flow.
     private SwitchBrowserProtocolCoordinator mSwitchBrowserProtocolCoordinator = null;
 
+    private boolean isBrokerRequest = false;
+    private boolean isEstsRequest = false;
```
shahzaibj: TBH, I don't think we make requests anywhere besides eSTS so all of our requests are to eSTS. So not sure what this check is doing

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -290,6 +302,9 @@ private void setUpWebView(@NonNull final View view,
         mWebView.getSettings().setUserAgentString(
                 userAgent + AuthenticationConstants.Broker.CLIENT_TLS_NOT_SUPPORTED);
         mWebView.getSettings().setJavaScriptEnabled(true);
+        if (isBrokerRequest && isEstsRequest) {
```
shahzaibj: What happens once the URL changes within the webview? So we go from eSTS to ADFS for federated auth but the `isEstsRequest` would still say true because it was defined earlier in the flow. This would be a problem

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
shahzaibj: May be you intend to check for the "current" URL that we are navigating to within the webview? I think that would be ideal but I don't think your code is doing that at the moment

fadidurah: No i don't think it's doing that either. I'm not sure if this check is really necessary honestly. The whole point is to make sure the JS API will not be called by any malicious actors, but my understanding is that we already have validation that makes it so our webview is only launched in ESTS pages, and unless they redirect onto a 3rd party page where a malicious actor may have access, then our JS api should be secure. Let me know if this is correct.

fadidurah: Idea was that if we validate webview is originally being launched on an ests page, then we should be safe from there, assuming ests is not actively redirecting us to malicious pages. I think this section will need some further consideration.

shahzaibj: I don't think eSTS will redirect to malicious page, however, we should still limit the usage of this API from just the eSTS page itself (if possible). So if eSTS redirects to ADFS or any third party page then this JS API shouldn't be available to that 3p domain.

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Data class representing the JSON payload object received from AuthUX.
+ *
+ * @property correlationId The correlation ID for the request.
+ * @property actionName The name of the action being performed.
+ * @property actionComponent The component responsible for the action.
+ * @property params The parameters for the action, including function and data.
+ */
+data class AuthUxJsonPayloadObject(
```
shahzaibj: Ofcourse this is an object. We don't have to explicitly state it though. Let's just rename this to `AuthUxJsonPayload`

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
```diff
@@ -0,0 +1,86 @@
...
+    val correlationId: String?,
+
+    @SerializedName(SerializedNames.ACTION_NAME)
+    val actionName: String?,
+
+    @SerializedName(SerializedNames.ACTION_COMPONENT)
+    val actionComponent: String?,
+
+    @SerializedName(SerializedNames.PARAMS)
+    val params: AuthUxParams?
```
shahzaibj: Why is everything nullable? I think one of these fields being optional makes sense but it doesn't make sense for everything to be nullable

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
```diff
@@ -0,0 +1,86 @@
...
+)
+
+object SerializedNames {
+    const val CORRELATIONID = "correlationID"
+    const val ACTION_NAME = "action_name"
+    const val ACTION_COMPONENT = "action_component"
+    const val PARAMS = "params"
+    const val FUNCTION = "function"
+    const val DATA = "data"
+    const val SESSIONID = "sessionID"
```
shahzaibj: suggesting rename `SESSIONID` to `SESSION_ID`
and `NUMBERMATCH` to `NUMBER_MATCH`

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
```diff
@@ -0,0 +1,86 @@
...
+    const val NUMBERMATCH = "numberMatch"
+}
```
shahzaibj: nit: new line ending

common/src/test/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterfaceTest.kt
```diff
@@ -64,4 +71,20 @@ class AuthUxJavaScriptInterfaceTest {
...
+
+        // Should not get an exception
+    }
+
+    @Test
+    fun `test postMessageToBroker with non-json string`() {
+        // Call the method
+        authUxJavaScriptInterface.postMessageToBroker("NotAJson")
+
+        // Should not get an exception
```
shahzaibj, line 89: Can we also validate that message was actually posted?

common/src/test/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObjectTest.kt
```diff
@@ -0,0 +1,75 @@
+package com.microsoft.identity.common.internal.broker
```
shahzaibj: LICENSE

common/src/test/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObjectTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        val payload = gson.fromJson(json, AuthUxJsonPayloadObject::class.java)
+
+        assertNotNull(payload)
+        assertNull(payload.correlationId)
+        assertNull(payload.actionName)
+        assertNull(payload.actionComponent)
+        assertNull(payload.params)
+    }
+}
```
shahzaibj: nit: new line ending

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
fadidurah: Handling in next commits

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
fadidurah: Moving the handling to `AzureActiveDirectoryWebViewClient` and `handleUrl`

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
fadidurah: Moving the javascript handling to `AzureActiveDirectoryWebViewClient`, can setup javascript api upon method call after construction, and will re-evaluate upon each `handleUrl`

common/src/test/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObjectTest.kt
fadidurah: License is there, I think you are on an older commit

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
fadidurah: Renamed

common/src/test/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterfaceTest.kt
fadidurah: What can be done to validate this? Copilot seems to want to validate the log statement, but can't get it to work

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
fadidurah: I think allowing everything to be nullable made sense because we can have specific logging/handling for null fields in the  JSinterface. The object itself is only meant to communicate the payload, so i wanted to leave null handling to the JS interface.

shahzaibj: But what exactly is the payload contract between client and server? If the contract says that certain fields (for instance actionName) cannot ever be null then we should mark it as such here.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1222,6 +1222,16 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String REDIRECT_PREFIX = "msauth";
 
+        /**
+         * Prefix for AAD urls
+         */
+        public static final String AAD_URL_PREFIX = "https://login.microsoftonline.";
```
shahzaibj: We might not need the period in the end if we choose to validate the host part of the authority only (as opposed to using strings to validate prefixes)

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterface.kt
```diff
@@ -0,0 +1,155 @@
...
+         */
+        fun isValidUrlForInterface(url: String?): Boolean {
+            // If url is null, return false
+            if (url == null) {
+                return false
+            }
+
+            // Otherwise, make sure url is a valid url
+            // We only want to allow URLs that start with the AAD URL prefix or the MSA URL prefix
+            return url.startsWith(AuthenticationConstants.Broker.AAD_URL_PREFIX) ||
```
shahzaibj: Mentioned in another comment, probably better to convert String url to URL object and check the host part of the URL only for validation

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
fadidurah: Ok, had to create a custom deserializer, because gson does not recognize non-null annotation or kotlin implicit non null as part of type declaration. Also added tests for these required fields (`correlationID`, `action_name`, `action_component`, `params`)

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: host in the url still includes the `.com`, or `.us` or `.cn` in sovereign cloud scenarios. I did switch to using hosts specifically

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
shahzaibj: Hmm. Not a fan of this custom deserialization. Since you are using class defined in kotlin then why not using kotlin deserialization? I believe that should take care kotlin non-null implicit declarations

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJsonPayloadObject.kt
shahzaibj: Discussed offline....

Kotlinx deserialization requires separate dependency and so does Moshi and that can increase size of SDK. We can do it later. For now sticking to GSON. It isn't great but this is what we have in the code today.

fadidurah: kotlin deserialization requires adding an additional plugin and dependency. that didn't really seem worth considering we already had Gson and Moshi. For Moshi, there was a similar problem, basically we need another package as a dependency to have moshi working properly for kotlin. Only solution that didn't involve adding an additional dependency was using gson with custom serializer

shahzaibj (approved):

PR #2618, opened 2025-04-02
Enable the new Broker discovery on MSAL side by default, Fixes AB#3226173

rpdome:
Keeps the method to avoid introducing breaking change to OneAuth.

[AB#3226173](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3226173)

PR #2625, opened 2025-04-14
Revert #2618

rpdome:


fadidurah (approved):

p3dr0rv (approved):
You published the reason on teams' channel, it will be a good idea to put here as well

fadidurah (approved):

PR #2628, opened 2025-04-21
Allow Intune CE app to force broker discovery, Fixes AB#3237686

rpdome:
[AB#3237686](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3237686)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1162,6 +1162,18 @@ public final class AuthenticationConstants {
         public static final String LTW_APP_SHA512_DEBUG_SIGNATURE = "5PAhhZNSRRvq7vpTT5vrYJbSLh05AU8USf7oUTS239PEltebX87uGN7GhAe5244lJepwZ5RU4vu8N6ospXVOlg==";
 
         /**
+         * Signing certificate thumbprint of the DEBUG-signed Microsoft Intune app.
+         * Generated with SHA-512.
+         */
+        public static final String INTUNE_APP_SHA512_DEBUG_SIGNATURE = "F+Tat7A/mlOJCzRYEmj9DgLRHU2Nb0VSQjgZEyAehqW9+cOT0oYjkT/fa33hYcVMwUzaSy0hUOVt9KQtyFRnVQ==";
+
+        /**
+         * Signing certificate thumbprint of the RELEASE-signed Microsoft Intune app.
+         * Generated with SHA-512.
+         */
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1161,6 +1161,18 @@ public static String computeMaxHostBrokerProtocol() {
...
+         * Signing certificate thumbprint of the DEBUG-signed Microsoft Intune app.
+         * Generated with SHA-512.
+         */
+        public static final String INTUNE_APP_SHA512_DEBUG_SIGNATURE = "F+Tat7A/mlOJCzRYEmj9DgLRHU2Nb0VSQjgZEyAehqW9+cOT0oYjkT/fa33hYcVMwUzaSy0hUOVt9KQtyFRnVQ==";
+
+        /**
+         * Signing certificate thumbprint of the RELEASE-signed Microsoft Intune app.
+         * Generated with SHA-512.
+         */
+        public static final String INTUNE_APP_SHA512_RELEASE_SIGNATURE = "jPpMoaNvcxSLMX4yG4C3Gf86rtTqh33SqpuRKg4WOP+MnnpA52zZgvKLW76U4Cqqf68iaBk9W7k/jhciiSAtgQ==";
```
fadidurah, line 1174: I always wondered, is it ok to have these signatures in a public repo?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
rpdome: should be. cert thumbprint is meant to be public.

the thumbprint allows everyone else to validate if something is signed with a specific "private" key. The key itself cannot be public.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -160,6 +160,20 @@ data class BrokerData(val packageName : String,
...
+            AuthenticationConstants.Broker.INTUNE_APP_SHA512_DEBUG_SIGNATURE,
+            "debugIntuneCE"
+        )
+
+        @JvmStatic
+        val prodIntuneCE = BrokerData(
+            AuthenticationConstants.Broker.INTUNE_APP_PACKAGE_NAME,
+            AuthenticationConstants.Broker.INTUNE_APP_SHA512_RELEASE_SIGNATURE,
+            "prodIntuneCE"
+        )
```
mohitc1, line 175: This seems special case here.. that this is app is not a broker itself but is broker API client. Add a comment here?

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
rpdome: comment added in the place where this is used.

somalaya (approved):

PR #2630, opened 2025-04-25
Move camera logic to CameraPermissionRequestHandler and add SdmQrPinManager , Fixes AB#3121976

p3dr0rv:
- Move IBrokerRestrictionsManager.kt from broker to common
- Add SdmQrPinManager 
- Move camera consent code from WebViewAuthorizationFragment.java to CameraPermissionRequestHandler.kt
- Camera logic change: previously we show camera QR PIN consent dialog if request origin is from MS, otherwise we use default behavior. Now we use the SdmQrPinManager to determine if we should show the camera QR PIN consent dialog.  (see:  CameraPermissionRequestHandler.handle(request: PermissionRequest, context: Context)

Related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3089

Run manual UI test for validation https://identitydivision.visualstudio.com/Engineering/_testPlans/execute?planId=2007357&suiteId=2810290

[AB#3121976](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3121976)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/IBrokerRestrictionsManager.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/IBrokerRestrictionsManager.kt
@@ -1,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.broker
+
+import android.os.Bundle
+
+/**
+ * Helper class to read the app restrictions manager of the current broker app or another broker app.
+ */
+interface IBrokerRestrictionsManager {
+    companion object BrokerRestrictionsManagerKeys {
+        /** Keys to group the values in the bundle. */
+        const val BOOLEAN_VALUES_KEY = "booleanValuesKey"
+        const val STRING_VALUES_KEY = "stringValuesKey"
+
+        /** Keys to be used in the app restrictions manager to read the values. */
+        // String keys
+        const val PREFERRED_AUTH_CONFIG = "preferred_auth_config"
+        // Boolean keys
+        const val SUPPRESS_CAMERA_CONSENT = "sdm_suppress_camera_consent"
+
+        /**
+         * Creates a request bundle with the keys to be requested from the app restrictions manager.
+         * The keys are filtered based on the type of the keys.
+         *
+         * @param stringKeys Keys to be requested from the app restrictions manager.
+         * @param booleanKeys Keys to be requested from the app restrictions manager.
+         * @return Bundle with the keys to be requested from the app restrictions manager.
+         */
+        @JvmOverloads
+        fun buildMultiValueRequest(stringKeys: Set<String>? = null, booleanKeys: Set<String>? = null): Bundle {
+            val stringList = stringKeys?.let { ArrayList(it) }
+            val booleanList = booleanKeys?.let { ArrayList(it) }
+            return Bundle().apply {
+                this.putStringArrayList(STRING_VALUES_KEY, stringList)
+                this.putStringArrayList(BOOLEAN_VALUES_KEY, booleanList)
+            }
+        }
+    }
+    /**
+     * Gets the [key] value from the [brokerAppPackageName] restrictions manager.
+     * Returns null if the key is not found or failed to read the value.
+     */
+    fun getString(key: String, brokerAppPackageName: String, defaultValue: String? = null): String?
+
+    /**
+     * Gets the [key] value from the [brokerAppPackageName] restrictions manager.
+     * Returns null if the key is not found or failed to read the value.
+     */
+    fun getBoolean(key: String, brokerAppPackageName: String, defaultValue: Boolean = false): Boolean
+
+    /**
+     * Reads the keys from the [bundleOfKeys] provided
+     * and returns a bundle with the values from the [brokerAppPackageName] restrictions manager.
+     * <p>
+     * The provided bundle should contain a key determining the type of the keys to be requested
+     * and the value should be an array list of keys.
+     * Example: if the bundle contains a key "stringValues" with an array list of keys ["key1s", "key2s"]
+     * and a key "booleanValues" with an array list of keys ["key0b"].
+     * Then the returned bundle will contain the values for the keys "key1s", "key2s" and "key0b".
+     *
+     *
+     * @param bundleOfKeys Bundle of keys to be requested from the app restrictions manager.
+     * @return Bundle with the values from the app restrictions manager.
+     */
+    fun getMultiValues(brokerAppPackageName: String, bundleOfKeys: Bundle) : Bundle
+}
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
```diff
@@ -0,0 +1,104 @@
...
+        preferredAuthMethod = multiValues.getString(PREFERRED_AUTH_CONFIG)
+        if (preferredAuthMethod == PreferredAuthMethod.QR.value) {
+            isDeviceOnSdmQrPinAuth = true
+        }
+        isCameraConsentSuppressed = multiValues.getBoolean(SUPPRESS_CAMERA_CONSENT)
+        Logger.info(
+            TAG,
+            "preferredAuthMethod: $preferredAuthMethod," +
+                    " isCameraConsentSuppressed: $isCameraConsentSuppressed"
+        )
```
mohitc1: Any reason you're doing reads in initialize itself?

Can we do it when getPreferredAuthMethod() and isDeviceOnSdmQrPinAuth is called?

common/src/main/java/com/microsoft/identity/common/internal/broker/IBrokerRestrictionsManager.kt
```diff
@@ -0,0 +1,88 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.broker
+
+import android.os.Bundle
+
+/**
+ * Helper class to read the app restrictions manager of the current broker app or another broker app.
+ */
+interface IBrokerRestrictionsManager {
```
mohitc1: >IBrokerRestrictionsManager [](http://example.com/codeflow?start=10&length=26)

Rename to IRestrictionsManager?

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
p3dr0rv: Those values ​​are unlikely to change since they are set by Intune policy, so I was thinking of avoiding calls to the broker and restrictions manager by simply reading all the values ​​at once and holding them for a while until the broker process is killed.

mohitc1: Is GetPreferredAuthMethodMsalBrokerOperation called only once by caller?

This method is called by GetPreferredAuthMethodMsalBrokerOperation every time

mohitc1: Is this always supposed to be called before CameraPermissionRequestHandler runs?

p3dr0rv: Before making a token request,
1 Teams calls the GetPreferredAuthMethodMsalBrokerOperation method. 
2 If the preferred authentication method is QR+PIN, Teams displays a button for signing in using the camera. 
3 The user then logs in with QR+PIN.

Teams and other applications using this authentication method always call GetPreferredAuthMethodMsalBrokerOperation before making a token request with the camera. This is because they will only use QR+PIN authentication if the preferred method is QR+PIN, and to determine the method, they need to call GetPreferredAuthMethodMsalBrokerOperation.

p3dr0rv: I chose to implement it this way to avoid making unnecessary calls for users who will use the camera but not QR+PIN. However, I have no issue moving it to where getPreferredAuthMethod() and isDeviceOnSdmQrPinAuth are called. As mentioned, this was just an optimization, but I will move it to be on the safe side in case the pattern changes. In this case, I will pass the broker restriction manager when we initialize platform components.

p3dr0rv: Agreed approach - I will read the value every time it is required, but I will initialize the object in the Android broker platform component.

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
```diff
@@ -0,0 +1,89 @@
...
+    private const val TAG = "SdmQrPinManager"
+
+    private var brokerRestrictionsManager: IRestrictionsManager? = null
+
+    /**
+     * This method is used to initialize the SDM QR PIN manager with the broker restrictions manager.
+     * This is called when brokerAndroidBrokerPlatformComponentsFactory.create is called
+     * to ensure the manager is initialized with the correct broker restrictions manager.
+     */
+    fun initializeSdmQrPinManager(brokerRestrictionsManager: IRestrictionsManager) {
```
mohitc1: >brokerRestrictionsManager [](http://example.com/codeflow?start=34&length=25)

nit: just restrictionsManager

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
```diff
@@ -0,0 +1,89 @@
...
+        if (this.brokerRestrictionsManager == null) {
+            this.brokerRestrictionsManager = brokerRestrictionsManager
+        }
+    }
+
+    /**
+     * This method checks if the device preferred authentication config is set.
+     */
+    fun getPreferredAuthConfig(): String? {
+        val methodTag = "$TAG:getPreferredAuthMethod"
```
mohitc1: >getPreferredAuthMethod [](http://example.com/codeflow?start=30&length=22)

fix tag

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
```diff
@@ -0,0 +1,89 @@
...
+        val defaultValue = null
+        return brokerRestrictionsManager?.getString(
+            key = PREFERRED_AUTH_CONFIG,
+            brokerAppPackageName = BrokerData.prodMicrosoftAuthenticator.packageName,
+            defaultValue = defaultValue
+        ) ?: run {
+            Logger.warn(methodTag, "Broker restrictions manager is not initialized.")
```
mohitc1, line 71: > Logger.warn(methodTag, "Broker restrictions manager is not initialized.") [](http://example.com/codeflow?start=11&length=74)

I think if getString returns null, you'd get into run block, which is not correct semantically

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequestHandler.kt
```diff
@@ -70,15 +114,16 @@ class CameraPermissionRequestHandler {
...
      * A request is considered valid if:
      * - The request is for the camera resource only.
      * - The request is not a repeated request (i.e., not for the same origin and resources).
-     *
+     * We can only grant or deny permissions for video capture/camera.
+     * To avoid unintentionally granting requests for not defined permissions
      */
-    fun setIfValid(request: PermissionRequest): Boolean {
+    private fun isValid(request: PermissionRequest): Boolean {
         val methodTag = "$TAG:setIfValid"
```
mohitc1: >setIfValid [](http://example.com/codeflow?start=30&length=10)

nit: fix tags,, will help with debugging

mohitc1 (approved):
Left some suggestions

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
p3dr0rv: Yes, you are right correcting this.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequestHandler.kt
p3dr0rv: checking tags

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * This method checks if the device preferred authentication config is set.
+     */
+    fun getPreferredAuthConfig(): String? {
+        val methodTag = "$TAG:getPreferredAuthConfig"
+        val defaultValue = null
+        restrictionsManager?.let { manager ->
+            return manager.getString(
+                key = PREFERRED_AUTH_CONFIG,
+                brokerAppPackageName = BrokerData.prodMicrosoftAuthenticator.packageName,
```
somalaya: brokerAppPackageName is authenticator in this case?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CameraPermissionRequestHandler.kt
```diff
@@ -97,19 +142,92 @@ class CameraPermissionRequestHandler {
...
      */
-    private fun setCurrentPermissionRequest(permissionRequest: PermissionRequest) {
-        currentPermissionRequest = permissionRequest
-        isGranted = false
+    private fun showQrPinCameraRationale(context: Context) {
+        val methodTag = "$TAG:showQrPinCameraRationale"
+        val builder = AlertDialog.Builder(context)
+        builder.setMessage(R.string.qr_code_rationale_message)
+            .setTitle(R.string.qr_code_rationale_header)
+            .setCancelable(false)
```
somalaya, line 221: Why is cancelable set to false?

p3dr0rv: In the showQrPinCameraRationale method, the setCancelable(false) is used to ensure that the dialog cannot be dismissed by tapping outside the dialog or pressing the back button. This is likely done to make sure the user explicitly chooses one of the provided options (allow or block) for granting or denying the camera permission.

common/src/main/java/com/microsoft/identity/common/internal/broker/SdmQrPinManager.kt
p3dr0rv: yes, improving doc

somalaya (approved):

PR #2631, opened 2025-04-28
DUNA. Prevent phish attack in “switch_browser” flow. , Fixes AB#3250169

p3dr0rv:
DOC: https://microsoft-my.sharepoint-df.com/:w:/r/personal/sedemche_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B29055A71-DD9A-4E95-9E2D-6DBFB77F7823%7D&file=DUNA-switch_browser-phish%20attack.docx&action=default&mobileredirect=true&share=IQFxWgUpmt2VTp4tbb-3f3gjAb1Z62agYgf-0Tl4zsGqDBE

The document highlights a security vulnerability where malicious applications can exploit the "switch_browser" and "switch_browser_resume" endpoints, leading to phishing attacks through mimicked UI pages. The proposed mitigations include adding a state parameter to prevent such attacks, ensuring that both actions occur within a valid authentication flow. 

This state parameter allows the client application to validate the actions, by comparing the state in the in the authorization request against the state in the response (broker redirect).
image

[AB#3250169](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3250169)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -560,6 +560,10 @@ public final class AuthenticationConstants {
          */
         public static final String ACTION_URI = "action_uri";
 
+        /**
+         * String Query parameter key for the state blob.
+         */
+        public static final String STATE = "state";
     }
 
     /**
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
@@ -25,6 +25,8 @@ package com.microsoft.identity.common.internal.ui.webview.switchbrowser
 import android.net.Uri
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants.SWITCH_BROWSER
 import com.microsoft.identity.common.java.exception.ClientException
+import com.microsoft.identity.common.java.flighting.CommonFlight
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager
 import com.microsoft.identity.common.java.opentelemetry.SpanExtension
 import com.microsoft.identity.common.logging.Logger
 import io.opentelemetry.api.trace.StatusCode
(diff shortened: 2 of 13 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -560,6 +560,10 @@ public static final class SWITCH_BROWSER {
          */
         public static final String ACTION_URI = "action_uri";
 
+        /**
+         * String Query parameter key for the state blob.
+         */
+        public static final String STATE = "state";
```
shahzaibj, line 566: Did we not already have a constant for this?

shahzaibj: Ok, never mind. You are defining this under SwitchBrowserConstans. that's fine

shahzaibj (commented):
Is it possible to put this change behind a separate flight? I'm thinking just in case client changes for state go to prod before server changes then it would be good to have the ability to turn off just this flight and test the rest of the switch browser protocol.

Not necessary to put the whole change behind a flight but just the parts of code that validate the state and/or take a hard dependency on state param being present. Essentially turning off the flight should allow the flow to work (no exception thrown) even if state param is not returned by server. So client code can essentially proceed with assuming an empty string state param.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
```diff
@@ -69,51 +79,53 @@ object SwitchBrowserUriHelper {
...
+        if (STATE_VALIDATION_REQUIRED) {
+            if (state.isNullOrEmpty()) {
+                // This should never happen, but if it does, we should log it and throw.
+                val errorMessage = "State is null or empty"
+                val exception = ClientException(ClientException.MISSING_PARAMETER, errorMessage)
+                Logger.error(methodTag, errorMessage, exception)
+                throw exception
+            } else {
+                queryParams[SWITCH_BROWSER.STATE] = state
+            }
```
shahzaibj: I see similar code above, perhaps it can be refactored into its own method

shahzaibj (approved):

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
p3dr0rv: [done](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2631/commits/f84da49bc1a22a471ef29286d0acc8643a079853)

shahzaibj (approved):

PR #2632, opened 2025-04-28
Add flight to control AM API WorkProfile, Fixes AB#3138815

fadidurah:
Add flight to control AM API WorkProfile

[AB#3138815](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3138815)


common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -98,7 +98,13 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable the attachment of PRT header in cross cloud requests. Default is true.
      */
-    ENABLE_ATTACH_PRT_HEADER_WHEN_CROSS_CLOUD("EnableAttachPrtHeaderWhenCrossCloud", true);
+    ENABLE_ATTACH_PRT_HEADER_WHEN_CROSS_CLOUD("EnableAttachPrtHeaderWhenCrossCloud", true),
+
+    /**
+     * Flight to enable adding x-client-MN and x-client-WPAvailable extra query parameters
+     */
+    ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS("EnableAmApiWorkProfileExtraQueryParameters", true);
```
somalaya: Is the default value for this flight intended to be true?

somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
fadidurah: Yes, we want the parameters to be sent by default, we just have the flight in order to turn it off if needed

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/AuthorizationRequest.java
```diff
@@ -308,7 +312,15 @@ public String toString() {
     public URI getAuthorizationRequestAsHttpRequest() throws ClientException {
         try {
             final CommonURIBuilder builder = new CommonURIBuilder(getAuthorizationEndpoint());
-            builder.addParametersIfAbsent(ObjectMapper.serializeObjectHashMap(this));
+            final Map<String, Object> serializedMap = ObjectMapper.serializeObjectHashMap(this);
+
+            // If the flight is not enabled, do not include these fields in the query parameter map
+            if (!CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS)) {
+                serializedMap.remove(MicrosoftAuthorizationRequest.WP_AVAILABLE_EXTRA_PARAMETER_NAME);
+                serializedMap.remove(Device.PlatformIdParameters.MANUFACTURER);
```
shahzaibj: It seems like the fields were added and we're removing them? If the flight is disabled then why we added the fields in the first place?

fadidurah: The fields are marked as exposed in `MicrosoftAuthorizationRequest.java`, so that they will be picked up by the `ObjectMapper.serializeObjectHashMap(this)` call in this method. They will always be included in the map as they are fields in the class itself, so checking for flight here and removing them is our way to make sure they don't get attached to the request if the flight is false. To my knowledge there isn't a straightforward way to conditionally assign an annotation. This is the next best thing in my opinion.

fadidurah: I reworked it to not annotate them as exposed, but i had to mark them transient to avoid serialization. This seems to re-order them at the end of the query string, but i think this is fine.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -178,6 +184,29 @@ protected MicrosoftAuthorizationRequest(@SuppressWarnings(WarningType.rawtype_wa
...
 
+    @Override
+    public URI getAuthorizationRequestAsHttpRequest() throws ClientException {
+        final URI superRequestUri = super.getAuthorizationRequestAsHttpRequest();
+
+        // If the flight is enabled, add fields for AM API Work Profile
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS)) {
+
+            final CommonURIBuilder builder = new CommonURIBuilder(superRequestUri);
+            builder.addParameterIfAbsent(Device.PlatformIdParameters.MANUFACTURER, mDiagnosticMN);
```
shahzaibj: Where is `mDiagnosticMN` initialized?

shahzaibj: Can we just not initialize that if flight is OFF?

fadidurah: `mDiagnosticMN = Device.getManufacturer()`, it takes a value from the Device Static instance, which pulls it from device metadata.

I don't really see a benefit to not initializing this field if the flight is off. We aren't doing any additional computation, just fetching a static variable that's already set.

shahzaibj: The idea is that right now you are using the flight in a hacky way where you are setting these params on AuthorizationRequest regardless of flight status but then only refraining from adding it to the query params if the flight is OFF but I think the better approach would be to not touch the query params and rather check flight before creating and setting these on AuthorizationRequest object

shahzaibj: That way you don't have to mark them as transient

shahzaibj (approved):
Discussed offline. Not a fan of the pattern used for flighting and setting qp conditionally here. We have better patterns in the code today. That said not blocking on this but we should consolidate in the future.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
fadidurah: Had an offline discussion, will switch to using the same approach as switch browser field

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -141,10 +149,10 @@ public abstract class MicrosoftAuthorizationRequest<T extends MicrosoftAuthoriza
     @SerializedName("pc")
     private final String mPreferredAuthMethodCode;
 
-    @Expose()
+    // Transient to avoid adding this automatically, leave control to flight
```
shahzaibj: Remove legacy comment

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -174,8 +182,16 @@ protected MicrosoftAuthorizationRequest(@SuppressWarnings(WarningType.rawtype_wa
...
+
+
+        // If the flight is enabled, set the fields
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS)) {
+            mDiagnosticMN = Device.getManufacturer();
+            mWorkProfileAvailable = Device.isInPersonalProfileButClouddpcWorkProfileAvailable();
+        } else {
+            mDiagnosticMN = null;
+            mWorkProfileAvailable = null;
+        }
```
shahzaibj, line 193: Do we need the else block? The default is null anyway, right?

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
fadidurah: If they are final variables, then the else block is needed, getting an error about them not being initialized

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
fadidurah: oops, missed that

shahzaibj (approved):

PR #2633, opened 2025-04-29
Allow generating wrapping keys without PURPOSE_WRAP_KEY with Flight, Fixes AB#3260422

mohitc1:
Fixes [AB#3260422](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3260422)
This for incident where we keystore unwrap operation fails with invalid key exception 
```
java.security.InvalidKeyException: Failed to unwrap key
               at android.security.keystore2.AndroidKeyStoreCipherSpiBase.engineUnwrap(AndroidKeyStoreCipherSpiBase.java:807)
               at javax.crypto.Cipher.unwrap(Cipher.java:2459)
….
Caused by: javax.crypto.IllegalBlockSizeException
               at android.security.keystore2.AndroidKeyStoreCipherSpiBase.engineDoFinal(AndroidKeyStoreCipherSpiBase.java:628)
               at android.security.keystore2.AndroidKeyStoreCipherSpiBase.engineUnwrap(AndroidKeyStoreCipherSpiBase.java:805)
               ... 32 more
Caused by: android.security.KeyStoreException: -22 (internal Keystore code: -22 message: system/security/keystore2/src/operation.rs:836: KeystoreOperation::update

Caused by:
    0: system/security/keystore2/src/operation.rs:390: Update failed.
    1: Error::Km(r#KEY_EXPORT_OPTIONS_INVALID)) (public error code: 13 internal Keystore code: -22)
               at android.security.KeyStore2.getKeyStoreException(KeyStore2.java:386)
               at android.security.KeyStoreOperation.handleExceptions(KeyStoreOperation.java:78)
               at android.security.KeyStoreOperation.update(KeyStoreOperation.java:115)
               at android.security.keystore2.KeyStoreCryptoOperationChunkedStreamer$MainDataStream.update(KeyStoreCryptoOperationChunkedStreamer.java:222)
               at android.security.keystore2.KeyStoreCryptoOperationChunkedStreamer.update(KeyStoreCryptoOperationChunkedStreamer.java:156)
               at android.security.keystore2.KeyStoreCryptoOperationChunkedStreamer.doFinal(KeyStoreCryptoOperationChunkedStreamer.java:179)
               at android.security.keystore2.AndroidKeyStoreCipherSpiBase.engineDoFinal(AndroidKeyStoreCipherSpiBase.java:618)
```

The root cause is not yet known and the issue is only impacting certain make + model.

We had previously seen this issue with pixel 5 devices and had add fix. While the added code addressed it for some devices, the issue still is seen this time on different devices primarily moto g85.

Link PRs here.

In new key gen spec we are using PURPOSE_WRAP_KEY, which was introduced mainly for securely importing secret keys into keystore. Our logic of secret is different where we generate secret key in code and then call Cipher.wrap(and Cipher.unwrap) for wrapping/unwrapping. The implementation with PURPOSE_WRAP_KEY works fine in this case as well for most cases.

A possibility is that on some devices this purpose is not working when usage is not involving secure key import. Address this adding flighted change in ths PR to generate key spec without PURPOSE_WRAP_KEY (with just PURPOSE_ENCRYPT and PURPOSE_DECRYPT).
In prod we can enable this path for the customer and see if this works.

Changes
1. Added change to only use PURPOSE_ENCRYPT and PURPOSE_DECRYPT based on flight (ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITHOUT_PURPOSE_WRAP_KEY) - flight 2
2. Previous flight for using PURPOSE_WRAP_KEY renamed to ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITH_PURPOSE_WRAP_KEY - flight 1
3. Both enabled by default.  This means on On API >= 28 by default, PURPOSE_WRAP_KEY will be used. If we turn OFF flight 1, then flight 2 will be used. If flight 2 is enabled, use new key gen spec without PURPOSE_WRAP_KEY. If we turn OFF flight 2 as well, legacy spec will be used. On API level >= 23 (and < 28) only flight 2 based logic would execute, as PURPOSE_WRAP_KEY is available on > = 28
5. Refactored some code.


common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -103,7 +104,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to make the state parameter required for the switch browser protocol.
      */
-    SWITCH_BROWSER_PROTOCOL_REQUIRES_STATE("SwitchBrowserProtocolRequiresState", false);
+    SWITCH_BROWSER_PROTOCOL_REQUIRES_STATE("SwitchBrowserProtocolRequiresState", false),
+
+    /** Flight to enable the new key generation without PURPOSE_WRAP_KEY. Default is true.
+    * This is applicable for API >= 23
+    */
+    ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITHOUT_PURPOSE_WRAP_KEY("EnableNewKeyGenSpecForWrapWithoutPurposeWrapKey", true);
```
somalaya, line 117: We want this to be true by default?

mohitc1: yes. This way we don't use deprecated key spec.

Do you prefer it to be OFF?

fadidurah (approved):

somalaya (approved):

PR #2635, opened 2025-05-01
Cherrypick, change boolean to Boolean, Fixes AB#3261502

fadidurah:
Made a change to support work profile, but a refactor introduced possibility of null values for work profile is available field, which is causing exceptions in automation as boolean type cannot handle null value. Changing to Boolean.

[AB#3261502](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3261502)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
@@ -109,7 +109,7 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable adding x-client-MN and x-client-WPAvailable extra query parameters
      */
-    ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS("EnableAmApiWorkProfileExtraQueryParameters", false),
+    ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS("EnableAmApiWorkProfileExtraQueryParameters", true),
 
     /** Flight to enable the new key generation without PURPOSE_WRAP_KEY. Default is true.
      * This is applicable for API >= 23
--- a/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
@@ -79,7 +79,7 @@ public class Device {
     }
 
     @GuardedBy("sLock")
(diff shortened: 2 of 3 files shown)


common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -109,7 +109,7 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable adding x-client-MN and x-client-WPAvailable extra query parameters
      */
-    ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS("EnableAmApiWorkProfileExtraQueryParameters", false),
+    ENABLE_AM_API_WORKPROFILE_EXTRA_QUERY_PARAMETERS("EnableAmApiWorkProfileExtraQueryParameters", true),
```
mohitc1, line 112: >true [](http://example.com/codeflow?start=99&length=4)

why this change?

fadidurah: it should be enabled by default. Idea is we will make a PR in ecs to turn this off if we run into any issues in prod

common4j/src/main/com/microsoft/identity/common/java/platform/Device.java
```diff
@@ -89,7 +89,7 @@ public static void setIsInPersonalProfileButClouddpcWorkProfileAvailable(final b
     }
 
     @GuardedBy("sLock")
-    public static boolean isInPersonalProfileButClouddpcWorkProfileAvailable() {
+    public static Boolean isInPersonalProfileButClouddpcWorkProfileAvailable() {
```
mohitc1, line 92: >isInPersonalProfileButClouddpcWorkProfileAvailable [](http://example.com/codeflow?start=26&length=50)

who's caller? Can it handle null?

fadidurah: Caller can handle null, issue was that method return signature of `isInPersonalProfileButClouddpcWorkProfileAvailable` used `boolean` type, which cannot be `null`

fadidurah: Caller is in `MicrosoftAuthorizationRequest` btw

mohitc1 (approved):
:shipit:

p3dr0rv (approved):

PR #2640, opened 2025-05-12
Resource Account API only, Fixes AB#3230428

mohitc1:
Fixes [AB#3230428](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3230428)

Adding method for provisioning resource account in Android Broker. The method is exposed to OneAuth via BrokerMsalController

For details Refer: https://github.com/AzureAD/ad-accounts-for-android/pull/3093


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -94,21 +94,18 @@ private static final class SerializedNames {
...
      * The redirect uri for the request.
      */
-    @NonNull
     @SerializedName(SerializedNames.REDIRECT)
     private String mRedirect;
 
     /**
      * The client id of the application.
      */
-    @NonNull
```
shahzaibj: Why removing these NonNull annotations?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/ResourceAccountCommandParameters.java
```diff
@@ -0,0 +1,43 @@
...
+package com.microsoft.identity.common.java.commands.parameters;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.experimental.SuperBuilder;
+
+/**
+ * Command parameters for setting up Resource Account in broker. This object is part of API.
+ * The caller constructs this and pass to BrokerMSALlController.
```
shahzaibj:
```suggestion
 * The caller constructs this and pass to BrokerMsalController.
```

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -905,6 +905,26 @@ public List<ICacheRecord> getAccountsFromResultBundle(@NonNull final Bundle bund
         return JsonExtensions.getICacheRecordListFromJsonString(accountJson);
     }
 
+    /**
+     * Get resource account record from the result bundle. If successful, new account
+     * record part of ICachedRecord is returned.
+     * @param bundle The result bundle from the broker.
+     * @throws BaseException
+     */
+    public ICacheRecord resourceAccountRecordFromBundle(@NonNull final Bundle bundle) throws BaseException {
```
shahzaibj, line 914: Do we need this method? Could the caller just call the `getAccountsFromResultBundle`?

mohitc1: It wraps additional logic. I am checking if 0 or more than 1 accounts are returned.

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
mohitc1: BrokerRequest can then be used for operations other than AT/ATS which mandate this and can represent generic broker request blob (serializable) to broker. On Broker side AT/ATS operations do validate parameters they receive.

Let me know if you have different opinion, I can consider creating separate request object for this request.

logdog16 (approved):

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -264,6 +265,50 @@ public Bundle getRequestBundleForAcquireTokenInteractive(@NonNull final Interact
...
+
+        Logger.info(methodTag, "Constructing result bundle from ProvisionResourceAccount.");
+        final String extraOptions = parameters.getExtraOptions() != null ?
+                QueryParamsAdapter._toJson(parameters.getExtraOptions()) : null;
+
+        final BrokerRequest brokerRequest = BrokerRequest.builder()
+                .authority(parameters.getAuthority().getAuthorityURL().toString())
+                .extraOptions(extraOptions)
+                .homeAccountId(parameters.getHomeAccountId())
+                .userName(parameters.getLoginHint())
```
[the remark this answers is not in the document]
mohitc1: I'll keep it as optional

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
shahzaibj: Yes, but the way I see it the method is too generic (which is nice) and essentially is a wrapper to get just the first account record from cache. IIRC, we are doing something similar for SingleAccountMode PCA. At some point we probably should consolidate into 1 generic method.

shahzaibj: Also wondering if rather than returning the bundle and then letting MSAL validate it if Broker itself should be throwing in the first place?

shahzaibj (approved):
Responded to few comments, else lgtm

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
mohitc1: Yes, there will be check on broker side.. I don't have implementation of the broker operation right now, but it will ensure to return on the one account object that it just created.
Correspondingly the MSAL side impl may also change. the method signature `public ICacheRecord resourceAccountRecordFromBundle(@NonNull final Bundle bundle)` remains same

p3dr0rv (approved):

PR #2297, opened 2024-01-25
Add InvalidError for NativeAuth

SaurabhMSFT:
Corresponding PR in MSAL Repo : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2020

PR #2650, opened 2025-05-26
Enforce proxy in UrlConnectionHttpClient 

RodargDev:
This PR should resolve this issue where the proxy currently isn't enforced [#2297](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/2297)

The code is currently untested, im using it as a reference to see if you guys would accept a solution like this if it were to be tested.


RodargDev:
> @RodargDev please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.

RodargDev:
@nilo-ms or @p3dr0rv any chance you or anyone from your team can have a look at this because currently I don't really know what I can do to get this looked at.

RodargDev:
@p3dr0rv any chance you can have a look at these changes?

RodargDev:
@bgavrilMS Any chance you can have a look at this?

mohitc1:
Is proxy set on device? How? Usually, when proxy is setup by going to settings > connections > edit > update proxy, the default HttpURLConnection would go through this. Trying to understand, why do you have to explicitly supply proxy?

RodargDev:
> Is proxy set on device? How? Usually, when proxy is setup by going to settings > connections > edit > update proxy, the default HttpURLConnection would go through this. Trying to understand, why do you have to explicitly supply proxy?

We are developing an app where we connect to a hotspot with [android Wi-Fi bootstrap API](https://developer.android.com/develop/connectivity/wifi/wifi-bootstrap). 

The documentation notes: ```"Note: Creating a connection using this API does not provide an internet connection to the app or to the device. To provide an internet connection to the apps on a device, use the [Wi-Fi Suggestion API](https://developer.android.com/develop/connectivity/wifi/wifi-suggest) instead."```

Due to our app’s requirements, we cannot use the Wi-Fi Suggestion API, as it only adds network credentials without ensuring a direct connection to the device. Therefore, we rely on the P2P connection provided by the Bootstrap API. However, this API restricts internet access on the connected network, which prevents MSAL from refreshing access tokens while connected.

We have found a way to work around this limitation by starting a local proxy within the app whenever we connect to a Wi-Fi endpoint via the Bootstrap API. This proxy enables the app to access the internet even though the Bootstrap API normally restricts network connectivity. To ensure that all traffic including MSAL requests goes through this proxy, we configure it using [JVM system properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html#Proxies) as follows:

````kotlin
System.setProperty("https.proxyHost", "127.0.0.1")
System.setProperty("https.proxyPort", proxyPort.toString());

System.setProperty("http.proxyHost", "127.0.0.1")
System.setProperty("http.proxyPort", proxyPort.toString());
````

We expected the HTTP client used by MSAL to automatically respect these system properties, since the default behavior of HttpURLConnection is to do so. However, it appears that MSAL uses a HttpURLConnectionFactory that does not respect these proxy settings, even though this is the default behavior in the standard implementation.

Manually setting a proxy through device settings, as you described, would normally work. However, this approach would require our users to configure a proxy themselves and exit the app, which is not practical and not a solution for the given requirement. We expected MSAL for Android to implement this behavior automatically, similar to other libraries like .NET Xamarin MSAL. Unfortunately, it seems there is currently no way to configure a proxy for MSAL requests because the custom HttpURLConnection does not implement it.

PR #2658, opened 2025-05-29
Add capabilities parameter to the flow start operations, Fixes AB#3288182, Closed AB#3288182

Yuki-YuXin:
Fixes [AB#3288182](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3288182)

1. __BaseNativeAuthCommandParameters.java__ - Added `capabilities` field to the base command parameters class
2. __SignInInitiateRequest.kt__ - Added `capabilities` parameter to the sign-in initiate request
3. __SignUpStartRequest.kt__ - Added `capabilities` parameter to the sign-up start request
4. __ResetPasswordStartRequest.kt__ - Added `capabilities` parameter to the reset password start request
5. __NativeAuthRequestProvider.kt__ - Updated to pass capabilities from configuration to requests
6. __NativeAuthOAuth2Configuration.kt__ - Added `capabilities` property to the configuration class

pipeline failed due to the broker step.

### DIFF
--- a/common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProviderTest.kt
+++ b/common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProviderTest.kt
@@ -178,6 +178,24 @@ class NativeAuthRequestProviderTest {
     }
 
     @Test
+    fun testSignUpStartExpectedCapabilitiesStringAddedToRequestBody() {
+        every { mockConfig.capabilities } returns capabilities
+
+        val commandParameters = SignUpStartCommandParameters.builder()
+            .platformComponents(mock<PlatformComponents>())
+            .username(username)
+            .clientId(clientId)
+            .correlationId(correlationId)
+            .build()
+
+        val result = nativeAuthRequestProvider.createSignUpStartRequest(
+            commandParameters = commandParameters
+        )
+
+        assertEquals(capabilities, result.parameters.capabilities)
+    }
+
+    @Test
     fun testSignUpStartWithUnsetCorrelationIdShouldNotHaveHeader() {
         val commandParameters = SignUpStartCommandParameters.builder()
             .platformComponents(mock<PlatformComponents>())
@@ -517,9 +535,28 @@ class NativeAuthRequestProviderTest {
             .correlationId(correlationId)
             .build()
 
-        nativeAuthRequestProvider.createSignInInitiateRequest(
+        val result = nativeAuthRequestProvider.createSignInInitiateRequest(
             commandParameters = commandParameters
         )
+
+        assertEquals(capabilities, result.parameters.capabilities)
+    }
+
+    @Test
+    fun testSignInInitiateExpectedCapabilitiesStringAddedToRequestBody() {
+        every { mockConfig.capabilities } returns capabilities
+
+        val commandParameters = SignInStartCommandParameters.builder()
+            .platformComponents(mock<PlatformComponents>())
+            .username(username)
+            .correlationId(correlationId)
+            .build()
+
+        val result = nativeAuthRequestProvider.createSignInInitiateRequest(
+            commandParameters = commandParameters
+        )
+
+        assertEquals(capabilities, result.parameters.capabilities)
     }
 
     @Test
@@ -1032,6 +1069,23 @@ class NativeAuthRequestProviderTest {
         nativeAuthRequestProvider.createResetPasswordStartRequest(
             commandParameters = commandParameters
         )
+    }
+
+    @Test
+    fun testResetPasswordStartExpectedCapabilitiesStringAddedToRequestBody() {
+        every { mockConfig.capabilities } returns capabilities
+
+        val commandParameters = ResetPasswordStartCommandParameters.builder()
+            .platformComponents(mock<PlatformComponents>())
+            .username(username)
+            .correlationId(correlationId)
+            .build()
+
+        val result = nativeAuthRequestProvider.createResetPasswordStartRequest(
+            commandParameters = commandParameters
+        )
+
+        assertEquals(capabilities, result.parameters.capabilities)
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


spetrescu84 (approved):

common4j/src/test/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProviderTest.kt
```diff
@@ -159,6 +161,22 @@ class NativeAuthRequestProviderTest {
         )
     }
 
+    @Test
```
nilo-ms, line 165: I'm not sure you can add those tests here, but we should also have the following tests:
- capabilities is only added to the request body when is not empty or null
- the expected capabilities string is added to the request body.

Yuki-YuXin: _capabilities is only added to the request body when is not empty or null_
IOS and Android don't seem to be doing that. 

Currently, inside [NativeAuthRequestProvider.kt](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2658/files#diff-adc0a615e32c02a25a7bc22ad62d55a52219b3cc3cce6976ddd57b3454c0c2af)
`capabilities = config.capabilities` without checking if it's null or empty.  The request body will have a parameter `capabilities = null`

Similar in IOS?
```
func makeRequestBody(config: MSALNativeAuthInternalConfiguration) -> [String: String] {
        typealias Key = MSALNativeAuthRequestParametersKey

        return [
            Key.clientId.rawValue: config.clientId,
            Key.username.rawValue: username,
            Key.challengeType.rawValue: config.challengeTypesString,
            Key.capabilities.rawValue: config.capabilitiesString
        ].compactMapValues { $0 }
```

Refer:
```
func test_capabilities_shouldCreateCorrectBodyRequest() throws {
        XCTAssertNoThrow(config = try .init(clientId: DEFAULT_TEST_CLIENT_ID, authority: MSALCIAMAuthority(url: baseUrl), challengeTypes: [], capabilities: [.registrationRequired], redirectUri: nil))
        let params = MSALNativeAuthResetPasswordStartRequestParameters(
            context: MSALNativeAuthRequestContextMock(),
            username: DEFAULT_TEST_ID_TOKEN_USERNAME
        )

        let body = params.makeRequestBody(config: config)

        let expectedBodyParams = [
            "client_id": DEFAULT_TEST_CLIENT_ID,
            "username": DEFAULT_TEST_ID_TOKEN_USERNAME,
            "challenge_type": "redirect",
            "capabilities": "registration_required"
        ]

        XCTAssertEqual(body, expectedBodyParams)
```

Added tests for _the expected capabilities string is added to the request body._

nilo-ms: 1. 

> IOS and Android don't seem to be doing that.

That is not entireraly true, we remove nil value on iOS. [Check here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2635/files#diff-09eb577fbfd6bbdc9f21935b4ed4083c4190995584483d16d35c764ede0e20ea) for example. This is what ["compactMapValues"](https://developer.apple.com/documentation/swift/dictionary/compactmapvalues(_:)) does. We also have tests to check that nil value for "capabilities" is not added to the request body. Check the "[MSALNativeAuthRequestConfiguratorTests.swift](https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/2635/files#diff-143e7d7cc82a17e991db5ba99c4efab850826c7f6c0db8055c2798920b1ae2a4)", for each test, we verify that the body matches the expected one.
We should do the same for Android too

Yuki-YuXin: After investigation, I confirmed that the parameter capability = null will be filtered out in the request body. Please see the image below.
image

And here is the comment of the method `ObjectMapper.serializeObjectToFormUrlEncoded`

```
/**
     * Method for serializing the contents of an object as a Url Encoded string.  **Important to the implementation of
     * this method is the behavior of GSON which excludes null fields from the resulting JSON.**  A TreeMap was used to
     * simplify testing.... the resulting url encoded string is in alphabetical order (keys).
     *
     * @param object Object
     * @return String
     * @throws UnsupportedEncodingException thrown if encoding not supported
     */
    public static String serializeObjectToFormUrlEncoded(Object object) throws UnsupportedEncodingException
```

Yuki-YuXin: The parameter itself keep the null parameter but the request content doesn't.
image

nilo-ms: Thanks Yuki, can you please a new unit test to test this scenario?

Yuki-YuXin: Do you have any thoughts on where unit testing fits in?  There are `ObjectMapperTest` and `NativeAuthRequestProviderTest`. The common pattern of `NativeAuthRequestProviderTest` is to test the request (requestUrl, headers, parameters)

```
@Test
    fun testSignUpStartSuccess() {
        val commandParameters = SignUpStartCommandParameters.builder()
            .platformComponents(mock<PlatformComponents>())
            .username(username)
            .clientId(clientId)
            .userAttributes(userAttributes)
            .correlationId(correlationId)
            .build()

        val result = nativeAuthRequestProvider.createSignUpStartRequest(
            commandParameters = commandParameters
        )

        assertEquals(username, result.parameters.username)
        assertEquals(clientId, result.parameters.clientId)
        assertEquals(challengeType, result.parameters.challengeType)
        assertEquals(ApiConstants.MockApi.signUpStartRequestUrl, result.requestUrl)
        assertEquals(userAttributes.toJsonString(userAttributes), result.parameters.attributes)
        assertEquals(result.headers[AuthenticationConstants.AAD.CLIENT_REQUEST_ID], correlationId)
    }
```

Yuki-YuXin: Added `testSignInInitiateWithNonNullCapabilitiesIncludedInSerializedRequestBody` and `testSignInInitiateWithNullCapabilitiesFiltersOutFromSerializedRequestBody` inside `NativeAuthRequestProviderTest`

nilo-ms (approved):

PR #2659, opened 2025-05-29
Improve wrapped key loader

mohitc1:
AndroidWrappedKeyLoader currently has code cache secret key in memory, but there are multiple instances of AndroidWrappedKeyLoader active (one per BrokerPlatformComponents). So, each object gets is own cache and there's no benefit of caching. Moreover, this makes keystore reads more frequent and concurrent. 

Changes:
Update caching logic in AndroidWrappedKeyLoader by static ConcurrentHashMap so cache is shared across all ConcurrentHashMap. This fixes caching (effective) logic and reduces keystore unwrap and decrypt calls. A good side effect of this would be lesser number of failures due to Keystore unwrap calls that we have seen. 

AndroidKeyStoreUtil changes:
Load keystore always instead of caching, as it is recommended to not cache it statically.


common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyProvider.java
```diff
@@ -173,18 +172,27 @@ public String getKeyTypeIdentifier() {
...
+        }
 
-        if (key == null) {
-            key = readSecretKeyFromStorage();
+        SecretKey key = sKeyCacheMap.get(mFilePath);
+        if (key != null) {
+            return key;
         }
 
+        Logger.info(methodTag, "Key not in cache is empty, loading key from storage");
```
somalaya:
```suggestion
        Logger.info(methodTag, "Key not in cache or cache is empty, loading key from storage");
```

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyProvider.java
```diff
@@ -122,21 +127,15 @@ public class AndroidWrappedKeyProvider implements ISecretKeyProvider {
...
-        public SecretKey getData() {
-            if (!sSkipKeyInvalidationCheck &&
-                    (!AndroidKeyStoreUtil.canLoadKey(mAlias) || !getKeyFile().exists())) {
-                this.clear();
-            }
-            return super.getData();
-        }
-    };
+    // Exposed for testing only.
+    @Nullable
```
somalaya, line 132: Annotate with @VisibleForTesting?

somalaya (approved):

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -87,20 +87,16 @@ public class AndroidKeyStoreUtil {
...
     );
 
     private static synchronized KeyStore getKeyStore()
             throws KeyStoreException, CertificateException, NoSuchAlgorithmException, IOException {
-        if (mKeyStore == null){
-            mKeyStore = KeyStore.getInstance(ANDROID_KEY_STORE_TYPE);
-            mKeyStore.load(null);
-        }
-        return mKeyStore;
+        final KeyStore keyStore = KeyStore.getInstance(ANDROID_KEY_STORE_TYPE);
```
fadidurah, line 97: Is this a lazy init? If not, i think we should make it so (i'm still thinking about OOM errors lol)

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
mohitc1: No, it's new local object creation that's returned from the method. Just a util function

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyProvider.java
```diff
@@ -122,21 +128,17 @@ public class AndroidWrappedKeyProvider implements ISecretKeyProvider {
...
+    @VisibleForTesting
+    /* package */ SecretKey getKeyFromCache() {
```
p3dr0rv, line 134: @mohitc1, we need to run the validation here, otherwise it will fail the testLoadDeletedKeyStoreKey and testLoadDeletedKeyFile, these tests did not appear because the tests instrumented in PR are failing :(


validation:
        if (!sSkipKeyInvalidationCheck &&
                (!AndroidKeyStoreUtil.canLoadKey(mAlias) || !this.getKeyFile().exists())) {
            sKeyCacheMap.remove(mFilePath);
        }

//nit: we can put it inside a function

mohitc1: Thanks for catching that. Update the code. I will run it once locally.

mohitc1: validated and updated the test code

p3dr0rv (approved):

PR #2665, opened 2025-06-04
Fixing the sign in screens when edge to edge is enabled, Fixes AB#3293334

somalaya:
The Auth password page on LTW is edge-to-edge by default on devices running Android 15 if the app is targeting Android 15 (API level 35). Please refer to [this](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) for more details. Because of this feature enabled by default, I am seeing below cropped screen.

Attached screenshot below.
EdgeToEdge

Steps to reproduce:
1. set targetSdkVersion to 35 in BrokerHost app

2. launch MsalTestApp

3. click on Sign in button

4. See that the page is cropped on the top
-------------------------------------------------------------
Expected Behavior:
Should not be cropped

Fix : Adding padding in the activity showing the UI screens by getting the insets of the system bar as suggested in the  edge to edge documentation.

This is after fixing the issue
image

Fixes [AB#3293334](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3293334)

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -161,7 +161,7 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "com.squareup.moshi:moshi:$rootProject.ext.moshiVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
-    implementation "androidx.core:core:1.5.0"
+    implementation "androidx.core:core:$rootProject.ext.androidxCoreVersion"
     implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayoutVersion"
     //For Executive Order work
     implementation "com.yubico.yubikit:android:$rootProject.ext.yubikitAndroidVersion"
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -161,6 +161,7 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "com.squareup.moshi:moshi:$rootProject.ext.moshiVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
+    implementation "androidx.core:core:1.5.0"
     implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayoutVersion"
     //For Executive Order work
     implementation "com.yubico.yubikit:android:$rootProject.ext.yubikitAndroidVersion"
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
@@ -63,15 +63,19 @@ public class DualScreenActivity extends FragmentActivity {
     private void initializeContentView(){
         super.setContentView(R.layout.dual_screen_layout);
         if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_HANDLING_FOR_EDGE_TO_EDGE)) {
-            ViewCompat.setOnApplyWindowInsetsListener(findViewById(android.R.id.content), (view, insets) -> {
-                int topInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).bottom;
-                int bottomInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).bottom;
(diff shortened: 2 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
```diff
@@ -58,6 +62,17 @@ public void setContentView(int layoutResID) {
 
     private void initializeContentView(){
         super.setContentView(R.layout.dual_screen_layout);
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_HANDLING_FOR_EDGE_TO_EDGE)) {
+            ViewCompat.setOnApplyWindowInsetsListener(findViewById(android.R.id.content), (view, insets) -> {
+                int topInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).bottom;
```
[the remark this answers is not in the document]
melissaahn: If this is on purpose (topInset is set to bottom value), probably good to put a comment?

melissaahn (approved):

common/build.gradle
```diff
@@ -161,6 +161,7 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "com.squareup.moshi:moshi:$rootProject.ext.moshiVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
+    implementation "androidx.core:core:1.5.0"
```
rpdome: nit: put this in the version file.

Also check with oneauth and broker apps if they're on a different version of this library.

rpdome (approved):

common/build.gradle
somalaya: Confirmed with all the teams, no risk with this one.

rpdome (approved):

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
```diff
@@ -58,6 +62,21 @@ public void setContentView(int layoutResID) {
 
     private void initializeContentView(){
         super.setContentView(R.layout.dual_screen_layout);
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_HANDLING_FOR_EDGE_TO_EDGE)) {
+            try {
+                ViewCompat.setOnApplyWindowInsetsListener(findViewById(android.R.id.content), (view, insets) -> {
+                    int topInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).top;
+                    int bottomInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).bottom;
+                    int leftInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).left;
+                    int rightInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).right;
```
piotrskibapl, line 71: I think that `displayCutout()` should be added here in addition to `systemBars()`. Otherwise the activity content might be obstructed by the cutouts, e.g. when the device is rotated or there's a cutout outside of the toolbar area.

somalaya: int topInset = insets.getInsets(WindowInsetsCompat.Type.systemBars()).top;
This gives you the top inset, which may include:

The status bar height

Any display cutout (notch) at the top

So even though you're not calling getDisplayCutout() directly, systemBars() includes cutouts as well. This is by design — WindowInsetsCompat.Type.systemBars() is a convenient alias for:

- Status bar
- Navigation bar
- Cutout (if applicable)

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
piotrskibapl: While a cutout might be a part of the system bar, will it also work when the device is rotated, so that the cutout is on the left/right side of the device?

I can see that the [documentation](https://developer.android.com/develop/ui/views/layout/edge-to-edge?utm_source=chatgpt.com#cutout-insets) in its example uses both `systemBars()` and `displayCutout()`. I worked on that topic earlier today in my app, and I recall that the activity content was displayed under the camera cutout when systemBars was used only.

PR #2666, opened 2025-06-12
Replace AbstractSecretKeyLoader with ISecretKeyProvider, Fixes AB#3297746

p3dr0rv:
In the next PR:

* We are replacing AbstractSecretKeyLoader with ISecretKeyProvider.
* We are also delegating the secret key generation to a new class to separate concerns and improve code readability.

All these changes lay the groundwork for the next PR, where the new implementation will be introduced and the feature flag will be used to switch between the two implementations.

[AB#3297746](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3297746)


common/src/androidTest/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoaderTest.java
```diff
@@ -113,7 +113,7 @@ public void testGenerateKey() throws ClientException {
         final AndroidWrappedKeyLoader keyLoader = new AndroidWrappedKeyLoader(MOCK_KEY_ALIAS, MOCK_KEY_FILE_PATH, context);
         final SecretKey secretKey = keyLoader.generateRandomKey();
 
-        Assert.assertEquals(AES256KeyLoader.AES_ALGORITHM, secretKey.getAlgorithm());
+        Assert.assertEquals(keyLoader.getSecretKeyGenerator().getKeyAlgorithm(), secretKey.getAlgorithm());
```
rpdome: the goal here is to compare with the actual value (constant), not from what we get from the keyloader object.

Perhaps revert all the changes in this class?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoaderFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Factory class for creating wrapped key loaders specific to the Android platform.
+ *
+ * This object is responsible for creating the appropriate implementation of [ISecretKeyLoader]
+ * based on feature flag. It abstracts away the details of which
+ * specific loader implementation should be used, allowing for runtime switching between
+ * different implementations without affecting client code.
+ */
+object AndroidWrappedKeyLoaderFactory {
```
rpdome: Rather than a keyloader factory, it looks like you need a factory for storage layer instead?

two ways to tackle this

1. the new storage layer can decrypt (read) with both old and new keyloader. it would only write with the new keyloader

2.  the storage re-encrypts its data with the new keyloader, and use the new keyloader to read/write.

rpdome: If we simply use the flight to switch here - it would lead to user's data loss (since they can't decrypt existing data)

rpdome: e.g. put the flight in AndroidBrokerEncryptionManagerWithKeyStoreWrappedKey's getKeyLoaderForEncryption() and getKeyLoaderForDecryption(), if you're going with the first route.

p3dr0rv: Yeah, I’m planning to address this in the next PR, where I’ll be using strategy 1, see -> 
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/db8781af8f7746c3df4a51ee78f0fde2d82a661a/common/src/main/java/com/microsoft/identity/common/crypto/AndroidKeyStoreRsaKekManager.kt#L111

Instead of handling it at the storage layer as you mentioned, I’m just plugging in a new keyloader. This new keyloader will be easier to update if we need to introduce a new algorithm, whether for the key itself or the key-encryption-key.

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoaderFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @param keyIdentifier A unique identifier for the key being loaded
+     * @param fileName The name of the file where the wrapped key is stored
+     * @param context The Android application context needed for file and security operations
+     * @return An implementation of [ISecretKeyLoader] that can load the specified wrapped key
+     */
+    fun createWrappedKeyLoader(
+        keyIdentifier: String,
+        fileName: String,
+        context: android.content.Context
```
mohitc1: >android.content.Context [](http://example.com/codeflow?start=17&length=23)

nit: Context

common4j/src/main/com/microsoft/identity/common/java/crypto/key/ISecretKeyGenerator.kt
```diff
@@ -0,0 +1,64 @@
...
+     * This is used when generating random keys to determine the key strength.
+     */
+    val keySize: Int
+
+    /**
+     * The algorithm name for the key.
+     * This should be compatible with cryptographic providers, such as
+     * those used with KeyGenerator.getInstance(algorithm).
+     */
+    val keyAlgorithm: String
```
mohitc1: do we need these?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -177,12 +185,11 @@ public synchronized SecretKey getKey() throws ClientException {
         return key;
     }
 
-    @Override
     @NonNull
     protected SecretKey generateRandomKey() throws ClientException {
```
mohitc1, line 192: >protected [](http://example.com/codeflow?start=4&length=9)

private?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoaderFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Factory class for creating wrapped key loaders specific to the Android platform.
+ *
+ * This object is responsible for creating the appropriate implementation of [ISecretKeyLoader]
+ * based on feature flag. It abstracts away the details of which
+ * specific loader implementation should be used, allowing for runtime switching between
+ * different implementations without affecting client code.
+ */
+object AndroidWrappedKeyLoaderFactory {
+    const val WRAPPED_KEY_KEY_IDENTIFIER: String = "A001"
```
mohitc1: >WRAPPED_KEY_KEY_IDENTIFIER [](http://example.com/codeflow?start=14&length=26)

do you need this here? (maybe for next pr)

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -124,7 +124,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable handling the UI in edge to edge mode
      */
-    ENABLE_HANDLING_FOR_EDGE_TO_EDGE("EnableHandlingEdgeToEdge", true);
+    ENABLE_HANDLING_FOR_EDGE_TO_EDGE("EnableHandlingEdgeToEdge", true),
+
+    /**
+     * Flight to enable the new Android wrapped key loader.
+     */
+    ENABLE_NEW_ANDROID_WRAPPED_KEY_LOADER("EnableNewAndroidWrappedKeyLoader", true);
```
mohitc1: Do you need this in this PR?

You could add this in next PR and factory class

common4j/src/main/com/microsoft/identity/common/java/crypto/key/ISecretKeyLoader.kt
```diff
@@ -0,0 +1,88 @@
...
+
+import com.microsoft.identity.common.java.exception.ClientException
+import javax.crypto.SecretKey
+
+/**
+ * Interface defining how a [SecretKey] is loaded, cached, sourced, and used.
+ *
+ * [ISecretKeyLoader] provides a consistent abstraction layer for cryptographic key operations
+ */
+interface ISecretKeyLoader {
```
mohitc1: >ISecretKeyLoader [](http://example.com/codeflow?start=10&length=16)

Given it's functionality, I think, a better name would be ISecretKeyProvider.

your call.

mohitc1 (approved):
Left some suggestions

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoaderFactory.kt
p3dr0rv: move for next pr

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
p3dr0rv: move for next pr

common4j/src/main/com/microsoft/identity/common/java/crypto/key/ISecretKeyLoader.kt
p3dr0rv: change  name

common4j/src/main/com/microsoft/identity/common/java/crypto/key/ISecretKeyGenerator.kt
p3dr0rv: these are just informational getters, removed

common/src/main/java/com/microsoft/identity/common/crypto/AndroidAuthSdkStorageEncryptionManager.java
```diff
@@ -54,26 +53,27 @@ public class AndroidAuthSdkStorageEncryptionManager extends StorageEncryptionMan
      */
     public static final String WRAPPED_KEY_FILE_NAME = "adalks";
 
-    private final PredefinedKeyLoader mPredefinedKeyLoader;
-    private final AndroidWrappedKeyLoader mKeyStoreKeyLoader;
+    private final PredefinedKeyProvider mPredefinedKeyLoader;
```
rpdome: nit: rename variables accordingly?

common/src/main/java/com/microsoft/identity/common/crypto/AndroidWrappedKeyLoader.java
```diff
@@ -71,7 +71,15 @@
  * Instead, the actual key that we use to encrypt/decrypt data is 'wrapped/encrypted' with the keystore key
  * before it get saved to the file.
  */
-public class AndroidWrappedKeyLoader extends AES256KeyLoader {
+public class AndroidWrappedKeyLoader extends AbstractAES256SecretKeyProvider {
```
rpdome: nit: rename class, loader -> provider?

common4j/src/main/com/microsoft/identity/common/java/crypto/key/AbstractAES256SecretKeyProvider.java
```diff
@@ -22,42 +22,21 @@
...
 
+/**
+ * Abstract base class for loading AES-256 secret keys.
+ * <p>
+ * This class implements the {@link ISecretKeyProvider} interface and provides a default
+ * implementation for retrieving an AES-256 key generator. It serves as a base class
+ * for concrete implementations that handle the loading of AES-256 secret keys from
+ * various sources.
+ */
+public abstract class AbstractAES256SecretKeyProvider implements ISecretKeyProvider {
```
rpdome: doesn't look like this abstract class is overriding anything. do we need this?

if all we need are constants/static object - we can get rid of this class and
- expose KEY_ALGORITHM in AES256SecretKeyGenerator
- add a static getInstance() in AES256SecretKeyGenerator() to replace AES_256_KEY_GENERATOR

rpdome (approved):
Left a few suggestions. otherwise lgtm.

common4j/src/main/com/microsoft/identity/common/java/crypto/key/AbstractAES256SecretKeyProvider.java
p3dr0rv: done

mohitc1 (approved):

PR #2674, opened 2025-06-16
Add KeyStoreBackedSecretKeyProvider, Fixes AB#3274176

p3dr0rv:
## Summary
This PR has two main goals: to seamlessly update the encryption padding used in the Android KeyStore, and to lay the groundwork for integrating future cryptographic requirements with minimal changes.

---

### Phase 1: Replace `AndroidWrappedKeyProvider` with `KeyStoreBackedSecretKeyProvider`

- **Controller Flag:** `ENABLE_KEYSTORE_BACKED_SECRET_KEY_PROVIDER`

- **Behavior When Enabled:**
  - The factory will return `KeyStoreBackedSecretKeyProvider` instead of `AndroidWrappedKeyProvider`.
  - The new class behaves similarly to the previous one but introduces the ability to:
    - Retrieve the paddings supported by the KeyStore `KeyPair`
  - Maintain a list of available ciphers (`oaepCipherSpec` and `pkcs1CipherSpec`)
  - Select a compatible cipher for encryption/decryption based on what the `KeyPair` supports.
    - In this case, if the `KeyPair` only supports `PKCS1`, then all keys will be encrypted and decrypted using `PKCS1`.

- **Goal:**
  - Validate that the new provider functions equivalently to the existing one.
  > Note: No encryption changes are introduced in this phase.

### Tests

- `AndroidWrappedKeyProviderTest` has been parameterized to confirm both providers behave identically.
- `KeyStoreBackedSecretKeyProviderRolloutTest.java` includes test to validate migration and rollback

---

### Phase 2: Enable New Wrapped Secret Key Format with Metadata Support
Before introducing new key types, we need to ensure that all new keys include additional metadata that specifies how they were encrypted.

- **Controller Flag:** `ENABLE_NEW_WRAPPED_SECRET_KEY_FORMAT`

- **Behavior When Enabled:**
  - `WrappedSecretKey` will use an enhanced binary format with the following structure:
    - MAGIC BYTES (4 bytes)
    - SERIALIZER ID (4bytes)
    - Metadata length (4 bytes)
    - Metadata (format is decided by the serializer)
    - Raw encrypted key bytes

- **Goal:**
  - Future-proof the wrapped key by including metadata that specifies:
    - Key size
    - Algorithm
    - Transformation used for encryption
    - Key version  
  - This allows the system to handle multiple key types if formats change in the future.
  - Validate `WrappedSecretKey` can read both new and old keys.

### Tests
  - `WrappedSecretKeyTest`

---

### Phase 3: Add OAEP encryption padding.

- **Controller Flag:** `ENABLE_OAEP_WITH_SHA_AND_MGF1_PADDING`

- **Behavior When Enabled:**
  - `CryptoParameterSpecFactory` will now return two paddings: `ENCRYPTION_PADDING_RSA_PKCS1` and `ENCRYPTION_PADDING_RSA_OAEP` for the KeyGen specifications.
  - `KeyStoreBackedSecretKeyProvider` will generate a `KeyPair` that supports `ENCRYPTION_PADDING_RSA_OAEP`.
  - As a result, when a `SecretKey` needs to be wrapped, the provider will select OAEP, since the `KeyPair` now supports this padding and OAEP cipher support was already introduced in Phase 1.

- **Goal:**
  - All new keys will be encrypted /decrypted with OAEP
  - Existing keys will still be encrypted /decrypted with PKCS1

### Tests
- `KeyStoreBackedSecretKeyProviderRolloutTest.java` includes test to validate migration and rollback

---

### Telemetry
| span_name         | key_pair_gen_successful_method                                                              | elapsed_time_keypair_generation | secret_key_wrapping_cipher                                | secret_key_wrapping_operation                               |
|-------------------|----------------------------------------------------------------------------------------------|----------------------------------|------------------------------------------------------------|--------------------------------------------------------------|
| SecretKeyWrapping | null                                                                                        |                                  | CipherSpec(transformation='RSA/ECB/PKCS1Padding')         | KeyStoreBackedSecretKeyProvider:unwrapSecretKey             |
| KeyPairGeneration | KeyGenSpec(description='modern_spec_with_wrap_key', algorithm='RSA', encryptionPaddings='[PKCS1Padding]') | 581                              |                                                            |                                                              |
| KeyPairGeneration | KeyGenSpec(description='modern_spec_with_wrap_key', algorithm='RSA', encryptionPaddings='[PKCS1Padding]') | 294                              |                                                            |                                                              |

[AB#3274176](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3274176)
https://github.com/AzureAD/ad-accounts-for-android/pull/3166


common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,486 @@
...
+    override val key: SecretKey
+        get() {
+            val methodTag = "$TAG:getKey"
+
+            keyFromCache?.let {
+                Logger.info(
+                    methodTag,
+                    "Key is already cached, returning cached key with thumbprint: " +
+                            KeyUtil.getKeyThumbPrint(it)
+                )
```
mohitc1: Do we need this log?

It will generate a lot log lines and flood the PL logs, and reduce the amount of useful logs we'd have to debug

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,486 @@
...
+         * Support for sharedUserId is deprecated, however some applications still use this Android capability.
+         * See: https://developer.android.com/guide/topics/manifest/manifest-element
+         *
+         * To address apps in this scenario we will attempt to load an existing KeyPair
+         * instead of immediately generating a new key pair.  This will use the same keypair
+         * to encrypt the symmetric key generated separately for each
+         * application using a shared linux user id... and avoid these applications from
+         * stomping/overwriting one another's keypair.
+         */
+        val methodTag = "$TAG:generateRandomKey"
```
mohitc1: >generateRandomKey [](http://example.com/codeflow?start=30&length=17)

Tag

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,486 @@
...
+                Logger.info(
+                    methodTag,
+                    "Key loaded from storage and cached with thumbprint: " +
+                            KeyUtil.getKeyThumbPrint(it)
+                )
+                return it
+            }
+
+            val newKey = generateNewSecretKey()
+            sKeyCacheMap[filePath] = newKey
```
mohitc1, line 159: >sKeyCacheMap[filePath] = newKey [](http://example.com/codeflow?start=12&length=31)

shall we move this generateNewSecretKey. It writes to file, can as well write to in-memory cache.

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,486 @@
...
+        val methodTag = "$TAG:unwrapSecretKey"
+        val span = OTelUtility.createSpanFromParent(
+            SpanName.SecretKeyWrapping.name,
+            SpanExtension.current().spanContext
+        )
+
+        return try {
+            SpanExtension.makeCurrentSpan(span).use { _ ->
+                span.setAttribute(AttributeName.secret_key_wrapping_operation.name, methodTag)
+                val cipherParamsSpec = selectCompatibleCipherSpec(keyPair)
```
mohitc1: >selectCompatibleCipherSpec [](http://example.com/codeflow?start=39&length=26)

I think you should use the same cipher spec that you used for wrapping, instead of running selection.

What if the selection differs? say due to OS update

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+        private const val LEGACY_SPEC = "legacy_key_gen_spec"
+    }
+
+    // Feature flags to control which key generation specs to use
+    private val keySpecWithPurposeKey =
+        flightsProvider.isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITH_PURPOSE_WRAP_KEY)
+    private val keySpecWithoutPurposeKey =
+        flightsProvider.isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITHOUT_PURPOSE_WRAP_KEY)
+    private val enableKeyGenEncryptionPaddingRsaOaep =
+        flightsProvider.isFlightEnabled(CommonFlight.ENABLE_OAEP_WITH_SHA_AND_MGF1_PADDING)
```
mohitc1, line 88: This caches flight during object construction itself. Theoretcially can contain sale stale info. Can we directly check for flight when getPrioritizedKeyGenParameterSpecs is called?

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
p3dr0rv: not removed

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
p3dr0rv: I put it outside on purpose, to explicitly do the cache operations here in getKey

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
p3dr0rv: ## Key Wrapping and Cipher Selection Logic

If I understand correctly (IIUC), whenever we create a secret key:

1. We either:
   - Use an existing key pair in the Android Keystore, or  
   - Generate a new key pair for the given alias to wrap the secret key.
2. To select the cipher to wrap:
   - We choose the most secure option supported by the key pair.
   - If `oaepCipherSpec` is available, we use that.
   - Otherwise, we fall back to `pkcs1CipherSpec`.

And the encrypted data becomes dependent of that keypair

When unwrapping the key, the same logic is applied—we select the best cipher supported by the key pair, matching the logic used during wrapping.

---

## Potential Issues

1. **Key Pair Replacement**  
   If the original key pair is deleted and replaced with a new one, the new key may support different cipher specs.  
   However, unwrapping will fail regardless of the cipher due to the use of a different key pair.

2. **Cipher**
   If we stop supporting a specific cipher (e.g., OAEP or PKCS1), we won’t be able to unwrap keys that were previously wrapped with that cipher.

p3dr0rv: To make this more robust, we could create a metadata file alongside the encrypted file. This metadata would store details such as the cipher used, the key algorithm, and other relevant information. It would also make it easier to update or change the specifications of the wrapped key in the future.

p3dr0rv: Addressed

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+     * with legacy fallback always included for maximum compatibility.
+     *
+     * @return List of [IKeyGenSpec] objects ordered by preference (most modern first)
+     */
+    fun getPrioritizedKeyGenParameterSpecs(): List<IKeyGenSpec> {
+        val methodTag = "$TAG:getPrioritizedKeyGenParameterSpecs"
+        val specs = mutableListOf<IKeyGenSpec>()
+
+        // Add specs in order of preference
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && keySpecWithPurposeKey) {
```
mohitc1: >Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && keySpecWithPurposeKey [](http://example.com/codeflow?start=12&length=71)

can be in get()'s implemetnation

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+            // First priority: API 28+ with PURPOSE_WRAP_KEY if enabled
+            specs.add(keyGenParamSpecWithPurposeWrapKey)
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && keySpecWithoutPurposeKey) {
```
mohitc1: >Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && keySpecWithoutPurposeKey [](http://example.com/codeflow?start=12&length=74)

can be get()'s implementation

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+        private const val KEY_SIZE: Int = 2048
+
+        // Descriptive identifiers for different key generation specifications
+        private const val MODERN_SPEC_WITH_PURPOSE_WRAP_KEY = "modern_spec_with_wrap_key"
+        private const val MODERN_SPEC_WITHOUT_PURPOSE_WRAP_KEY = "modern_spec_without_wrap_key"
+        private const val LEGACY_SPEC = "legacy_key_gen_spec"
+    }
+
+    // Feature flags to control which key generation specs to use. (evaluated every time accessed)
+    private val keySpecWithPurposeKey get() =
```
somalaya: nit : keySpecWithWrapPurposeKey

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+        flightsProvider.isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITH_PURPOSE_WRAP_KEY)
+    private val keySpecWithoutPurposeKey get() =
```
somalaya: nit same as above, include "wrap" : keySpecWithoutWrapPurposeKey

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,506 @@
...
+            )
+            deleteSecretKeyFromStorage()
+            throw e
+        }
+    }
+
+    private fun wrapSecretKey(
+        secretKey: SecretKey,
+        keyPair: KeyPair
+    ): Pair<ByteArray, CipherSpec> {
```
mohitc1: >Pair<ByteArray, CipherSpec> [](http://example.com/codeflow?start=7&length=27)

Return WrappedSecretKey

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Represents a wrapped secret key with metadata for algorithm and cipher transformation.
+ *
+ * This class supports both old and new storage formats for wrapped keys.
+ * The new format includes metadata header for better compatibility and extensibility.
+ */
+class WrappedSecretKey(
+    val byteArray: ByteArray,
+    val algorithm: String,
+    val cipherTransformation: String
```
mohitc1: File reading/storing should be in Provider KeyStoreBackedSecretKeyProvider class itself.

You may expose method like serialize/deserialize on this class, and provider can call it.

Based on new format or oldformat, serialize can return accordingly.

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,506 @@
...
+        val methodTag = "$TAG:unwrapSecretKey"
+        val span = OTelUtility.createSpanFromParent(
+            SpanName.SecretKeyWrapping.name,
+            SpanExtension.current().spanContext
+        )
+
+        return try {
+            SpanExtension.makeCurrentSpan(span).use { _ ->
+                span.setAttribute(AttributeName.secret_key_wrapping_operation.name, "UNWRAP")
+                val cipherParamsSpec = getKeyPairCompatibleCipherSpecs(keyPair).firstOrNull { spec ->
```
mohitc1, line 307: >getKeyPairCompatibleCipherSpecs(keyPair) [](http://example.com/codeflow?start=39&length=40)

Why do we need to first get compatible and apply second filtering based on ciphertransform?

You can directly select spec, which matches wrappedSecretKey.cipherTransformation, single filtering

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Stores the wrapped secret key to file.
+     *
+     * @param file The file to store the key data
+     */
+    fun storeOnFile(file: File) {
+        val methodTag = "WrappedSecretKey:storeOnFile"
+        try {
+            val useNewFormat =
```
somalaya: nit : useNewSecretKeyFormat

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+                    .getFlightsProvider()
+                    .isFlightEnabled(CommonFlight.ENABLE_NEW_WRAPPED_SECRET_KEY_FORMAT)
+
+            if (useNewFormat) {
+                storeOnFileNewFormat(file)
+            } else {
+                FileUtil.writeDataToFile(byteArray, file)
+            }
+        } catch (e: Exception) {
+            Logger.error(methodTag, "Failed to store key on disk", e)
```
somalaya:
```suggestion
           val errorMessage = "Failed to store key on disk"
            Logger.error(methodTag, errorMessage , e)
            throw ClientException(ClientException.IO_ERROR, errorMessage , e)
```

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+import java.nio.ByteBuffer
+
+/**
+ * Represents a wrapped secret key with metadata for algorithm and cipher transformation.
+ *
+ * This class supports both old and new storage formats for wrapped keys.
+ * The new format includes metadata header for better compatibility and extensibility.
+ */
+class WrappedSecretKey(
+    val byteArray: ByteArray,
```
somalaya: var name can be wrappedKeyData

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+        fun loadFromFile(file: File, fileSize: Int): WrappedSecretKey? {
+            val methodTag = "$TAG:loadFromFile"
+
+            if (!file.exists()) {
+                Logger.warn(methodTag, "Key file does not exist")
+                return null
+            }
+
+            try {
+                val rawData = FileUtil.readFromFile(file, fileSize)
```
somalaya: is this actually raw data or wrapped data/secret key?

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
```diff
@@ -0,0 +1,226 @@
...
+                    loadFromOldFormat(rawData)
+                }
+            } catch (e: Exception) {
+                Logger.error(methodTag, "Failed to load key from file", e)
+                throw ClientException(ClientException.KEY_LOAD_FAILURE, "Failed to load key from file", e)
+            }
+        }
+
+        /**
+         * Determines if the raw data is in new binary format or old binary format.
```
somalaya:
```suggestion
         * Determines if the raw data uses the new wrapped key format.

File: common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt [hunk 18]
```diff
@@ -0,0 +1,226 @@
...
+         *
+         * @param rawData The raw bytes read from file
+         * @return true if data is in new format, false if old format
+         */
+        private fun isNewFormat(rawData: ByteArray): Boolean {
+            if (rawData.size < 8) return false
```
somalaya: is the number 8 coming from 4 bytes for header, 4 for metadata length? If yes, please specify it in the comment

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -432,6 +447,15 @@ public enum AttributeName {
     /**
      * Records the time (in milliseconds) spent on flight check for webcp.
      */
-    web_cp_flight_get_time
+    web_cp_flight_get_time,
 
+    /**
+     * Records the Cipher transformation used to encrypt/decrypt the keypair.
```
somalaya, line 483: is it for encrypt/decrypt or wrap/unwrap?

somalaya:
Is there any reason why Phase 2 and phase 3 should not be done together?

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
p3dr0rv: reworked

common/src/main/java/com/microsoft/identity/common/crypto/WrappedSecretKey.kt
p3dr0rv: renamed

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
p3dr0rv: encrypt/decrypt https://developer.android.com/reference/javax/crypto/Cipher

p3dr0rv:
> Is there any reason why Phase 2 and phase 3 should not be done together?

There’s no limitation in the code; I chose this approach so that in case of a rollback, it’s easier to move from one state to another. It also provides more granularity in the release, allowing us to ensure that each step of the new changes works as expected.

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
p3dr0rv: We need to ensure that the KeyPair we use for unwrapping is capable of performing that operation. If the KeyPair does not support the padding required by the cipher, it will not be able to unwrap the key.

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
p3dr0rv: Could you explain this in more detail? I’m not following.

mohitc1: get() for this property can implementation like    

private val keySpecWithWrapPurposeKey get() = Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && 
        flightsProvider.isFlightEnabled(CommonFlight.ENABLE_NEW_KEY_GEN_SPEC_FOR_WRAP_WITH_PURPOSE_WRAP_KEY)

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeyLegacySerializer.kt
```diff
@@ -0,0 +1,100 @@
...
+ * - Default algorithm: AES
+ * - Default cipher transformation: RSA/ECB/PKCS1Padding
+ *
+ * This serializer is essential for maintaining compatibility when migrating from older
+ * key storage formats to newer metadata-aware formats.
+ *
+ * @see IWrappedSecretKeySerializer
+ * @see WrappedSecretKey
+ */
+class WrappedSecretKeyLegacySerializer : IWrappedSecretKeySerializer {
```
mohitc1, line 44: >WrappedSecretKeyLegacySerializer [](http://example.com/codeflow?start=6&length=32)

Will this be used with new KeyProvider?

I think KeyProvider should be able to read/deserialize old values. But if new wrapped key or secretkey is getting generated, it should be new format.

p3dr0rv: The new KeyProvider will be able to read and deserialize both old and new data formats.
When it comes to writing, the new provider supports both formats. During Phase 1, it will continue to write using the old format. In Phase 2, it will switch to writing in the new format.


Why this phased approach?
This strategy allows us to validate the new implementation while maintaining a clean rollback path. By starting with the old format, we can ensure the new provider behaves identically to the legacy one. Once validated, we move to Phase 2 to test writing in the new format.

In case we encounter issues during Phase 1, we can safely revert to the previous provider since it will still be able to read keys generated by the new one. However, if we were to immediately adopt the new format, a rollback would be problematic—because the legacy provider wouldn’t be able to read those newly formatted keys.

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeyJsonObjectSerializer.kt
```diff
@@ -0,0 +1,171 @@
...
+class WrappedSecretKeyJsonObjectSerializer: IWrappedSecretKeySerializer {
+
+    companion object {
+        private const val TAG = "WrappedSecretKeyJsonObjectSerializer"
+
+        /** Version identifier for the JSON-based format */
+        const val VERSION = 1
+
+        /** Header identifier used to recognize this format */
+        private const val NEW_FORMAT_HEADER_IDENTIFIER = 0x00FF3C01
```
mohitc1: >NEW_FORMAT_HEADER_IDENTIFIER [](http://example.com/codeflow?start=26&length=28)

what is significance of the header?

p3dr0rv: Reworked as discussed offline, the header will only represent a unique identifier to indicate the byte array contains metadata

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeyBinaryStreamSerializer.kt
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @param wrappedSecretKey The wrapped secret key containing metadata to serialize
+     * @return Binary encoded metadata as byte array
+     */
+    @Throws(java.io.IOException::class)
+    override fun serializeMetadata(wrappedSecretKey: WrappedSecretKey): ByteArray {
+        return ByteArrayOutputStream().use { baos ->
+            DataOutputStream(baos).use { dos ->
+                // Write algorithm (writeUTF includes length prefix)
+                dos.writeUTF(wrappedSecretKey.algorithm)
```
somalaya, line 63: nit : change baos bytesOutputStream, similarly change daos

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -0,0 +1,527 @@
...
+        val compatibleSpecs = availableCipherSpecs.filter { spec ->
+            supportedPaddings.any { padding ->
+                spec.padding.contains(padding, ignoreCase = true)
+            }
+        }
+        Logger.verbose(methodTag, "Found ${compatibleSpecs.size} compatible cipher specs: $compatibleSpecs")
+        return compatibleSpecs
+    }
+
+
```
somalaya: nit : extra line

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeyLegacySerializer.kt
```diff
@@ -0,0 +1,108 @@
...
+
+
+/**
+ * Legacy serializer for [WrappedSecretKey] that maintains backward compatibility.
+ *
+ * Handles the original format with raw key data only, applying default metadata
+ * during deserialization for proper key reconstruction.
+ *
+ * **Format:**
+ * - **ID**: 0 (legacy)
```
somalaya, line 33: Can this be called as VERSION_NUMBER or VERSION instead of ID?

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/crypto/KeyGenSpec.kt
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @property purposes Key usage purposes (e.g., ENCRYPT, DECRYPT, WRAP_KEY)
+ * @property digestAlgorithms Supported digest algorithms (e.g., SHA-256, SHA-512)
+ * @property keyAlias Unique identifier for the key in KeyStore
+ * @property keySize RSA key size in bits (typically 2048)
+ * @property description Human-readable identifier for logging
+ * @property algorithm Key generation algorithm (typically "RSA")
+ * @property encryptionPaddings Supported padding schemes (e.g., PKCS1, OAEP)
+ */
+data class KeyGenSpec(
```
mohitc1, line 62: nit: can this whole class be annotated as M?

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeySerializerWithMetadata.kt
```diff
@@ -0,0 +1,191 @@
...
+     * Deserializes metadata from its byte array representation.
+     *
+     * Concrete implementations must provide format-specific metadata deserialization
+     * that corresponds to their [serializeMetadata] implementation.
+     *
+     * @param metadataByteArray The serialized metadata bytes
+     * @return The deserialized metadata object
+     * @throws Exception if metadata format is invalid or corrupted
+     */
+    abstract fun deserializeMetadata(metadataByteArray: ByteArray): WrappedSecretKeyMetadata
```
mohitc1: >abstract [](http://example.com/codeflow?start=4&length=8)

protected?

common/src/main/java/com/microsoft/identity/common/crypto/wrappedsecretkey/WrappedSecretKeySerializerWithMetadata.kt
```diff
@@ -0,0 +1,191 @@
...
+ * - **Metadata Length**: Size of the metadata section in bytes
+ * ** Body components:
+ * - **Metadata**: Format-specific metadata (JSON, binary, etc.)
+ * - **Raw Key Data**: The actual wrapped secret key bytes
+ *
+ * @see IWrappedSecretKeySerializer
+ * @see WrappedSecretKeySerializerManager
+ */
+
+abstract class WrappedSecretKeySerializerWithMetadata: IWrappedSecretKeySerializer {
```
mohitc1: >WrappedSecretKeySerializerWithMetadata [](http://example.com/codeflow?start=15&length=38)

Can we just call it  AbstractWrappedSecretKeySerializer?

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/crypto/KeyGenSpec.kt
p3dr0rv: In Kotlin there is no M Lombok so the way to do it is in the data object

mohitc1 (approved):

PR #2679, opened 2025-06-18
Use tdbr claim to route telemetry traffic to EU region, Fixes AB#3200872

fadidurah:
Introduce Tracking for a new TDBR Claim that ESTS will send as part of ClientInfo to denote EU Home tenants, then fetch the claim to route telemetry to EU when applicable. This will be used to better comply with European Data laws

[AB#3200872](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3200872)


common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1421,4 +1426,23 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
                     "So, this is not likely a broker version supported issue. Continuing.");
         }
     }
+
+    private void trackTelemetryRegionFromResultBundle(@NonNull final Bundle resultBundle) throws BaseException {
+        final String methodTag = TAG + ":trackTelemetryRegionFromResultBundle";
+        final BrokerResult brokerResult = new MsalBrokerResultAdapter().brokerResultFromBundle(resultBundle);
```
shahzaibj: Is this a duplicate deserialization? I suspect we unpack the bundle into BrokerResult already somewhere in the code. If yes, then let's just pass BrokerResult here (as opposed to passing a bundle and deserializing again)

common4j/src/main/com/microsoft/identity/common/java/util/IPlatformUtil.java
```diff
@@ -129,4 +131,11 @@ public interface IPlatformUtil {
      */
     @Nullable
     List<Map.Entry<String, String>> updateWithAndGetPlatformSpecificExtraQueryParameters(@Nullable List<Map.Entry<String, String>> originalList);
+
+    /**
+     * Store telemetry region by tenant.
+     *
+     * @param clientInfo the client info containing tenant information and tdbr claim
+     */
+    void storeTelemetryRegionByTenant(@NonNull IStorageSupplier supplier, @NonNull ClientInfo clientInfo);
```
shahzaibj: I honestly think there is no reason for this method to in the platform util. This method can easily be in its own class or any class related to telemetry helpers. and it would still give you platform specific flexibility because you are passing a generic `IStorageSupplier` to the params which takes care of that already (so no need to do double platform agnostic-ness by putting it on platform util)

fadidurah: Done

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
fadidurah: I refactored some of the code to remove duplicate deserialization

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1322,6 +1326,9 @@ private void saveMsaAccountToCache(final @NonNull Bundle resultBundle,
 
             try {
                 final ClientInfo clientInfo = new ClientInfo(brokerResult.getClientInfo());
+                // Store the telemetry region info in shared preferences
+                EUClaimStorageUtil.Companion.storeTelemetryRegionByTenant(mComponents.getStorageSupplier(), clientInfo);
```
shahzaibj: You're having to basically invoke this in a number of places ... instead of doing so much work... why not just invoke this method directly in the `ClientInfo` constructor? that way you invoke if once and it gets the job done? the rest of the code becomes significantly cleaner

shahzaibj: Ok, I see. you also need to supply storage supplier..

fadidurah: Honestly, this was my original idea. Just invoke it whenever we create Client Info. But the issue i ran into with that is that we would end up calling it multiple times in a single flow most likely. I think doing at when broker is about to save an account makes sense. I didn't want the additional shared preferences calls, I figured they would be somewhat costly.

shahzaibj: but still, I'd imagine there would be a would be a way to do this within ClientInfo... maybe a new constructor in that class that can take in `StorageSupplier`

shahzaibj: Let's just think about it a little bit...if there is a way to do this from within ClientInfo - that way caller doesn't have to remember to then invoke `storeTelemetryRegionByTenant` in so many places because it would be guaranteed to be happening every time ClientInfo is constructed

shahzaibj: > Honestly, this was my original idea. Just invoke it whenever we create Client Info. But the issue i ran into with that is that we would end up calling it multiple times in a single flow most likely. I think doing at when broker is about to save an account makes sense. I didn't want the additional shared preferences calls, I figured they would be somewhat costly.

Oh, so we are constructing ClientInfo multiple times even in the same flow for a single request?

fadidurah: Yeah, I put a checkpoint in ClientInfo constructor, called AT, and the breakpoint was hit 6 or 7 times. I didn't test other flows though

fadidurah: There is a check to only access the shared preferences when the tdbr claim is not null, so that should cut some of the noise. But i also saw multiple client Infos being created with the tdbr claim attached. I think this approach is better from a perf standpoint.

shahzaibj: Ok. Makes sense.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
shahzaibj, line 1: Just realized, this file is for MSAL - we don't currently track telemetry for MSAL. We only do so for Broker. Is there a reason we are tracking the telemetry region in this file?

fadidurah: Yeah makes sense, i've removed the code here

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -131,6 +131,11 @@ public enum CommonFlight implements IFlightConfig {
      */
     ENABLE_WEB_CP_IN_WEBVIEW("EnableWebCpInWebView", false),
 
+    /**
+     * Flight to enable using the new EU TDBR Claim sent through ClientInfo
+     */
+    ENABLE_USING_TDBR_CLAIM_FOR_EU_ROUTING("EnableUsingTdbrClaimForEuRouting", true),
```
shahzaibj: Why not go default false?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -27,6 +27,10 @@
  * NOTE : Any changes to this enum should also be made in the corresponding enum in Broker.
  */
 public enum AttributeName {
+    /**
+     * The tenant id for the home tenant of the account for which PRT is required.
+     */
+    tenant_id,
```
shahzaibj: Weren't we already tracking tenant_id?

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
fadidurah: Same reason as in the other PR. I don't plan to really roll this out. The reason i'm adding this flight is to follow our standard of at least having an emergency shut off in case this causes issues in PROD. Have it as true by default, disable it if need be.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
fadidurah: This addition is because i now track tenant id whenever we get the tdbr claim. This way we can have it populated for the telemetry region decision later in this same flow. This happens in common code, so i needed to add it here. I think in other scenarios (in broker), it gets populated into span either from broker command parameters or wpj data for when accounts are already present in our storage.

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/EUClaimStorageUtil.kt
```diff
@@ -0,0 +1,85 @@
...
+            )
+            val tdbrValueStore = supplier.getUnencryptedNameValueStore(
+                ClientInfo.TDBR_CLAIM,
+                String::class.java
+            )
+            tdbrValueStore.put(tenantId, tdbrClaim)
+
+            // Attach tenant id to the current span
+            SpanExtension.current().setAttribute(AttributeName.tenant_id.name, tenantId)
+        }
```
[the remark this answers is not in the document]
fadidurah: That work is in broker repo

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/EUClaimStorageUtil.kt
```diff
@@ -0,0 +1,85 @@
...
+                    methodTag,
+                    "Received no tdbr claim, not storing anything in shared preferences.."
+                )
+                return
+            }
+
+            // Store the tdbr claim for a specific tenant ID
+            Logger.info(
+                methodTag,
+                "Storing telemetry region by tenant: $tenantId, TDBR Claim: $tdbrClaim"
```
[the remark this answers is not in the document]
fadidurah: These are not sensitive fields

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/EUClaimStorageUtil.kt
```diff
@@ -0,0 +1,85 @@
...
+        @JvmStatic
+        fun storeTelemetryRegionByTenant(
+            supplier: IStorageSupplier,
+            clientInfo: ClientInfo
+        ) {
+            val methodTag = "$TAG:storeTelemetryRegionByTenant"
+
+            val tenantId = clientInfo.utid
+            val tdbrClaim = clientInfo.tdbrClaim
+
```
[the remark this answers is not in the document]
fadidurah: @copilot add tests here

melissaahn (approved):

shahzaibj (approved):

shahzaibj (approved):

PR #2682, opened 2025-06-24
Construct broker app link redirect URI based on broker pkg name + move under /androidbroker path

shahzaibj:
This basically does a couple of changes:

- 1st: Received feedback from eSTS folks to move app links under `androidbroker/` path as opposed to the root of the eSTS domain.
- 2nd: for each broker app specific app link, we were using a friendly identifier at the end such as /authenticator to denote authapp and `/ltw` to denote ltw. I've moved away from these custom friendly names and decided to just put the pkg name into the path. This allows us to declare an intent filter for it directly in our broker's intent filter as opposed to asking partner teams to put an intent filter in their manifest. 

Related Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3139


p3dr0rv (approved):
:shipit:

fadidurah (approved):

PR #2686, opened 2025-06-25
Flighting webcp in webview, Fixes AB#3307594

somalaya:
1. Flighting for webcp in webview is divided into 3 parts
- Not flighting this feature for brokerless scenarios. The flight will be off by default for brokerless scenarios as Intune team is ok to only target brokered flows since this is the majority case.
- Now coming to brokered scenarios, we have below
    -      Simple flighting with ENABLE_WEB_CP_IN_WEBVIEW which will be used for rollout starting October.
    -      Using tenantId list TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW. This flight will be used first in the months of August and September. This will be 100% rolled out in the beginning itself. This flight will hold upto 25 tenantIds. See below snapshot 
image

2. I intentionally added a global variable to keep the flight value instead of calling isWebCpInWebviewFeatureEnabled method because I do not want to invoke code to get tenantId multiple times. Also, this value must remain the same once the webcp url is loaded in webview. 

3. Apart from this, I have added a special method to handle google enrollment URL which has extra intent flags needed for this scenario.

Related PR : https://github.com/AzureAD/ad-accounts-for-android/pull/3140

Fixes [AB#3307594](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3307594)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -644,20 +644,24 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
         }
 
         // Else, check if the tenant is in the list of tenants that have this feature enabled.
-        final String username = StringExtensions.getUrlParameters(originalUrl).get(AuthenticationConstants.AAD.LOGIN_HINT);
-        if (!StringUtil.isNullOrEmpty(username)) {
-            final String tenantId = CommonTenantInfoProvider.INSTANCE.getTenantId(username);
-            if (StringUtil.isNullOrEmpty(tenantId)) {
-                return false;
-            }
-            final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
-            final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
-            Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
-            mIsWebCpInWebViewFeatureEnabled = isFlightEnabledForCurrentTenant;
-            return isFlightEnabledForCurrentTenant;
+        final String tenantId = getTenantIdFromUrl(originalUrl);
+        if (StringUtil.isNullOrEmpty(tenantId)) {
+            return false;
         }
+        
+        final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
+        final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
+        Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
+        mIsWebCpInWebViewFeatureEnabled = isFlightEnabledForCurrentTenant;
+        return isFlightEnabledForCurrentTenant;
+    }
 
-        return false;
+    private String getTenantIdFromUrl(@NonNull final String url) {
+        final String username = StringExtensions.getUrlParameters(url).get(AuthenticationConstants.AAD.LOGIN_HINT);
+        if (StringUtil.isNullOrEmpty(username)) {
+            return null;
+        }
+        return CommonTenantInfoProvider.INSTANCE.getTenantId(username);
     }
 
     // WebCP enrollment URL is a guided enrollment page showing instructions on how to enroll within the productivity app.
@@ -693,7 +697,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
             getActivity().startActivity(intent);
         } catch (android.content.ActivityNotFoundException e) {
-            Logger.error(methodTag,"Failed to open the Authenticator application.", e);
+            Logger.error(methodTag,"Failed to open the intent for google enrollment.", e);
         }
     }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -641,29 +641,35 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     protected boolean isWebCpInWebviewFeatureEnabled(@NonNull final String originalUrl) {
         final String methodTag = TAG + ":isWebCpInWebviewFeatureEnabled";
-        if (!ProcessUtil.isRunningOnAuthService(getActivity().getApplicationContext())) {
-            // Enabling webcp in webview feature for brokered flows only for now.
+        try {
+            if (!ProcessUtil.isRunningOnAuthService(getActivity().getApplicationContext())) {
+                // Enabling webcp in webview feature for brokered flows only for now.
+                return false;
+            }
+
+            if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW)) {
+                // Directly enabled via flight rollout.
+                Logger.info(methodTag, "WebCP in WebView feature is enabled.");
+                mIsWebCpInWebViewFeatureEnabled = true;
+                return true;
+            }
+
+            // Else, check if the home tenant is in the list of tenants that have this feature enabled.
+            final String tenantId = getHomeTenantIdFromUrl(originalUrl);
+            if (StringUtil.isNullOrEmpty(tenantId)) {
+                return false;
+            }
+
+            final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
+            final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
+            Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
+            mIsWebCpInWebViewFeatureEnabled = isFlightEnabledForCurrentTenant;
+            return isFlightEnabledForCurrentTenant;
+        } catch (final Throwable throwable) {
+            // Catching any unexpected exceptions to avoid breaking the flow. We will anyway remove this block once the feature is fully rolled out.
+            Logger.error(methodTag, "Failed to check if WebCP in WebView feature is enabled.", throwable);
             return false;
         }
-
-        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW)) {
-            // Directly enabled via flight rollout.
-            Logger.info(methodTag, "WebCP in WebView feature is enabled.");
-            mIsWebCpInWebViewFeatureEnabled = true;
-            return true;
-        }
-
-        // Else, check if the home tenant is in the list of tenants that have this feature enabled.
-        final String tenantId = getHomeTenantIdFromUrl(originalUrl);
-        if (StringUtil.isNullOrEmpty(tenantId)) {
-            return false;
-        }
-        
-        final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
-        final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
-        Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
-        mIsWebCpInWebViewFeatureEnabled = isFlightEnabledForCurrentTenant;
-        return isFlightEnabledForCurrentTenant;
     }
 
     private String getHomeTenantIdFromUrl(@NonNull final String url) {
@@ -716,8 +722,9 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             // Important for the enrollment flow to work correctly.
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
             getActivity().startActivity(intent);
-        } catch (android.content.ActivityNotFoundException e) {
+        } catch (final ActivityNotFoundException e) {
             Logger.error(methodTag,"Failed to open the intent for google enrollment.", e);
+            throw e;
         }
     }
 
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -328,10 +330,10 @@ else if (isRedirectUrl(formattedURL)) {
             } else if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_ATTACH_PRT_HEADER_WHEN_CROSS_CLOUD) && isCrossCloudRedirect(formattedURL)) {
                 Logger.info(methodTag,"Navigation contains cross cloud redirect.");
                 processCrossCloudRedirect(view, url);
-            } else if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW) && isWebCpEnrollmentUrl(url)) {
+            } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpEnrollmentUrl(url)) {
                 Logger.info(methodTag,"Navigation contains web cp enrollment url.");
                 processWebCpEnrollmentUrl(view, url);
-            } else if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW) && isWebCpAuthorizeUrl(url)) {
+            } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpAuthorizeUrl(url)) {
```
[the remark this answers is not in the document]
somalaya: Do not want to use new flight value in the same auth context

common4j/src/main/com/microsoft/identity/common/java/interfaces/ITenantInfoProvider.kt
```diff
@@ -0,0 +1,33 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.java.interfaces
+
+/**
+ * Consumer of commons needs to implement [ITenantInfoProvider] interface
+ * and set it using CommonTenantInfoProvider.initializeCommonTenantInfoProvider(@NonNull tenantInfoProvider: ITenantInfoProvider)
+ * to provide tenantInfo holder to common module.
+ */
+interface ITenantInfoProvider {
```
mohitc1, line 30: >ITenantInfoProvider [](http://example.com/codeflow?start=10&length=19)

ON broker would this query IBrokerAccount for home tenantid?

somalaya: Yes, something like this code snapshot from the corresponding broker pR
image

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -609,6 +610,39 @@ protected void loadDeviceCaUrl(@NonNull final String originalUrl, @NonNull final
...
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW)) {
+            // Directly enabled via flight rollout.
+            Logger.info(methodTag, "WebCP in WebView feature is enabled.");
+            mIsWebCpInWebViewFeatureEnabled = true;
+            return true;
+        }
+
+        // Else, check if the tenant is in the list of tenants that have this feature enabled.
+        final String username = StringExtensions.getUrlParameters(originalUrl).get(AuthenticationConstants.AAD.LOGIN_HINT);
+        if (!StringUtil.isNullOrEmpty(username)) {
```
shahzaibj: I guess this can be simplified to 

```java
if (StringUtil.isNullOrEmpty(username)) {
  return false
}
```

Then the code below can move out of the if

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -609,6 +610,39 @@ protected void loadDeviceCaUrl(@NonNull final String originalUrl, @NonNull final
...
+            final String tenantId = CommonTenantInfoProvider.INSTANCE.getTenantId(username);
+            if (StringUtil.isNullOrEmpty(tenantId)) {
+                return false;
+            }
+            final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
+            final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
+            Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
```
shahzaibj, line 665: This can potentially become a helper method

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -631,6 +665,21 @@ public void run() {
...
+        final String methodTag = TAG + ":openGoogleEnrollmentUrl";
+        Logger.info(methodTag, "Opening Google enrollment URL");
+        try {
+            final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+            // Add flags to ensure the activity is launched in a new task and clears the current task stack.
+            // Important for the enrollment flow to work correctly.
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+            getActivity().startActivity(intent);
+        } catch (android.content.ActivityNotFoundException e) {
+            Logger.error(methodTag,"Failed to open the Authenticator application.", e);
```
shahzaibj: Is it authenticator application? Or rather a browser app?

common/src/test/java/com/microsoft/identity/common/internal/mocks/MockCommonFlightsManager.java
```diff
@@ -0,0 +1,49 @@
...
+import com.microsoft.identity.common.java.flighting.IFlightsProvider;
+
+import org.jetbrains.annotations.NotNull;
+
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+@Setter
+@Accessors(prefix = "m")
+public class MockCommonFlightsManager implements IFlightsManager {
```
shahzaibj, line 35: Did we not already have a mock implementation?

common4j/src/main/com/microsoft/identity/common/java/broker/CommonTenantInfoProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Consumer of commons needs to implement [ITenantInfoProvider] interface
+ * and set it using CommonTenantInfoProvider.initializeCommonTenantInfoProvider(@NonNull tenantInfoProvider: ITenantInfoProvider)
+ * to provide tenantInfo to common module.
+ */
+object CommonTenantInfoProvider : ITenantInfoProvider{
+    private val TAG = CommonTenantInfoProvider::class.java.simpleName
+    private var mTenantInfoProvider: ITenantInfoProvider? = null
+
+    // Note : This method should only be invoked by broker module.
```
shahzaibj: How do we ensure that?

shahzaibj: Also what would go wrong if someone else invokes it?

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -129,7 +129,9 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable the Web CP in WebView.
      */
-    ENABLE_WEB_CP_IN_WEBVIEW("EnableWebCpInWebView", false);
+    ENABLE_WEB_CP_IN_WEBVIEW("EnableWebCpInWebView", false),
+
+    TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW("TenantListToEnableWebCpInWebView", "");
```
shahzaibj, line 142: javadoc

common/src/test/java/com/microsoft/identity/common/internal/mocks/MockCommonFlightsManager.java
somalaya: No, we did not have it in common

common4j/src/main/com/microsoft/identity/common/java/broker/CommonTenantInfoProvider.kt
somalaya: It just means that right now we only support getting tenant related information at this stage using broker account storage. May be in the future, if we can provide tenant information from common module itself, we can invoke it. Since the comment is confusing, i would rather remove it.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -620,7 +658,7 @@ private void processWebCpEnrollmentUrl(@NonNull final WebView view, @NonNull fin
         view.stopLoading();
         Logger.info(methodTag, "Loading WebCP enrollment url in browser.");
         // This is a WebCP enrollment URL, so we need to open it in the browser (it does not work in WebView as google enrollment is enforced to be done in browser).
-        openLinkInBrowser(url);
+        openGoogleEnrollmentUrl(url);
         // We need to return MDM_FLOW result code as the enrollment is done in browser. But this may sometimes take a few seconds to launch the intent.
```
shahzaibj: We should add details here about "why" this can sometimes take a few seconds

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -631,6 +669,21 @@ public void run() {
...
+        final String methodTag = TAG + ":openGoogleEnrollmentUrl";
+        Logger.info(methodTag, "Opening Google enrollment URL");
+        try {
+            final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+            // Add flags to ensure the activity is launched in a new task and clears the current task stack.
+            // Important for the enrollment flow to work correctly.
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+            getActivity().startActivity(intent);
+        } catch (android.content.ActivityNotFoundException e) {
+            Logger.error(methodTag,"Failed to open the intent for google enrollment.", e);
```
shahzaibj, line 726: In this case are we still returning the MDM_FLOW (can't remember exact name) error back to the client app?

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -609,6 +610,43 @@ protected void loadDeviceCaUrl(@NonNull final String originalUrl, @NonNull final
...
+        return isFlightEnabledForCurrentTenant;
+    }
+
+    private String getTenantIdFromUrl(@NonNull final String url) {
+        final String username = StringExtensions.getUrlParameters(url).get(AuthenticationConstants.AAD.LOGIN_HINT);
+        if (StringUtil.isNullOrEmpty(username)) {
+            return null;
+        }
+        return CommonTenantInfoProvider.INSTANCE.getTenantId(username);
+    }
```
mohitc1, line 681: there's utid parameter in MicrosoftAuthorizationRequest object, can you rather use that?

common4j/src/main/com/microsoft/identity/common/java/broker/CommonTenantInfoProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val TAG = CommonTenantInfoProvider::class.java.simpleName
+    private var mTenantInfoProvider: ITenantInfoProvider? = null
+
+    fun initializeCommonTenantInfoProvider(tenantInfoProvider: ITenantInfoProvider) {
+        val methodTag = "$TAG:initializeCommonTenantInfoProvider"
+        Logger.info(methodTag, "Initializing common tenant information provider with " + tenantInfoProvider.javaClass.simpleName)
+        mTenantInfoProvider = tenantInfoProvider
+    }
+
+    override fun getTenantId(username: String): String? {
```
mohitc1: >getTenantId [](http://example.com/codeflow?start=17&length=11)

If intention is homeTenantId, could you rename it homeTenantId? Otherwise it's too generic

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Sure, any intent launch takes few milli seconds in general. Sometimes, when the phone is slow or due to some device battery issues, it could take a few seconds. Generally, it is not very evident because the launching activity is not killed in most scenarios.  In this particular case though,  we are killing the activity that is launching the intent. Hence the transition could be poor.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Are you referring to this file? https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java

somalaya: utid as it is cannot be used as it is not getting set correctly (is null), since we are NOT setting it using the broker account but using the request parameter's account record. Setting this also needs some changes.
The change itself is not too big/major but it makes me scatter the logic to add flight. I prefer the current way of doing things in the PR.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Yes this is still same as what I introduced in initial webcp support PR. Only difference is adding some intent flags. The calling method returns MDM_FLOW error

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -609,6 +610,43 @@ protected void loadDeviceCaUrl(@NonNull final String originalUrl, @NonNull final
...
+        
+        final String tenantIdList = CommonFlightsManager.INSTANCE.getFlightsProvider().getStringValue(CommonFlight.TENANT_LIST_TO_ENABLE_WEB_CP_IN_WEBVIEW);
+        final boolean isFlightEnabledForCurrentTenant = !StringUtil.isNullOrEmpty(tenantIdList) && tenantIdList.contains(tenantId);
+        Logger.info(methodTag, "TenantId list is empty? " + StringUtil.isNullOrEmpty(tenantIdList) + ", Is current tenantId in list? " + isFlightEnabledForCurrentTenant);
+        mIsWebCpInWebViewFeatureEnabled = isFlightEnabledForCurrentTenant;
+        return isFlightEnabledForCurrentTenant;
+    }
+
+    private String getHomeTenantIdFromUrl(@NonNull final String url) {
+        final String username = StringExtensions.getUrlParameters(url).get(AuthenticationConstants.AAD.LOGIN_HINT);
```
mohitc1, line 676: >final String username = StringExtensions.getUrlParameters(url).get(AuthenticationConstants.AAD.LOGIN_HINT); [](http://example.com/codeflow?start=8&length=107)

Is it guaranteed that, login hint will always be there in this URL?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
mohitc1: okay.

mohitc1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: I am seeing this code in eSTS [url has login hint param](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmsazure.visualstudio.com%2FOne%2F_git%2FESTS-Main%3Fpath%3D%2Fsrc%2FProduct%2FMicrosoft.AzureAD.ESTS%2FSts%2FConditionalAccess%2FDevicePolicyError.cs%26version%3DGBmaster%26_a%3Dcontents%26line%3D1106%26lineStyle%3Dplain%26lineEnd%3D1114%26lineStartColumn%3D1%26lineEndColumn%3D1&data=05%7C02%7CSowmya.Malayanur%40microsoft.com%7Cbf4222b96017481b0ad908ddaeacd4ae%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638858779540329792%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=tlC8bO3nTh9YFGl5gHnPVzb%2FPL2iKjihvVsY%2FtZUikk%3D&reserved=0) which always adds the device  ca param (ismdmurl=1) and upn.

mohitc1 (approved):

PR #2691, opened 2025-06-26
Updating handling of ssl error received in Android WebView's onReceivedSslError callback, Fixes AB#3268908

mohitc1:
Update:
After talking to @iulico-1 from OneAuth.
1. First rolling out with Broker and keeping disabled for OneAuth (and MSAL Android). Added SSL error telemetry
2. Enable for oneauth after we see no issues in Broker. So, the toggle code to let OneAuth enable/disable is removed from the PR.
3. Investigate how we can pass the error to OneAuth to track SSL failure and User canceled the flow because of SSL error.

=====================

Fixes [AB#3268908](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3268908)

**Summary:**  
This PR updates the handling of SSL errors in the Android WebView used for authentication. The change is in response to an incident where the MSA sign-up flow failed due to an expired certificate on an MSA UX web page. The expired certificate was associated with a resource loaded by the page; however, the existing SSL error handling in Common’s WebView would cancel the loading of that resource and subsequently terminate the entire authentication flow. (See PBI for details.)

**Solution**
**Option 1**
We simply log and stop loading the resource that caused the error by calling `SssHandler.cancel()` but not stop the flow by erroring out. This is a simple change. The only thing is the main URL (the url loading in the WebView) itself has SSL error then the flow would end up in *blank page*. This is as bad as finishing the flow but does not inform calling app of the error via exception.

But if sub-resource has the error, then the flow continues.

**Flighting**
**New SSL Error Handler**: 
   - ~~Adds a new intent key (`WEB_VIEW_NEW_SSL_ERROR_HANDLER_ENABLED`) in `AuthenticationConstants.java` to control whether the new SSL error handler logic is enabled. This can be used by OneAuth to opt-in or opt-out. If the key is not passed, then Common's default logic based on flighting will be used. If the key passed, that will be used (primarily meant for OneAuth/MSALCPP).~~. The toggle would be added later if required. This simplifies current logic to only rely on Common/broker's flighting.
   - Adds a new flight value is well defaulting to `false`. So, OneAuth and MSAL would also use default flight value and continue using current behavior. For Broker, we would enable new flow with default `true` (https://github.com/AzureAD/ad-accounts-for-android/pull/3141), which can be controlled remotely to turn off if required.

**Option 2** (Not this PR, added for reference)
The objective of this update is to improve SSL error handling so that the flow is only cancelled if an SSL error occurs on the main frame resource (the primary page being loaded). If an SSL error affects a sub-resource, only the loading of that specific resource is cancelled, and the overall authentication flow continues. This approach helps prevent unnecessary failures caused by SSL errors on secondary or non-critical resources. (https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2688)

The con of the approach is client needs to track the active url that's about and being loaded in WebView. The tracking adds cost of developing and maintaining to solution without significant benefit over option 1.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -172,16 +181,17 @@ public void onReceivedSslError(final WebView view,
                                    final SslError error) {
         // Developer does not have option to control this for now
         super.onReceivedSslError(view, handler, error);
-        handler.cancel();
```
iulico-1, line 175: > handler.cancel(); [](http://example.com/codeflow?start=7&length=18)

Should we ensure that we have telemetry events in OneAuth and broker telemetry.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -172,16 +181,17 @@ public void onReceivedSslError(final WebView view,
...
 
         final String errMsg = "Received SSL Error during request. For more info see: " + SSL_HELP_URL;
```
iulico-1: For the OneAuth path it would be good to capture that an TLS error occurred and if the flow ended-up failing (like when the SSL error is for the top level domain not an iframe) we would be in OneAuth a UserCancelException + an error code indicating that there was a TLS error in the mix that we can log and monitor through MATS. 

Please work with @nickbopp on the exact telemetry contract with OneAuth.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -172,16 +181,17 @@ public void onReceivedSslError(final WebView view,
...
 
         Logger.error(TAG + ":onReceivedSslError", errMsg, null);
 
-        // Send the result back to the calling activity
-        mCompletionCallback.onChallengeResponseReceived(
-                RawAuthorizationResult.fromException(
-                        new ClientException("Code:" + ERROR_FAILED_SSL_HANDSHAKE, error.toString())));
+        if (!mShouldPreserveFlowOnSslError) {
```
iulico-1: In general I agree with your stand that going with a simpler solution where we don't need to track and match URLs and RU to identify the top level domain is preferable. 

Let's just ensure we have telemetry to differentiate between user cancellations that will happen because of TLS + rest cases.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
mohitc1: Add broker telemetry to track if SSL error was seen in web view flow.

For OneAuth, this needs different wiring which I don't see we currently have. So, that would be separate change.

As we discussed, I will first rollout this for broker and keep disabled for OneAuth and MSAL Android. Once we have seen that there's no change in issues in broker. We will expand change to OneAuth. Meanwhile, I can work with OneAuth how can we track this in OneAuth telemetry. This may be non-trivial work.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -171,17 +176,19 @@ public void onReceivedSslError(final WebView view,
...
-        Logger.error(TAG + ":onReceivedSslError", errMsg, null);
-
-        // Send the result back to the calling activity
-        mCompletionCallback.onChallengeResponseReceived(
-                RawAuthorizationResult.fromException(
-                        new ClientException("Code:" + ERROR_FAILED_SSL_HANDSHAKE, error.toString())));
+        final String errMsg = "Received SSL Error during request. For more info see: " + SSL_HELP_URL + ". Error: " + error.toString();
+
+        Logger.warn(methodTag, errMsg);
+        SpanExtension.current()
```
somalaya: The span here would be AcquireTokenInteractive?

somalaya:
So, I guess we are not doing WEB_VIEW_NEW_SSL_ERROR_HANDLER_ENABLED being passed from OneAuth for opt-in or opt-out in this PR? If we are not, then let's update the PR description with the chosen option?

mohitc1:
> So, I guess we are not doing WEB_VIEW_NEW_SSL_ERROR_HANDLER_ENABLED being passed from OneAuth for opt-in or opt-out in this PR? If we are not, then let's update the PR description with the chosen option?

Updated

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
mohitc1: I expected it to be ATIInteractively but seems like it's lost when switching threads. I referred to code in `AzureActiveDirectoryWebViewClient`  where we are doing         `SpanExtension.current().setAttribute(
                AttributeName.is_sso_nonce_found_in_ests_request.name(), nonceQueryParam != null
        );`

somalaya: Yes, SpanExtension.current().setAttribute( AttributeName.is_sso_nonce_found_in_ests_request.name(), nonceQueryParam != null ); never really got captured correctly. Same is the case with other places in the code where we are not creating new spans. 
Do you want to remove this completely or add a new span?

melissaahn: Yep, when revisiting the passkey telemetry recently, I found that there isn't a way to revive a span in the context of AuthorizationActivity; you can only link it to a new span using spanContext.

mohitc1: Moved to using counter.

mohitc1: > Yep, when revisiting the passkey telemetry recently, I found that there isn't a way to revive a span in the context of AuthorizationActivity; you can only link it to a new span using spanContext.

we would need maintain span in some span store/span holder .

Updated telemetry here to use counter.

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -171,17 +184,21 @@ public void onReceivedSslError(final WebView view,
                                    final SslErrorHandler handler,
                                    final SslError error) {
         // Developer does not have option to control this for now
+        final String methodTag = TAG + ":onReceivedSslError";
         super.onReceivedSslError(view, handler, error);
```
nickbopp, line 188: I don't see 'cancel' being called anymore in the flight-off scenario.  But it seems like tests are validating that it is.  Is the job getting done `super` method invocation?

nickbopp (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
mohitc1: yes, super calls it hence removed.

melissaahn (approved):

PR #2693, opened 2025-06-27
Add telemetry for Auth Ux Javascript and Number Match, Fixes AB#3308132

fadidurah:
Add telemetry for Auth Ux Javascript and Number Match. Also have a broker PR where we copy the attribute name.

[AB#3308132](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3308132)


siddhijain:
Could we wait until I merge the other PR with the ContentProvider changes as both the PRs are changing the same files?

PR #2704, opened 2025-07-02
[Common] Update Broker ATS for Resource Accounts, Fixes AB#3230426

mohitc1:
Don't return UiRequiredException when the acquire token silent is for resource account. It's not expected to start interactive request for an RA account, so instead we return ServiceException.

On top of it, there will be broker side PR to handle interrupt internally if ATS fails with relevant errors. This will be separate change.

Fixes [AB#3230426](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3230426)


common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
```diff
@@ -222,11 +222,25 @@ private static boolean shouldBeConvertedToUiRequiredException(final String oAuth
...
+     */
+    static ServiceException getExceptionFromTokenErrorResponse(
+            @NonNull final TokenErrorResponse errorResponse,
+            final boolean allowConvertToUiRequiredException
+    ) {
 
         final ServiceException outErr;
 
-        if (shouldBeConvertedToUiRequiredException(errorResponse.getError())) {
+        if (allowConvertToUiRequiredException && shouldBeConvertedToUiRequiredException(errorResponse.getError())) {
```
siddhijain, line 243: Where will the value of the parameter be set to false for Silent token requests?

mohitc1: It's false by default. It will be set in brokered flow if the request is for resource account. That will be separate PR.

somalaya, line 243: Can you capture why this needs to be false for RA accounts as well in this or broker PR?

somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/ExceptionAdapter.java
mohitc1: It's there at caller level. At this method it's just flag (theoretically applies to other scenarios if need be), which is documented in params.
I am updating caller with better doc

siddhijain (approved):

PR #2712, opened 2025-07-17
Fix webview overlap with keyboard on SDK 35

Alexeyca:
This fix the issue that arises after updating the SDK to version 35.
WebView overlaps with the keyboard, preventing users from typing their credentials.
screen_with_keyboard


Alexeyca:

PR #2719, opened 2025-07-28
Adding NoSuchMethodError to catch blocks

melissaahn:
### Summary
The UIAutomation tests failed due to uncaught NoSuchMethodErrors. The methods I added as a part of my Baggage PRs do have catch blocks meant for this purpose, but they were catching a generic exception (which NoSuchMethodError does not fall into).

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/BaggageExtension.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/BaggageExtension.java
@@ -55,7 +55,7 @@ public class BaggageExtension {
                 return SpanExtension.NoopScope.INSTANCE;
             }
             return baggage.storeInContext(Context.current()).makeCurrent();
-        } catch (final Exception e) {
+        } catch (final Exception | NoSuchMethodError e) {
             Logger.error(TAG + ":makeBaggageCurrent", "Failed to make baggage current", e);
             return SpanExtension.NoopScope.INSTANCE;
         }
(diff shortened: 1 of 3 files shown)


siddhijain (approved):

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/BaggageExtension.java
```diff
@@ -55,7 +55,7 @@ public static Scope makeBaggageCurrent(final Baggage baggage) {
                 return SpanExtension.NoopScope.INSTANCE;
             }
             return baggage.storeInContext(Context.current()).makeCurrent();
-        } catch (final Exception e) {
+        } catch (final Exception | NoSuchMethodError e) {
```
shahzaibj, line 58: If you captured generic `Throwable` that would've covered both Exceptions and Errors

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/BaggageExtension.java
melissaahn: So Copilot (see some of the comments above) mentioned this as well, and I changed it, but then Copilot gave conflicting advice and said that using Throwable isn't advisable because it could catch Errors which shouldn't be caught? So I decided to go with my original route, since both options kinda have similar pros and cons.

PR #2722, opened 2025-07-29
Allow Silent/Invisible authorization flow for Copilot SWAG flow, Fixes AB#3330822

mohitc1:
This is per Copilot's requirement for SWAG (Sign in with Apple/Google flow). 

Copilot currently has its own auth stack (not MSA) to support SWAG flow. Now, they are migrating to MSA. To allow migration of existing copilot consumer account to MSA, the protocol requires them to pass an assertion in header in /authorize call. The assertion is processed by MSA (with help of Copilot backend) to sign-up/sign-in account into MSA. The flow is supposed to run without required active user interaction as assertion should help migrate the account.

The requirement is to make the flow appear fully silent by displaying no UI. Currently the UI elements are activity, fragment, webview and MSA UX inside the webview. There's a new extra query param proposed display=none, which signals MSA to not show any UI or interrupt user for interaction. On client side also, in such case we need hide any UI, like WebView. Since, the WebView itself is hosted in Fragment, in Activity, need to hide those as well. The flow should still run in the WebView, just does not appear to user, so these UIs need to be invisible.

The change allows OneAuth to specify via `AuthorizationActivityParameters` (this is where OneAuth/MSALCPP interacts with Android Common) if silent flow is required. OneAuth can specify that in case when display=none.

Changes
1. Added `SilentWebViewAuthorizationFragment` and `SilentAuthorizationActivity` which hide all UI and appear transparent. Correponding theme added in Common's AndroidManifest.xml
2. Exposed new param `webViewEnableSilentAuthorizationFlowTimeOutMs` which indicates silent flow is required and also timeout amount for which the flow with run. The timeout saves in cases, if the flow takes too long or if there's an unexpected interrupt in the flow, which won't be visible to the user.
3. In `AuthorizationActivityFactory`, check the param is passed, if so instead of creating `AuthorizationActivity` create `SilentAuthorizationActivity`. This is only applicable to WebView flows.
4. SilentWebViewAuthorizationFragment has logic of executing the flow and mostly derives from `WebViewAuthorizationFragment` with addition of hiding the view and handling timeout. 
5. There's separate `RawAuthorizationResult` sent back in case of timeout, to notify OneAuth/App of the failure.

Fixed [AB#3330822](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3330822)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SilentAuthorizationActivity.kt
```diff
@@ -0,0 +1,51 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.providers.oauth2
+
+import android.view.View
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import androidx.fragment.app.Fragment
+
+
```
melissaahn: Would be good to have description here similar to SilentWebAuthorizationFragment

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityFactoryTest.java
```diff
@@ -180,4 +185,76 @@ public void testGetAuthorizationActivityIntentWithGoogleCredential() {
...
+        assertEquals(sourceLibraryName, resultIntent.getStringExtra(PRODUCT));
+        assertEquals(sourceLibraryVersion, resultIntent.getStringExtra(VERSION));
+        assertEquals(requestUrl, resultIntent.getStringExtra(REQUEST_URL));
+        assertEquals(10000L, resultIntent.getLongExtra(WEB_VIEW_SILENT_AUTHORIZATION_FLOW_TIMEOUT, 0));
+    }
+
+    @Test
+    public void testGetAuthorizationFragmentFromStartIntentWebView() {
+        // Create intent with silent flow enabled
+        final Intent silentFlowIntent = new Intent();
```
melissaahn: I think this variable name and comment were copied?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SilentWebViewAuthorizationFragment.kt
```diff
@@ -0,0 +1,92 @@
...
+ * and notifying user. It is equivalent to the silent flow. Such flow is non-ideal but may be required by protocol.
+ * Current requirement is for SWAG (Sign-in-with Apple/Google) flow for Copilot, to allow them to silently migrate
+ * users from their stack to MSA.
+ */
+class SilentWebViewAuthorizationFragment : WebViewAuthorizationFragment() {
+
+    companion object {
+        private const val TAG = "SilentWebViewAuthorizationFragment"
+        private const val DEFAULT_WEB_VIEW_SILENT_AUTHORIZATION_FLOW_TIME_OUT = 5000L
+        private const val MIN_WEB_VIEW_SILENT_AUTHORIZATION_FLOW_TIMEOUT = 1000L
```
melissaahn, line 49: Supposing something goes wrong on the MSA side and the user can't be silently signed in, will MSA immediately send back an error, or is there any chance in which we just time out?

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SilentWebViewAuthorizationFragment.kt
mohitc1: yes, MSA flow should return an error, which would be part of RawAuthorizationResult. In case it is not, then timeout helps break the flow. Timeout is exception case indicating a bug.

rpdome (approved):

PR #2726, opened 2025-08-05
Fix for enrollment id not sent to /token from local controller

shahzaibj:
Fix for ICM: https://portal.microsofticm.com/imp/v5/incidents/details/663024585/summary

What happens is follows:

- MAM compliant app makes ATS call to Broker
- Broker doesn't have PRT for account so fallbacks to local RT (BrokerLocalController)
- Broker correctly queries MAM enrollment id from CP content providers and populates it on CommandParameters
- Now request goes through BrokerLocalController which apparently uses BaseController (as opposed to AbstractBrokerController) for creating token request to renew AT with RT.
- BaseController ignores the MAM enrollment id from CommandParameters and doesn't attach it to the request.
- ESTS throws IntuneAppProtectionPolicyRequired error because it doesn't see enrollment ID in request

Had BrokerLocalController actually relied on its immediate parent class i.e. AbstractBrokerLocalController this would've worked but that's not the case.

Changing the above is a big refactor.

To fix the issue, I'm just attaching MAM enrollment ID (if present on CommandParameters) to the request in BaseController.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
@@ -872,6 +872,14 @@ public abstract class BaseController {
                 ((MicrosoftTokenRequest) refreshTokenRequest).setPKeyAuthHeaderAllowed(
                         ((BrokerSilentTokenCommandParameters) parameters).isPKeyAuthHeaderAllowed()
                 );
+
+                // set enrollment id if available
+                // This is used to support MAM scenarios where the broker needs to send the enrollment id to the service
+                // For BrokerLocalController, PRT is not used but RT may have device claim so MAM enrollment is still possible
+                // as long as the broker sends the enrollment id.
+                if (!StringUtil.isNullOrEmpty(parameters.getMamEnrollmentId())) {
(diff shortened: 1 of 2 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/BaseController.java
```diff
@@ -872,6 +872,14 @@ protected TokenResult performSilentTokenRequest(
                 ((MicrosoftTokenRequest) refreshTokenRequest).setPKeyAuthHeaderAllowed(
                         ((BrokerSilentTokenCommandParameters) parameters).isPKeyAuthHeaderAllowed()
                 );
+
+                // set enrollment id if available
+                // This is used to support MAM scenarios where the broker needs to send the enrollment id to the service
+                // For BrokerLocalController, PRT is not used but RT may have device claim so MAM enrollment is still possible
```
mohitc1: >MAM enrollment [](http://example.com/codeflow?start=94&length=14)

nit: satisfying MAM-CA

mohitc1 (approved):
:shipit:

mohitc1:
Update Changelog?

melissaahn (approved):

PR #2730, opened 2025-08-07
Remove System Gesture Inset

rpdome:
Removing the system gestures inset 
https://developer.android.com/develop/ui/views/layout/edge-to-edge#system-gesture-insets


nickbopp (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/DualScreenActivity.java
```diff
@@ -88,10 +88,9 @@ private void initializeContentView(){
                     // Set the padding of the view to the insets of system bars, display cutout, system gestures, and Input (keyboards).
```
mohitc1: >system gestures, [](http://example.com/codeflow?start=97&length=16)

nit: remove

mohitc1 (approved):
:shipit:

PR #2731, opened 2025-08-08
Handle BadTokenException gracefully in CBA dialogs, Fixes AB#3346402

melissaahn:
### Summary
Fixes [AB#3346402](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3346402)

If activities are destroyed or in the process of being destroyed while a CBA dialog is going to pop up, a BadTokenException will get thrown, and this is currently not being handled by the CBA logic. This PR adds try/catch clauses to the "show" methods of the DialogHolder, which log the exception and call the appropriate callback methods to finish out the CBA flow as a cancellation.

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/DialogHolderTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/DialogHolderTest.java
@@ -34,7 +34,10 @@ import org.robolectric.annotation.Config;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import androidx.annotation.NonNull;
@@ -57,7 +60,8 @@ public class DialogHolderTest {
 
     @Test
     @Config(shadows={ShadowUserChoiceDialog.class})
-    public void testExceptionThrownFromUserChoiceDialog() {
+    public void testExceptionThrownFromUserChoiceDialog() throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(1);
         mDialogHolder.showUserChoiceDialog(
                 new UserChoiceDialog.PositiveButtonListener() {
                     @Override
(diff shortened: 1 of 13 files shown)


common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/ShadowSmartcardPinDialog.java
```diff
@@ -0,0 +1,48 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.internal.ui.webview.certbasedauth;
+
+import android.view.WindowManager;
```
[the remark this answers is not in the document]
melissaahn: Not correct; we can see that WindowManager is used in the show() method

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/DialogHolderTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @Test
+    @Config(shadows={ShadowSmartcardPromptDialog.class})
+    public void testExceptionThrownFromSmartcardPromptDialog() {
+        mDialogHolder.showSmartcardPromptDialog(
+                new ICancelCbaCallback() {
+                    @Override
+                    public void onCancel() {
+                        // If we get here, the test has passed.
```
fadidurah, line 163: Not sure if necessary, but we can consider having a latch after the .showSmartcardPromptDialog, that is unlocked inside `onCancel()`. That way, test will not pass unless `onCancel()` is called. I think you can set a timeout too so test doesn't run forever if it doesn't hit the callback

fadidurah (approved):

somalaya (approved):

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/certbasedauth/DialogHolderTest.java
melissaahn: Great suggestion; implemented in the latest commit.

fadidurah (approved):

PR #2732, opened 2025-08-08
Handling https scheme for device enrollment link, Fixes AB#3344894

somalaya:
Issue : In my testing, I noticed that OneAuth team started sending a config for a confidential client and that there is a corner case for device enrollment url's scheme on eSTS side. Our contract with eSTS has been to send "browser://" scheme for device enrollment url redirect and we enable webcp flow once we notice the url is of browser scheme and has device enrollment parameters.
 
But eSTS's corner case is that when the calling app is a web app, they pass https:// scheme instead. My logic to handle webcp flow is not triggered in this case and breaks the user's sign in.
Although this is a corner case scenario, we may start seeing issues if consumers of OneAuth or MSAL register themselves as webapp/confidential clients unknowingly. 

Link to ests code where they replace browser scheme with https in specific cases : https://msazure.visualstudio.com/One/_git/ESTS-Main?path=/src/Product/Microsoft.AzureAD.ESTS/Sts/ConditionalAccess/DevicePolicyError.cs&version=GBmaster&_a=contents&line=497&lineStyle=plain&lineEnd=503&lineStartColumn=1&lineEndColumn=10
 
My fix : Is a small change to start supporting webcp flow with https:// scheme as well. 

Fixes [AB#3344894](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3344894)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1283,6 +1283,11 @@ public final class AuthenticationConstants {
         public static final String BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "&ismdmurl=1";
 
         /**
+         * The scheme for HTTPS URLs.
+         */
+        public static final String HTTPS_SCHEME = "https://";
+
+        /**
          * Activity name to launch company portal.
          */
         public static final String COMPANY_PORTAL_APP_LAUNCH_ACTIVITY_NAME = Broker.COMPANY_PORTAL_APP_PACKAGE_NAME + ".views.SplashActivity";
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1285,7 +1285,7 @@ public final class AuthenticationConstants {
         /**
          * The scheme for HTTPS URLs.
          */
-        public static final String HTTPS_SCHEME = "https://";
+        public static final String HTTPS_SCHEME = "https";
 
         /**
          * Activity name to launch company portal.
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -586,6 +586,11 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
         }
     }
 
+    /**
+     * Processed device CA requests detected in the web flow.
+     * @param view The {@link WebView} instance in which the request originated.
+     * @param url  The URL representing the device CA request.
+     */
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -343,6 +343,9 @@ else if (isRedirectUrl(formattedURL)) {
                 processWebCpEnrollmentUrl(view, url);
             } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpAuthorizeUrl(url)) {
                 processWebCpAuthorize(view, url);
+            } else if (isDeviceCaRequest(url) && isHttpsScheme(url)) {
```
rpdome, line 346: can't we just use isDeviceCaRequest() here and remove the if block in processWebsiteRequest()?

 = isDeviceCaRequest will be executed if ismdmurl exists, doesn't matter if it's https or browser://

moumighosh (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Not sure if it is being used for other purposes. I do not want to touch it right now :P

I will explore if it is safe to do this but won't do it as part of hotfix.

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1282,6 +1282,11 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER = "&ismdmurl=1";
 
+        /**
+         * The scheme for HTTPS URLs.
+         */
+        public static final String HTTPS_SCHEME = "https://";
```
shahzaibj: Do we need the `://` as part of the scheme itself?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
shahzaibj, line 346: What if it's deviceCaRequest true but NOT https scheme? That's the general case, right? Where is that getting handled now?

somalaya: It is already getting handled in if (isWebsiteRequestUrl) check

somalaya: Even before we reach the if condition I added
image

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
somalaya: I guess not, changed.

shahzaibj (approved):

PR #2733, opened 2025-08-10
Securing handling added for webapps with webcp flight, Fixes AB#3344894

somalaya:
In this PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2732, I added a new check to detect device CA flow and follow our current pattern of handling the device CA requests. But I did not secure it with a flight. 

There are 2 ways of adding a flight for this
1. Add a new flight that is independent of webcp feature 
2. Depend on webcp feature only. 
I went with 2nd option, because I only really need these changes when webcp flight is ON so that users are not blocked by new flow. So, I am securing it with webcp flight.

Fixes [AB#3344894](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3344894)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -343,7 +343,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
                 processWebCpEnrollmentUrl(view, url);
             } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpAuthorizeUrl(url)) {
                 processWebCpAuthorize(view, url);
-            } else if (isDeviceCaRequest(url) && isHttpsScheme(url)) {
+            } else if (isDeviceCaRequest(url) && isHttpsScheme(url) && isWebCpInWebviewFeatureEnabled(url)) {
                  // Special handling for device CA requests due to a corner case in eSTS for webapps/confidential clients, which should be handled by the WebView.
                 processDeviceCaRequest(view, url);
             } else {
@@ -358,6 +358,10 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             view.stopLoading();
         }
         return true;
+    }
+
+    private boolean shouldHandleWebAppCaRequest(@NonNull final String url) {
+        return isWebCpInWebviewFeatureEnabled(url);
     }
 
     private boolean isUriSSLProtected(@NonNull final String url) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -343,7 +343,7 @@ else if (isRedirectUrl(formattedURL)) {
                 processWebCpEnrollmentUrl(view, url);
             } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpAuthorizeUrl(url)) {
                 processWebCpAuthorize(view, url);
-            } else if (isDeviceCaRequest(url) && isHttpsScheme(url)) {
+            } else if (isDeviceCaRequest(url) && isHttpsScheme(url) && isWebCpInWebviewFeatureEnabled(url)) {
```
shahzaibj, line 346: Is this an existing method?

somalaya: Yes an existing method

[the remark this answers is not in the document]
somalaya: mIsWebCpInWebViewFeatureEnabled is not completely initialized with the flight value unless we invoke isWebCpInWebviewFeatureEnabled(url) method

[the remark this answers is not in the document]
somalaya: Hence cannot use the value directly

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
mohitc1, line 346: >isDeviceCaRequest(url) && isHttpsScheme(url) [](http://example.com/codeflow?start=23&length=44)

Does this not apply to pre-WebCP flows (https + device ca)?

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: As discussed offline, it does apply to pre-webcp flows. But I only found that the users get stuck with webcp flow enabled. In pre-webcp, they are good! So, I did not want to touch this scenario for existing customers just yet (at-least not as part of a hotfix).

PR #2737, opened 2025-08-15
Fix dualscreenactivity darkmode

mohitc1:
Issue:
Color constant value "white" is defined in both Android Common and Auth App
In Common, this is defined as FFFFFF which is real white.

In Auth App, defined as [background_1](https://msazure.visualstudio.com/One/_git/AD-MFA-phonefactor-phoneApp-android?path=/PhoneFactor/CommonUiLibrary/src/main/res/values/colors.xml&version=GBworking&line=5&lineEnd=6&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), which itself it defined as FFFFFF (white) in day mode - light and 000000 in night mode. 000000 is actually black.

Due to same names, the final APK uses the one from auth app, and effectively makes color of DualScreenActivity black (000000) in dark theme/night mode.

Fix
Change the constant name in Android common. Named it with prefix library package suffix com_microsoft_identity_common, to avoid any conflicts.
Going forward we should name resources with namespace prefix to avoid resource merging issues. (Perhaps rename existing ones)

Also change them for DualScreenActivity from Base.Theme.AppCompat.Light to Theme.AppCompat.Light (recommended)

Auth App should also address this: Using "white" to indirectly define both white and black, is misleading and causes confusion.

### DIFF
--- a/common/src/main/res/values/styles.xml
+++ b/common/src/main/res/values/styles.xml
@@ -60,6 +60,6 @@
     <!-- Theme for DualScreenActivity.
         Force set to a light theme (to status and navigation bars) since broker/common activities always have white background. -->
     <style name="DualScreenActivityTheme" parent="Theme.AppCompat.Light">
-        <item name="android:background">@color/white</item>
+        <item name="android:background">@color/com_microsoft_identity_common_white</item>
     </style>
 </resources>
--- a/common/src/main/res/values/styles.xml
+++ b/common/src/main/res/values/styles.xml
@@ -59,7 +59,7 @@
 
     <!-- Theme for DualScreenActivity.
         Force set to a light theme (to status and navigation bars) since broker/common activities always have white background. -->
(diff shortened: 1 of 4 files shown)


rpdome (approved):

siddhijain (approved):

fadidurah (approved):

common/src/main/res/values/styles.xml
```diff
@@ -59,7 +59,7 @@
 
     <!-- Theme for DualScreenActivity.
         Force set to a light theme (to status and navigation bars) since broker/common activities always have white background. -->
-    <style name="DualScreenActivityTheme" parent="Base.Theme.AppCompat.Light">
```
fadidurah, line 62: why removing base?

mohitc1: replaced with Theme.AppCompat.Light, recommended one.

PR #2742, opened 2025-08-22
Move ests telemetry related code behind a feature flag AB#3352490, Fixes AB#3352490

siddhijain:
Emitting eSTS telemetry is an expensive operation that impacts the performance of calls. Therefore, we are placing it behind a feature flag to avoid executing telemetry-related operations. [AB#3352490](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3352490)

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
@@ -57,31 +57,4 @@ public class TelemetryUtils {
             return s;
         }
     }
-
-    /**
-     * Functional interface for executing ests telemetry-related code.
-     */
-    public interface EstsAction {
-        void execute();
-    }
-
-    /**
-     * Checks if ESTS telemetry related code should be executed based on feature flag, and executes
-     * the provided action.
-     *
-     * @param action The action to execute if ESTS telemetry should be emitted
-     */
-    public static void executeIfEstsTelemetryEnabled(@NonNull final EstsAction action) {
-        final String methodTag = TAG + ":executeIfEstsTelemetryEnabled";
-        final boolean shouldSkipEstsTelemetry = CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_ESTS_TELEMETRY);
-
-        if (!shouldSkipEstsTelemetry) {
-            Logger.info(methodTag, "Executing ESTS telemetry action");
-            action.execute();
-            SpanExtension.current().setAttribute(AttributeName.skipped_ests_telemetry.name(), false);
-        } else {
-            Logger.info(methodTag, "Skipping ESTS telemetry action due to feature flag");
-            SpanExtension.current().setAttribute(AttributeName.skipped_ests_telemetry.name(), true);
-        }
-    }
 }
--- a/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
@@ -22,11 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.eststelemetry;
 
-import com.microsoft.identity.common.java.flighting.CommonFlight;
-import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
-import com.microsoft.identity.common.java.logging.Logger;
-import com.microsoft.identity.common.java.opentelemetry.AttributeName;
-import com.microsoft.identity.common.java.opentelemetry.SpanExtension;
 import com.microsoft.identity.common.java.telemetry.TelemetryEventStrings;
 import com.microsoft.identity.common.java.util.StringUtil;
 
@@ -34,8 +29,6 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import lombok.NonNull;
 
 public class TelemetryUtils {
-
-    private static final String TAG = TelemetryUtils.class.getSimpleName();
 
     static boolean getBooleanFromString(@Nullable final String val) {
         return val != null && val.equals(SchemaConstants.Value.TRUE);
--- a/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
@@ -22,6 +22,11 @@
 // THE SOFTWARE.
 package com.microsoft.identity.common.java.eststelemetry;
 
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
+import com.microsoft.identity.common.java.logging.Logger;
+import com.microsoft.identity.common.java.opentelemetry.AttributeName;
+import com.microsoft.identity.common.java.opentelemetry.SpanExtension;
 import com.microsoft.identity.common.java.telemetry.TelemetryEventStrings;
 import com.microsoft.identity.common.java.util.StringUtil;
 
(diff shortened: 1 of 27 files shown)


common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
```diff
@@ -50,4 +57,31 @@ static String getSchemaCompliantString(final String s) {
...
+        final boolean shouldSkipEstsTelemetry = CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_ESTS_TELEMETRY);
+
+        if (!shouldSkipEstsTelemetry) {
+            Logger.info(methodTag, "Executing ESTS telemetry action");
+            action.execute();
+            SpanExtension.current().setAttribute(AttributeName.skipped_ests_telemetry.name(), false);
+        } else {
+            Logger.info(methodTag, "Skipping ESTS telemetry action due to feature flag");
+            SpanExtension.current().setAttribute(AttributeName.skipped_ests_telemetry.name(), true);
+        }
```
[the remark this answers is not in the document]
siddhijain: I would like to stick to the current implementation as the suggested change does not create clarity on why a feature flag value is used instead of the actual value of whether ests telemetry was skipped.

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
```diff
@@ -50,4 +57,31 @@ static String getSchemaCompliantString(final String s) {
...
+    public static void executeIfEstsTelemetryEnabled(@NonNull final EstsAction action) {
+        final String methodTag = TAG + ":executeIfEstsTelemetryEnabled";
+        final boolean shouldSkipEstsTelemetry = CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_ESTS_TELEMETRY);
+
+        if (!shouldSkipEstsTelemetry) {
+            Logger.info(methodTag, "Executing ESTS telemetry action");
+            action.execute();
+            SpanExtension.current().setAttribute(AttributeName.skipped_ests_telemetry.name(), false);
+        } else {
+            Logger.info(methodTag, "Skipping ESTS telemetry action due to feature flag");
```
fadidurah: Logging might be excessive here to be honest, since we already tracking in telemetry. I think it's safe to keep this for now fror debugging but once this feature is fully rolled out, I think we should remove the log statements to not spam the log files. IF we also notice that these logs are not showing up a lot in our log files, we can keep this.

siddhijain: sounds good. Once the feature flag is rolled out to 100% of the users, we can safely get rid of the class as well :)

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
```diff
@@ -50,4 +57,31 @@ static String getSchemaCompliantString(final String s) {
...
+        void execute();
+    }
+
+    /**
+     * Checks if ESTS telemetry related code should be executed based on feature flag, and executes
+     * the provided action.
+     *
+     * @param action The action to execute if ESTS telemetry should be emitted
+     */
+    public static void executeIfEstsTelemetryEnabled(@NonNull final EstsAction action) {
```
mohitc1: >EstsAction [](http://example.com/codeflow?start=68&length=10)

Runnable should be enough

mohitc1: >executeIfEstsTelemetryEnabled [](http://example.com/codeflow?start=23&length=29)

This ill be called multiple times, so we would emit same value in Span multiple times, only latest one persisting.

mohitc1:
How about we make EstsTelemetry.getInstance() return NoOpEstsTelemetry when flight is enabled? Only change needed would be EstsTelemetry class, instead of updating every call.

common4j/src/main/com/microsoft/identity/common/java/eststelemetry/TelemetryUtils.java
siddhijain: Good point but closing the comment since I moved to using Noop telemetry.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
```diff
@@ -419,7 +421,9 @@ class NativeAuthRequestProvider(private val config: NativeAuthOAuth2Configuratio
         headers[AuthenticationConstants.SdkPlatformFields.PRODUCT] = LibraryInfoHelper.getLibraryName()
         headers[AuthenticationConstants.SdkPlatformFields.VERSION] = LibraryInfoHelper.getLibraryVersion()
         headers.putAll(Device.getPlatformIdParameters())
-        headers.putAll(EstsTelemetry.getInstance().telemetryHeaders)
+        if (!CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_ESTS_TELEMETRY)) {
+            headers.putAll(EstsTelemetry.getInstance().telemetryHeaders)
+        }
```
mohitc1: do we still need this?

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
```diff
@@ -238,7 +240,9 @@ protected HttpResponse performTokenRequest(final GenericTokenRequest request) th
         headers.putAll(Device.getPlatformIdParameters());
         headers.put(AuthenticationConstants.SdkPlatformFields.PRODUCT, sourceLibraryName);
         headers.put(AuthenticationConstants.SdkPlatformFields.VERSION, sourceLibraryVersion);
-        headers.putAll(EstsTelemetry.getInstance().getTelemetryHeaders());
+        if (!CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_ESTS_TELEMETRY)) {
+            headers.putAll(EstsTelemetry.getInstance().getTelemetryHeaders());
+        }
```
mohitc1: do we still need this?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/NativeAuthRequestProvider.kt
siddhijain: No - I will remove it.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
```diff
@@ -33,6 +33,8 @@
 import com.microsoft.identity.common.java.dto.IAccountRecord;
 import com.microsoft.identity.common.java.eststelemetry.EstsTelemetry;
 import com.microsoft.identity.common.java.exception.ClientException;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
```
mohitc1: nit: imports

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
siddhijain: Nopes - removed in the latest commit.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OAuth2Strategy.java
siddhijain: yeah - removed.

PR #2744, opened 2025-08-25
 Fix for SDL violation in device pop scenarios, Fixes AB#3284510

somalaya:
- 1 of the SDL assessments has a requirement that we should avoid using multiple crypto purposes with the same key.
- There are multiple places where we are using multiple purposes and device pop scenario is one of them.
- In Device pop scenario, we do not use encrypt/decrypt methods anywhere in our code. Only signing is used. Hence removing the ununsed purposes encrypt, decrypt from the key gen spec. 
- Also, encrypt and decrypt methods added in [IDevicePopManager.java](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/compare/somalaya/devicePopSdlFix?expand=1#diff-2ca2fe49df086f6ee051aeeedcc5bf3bfd53b1b3390e2a8d1143af2cfc74da7e) are probably for maintaining all the crypto operations but I have deprecated them as they are not getting used.
- Added telemetry in mintSignedHttpRequest method. 

Note : The flight to remove encryption paddings in initialize method of AndroidDevicePopManager is not going to be enabled until we observe telemetry added in this PR first. Idea is that the telemetry should prove that encrypt/decrypt methods are not getting used. Only then, we can enable the flight! So, this is going to be enabled in Jan/Feb next year.

Fixes [AB#3284510](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3284510)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
@@ -357,7 +357,7 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
                             final boolean enableImport,
                             final boolean trySetAttestationChallenge) throws InvalidAlgorithmParameterException {
         final boolean unnecessaryCryptoPurposesDisabled =
-                CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_UNNECESSARY_CRYPTO_PURPOSES);
+                CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_UNNECESSARY_CRYPTO_PURPOSES_FROM_DEVICE_POP_MANAGER);
 
         int purposes = KeyProperties.PURPOSE_SIGN | KeyProperties.PURPOSE_VERIFY;
         if (!unnecessaryCryptoPurposesDisabled) {
@@ -383,7 +383,8 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
                               final int keySize,
                               final boolean useStrongbox,
                               final boolean trySetAttestationChallenge,
-                              final int purposes) throws InvalidAlgorithmParameterException {
+                              final int purposes,
+                              final boolean unnecessaryCryptoPurposesDisabled) throws InvalidAlgorithmParameterException {
         KeyGenParameterSpec.Builder builder;
 
         builder = new KeyGenParameterSpec.Builder(
@@ -398,6 +399,13 @@ public class AndroidDevicePopManager extends AbstractDevicePopManager {
                         KeyProperties.DIGEST_SHA1,
                         KeyProperties.DIGEST_SHA256
                 );
+
+        if (!unnecessaryCryptoPurposesDisabled) {
+            builder.setEncryptionPaddings(
+                    KeyProperties.ENCRYPTION_PADDING_RSA_OAEP,
+                    KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1
+            );
+        }
 
         if (trySetAttestationChallenge && Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             builder = setAttestationChallenge(builder);
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
@@ -40,6 +40,8 @@ import androidx.annotation.RequiresApi;
 
 import com.microsoft.identity.common.java.crypto.IKeyStoreKeyManager;
 import com.microsoft.identity.common.java.crypto.SecureHardwareState;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
 import com.microsoft.identity.common.java.platform.AbstractDevicePopManager;
 import com.microsoft.identity.common.logging.Logger;
 import com.nimbusds.jose.crypto.impl.RSAKeyUtils;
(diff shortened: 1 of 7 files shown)


common4j/src/main/com/microsoft/identity/common/java/platform/AbstractDevicePopManager.java
```diff
@@ -877,14 +883,23 @@ public String mintSignedAccessToken(@Nullable String httpMethod,
...
+        final Span span = OTelUtility.createSpan(SpanName.DevicePopMintSignedAccessToken.name());
+        try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
+            return mintSignedHttpRequestInternal(
+                    httpMethod,
+                    timestamp,
+                    requestUrl,
+                    accessToken,
+                    nonce,
+                    clientClaims
+            );
```
mohitc1, line 895: span success?

mohitc1:
> 

Is this needed if purpose does not have encrypt/decrypt/wrap?

If not, then this method is identical to initialize23, since caller decides if wrap key is to be set or not.

---
Refers to: common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java:470 in 42bf650. [](commit_id = 42bf650db7c308f6fd344527c16fc0de82d2b93c, deletion_comment = True)

somalaya:
> > 
> 
> Is this needed if purpose does not have encrypt/decrypt/wrap?
> 
> If not, then this method is identical to initialize23, since caller decides if wrap key is to be set or not.
> 
> Refers to: common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java:470 in [42bf650](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/42bf650db7c308f6fd344527c16fc0de82d2b93c). [](commit_id = [42bf650](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/42bf650db7c308f6fd344527c16fc0de82d2b93c), deletion_comment = True)

If you meant the encryption padding, then yes it is not needed when purpose is not encrypt/decrypt/wrap. I have skipped adding this padding when the flight is ON in 5c852dbfbd0539b64aa05a6a2bf21fb239ccc6ed . Once the flight is completely ON, we can remove initialize23 and initalize28 and just add a check for Build version to use strongbox in API >=28.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -385,9 +397,6 @@ private void initialize23(@androidx.annotation.NonNull final KeyPairGenerator ke
                         KeyProperties.DIGEST_NONE,
                         KeyProperties.DIGEST_SHA1,
                         KeyProperties.DIGEST_SHA256
-                ).setEncryptionPaddings(
-                        KeyProperties.ENCRYPTION_PADDING_RSA_OAEP,
-                        KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1
```
mohitc1, line 390: setEncryptionPaddings is removed unconditionally.

If purposes has encrypt/decrypt, shouldn't this be kept?

Instead or purposes you can pass the use unnecessaryCryptoPurposesDisabled flag to these initialize* methods, and let each method decide its purposes.

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -155,7 +155,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable WebView security settings to prevent unauthorized access.
      */
-    ENABLE_WEBVIEW_SECURITY_SETTINGS("EnableWebViewSecuritySettings", false);
+    ENABLE_WEBVIEW_SECURITY_SETTINGS("EnableWebViewSecuritySettings", false),
+
+    /**
+     * Flight to disable the unnecessary crypto operation purposes in device pop manager like encrypt, decrypt and wrap.
+     */
+    DISABLE_UNNECESSARY_CRYPTO_PURPOSES("DisableUnnecessaryCryptoPurposes", false);
```
mohitc1: >DISABLE_UNNECESSARY_CRYPTO_PURPOSES [](http://example.com/codeflow?start=4&length=35)

DISABLE_UNNECESSARY_CRYPTO_PURPOSES_FROM_DEVICE_POP_MANAGER ?

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
somalaya: Are you referring to this commit? 
image
If yes, it is removed unconditionally but added immediately conditionally based on this boolean unnecessaryCryptoPurposesDisabled

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -348,18 +350,30 @@ private KeyPairGenerator getInitializedRsaKeyPairGenerator(@androidx.annotation.
...
+        if (!unnecessaryCryptoPurposesDisabled) {
+            // If the flight is not enabled, we can use the key for encryption and decryption.
+            purposes |= KeyProperties.PURPOSE_ENCRYPT | KeyProperties.PURPOSE_DECRYPT;
+        }
+
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
             initializePre23(context, keyPairGenerator, keySize);
         } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) {
-            initialize23(keyPairGenerator, keySize, useStrongbox, trySetAttestationChallenge);
+            initialize23(keyPairGenerator, keySize, useStrongbox, trySetAttestationChallenge, purposes);
```
mohitc1: >purposes [](http://example.com/codeflow?start=94&length=8)

you need to pass 6th param unnecessaryCryptoPurposesDisabled :)

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -348,18 +350,30 @@ private KeyPairGenerator getInitializedRsaKeyPairGenerator(@androidx.annotation.
...
+    private void initialize(@NonNull final Context context,
+                            @NonNull final KeyPairGenerator keyPairGenerator,
                             final int keySize,
                             final boolean useStrongbox,
                             final boolean enableImport,
                             final boolean trySetAttestationChallenge) throws InvalidAlgorithmParameterException {
+        final boolean unnecessaryCryptoPurposesDisabled =
+                CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_UNNECESSARY_CRYPTO_PURPOSES_FROM_DEVICE_POP_MANAGER);
+
+        int purposes = KeyProperties.PURPOSE_SIGN | KeyProperties.PURPOSE_VERIFY;
```
p3dr0rv, line 362: Nit: Consider moving the definition of purposes inside each initializeX method — I believe it makes things a little easier to read.

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java
```diff
@@ -348,18 +350,30 @@ private KeyPairGenerator getInitializedRsaKeyPairGenerator(@androidx.annotation.
...
-    private void initialize(@androidx.annotation.NonNull final Context context,
-                            @androidx.annotation.NonNull final KeyPairGenerator keyPairGenerator,
+    private void initialize(@NonNull final Context context,
+                            @NonNull final KeyPairGenerator keyPairGenerator,
                             final int keySize,
                             final boolean useStrongbox,
                             final boolean enableImport,
                             final boolean trySetAttestationChallenge) throws InvalidAlgorithmParameterException {
+        final boolean unnecessaryCryptoPurposesDisabled =
+                CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_UNNECESSARY_CRYPTO_PURPOSES_FROM_DEVICE_POP_MANAGER);
```
p3dr0rv, line 360: nit: ask copilot for a better name :p

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -181,7 +181,12 @@ public enum CommonFlight implements IFlightConfig {
      * Flight to enable OpenID issuer validation code which validates issuer against the open id well known
      * config endpoint and only reports the failure result.
      */
-    ENABLE_OPENID_ISSUER_VALIDATION_REPORTING("EnableOpenIdIssuerValidationReporting", true);
+    ENABLE_OPENID_ISSUER_VALIDATION_REPORTING("EnableOpenIdIssuerValidationReporting", true),
+
+    /**
+     * Flight to disable the unnecessary crypto operation purposes in device pop manager like encrypt, decrypt and wrap.
+     */
+    DISABLE_UNNECESSARY_CRYPTO_PURPOSES_FROM_DEVICE_POP_MANAGER ("DisableUnnecessaryCryptoPurposes", false);
```
mohitc1: >DisableUnnecessaryCryptoPurposes [](http://example.com/codeflow?start=66&length=32)

Update this string as well?

mohitc1:
>     }

this sets it to null. Do we really need this method? if we delete this method, initialize23 and initialize28 can be simplified.

initialize23() also does not need to worry about Strongbox. If strong box is needed call initialize28

---
Refers to: common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java:424 in 9691151. [](commit_id = 9691151d60a2699eba730ea437639709da769024, deletion_comment = False)

mohitc1 (dismissed):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
somalaya: Done

somalaya:
> > ```
> > }
> > ```
> 
> this sets it to null. Do we really need this method? if we delete this method, initialize23 and initialize28 can be simplified.
> 
> initialize23() also does not need to worry about Strongbox. If strong box is needed call initialize28
> 
> Refers to: common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDevicePopManager.java:424 in [9691151](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/9691151d60a2699eba730ea437639709da769024). [](commit_id = [9691151](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/9691151d60a2699eba730ea437639709da769024), deletion_comment = False)

Coming in next PR with a different flight

p3dr0rv (approved):

mohitc1 (approved):

PR #2747, opened 2025-08-28
Add aea app to DUNA browser safelist

shahzaibj:
Amazon wants to use their own browser as CCT provider so adding this to browser safelist for DUNA.


p3dr0rv (approved):
LGTM, add unit to avoid someone break this on the future

melissaahn (approved):

PR #2749, opened 2025-08-29
Adjust Ui Automation to handle both old and new Authenticator Settings Ui, Fixes AB#3330305

fadidurah:
Authenticator is rolling out React Native settings page. This PR is to update our E2E Ui Automation validation to handle both the old Ui and the new to maintain success rate as the feature rolls out.

Using regex is a more efficient way to handle multiple possible text fields rather than a typical try-catch, we don't need to wait and see which one is available, the regex can select either one consistently.

Validation:
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1526083&view=results

[AB#3330305](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3330305)


siddhijain (approved):

mohitc1 (approved):
:shipit:

PR #2750, opened 2025-08-29
 Fix Switch browser back stack error

p3dr0rv:
## Problem
The current **DUNA** implementation launches `BrokerAuthorizationActivity` with  
`FLAG_ACTIVITY_NEW_TASK` and `FLAG_ACTIVITY_CLEAR_TASK`.  

This causes:
- Device registration failures  
- Other authentication flows failing  

The issue arises because the activity stack is cleared, preventing proper continuation of dependent flows.

---

## Solution
Introduce **`SwitchBrowserActivity`** to handle the browser switch during the DUNA authentication flow.  

This approach:
- Isolates the browser-based authentication  
- Preserves the existing activity stack for device registration and other auth processes  
- Leaves non-DUNA flows unchanged  

---

## Activity Stack Behavior

1. **Task 1:** `BrokerAuthorizationActivity` launches `SwitchBrowserActivity` in **Task 2**  
2. **Task 1:** `BrokerAuthorizationActivity` | **Task 2:** `SwitchBrowserActivity` → Chrome Custom Tabs  
3. **Task 1:** `BrokerAuthorizationActivity` | **Task 2:** `SwitchBrowserActivity` → Chrome Custom Tabs | **Task 3:** `BrokerBrowserRedirectActivity`  
4. **Task 1:** `BrokerAuthorizationActivity` | **Task 2:** `SwitchBrowserActivity` → Chrome Custom Tabs | **Task 3:** `BrokerBrowserRedirectActivity` (finishes)  

---

## Flow Overview of `SwitchBrowserActivity`

1. `WebViewAuthorizationFragment` receives a **SwitchBrowser challenge**  
2. `SwitchBrowserActivity` is launched with browser configuration parameters  
3. Launches external browser (Custom Tabs or standard browser)  
4. User completes authentication in the external browser  
5. Browser redirects to `BrokerBrowserRedirectActivity` in a new task  
6. `BrokerBrowserRedirectActivity` forwards redirect result back to `SwitchBrowserActivity` via `onNewIntent()`  
7. `SwitchBrowserActivity` passes the result to `WebViewAuthorizationFragment`  
8. `SwitchBrowserActivity` finishes and removes itself from the task stack  

---

## Security Note
- `SwitchBrowserActivity` is **not exported**  
- Can only be launched from inside the app  
- Prevents external apps from triggering unwanted browser switches


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
```diff
@@ -0,0 +1,239 @@
...
+        Logger.info(
+            methodTag,
+            "Launching switch browser request on browser: $browserPackageName, Custom Tabs supported: $browserSupportsCustomTabs"
+        )
+
+        // Launch browser based on Custom Tabs support
+        if (browserSupportsCustomTabs) {
+            // Use Custom Tabs for better user experience and security
+            Logger.info(methodTag, "Launching Custom Tabs intent for DUNA authentication")
+            val customTabsIntent = CustomTabsIntent.Builder().build().apply {
```
shahzaibj: Not using `CustomTabsManager` helper to create custom tab intent? Why?

We need to make sure the case for app link that I fixed last week is still handled appropriately. 

Referring to this: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2745/files

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
```diff
@@ -0,0 +1,239 @@
...
+    override fun onNewIntent(intent: Intent?) {
+        val methodTag = "$TAG:onNewIntent"
+        super.onNewIntent(intent)
+        // Update the activity's intent with the new intent containing the auth result
+        Logger.info(methodTag, "On new intent received.")
+        setIntent(intent)
+
+        if (intent != null) {
+            if (intent.hasExtra(PROCESS_URI)) {
+                // Handle scenario where a new browser switch request is received while one is already in progress
```
shahzaibj, line 176: What happens to existing request in this case?

p3dr0rv: we drop the existing request, and we start a new one.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
p3dr0rv: "I thought it would be simpler than using CustomTabsManager, but good point— even if I replace it, I’ll still need that flag. So, I’ll just reuse CustomTabsManager."

shahzaibj (approved):

melissaahn (approved):

PR #2753, opened 2025-08-30
Add a couple more attributes to track key store errors, Fixes AB#3327441

fadidurah:
Add a couple more attributes to track key store errors

[AB#3327441](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3327441)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -474,11 +474,11 @@ public class AndroidKeyStoreUtil {
     }
 
 
-    private static @Nullable KeyStoreException findKeyStoreException(@NonNull Throwable throwable) {
+    private static @Nullable android.security.KeyStoreException findKeyStoreException(@NonNull Throwable throwable) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             while (throwable != null) {
                 if (throwable instanceof android.security.KeyStoreException) {
-                    return (KeyStoreException) throwable;
+                    return (android.security.KeyStoreException) throwable;
                 }
                 throwable = throwable.getCause();
             }
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -89,7 +89,7 @@ public class AndroidKeyStoreUtil {
     private AndroidKeyStoreUtil() {
     }
 
-    public static final LongCounter sFailedAndroidKeyStoreUnwrapOperationCount = OTelUtility.createLongCounter(
+    private static final LongCounter sFailedAndroidKeyStoreUnwrapOperationCount = OTelUtility.createLongCounter(
             "failed_keystore_key_unwrap_operation_count",
             "Number of failed Android KeyStore unwrap operations"
     );
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -450,7 +450,7 @@ public class AndroidKeyStoreUtil {
         );
         if (exception instanceof InvalidKeyException) {
             final KeyStoreException keyStoreException = findKeyStoreException(exception);
-            String ksMessage = keyStoreException != null ? keyStoreException.getMessage() : "No Keystore Excerption Found";
+            String ksMessage = keyStoreException != null ? keyStoreException.getMessage() : "No Keystore Exception Found";
             if (ksMessage == null) {
                 ksMessage = "";
             }
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -455,11 +455,11 @@ public class AndroidKeyStoreUtil {
             if (keyStoreException != null) {
                 ksMessage = keyStoreException.getMessage();
                 if (ksMessage == null) {
-                    ksMessage = "";
+                    ksMessage = "Keystore exception found, no error message";
                 }
                 ksNumericErrorCode = getNumericErrorCodeFromKeyStoreException(keyStoreException);
             } else {
-                ksMessage = "No Keystore Exception Found";
+                ksMessage = "No keystore exception found";
                 ksNumericErrorCode = "";
             }
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -514,16 +514,20 @@ public class AndroidKeyStoreUtil {
      * @return The found KeyStoreException, or null if none was found or the API level is below 33.
      */
     private static @Nullable android.security.KeyStoreException findKeyStoreException(@NonNull Throwable throwable) {
-        // Check up to a max depth to avoid infinite loops in case of circular references
-        int count = 0;
-        while (throwable != null && count < KEYSTORE_EXCEPTION_CAUSE_CHAIN_MAX_DEPTH) {
-            if (throwable instanceof android.security.KeyStoreException) {
-                return (android.security.KeyStoreException) throwable;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            // Check up to a max depth to avoid infinite loops in case of circular references
+            int count = 0;
+            while (throwable != null && count < KEYSTORE_EXCEPTION_CAUSE_CHAIN_MAX_DEPTH) {
+                if (throwable instanceof android.security.KeyStoreException) {
+                    return (android.security.KeyStoreException) throwable;
+                }
+                throwable = throwable.getCause();
+                count++;
             }
-            throwable = throwable.getCause();
-            count++;
-        }
 
-        return null;
+            return null;
+        } else {
+            return null;
+        }
     }
 }
--- a/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
@@ -47,8 +47,6 @@ import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import java.security.spec.AlgorithmParameterSpec;
 import java.util.Locale;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.crypto.Cipher;
 import javax.crypto.IllegalBlockSizeException;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +449,18 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
                 exception
         );
         if (exception instanceof InvalidKeyException) {
+            final android.security.KeyStoreException keyStoreException = findKeyStoreException(exception);
```
fadidurah: @mohitc1 

Noticed something interesting. There are 2 Keystore exception types: java.security.KeyStoreException and android.security.KeyStoreException.

In existing code, we seem to expect java.security one, but the stack trace where we see the keystore exception for failed operation uses android.security one.

java.security.InvalidKeyException: Keystore operation failed

	at android.security.keystore2.KeyStoreCryptoOperationUtils.getInvalidKeyException(KeyStoreCryptoOperationUtils.java:128)
 .......
Caused by: android.security.KeyStoreException: -22 (internal Keystore code: -22 message: system/security/keystore2/src/security_level.rs:358: Failed to begin operation.
 
Caused by:
    0: system/security/keystore2/src/security_level.rs:868
    1: system/security/keystore2/src/utils.rs:200: Calling km_op.
    2: Error::Km(r#KEY_EXPORT_OPTIONS_INVALID)) (public error code: 13 internal Keystore code: -22)
	at android.security.KeyStore2.getKeyStoreException(KeyStore2.java:386)
	at android.security.KeyStoreSecurityLevel.createOperation(KeyStoreSecurityLevel.java:120)

	at android.security.keystore2.AndroidKeyStoreCipherSpiBase.ensureKeystoreOperationInitialized(AndroidKeyStoreCipherSpiBase.java:349)

	... 40 more
	
Was wondering if you had any context here.

fadidurah: I think if our existing handling is looking for java.security.KeyStoreException, but android OSs are throwing android.security.KeyStoreException, then we are probably mishandling some exceptions

mohitc1: No, it's both. 
final KeyStore keyStore = KeyStore.getInstance(ANDROID_KEY_STORE_TYPE); > this throws java.security.KeyStoreException

while when the keystore used is AndroidKeyStore then they can also throw android.security.KeyStoreException, which is different object.

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -464,4 +473,30 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+            return null;
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Use Regex to pull out the internal error code from the key store exception message
+     * @param message the exception message
+     * @return the internal error code, or "N/A" if it can't be found
```
mohitc1: >or "N/A" if it can't be found [](http://example.com/codeflow?start=40&length=29)

Update this. You're returning only ""

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -464,4 +473,30 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+     */
+    private static String extractInternalKeystoreCode(final String message) {
+        if (message == null) return "";
+        Pattern pattern = Pattern.compile("internal Keystore code:\\s*(-?\\d+)");
+        Matcher matcher = pattern.matcher(message);
+        return matcher.find() ? matcher.group(1) : "";
+    }
```
mohitc1: You can skip this, since you already adding entire message. We can do parsing in telemetry

Instead add getNumericErrorCode() 

https://developer.android.com/reference/android/security/KeyStoreException#getNumericErrorCode() 

This respresents 13 in

Error::Km(r#KEY_EXPORT_OPTIONS_INVALID)) (public error code: 13 internal Keystore code: -22)

and -22 = KEY_EXPORT_OPTIONS_INVALID which is part of message.

fadidurah: Parsing the message is going to be expensive in telemetry though, we will have very slow queries. a separate field is appropriate I think. Do we really need to track the whole message in telemetry? Is the internal error code enough?

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
fadidurah: Yup, missed this

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
fadidurah: I switched to using "getNumericErrorCode"

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
fadidurah: I see, so ok to leave existing implementation?

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +449,27 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+                ksNumericErrorCode = "";
+            }
+
             final Attributes attributes = Attributes.builder()
                     .put(AttributeName.keystore_operation.name(), "unwrap")
                     .put(AttributeName.error_code.name(), errCode)
                     .put(AttributeName.error_type.name(), clientException.getClass().getSimpleName())
                     .put(AttributeName.keystore_exception_stack_trace.name(), ThrowableUtil.getStackTraceAsString(clientException))
+                    .put(AttributeName.keystore_exception_message.name(), ksMessage)
+                    .put(AttributeName.keystore_internal_error_code.name(), ksNumericErrorCode)
```
mohitc1: >keystore_internal_error_code [](http://example.com/codeflow?start=39&length=28)

rename this because numeric errorcode is different from internal. internal would part of ksMessage.
e.g
intenral = -22
when numeric is 13

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
mohitc1: yes

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +449,27 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
                 exception
         );
         if (exception instanceof InvalidKeyException) {
+            final android.security.KeyStoreException keyStoreException = findKeyStoreException(exception);
```
mohitc1: >KeyStoreException [](http://example.com/codeflow?start=35&length=17)

Seems this object >= 33. in this line it used without protection and can cause NoClassDefError exception

Refactor such that all references to this android.security.Exception are API safe.

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +449,27 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+                ksNumericErrorCode = getNumericErrorCodeFromKeyStoreException(keyStoreException);
+            } else {
+                ksMessage = "No keystore exception found";
+                ksNumericErrorCode = "";
+            }
+
             final Attributes attributes = Attributes.builder()
                     .put(AttributeName.keystore_operation.name(), "unwrap")
                     .put(AttributeName.error_code.name(), errCode)
                     .put(AttributeName.error_type.name(), clientException.getClass().getSimpleName())
```
mohitc1: >clientException.getClass().getSimpleName() [](http://example.com/codeflow?start=58&length=42)

this is always ClientExeption. we rather replace it with excception.getMessage(), here exception = InvalidKeyEception. this can be helpful if InvalidKeyException is not caused by KeyStoreException. You will have some info.

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +449,27 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
                     .put(AttributeName.keystore_exception_stack_trace.name(), ThrowableUtil.getStackTraceAsString(clientException))
```
mohitc1: >clientException [](http://example.com/codeflow?start=114&length=15)

here to better would be exception not clientException

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -464,4 +482,38 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+     * @return The found KeyStoreException, or null if none was found or the API level is below 33.
+     */
+    private static @Nullable android.security.KeyStoreException findKeyStoreException(@NonNull Throwable throwable) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            while (throwable != null) {
+                if (throwable instanceof android.security.KeyStoreException) {
+                    return (android.security.KeyStoreException) throwable;
+                }
+                throwable = throwable.getCause();
+            }
```
mohitc1, line 526: I hope this never becomes infinite?

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -447,11 +454,38 @@ public static synchronized SecretKey unwrap(@NonNull final byte[] wrappedKeyBlob
...
+                    ksMessage = "No keystore exception found";
+                    errorType = "InvalidKeyException";
+                    ksNumericErrorCode = "";
+                }
+            } else {
+                ksMessage = "API Level below 33, keystore exception not available";
+                errorType = "InvalidKeyException";
+                ksNumericErrorCode = "";
+            }
+
```
mohitc1: We can separate this in to a method

addAttributesForKeystoreExxception(invalidKeyException, attributesbuilder)

mohitc1 (approved):
Left a suggestion

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
fadidurah: Yeah i dont really like how this ended up looking either, i'll make a helper method

p3dr0rv (approved):

p3dr0rv (approved):

PR #2754, opened 2025-09-03
Add query parameter for Android Release OS Version, Fixes AB#3351713

fadidurah:
This is needed for WebCP work. Intune Portal uses android release os versions, not api level versions. Need this additional query parameter to facilitate flow.

[AB#3351713](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3351713)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -82,6 +82,11 @@ public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
     }
 
     @Override
+    public @NonNull String getAndroidReleaseOs() {
+        return android.os.Build.VERSION.RELEASE;
+    }
+
+    @Override
     @NonNull
     public String getDeviceModel() {
         return Build.MODEL;
--- a/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
@@ -73,12 +73,12 @@ public class AndroidDeviceMetadata extends AbstractDeviceMetadata {
 
     @Override
     public @NonNull String getOsForMats() {
-        return android.os.Build.VERSION.RELEASE;
+        return getAndroidReleaseOs();
     }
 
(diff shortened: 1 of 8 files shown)


rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
```diff
@@ -81,6 +81,11 @@ public String getCpu() {
         return android.os.Build.VERSION.RELEASE;
     }
 
+    @Override
```
Veena11, line 84: Curious why we have three different APIs for returning android.os.Build.VERSION.RELEASE;
Could this all be replaced by single API ?

siddhijain (approved):

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
```diff
@@ -119,6 +119,12 @@ public abstract class MicrosoftAuthorizationRequest<T extends MicrosoftAuthoriza
     @SerializedName("x-client-OS")
     private final String mDiagnosticOS;
 
+    @Expose()
```
Veena11, line 122: ahh so it is a new request header , not really a query param

fadidurah: It is added as a query parameter, having it exposed in this file will add it to the list of query parameters down the line.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidDeviceMetadata.java
fadidurah: changed them to all use getAndroidReleaseOs()

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/MicrosoftAuthorizationRequest.java
Veena11: aah I see, thanks for the info

Veena11 (approved):

rpdome (approved):

PR #2756, opened 2025-09-08
[Common] Update proguard rules, Fixes AB#3340920

mohitc1:
Updated proguard rules
1. Keep annotation at AndroidBase64, since it is used by reflection
2. removed com.microsoft.identity* from keeps but keeping names to help with logging and exception. So, names are not obfuscated.
3. Updated build.gradle for jcipAnnotationVersion dependency as it provides runtime objects, to avoid any issues.
4. Added donotwarn for compile time classes (annotations)
5. Major version bump as it can require consumers to adjust their code to avoid build time/runtime issues.

Fixes [AB#3340920](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3340920)

Run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1532428&view=results

### DIFF
--- a/common/consumer-rules.pro
+++ b/common/consumer-rules.pro
@@ -50,7 +50,7 @@
 # keep everything in this package from being removed or renamed
 -keep class io.opentelemetry.** { *; }
 
-# keep names
+# keep names only to allow keeping them logs and exceptions
 -keepnames class com.microsoft.identity.** { *; }
 
 # Prevent R8 from leaving Data object members always null
@@ -65,6 +65,7 @@
   *;
 }
 
+# Runtime annotations
 -keep class net.jcip.annotations.GuardedBy
 -keep class net.jcip.annotations.Immutable
 -keep class net.jcip.annotations.ThreadSafe
@@ -77,4 +78,3 @@
 -dontwarn com.google.auto.value.AutoValue$CopyAnnotations
 -dontwarn com.google.auto.value.AutoValue
 -dontwarn com.google.auto.value.extension.memoized.Memoized
-
--- a/common/consumer-rules.pro
+++ b/common/consumer-rules.pro
@@ -50,9 +50,13 @@
 # keep everything in this package from being removed or renamed
 -keep class io.opentelemetry.** { *; }
 
+# keep names
+-keepnames class com.microsoft.identity.** { *; }
+
 # Prevent R8 from leaving Data object members always null
--keepclassmembers,allowobfuscation class * {
+-keepclassmembers class * {
   @com.google.gson.annotations.SerializedName <fields>;
+  @com.squareup.moshi.Json <fields>;
 }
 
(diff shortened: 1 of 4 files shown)


common/consumer-rules.pro
```diff
@@ -50,13 +50,31 @@
 # keep everything in this package from being removed or renamed
 -keep class io.opentelemetry.** { *; }
 
+# keep names only to allow keeping them logs and exceptions
+-keepnames class com.microsoft.identity.** { *; }
+
 # Prevent R8 from leaving Data object members always null
--keepclassmembers,allowobfuscation class * {
+-keepclassmembers class * {
```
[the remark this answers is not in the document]
mohitc1: yes, obfuscation not needed.

fadidurah (approved):

common/consumer-rules.pro
```diff
@@ -18,6 +18,33 @@
 
 ##---------------Begin: proguard configuration for Common  --------
 # Intentionally blank, left to consumers of common to implement.
+# keep with optmizations and shrinking, but do not obfuscate
+-keep,allowoptimization,allowshrinking class !com.microsoft.identity.common.java.nativeauth.**, !com.microsoft.identity.common.nativeauth.**, com.microsoft.identity.** { *; }
+-keep class * extends com.microsoft.identity.common.java.authorities.Authority { *; }
+-keep class * extends com.microsoft.identity.common.java.authorities.AzureActiveDirectoryAudience { *; }
```
fadidurah, line 23: Why are we keeping these?

mohitc1: Without these I see serialization issues by GSON. When default constructor is optimized out, GSON fails to correctly deserialize.

To be safe all objects which get serialized/deserialized, I keep them wholly. others are optmized/shrunk due to first line.

shahzaibj (approved):

common/consumer-rules.pro
```diff
@@ -37,26 +64,27 @@
...
 -keep class * extends com.google.gson.TypeAdapter
 -keep class * implements com.google.gson.TypeAdapterFactory
 -keep class * implements com.google.gson.JsonSerializer
 -keep class * implements com.google.gson.JsonDeserializer
+-keep class com.google.gson.reflect.TypeToken { *; }
+-keep class * extends com.google.gson.reflect.TypeToken { *; }
 
+##---------------Begin: proguard configuration for OpenTelemetry  --------
 # keep everything in this package from being removed or renamed
 -keep class io.opentelemetry.** { *; }
```
shahzaibj, line 77: Do we want to optimize further within otel code?

mohitc1: I had tried, the gain was small 40-45KB, but not sure of side effects. So, I left it in this. Will spend time on this and nimbus and add on top as separate PR.

fadidurah (approved):

PR #2760, opened 2025-09-12
Native auth: Update Email OTP MFA to Match EC Implementation, Fixes AB#3351233

nilo-ms:
This PR updates the SDK to match the latest flow from EC.

In this new flow, the developer must always supply an auth Method to the /oauth2/v2.0/challenge endpoint which means once the .mfaRequired error is received from token endpoint, the /oauth2/v2.0/introspect endpoint needs to be called to retrieve the methods which are automatically returned to the external developer.

Furthermore, whenever calling the /token endpoint is called with an MFA Email OTP code, the grant type should be mfa_oob

Fixes [AB#3351233](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3351233)

MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2379


shenj (approved):

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -1739,21 +1669,21 @@ class NativeAuthMsalController : BaseNativeAuthController() {
         )
     }
 
-    private fun performSignInSelectedChallengeCall(
+    private fun performSignInSelectedAuthMethodCall(
```
spetrescu84, line 1672: I am not clear what is the difference between `performSignInChallengeCall` and `performSignInSelectedAuthMethodCall`, the first one sends the challenge or retrieves the methods? The naming looks different than iOS as we have the retrieve of authMethods and then one gets challenged. Is there an extra step here?

mustafamizrak (approved):

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
nilo-ms: The difference is the following:

- performSignInChallengeCall: when the SDK hits the /oauth/challenge endpoint as part of the signIn first factor
- performSignInSelectedAuthMethodCall: when the SDK hits the /oauth/challenge endpoint as part of the MFA specifying the id of the auth method selected

I renamed this method to clarify that this method (performSignInSelectedAuthMethodCall) involves the auth method selection as part of the signIn (first + second factor)

spetrescu84: Ok, clear now, didn't think in that context.

spetrescu84 (approved):

PR #2763, opened 2025-09-17
Fadi/22.1.0 js api log

fadidurah:


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -74,6 +75,10 @@ public abstract class OAuth2WebViewClient extends WebViewClient {
     private final OnPageLoadedCallback mPageLoadedCallback;
     private final Activity mActivity;
 
+    public boolean mAuthUxJavaScriptInterfaceAdded = false;
+
+    public static boolean mIsRunningInAuthProcess = false;
```
mohitc1: >mIsRunningInAuthProcess [](http://example.com/codeflow?start=26&length=23)

just keep in this class.

private boolean mIsRunningInAuthProcess

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -74,6 +75,10 @@ public abstract class OAuth2WebViewClient extends WebViewClient {
     private final OnPageLoadedCallback mPageLoadedCallback;
     private final Activity mActivity;
 
+    public boolean mAuthUxJavaScriptInterfaceAdded = false;
```
mohitc1, line 78: >public [](http://example.com/codeflow?start=4&length=6)

private

fadidurah: it's being used in AzureActiveDirectoryWebViewClient

PR #2767, opened 2025-09-22
Added error handling when webcp redirects have browser protocol, Fixes AB#3385536

somalaya:
Issue : There is a recent change made in WebCP side to pass browser protocol redirects for China market urls, 3rd part MDM urls and also the Url to playstore launch for installing CP. This is an unexpected change on their side and we did not have proper handling for these cases.

Fix : 
1. Added try/catch block around the operations in processWebsiteRequest method.
2. Returning MDM_FLOW as the result instead of CANCELLED result currently being returned when we open link in a browser for webcp urls.
3. Added telemetry for operations in processWebsiteRequest method.
4. Made some minor improvements in AzureActiveDirectoryWebViewClient class to avoid duplicate code for getting broker package name from url.

Related broker with just an attribute added : https://github.com/AzureAD/ad-accounts-for-android/pull/3228

Fixes [AB#3385536](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3385536)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -131,7 +131,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     private HashMap<String, String> mRequestHeaders;
     private String mRequestUrl;
     private boolean mAuthUxJavaScriptInterfaceAdded = false;
-    private boolean mIsWebCpInWebViewFeatureEnabled = false;
+    private boolean mInWebCpFlow = false;
 
     private final String mUtid;
 
@@ -339,10 +339,10 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
             } else if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.ENABLE_ATTACH_PRT_HEADER_WHEN_CROSS_CLOUD) && isCrossCloudRedirect(formattedURL)) {
                 Logger.info(methodTag,"Navigation contains cross cloud redirect.");
                 processCrossCloudRedirect(view, url);
-            } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpEnrollmentUrl(url)) {
+            } else if (mInWebCpFlow && isWebCpEnrollmentUrl(url)) {
                 Logger.info(methodTag,"Navigation contains web cp enrollment url.");
                 processWebCpEnrollmentUrl(view, url);
-            } else if (mIsWebCpInWebViewFeatureEnabled && isWebCpAuthorizeUrl(url)) {
+            } else if (mInWebCpFlow && isWebCpAuthorizeUrl(url)) {
                 processWebCpAuthorize(view, url);
             }  else if (isDeviceCaRequest(url) && isHttpsScheme(url) && isWebCpInWebviewFeatureEnabled(url)) {
                 // Special handling for device CA requests due to a corner case in eSTS for webapps/confidential clients, which should be handled by the WebView.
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -779,21 +839,13 @@ private boolean processPlayStoreURL(@NonNull final WebView view, @NonNull final
     private boolean processPlayStoreURLForBrokerApps(@NonNull final WebView view, @NonNull final String url) {
         final String methodTag = TAG + ":processPlayStoreURL";
 
-        final String appPackageName = (url.contains(COMPANY_PORTAL_APP_PACKAGE_NAME) ?
-                COMPANY_PORTAL_APP_PACKAGE_NAME : AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
+        final String appPackageName = getBrokerAppPackageNameFromUrl(url);
         Logger.info(methodTag, "Request to open PlayStore to install package : '" + appPackageName + "'");
 
         try {
             final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(PLAY_STORE_INSTALL_APP_PREFIX + appPackageName));
```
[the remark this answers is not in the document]
somalaya: Removed intentionally as it is not needed

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -574,19 +575,78 @@ private void processSwitchBrowserRequest(@NonNull final String url) {
...
+        returnResult(RawAuthorizationResult.ResultCode.MDM_FLOW);
+    }
+
+    /**
+     * Checks if it is a redirect to Playstore to install Company Portal app.
+     * @param url The URL to check.
+     * @return true if it is a redirect to Playstore to install Company Portal app.
+     */
+    private boolean isRedirectToPlaystoreToInstallCp(@NonNull final String url) {
+       if (url.contains(PLAY_STORE_INSTALL_APP_PREFIX) && url.contains(COMPANY_PORTAL_APP_PACKAGE_NAME)) {
```
rpdome: PLAY_STORE_INSTALL_APP_PREFIX + COMPANY_PORTAL_APP_PACKAGE_NAME?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -574,19 +575,78 @@ private void processSwitchBrowserRequest(@NonNull final String url) {
...
 
+    /**
+     * Handles the default browser redirect case for website requests.
+     * @param methodTag The method tag for logging
+     * @param url The URL to open in browser
+     */
+    private void handleBrowserRedirect(@NonNull final String methodTag, @NonNull final String url) {
+        Logger.info(methodTag, "Not a device CA request. Redirecting to browser.");
+        openLinkInBrowser(url);
+        final RawAuthorizationResult.ResultCode resultCode = mIsWebCpInWebViewFeatureEnabled
```
rpdome: this mIsWebCpInWebViewFeatureEnabled is set via flight, right? wouldn't this make every redirect (including non-cp flows) return MDM_FLOW?

somalaya: It is set via flight but only after we hit the CA block page.
So first we hit the CA block page ---> set mIsWebCpInWebViewFeatureEnabled = true ---> then the following URLs will all be redirects from webcp. So, launching any of them to browser should return MDM_FLOW

somalaya: Changed var name to mInWebCpFlow to correctly reflect the intention.

rpdome (approved):

melissaahn (approved):

melissaahn (approved):

PR #2769, opened 2025-09-23
Add passkey registration support for WebView, Fixes AB#3385532

p3dr0rv:
[AB#3385532](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3385532)
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/20357

This PR implements passkey registration support for MSAL/Broker/OneAuth on Android WebView through a WebMessageListener bridge. It extends the existing authentication-only passkey functionality to include full registration capabilities, leveraging the standard Android Credential Manager.

🎯 Key Features
Protocol Version Update
* Current: x-ms-PassKeyAuth: 1.0/passkey (authentication only)
* New: x-ms-PassKeyAuth: 1.1/passkey (registration + authentication)

**Passkey and Credential Manager Integration:**

* Added new `CredentialManagerHandler` class to encapsulate passkey creation and retrieval using the Android Credential Manager API, including version checks and logging. This simplifies and centralizes interactions with the credential APIs.
* Introduced `PasskeyReplyChannel` class to standardize communication of WebAuthn responses (success and error) back to JavaScript via `JavaScriptReplyProxy`, with detailed error mapping to DOMException types per the WebAuthn specification.

**Protocol Version Handling:**

* Updated protocol version validation in `FidoChallengeField` to accept both `1.0` and `1.1` as supported passkey protocol versions, improving compatibility with newer protocol versions.

**Dependency Management:**

* Added `androidx.webkit:webkit` as a dependency to support enhanced WebView and JavaScript interaction features.

> Telemetry will be added in following PR


p3dr0rv:
Assemble consumers will fail because webkit version is not defined in those libraries yet.

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -58,7 +58,7 @@ public enum CommonFlight implements IFlightConfig {
      * Flight to be able to disable/rollback the passkey feature in broker if necessary.
      * This will be set to true by default.
```
melissaahn: Javadocs need to be updated

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -470,9 +494,69 @@ private SwitchBrowserProtocolCoordinator getSwitchBrowserCoordinator() {
...
+     *
+     * @param requestHeaders The request headers to modify.
+     */
+    private void injectPasskeyProtocolHeader(@NonNull final HashMap<String, String> requestHeaders) {
+        final String methodTag = TAG + ":injectPasskeyProtocolHeader";
+        final Uri authRequestUri = Uri.parse(mAuthorizationRequestUrl);
+        final String webAuthNQueryParameter = authRequestUri.getQueryParameter(FidoConstants.WEBAUTHN_QUERY_PARAMETER_FIELD);
+        final boolean hasWebAuthNQueryParameter = !StringUtil.isNullOrEmpty(webAuthNQueryParameter);
+
+        if (!hasWebAuthNQueryParameter) {
```
melissaahn: nit: no need for the hasWebAuthNQueryParameter variable if it's only going to be used for this condition

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyWebListener.kt
```diff
@@ -0,0 +1,364 @@
...
+        /** JSON key for the request payload field. */
+        const val REQUEST_KEY = "request"
+
+        /** Name of the JavaScript message port interface. */
+        private const val INTERFACE_NAME = "__webauthn_interface__"
+
+        /**
+         * Minified JavaScript code that intercepts WebAuthn API calls.
+         *
+         * ⚠️ IMPORTANT: This is the MINIFIED version of js-bridge.js
```
melissaahn, line 257: I didn't know it was possible to put emojis in docs ⚠️

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyReplyChannel.kt
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Sends a message to JavaScript via the reply proxy.
+     */
+    @SuppressLint("RequiresFeature", "Only called when feature is available")
+    private fun send(message: ReplyMessage) {
+        val methodTag = "$TAG:send"
+        try {
+            replyProxy.postMessage(message.toString())
+        } catch (t: Throwable) {
+            Logger.error(methodTag, "Reply message failed", t)
```
melissaahn, line 224: Could we potentially get stuck if an exception was thrown here and ESTS UX was still waiting for a response from us? Not sure what the mitigation would be in this case if the actual communication channel wasn't working, but just curious what the user might experience.

melissaahn (approved):

common/src/main/assets/js-bridge.js
```diff
@@ -0,0 +1,205 @@
...
+__webauthn_hooks__.originalGetFunction = navigator.credentials.get;
+__webauthn_hooks__.originalCreateFunction = navigator.credentials.create;
+navigator.credentials.get = __webauthn_hooks__.get;
+navigator.credentials.create = __webauthn_hooks__.create;
+// Some sites test that `typeof window.PublicKeyCredential` is
+// `function`.
+window.PublicKeyCredential = (function () { });
+window.PublicKeyCredential.isUserVerifyingPlatformAuthenticatorAvailable =
+    function () {
+        return Promise.resolve(false);
```
melissaahn: Just double checking my understanding: this part controls if the passkey option is visible correct? By always returning false here, this ensures that we can control the visibility with webauthn=1 instead?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyReplyChannel.kt
p3dr0rv: I agree this is a valid concern. I’ve refactored the error handling so that whenever an error occurs in the communication layer, it is now raised properly.

common/src/main/assets/js-bridge.js
p3dr0rv: This JS method basically tells the webpage whether the current device has a built-in (platform) authenticator — like a fingerprint sensor, Face ID, or passkey support, so this by default is false in WebView is false, https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/isUserVerifyingPlatformAuthenticatorAvailable_static.

In our case, since we’re implementing the credential manager integration, we could actually return **true**. I believe [ESTS skips this validation](https://dev.azure.com/msazure/One/_search?action=contents&text=SkipFidoStandardSupport&type=code&lp=code-Team&filters=ProjectFilters%7BOne%7DRepositoryFilters%7BESTS-Main%7D&pageSize=25&result=DefaultCollection/One/ESTS-Main/GBmaster//src/Product/Microsoft.AzureAD.ESTS.Config/Teams/UXTeam.config.json), which is why it worked before, but it makes more sense to explicitly return **true** here.

I will double check with Suresh

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -470,9 +494,68 @@ private SwitchBrowserProtocolCoordinator getSwitchBrowserCoordinator() {
...
+        } else {
+            if (requestHeaders.containsKey(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME)) {
+                Logger.verbose(methodTag, "Passkey protocol header already exists in request headers  "
+                        + requestHeaders.get(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME));
+                return;
+            }
+        }
+        Logger.verbose(methodTag, "Injecting Passkey protocol header for auth only.");
+        requestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_AUTH_ONLY);
+    }
```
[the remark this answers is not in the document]
p3dr0rv: fixed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyWebListener.kt
```diff
@@ -0,0 +1,361 @@
...
+    private suspend fun handleGetFlow(
+        credentialManagerHandler: CredentialManagerHandler,
+        message: String,
+        reply: PasskeyReplyChannel
+    ) {
+        runCatching { credentialManagerHandler.getPasskey(message) }
+            .onSuccess { credentialResponse ->
+                reply.postSuccess(
+                    (credentialResponse.credential as PublicKeyCredential).authenticationResponseJson
+                )            }
```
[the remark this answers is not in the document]
p3dr0rv: solved

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -349,7 +353,35 @@ public Bitmap getDefaultVideoPoster() {
...
+             * @param consoleMessage The console message from the WebView
+             * @return returns true if we handled the message. False will let the default handler handle it.
+             */
+            @Override
+            public boolean onConsoleMessage(@NonNull final ConsoleMessage consoleMessage) {
+                // Note: Decide what we are interested in logging and what level.
+                super.onConsoleMessage(consoleMessage);
+                final String methodTag = TAG + ":onConsoleMessage";
+                final String errorMessage = consoleMessage.message() + " -- From line " +
+                        consoleMessage.lineNumber() + " of " + consoleMessage.sourceId();
```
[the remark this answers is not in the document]
p3dr0rv: change to log PII

common/src/main/assets/js-bridge.js
p3dr0rv: from Suresh:
image

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/CredentialManagerHandler.kt
```diff
@@ -0,0 +1,81 @@
...
+import android.os.Build
+import androidx.credentials.CreatePublicKeyCredentialRequest
+import androidx.credentials.CreatePublicKeyCredentialResponse
+import androidx.credentials.CredentialManager
+import androidx.credentials.GetCredentialRequest
+import androidx.credentials.GetCredentialResponse
+import androidx.credentials.GetPublicKeyCredentialOption
+import com.microsoft.identity.common.logging.Logger
+
+class CredentialManagerHandler(private val activity: Activity) {
```
fadidurah, line 40: Javadoc

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -355,7 +359,35 @@ public Bitmap getDefaultVideoPoster() {
...
+                    case ERROR:
+                        Logger.errorPII(methodTag, errorMessage, null);
+                        break;
+                    case WARNING:
+                        Logger.warnPII(methodTag, errorMessage);
+                        break;
+                    default:
+                        Logger.infoPII(methodTag, errorMessage);
+                }
+                return true;
```
fadidurah: I'm a little concerned about how many additional logs this will create in our logs. Have we tested this to see how much noise this adds?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/JsScriptRecord.kt
```diff
@@ -0,0 +1,74 @@
...
+     * @param url The URL to check against the allowed list.
+     * @return `true` if the script can execute for this URL, `false` otherwise.
+     */
+    fun isAllowedForUrl(url: String): Boolean {
+        // No restrictions — allowed for any URL
+        if (allowedUrls == null) return true
+
+        // Check if the URL starts with any allowed prefix
+        return allowedUrls.any { prefix ->
+            if (SOVEREIGN_CLOUD_URL_WITH_EXTRA_VALIDATION.contains(prefix) && url.startsWith(prefix)) {
```
fadidurah: We should validate on suffix rather than prefix. What if url was https://login.microsoftonline.us.someDomain.com

fadidurah: You can check `isValidUriForInterface` in `AuthUxJavaScriptInterface`

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1218,4 +1235,20 @@ private Span createSpanWithAttributesFromParent(@NonNull final String spanName)
...
     }
+
+    /**
+     * Add a JavaScript to be executed in onPageStarted.
+     * If allowedUrls is null, the script will be executed for all URLs.
+     * If allowedUrls is non-null, the script will be executed only for URLs that start with any of the allowed origins.
+     * @param script JavaScript code to be executed.
+     * @param allowedUrls Set of allowed URL origins.
+     */
+    public void addOnPageStartedScript(
```
fadidurah, line 1246: Where do we call this?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyWebListener.kt
```diff
@@ -0,0 +1,383 @@
...
+         *
+         * When updating:
+         * 1. Modify the source file (js-bridge.js) with your changes
+         * 2. Minify the updated JavaScript code
+         * 3. Replace the string below with the new minified version
+         * 4. Verify the minified code works correctly through testing
+         *
+         * DO NOT modify this constant directly - always update the source file first!
+         */
+        private const val WEB_AUTHN_INTERFACE_JS_MINIFIED = """
```
fadidurah, line 269: So we use this constant, not the contents of the js-bridge.js file?

fadidurah: nvm, i see it is a fallback

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -476,9 +500,69 @@ private SwitchBrowserProtocolCoordinator getSwitchBrowserCoordinator() {
...
+            Logger.verbose(methodTag, "Injecting Passkey protocol header for broker request: "
+                    + passkeyProtocolHeaderValue);
+            requestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, passkeyProtocolHeaderValue);
+        } else {
+            if (requestHeaders.containsKey(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME)) {
+                Logger.verbose(methodTag, "Passkey protocol header already exists in request headers  "
+                        + requestHeaders.get(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME));
+            } else {
+                Logger.verbose(methodTag, "Injecting Passkey protocol header for auth only.");
+                requestHeaders.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_AUTH_ONLY);
```
fadidurah, line 535: Should we check the version here to be 1.1?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyReplyChannel.kt
```diff
@@ -0,0 +1,228 @@
...
+        /**
+         * Success message containing credential data.
+         *
+         * @property json JSON string with credential response data.
+         * @property type Request type that succeeded.
+         */
+        class Success(val json: String, override val type: String) : ReplyMessage() {
+            override val status = SUCCESS_STATUS
+            override val data: JSONObject =
+                runCatching { JSONObject(json) }.getOrElse { JSONObject() }
```
fadidurah, line 94: When we say credential data, how is it stored? Is it encrypted at all? This might be a security concern as it's being stored in memory, so we might want to be careful here. I'm not super familiar with the feature so let me know

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: Use version 1.1 only if the client specifies it; otherwise, default to 1.0.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
p3dr0rv: removed

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/JsScriptRecord.kt
p3dr0rv: updated

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyReplyChannel.kt
p3dr0rv: We do not persist any credential material. During passkey creation the private key remains in the provider’s keystore; our application only receives the public key and the creation response (attestation). We transmit that response to the server so the server can register the credential. Any credential data present on the client exists only briefly in memory.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/JsScriptRecord.kt
```diff
@@ -0,0 +1,88 @@
...
+     * @param url The URL to check against the allowed list.
+     * @return `true` if the script can execute for this URL, `false` otherwise.
+     */
+    fun isAllowedForUrl(url: String): Boolean {
+        // No restrictions — allowed for any URL
+        if (allowedUrls == null) return true
+
+        val uri = url.toUri()
+
+        // Check if the URL matches any allowed prefix
```
fadidurah: updated as we don't validate on prefix anymore. We are checking the url for exact match now right?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/JsScriptRecord.kt
p3dr0rv: yes, The comment is incorrect. updating

fadidurah (approved):

PR #2671, opened 2025-06-12
Using Baggage to propagate attributes from parent Span, Fixes AB#3262849

melissaahn:
### Summary
The passkey telemetry dashboard (and maybe some others) has had issues with latency due to joins being done between the Fido span and the parent span related to acquire token interactive. These join statements by trace id allowed us to associate attributes (such as tenant id) to the Fido span that weren't available in the context of the Fido classes.
In order to directly save the attributes onto the Fido span itself, we can make use of [Otel's Baggage](https://opentelemetry.io/docs/specs/otel/baggage/api/), which is meant to help hold data across multiple spans (from parent to children). We can then use an implementation of [Otel's TextMapPropagator](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) to serialize the Otel Context and pass it to the AuthenticationActivity as an Extra (similar to what we are doing for spanContext). This Context can then be accessed by the WebViewClient and Fido related classes, where they will retrieve the Baggage.

This PR has the changes to implement this behavior and adds a few static helper classes for TextMapPropagator and Baggage (the baggage class is more relevant for the corresponding Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3119). 

While this PR tackles attribute propagation for AuthorizationActivity specifically, Baggage and the helper classes can be used in other areas of the code where propagation is needed as well.

[AB#3262849](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3262849)

PR #2770, opened 2025-09-23
Use baggage for spans, Fixes AB#3385536

somalaya:
Using baggage to get parent span's (AcquireTokenInteractive) correlationId, calling app package name and setting them on the child spans created during webcp redirects in webview.

Why this needed?
- We do not have correlationId set on webcp spans as of today. Although, we can check for the parent span and get correlationId from there, it is just an additional step. I want to avoid this.
- We cannot directly set attributes to indicate that is webcp flow on the parent span (ATI) since the context passed to the activity is immutable. Recommended way is to create a child span and set attributes as needed on it.
- We also cannot get parent span's attributes on child span directly using the context passed. Support for this was added by using baggage in this PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2671. So, I am leveraging this. 

Related broker PR to remove the redundant attributes : https://github.com/AzureAD/ad-accounts-for-android/pull/3230

Fixes [AB#3385536](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3385536)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -583,9 +583,7 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     protected void processWebsiteRequest(@NonNull final WebView view, @NonNull final String url) {
         final String methodTag = TAG + ":processWebsiteRequest";
         view.stopLoading();
-        final SpanContext spanContext = getActivity() instanceof AuthorizationActivity ? ((AuthorizationActivity) getActivity()).getSpanContext() : null;
-        final Span span = spanContext != null ?
-                OTelUtility.createSpanFromParent(SpanName.ProcessWebsiteRequest.name(), spanContext) : OTelUtility.createSpan(SpanName.ProcessWebsiteRequest.name());
+        final Span span = createSpanWithAttributesFromParent(SpanName.ProcessWebsiteRequest.name());
         span.setAttribute(AttributeName.is_in_web_cp_flow.name(), mInWebCpFlow);
         try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
             if (isDeviceCaRequest(url)) {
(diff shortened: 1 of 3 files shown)


melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1138,4 +1135,33 @@ public void finalizeBeforeSendingResult(@NonNull final RawAuthorizationResult re
         ((AbstractSmartcardCertBasedAuthChallengeHandler<?>)mCertBasedAuthChallengeHandler).promptSmartcardRemovalForResult(callback);
     }
 
+    /**
+     * Create a span with parent span context if available.
+     * @param spanName Name of the span to be created.
+     * @return Created {@link Span}
+     */
+    private Span createSpanWithAttributesFromParent(@NonNull final String spanName) {
```
fadidurah, line 1211: Should we have this in SpanExtension instead, or some central place to access in other places? This seems generally useful

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Sure, I can think of making it generally available to all spans. I have more telemetry changes coming in, will put them in that PR.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -739,7 +739,7 @@ protected boolean isWebCpInWebviewFeatureEnabled(@NonNull final String originalU
             final int waitForFlightsTimeOut = CommonFlightsManager.INSTANCE.getFlightsProvider().getIntValue(CommonFlight.WEB_CP_WAIT_TIMEOUT_FOR_FLIGHTS);
             final boolean isWebCpFlightEnabled = CommonFlightsManager.INSTANCE.getFlightsProviderForTenant(homeTenantId, waitForFlightsTimeOut).isFlightEnabled(CommonFlight.ENABLE_WEB_CP_IN_WEBVIEW);
             SpanExtension.current().setAttribute(AttributeName.web_cp_flight_get_time.name(), (System.currentTimeMillis() - webCpGetFlightStartTime));
-
+            SpanExtension.current().setAttribute(AttributeName.tenant_id.name(), homeTenantId);
```
fadidurah, line 742: Seems you are adding tenant id here, which is good

fadidurah (approved):

PR #2772, opened 2025-09-25
Native auth: add new SDK error when auth method is blocked, Fixes AB#3313635

nilo-ms:
This PR adds a new SDK error with instructions when the auth method is blocked during MFA flow.

MSAL PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2384

Fixes [AB#3313635](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3313635)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
+++ b/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
@@ -482,7 +482,7 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                         correlationId = result.correlationId
                     )
                 }
-                is SignInChallengeApiResult.PasswordRequired -> {
+                is SignInChallengeApiResult.PasswordRequired, is SignInChallengeApiResult.BlockedAuthMethod -> {
                     Logger.warnWithObject(
                         TAG,
                         result.correlationId,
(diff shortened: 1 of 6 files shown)


andwhysoft (approved):

spetrescu84 (approved):

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
```diff
@@ -2680,6 +2687,21 @@ class NativeAuthMsalController : BaseNativeAuthController() {
                     correlationId = result.correlationId
                 )
             }
+
+            is SignInChallengeApiResult.BlockedAuthMethod -> {
+                Logger.warnWithObject(
+                    TAG,
+                    result.correlationId,
+                    "Unexpected result: ",
```
mustafamizrak, line 2695: <NIT> Giving more specific error message like  "Auth Method Blocked:"

mustafamizrak (approved):

common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/NativeAuthMsalController.kt
nilo-ms: This result is not expected here, for this reason we're not giving a specific error

spetrescu84 (approved):

PR #2773, opened 2025-09-26
Update Java Script Api Exposure, Fixes AB#3385083

fadidurah:
This PR will move the code that exposes the java script api from handleUrl on pageStarted. This will make it so we use the url that we are actually planning to load in the webview to validate whether or not to exposue the JS interface. previously, most pages were being validated correctly in handleUrl, but in certain redirects, we may have been missing some urls that should have had the url exposed. This change addresses those gaps by checking every url the webview is about to load.

[AB#3385083](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3385083)


common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterface.kt
```diff
@@ -53,32 +53,32 @@ class AuthUxJavaScriptInterface {
...
-            // Otherwise, make sure url is a valid url
-            // We only want to allow URLs that have the AAD or MSA url hosts
-            return host.startsWith(AuthenticationConstants.Broker.AAD_URL_HOST_PREFIX) ||
-                    host.startsWith(AuthenticationConstants.Broker.MSA_URL_HOST_PREFIX)
+            // Otherwise, make sure uri is a valid uri
+            // We only want to allow URIs that have the AAD or MSA uri hosts
+            return host.endsWith(AuthenticationConstants.Broker.AAD_GLOBAL_URL_HOST_SUFFIX) ||
+                    host.endsWith(AuthenticationConstants.Broker.AAD_US_URL_HOST_SUFFIX) ||
+                    host.endsWith(AuthenticationConstants.Broker.AAD_CHINA_URL_HOST_SUFFIX) ||
+                    host.endsWith(AuthenticationConstants.Broker.MSA_URL_HOST)
```
siddhijain: This feature is for AAD only so we do not need MSA_URL_HOST check. Also, have your confirmed that the AAD list is exhaustive?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -144,6 +143,7 @@ public AzureActiveDirectoryWebViewClient(@NonNull final Activity activity,
         mRedirectUrl = redirectUrl;
         mCertBasedAuthFactory = new CertBasedAuthFactory(activity);
         mSwitchBrowserRequestHandler = switchBrowserRequestHandler;
+        mIsRunningInAuthProcess = ProcessUtil.isRunningOnAuthService(getActivity().getApplicationContext());
```
mohitc1: > mIsRunningInAuthProcess = ProcessUtil.isRunningOnAuthService(getActivity().getApplicationContext()); [](http://example.com/codeflow?start=7&length=101)

This can all be in OAuth2WebViewClient.

fadidurah: Losing my mind. I thought i had checked this and it wouldn't let me, but definitely me missing something. moved it

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterface.kt
fadidurah: Will we expose the API in the future for any MSA enabled scenarios? I mean outside of number match

These three hosts will cover global, china, and us gov clouds, should be exhaustive, but i'll double check with iOS. for our purposes, i would think global and us gov should be plenty

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1223,15 +1223,24 @@ public static String computeMaxHostBrokerProtocol() {
         public static final String REDIRECT_PREFIX = "msauth";
 
         /**
-         * Prefix for AAD urls
+         * Suffix for global AAD urls
          */
-        public static final String AAD_URL_HOST_PREFIX = "login.microsoftonline.";
+        public static final String AAD_GLOBAL_URL_HOST_SUFFIX = ".microsoftonline.com";
```
mohitc1, line 1228: >public static final String AAD_GLOBAL_URL_HOST_SUFFIX = ".microsoftonline.com"; [](http://example.com/codeflow?start=8&length=79)

why this change in URL validation logic?

fadidurah: Checked with ios team, they said validating based on "startsWith" doesn't work because it would validate a malicious domain such as login.microsoftonline.my-scam-site.com. With this change, we validate on endsWith

common/src/main/java/com/microsoft/identity/common/internal/broker/AuthUxJavaScriptInterface.kt
fadidurah: We can remove msa check for now, i guess add it in the future when we enable a feature for MSA

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -78,6 +78,8 @@ public abstract class OAuth2WebViewClient extends WebViewClient {
     @VisibleForTesting
     public static ExpectedPage mExpectedPage = null;
 
+    protected boolean mAuthUxJavaScriptInterfaceAdded = false;
```
mohitc1, line 81: >protected [](http://example.com/codeflow?start=4&length=9)

can be private now

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
```diff
@@ -248,4 +264,10 @@ private void checkStartUrl(final String url) {
                     + " Path: " + uri.getPath());
         }
     }
+
+    public boolean shouldExposeJavaScriptInterface(final String url) {
```
mohitc1: >public [](http://example.com/codeflow?start=4&length=6)

private

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
mohitc1: Thanks. I will add that whenever we have had to ensure trusted urls are being used, we have at many places used "cloud" derived from the url and use it for comparison. Here we are using URL validation directly. Just calling out the difference. Perhaps it's better to validate on url here than cloud.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1223,15 +1223,39 @@ public static String computeMaxHostBrokerProtocol() {
...
 
         /**
-         * Prefix for MSA urls
+         * Suffix for China AAD urls
          */
-        public static final String MSA_URL_HOST_PREFIX = "login.live.";
+        public static final String AAD_CHINA_URL_HOST_SUFFIX = ".microsoftonline.cn";
 
+        /**
```
siddhijain, line 1235: Still missing few from the list 
image

siddhijain (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
fadidurah: We have all of these already, in AuthUxJavaScriptInterface is the full list. We just do not include msa url

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
fadidurah: protected, we use it in AzureActiveDirectoryWebViewClient

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/OAuth2WebViewClient.java
fadidurah: used in AzureActiveDirectoryWebViewClient

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
siddhijain: I thought we said we will include MSA url as well until Veena confirms that it is not needed. Lets exactly match with what iOS has.

siddhijain: nvm Veena confirmed that this is not needed.

siddhijain (approved):

mohitc1 (approved):

PR #2774, opened 2025-09-26
getAllSsoTokens method for Edge, Fixes AB#3361797

melissaahn:
### Summary
We have a getSsoToken for OneAuth/Edge which returns a PRT cookie corresponding to a given account. They have asked us to also implement a method which gets all (well, we're capping it at the 3 most recent) PRT cookies based on Environment/cloud, which we can get from a passed in Authority value. 
The changes in common for this method involve defining the method in BrokerMsalController, updating the appropriate adapter classes, creating the AcquirePrtSsoTokenBatchResult class, and bumping the broker protocol version.

Corresponding Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3233

Fixes [AB#3361797](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3361797)


common4j/src/main/com/microsoft/identity/common/java/commands/AcquirePrtSsoTokenResult.java
```diff
@@ -89,4 +89,10 @@ public class AcquirePrtSsoTokenResult {
     @SerializedName("telemetry")
     private final @NonNull Map<String, Object> mTelemetry;
 
+    /**
+     * The time when the token was acquired, in milliseconds since epoch.
+     */
+    @SerializedName("AcquisitionTimeMillis")
+    private final @Nullable Long mAcquisitionTimeMillis;
```
melissaahn, line 96: From my understanding, supposing a client with an older MSAL version calls the getSsoToken method, their version of AcquirePrtSsoTokenResult won't have this mAcquisitionTimeMillis, but GSON should be ignoring any parameters which they don't recognize, and thus shouldn't throw an exception or cause a crash.
Folks can correct me if I'm wrong here (and I'll be sure to test this out)

common4j/src/main/com/microsoft/identity/common/java/commands/AcquirePrtSsoTokenResult.java
```diff
@@ -89,4 +89,10 @@ public class AcquirePrtSsoTokenResult {
     @SerializedName("telemetry")
     private final @NonNull Map<String, Object> mTelemetry;
 
+    /**
+     * The time when the token was acquired, in milliseconds since epoch.
+     */
+    @SerializedName("AcquisitionTimeMillis")
```
shahzaibj:
```suggestion
    @SerializedName("acquisitionTimeMillis")
```

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1188,6 +1191,66 @@ public void putValueInSuccessEvent(@NonNull final ApiEndEvent event,
         });
     }
 
+    /**
+     * Get SSO tokens from broker by environment from given request authority.
+     * Note: The result is limited to the 3 most recent tokens.
```
shahzaibj: What does recent mean in this context?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
```diff
@@ -67,7 +67,7 @@ public enum SpanName {
     SecretKeyWrapping,
     WrappedKeyAlgorithmIdentifier,
     ProcessWebCpRedirects,
-
     ProvisionResourceAccount,
-    ProcessWebsiteRequest
+    ProcessWebsiteRequest,
+    GetAllSsoTokens
```
shahzaibj: How often will this be invoked?

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -232,6 +232,24 @@ public BrokerRequest brokerRequestFromSilentOperationParameters(@NonNull final S
...
+        bundle.putString(AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY, negotiatedBrokerProtocolVersion);
+        if (parameters.getRequestAuthority() != null) {
+            bundle.putString(AuthenticationConstants.Broker.REQUEST_AUTHORITY, parameters.getRequestAuthority());
+        }
+        if (parameters.getSsoUrl() != null) {
+            bundle.putString(AuthenticationConstants.Broker.BROKER_SSO_URL_KEY, parameters.getSsoUrl());
+        }
+        if (parameters.getCorrelationId() != null) {
+            bundle.putString(AuthenticationConstants.Broker.ACCOUNT_CORRELATIONID, parameters.getCorrelationId());
+        }
```
mohitc1, line 248: addRequiredBrokerProtocolVersionToRequestBundle() is missing here.

Would caller provide required min version? if so, this call is needed and then is validated on broker side.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -232,6 +232,24 @@ public BrokerRequest brokerRequestFromSilentOperationParameters(@NonNull final S
         return requestBundle;
     }
 
+    public Bundle getRequestBundleForAllSsoTokens(
+            @NonNull final AcquirePrtSsoTokenCommandParameters parameters,
+            @NonNull final String negotiatedBrokerProtocolVersion) {
```
mohitc1: Nonnull


Also do you want to add extra validation on presence of required parameters?

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: most recently acquired PRTs. I will update the javadoc.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/SpanName.java
melissaahn: Discussed offline; this is going to be called multiple times per login session. So I will set a lower sampling rate, and once I get a better idea of an average amount of calls, I'll be able to control that rate via ECS.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
melissaahn: Thank you for catching that. Added that method in, and it should be getting validated with the BrokerUtils.validateRequiredBrokerProtocolVersion(components, requestBundle) method on the broker side.

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
melissaahn: Added the Nonnull; right now the validation is done on the broker side, which matches what is done for the existing getSsoToken method. Though I will consider updating both of these potentially if there's a big difference in validating earlier.

common4j/src/main/com/microsoft/identity/common/java/commands/AcquirePrtSsoTokenBatchResult.java
```diff
@@ -0,0 +1,74 @@
...
+import lombok.NonNull;
+import lombok.experimental.Accessors;
+
+/**
+ * A DTO for the results of a batch AcquirePrtSsoToken request.
+ */
+@Builder
+@Getter
+@Accessors(prefix = "m")
+public class AcquirePrtSsoTokenBatchResult {
```
p3dr0rv, line 42: you could also use a Kotlin data class, to avoid Lombok and boilerplate code.
Unless is mixed Java+ Kotlin usage

p3dr0rv (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/AcquirePrtSsoTokenBatchResult.java
melissaahn: True, though this class is referenced in both Java and Kotlin contexts. I'll keep this in mind for next time

PR #2775, opened 2025-09-28
Fix for app link redirect from CCT due to forced browser preference, Fixes AB#3387976

shahzaibj:
Fixes [AB#3387976](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3387976)

The previous fix seems to have addressed the issue only partially. 
Previous fix: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2745

Basically what's happening is that even with that flag CCT will sometimes still launch the URL in browser and this is happening because we're setting the browser's pkg name explicitly on the intent which tells the OS that this MUST be opened in the browser.

Checked CCT logs and they print something like:

`Launching in browser as Browser pkg was explicitly set on the Intent`

Once we remove the explicit pkg name then OS finds the best possible app on the device to handle the intent. For an app link that's the the broker hosting app.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
@@ -147,7 +147,6 @@ class SwitchBrowserActivity : FragmentActivity() {
             Logger.warn(methodTag, "CustomTabsService is NOT supported")
             browserIntent = Intent(Intent.ACTION_VIEW)
         }
-        browserIntent.setPackage(browserPackageName)
         browserIntent.setData(processUri.toUri())
         startActivity(browserIntent)
     }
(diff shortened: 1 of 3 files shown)


melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
```diff
@@ -147,7 +147,6 @@ class SwitchBrowserActivity : FragmentActivity() {
             Logger.warn(methodTag, "CustomTabsService is NOT supported")
             browserIntent = Intent(Intent.ACTION_VIEW)
         }
-        browserIntent.setPackage(browserPackageName)
```
p3dr0rv, line 150: I’m wondering if we should set the intent package in the else condition when CCT isn’t supported, so that the desired browser is used

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/SwitchBrowserActivity.kt
shahzaibj: Good point. Let me try this out.

shahzaibj: Addressed

p3dr0rv (approved):

PR #2781, opened 2025-10-09
Refactor get-account-by-local-account-id FixesAB#3397557, Fixes AB#3397557

siddhijain:
Fixes[AB#3397557](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3397557)

This pull request refactors and optimizes account and credential filtering logic within the token cache implementation. The main focus is on improving performance and readability by streamlining filtering conditions and reducing unnecessary checks, especially in the methods that retrieve accounts and credentials. Additionally, some redundant imports are removed and logging is enhanced for better traceability.

### Filtering Logic Improvements

* Refactored the filtering process in `AbstractAccountCredentialCache.java` by immediately continuing to the next iteration when a match condition fails, rather than accumulating match state across conditions. This makes the code more efficient and easier to read. 

### Account Retrieval Logic

* Rewrote the logic for retrieving accounts by local account ID and environment in `MsalOAuth2TokenCache.java`, now filtering accounts and credentials separately and checking for associated credentials before returning an account. This avoids unnecessary environment checks and improves clarity.

### Codebase Maintenance

* Removed unused imports and streamlined import statements in `MsalOAuth2TokenCache.java` for better maintainability. 

These changes collectively improve the efficiency, readability, and maintainability of the account and credential filtering logic in the token cache.


common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1202,6 +1232,58 @@ private void setToCacheRecord(@NonNull final CacheRecord.CacheRecordBuilder targ
...
+
+        for (final AccountRecord account : accountsFilteredByEnvAndLocalAccountId) {
+            if (accountHasCredential(account, appCredentials)) {
+                return account;
+            }
+        }
+
+        return null;
+    }
+
```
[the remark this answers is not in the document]
siddhijain: yes - this was created for temporary test and should be removed. I will remove this.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -115,21 +115,24 @@ protected List<AccountRecord> getAccountsFilteredByInternal(@Nullable final Stri
...
             }
 
             if (mustMatchOnEnvironment) {
                 matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(environment, account.getEnvironment());
+                if(!matches) continue;
             }
 
             if (mustMatchOnRealm) {
                 matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(realm, account.getRealm());
+                if(!matches) continue;
```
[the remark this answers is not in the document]
siddhijain: Done

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1092,16 +1091,45 @@ public AccountRecord getAccountByLocalAccountId(@Nullable final String environme
...
+        for (AccountRecord accountRecord: accountRecordList) {
+            if (accountRecord.getLocalAccountId().equals(localAccountId)
+                    && accountRecord.getEnvironment().equals(environment)) {
+                accountsFilteredByEnvAndLocalAccountId.add(accountRecord);
+            }
+        }
+
+        Logger.verbose(
+                TAG + methodName,
+                "Found " + accountsFilteredByEnvAndLocalAccountId.size() + " accounts for this environment" +
```
siddhijain: Ideally there should be only 1 account with a local_account_id? If that's the cache the logic to check each account against each credential can be improved.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -115,21 +115,24 @@ protected List<AccountRecord> getAccountsFilteredByInternal(@Nullable final Stri
         for (final AccountRecord account : allAccounts) {
             boolean matches = true;
```
melissaahn: Q: Does the matches variable have to be used here? would it be valid for the logic to be simplified to something like
```
if (mustMatchOnHomeAccountId && !StringUtil.equalsIgnoreCaseTrimBoth(homeAccountId, account.getHomeAccountId()) continue
... (same sort of logic for env and realm)
matchesAccounts.add(account);
```

siddhijain: Yeah, the problem is that there are too many conditions to check, which will make this form of condition matching really large.

melissaahn: But each condition block is independent right? If I were to expand upon what I was thinking:
```
if (mustMatchOnHomeAccountId && !StringUtil.equalsIgnoreCaseTrimBoth(homeAccountId, account.getHomeAccountId()) continue
if (mustMatchOnEnvironment && !StringUtil.equalsIgnoreCaseTrimBoth(environment, account.getEnvironment()) continue
if (mustMatchOnRealm && !StringUtil.equalsIgnoreCaseTrimBoth(realm, account.getRealm()) continue

matchesAccounts.add(account);
```
Since we're failing fast now, it doesn't really matter what the result of the previous if statement was (like if we got to the environment part, we know that home account id matched or account id just didn't matter), so there isn't really a need to carry the matches variable anymore, right?

siddhijain: Thanks for expanding! You're right. That is another way of simplifying this check.

siddhijain: Updated.

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -195,58 +183,67 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
 
             if (mustMatchOnHomeAccountId) {
                 matches = StringUtil.equalsIgnoreCaseTrimBoth(homeAccountId, credential.getHomeAccountId());
+                if (!matches) continue;
```
melissaahn: Like the instance above, I think that a lot of these other instances of "matches" can also be consolidated in the same manner.

siddhijain: Having a separate variable here improves readability and clarity, imo, since the method contains more than just simple string comparisons and some if conditions include additional nested logic.

melissaahn: I see what you mean for maybe the later conditions, but the first ones can definitely be simplified to 
```
            if (mustMatchOnHomeAccountId
                    && !StringUtil.equalsIgnoreCaseTrimBoth(homeAccountId, credential.getHomeAccountId())) continue;

            if (mustMatchOnEnvironment
                    && !StringUtil.equalsIgnoreCaseTrimBoth(environment, credential.getEnvironment())) continue;

            if (mustMatchOnCredentialType
                    && !StringUtil.equalsIgnoreCaseTrimBoth(credentialType.name(), credential.getCredentialType())) continue;

            if (mustMatchOnClientId
                    && !StringUtil.equalsIgnoreCaseTrimBoth(clientId, credential.getClientId())) continue;
```

I do think some of the other ones can be simplified as well (I'll write more targeted comments there), but I'd argue that some of these conditions would actually be more readable without that matches variable, especially when we're failing fast

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -195,58 +183,67 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
...
                     final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
-                    matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(mamEnrollmentIdentifier, accessToken.getMamEnrollmentIdentifier());
+                    matches = StringUtil.equalsIgnoreCaseTrimBoth(mamEnrollmentIdentifier, accessToken.getMamEnrollmentIdentifier());
                 } else {
                     Logger.verbose(TAG, "Query specified mamEnrollmentIdentifier match, but credential type does not have MAM enrollment identifier");
                 }
+                if (!matches) continue;
             }
 
             if (mustMatchOnRealm && credential instanceof AccessTokenRecord) {
```
melissaahn: These two "mustMatchOnRealm" clauses should be combined such that:
```
            if (mustMatchOnRealm) {
                if (credential instanceof AccessTokenRecord) {
                    if (!StringUtil.equalsIgnoreCaseTrimBoth(realm, ((AccessTokenRecord) credential).getRealm())) continue;
                } else if (credential instanceof IdTokenRecord) {
                    if (!StringUtil.equalsIgnoreCaseTrimBoth(realm, ((IdTokenRecord) credential).getRealm())) continue;
                }
            }
```

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -263,24 +260,27 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
...
-                        matches = matches && StringUtil.equalsIgnoreCaseTrimBoth(requestedClaims, accessToken.getRequestedClaims());
+                        matches = StringUtil.equalsIgnoreCaseTrimBoth(requestedClaims, accessToken.getRequestedClaims());
                     }
                 } else {
                     Logger.verbose(TAG, "Query specified requested_claims-match, but attempted to match with non-AT credential type.");
                 }
+                if (!matches) continue;
             }
 
             if (matches) {
```
melissaahn: We don't need to check matches here I think because we're failing fast everywhere else right?

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
```diff
@@ -263,24 +260,27 @@ protected List<Credential> getCredentialsFilteredByInternal(@NonNull final List<
                                     || authScheme.equalsIgnoreCase(PopAuthenticationSchemeInternal.SCHEME_POP)
                     );
                 } else {
-                    matches = matches && authScheme.equalsIgnoreCase(atType);
+                    matches = authScheme.equalsIgnoreCase(atType);
                 }
+                if (!matches) continue;
             }
 
             if(mustMatchOnKid && credential instanceof AccessTokenRecord) {
```
melissaahn, line 248:
````  
 if (mustMatchOnKid && credential instanceof AccessTokenRecord) {
        final AccessTokenRecord accessToken = (AccessTokenRecord) credential;
        if (!kid.equalsIgnoreCase(accessToken.getKid()))  continue;
 }
```

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
siddhijain: Updated

common4j/src/main/com/microsoft/identity/common/java/cache/AbstractAccountCredentialCache.java
siddhijain: Agree. I updated all the conditions.

melissaahn (approved):
Thank you!

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1092,16 +1091,45 @@ public AccountRecord getAccountByLocalAccountId(@Nullable final String environme
...
-        final List<AccountRecord> accounts = getAccounts(environment, clientId);
-
         Logger.verbosePII(
                 TAG + methodName,
                 "LocalAccountId: [" + localAccountId + "]"
         );
 
-        for (final AccountRecord account : accounts) {
-            if (localAccountId.equals(account.getLocalAccountId())) {
+        List<AccountRecord> accountsFilteredByEnvAndLocalAccountId = new ArrayList<>();
```
mohitc1: >List<AccountRecord> accountsFilteredByEnvAndLocalAccountId = new ArrayList<>(); [](http://example.com/codeflow?start=8&length=79)

final, here and other places

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1092,16 +1091,45 @@ public AccountRecord getAccountByLocalAccountId(@Nullable final String environme
...
+        List<AccountRecord> accountRecordList = mAccountCredentialCache.getAccounts();
```
mohitc1: >List<AccountRecord> accountRecordList = mAccountCredentialCache.getAccounts(); [](http://example.com/codeflow?start=8&length=78)

A couple more points on optimization

1. If accountRecordList size is zero, return early. This can be added in getAccounts(environment, client) method as well.
2. We are anyway calling getCredentialsFilteredBy() after getting accounts, so we can call that first to get appCredentials and then iterate over accounts

for (AccountRecord accountRecord: accountRecordList) {
            if (accountRecord.getLocalAccountId().equals(localAccountId)
                    && accountRecord.getEnvironment().equals(environment) && accountHasCredential(account, appCredentials)) {
               return accountRecord;
            }
        }

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -1355,20 +1383,17 @@ private CredentialType getAccessTokenCredentialTypeForAuthenticationScheme(
...
 
         Logger.verbosePII(
                 TAG + methodName,
                 "HomeAccountId: [" + accountHomeId + "]"
-                        + "\n"
-                        + "Environment: [" + accountEnvironment + "]"
         );
 
+        // Since we already filtered accounts and credentials by environment, there is no need to check
+        // environment again
```
mohitc1, line 1395: This assumes caller would have done this. And current callers do that, but as an independent util method, it should still match all environment.

siddhijain: The method does not specify anywhere that it will match environment. Since the current callers already handle environment filtering, if there’s ever a need to filter by environment, a separate overloaded method can be created.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
siddhijain: Good suggestion.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
siddhijain: Blame it to Kotlin :D

siddhijain: Done

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
mohitc1: My point was "is accountHasCredential() answering correctly that account owns the credential?"

matching homeaccountId + environment is internal logic which concludes right association of account and credential.

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
siddhijain: Sure. Updated documentation to make it clear that this method only filter on homeAccountId.

mohitc1 (approved):

PR #2784, opened 2025-10-10
Take flight value for whether to show webcp flow in weview or not in brokerless scenarios. Fixes AB#3407952

somalaya:
1. In brokerless scenarios, we cannot control flighting the features.
2. WebCp in webview feature is planned to be first rolled out in brokered scenarios using flighting (slowing going from 25 to 100% in 4 weeks)
3. Once we complete step 2 above, we want to enable the feature by default in brokerless scenarios as well.
4. But in case there are any issues found, we also want to be able to provide a kill switch to calling apps to turn the feature off.
5. This PR takes a boolean value from OneAuth to decide whether to show webcp in webview or not.
6. Changes are made in OneAuth side as well for passing the value.

Fixes [AB#3407952](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3407952)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityParameters.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityParameters.kt
@@ -41,7 +41,7 @@ import com.microsoft.identity.common.java.ui.AuthorizationAgent
  * @param sourceLibraryVersion       Product version to be of library making the request
  * @param utid                       The tenant unique id, if applicable
  * @param webViewEnableSilentAuthorizationFlowTimeOutMs                 If set to a non-null value, this indicates that the flow is silent and specifies the timeout for the silent authorization flow in milliseconds.
- * @param webViewWebcpEnabled        This parameter controls whether webcp URLs should be handled within the WebView or redirected to external browser
+ * @param isWebViewWebCpEnabled        This parameter controls whether webcp URLs should be handled within the WebView or redirected to external browser
  */
 data class AuthorizationActivityParameters @JvmOverloads constructor(
     val context: Context,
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityParameters.kt
```diff
@@ -57,5 +58,6 @@ data class AuthorizationActivityParameters @JvmOverloads constructor(
      * The tenant unique id
      */
     val utid: String? = null,
-    val webViewEnableSilentAuthorizationFlowTimeOutMs: Long? = null
+    val webViewEnableSilentAuthorizationFlowTimeOutMs: Long? = null,
```
p3dr0rv, line 61: where is this param used?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -704,9 +710,13 @@ protected boolean isWebCpInWebviewFeatureEnabled(@NonNull final String originalU
         final String methodTag = TAG + ":isWebCpInWebviewFeatureEnabled";
         try {
             if (!ProcessUtil.isRunningOnAuthService(getActivity().getApplicationContext())) {
-                // Enabling webcp in webview feature for brokered flows only for now.
-                Logger.info(methodTag, "Not running on AuthService, skipping WebCP in WebView feature check.");
-                return false;
+                if (mIsWebViewWebCpEnabledInBrokerlessCase) {
```
p3dr0rv: Why don't we assign mIsWebViewWebCpEnabledInBrokerlessCase to mInWebCpFlow and then log the status?

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/AuthorizationActivityParameters.kt
somalaya: It is not something that I added but I think it is being used AuthorizationActivityFactory

fadidurah (approved):

fadidurah (approved):

PR #2787, opened 2025-10-17
Web Apps account id registry, Fixes AB#3406762

melissaahn:
[AB#3406762](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3406762)
### Summary
The WebAppsAccountIdRegistry maps account ids to sets of client ids. This class is created for the purpose of the Edge TB APIs work, where it will be used for the getToken operation (silent case where we want to check that the user has signed into this webapp before) and the SignOut operation (to remember which web apps we need to sign out of).

Broker PR for the SignOut operation using this registry will follow.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
@@ -27,6 +27,7 @@ import com.microsoft.identity.common.java.cache.IMultiTypeNameValueStorage
 import com.microsoft.identity.common.java.interfaces.IStorageSupplier
 import com.microsoft.identity.common.java.util.ObjectMapper
 import com.microsoft.identity.common.logging.Logger
+import com.squareup.moshi.Json
 import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.read
 import kotlin.concurrent.write
@@ -43,7 +44,6 @@ class WebAppsAccountIdRegistry private constructor(
     companion object {
         private val TAG = WebAppsAccountIdRegistry::class.java.simpleName
         private const val WEBAPPS_ACCOUNT_ID_REGISTRY_STORAGE_KEY = "WebAppsAccountIdRegistryStorageKey"
-        private const val MAP_JSON_KEY = "MapJsonKey"
         private val rwLock = ReentrantReadWriteLock()
 
         /**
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,185 @@
...
+ * This is used to track which web applications are associated with which accounts,
+ * enabling coordinated sign-out and management of web app sessions.
+ */
+class WebAppsAccountIdRegistry private constructor(
+    private val storage: IMultiTypeNameValueStorage
+){
+    private val TAG = WebAppsAccountIdRegistry::class.java.simpleName
+    private val MAP_JSON_KEY = "MapJsonKey"
+    private val lock = Any()
+
```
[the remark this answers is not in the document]
melissaahn: Done

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Add a client ID to the set associated with the given account ID.
+     *
+     * @param accountId The account ID.
+     * @param clientId The client ID to add.
+     */
+    fun addClient(accountId: String, clientId: String) {
+        synchronized(lock) {
```
rpdome: use ReentrantReadWriteLock?
(if you expect # of read >> # of write)

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,185 @@
...
+         * @return A new instance of [WebAppsAccountIdRegistry].
+         */
+        fun create(supplier: IStorageSupplier): WebAppsAccountIdRegistry {
+            val store = supplier.getEncryptedFileStore(WEBAPPS_ACCOUNT_ID_REGISTRY_STORAGE_KEY)
+            return WebAppsAccountIdRegistry(store)
+        }
+    }
+
+    @Volatile
+    private var cache: MutableMap<String, MutableSet<String>>? = null
```
rpdome: if we only initialize one instance of this **_per request_** (not a static object), and only have 1 read/write at most **_per request_**, then having a cache wouldn't help much. (it'll just add complexity and overhead)

What's your use case?

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
rpdome: Also

If you plan to create one object per request, then the lock should be static.

If not (think singleton pattern), this object-level synchronization is fine.

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
melissaahn: True, we are initializing this once per request, but I would say that the number of read/write will be more like 2 per request. For the sign out, we need to get all the client ids and then remove the account. For the getToken, I'm thinking it will be at least contains and potentially addClient.

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
melissaahn: Switched to using a static ReentrantReadWriteLock: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2787/commits/addd84a81c351a3f65f93c256af69a067f546fdb

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
melissaahn: Lmk if a cache still seem like overkill for the usecases

melissaahn: Removed the cache: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2787/commits/3672ba1bc0c49b672cc0983ebc6efdaaa1bf25b9

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        /**
+         * Factory method to create an instance of [WebAppsAccountIdRegistry].
+         *
+         * @param supplier The storage supplier.
+         * @return A new instance of [WebAppsAccountIdRegistry].
+         */
+        fun create(supplier: IStorageSupplier): WebAppsAccountIdRegistry {
+            val store = supplier.getEncryptedFileStore(WEBAPPS_ACCOUNT_ID_REGISTRY_STORAGE_KEY)
+            return WebAppsAccountIdRegistry(store)
```
mohitc1, line 55: >WebAppsAccountIdRegistry [](http://example.com/codeflow?start=19&length=24)

you storage has this format
MapJsonKey : {<json of map>}

why not following?
accountid1 : {<json of set>}
accountid2 : {<json of set>}
....

you would not have to read and write entire thing always.

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,171 @@
...
+        private val rwLock = ReentrantReadWriteLock()
+
+        /**
+         * Factory method to create an instance of [WebAppsAccountIdRegistry].
+         *
+         * @param supplier The storage supplier.
+         * @return A new instance of [WebAppsAccountIdRegistry].
+         */
+        fun create(supplier: IStorageSupplier): WebAppsAccountIdRegistry {
+            val store = supplier.getEncryptedFileStore(WEBAPPS_ACCOUNT_ID_REGISTRY_STORAGE_KEY)
```
mohitc1, line 54: >store [](http://example.com/codeflow?start=16&length=5)

Could you check if operations on this are already synchronized? perhaps we don't need lock here #Resolved

melissaahn: I see that the SharedPreferencesFileManager does make use of a lock. But maybe it doesn't hurt to have a lock logic in place? Also this store value returned could change to a diff implementation in the future I suppose (like something which uses jetpack data storage)

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
melissaahn: For some reason I implemented this way at first, but it doesn't really make sense when looking at it again.
I updated the structure to what you had proposed.

common/src/main/java/com/microsoft/identity/common/internal/cache/WebAppsAccountIdRegistry.kt
```diff
@@ -0,0 +1,189 @@
...
+        return ObjectMapper.serializeObjectToJsonString(set)
+    }
+
+    /**
+     * Load the account entry from storage.
+     *
+     * @param accountId The account ID to load.
+     * @return A set of client IDs associated with the given account ID. Or an empty set if none exist.
+     */
+    private fun loadAccount(accountId: String): MutableSet<String>{
```
mohitc1: >loadAccount [](http://example.com/codeflow?start=16&length=11)

nit: load/save ClientIdsForAccount()

melissaahn: Done! And also did the homeAccountId renames

melissaahn:
Added the skip consumer check tag because the Assemble consumers check is failing due to a reason outside of this PR (this PR had passed this yesterday, and it anyways only includes new changes)

mohitc1 (changes requested):
:clock1:

mohitc1 (approved):
:shipit:

PR #2793, opened 2025-10-23
Edge TB: Expose WebApps APIs, Fixes AB#3409879

melissaahn:
### Summary:

In this PR, we:
- Expose the WebApp APIs via BrokerMsalController. OneAuth is already using this class to call similar Edge APIs.
    - getSupportedWebAppContracts(String minBrokerProtocolVersion): this method is called by OneAuth to make sure that the broker can handle the request.
    - executeWebAppRequest(String Request, String minBrokerProtocolVersion): this method handles requests for GetToken, GetCookies, and SignOut.
- Implement the bundle helper methods for creating the bundle and extracting the results from the result bundle.
- Adds the bundle keys to AuthenticationConstants, where both common and broker can access them.

Related broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3256

[AB#3409879](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3409879/)


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -1044,4 +1047,38 @@ public AadDeviceIdRecord aadDeviceIdRecordFromBundle(@NonNull final Bundle resul
         }
         return aadDeviceIdRecord;
     }
+
+    /**
+     * Gets the supported web apps contract string from the result bundle.
+     * @param resultBundle The result bundle from the broker.
+     */
+    @NonNull
+    public String getSupportedWebAppsContractFromBundle(Bundle resultBundle) throws ClientException {
```
shahzaibj:
```suggestion
    public String getSupportedWebAppsContractFromBundle(final Bundle resultBundle) throws ClientException {
```

shahzaibj (approved):

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
melissaahn: Done

p3dr0rv (approved):

PR #2794, opened 2025-10-27
Reset flights between tests, Fixes AB#3413737

somalaya:
The testcase testUrlOverrideHandlesNonceRedirectUrl was failing because there was a new testcase introduced in the class where we were setting flights but we missed resetting them before the next testcase runs.
Fix : Resetting flights between the UTs
Additional fix : There was a change that was relying on the state of flights set in previous tests which was incorrect. Correcting this as well.
Fixes [AB#3413737](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3413737)


rpdome (approved):

mohitc1 (approved):
:shipit:

PR #2795, opened 2025-10-27
Add OpenTelemetry support for passkey operations, Fixes AB#3412004

p3dr0rv:
This pull request introduces OpenTelemetry tracing to the passkey (WebAuthN) flow, ensuring that all interactions with the JavaScript reply channel and WebView client are properly traced for observability. It also refactors error handling in the passkey reply channel to use exceptions rather than plain strings, and propagates the tracing context (`SpanContext`) throughout the relevant classes. Additionally, the pull request cleans up and simplifies some internal logic and improves test coverage.

[AB#3412004](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3412004)


somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -487,5 +487,15 @@ public enum AttributeName {
     /**
      * Records if current flow is in webcp flow.
      */
-    is_in_web_cp_flow
+    is_in_web_cp_flow,
+
+    /**
+     * Passkey operation type (e.g., registration, authentication).
+     */
+    passkey_operation_type,
```
melissaahn, line 495: Just to make sure, we should not expect to see the type be authentication until support is extended on the ESTS side to allow authentication, right? auth will be represented by the telemetry set for the legacy fido classes?

melissaahn (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
p3dr0rv: Yes, all operations should remain of type "registration" until ESTS decides to update their code to include authentication as well. At that point, a new type will be introduced. The idea of defining it now is simply to be prepared for when it’s needed. All authentication requests will continue to stay within the Fido span.

PR #2796, opened 2025-10-28
Add callback for OneAuth for measuring Broker Discovery Client Perf, Fixes AB#3416279

rpdome:
[AB#3416279](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3416279)

This will allow OneAuth to wire telemetry into BrokerDiscoveryClient.getActiveBroker()

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -22,6 +22,7 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.common.internal.activebrokerdiscovery
 
+import IBrokerDiscoveryClientTelemetryCallback
 import android.content.Context
 import android.os.Bundle
 import com.microsoft.identity.common.exception.BrokerCommunicationException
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -99,6 +108,8 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
 
         const val ERROR_BUNDLE_KEY = "ERROR_BUNDLE_KEY"
 
+        public val sTelemetryCallback: IBrokerDiscoveryClientTelemetryCallback? = null
```
nickbopp: Could this be a function level parameter (optional)? We can potentially be running multiple requests at a time, though the Broker object is shared. The telemetry data is per-request on our side.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -275,19 +275,36 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
+        return runBlocking {
+            return@runBlocking getActiveBrokerAsync(shouldSkipCache, telemetryCallback)
+        }
+    }
+
+    private suspend fun getActiveBrokerAsync(shouldSkipCache:Boolean,
+                                             telemetryCallback: IBrokerDiscoveryClientTelemetryCallback?): BrokerData?{
         val methodTag = "$TAG:getActiveBrokerAsync"
         classLevelLock.withLock {
             if (!shouldSkipCache) {
```
nickbopp, line 298: Can we get a callback here for when the lock has actually been acquired.

nickbopp (changes requested):
Looking for a tag to measure time waiting for lock, and then we're good.  Chase and I looked together so I'm signing off for both of us.

nickbopp (approved):

p3dr0rv (approved):

somalaya (approved):

PR #2803, opened 2025-11-04
Edge TB: changes related to GetToken, Fixes AB#3317102

melissaahn:
### Summary

Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3267

Changes related to WebApps getToken are done in this PR, and it also notably contains refactoring (such as moving some files from broker to common) and some additions to core classes such as AccountChooser and AccountChooserActivity.

To optimize the number of IPCs and repeated code, the getToken operation will be completed as follows:

1. Caller calls BrokerMsalController.ExecuteWebAppsRequest.
2. BrokerMsalController takes the request and calls the broker, which will start the ExecuteWebAppsRequestOperation.
3. ExecuteWebAppsRequestOperation takes a peek at the method and delegates to the correct sub operation.
4. For the WebAppsGetTokenSubOperation, determine if silent is possible. Otherwise, force an interactive request and send an intent back to BrokerMsalController.
5. If silent is possible, try a silent request. If a token is returned, this will get formed into a response and sent back to BrokerMsalController. If an exception is thrown, the operation determines if an interactive request should be made. If true, an intent is sent back. If false, the exception is formed into a response and sent back to BrokerMsalController.
6. Back at BrokerMsalController, if a token response or error response is received, this response string is returned as-is. If an intent is received, BrokerMsalController will start the intent (you'll see some refactoring with the existing acquireToken method to help reusing logic). 
7. AccountChooserActivity will handle the request and knows that it is from a webapp. It will return either a token response or an error response in the protocol format.
8. BrokerMsalController returns this response as-is.

[AB#3317102](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3317102)


common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsTokenResponsePayload.kt
```diff
@@ -0,0 +1,57 @@
...
+    val userName: String,
+
+    @SerializedName(FIELD_HOME_ACCOUNT_ID)
+    val homeAccountId: String,
+
+    @SerializedName(FIELD_EXPIRES_IN)
+    val expiresIn: String,
+
+    @SerializedName(FIELD_ID_TOKEN)
+    val idToken: String,
```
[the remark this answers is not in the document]
melissaahn: I can document that this should never be logged, but is that going to be enough...

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1597,4 +1634,210 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+                    acquireToken(interactiveParams).getLocalAuthenticationResult();
+
+            final WebAppsTokenResponse response = new WebAppsTokenResponse(
+                    interactiveResult.getAccountRecord().getUsername(),
+                    interactiveResult.getUniqueId(),
+                    interactiveResult.getAccessTokenRecord().getExpiresOn(),
+                    interactiveResult.getIdToken(),
+                    interactiveResult.getAccessToken(),
+                    null
+            );
```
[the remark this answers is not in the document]
melissaahn: I will add a null check on acquireToken; I would think that if we get a non-null result for acquireToken, those properties would not be null. But I also think that the catch block would catch things.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -77,6 +77,32 @@ public String toString() {
         return this.name().toLowerCase(Locale.ROOT);
     }
 
+    /**
+     * Utility method to map string to OpenIdConnectPromptParameter
+     *
+     * @param promptParameterString
+     * @return
```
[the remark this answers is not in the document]
melissaahn: Copilot is doing a good job with these documentation comments now.

common4j/src/main/com/microsoft/identity/common/java/util/IPlatformUtil.java
```diff
@@ -67,6 +67,14 @@ public interface IPlatformUtil {
      */
     boolean isValidCallingApp(@NonNull final String redirectUri, @NonNull final String packageName);
 
+    /**
+     * Validates that the calling uid belongs to an acceptable app for web apps.
+     *
+     * @param callingUid the calling uid to validate
+     * @throws IllegalArgumentException if the calling uid is not valid
+     */
+    boolean isValidCallingAppForWebApps(@NonNull int callingUid) throws ClientException;
```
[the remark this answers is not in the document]
melissaahn: Not true- it's used in the BrokerAndroidPlatformUtils implementation.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/OpenIdConnectPromptParameter.java
```diff
@@ -77,6 +77,32 @@ public String toString() {
...
+            case "login":
+                return LOGIN;
+            case "consent":
+                return CONSENT;
+            case "create":
+                return CREATE;
+            default:
+                return UNSET;
+        }
+    }
```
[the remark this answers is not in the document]
melissaahn: Done

melissaahn:
Regarding the failing assemble consumers being caused by the validate broker step, it's failing due to a breaking change:
```
  D:\a\_work\1\s\AADAuthenticator\src\msal\java\com\microsoft\identity\client\MicrosoftAuthServiceOperation.java:251: error: unreported exception ClientException; must be caught or declared to be thrown
              parameters.validate();
                                 ^
  D:\a\_work\1\s\AADAuthenticator\src\msal\java\com\microsoft\identity\client\MicrosoftAuthServiceOperation.java:294: error: unreported exception ClientException; must be caught or declared to be thrown
              parameters.validate();
                                 ^
  D:\a\_work\1\s\AADAuthenticator\src\msal\java\com\microsoft\identity\client\MicrosoftAuthServiceOperation.java:458: error: unreported exception ClientException; must be caught or declared to be thrown
              parameters.validate();
```
              
The fix is to declare ClientException to be thrown, and that is made in the broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/3267/files#diff-e67c2e127c25d14625ad43c557f2aeb3357591fb8c965b6df8f641c880a2dc86

So this is why I will add the "skips consumer check" (once I get this PR reviewed)

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsTokenResponsePayload.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @SerializedName(FIELD_ACCESS_TOKEN)
+    val accessToken: String,
```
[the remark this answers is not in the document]
melissaahn: Resolving to consolidate to one comment

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1597,4 +1634,287 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+            }
+            final ILocalAuthenticationResult interactiveResult = rawResult.getLocalAuthenticationResult();
+            final WebAppsTokenResponse response = new WebAppsTokenResponse(
+                    interactiveResult.getAccountRecord() != null ? interactiveResult.getAccountRecord().getUsername() : null,
+                    interactiveResult.getUniqueId(),
+                    interactiveResult.getAccessTokenRecord() != null ? interactiveResult.getAccessTokenRecord().getExpiresOn() : null,
+                    interactiveResult.getIdToken(),
+                    interactiveResult.getAccessToken(),
+                    null
+            );
```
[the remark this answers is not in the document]
melissaahn: While this is a good warning, don't we already do this in our regular token operation logic? Is this really a big concern?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1597,4 +1634,287 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+     *
+     * @param webAppsRequest               web apps interactive request
+     * @param requiredParams               additional required parameters for web app request.
+     * @param minBrokerProtocolVersion     minimum broker protocol version the caller requires.
+     * @return BrokerInteractiveTokenCommandParameters
+     * @throws BaseException
+     */
+    private BrokerInteractiveTokenCommandParameters buildInteractiveTokenParametersForWebApps(@NonNull final WebAppsInteractiveRequest webAppsRequest,
+                                                                                              @NonNull final WebAppsAdditionalRequiredParameters requiredParams,
+                                                                                              @NonNull final String minBrokerProtocolVersion) throws BaseException {
```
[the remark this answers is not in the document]
melissaahn: I will be writing unit tests for this path (it will take some time) and it may not come in this PR due to the size.

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsSupportedContracts.kt
```diff
@@ -0,0 +1,45 @@
...
+        const val GET_COOKIES = "GetCookies"
+    }
+
+    override fun toString(): String {
+        return contracts.joinToString(
+            separator = ", ",
+            prefix = "[",
+            postfix = "]")
+    }
+}
```
fadidurah: new line

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsTokenResponsePayload.kt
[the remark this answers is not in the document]
fadidurah: We can override toString i think to make sure it doesn't get logged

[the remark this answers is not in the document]
fadidurah: nvm copilot already mentioned it

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsGetTokenSubOperationRequest.kt
```diff
@@ -0,0 +1,105 @@
...
+        const val FIELD_CLAIMS = "claims"
+        const val FIELD_IS_SECURITY_TOKEN_SERVICE = "isSts"
+        const val FIELD_NONCE = "nonce"
+        const val FIELD_STATE = "state"
+        const val FIELD_LOGIN_HINT = "loginHint"
+        const val FIELD_INSTANCE_AWARE = "instanceAware"
+        const val FIELD_EXTRA_PARAMETERS = "extraParameters"
+        const val DEFAULT_AUTHORITY = "https://login.microsoftonline.com/common"
+    }
+}
```
fadidurah: some files missing EOF new line

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsGetTokenSubOperationRequest.kt
```diff
@@ -0,0 +1,105 @@
...
+    // Optional; additional extra query parameters to include in the token request.
+    // Note: PoP token parameters will come through here.
+    @SerializedName(FIELD_EXTRA_PARAMETERS)
+    val extraParameters: Map<String, String>? = null
+) {
+    companion object {
+        const val FIELD_HOME_ACCOUNT_ID = "accountId"
+        const val FIELD_CLIENT_ID = "clientId"
+        const val FIELD_AUTHORITY = "authority"
+        const val FIELD_SCOPES = "scope"
```
fadidurah, line 88: scope or scopes?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1442,10 +1468,50 @@ public void performPrerequisites(@NonNull final IIpcStrategy strategy) throws Ba
...
+                                            request,
+                                            WebAppsGetTokenSubOperationEnvelope.class
+                                        );
+                                final WebAppsGetTokenSubOperationRequest getTokenRequest = envelope.getRequest();
+                                // If need to do interactive right away, do it now.
+                                // Otherwise, just let the broker handle the silent token acquisition.
+                                if (shouldForceInteractive(getTokenRequest, additionalRequiredParams.getCanShowUi())) {
+                                    mergedExtraArgs.putAll(
+                                            performInteractiveAndSerialize(envelope, additionalRequiredParams, minBrokerProtocolVersion)
+                                    );
```
mohitc1: Not clear here. Why do you need to send it to broker?
Don't you have full result already?

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
```diff
@@ -324,4 +330,8 @@ private boolean isValidHubRedirectURIForNAATests(String redirectUri) {
                 || redirectUri.equals("msauth://com.microsoft.teams/fcg80qvoM1YMKJZibjBwQcDfOno=")
                 || redirectUri.equals("https://login.microsoftonline.com/common/oauth2/nativeclient"));
     }
+
+    protected Context getContext() {
```
fadidurah, line 334: Where is this one used?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -314,4 +314,52 @@ private static List<AzureActiveDirectoryCloud> deserializeClouds(final String js
         return new Gson().fromJson(jsonCloudArray, listType);
     }
 
+    /**
+     * Builds and validates the authority from the WebApp sender URL.
```
fadidurah, line 318: Is this method only for this webApp functionality? we can rename it to include that i think. Maybe rename parameter to mention web app

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1597,4 +1677,360 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+     * @param request                       The original web app request string.
+     * @param phaseOneResultBundle          The result bundle returned from the first phase, containing the interactive request payload.
+     * @param negotiatedBrokerProtocolVersion The negotiated broker protocol version.
+     * @param minBrokerProtocolVersion      The minimum broker protocol version the caller requires.
+     * @param additionalRequiredParams      Additional required parameters for the web app request.
+     * @param additionalArgs                Additional arguments as needed; will be updated with the result.
+     * @return The result string from the broker after the second phase completes.
+     * @throws BaseException If the interactive request payload is missing or token acquisition fails.
+     */
+    private String handleInteractivePhase(@NonNull final String request,
```
fadidurah: Let's mention webapp in method name

fadidurah: "interactivePhase" can be vague i think

fadidurah (approved):

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/WebAppsGetTokenSubOperationRequest.kt
melissaahn: "scope"... just a part of the defined protocol which we don't have control over.

common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java
melissaahn: AndroidBrokerPlatformUtil.isValidCallingAppForWebApps... used for validating that the caller app is in our designated allowlist

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
melissaahn: Done... Now named buildAndValidateAuthorityFromWebAppSender.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: This method has since been taken out, but I'll take this feedback for other helper methods

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1379,22 +1379,32 @@ public static String computeMaxHostBrokerProtocol() {
         /**
          * String for broker webapps get contracts result.
          */
-        public static final String BROKER_WEBAPPS_GET_CONTRACTS_RESULT = "contracts";
+        public static final String BROKER_WEBAPPS_GET_CONTRACTS_RESULT = "web_apps_contracts";
```
melissaahn, line 1382: Just making these values more specific

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
```diff
@@ -281,4 +283,15 @@ private static final class SerializedNames {
...
+    @Nullable
+    @SerializedName(SerializedNames.REQUEST_TYPE)
+    private String mRequestType;
+
+    /**
+     * State for web apps requests. Make sure not to log this.
+     */
+    @Nullable
+    @SerializedName(SerializedNames.WEB_APPS_STATE)
+    private String mWebAppsState;
```
melissaahn, line 296: To allow AccountChooserActivity to return a result as a string we can return directly to the caller (without BrokerMsalController interfering too much), we pass the state webapp parameter so that AccountChooserActivity can include it in the result.
I think this is also the more correct approach vs BrkerMsalController injecting it at the end, as the state parameter should be included at the time when the token result is getting created.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -351,14 +372,47 @@ private String tryGetNegotiatedProtocolVersionFromHelloCache(
     public AcquireTokenResult acquireToken(final @NonNull InteractiveTokenCommandParameters parameters)
             throws BaseException, InterruptedException, ExecutionException {
         final String methodTag = TAG + ":acquireToken";
+        final AcquireTokenResult result;
+        try {
+            final Bundle resultBundle = acquireTokenInternal(parameters);
```
melissaahn: I factored out logic into acquireTokenInternal in order to allow executeWebAppsRequest to reuse interactive request logic and inject the intent directly + get the custom webapp result directly from the bundle. The acquireToken method is not changed otherwise.

common/src/main/java/com/microsoft/identity/common/internal/util/WebAppsUtil.kt
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft Corporation.
```
melissaahn, line 1: One of the classes moved from broker to common.

common4j/src/main/com/microsoft/identity/common/java/commands/webapps/MatsProperties.kt
```diff
@@ -0,0 +1,39 @@
...
+package com.microsoft.identity.common.java.commands.webapps
+
+import com.google.gson.annotations.SerializedName
+
+/**
+ * Holds the MATS data blob which will be passed to MSAL JS for telemetry purposes.
+ * Note: This class isn't going to be used until the design for the MATS blob is complete.
+ * Once that is complete, we will need to pass this data for each error object and getToken suboperation.
+ */
+data class MatsProperties(
```
melissaahn, line 32: This is being created for the future when the design for the MatsProperties schema is complete.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1425,60 +1450,126 @@ public void putValueInSuccessEvent(@NonNull final ApiEndEvent event,
...
+
+                            final String result = mResultAdapter.getExecuteWebAppRequestResultFromBundle(resultBundle);
+                            if (resultBundle.containsKey(BROKER_WEB_APPS_ERROR_RESULT)
+                                    && subMethod.equals(WebAppsSupportedContracts.GET_TOKEN)) {
+                                final WebAppsGetTokenSubOperationEnvelope envelope =
+                                        ObjectMapper.deserializeJsonStringToObject(
+                                                request,
+                                                WebAppsGetTokenSubOperationEnvelope.class
+                                        );
+                                final WebAppsGetTokenSubOperationRequest getTokenRequest = envelope.getRequest();
```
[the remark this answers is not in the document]
melissaahn: Implemented something similar

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1600,4 +1693,148 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+        final Map<String, String> extraQueryParamsMap = tempMap != null
+                ? tempMap
+                : Collections.emptyMap();
+        final List<Map.Entry<String, String>> queryParams = new ArrayList<>(extraQueryParamsMap.size() + 1);
+        for (Map.Entry<String, String> e : extraQueryParamsMap.entrySet()) {
+            queryParams.add(new java.util.AbstractMap.SimpleEntry<>(e.getKey(), e.getValue()));
+        }
+
+        if (!StringUtil.isNullOrEmpty(webAppsRequest.getNonce())) {
+            queryParams.add(new java.util.AbstractMap.SimpleEntry<>(PRT_NONCE, webAppsRequest.getNonce()));
```
mohitc1: >PRT_NONCE [](http://example.com/codeflow?start=68&length=9)

this isn't query parameter right?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1600,4 +1693,148 @@ private void verifyBrokerVersionIsSupported(@Nullable final Bundle resultBundle,
...
+
+        Map<String, String> tempMap = webAppsRequest.getExtraParameters();
+        final Map<String, String> extraQueryParamsMap = tempMap != null
+                ? tempMap
+                : Collections.emptyMap();
+        final List<Map.Entry<String, String>> queryParams = new ArrayList<>(extraQueryParamsMap.size() + 1);
+        for (Map.Entry<String, String> e : extraQueryParamsMap.entrySet()) {
+            queryParams.add(new java.util.AbstractMap.SimpleEntry<>(e.getKey(), e.getValue()));
+        }
+
```
mohitc1: simplify to

 final List<Map.Entry<String, String>> queryParams = new ArrayList<>(extraQueryParamsMap.entrySet());

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1425,60 +1450,128 @@ public void putValueInSuccessEvent(@NonNull final ApiEndEvent event,
...
+                if (shouldForceInteractiveRequestForWebApp(getTokenRequest, additionalRequiredParams.getCanShowUi())) {
+                    AzureActiveDirectory.buildAndValidateAuthorityFromWebAppSender(envelope.getSender());
+                    final BrokerInteractiveTokenCommandParameters interactiveParams =
+                            buildInteractiveTokenParametersForWebApps(getTokenRequest, additionalRequiredParams, minBrokerProtocolVersion);
+                    final Bundle resultBundle = acquireTokenInternal(interactiveParams);
+                    return mResultAdapter.getExecuteWebAppRequestResultFromBundle(resultBundle);
+                }
+            } else {
+                envelope = null;
+            }
```
mohitc1: nit: just initialize with null

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
```diff
@@ -1425,60 +1450,128 @@ public void putValueInSuccessEvent(@NonNull final ApiEndEvent event,
...
+                            if (resultBundle == null) {
+                                throw mResultAdapter.getExceptionForEmptyResultBundle();
+                            }
+                            verifyBrokerVersionIsSupported(resultBundle, minBrokerProtocolVersion);
+
+                            final String result = mResultAdapter.getExecuteWebAppRequestResultFromBundle(resultBundle);
+                            if (resultBundle.containsKey(BROKER_WEB_APPS_ERROR_RESULT)
+                                    && envelope != null) {
+                                final WebAppsGetTokenSubOperationRequest getTokenRequest = envelope.getRequest();
+                                if (canFallbackToInteractiveRequestForWebApp(getTokenRequest, additionalRequiredParams.getCanShowUi())) {
```
mohitc1: should all errors fallback to interactive? In normal AT/ATS it's  invalid_grant and interaction_required

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: I think though since it's being used in an inner method, if I were to get rid of this else clasue and just init in the beginning with no final, that would not be allowed?

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: I am double checking with the iOS folks as to how it's being done on their side.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: You're right, and as discussed, I'm taking this logic out; nonce logic will come in a diff PR.

common/src/main/java/com/microsoft/identity/common/internal/controllers/BrokerMsalController.java
melissaahn: 9b46d8884a8a39c7b28dadabca4ad59874b91426 <- made changes here

melissaahn:
image
We can see here that the assemble consumers check is failing due to the broker and linux check, but this is because I'm moving some classes from broker to common (and thus some references are getting broken). The broker PR shows all parts passing:
image
image

common/src/main/java/com/microsoft/identity/common/internal/request/MsalBrokerRequestAdapter.java
```diff
@@ -138,6 +141,11 @@ public BrokerRequest brokerRequestFromAcquireTokenParameters(@NonNull final Inte
             brokerRequestBuilder.signInWithGoogleCredential(androidInteractiveTokenCommandParameters.getSignInWithGoogleCredential());
         }
 
+        if (parameters instanceof BrokerInteractiveTokenCommandParameters) {
+            brokerRequestBuilder.requestType(((BrokerInteractiveTokenCommandParameters) parameters).getRequestType().name());
+            brokerRequestBuilder.webAppsState(((BrokerInteractiveTokenCommandParameters) parameters).getWebAppsState());
```
mohitc1, line 146: is this needed now?

melissaahn: Yes; this is a part of the work to have AccountChooserActivity return a custom webapps protocol response. the request type (WEB_APPS) is how AccountChooserActivity knows it's a webapp request, and webapp state is a parameter which needs to be passed through the request and returned in the response

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
mohitc1, line 296: are these needed now?

mohitc1 (approved):
Left suggestion

common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java
melissaahn: Offline we talked about instead creating the command parameters within AccountChooserActivity instead; agreed that this is not a blocker but an optimization I should do if I have time, or next release.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
```diff
@@ -122,6 +117,19 @@ public void validate() throws ArgumentException {
                     "mCallerPackageName", "Broker Account is null"
             );
         }
-
+        final IPlatformUtil platformUtil = getPlatformComponents().getPlatformUtil();
+        if (!CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.DISABLE_WEB_APPS_API)
+                && getRequestType() == BrokerRequestType.WEB_APPS) {
+            // For web apps, we have a different redirect URI from our standard Android one.
```
p3dr0rv: nit: give more context to reader, what is the difference?

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
```diff
@@ -118,7 +125,14 @@ public void validate() throws ArgumentException {
                         "OAuth2Cache not an instance of BrokerOAuth2TokenCache"
                 );
             }
-            if (!getPlatformComponents().getPlatformUtil().isValidCallingApp(getRedirectUri(), getCallerPackageName())) {
+            final IPlatformUtil platformUtil = getPlatformComponents().getPlatformUtil();
```
p3dr0rv, line 128: logic seems duplicated, feels like validation can be contain in abstract.

p3dr0rv (approved):
ensure you have enough test for all the changes introduced.

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerInteractiveTokenCommandParameters.java
melissaahn: I want to make sure that this calling app check and flight stay close to the existing app validation logic without moving the existing redirect uri/app validation logic, so I won't move this around for now (may consider an optimization later)

common4j/src/main/com/microsoft/identity/common/java/commands/parameters/BrokerSilentTokenCommandParameters.java
melissaahn: Noted: 09792308737421739c1cd1f4043ad01f4b331d7b

mohitc1 (approved):

PR #2809, opened 2025-11-10
Update Lab API Calling Pattern, Fixes AB#3424041

fadidurah:
The intermediate layer between our client code and the back end Lab api functions is no longer working, working theory is an azure security policy that was enforced. This PR updates our calling pattern to make it so we interface directly with the lab api's backend. The PR also introduces a Key Vault api in our code to access key vault secrets by using the automation certificate.

Validation runs:
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1555631&view=results
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1555744&view=results

[AB#3424041](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3424041)

### DIFF
--- a/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/PolicyHelper.java
+++ b/testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/PolicyHelper.java
@@ -83,7 +83,7 @@ public class PolicyHelper {
      * @param policy Disable Policy can be used for GlobalMFA, MAMCA, MDMCA, MFAONSPO, MFAONEXO. (optional)
      * @return boolean value indicating policy is disabled or not for the upn.
      */
-    public boolean disablePolicy(@NonNull final String upn, @NonNull final String policy) {
+    public boolean disablePolicy(@NonNull final String upn, @NonNull final String policy) throws LabApiException {
         instance.setupApiClientWithAccessToken();
 
         try {
@@ -98,8 +98,6 @@ public class PolicyHelper {
         } catch (final ApiException e) {
             Logger.error(TAG," Bad Request : Disable Policy can be used only for Locked users. ",e);
             throw new AssertionError(e);
-        } catch (LabApiException e) {
-            throw new RuntimeException(e);
         }
     }
 }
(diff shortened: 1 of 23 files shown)


testutils/src/main/java/com/microsoft/identity/internal/testutils/labutils/PolicyHelper.java
```diff
@@ -55,15 +57,21 @@ public class PolicyHelper {
...
+
             final CustomSuccessResponse enablePolicyResult = enablePolicyApi.apiEnablePolicyPut(upn, policy);
             final String expectedResult = (policy +" Enabled for user : " + upn).toLowerCase();
             Assert.assertNotNull(enablePolicyResult);
             return enablePolicyResult.toString().toLowerCase().contains(expectedResult);
         } catch (final ApiException e) {
             Logger.error(TAG,"Bad Request : Enable Policy can be used only for Locked users.",e);
             throw new AssertionError(e);
+        } catch (LabApiException e) {
+            throw new RuntimeException(e);
```
p3dr0rv: nit: log?

fadidurah: this was removed

p3dr0rv (approved):
Overall, it looks good. However, I believe that getKeyVaultSecret should not be part of LabClient, as this causes changes to propagate through multiple layers of the code and results in repetitive patterns. It might be better to add it to the APIClient or another appropriate class.
It might be worth checking with the Copilot agent to see if it has any recommendations on how to avoid this issue.

rpdome (approved):

fadidurah:
> Overall, it looks good. However, I believe that getKeyVaultSecret should not be part of LabClient, as this causes changes to propagate through multiple layers of the code and results in repetitive patterns. It might be better to add it to the APIClient or another appropriate class. It might be worth checking with the Copilot agent to see if it has any recommendations on how to avoid this issue.

I think this is fair. reason i have it in LabClient is because we are using it for the purposes of calling these lab apis. If we wanted to call it outside of lab client we can still do that, but we don't really implement ApiClient for anything else and we would have to add that implementation if we wanted to call the function outside of LabClient. Right now we just don't have a reason to

rpdome (approved):

PR #2817, opened 2025-11-18
Cache Active Broker In Memory (BrokerDiscoveryClient), Fixes AB#3429928

rpdome:
[AB#3429928](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3429928)

expose a new method for oneauth, ```getActiveBrokerWithInMemoryCache()```, which will cache the result in memory for the process lifetime.


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -274,112 +283,207 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
+
+        // Fall back to AccountManager (for older brokers).
+        cache.setShouldUseAccountManagerForTheNextMilliseconds(
+            TimeUnit.MINUTES.toMillis(
+                60
+            )
+        )
+
+        return getFromAccountManagerAndCacheResult(telemetryCallback)
     }
```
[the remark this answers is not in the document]
rpdome: Intentional. We won't cache accountmanager result, but we'll set up the flag setShouldUseAccountManagerForTheNextMilliseconds, which will skip IPC for the next 60 seconds.

rpdome:
todo: based on a discussion with nick

expose a new function "getActiveBrokerWithInMemoryCache" which will 
1. use in memory cache
2. only perform signing cert check (skip everything else).

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -273,9 +289,59 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
...
+     * Make sure the [BrokerData] that stays in the memory cache (if any) is valid.
+     **/
+    private fun validateInMemoryCacheValue(
+        data: BrokerData,
+        telemetryCallback: IBrokerDiscoveryClientTelemetryCallback?,
+    ): Boolean {
+        val timeStartIsValidBroker = System.nanoTime()
+        val isValidBroker = isValidBroker(data)
+        telemetryCallback?.onFinishCheckingIfValidBroker(System.nanoTime() - timeStartIsValidBroker)
+        return isValidBroker
```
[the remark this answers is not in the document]
rpdome: This is intentional. we're stripping "non-P0" check in this call for perf optimization.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -273,9 +289,59 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
         }
     }
 
+    override fun getActiveBrokerWithInMemoryCache(telemetryCallback: IBrokerDiscoveryClientTelemetryCallback?): BrokerData?{
+        cachedData?.let {
+            if (it.brokerData == null) {
```
siddhijain, line 294: brokerData == null means no active broker was found the last time. Should we not do a broker discovery again?

siddhijain: never mind. just read that the cached data will be used for the lifetime of an app.

somalaya, line 294: What happens when let's say teams app is active, brokerData is null and we install AuthApp at that time? Will it not be doing broker discovery? I believe it will still do a brokered auth using legacy active broker app mechanism?

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -273,9 +289,59 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
         }
     }
 
+    override fun getActiveBrokerWithInMemoryCache(telemetryCallback: IBrokerDiscoveryClientTelemetryCallback?): BrokerData?{
```
somalaya, line 292: Do you plan to use flight (defaulted to true) to decide if in memory cache should be used or do we not think that this would need a flight?

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
rpdome: It won't. OneAuth will keep using local auth for the process lifetime.

This was an agreed upon trade off for the perf improvement.

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
rpdome: Not in our layer. This will be in OneAuth side (they can use the flight to determine if they want to use the old API or the new API)

siddhijain (approved):

nickbopp (approved):

somalaya (approved):

PR #2818, opened 2025-11-18
Enable Broker Discovery by default in MSAL/Broker API, Fixes AB#3430520

rpdome:
[AB#3430520](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3430520)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
@@ -38,12 +38,7 @@ import kotlinx.coroutines.sync.withLock
 class BrokerDiscoveryClientFactory {
 
     companion object {
-
-        private val TAG = BrokerDiscoveryClientFactory::class.simpleName
-
-        @Volatile
-        private var IS_NEW_DISCOVERY_ENABLED = false
-
+        
         @Volatile
         private var clientSdkInstance: IBrokerDiscoveryClient? = null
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
```diff
@@ -56,25 +51,11 @@ class BrokerDiscoveryClientFactory {
         private val lock = Mutex()
 
         /**
-         * If set to true, the new Broker discovery mechanism will be enabled.
-         * This is currently turned off by default - until we're ready to ship the feature.
+         * Do nothing, since this is always enabled now.
+         * Keep it for now to prevent breaking OneAuth.
```
fadidurah, line 55: We should note that developers cannot use this to turn the feature off. Not sure if anyone was doing that to begin with, but doesn't hurt lol

also, can we deprecate this method?

fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
rpdome: Let me talk with nick first whether marking this as deprecate (or removing it entirely) is okay.

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClientFactory.kt
rpdome: I will refrain from adding deprecated for now. Don't wanna deal with potential pipeline issues.

PR #2819, opened 2025-11-18
Determine whether broker app opts out from battery optimization, Fixes AB#3428706

rpdome:
[AB#3428706](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3428706)

The value would also be cached in memory (since it should rarely change).
The caller can invoke this method right after it gets the package name via BrokerDiscoveryClient.getActiveBroker().

(NOTE: OneAuth would **_also_** need to invoke getDeviceIdleMode(), which will return the device status. 
These two values shall be emitted together.
For that method, given that that value fluctuates based on usage, we are not going to cache it)


common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -130,8 +136,43 @@ public String getPowerOptimizationSettings(@NonNull final Context context){
      * @param connectionContext Context used to query if app is ignoring battery optimizations.
      * @return true if the given application package name is on the device's power allow list.
      */
```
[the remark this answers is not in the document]
rpdome: This is no longer needed since our min sdk is 24.

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -130,8 +136,43 @@ public String getPowerOptimizationSettings(@NonNull final Context context){
...
+                return "SecurityException";
+            } catch (IllegalArgumentException e) {
+                Logger.error(methodTag, "IllegalArgumentException when checking battery optimization status for package: " + packageName, e);
+                return "IllegalArgumentException";
+            } catch (Exception e) {
+                Logger.error(methodTag, "Unknown Exception when checking battery optimization status for package: " + packageName, e);
+                return "UnknownException";
+            }
+        });
+    }
```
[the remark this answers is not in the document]
rpdome: No longer applicable since our minSdk is 24.

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -130,8 +136,48 @@ public String getPowerOptimizationSettings(@NonNull final Context context){
...
+        return batteryOptOutCache.computeIfAbsent(packageName, key -> {
+            try {
+                final PowerManager powerManager = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+                if (powerManager == null) {
+                    Logger.error(methodTag, "PowerManager is null for package: " + packageName, null);
+                    return "CannotRetrievePowerManager";
+                }
+
+                if (powerManager.isIgnoringBatteryOptimizations(key)) {
+                    return "OptOut";
```
siddhijain: nit - move this to a constant file / enum.

siddhijain: Can we return boolean true / false instead?

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -130,8 +136,48 @@ public String getPowerOptimizationSettings(@NonNull final Context context){
...
+     *
+     * @param packageName The package name to check.
+     * @param context The context to use for PowerManager.
+     * @return "OptOut" if the app is opted out, "NotOptOut" if not, or exception type string.
+     */
+    public String isAppOptedOutFromBatteryOptimization(@NonNull final String packageName, @NonNull final Context context) {
+        final String methodTag = TAG + ":isAppOptedOutFromBatteryOptimization";
+
+        return batteryOptOutCache.computeIfAbsent(packageName, key -> {
+            try {
```
siddhijain, line 194: When will the cache be invalidated? If an app has opted out of battery optimization but its status changes later, how will we know?

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
rpdome: boolean true / false instead << considered this, but i think this will also tell whether this android API is reliable enough for us to use as a signal as well. (we don't want to send an error as "not opted out")

(@nickbopp will send out the doc soon, but basically for Jan, we're thinking of collecting data only before making riskier moves - like using these two flags to determine if broker should be used)

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
rpdome: this is by design. the value shall remain the same until the next app boot.

(we expect that the opt in -> opt out would rarely happen - once per app install)

siddhijain (approved):

nickbopp (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -99,6 +109,40 @@ public String getDeviceIdleMode(@NonNull final Context context){
...
+     * Gets the Device Doze Mode Status.
+     * 
+     * This is exposed to OneAuth.
+     *
+     * @param context The context to use for PowerManager.
+     * @return a {@link DeviceDozeModeStatus}
+     */
+    @NonNull
+    public DeviceDozeModeStatus getDeviceDozeModeStatus(@NonNull final Context context){
+        final String methodTag = TAG + ":getDeviceIdleMode";
```
mohitc1: >getDeviceIdleMode [](http://example.com/codeflow?start=41&length=17)

tag

common/src/main/java/com/microsoft/identity/common/internal/DeviceDozeModeStatus.kt
```diff
@@ -0,0 +1,31 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.adal.internal
+
+enum class DeviceDozeModeStatus {
```
mohitc1, line 29: nit: doc

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
```diff
@@ -130,8 +170,41 @@ public String getPowerOptimizationSettings(@NonNull final Context context){
...
+     * Caches the result in memory using computeIfAbsent for thread safety.
+     * Returns a string indicating the result or exception type.
+     *
+     * This is exposed to OneAuth.
+     *
+     * @param packageName The package name to check.
+     * @param context The context to use for PowerManager.
+     * @return a {@link BatteryOptimizationStatus}
+     */
+    public BatteryOptimizationStatus isAppOptedOutFromBatteryOptimization(@NonNull final String packageName, @NonNull final Context context) {
```
mohitc1, line 190: >packageName [](http://example.com/codeflow?start=96&length=11)

PowerManager does not need the package name looking at the implementation. Are context and package for same app? can we just take context as input then? #Resolved

common/src/main/java/com/microsoft/identity/common/adal/internal/BatteryOptimizationStatus.kt
```diff
@@ -0,0 +1,33 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.common.adal.internal
```
mohitc1: >adal.internal [](http://example.com/codeflow?start=38&length=13)

nit: can this be in different non-adal package?

common/src/main/java/com/microsoft/identity/common/adal/internal/PowerManagerWrapper.java
rpdome: the intention here is to have OneAuth query broker app, so package name and context owner are different.

mohitc1 (approved):
:shipit:

mohitc1 (approved):

PR #2820, opened 2025-11-19
revert function key append, update HTTP request methods, Fixes AB#3430779

fadidurah:
We no longer need to pass the functions keys when calling lab api, just having the certificate token in headers is enough when making the request. All Lab APIs are now working again, needed to update their http request methods as well, all should use POST.

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1560386&view=results

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1560092&view=results

[AB#3430779](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3430779), 3429582

### DIFF
--- a/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
+++ b/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
@@ -229,10 +229,7 @@ public class LabClient implements ILabClient {
                 mLabApiAuthenticationClient.getAccessToken()
         );
 
-        final String createTempUserFunctionCode = getKeyVaultSecret(
-                CreateTempUserApi.AZURE_FUNCTION_CODE_SECRET_NAME
-        );
-        final CreateTempUserApi createTempUserApi = new CreateTempUserApi(createTempUserFunctionCode);
+        final CreateTempUserApi createTempUserApi = new CreateTempUserApi();
         createTempUserApi.getApiClient().setReadTimeout(TEMP_USER_API_READ_TIMEOUT);
         final TempUser tempUser;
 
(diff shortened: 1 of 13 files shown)


LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
```diff
@@ -329,14 +326,10 @@ public boolean deleteDevice(@NonNull final String upn,
...
         );
-
-        final String deleteDeviceFunctionCode = getKeyVaultSecret(
-                DeleteDeviceApi.AZURE_FUNCTION_CODE_SECRET_NAME
-        );
-        final DeleteDeviceApi deleteDeviceApi = new DeleteDeviceApi(deleteDeviceFunctionCode);
+        final DeleteDeviceApi deleteDeviceApi = new DeleteDeviceApi();
 
         try {
-            final CustomSuccessResponse successResponse = deleteDeviceApi.apiDeleteDeviceDelete(
```
melissaahn, line 339: Out of curiosity, when is CustomSuccessResponse expected as a result vs a general string?

melissaahn (approved):

LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/client/LabClient.java
fadidurah: These Response classes are generated code, from a few years ago, they map to a particular json schema that lab was sending at one point. In the case of delete Device, that schema is no longer there due to bypassing the msidlabs layer, it's just sending us a string. This can should probably be changed on lab's side, but i don't really think such a change is worth it as it may cause another outage.

rpdome (approved):

PR #2826, opened 2025-12-03
Bug fix : Selecting text in login page is causing the input field to be hidden, Fixes AB#3447997

somalaya:
**Problem**: Adding android:background to an Activity theme causes input text fields to disappear when selected as reported in this bug https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3383676

**Root Cause**: The android:background attribute sets a background drawable at the view level, which interferes with EditText's own background drawable that handles the focus state, underline, and cursor rendering.

**Solution**: Changed from android:background to android:windowBackground. The android:windowBackground attribute sets the background at the window level (behind all views) without affecting individual view backgrounds, allowing EditText fields to maintain their focus states and remain visible.
The style now correctly uses android:windowBackground which will set the white background for the entire activity window without interfering with input fields' rendering and focus behavior.

Fixes [AB#3447997](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3447997)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -575,6 +575,7 @@ Version 6.0.0
 - [PATCH] Added Base64 encode for the AuthorizationRequest state parameter (#1750)
 - [PATCH] Fix missing authority_url when creating the authority audience (#1753)
 - [PATCH] Add filter for broker telemetry event fields. (#1793)
+- [MINOR] Bug fix : Selecting text in login page is causing the input field to be hidden (#2826)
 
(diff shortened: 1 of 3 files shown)


richardtz12 (approved):

mohitc1:
Add changelog

rohitluthra (approved):

rohitluthra (approved):

PR #2830, opened 2025-12-06
Move our Libraries to only use artifact feed for packages, no Maven Central, Fixes AB#3395409

fadidurah:
Received an s360 item asking our libraries to always pull packages from feeds, and not declare mavenCentral as a repository to source dependencies. The feed itself has maven as an upstream. We cannot use AndroidADAL feed as that one has some versions deleted (once a version is deleted from a feed, it cannot be restored, even if an upstream has it). Created a new feed to support this https://identitydivision.visualstudio.com/Engineering/_artifacts/feed/NewAndroid. We pull packages from this new feed, but we still publish our versions to AndroidADAL. The new feed has AndroidADAL as an upstream still, so our existing collection of artifacts is still accessible.

I also took this chance to consolidate the Maven VSTS Username and Access Token fields, which are fields used to authenticate and pull artifacts from the feed. Previously, each library had their own names for these fields with an identifier based on the library in question. These PRs make it so all libraries use the same names, `ENV_VSTS_MVN_CRED_USERNAME` and `ENV_VSTS_MVN_CRED_ACCESSTOKEN`.

**Validation**
Weekly: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1568877&view=results (LTW msal automation app failed because ltw branch does not have these changes, will work once this is in dev and we rebase.
Consumers: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1568839&view=results
Test Library: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1568872&view=results

[AB#3395409](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3395409)

### DIFF
--- a/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/authentication/LabApiAuthenticationClient.java
+++ b/LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/authentication/LabApiAuthenticationClient.java
@@ -51,7 +51,6 @@ public class LabApiAuthenticationClient implements IAccessTokenSupplier {
     private final static int ATTEMPT_RETRY_WAIT = 3;
     private final String mLabCredential;
     private final String mLabCertPassword;
-    private final String defaultScope = LabConstants.DEFAULT_LAB_SCOPE;
     private final String mClientId;
 
     public LabApiAuthenticationClient(@NonNull final String labSecret) {
@@ -124,7 +123,7 @@ public class LabApiAuthenticationClient implements IAccessTokenSupplier {
         if (customScope != null) {
             authScope = customScope;
         } else {
-            authScope = defaultScope;
+            authScope = LabConstants.DEFAULT_LAB_SCOPE;
         }
 
         final IConfidentialAuthClient confidentialAuthClient = new Msal4jAuthClient();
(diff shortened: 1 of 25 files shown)


LabApiUtilities/src/main/com/microsoft/identity/labapi/utilities/authentication/LabApiAuthenticationClient.java
```diff
@@ -124,7 +123,7 @@ private String getAccessTokenInternal(final String customScope) throws LabApiExc
         if (customScope != null) {
             authScope = customScope;
         } else {
-            authScope = defaultScope;
+            authScope = LabConstants.DEFAULT_LAB_SCOPE;
```
fadidurah, line 126: This was a spotbug, not sure how it wasn't caught in it's original PR

shahzaibj (approved):

p3dr0rv (approved):
It looks like we’re now injecting the credentials in places where we weren’t before. Why is that?

fadidurah:
> It looks like we’re now injecting the credentials in places where we weren’t before. Why is that?

You don't need credentials to access maven central. i think some places were relying on that. also, some pipelines had the username set in variables. I decided to just code it into the yaml so we don't have to maintain variables.

PR #2831, opened 2025-12-10
AB#3455870 Adds Authentication Constants to be used for broker latency timestamp in response, Fixes AB#3455870

Prvnkmr337:
[AB#3455870](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3455870) Adding Authentication constants necessary to support passing AST request latency timestamps from Broker to OneAuth.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1564,16 +1564,6 @@ public final class AuthenticationConstants {
         public static final String BROKER_APP_LINK_REDIRECT_URL_PATH_PREFIX = "androidbroker";
 
         /**
-         * Timestamp when the broker received the authentication request.
-         */
-        public static final String BROKER_REQUEST_RECEIVED_TIMESTAMP = "broker_request_received_timestamp";
-
-        /**
-         * Timestamp when the broker generated the authentication response.
-         */
-        public static final String BROKER_RESPONSE_GENERATION_TIMESTAMP = "broker_response_generation_timestamp";
-
-        /**
          * Bundle identifiers for x-ms-clitelem info.
          */
         public static final class CliTelemInfo {
--- a/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
+++ b/common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
@@ -1564,6 +1564,16 @@ public final class AuthenticationConstants {
         public static final String BROKER_APP_LINK_REDIRECT_URL_PATH_PREFIX = "androidbroker";
 
         /**
+         * Timestamp when the broker received the authentication request.
+         */
+        public static final String BROKER_REQUEST_RECEIVED_TIMESTAMP = "broker_request_received_timestamp";
+
+        /**
+         * Timestamp when the broker generated the authentication response.
+         */
+        public static final String BROKER_RESPONSE_GENERATION_TIMESTAMP = "broker_response_generation_timestamp";
+
+        /**
          * Bundle identifiers for x-ms-clitelem info.
          */
         public static final class CliTelemInfo {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1563,6 +1563,16 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String BROKER_APP_LINK_REDIRECT_URL_PATH_PREFIX = "androidbroker";
 
+        /**
+         * Timestamp when the broker received the authentication request.
+         */
+        public static final String BROKER_REQUEST_RECEIVED_TIMESTAMP = "broker_request_received_timestamp";
```
siddhijain: Adal is deprecated. The constant can be defined in the Broker itself.

mohitc1: Since this will be read in BrokerMsalController. This should be in common. But yes, better not in *adal* package.

mohitc1:
Would BrokerMSALController change be separate PR?

We should also evaluate if we need broker protocol version update and adjust logic around it. Usually, when it's change in broker API, it's recommended. Only if both BrokerMSALController and Broker are version >= new protocol version added, the new contract is used. This is to ensure on a device, if broker is old, then client side should not expect new parameters and likewise, if client side is old version it should work fine with new broker response.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
Prvnkmr337: Got it, should this change then be in **'com.microsoft.identity.common.java.AuthenticationConstants'** instead of **'com.microsoft.identity.common.adal.internal.AuthenticationConstants'**  ?

Prvnkmr337:
> Would BrokerMSALController change be separate PR?
> 
> We should also evaluate if we need broker protocol version update and adjust logic around it. Usually, when it's change in broker API, it's recommended. Only if both BrokerMSALController and Broker are version >= new protocol version added, the new contract is used. This is to ensure on a device, if broker is old, then client side should not expect new parameters and likewise, if client side is old version it should work fine with new broker response.

Yes, I was planning changes to BrokerMSALController as separate PR. Reason being I was waiting for inputs from OneAuth team on format to return latency values from BrokerMSALController to OneAuth's Broker Client.

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
mohitc1: You can put there

mohitc1:
Refer here [https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2F%5BAndroid%5D%20Broker%20API%2Fbroker_protocol_versions.md&version=GBdev&_a=preview](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2F%5BAndroid%5D%20Broker%20API%2Fbroker_protocol_versions.md&version=GBdev&_a=preview) . 

Here the changes are addition of new params in response and not new API or core functionality change. Client side can just look for them; if not found, then it can ignore. So version update is not strictly needed as nothing should break on whether they are present or not.

---
In reply to: [3647726937](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2831#issuecomment-3647726937) [](http://example.com/codeflow?ancestors=3647726937)

siddhijain (approved):

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
Prvnkmr337: Updated

mohitc1 (dismissed):
:shipit:

mohitc1 (approved):
:shipit:

mohitc1 (approved):

PR #2832, opened 2025-12-12
Fixes AB#3429389 cleanup + optimized saveAndLoadAggregatedAccountData that identifies …

siddhijain:
Fixes [AB#3429389](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3429389)

This PR cleans up unused classes and methods that have not been used and have been deprecated for the last 3 years.

It adds an optimized saveAndLoadAggregatedAccount() method that identifies the targetCache, on which save() and loadAggregatedAccountData() methods should be called, once. The current SaveAndLoadAggregatedAccount() method performs 2 disjoint operations - 1. save() -> finds the token cache (FOCI or ProcessUId), save the accounts and credentials to the cache. 2. loadAggregatedAccountData -> finds the token cache again (by querying metadata from the storage and performing multiple operations) and then, loading the aggregated account data.
This PR gets rid of the duplicate 'finding the token cache' part.


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -512,5 +512,10 @@ public enum AttributeName {
     /**
      *  Elapsed time (in milliseconds) spent in executing the loadAggregatedAccountData() method in BrokerOAuth2TokenCache.
      */
-    elapsed_time_load_aggregated_account_data
+    elapsed_time_load_aggregated_account_data,
+
+    /**
+     * Elapsed time (in milliseconds) in loading accounts and credentials to in-memory cache.
```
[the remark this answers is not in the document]
siddhijain: This classification will be done in the broker repo once i add the attribute there.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1782,4 +1703,103 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
...
+     * This method saves the provided account and credential records to the appropriate cache
+     * (FOCI or process UID cache based on familyId), then loads and returns aggregated account
+     * data including any guest tenant credentials.
+     * <p>
+     * <b>Optimization:</b> Unlike the original synchronized implementation which called
+     * {@code getTokenCacheForClient} twice (once during save, once during load), this method
+     * determines the target cache once and reuses it for both operations, reducing overhead.
+     * <p>
+     * This method does NOT use synchronization. Thread safety is provided
+     * by the underlying cache implementations (MsalOAuth2TokenCache and MicrosoftFamilyOAuth2TokenCache).
```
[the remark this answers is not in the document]
siddhijain: The current synchronized block does not protect the saveAndLoadAggregatedAccount() method from multiple threads since the lock is acquired at the object level and a new BrokerOuth2TokenCache instance is created by each thread. So the synchronized block is useless in the current implementation. That is why this PR got rid of it. 
There is already a work item in place that will take care of all the potential issues identified above. https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3407692

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
```diff
@@ -1365,4 +1243,549 @@ public IFlightsProvider getFlightsProvider() {
...
+                        SESSION_KEY,
+                        CredentialType.IdToken
+                );
+
+        mFociCredentialCache.saveAccount(guestBundle.mGeneratedAccount);
+        mFociCredentialCache.saveCredential(guestBundle.mGeneratedIdToken);
+
+        configureMocksForFoci();
+        final List<ICacheRecord> result = mBrokerOAuth2TokenCache.saveAndLoadAggregatedAccountData(
+                mockStrategy, mockRequest, mockResponse);
```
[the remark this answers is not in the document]
siddhijain: updated.

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -187,6 +187,11 @@ public enum CommonFlight implements IFlightConfig {
      */
     USE_IN_MEMORY_CACHE_FOR_ACCOUNTS_AND_CREDENTIALS("UseInMemoryCacheForAccountsAndCredentials", false),
 
+    /**
+     * Flight to control whether or not to use in memory cache for accounts and credentials.
```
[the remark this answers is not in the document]
siddhijain: updated

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1782,4 +1703,103 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
...
+                clientId,
+                applicationIdentifier,
+                mamEnrollmentIdentifier,
+                target,
+                cacheRecord.getAccount(),
+                authScheme
+        );
+        OTelUtility.recordElapsedTime(AttributeName.elapsed_time_cache_save_and_load_aggregated_account_data.name(),
+                saveAndLoadStartTime);
+        return listICacheRecord;
```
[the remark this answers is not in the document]
siddhijain: updated

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1782,4 +1703,103 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
...
+            // Save to the foci cache....
+            targetCache = mFociCache;
+            Logger.info(TAG + methodName, "Saving data to FOCI cache");
+        } else {
+            // Save to the processUid cache... or create a new one
+            targetCache = initializeProcessUidCache(
+                    getComponents(),
+                    mUid
+            );
+            Logger.info(TAG + methodName, "Saving data to Process Uid cache");
```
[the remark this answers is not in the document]
siddhijain: The optimized method intentionally uses [mUid] directly, which is consistent with the original [save()] method (line 233). 

This actually eliminates a discrepancy between save() and load() where they could target different caches. The [updateApplicationMetadataCache()] call ensures metadata is updated with [mUid], making subsequent operations consistent. 

The suggested changes will actually negate the optimization.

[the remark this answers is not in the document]
siddhijain: @mohitc1 I would like to know your thoughts here. From my research (and that is where this optimization came from), both save and loadAggregatedAccountData() operations should be performed on the same cache (FOCI or processUid). Is this a real world scenario where we save to one cache but load data from the other?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
```diff
@@ -79,6 +81,7 @@ public SharedPreferencesAccountCredentialCacheWithMemoryCache(
 
     private void load() {
```
mohitc1, line 82: >load [](http://example.com/codeflow?start=17&length=4)

load() is called on different thread. I doubht (SpanExtension.current()) would track it

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -325,32 +253,25 @@ public ICacheRecord save(@NonNull AccountRecord accountRecord,
         return result;
     }
 
-    @Deprecated
-    public synchronized List<ICacheRecord> saveAndLoadAggregatedAccountData(
```
mohitc1, line 329: >saveAndLoadAggregatedAccountData [](http://example.com/codeflow?start=43&length=32)

Appears that similar method in MsalOauth2TokenCache can also be deleted. Therefore MsalOauth2TokenCache.save() also.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1782,4 +1703,109 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
...
+     * @return A List of ICacheRecords containing the saved account and all associated credentials
+     *         across tenants (home + guest). May include multiple records if guest tenant tokens exist.
+     * @throws ClientException If an error occurs during save or load operations.
+     */
+    private List<ICacheRecord> saveAndLoadAggregatedAccountDataOptimized(
+            @NonNull AccountRecord accountRecord,
+            @NonNull IdTokenRecord idTokenRecord,
+            @NonNull AccessTokenRecord accessTokenRecord,
+            @Nullable RefreshTokenRecord refreshTokenRecord,
+            @Nullable String familyId,
```
mohitc1: all finals?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCacheWithMemoryCache.java
siddhijain: Good point. I will remove this change from the PR and think about a better way to track this.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: Removed

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: done

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -1782,4 +1703,109 @@ private MsalOAuth2TokenCache getTokenCacheForClient(@NonNull final String client
...
+            final @NonNull AbstractAuthenticationScheme authScheme
+    ) throws ClientException{
+        final String methodName = ":saveAndLoadAggregatedAccountDataOptimized(accountRecord, idTokenRecord, accessTokenRecord, " +
+                "refreshTokenRecord, familyId, authScheme)";
+
+        final ICacheRecord cacheRecord;
+        final long saveAndLoadStartTime = System.currentTimeMillis();
+
+        final boolean isFoci = !StringUtil.isNullOrEmpty(familyId);
+        final MsalOAuth2TokenCache targetCache;
```
mohitc1: We don't any synchronization here?

siddhijain: No. I’ll implement locking at the cache layer. The current method’s synchronization is ineffective because each thread creates its own BrokerOAuth2TokenCache instance, so there’s no shared state to protect.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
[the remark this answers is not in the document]
mohitc1: I don't have prior knowledge. But reading through code, yes, they should be same. You should read cache what you just wrote to. So, this looks fine to me.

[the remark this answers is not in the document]
mohitc1: one question though, is this change saving cost on storage I/O?

For foci, I don't see advantage
For non-Foci, it's creating new MSALOAuth2Cache, object but are there extra I/O?

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
```diff
@@ -1365,4 +1243,488 @@ public IFlightsProvider getFlightsProvider() {
...
+        boolean foundMetadata = false;
+        for (BrokerApplicationMetadata metadata : allMetadata) {
+            if (CLIENT_ID.equals(metadata.getClientId()) &&
+                    ENVIRONMENT.equals(metadata.getEnvironment())) {
+                foundMetadata = true;
+                assertEquals("Metadata should have family ID", "1", metadata.getFoci());
+                break;
+            }
+        }
+        assertTrue("Should find metadata for the saved client", foundMetadata);
```
mohitc1, line 1336: Move to separate internal method and make both testSaveAndLoadAggregatedAccountData_FociApp_FlightDisabled and testSaveAndLoadAggregatedAccountData_FociApp_Enabled call it.

This way we strictly enforce in code that method's input and output is same irrespective flight enabled/disabled

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
mohitc1: It's ineffective because, callers are creating different objects. 

But from a single instance's of this class perspective, it's moving from being thread-safe to thread-unsafe. If flow wants to use same object, the flow would have to guarantee thread-safety at caller level.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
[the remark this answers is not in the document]
siddhijain: The optimized code path saves:

1 disk I/O operation (SharedPreferences read from application metadata cache)
1 JSON deserialization of the entire **application metadata** list
1 iteration through all registered application metadata entries
1 decryption operation
1 cache object resolution/initialization call

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: Not at the caller level - the locks should be applied at the lowest level possible (storage layer) not at the BrokerOAuth2TokenCache layer.

The current MsalOauth2TokenCache layer and 
SharedPreferencesAccountCredentialCache - both have locks in place (thought I will be refactoring that in the later PRs) - so a synchronization at this level was redundant.

mohitc1: okay. Does that mean, even if we operate on same instance of BrokerOAuth2Cache concurrently from two different threads, current locks/synchronization at lower layers would ensure no issues are seen?

shahzaibj: What is the storage layer? Do we mean SharedPreferencesFileManager?

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: Storage layer is SharedPreferencesAccountCredentialCache  and SharedPreferencesAccountCredentialCacheWithInMemoryCache.

siddhijain: > Does that mean, even if we operate on same instance of BrokerOAuth2Cache concurrently from two different threads, current locks/synchronization at lower layers would ensure no issues are seen?

yes - the save is protected at SharedPreferencesAccountCredentialCache layer and the load is protected at MsalOauth2TokenCache layer.

common/src/test/java/com/microsoft/identity/common/BrokerOAuth2TokenCacheTest.java
siddhijain: updated this and the other test methods.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: Discussed offline: Although the goal is to move locks and synchronization closer to the storage layer, since locks are not yet implemented at the storage level (MetadataCache layer), we will keep them here for now for completeness. They will be removed as part of the PR that adds locks at the storage layer.

mohitc1 (approved):
:shipit:

PR #2839, opened 2025-12-18
Add additional allowed origins for PasskeyWebListener, Fixes AB#3465377

p3dr0rv:
This pull request updates the list of allowed origins in the `PasskeyWebListener` class to support additional sovereign cloud identity endpoints. This change improves compatibility with new regional identity services.

Expansion of allowed origins:

* Added support for three new sovereign cloud identity endpoints: `login.sovcloud-identity.fr`, `login.sovcloud-identity.de`, and `login.sovcloud-identity.sg` to the allowed origins list in `PasskeyWebListener.kt`.
[AB#3465377](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3465377)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyOriginRulesManager.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            // Check if it's a sovereign cloud origin (requires FIDO path)
+            if (SOVEREIGN_CLOUD_ORIGINS.contains(origin)) {
+                return hasFidoPathSegment(uri.path)
+            }
+
+            false
+        } catch (throwable: Throwable) {
+            Logger.error(TAG, "Error validating origin for URL: $url.", throwable)
```
somalaya: Is it safe to log full URL?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyOriginRulesManager.kt
```diff
@@ -0,0 +1,148 @@
...
+     * @param url The URL to validate
+     * @return `true` if the URL is from an allowed origin and meets all validation rules, `false` otherwise
+     */
+    @JvmStatic
+    fun isAllowedOrigin(url: String): Boolean {
+        return try {
+            val uri = URI(url)
+
+            // Validate scheme
+            if (uri.scheme != HTTPS_SCHEME) {
```
somalaya: Can be changed to cover any corner cases
if (uri.scheme?.lowercase() != HTTPS_SCHEME) {
    return false
}

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyOriginRulesManager.kt
```diff
@@ -0,0 +1,148 @@
...
+ * - Requires exact domain matching
+ * - Case-insensitive FIDO path matching for sovereign clouds
+ */
+object PasskeyOriginRulesManager {
+    private const val TAG = "PasskeyOriginRulesManager"
+    private const val HTTPS_SCHEME = "https"
+    private const val FIDO_SEGMENT = "fido"
+
+    // Production origins - allow any path
+    private val PRODUCTION_ORIGINS = setOf(
```
melissaahn, line 50: nit: if pre-prod origins are getting removed (btw, is there a particular reason why pre-prod support is getting removed?), maybe the name should be something like public_cloud_origins or similar

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1232,17 +1229,10 @@ private Span createSpanWithAttributesFromParent(@NonNull final String spanName)
 
     /**
      * Add a JavaScript to be executed in onPageStarted.
-     * If allowedUrls is null, the script will be executed for all URLs.
-     * If allowedUrls is non-null, the script will be executed only for URLs that start with any of the allowed origins.
+     * The script will be executed only for URLs that are allowed by {@link PasskeyOriginRulesManager}.
      * @param script JavaScript code to be executed.
-     * @param allowedUrls Set of allowed URL origins.
      */
-    public void addOnPageStartedScript(
```
melissaahn, line 1240: q; is there a particular reason why we moved away from an array of scripts (which I assumed was implemented for future js bridge scripts to be added for diff features) to just having separate variables for each script?

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/PasskeyOriginRulesManager.kt
p3dr0rv: add them again

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
p3dr0rv: I’ve reconsidered and don’t think it makes sense to generalize this, as it doesn’t have a clear use case and will likely be removed once the new passkey library rolls out.

somalaya (approved):

somalaya (approved):

PR #2847, opened 2025-12-26
Add validation for number match parameter values, Fixes AB#3446255

fadidurah:
Add validation for number match parameter values.

number match value is a 2 digit numerical value.
session ID is either a 8 character alphanumeric value or a GUID.

[AB#3446255](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3446255)


common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +55,51 @@ class NumberMatchHelper {
...
+
+            if (numberMatch.isNullOrEmpty()) {
+                Logger.warn(TAG, "Number match is null or empty.")
+                return false
+            }
+
+            // Regex for 2-digit number string
+            val twoDigitRegex = Regex("^\\d{2}$")
+            if (!twoDigitRegex.matches(numberMatch)) {
+                Logger.warn(TAG, "Number match is not a valid 2-digit numerical string. Value: $numberMatch")
```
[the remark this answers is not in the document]
fadidurah: Debugging format is important here. number Match value is not a credential, and is shown visually on the screen during the broker WebView session. Session id is also not a credential value, and is simply used to match up with a particular mfa call that is only relevant for the current authentication session. no need to redact these values from logs.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +55,51 @@ class NumberMatchHelper {
...
+                return false
+            }
+
+            if (numberMatch.isNullOrEmpty()) {
+                Logger.warn(TAG, "Number match is null or empty.")
+                return false
+            }
+
+            // Regex for 2-digit number string
+            val twoDigitRegex = Regex("^\\d{2}$")
```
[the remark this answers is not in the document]
fadidurah: Done, moved to fields in the companion object, now they are only initialized once

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +55,51 @@ class NumberMatchHelper {
...
         }
 
+        private fun checkSessionIdAndNumberMatchAreValid(sessionId: String?, numberMatch: String?) : Boolean {
+            if (sessionId.isNullOrEmpty()) {
+                Logger.warn(TAG, "Session ID is null or empty.")
+                return false
+            }
+
+            // Regex for GUID: 8-4-4-4-12 hex digits, case-insensitive
+            val guidRegex = Regex("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$")
```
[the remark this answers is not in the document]
fadidurah: We do not need that strict of validation of the GUIDs, just that the format cannot be something malicious. As long as the same value is sent by server side to the authenticator app via MFA Notification we will be able to complete the flow.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +55,51 @@ class NumberMatchHelper {
...
 
-            // If both parameters are non-null, add a new entry to the hashmap
-            if (sessionId != null && numberMatch != null) {
-                numberMatchMap[sessionId] = numberMatch
+            // Validate that the sessionId and numberMatch are in expected formats
+            val parametersValid = checkSessionIdAndNumberMatchAreValid(sessionId, numberMatch)
+
+            // If both parameters are proper format, add a new entry to the hashmap
+            if (parametersValid) {
+                numberMatchMap[sessionId!!] = numberMatch!!
```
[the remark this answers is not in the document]
fadidurah: I do not think this is a likely or even possible scenario. This method is only called by the active webview instance. Broker does not allow multiple webview instances to be active  at the same time.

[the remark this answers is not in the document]
fadidurah: Will use ConcurrentHashMap anyway

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +55,51 @@ class NumberMatchHelper {
...
+                return false
+            }
+
+            // Regex for GUID: 8-4-4-4-12 hex digits, case-insensitive
+            val guidRegex = Regex("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$")
+            // Regex for 8-character alphanumeric string
+            val alphaNum8Regex = Regex("^[A-Za-z0-9]{8}$")
+
+            if (!guidRegex.matches(sessionId) && !alphaNum8Regex.matches(sessionId)) {
+                Logger.warn(TAG, "Session ID is not a valid GUID or 8-character alphanumeric string. Value: $sessionId")
```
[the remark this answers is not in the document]
fadidurah: Same as an earlier comment. there is no need for secrecy with these values.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -42,7 +43,16 @@ class NumberMatchHelper {
...
+        val numberMatchMap: ConcurrentHashMap<String, String> = ConcurrentHashMap()
+
+        // Regex for GUID: 8-4-4-4-12 hex digits, case-insensitive
+        val guidRegex = Regex("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$")
+
+        // Regex for 8-character alphanumeric string
+        val alphaNum8Regex = Regex("^[A-Za-z0-9]{8}$")
+
+        // Regex for 2-digit number string
+        val twoDigitRegex = Regex("^\\d{2}$")
```
siddhijain, line 55: I think a valid number can be single or double digits.

siddhijain:
Was this PR raised to take care of the S360 item? I think the S360 item is complaining about Javascript API being unauthenticated but this PR is fixing the validation part of the data being passed to the API.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +65,48 @@ class NumberMatchHelper {
...
+            val parametersValid = checkSessionIdAndNumberMatchAreValid(sessionId, numberMatch)
+
+            // If both parameters are proper format, add a new entry to the hashmap
+            if (parametersValid) {
+                numberMatchMap[sessionId!!] = numberMatch!!
                 span.setAttribute(AttributeName.stored_number_match_entry.name, true)
             }
-            // If either parameter is null, do nothing
+            // If either parameter is misformatted, do nothing
             else {
```
richardtz12, line 77: I initially questioned why we got rid of the logging statements here but see that they've been moved into the checkSessionIdAndNumberMatchAreValid() method.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
```diff
@@ -55,20 +65,48 @@ class NumberMatchHelper {
...
+         * Validate that sessionId is non-null and either a GUID or an 8-character alphanumeric string,
+         * and that numberMatch is a non-null 2-digit numerical string.
+         */
+        private fun checkSessionIdAndNumberMatchAreValid(sessionId: String?, numberMatch: String?) : Boolean {
+            if (sessionId.isNullOrEmpty()) {
+                Logger.warn(TAG, "Session ID is null or empty.")
+                return false
+            }
+
+            if (!guidRegex.matches(sessionId) && !alphaNum8Regex.matches(sessionId)) {
```
richardtz12, line 92: Perhaps for my own learning, but can sessionId be of two different formats? both guidRegex and alphaNum8Regex? Why is this the case?

richardtz12 (approved):

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
fadidurah: Depending on the auth flow, the session id value is different, for MFA and passkey, one gets a guid and the other a simple 8 character string

fadidurah:
> Was this PR raised to take care of the S360 item? I think the S360 item is complaining about Javascript API being unauthenticated but this PR is fixing the validation part of the data being passed to the API.

Yes, this is for the s360 item. The work item's name is a little off i think. From the work item:

"The service team has confirmed that no authentication or validation is applied to the sessionId or numberMatch data received via the Javascript API, except for null checks. This fails to meet the requirement to authenticate all sign-in requests and notification payloads received by the Broker through the Javascript API.

This workitem is linked to the TMP review: bookidsecrev95789 and can be accessed here: https://threatmodelingportal.security.azure/reviewdetail/2d8dd855-3651-4628-8246-32e22a09ff38/8d0d308e-cd5c-44a3-9518-43eeeb424b57."

There is not way for us to implement "Authentication" for these java script calls on android without server sending us a token to validate as a parameter, but i don't think that is tenable, the only thing we can do is only exposing the Api when it's a valid url being loaded, and now with this PR also double checking that we only save proper values when we get numberMatch payloads. I spoke to richard about this during the threat model review, that's the conclusion we reached.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
fadidurah: I'm not sure, Say the number match is "05", would they send 05 or 5 as the string value? Let's check with arnav

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
fadidurah: Yup, the validation and logging are in the helper method.

common/src/main/java/com/microsoft/identity/common/internal/numberMatch/NumberMatchHelper.kt
siddhijain: okay Arnav confirmed that this will always be a 2 digit number so we are good here https://msazure.visualstudio.com/One/_git/ESTS-Main?path=/src/Product/Microsoft.AzureAD.ESTS/WebFE.Controllers/RemoteAuthHelper.cs&version=GBmaster&_a=contents

siddhijain:
The PR looks good. Lets test the same and cross device number matching flows before merging this in.

siddhijain (approved):

richardtz12 (approved):

PR #2865, opened 2026-01-16
Add flight-controlled thread pool expansion for silent requests, Fixes AB#3492151

Prvnkmr337:
Fixes: [AB#3492151](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3492151)

Problem:
The Broker's silent token request thread pool (sSilentExecutor) is fixed at 5 threads. Under high load scenarios with multiple concurrent ATS requests from various client apps (Teams, Outlook, OneAuth), this can cause request queuing, increased latency, and potential timeouts.

Solution:
Introduce a flight-controlled mechanism to expand the thread pool from 5 to 8 threads, applying ONLY to the Broker module (not MSAL client apps).

Changes:
- CommonFlight.java: Add ENABLE_EXPANDED_BROKER_SILENT_THREAD_POOL flight (default: false)
- CommandDispatcher.java: Add initializeSilentExecutorForBroker() and resetSilentRequestExecutorWithSize() methods with graceful executor shutdown handling
- AndroidBrokerFlightsManager.kt: Call initializeBrokerSilentExecutor() during Broker initialization after flights are loaded

Key design decisions:
- Uses Android process isolation: Broker runs in separate process, so static sSilentExecutor is independent from MSAL client apps
- Broker-only initialization: AndroidBrokerFlightsManager is Broker-specific code that MSAL apps never call
- Flight-controlled: Safe rollout via ECS with instant rollback capability
- Graceful shutdown: Existing executor is properly terminated before creating expanded pool


changelog.txt
```diff
@@ -2,6 +2,7 @@ vNext
 ----------
 - [MINOR] Skip account aggregation when responding to AcquireTokenSilent api call for OneAuth (#2863)
 - [MINOR] Introduce locks at NameValueStorageFileManagerSimpleCacheImpl layer (#2842)
+- [MINOR] Increase CommandDispatcher thread pool size to 8
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,6 +889,61 @@ public static void stopSilentRequestExecutor() {
...
+        }
+    }
+
+    /**
+     * Resets the SilentRequestsExecutor with a custom thread pool size.
+     * <p>
+     * This gracefully shuts down the existing executor before creating a new one.
+     * </p>
+     *
+     * @param poolSize the desired thread pool size
```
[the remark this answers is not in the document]
Prvnkmr337: Updated to make to make it complete description.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,6 +889,61 @@ public static void stopSilentRequestExecutor() {
...
+                .getFlightsProvider()
+                .getBooleanValue(CommonFlight.ENABLE_EXPANDED_BROKER_SILENT_THREAD_POOL);
+
+        if (useExpandedPool) {
+            Logger.info(methodTag, "Flight enabled: Expanding silent thread pool size for Broker");
+            resetSilentRequestExecutorWithSize(SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED);
+        } else {
+            Logger.info(methodTag, "Flight disabled: Using default silent thread pool size for Broker");
+        }
+    }
```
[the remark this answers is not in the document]
Prvnkmr337: Included telemetry only inside initializeSilentExecutorForBroker; resetSilentRequestExecutorWithSize not required.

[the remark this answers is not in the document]
Prvnkmr337: Added necessary tests.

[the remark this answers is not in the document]
Prvnkmr337: File AndroidBrokerFlightsManager  is part of Broker module in a different repo and not part of this Common module repo. Separate PR will be created for Broker Module. Hence this is not a valid concern

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,6 +889,61 @@ public static void stopSilentRequestExecutor() {
...
+            }
+        } catch (final InterruptedException e) {
+            Logger.warn(methodTag, "Interrupted while waiting for executor shutdown");
+            sSilentExecutor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
+
+        sSilentExecutor = Executors.newFixedThreadPool(poolSize);
+        Logger.info(methodTag, "Silent Executor reset complete with " + poolSize + " threads");
+    }
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed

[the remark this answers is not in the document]
Prvnkmr337: Fixed

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,6 +889,61 @@ public static void stopSilentRequestExecutor() {
...
+                Logger.warn(methodTag, "Executor did not terminate gracefully, forcing shutdown");
+                sSilentExecutor.shutdownNow();
+            }
+        } catch (final InterruptedException e) {
+            Logger.warn(methodTag, "Interrupted while waiting for executor shutdown");
+            sSilentExecutor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
+
+        sSilentExecutor = Executors.newFixedThreadPool(poolSize);
```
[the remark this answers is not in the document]
Prvnkmr337: accessed only during init. not an issue

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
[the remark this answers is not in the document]
Prvnkmr337: fixed

somalaya:
Just curious, how did we arrive at the number for threads allowed for silent requests as 8? Is it because we did some stress testing and confirmed that it is performing better? Or are we going to try it out in PROD directly?

Prvnkmr337:
> Just curious, how did we arrive at the number for threads allowed for silent requests as 8? Is it because we did some stress testing and confirmed that it is performing better? Or are we going to try it out in PROD directly?

The count is based on experiment being done on client side(WXP) thread pool size.  For context the flow is:
WXP  > OneAuth > IPC > Broker
 
So, 

1. WXP experimenting with 8 pool size
2. OneAuth communicates with Broker via IPC
3. Broker we want to match similar count, else 8 on WXP alone won't be beneficial

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,15 +892,97 @@ public static void stopSilentRequestExecutor() {
...
+     *
+     * @param poolSize the desired thread pool size. Must be positive non-zero integer.
+     */
+    private static void resetSilentRequestExecutorWithSize(final int poolSize) {
+        final String methodTag = TAG + ":resetSilentRequestExecutorWithSize";
+        Logger.info(methodTag, "Resetting silent Executor with pool size: " + poolSize);
+
+        synchronized (sExecutorLock) {
+            // Gracefully shutdown existing executor
+            sSilentExecutor.shutdown();
```
somalaya: Could you confirm if below scenario is possible? 

Thread A calls resetSilentRequestExecutorWithSize() and enters synchronized (sExecutorLock) and runs:  sSilentExecutor.shutdown()

Thread B calls submitAndForgetReturningFuture() method and enters synchronized (mapAccessLock) and runs: sSilentExecutor.execute(...)

Would Thread B's execution not fail?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,15 +892,97 @@ public static void stopSilentRequestExecutor() {
...
+     * </p>
+     * <p>
+     * This enables Broker to handle more concurrent silent token requests (8 vs 5 threads)
+     * without affecting MSAL client apps which run in separate processes.
+     * </p>
+     */
+    public static void initializeSilentExecutorForBroker() {
+        final String methodTag = TAG + ":initializeSilentExecutorForBroker";
+
+        final boolean useExpandedPool = CommonFlightsManager.INSTANCE
```
siddhijain: should the flight check be here on in the broker? Can broker only call this method if the flight is enabled?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: +1, yes this is valid case.

To avoid race condition, removed **sExecutorLock** and reused **mapAccessLock** to ensure single lock is used. this avoids race condition.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: +1, Move the flight checking logic to Broker and update related tests accordingly.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,15 +891,89 @@ public static void stopSilentRequestExecutor() {
             sSilentExecutor.shutdownNow();
         }
     }
+    /**
+     * Initializes the silent executor with expanded thread pool size (8 threads).
+     * <p>
+     * This method should ONLY be called by Broker during its initialization phase
+     * when the flight check determines that expanded pool is enabled.
+     * MSAL client apps should NOT call this method - they will use the default pool size.
```
somalaya, line 1096: Since you moved the flight check to broker in this iteration, if you want to avoid MSAL clients to avoid calling this method (in spite of the very clear Javadoc), you can use throwIfNotInvokedByBroker method.

changelog.txt
```diff
@@ -2,6 +2,8 @@ vNext
 ----------
 - [MINOR] Skip account aggregation when responding to AcquireTokenSilent api call for OneAuth (#2863)
 - [MINOR] Introduce locks at NameValueStorageFileManagerSimpleCacheImpl layer (#2842)
+- [MINOR] Add flight-controlled option (ENABLE_EXPANDED_BROKER_SILENT_THREAD_POOL) to expand Broker
+CommandDispatcher silent thread pool to 8 threads (default remains 5; MSAL clients unaffected). Fixes AB#3492151 (#2865)
```
siddhijain: nit - Remove the work item number from the change log. We can also remove the actual flight name here.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,15 +891,89 @@ public static void stopSilentRequestExecutor() {
...
+        synchronized (mapAccessLock) {
+            // Gracefully shutdown existing executor
+            sSilentExecutor.shutdown();
+            boolean terminated = false;
+            try {
+                // Wait for graceful completion (500ms timeout)
+                terminated = sSilentExecutor.awaitTermination(EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+                if (!terminated) {
+                    Logger.warn(methodTag, "Executor did not terminate gracefully within " + EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS + "ms, forcing shutdown");
```
siddhijain: what if there are requests in progress and we forcefully shutdown the executor?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -888,15 +891,89 @@ public static void stopSilentRequestExecutor() {
...
+                    terminated = sSilentExecutor.awaitTermination(EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                }
+            } catch (final InterruptedException e) {
+                Logger.warn(methodTag, "Interrupted while waiting for executor shutdown");
+                sSilentExecutor.shutdownNow();
+                Thread.currentThread().interrupt();
+            }
+
+            if (!terminated) {
+                Logger.error(methodTag, "Executor did not fully terminate. Creating new executor anyway.", null);
```
siddhijain: Also what happens here if there are 2 Executors now? 1 with the thread pool size of 5 and another one with a size of 8?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Request in progress will be interrupted; exception will be caught in submitSilentReturningFuture and set on respective future via set Exception.

NOTE:
This method should ONLY be called:
1. During Broker initialization (no requests active yet)
2. After global signout in Shared Device Mode (no active requests expected)

In both cases, there should be no in-progress requests. If there are (edge case), they will fail with an exception, which is the expected behavior during a global signout scenario where all user state is being cleared anyway.

Also, behavior was similar before the PR; it simply replaced the executor reference without any shutdown. This meant:
1. Old executor kept running with orphaned threads
2. In-progress tasks continued executing on the old executor
3. Potential thread/resource leak

With changes in PR we now properly shutdown the executor:
1. Attempt graceful shutdown (500ms for tasks to complete)
2. Force shutdown if needed (with logging of dropped tasks)
3. Synchronized to prevent race conditions

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: There's only one static reference **sSilentExecutor** so we can't have two executors being actively used simultaneously. When we create a new executor at line 965, the old reference is replaced.

However, if we reach line 960-962 (executor didn't fully terminate), the old executor's threads may still be running in the background temporarily until their tasks complete or they're garbage collected. This is an edge case that:


1. Should rarely occur - we wait 500ms for graceful shutdown, then call **shutdownNow()** which interrupts threads, then wait another 1000ms
2. Has minimal impact - old tasks complete on old threads, new tasks use new executor
3. Self-resolves - old executor is garbage collected once threads complete

This scenario is also no worse than before this PR, where **resetSilentRequestExecutor()** simply created a new executor without any shutdown, potentially leaving the old executor running indefinitely.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: <capturing offline discussion here>

1. Based on my code analysis, throwIfNotInvokedByBroker is part of BrokerHostingAppPublicApi
2. My PR change initializeSilentExecutorWithExpandedPool() is in common module

Common module cannot referrer Broker API. 

I understand the problem identified here that any non-Broker module ex: MSAL can call this API and change the thread pool size. I would like to take others thought, if we need more protection here or java doc is sufficient for now.

vis: @siddhijain

siddhijain: I think it is good to have that protection in place - Check the calling package and throw exception if anyone other than the Broker calls this method. If the javadoc is not read (which is generally the case), we won't even know if the MSAL clients start using thread pool size of 8.

Prvnkmr337: In latest revision added a check for calling package name for Broker apps else throw exception.

siddhijain (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
somalaya: Do we have telemetry to monitor these exceptions?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+     *
+     * @param callingPackageName the package name of the calling application for validation
+     * @throws ClientException if called from a non-Broker application
+     */
+    public static void initializeSilentExecutorWithExpandedPool(@NonNull final String callingPackageName) throws ClientException {
+        final String methodTag = TAG + ":initializeSilentExecutorWithExpandedPool";
+        Logger.info(methodTag, "callingPackageName = " + callingPackageName);
+
+        // Validate caller is a Broker application
+        if (!isBrokerPackageName(callingPackageName)) {
```
rpdome, line 1110: Use ProcessUtil.isBrokerProcess()

That way, you can test with Debug Broker apps as well.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+                sSilentExecutor.shutdownNow();
+                Thread.currentThread().interrupt();
+            }
+
+            if (!terminated) {
+                Logger.error(methodTag, "Executor did not fully terminate. Creating new executor anyway.", null);
+
+                // Emit telemetry for incomplete termination
+                SpanExtension.current().setAttribute(
+                        AttributeName.silent_executor_incomplete_termination.name(),
```
rpdome: Do we want to track how the termination actually happens? 
e.g.
- Terminated successfully 
- Force shutdown succeeded.
- Force shutdown failed.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+                        SpanExtension.current().setAttribute(
+                                AttributeName.silent_executor_forced_shutdown_dropped_tasks.name(),
+                                droppedTasks.size()
+                        );
+                    }
+
+                    // Ensure complete shutdown after forced termination
+                    terminated = sSilentExecutor.awaitTermination(EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                }
+            } catch (final InterruptedException e) {
```
rpdome: Is it possible for this to get another kind of (runtime) exception?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+     * or after global signout in Shared Device Mode when no silent requests are active.
+     * </p>
+     *
+     * @param poolSize the desired thread pool size. Must be positive non-zero integer.
+     */
+    private static void resetSilentRequestExecutorWithSize(final int poolSize) {
+        final String methodTag = TAG + ":resetSilentRequestExecutorWithSize";
+        Logger.info(methodTag, "Resetting silent Executor with pool size: " + poolSize);
+
+        synchronized (mapAccessLock) {
```
rpdome, line 1151: Seems like this mapAccessLock is for protecting sExecutingCommandMap, but i'm not seeing sExecutingCommandMap being invoked here.

(the silentExecutor itself is not protected because it's currently not expected to be changed?)

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+        // Validate caller is a Broker application
+        if (!isBrokerPackageName(callingPackageName)) {
+            Logger.error(methodTag, "Method called from non-Broker application: " + callingPackageName, null);
+            throw new ClientException(
+                    ErrorStrings.BROKER_ONLY_OPERATION,
+                    "This operation is only available for Broker applications."
+            );
+        }
+
+        SpanExtension.current().setAttribute(
```
rpdome: Have you checked whether this is actually emitted? 
(I'm not sure whether there is actually a "current" span when initializeBrokerGlobalStates is invoked during process boot.)

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -521,6 +521,21 @@ public static final class AAD {
...
 
+        /**
+         * Azure Authenticator app package name.
+         */
+        public static final String AZURE_AUTHENTICATOR_APP_PACKAGE_NAME = "com.azure.authenticator";
+
+        /**
+         * Microsoft Authenticator app package name.
+         */
+        public static final String MICROSOFT_AUTHENTICATOR_APP_PACKAGE_NAME = "com.microsoft.azureauthenticator";
```
shahzaibj: Sorry, which app is this?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
         }
     }
+
+    /**
+     * Checks if the given package name belongs to a Broker application.
+     *
+     * @param packageName the package name to check
+     * @return true if the package name is a Broker app, false otherwise
+     */
+    private static boolean isBrokerPackageName(@Nullable final String packageName) {
```
shahzaibj, line 1082: Could move this to BrokerData and leverage some of the existing code there

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +866,131 @@ public static void stopSilentRequestExecutor() {
...
+        if (StringUtil.isNullOrEmpty(packageName)) {
+            return false;
+        }
+        return AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME.equals(packageName) ||
+                AuthenticationConstants.Broker.MICROSOFT_AUTHENTICATOR_APP_PACKAGE_NAME.equals(packageName) ||
+                AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME.equals(packageName);
```
shahzaibj: What about LTW?

and not sure why we have 2 authenticators. We should leverage BrokerData or ProcessUtil as Dome suggested

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: @rpdome 
_ProcessUtil_ is in Android specific **Common** module.
However, _CommandDispatcher_ is in pure java **common4j** module. 

And, _isBrokerProcess_ required **application context** which is not available in CommandDispatcher. So, I think current form is good for now. we can include debug process package name if needed. 

Let me know your thoughts.

rpdome: In that case, let's add a helper function in BrokerData, utilizing getKnownBrokerApps

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Yes, it's possible to get runtime exceptions:

1. SecurityException - from shutdown() or shutdownNow() if a security manager is present and denies permission
2. NullPointerException - from SpanExtension.current().setAttribute() if the span context is null
3. IllegalArgumentException - from Executors.newFixedThreadPool() if poolSize is invalid (though this is controlled internally)

However, the risk is low .

If we want to be defensive, we can wrap with a catch (RuntimeException e) block to ensure the executor is always reset, even if something unexpected happens.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: @rpdome 
BrokerData is too in common module. CommandDispatcher (in common4j) cannot invoke BrokerData in common module.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
Prvnkmr337: Sorry, this is incorrect, will remove it.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: 1. The purpose of adding this check in  CommandDispatcher is to ensure that initializeSilentExecutorWithExpandedPool() can be invoked only by the Broker app. This API is called directly from AndroidBrokerPlatformComponentsFactory, which resides within the Broker.
2. However, since CommandDispatcher lives in the common4j module, it does not have any awareness of Broker-specific context.
3. Additionally, all Broker-related package validation logic exists in the Common module and is not accessible from CommandDispatcher (common4j).

Also, moving the broker package validation to Common module will not be helpful since the _CommandDispatcher.initializeSilentExecutorWithExpandedPool()_ will still be public static accessible from anywhere. So, best to keep the validation inside CommandDispatcher.

Hence, we may have to maintains a list of package names in common4j and validate it. Kindly share your thoughts.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Here mapAccessLock  is used toi protect _silentExecutor_ ; where it is reset with new thread pool

mapAccessLock is used to maintain consistency in:
1. scheduling new request to silentExecutor inside **submitSilentReturningFuture**
2. resetting silentExecutor with new threadpool executor inside **resetSilentRequestExecutorWithSize**

.. thereby avoiding any racing condition

This was made in response to comment: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2865/#discussion_r2714616054 from Sowmya

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Yes, span is available. Tested with Broker  Host app.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -156,6 +156,20 @@ public enum AttributeName {
...
+
+    /**
+     * Number of tasks dropped when silent executor was forcefully shutdown during reset.
+     */
+    silent_executor_forced_shutdown_dropped_tasks,
+
+    /**
+     * Indicates the silent executor did not fully terminate before creating new executor.
+     */
+    silent_executor_incomplete_termination,
```
somalaya: Please ensure that these are also added to the broker's AttributeName.java as well. The attributes don't actually get emitted without adding to that class.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Added logic to capture termination outcome details

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Responded to Dome's feedback, please refer here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2865/#discussion_r2734419983

Kindly share your thoughts

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: 1. Removed incorrect Authenticator name and included right LTW package name.
2. Regarding using BrokerData or ProcessUtil responded with details here: Responded to Dome's feedback, please refer here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2865/#discussion_r2734419983

Kindly share your thoughts

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +878,155 @@ public static void stopSilentRequestExecutor() {
...
+        }
+
+        SpanExtension.current().setAttribute(
+                AttributeName.silent_executor_pool_size.name(),
+                SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED
+        );
+
+        Logger.info(methodTag, "Expanding silent thread pool size for Broker to " + SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED);
+        resetSilentRequestExecutorWithSize(SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED);
+    }
```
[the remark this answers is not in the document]
Prvnkmr337: Flight check is added at the Broker module where initializeSilentExecutorWithExpandedPool is invoked

changelog.txt
```diff
@@ -4,6 +4,8 @@ vNext
 - [MINOR] Cleanup code related to emitting ests telemetry (#2868)
 - [MINOR] Skip account aggregation when responding to AcquireTokenSilent api call for OneAuth (#2863)
 - [MINOR] Introduce locks at NameValueStorageFileManagerSimpleCacheImpl layer (#2842)
+- [MINOR] Add flight-controlled option to expand Broker CommandDispatcher silent thread pool
+ to 8 threads (default remains 5; MSAL clients unaffected) (#2865)
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +878,155 @@ public static void stopSilentRequestExecutor() {
...
+     * This gracefully shuts down the existing executor before creating a new one.
+     * </p>
+     * <p>
+     * <strong>IMPORTANT:</strong> This method should ONLY be called during Broker initialization
+     * or after global signout in Shared Device Mode when no silent requests are active.
+     * </p>
+     *
+     * @param poolSize the desired thread pool size. Must be positive non-zero integer.
+     */
+    private static void resetSilentRequestExecutorWithSize(final int poolSize) {
```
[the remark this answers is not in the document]
Prvnkmr337: resetSilentRequestExecutorWithSize is a private method and always invoked within CommandDispatcher with valid constant. Hence, argument validation is not mandatory.

Instead, added a check for protection and sets to default value in case of 0 or negative value.

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -410,6 +412,114 @@ public void testResetSilentRequestExecutor() throws Exception {
...
+    }
+
+    /**
+     * Tests that initializeSilentExecutorWithExpandedPool() works with all valid Broker packages.
+     */
+    @Test
+    public void testInitializeSilentExecutorWithExpandedPool_AllBrokerPackages() throws Exception {
+        final String[] brokerPackages = {
+                AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME,
+                AuthenticationConstants.Broker.MICROSOFT_AUTHENTICATOR_APP_PACKAGE_NAME,
```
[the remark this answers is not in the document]
Prvnkmr337: Removed and updated with LTW package name

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -862,15 +878,155 @@ public static void stopSilentRequestExecutor() {
...
+                        // Emit telemetry for dropped tasks due to forced shutdown
+                        SpanExtension.current().setAttribute(
+                                AttributeName.silent_executor_forced_shutdown_dropped_tasks.name(),
+                                droppedTasks.size()
+                        );
+                    }
+
+                    // Ensure complete shutdown after forced termination
+                    terminated = sSilentExecutor.awaitTermination(EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                }
```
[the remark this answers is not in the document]
Prvnkmr337: Updated with granular lock to avoid deadlock scenario,

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
shahzaibj: Maybe let's move `BrokerData` to common4j and leverage that then?

Prvnkmr337: @shahzaibj  Discussed the same with Dome but moving BrokerData  will outside the scope of  this task. So, planning to stick with current method of using against package names; as it is needed only for validation purpose.

common4j/src/main/com/microsoft/identity/common/java/broker/BrokerPerformanceMetrics.kt
```diff
@@ -39,7 +40,14 @@ class BrokerPerformanceMetrics(
...
      */
-    val brokerResponseGenerationTimestamp: Long
+    val brokerResponseGenerationTimestamp: Long,
+
+    /**
+     * The configured pool size of the silent executor at the time of the request.
+     * Helps diagnose concurrency-related performance issues.
+     * Defaults to the standard silent executor pool size.
+     */
+    val silentExecutorPoolSize: Int = CommandDispatcher.getDefaultSilentExecutorPoolSize()
```
shahzaibj: Is this correct? Why invoking default here?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -98,7 +99,10 @@ public class CommandDispatcher {
 
     private static final String TAG = CommandDispatcher.class.getSimpleName();
     private static final int SILENT_REQUEST_THREAD_POOL_SIZE = 5;
+    private static final int SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED = 8;
```
rpdome, line 102: btw - you could actually turn this into a flight with default value = 8. 

it will allow you to fudge with the size without shipping the code

(not blocking - something you could consider)

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -513,6 +548,11 @@ public static FinalizableResultFuture<CommandResult> submitSilentReturningFuture
                     AttributeName.silent_requests_queue_size.name(),
                     queueSize
             );
+            // Report the actual pool size being used
+            SpanExtension.current().setAttribute(
+                    AttributeName.silent_executor_pool_size.name(),
+                    ((ThreadPoolExecutor)sSilentExecutor).getCorePoolSize()
```
rpdome: nit: use getSilentExecutorPoolSize()  for simplicity?

common4j/src/main/com/microsoft/identity/common/java/broker/BrokerPerformanceMetrics.kt
Prvnkmr337: Ya, this need not be optional param, I can remove the " CommandDispatcher.getDefaultSilentExecutorPoolSize()". And force caller to always pass value.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
rpdome: as @shahzaibj mentioned in the chat. let's clear sExecutingCommandMap too.

common4j/src/main/com/microsoft/identity/common/java/broker/BrokerPerformanceMetrics.kt
rpdome: i think he meant to say "use getSilentExecutorPoolSize() instead"?

rpdome (approved):

PR #2871, opened 2026-01-22
Fixes AB#3496484 Add utility and attributes that calculate elapsed_time using nanos

siddhijain:
Fixes [AB#3496484](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3496484) From the telemetry, we have seen cases where total elapsed_time is less than the time taken by a certain operation. We believe it is due to the way elapsed time is calculated using System.milliseconds() which is dependent on external factors like System clock change etc.
This PBI is to start using nano to calculate starttime and endtime of an operation and then, convert it to milliseconds before emitting telemetry.
We're testing this with some fields first so that we can compare and tell whether the perf gain in the upcoming release is due to currentTimeMillis() vs nanoTime(), or an actual fix.

### DIFF
--- a/common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
@@ -317,6 +317,7 @@ public class BrokerOAuth2TokenCache
                                                          final @NonNull ICacheRecord cacheRecord) {
         final String methodName = ":loadAggregatedAccountData";
         final long loadStartTime = System.currentTimeMillis();
+        final long loadStartTimeInNanos = System.nanoTime();
 
         final String clientId = cacheRecord.getAccessToken().getClientId();
         final String target = cacheRecord.getAccessToken().getTarget();
@@ -347,6 +348,8 @@ public class BrokerOAuth2TokenCache
         );
         OTelUtility.recordElapsedTime(AttributeName.elapsed_time_load_aggregated_account_data.name(),
                 loadStartTime);
+        OTelUtility.recordElapsedTimeFromNanos(AttributeName.elapsed_time_load_aggregated_account_data_using_nanos.name(),
+                loadStartTimeInNanos);
         return cacheRecordList;
     }
 
--- a/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
@@ -203,6 +203,11 @@ public enum AttributeName {
     elapsed_time_cache_get_account_by_local_account_id,
 
     /**
+     * The time (in milliseconds), calculated using nanos, spent in executing the getAccountByLocalAccountId method in OAuth2TokenCache.
+     */
+    elapsed_time_cache_get_account_by_local_account_id_using_nanos,
+
+    /**
      * The time (in milliseconds) spent in executing the getAccountWithAggregatedAccountDataByLocalAccountId method in OAuth2TokenCache.
      */
     elapsed_time_cache_get_account_with_aggregated_account_data_by_local_account_id,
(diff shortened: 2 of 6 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.kt
```diff
@@ -115,4 +116,17 @@ object OTelUtility {
...
+     *
+     * @param attributeName The name of the attribute to record in the telemetry.
+     * @param startTimeNanos The start time in nanoseconds.
+     *                       The elapsed time is calculated using System.nanoTime() and converted to milliseconds.
+     */
+    @JvmStatic
+    fun recordElapsedTimeFromNanos(attributeName: String, startTimeNanos: Long) {
+        val elapsedTimeMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTimeNanos)
+        SpanExtension.current().setAttribute(attributeName, elapsedTimeMillis)
+    }
```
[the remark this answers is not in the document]
siddhijain: This utility method is not doing anything except for using TimeUnit.NANOSECONDS api so skipping test is fine here.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -202,6 +202,11 @@ public enum AttributeName {
      */
     elapsed_time_cache_get_account_by_local_account_id,
 
+    /**
+     * The time (in milliseconds), calculated using nanos, spent in executing the getAccountByLocalAccountId method in OAuth2TokenCache.
```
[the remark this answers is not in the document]
siddhijain: Yes - this is a part of the Broker repo PR.

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
```diff
@@ -317,6 +317,7 @@ private List<ICacheRecord> loadAggregatedAccountData(final @NonNull AbstractAuth
                                                          final @NonNull ICacheRecord cacheRecord) {
         final String methodName = ":loadAggregatedAccountData";
         final long loadStartTime = System.currentTimeMillis();
+        final long loadStartTimeInNanos = System.nanoTime();
```
wzhipan, line 320: Seems the operations being timed are always short? If it is long enough that the CPU might enter deep sleep, SystemClock.elapsedRealtimeNanos() is the safer choice, as nanoTime() will stop counting.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.kt
```diff
@@ -115,4 +116,17 @@ object OTelUtility {
         val elapsedTimeMillis = endTimeMillis - startTimeMillis
         SpanExtension.current().setAttribute(attributeName, elapsedTimeMillis)
     }
+
+    /**
+     * Records the elapsed time in milliseconds since the provided start time in nanoseconds as a span attribute.
+     *
+     * @param attributeName The name of the attribute to record in the telemetry.
+     * @param startTimeNanos The start time in nanoseconds.
+     *                       The elapsed time is calculated using System.nanoTime() and converted to milliseconds.
```
wzhipan, line 125: If you only need ms accuracy, use SystemClock.elapsedRealtime() instead

rpdome (approved):
Let's update the description that you're testing this with some fields first (so that you can tell whether the perf gain in the upcoming release is due to currentTimeMillis() vs nanoTime(), or an actual fix.)

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/BrokerOAuth2TokenCache.java
siddhijain: System.nanoTime() is fine for our use case because:

1. ATS is a synchronous, active operation - The CPU cannot enter deep sleep while our thread is actively executing code (making network calls, parsing responses, etc.)
2. Deep sleep only affects idle periods - If we're measuring startTime to endTime within a single method execution path, the device won't sleep in between
3. The anomalies we saw were from currentTimeMillis() - Those were due to NTP adjustments and non-monotonic behavior, not sleep issues

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/OTelUtility.kt
siddhijain: same as above.

wzhipan (approved):

PR #2873, opened 2026-01-27
Remove Microsoft Duo dual-screen support

joshfriend:
This is a longstanding issue people experience when trying to integrate the library for the first time: https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1027

It seems VERY reasonable to me that the display-mask dependency is removed from this library because the phone it supported was cancelled years ago after selling functionally zero units. 

Once this is merged and released, the main MSAL library can have all the weird custom proguard rules, ai instructions, and help doc pertaining to this library removed once and for all.


FalconeGiuseppe (approved):
LGTM, DualScreenActivity is not more used since Surface Duo line was cancelled

PR #2881, opened 2026-02-05
Remove io.opentelemetry rule from consumer-rules.pro, Fixes AB#3481336

mohitc1:
Fixes [AB#3481336](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3481336)

Currently, consumer proguard rules add keep rule for inter opentelemetry library. This is blanket and not good practice. It was also flagged by Teams client.

In my evaluation, I see that we can completely remove the keep rule. With that only required classes are retained.

Tested app builds and core scenarios manually.

Automation will be validated in upcoming weekly run.


rpdome (approved):

shahzaibj (approved):
Let's make sure to test this prior to shipping

PR #2897, opened 2026-02-18
Add flight thread pool size based on CPU count for Silent Broker Token request, Fixes AB#3520568

Prvnkmr337:
[AB#3520568](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3520568)
1.  Implements [AB#3520568](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3520568)
2. Updated default thread pool size to 12. Reason, WXP client uses 12 thread pool count and also android Binder IPC thread pool size is 16. Therefore, for better performance default thread count increased to 12.
3. Also, added logic to enable dynamic thread pool size based on device processor count with ThreadpoolExecutor supported core thread pool size and maximum pool size. This protected behind flight controlled from broker.

### DIFF
--- a/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
+++ b/common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
@@ -112,7 +112,7 @@ public class CommandDispatcherTest {
     private static final long TASK_STABILIZATION_DELAY_MS = 100;
 
     /** Thread pool size for silent requests (matches CommandDispatcher.SILENT_REQUEST_THREAD_POOL_SIZE) */
-    private static final int SILENT_THREAD_POOL_SIZE = 5;
+    private static final int SILENT_THREAD_POOL_SIZE = 12;
 
     /** Number of concurrent requests for state collision test */
     private static final int CONCURRENT_REQUEST_COUNT = 20;
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -106,8 +106,10 @@ public class CommandDispatcher {
     private static final int DCF_REQUEST_THREAD_POOL_SIZE = 5;
     private static final int EXECUTOR_GRACEFUL_TERMINATION_TIMEOUT_MS = 500;
     private static final int EXECUTOR_FORCED_TERMINATION_TIMEOUT_MS = 1000;
+    // Cache the pool size for the session
+    private static final int SILENT_REQUEST_THREAD_POOL_SIZE = computeSilentRequestThreadPoolSize();
     private static ExecutorService sInteractiveExecutor = Executors.newSingleThreadExecutor();
-    private static ExecutorService sSilentExecutor = Executors.newFixedThreadPool(getSilentRequestThreadPoolSize());
+    private static ExecutorService sSilentExecutor = Executors.newFixedThreadPool(SILENT_REQUEST_THREAD_POOL_SIZE);
     private static final ExecutorService sDCFExecutor = Executors.newFixedThreadPool(DCF_REQUEST_THREAD_POOL_SIZE);
     private static final Object sLock = new Object();
     private static InteractiveTokenCommand sCommand = null;
@@ -162,15 +164,25 @@ public class CommandDispatcher {
     private static final String TIMEOUT_MSG_UNKNOWN_STATE = 
         "Unknown state '%s'.";
 
-    // Get the actual default pool size from flight configuration
-    private static int getSilentRequestThreadPoolSize() {
+    // Compute the thread pool size based on flight configuration
+    private static int computeSilentRequestThreadPoolSize() {
         if (CommonFlightsManager.INSTANCE.getFlightsProvider() != null) {
             boolean useIncreasedPoolSize = CommonFlightsManager.INSTANCE.getFlightsProvider()
                     .getBooleanValue(CommonFlight.USE_INCREASED_DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE);
             return useIncreasedPoolSize ? DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE
                     : LEGACY_SILENT_REQUEST_THREAD_POOL_SIZE;
         }
-        return DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE;
+        return LEGACY_SILENT_REQUEST_THREAD_POOL_SIZE;
+    }
+
+    /**
+     * Returns the cached thread pool size for silent requests.
+     * This value is computed once during class initialization based on flight configuration.
+     *
+     * @return the cached pool size for the session
+     */
+    private static int getSilentRequestThreadPoolSize() {
+        return SILENT_REQUEST_THREAD_POOL_SIZE;
     }
 
     /**
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -580,13 +580,6 @@ public class CommandDispatcher {
                     getSilentExecutorPoolSize()
             );
 
-            boolean isIncreasedDefaultSilentRequestPoolSize = CommonFlightsManager.INSTANCE.getFlightsProvider()
-                    .getBooleanValue(CommonFlight.USE_INCREASED_DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE);
-            SpanExtension.current().setAttribute(
-                    AttributeName.flight_increased_default_silent_request_pool_size.name(),
-                    isIncreasedDefaultSilentRequestPoolSize
-            );
-
             commandExecutor.execute(OtelContextExtension.wrap(new Runnable() {
                 @Override
                 public void run() {
--- a/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
@@ -88,6 +88,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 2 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -1150,7 +1152,12 @@ private static void resetSilentRequestExecutorWithSize(final int poolSize) {
         final ExecutorService oldExecutor;
         synchronized (mapAccessLock) {
             oldExecutor = sSilentExecutor;
-            sSilentExecutor = Executors.newFixedThreadPool(effectivePoolSize);
+            sSilentExecutor = new ThreadPoolExecutor(
+                    poolSize, // core pool size
+                    poolSize*2, // max pool size
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -1150,7 +1152,12 @@ private static void resetSilentRequestExecutorWithSize(final int poolSize) {
...
+                    30L, // keep-alive time for idle threads above core pool size
+                    TimeUnit.SECONDS,
+                    new LinkedBlockingQueue<>());
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
```diff
@@ -112,7 +112,7 @@ public class CommandDispatcherTest {
     private static final long TASK_STABILIZATION_DELAY_MS = 100;
 
     /** Thread pool size for silent requests (matches CommandDispatcher.SILENT_REQUEST_THREAD_POOL_SIZE) */
-    private static final int SILENT_THREAD_POOL_SIZE = 5;
+    private static final int SILENT_THREAD_POOL_SIZE = 12;
```
siddhijain, line 115: 12 might be the ideal number but 5 -> 12 is a big jump. What if something goes wrong here? How are we planning to monitor this change? Can this be behind a flight that is true by default?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -1150,7 +1152,12 @@ private static void resetSilentRequestExecutorWithSize(final int poolSize) {
         final ExecutorService oldExecutor;
         synchronized (mapAccessLock) {
             oldExecutor = sSilentExecutor;
-            sSilentExecutor = Executors.newFixedThreadPool(effectivePoolSize);
+            sSilentExecutor = new ThreadPoolExecutor(
```
mohitc1, line 1180: Could you provide explainer on why using this ThreadPoolExecutor() (vs older newFixedThreadPool(..))?

Prvnkmr337: We switched from Executors.newFixedThreadPool() to explicitly constructing ThreadPoolExecutor to control queue capacity and overload behavior.

 A fixed thread pool uses an unbounded queue, which can hide overload and inflate latency under bursty silent-request traffic. With ThreadPoolExecutor, we can bound the queue, allow limited thread growth (core→max), and improve observability (queue depth / active thread count) which aligns with existing perf telemetry needs.

common/src/androidTest/java/com/microsoft/identity/common/CommandDispatcherTest.java
Prvnkmr337: Agree, I can sure add a flight as suggested. Also added a metric to track this flight in CommanDispatcher.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
mohitc1: seems then under load, queued request would fail with some error instead of staying queued and failing with timeout later. Is that correct?

Prvnkmr337: Yes That right, this enables faster fail under heavy load with RejectedExecutionException .

Meaning, during heavy load, with **newFixedThreadPool** where unbounded queue will result in request waiting in queue for longer and then timeout. With **ThreadPoolExecutor**, under heavy load request, beyond max pool size, will be rejected earlier instead of sitting in the queue and timing out later.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -159,6 +162,17 @@ private enum RequestState {
     private static final String TIMEOUT_MSG_UNKNOWN_STATE = 
         "Unknown state '%s'.";
 
+    // Get the actual default pool size from flight configuration
+    private static int getSilentRequestThreadPoolSize() {
```
mohitc1: instead of calling this everytime whenever it is used. Why not just store it in static member once for that session of broker?

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -206,7 +206,12 @@ public enum CommonFlight implements IFlightConfig {
     /**
      * Flight to enable the use of locks in name value storage to prevent concurrent access issues.
      */
-    USE_LOCKS_IN_NAME_VALUE_STORAGE("UseLocksInNameValueStorage", false);
+    USE_LOCKS_IN_NAME_VALUE_STORAGE("UseLocksInNameValueStorage", false),
+    /**
+     * Flight to enable increased thread pool size for silent requests.
+     * When true, uses 12 threads. When false, uses legacy 5 threads.
+     */
+    USE_INCREASED_DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE("UseIncreasedSilentRequestThreadPoolSize", true);
```
mohitc1: This will make it 12 for non-broker flows. To keep it enabled by default, keep it false here
and true BrokerFlight enum

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Fixed

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
Prvnkmr337: Fixed, also updated **CommandDispatcher.computeSilentRequestThreadPoolSize()** to return pool size 5 when _getFlightsProvider()_ is null (edge case).

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -98,11 +99,15 @@
 public class CommandDispatcher {
 
     private static final String TAG = CommandDispatcher.class.getSimpleName();
-    private static final int SILENT_REQUEST_THREAD_POOL_SIZE = 5;
-    private static final int SILENT_REQUEST_THREAD_POOL_SIZE_EXPANDED = 8;
+    private static final int DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE = 12;
+    private static final int LEGACY_SILENT_REQUEST_THREAD_POOL_SIZE = 5;
+    // CPU core count is used as a reference for sizing SilentRequestThreadPool.
+    private static final int CPU_CORE_COUNT = Runtime.getRuntime().availableProcessors();
```
mohitc1, line 105: I hope this does not throw exception. An exception would fail the class to load.

mohitc1 (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Verified official [doc](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#availableProcessors--)and checked with Copilot; it does not throw any checked exception.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -161,6 +161,8 @@ public enum AttributeName {
      */
     silent_executor_pool_size,
 
+    flight_increased_default_silent_request_pool_size,
```
siddhijain: nit - rename to a variable denoting boolean value? is_increased_default_pool_size_used?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -554,6 +580,13 @@ public static FinalizableResultFuture<CommandResult> submitSilentReturningFuture
                     getSilentExecutorPoolSize()
             );
 
+            boolean isIncreasedDefaultSilentRequestPoolSize = CommonFlightsManager.INSTANCE.getFlightsProvider()
+                    .getBooleanValue(CommonFlight.USE_INCREASED_DEFAULT_SILENT_REQUEST_THREAD_POOL_SIZE);
+            SpanExtension.current().setAttribute(
```
siddhijain: can we just emit the silent thread pool size here? The flight value can change in between but the thread pool size is set at the beginning?

Prvnkmr337: silent_executor_pool_size is reported in Line#578 already, Shall I just remove this metric then?

siddhijain: yes I think so.

siddhijain (approved):

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
Prvnkmr337: Removed

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
Prvnkmr337: NA, as its usage has been removed.

mohitc1 (approved):

mohitc1 (approved):

PR #2898, opened 2026-02-19
Edge TB: Allow For Lookup Mode, Fixes AB#3501322

melissaahn:
### Summary
Corresponding Broker PR: https://github.com/identity-authnz-teams/ad-accounts-for-android/pull/73

Lookup mode is an ESTS feature where nativebroker=1 and nativebroker_mode=Lookup are sent as extra query parameters to the authorize endpoint, and ESTS is supposed to return information regarding if an account is signed in (like account id and username). Notably, the response from ESTS for this mode has idtoken, AT, and scope set to "none". But we still need to forward this response back to the caller for Edge TB, and my understanding is that the only information ESTS really needs is the account id and username.

This PR allows for lookup mode by creating a new private constructor and factory method in IDToken which allows us to have an IdToken with a "none" value.  And since idToken is not given to us, but the caller (also ESTS) says that they need the username, I implemented some logic in broker to try to retrieve the username from the cache with the account id. I don't think we should necessarily block on sending back the response if the username can't be found, so that's why the webapp account item username is now nullable (but in all other scenarios, this should not be nullable at all).

[AB#3501322](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3501322)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -174,15 +174,12 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
         final Bundle resultBundle = new Bundle();
 
         final String homeAccountId = authenticationResult.getUniqueId();
-        String clientInfo = authenticationResult.getAccountRecord().getClientInfo();
+        final String clientInfo = WebAppsUtil.homeAccountIdToClientInfo(homeAccountId);
         if (clientInfo == null) {
-            clientInfo = WebAppsUtil.homeAccountIdToClientInfo(homeAccountId);
-            if (clientInfo == null) {
-                throw new ClientException(
-                        ErrorStrings.UNKNOWN_ERROR,
-                        errorMessagePrefix + "clientInfo could not be derived from homeAccountId."
-                );
-            }
+            throw new ClientException(
+                    ErrorStrings.UNKNOWN_ERROR,
+                    errorMessagePrefix + "clientInfo could not be derived from homeAccountId."
+            );
         }
         // Some parameters can be null, so double checking.
         final String expiresOn = WebAppsUtil.requireNotNullOrEmpty(authenticationResult.getAccessTokenRecord().getExpiresOn(), WebAppsGetTokenSubOperationResponse.FIELD_EXPIRES_IN);
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -168,24 +168,17 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
                                                            @Nullable final String negotiatedBrokerProtocolVersion,
                                                            @Nullable final String state) throws BaseException {
         final String methodTag = TAG + ":bundleFromAuthenticationResultForWebApps";
-        final String errorMessagePrefix = "Received a successful interactive result, but: ";
         Logger.info(methodTag, "Constructing result bundle from ILocalAuthenticationResult");
 
         final Bundle resultBundle = new Bundle();
 
         final String homeAccountId = authenticationResult.getUniqueId();
-
-        final String clientInfo = WebAppsUtil.homeAccountIdToClientInfo(homeAccountId);
-        if (StringUtil.isNullOrEmpty(clientInfo)) {
-            throw new ClientException(
-                    ErrorStrings.UNKNOWN_ERROR,
-                    errorMessagePrefix + "clientInfo could not be derived from homeAccountId."
-            );
-        }
+        final String clientInfo = authenticationResult.getAccountRecord().getClientInfo();
         // Some parameters can be null, so double checking.
         final String expiresOn = WebAppsUtil.requireNotNullOrEmpty(authenticationResult.getAccessTokenRecord().getExpiresOn(), WebAppsGetTokenSubOperationResponse.FIELD_EXPIRES_IN);
         final String idToken = WebAppsUtil.requireNotNullOrEmpty(authenticationResult.getIdToken(), WebAppsGetTokenSubOperationResponse.FIELD_ID_TOKEN);
-        // When ESTS makes a lookup mode request, id token is set to "none". There is no username in this case, but we shouldn't block on that.
+        // When ESTS makes a lookup mode request, id token is set to "none". We have logic to get the username from the account data storage.
+        // However, in the case where we don't find the username in the cache (for whatever reason), I don't think we should block the lookup mode response from being sent back.
         final String username = Objects.equals(authenticationResult.getAccountRecord().getUsername(), SchemaUtil.MISSING_FROM_THE_TOKEN_RESPONSE) ? null : authenticationResult.getAccountRecord().getUsername();
         final WebAppsAccountItem accountItem = new WebAppsAccountItem(username, homeAccountId, null);
 
--- a/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
@@ -168,12 +168,22 @@ public class MsalBrokerResultAdapter implements IBrokerResultAdapter {
                                                            @Nullable final String negotiatedBrokerProtocolVersion,
                                                            @Nullable final String state) throws BaseException {
         final String methodTag = TAG + ":bundleFromAuthenticationResultForWebApps";
+        final String errorMessagePrefix = "Received a successful interactive result, but: ";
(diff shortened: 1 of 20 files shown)


common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
```diff
@@ -172,18 +174,22 @@ public Bundle bundleFromAuthenticationResultForWebApps(@NonNull final ILocalAuth
...
+        String clientInfo = authenticationResult.getAccountRecord().getClientInfo();
         if (StringUtil.isNullOrEmpty(clientInfo)) {
-            throw new ClientException(
-                    ErrorStrings.UNKNOWN_ERROR,
-                    errorMessagePrefix + "clientInfo could not be derived from homeAccountId."
-            );
+            clientInfo = WebAppsUtil.homeAccountIdToClientInfo(homeAccountId);
+            if (StringUtil.isNullOrEmpty(clientInfo)) {
+                throw new ClientException(
+                        ErrorStrings.UNKNOWN_ERROR,
```
p3dr0rv, line 182: is ok that we throw if this value is null?

p3dr0rv (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/internal/result/MsalBrokerResultAdapter.java
melissaahn: clientInfo is a required field, so if we've tried twice to get it and it's still null, I do think we should throw an exception and return an error back to the caller here.
Username is technically required too, but because ESTS does actually have the capability to retrieve username on their end with the account id (they just haven't made that change yet on their end), that's kinda why I decided not to block on that.

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -301,8 +303,10 @@ public AccountRecord createAccountRecord(
                     );
             accountRecord = new AccountRecord(azureActiveDirectoryAccount);
         } else {
+            final String idTokenValue = microsoftStsTokenResponse.getIdToken();
```
mohitc1, line 304: I think we should update createIdTokenRecord as well, to accept none. SchemaUtil.getCredentialTypeFromVersion(
tokenResponse.getIdToken() expects valid idtoken string. It likely runs into illegalstateexception which is silently ignored. Instead if it expects "none" and handles correctly, we won't be registering that as error log.

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -301,8 +303,10 @@ public AccountRecord createAccountRecord(
...
             final MicrosoftStsAccount microsoftStsAccount = new MicrosoftStsAccount(
-                    new IDToken(microsoftStsTokenResponse.getIdToken()),
+                    Objects.equals(idTokenValue, AuthenticationConstants.Broker.LOOKUP_MODE_EXPECTED_TOKEN_VALUE) ?
```
mohitc1: If possible, comprehensive isLookupMode() check like you have in controller? So, just "none" value of the field does not get accepted in (erroneously) other flows.

melissaahn: I am making a change right now to instead put isLookupMode() in tokencommandparameters instead. Then it can be called from here as well.

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
melissaahn: See 8a12e36faaa718b45ca20fbbe96cf1ffcfd27fc0

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
melissaahn: 8a12e36faaa718b45ca20fbbe96cf1ffcfd27fc0

common/src/test/java/com/microsoft/identity/common/internal/request/MsalBrokerResultAdapterTests.kt
```diff
@@ -468,4 +471,191 @@ class MsalBrokerResultAdapterTests {
...
+        assertTrue(resultString.contains(mockState))
+        assertTrue(resultString.contains(WebAppsGetTokenSubOperationResponse.FIELD_ACCOUNT))
+        assertTrue(resultString.contains(WebAppsAccountItem.FIELD_HOME_ACCOUNT_ID))
+
+        // The username field in the account object should be null
+        // We verify that the special MISSING_FROM_THE_TOKEN_RESPONSE value is NOT in the result
+        Assert.assertFalse(
+            "Username should be null/absent, not the MISSING_FROM_THE_TOKEN_RESPONSE value",
+            resultString.contains(SchemaUtil.MISSING_FROM_THE_TOKEN_RESPONSE)
+        )
```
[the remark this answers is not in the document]
melissaahn: This is fine

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -301,8 +303,10 @@ public AccountRecord createAccountRecord(
                     );
             accountRecord = new AccountRecord(azureActiveDirectoryAccount);
         } else {
+            final String idTokenValue = microsoftStsTokenResponse.getIdToken();
             final MicrosoftStsAccount microsoftStsAccount = new MicrosoftStsAccount(
-                    new IDToken(microsoftStsTokenResponse.getIdToken()),
+                    parameters.isLookupMode() ?
+                            IDToken.createForLookup(idTokenValue) : new IDToken(idTokenValue),
```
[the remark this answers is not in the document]
melissaahn: It's a good check; we don't currently cache these responses if in lookup mode (we have checks for that)

common4j/src/main/com/microsoft/identity/common/java/cache/MicrosoftStsAccountCredentialAdapter.java
```diff
@@ -429,8 +433,9 @@ public IdTokenRecord createIdTokenRecord(
...
         idToken.setHomeAccountId(accountRecord.getHomeAccountId());
         idToken.setRealm(accountRecord.getRealm());
-        idToken.setCredentialType(SchemaUtil.getCredentialTypeFromVersion(
-                tokenResponse.getIdToken())
+        idToken.setCredentialType(parameters.isLookupMode() ?
+                CredentialType.IdToken.name() : SchemaUtil.getCredentialTypeFromVersion(
+                        tokenResponse.getIdToken())
         );
         idToken.setClientId(parameters.getClientId());
         idToken.setSecret(tokenResponse.getIdToken());
```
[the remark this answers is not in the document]
melissaahn: We aren't caching (yet. Might in the future and will need to handle that at that point), so this shouldn't be an issue. The whole point of this PR is to allow for "none" as a token value in the specific lookup mode case.

common4j/src/main/com/microsoft/identity/common/java/providers/oauth2/IDToken.java
```diff
@@ -211,6 +211,38 @@ public IDToken(final String rawIdToken) throws ServiceException {
...
+    /**
+     * Private constructor for lookup mode only.
+     * Creates an IDToken instance with empty token claims.
+     *
+     * @param rawIdToken raw ID token
+     * @param isLookupMode flag indicating this is for lookup purposes only
+     */
+    private IDToken(final String rawIdToken, final boolean isLookupMode) {
+        if (StringUtil.isNullOrEmpty(rawIdToken)) {
+            throw new IllegalArgumentException("null or empty raw idtoken");
```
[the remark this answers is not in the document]
melissaahn: This is fine, because we need a way to create IDToken without going through parseJWT.

common4j/src/main/com/microsoft/identity/common/java/AuthenticationConstants.java
```diff
@@ -608,7 +608,22 @@ public static final class Broker {
...
+        /**
+         * Value which ESTS sends for nativebroker extra query param.
+         */
+        public static final String NATIVEBROKER_VALUE = "1";
+        /**
+         * Key for nativebroker mode sent in the extra query param by ESTS.
+         */
+        public static final String NATIVEBROKER_MODE_KEY = "nativebroker_mode";
+        /**
+         * Value for nativebroker mode sent in the extra query param by ESTS when the request is from lookup.
```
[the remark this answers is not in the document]
melissaahn: Nope, ests is the caller in this case and does send these parameters.

mohitc1 (approved):
:shipit:

PR #2900, opened 2026-02-20
Ignore setShouldTrustDebugBrokers if is not a Debug build, Fixes AB#3515110

p3dr0rv:
This pull request updates the logic for setting the `shouldTrustDebugBrokers` flag in `BrokerData.kt` to ensure it can only be modified in debug builds, and improves logging to clarify when the setting is ignored.

[AB#3515110](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3515110)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
@@ -93,9 +93,10 @@ data class BrokerData(
                 sShouldTrustDebugBrokers = value
                 Logger.info(methodTag, "Setting shouldTrustDebugBrokers to $value." +
                         " This should only be used for testing purposes.")
+            } else {
+                Logger.warn(methodTag, "Attempting to set shouldTrustDebugBrokers to $value. " +
+                        "This change will be ignored since the current build is not a debug build.")
             }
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerData.kt
```diff
@@ -89,10 +89,14 @@ data class BrokerData(
         @JvmStatic
         fun setShouldTrustDebugBrokers(value: Boolean) {
             val methodTag = "$TAG:setShouldTrustDebugBrokers"
-            if (!BuildConfig.DEBUG && value) {
-                Logger.warn(methodTag, "You are forcing to trust debug brokers in non-debug builds.")
+            if (BuildConfig.DEBUG) {
```
somalaya, line 92: Does this work with all of our UI tests? Don't we have test(s) which genuinely test where we validate that the call from unrecognized broker should fail (in local flights = false run)?

somalaya: Nvm, I see you already modified the test TestCase1600567 in a separate PR to not run when both local flights is true and false. Only runs with ecs where BuildConfig.DEBUG check is release. So, my question is not applicable.

somalaya (approved):

rpdome (approved):

PR #2910, opened 2026-02-25
Fixes AB#3527654 Remove Lru cache + few optimizations

siddhijain:
Fixes [AB#3527654](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3527654) With the introduction of in-memory cache, the use of LruCache is redundant and can be removed. This PR also introduces few code optimizations in terms of encryption / decryption.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
@@ -24,8 +24,6 @@ package com.microsoft.identity.common.internal.cache;
 
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.util.LruCache;
-
 import androidx.annotation.GuardedBy;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -54,10 +52,7 @@ import java.util.concurrent.ConcurrentMap;
 public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage {
 
     private static final String TAG = SharedPreferencesFileManager.class.getSimpleName();
-
     private final Object cacheLock = new Object();
-    @GuardedBy("cacheLock")
-    private final LruCache<String, String> fileCache = new LruCache<>(256);
     @GuardedBy("cacheLock")
     private final SharedPreferences mSharedPreferences;
     private final KeyAccessorStringAdapter mEncryptionManager;
@@ -65,7 +60,7 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
     private final String mSharedPreferencesFileName;
     // This is making a huge assumption - that we don't need to separate this cache by context.
     private static final ConcurrentMap<String, SharedPreferencesFileManager> objectCache =
-            new ConcurrentHashMap<String, SharedPreferencesFileManager>(16, 0.75f, 1);
+            new ConcurrentHashMap<>(16, 0.75f, 1);
 
 
     /**
@@ -139,12 +134,6 @@ public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage 
         final String methodTag = TAG + ":putString";
 
         synchronized (cacheLock) {
-            if (value != null) {
-                fileCache.put(key, value);
-            } else {
-                fileCache.remove(key);
-            }
-
             final SharedPreferences.Editor editor = mSharedPreferences.edit();
 
             if (null == mEncryptionManager || StringUtil.isNullOrEmpty(value)) {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -226,9 +212,19 @@ public final Map<String, String> getAll() {
...
+                final String storedValue = entry.getValue();
+                if (StringUtil.isNullOrEmpty(storedValue)) {
+                    continue;
+                }
+
+                final String decryptedValue;
+                try {
+                    decryptedValue = mEncryptionManager.decrypt(storedValue);
+                } catch (final ClientException e) {
+                    Logger.error(TAG + ":getAll", "Failed to decrypt value", null);
```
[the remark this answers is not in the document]
siddhijain: added tests

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -259,7 +255,19 @@ public boolean hasNext() {
...
+                            if (StringUtil.isNullOrEmpty(storedValue)) {
+                                continue;
+                            }
+
+                            final String decryptedValue;
+                            try {
+                                decryptedValue = mEncryptionManager.decrypt(storedValue);
+                            } catch (final ClientException e) {
+                                Logger.error(TAG + ":getAllFilteredByKey", "Failed to decrypt value", null);
+                                continue;
```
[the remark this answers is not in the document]
siddhijain: added tests

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -54,18 +52,15 @@
 public class SharedPreferencesFileManager implements IMultiTypeNameValueStorage {
 
     private static final String TAG = SharedPreferencesFileManager.class.getSimpleName();
-
     private final Object cacheLock = new Object();
```
nickbopp: I see mSharedPreferences is still guarded by this, but it is internally threadsafe. Do you know what operations we're doing that still need to remain atomic/exclusive due to this lock? Can we drop it?

nickbopp: I don't think any of the operations we are doing around the use of sharedpreferences require this lock now that the matching cache structure is gone, but please double check.

siddhijain: Agreed. Confirmed that the cacheLock is no longer needed since each method is independently thread-safe with SharedPreferences and we no longer need this lock for atomicity.

siddhijain: Removed in the latest commit.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -138,67 +129,54 @@ public final void putString(
...
     public final String getString(final String key) {
         final String methodTag = TAG + ":getString";
+        final String storedValue;
 
-        synchronized (cacheLock) {
-            String memCache = fileCache.get(key);
-            if (memCache != null) {
-                return memCache;
-            }
+        storedValue = mSharedPreferences.getString(key, null);
```
wzhipan: Why not assign the value on the same line?

siddhijain: yes good point.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -226,9 +204,19 @@ public final Map<String, String> getAll() {
...
+                if (StringUtil.isNullOrEmpty(storedValue)) {
+                    continue;
+                }
+
+                final String decryptedValue;
+                try {
+                    decryptedValue = mEncryptionManager.decrypt(storedValue);
+                } catch (final ClientException e) {
+                    Logger.error(TAG + ":getAll", "Failed to decrypt value", null);
+                    continue;
```
wzhipan, line 215: Hopefully it wouldn't reach this exception, but if it did, is it reasonable to keep the encrypted value?

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
```diff
@@ -226,9 +204,19 @@ public final Map<String, String> getAll() {
...
+                final String storedValue = entry.getValue();
+                if (StringUtil.isNullOrEmpty(storedValue)) {
+                    continue;
+                }
+
+                final String decryptedValue;
+                try {
+                    decryptedValue = mEncryptionManager.decrypt(storedValue);
+                } catch (final ClientException e) {
+                    Logger.error(TAG + ":getAll", "Failed to decrypt value", null);
```
mohitc1: >TAG + ":getAll" [](http://example.com/codeflow?start=33&length=15)

methodTag

mohitc1 (approved):
:shipit:

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
siddhijain: yeah since we cant deterministically say that the encrypted value is corrupted.

common/src/main/java/com/microsoft/identity/common/internal/cache/SharedPreferencesFileManager.java
siddhijain: updated

wzhipan (approved):

PR #2915, opened 2026-02-26
Remove test for trustDebugBrokerFlag, Fixes AB#3529118

p3dr0rv:
Removed test for trustDebugBrokerFlag in BuildConfigTest.

why?
Now we have scheduled and manual builds with trustDebugBrokerFlag for testing

[AB#3529118](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3529118)

### DIFF
--- a/common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
+++ b/common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
@@ -39,12 +39,4 @@ class BuildConfigTest {
 
         assert(!BuildConfig.bypassRedirectUriCheck)
     }
-
-    @Test
-    fun failIfTrustDebugBrokerFlagEnabled(){
-        // Do not run this on scheduled test.
-        Assume.assumeFalse(isBuildReason(BuildReason.Schedule))
-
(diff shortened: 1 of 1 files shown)


wzhipan (approved):

common/src/test/java/com/microsoft/identity/common/BuildConfigTest.kt
```diff
@@ -39,12 +39,4 @@ class BuildConfigTest {
 
         assert(!BuildConfig.bypassRedirectUriCheck)
     }
-
-    @Test
-    fun failIfTrustDebugBrokerFlagEnabled(){
```
rpdome, line 44: should we still keep this for the monthly ecs run?
(as a sanity check, so that we don't accidentally ship with this flag enabled)

p3dr0rv: the monthly pipeline has its own validation mechanism.
https://identitydivision.visualstudio.com/Engineering/_git/AuthClientAndroidPipelines?path=/templates/verification/flight-verification-for-official-release.yml

rpdome (approved):

PR #3010, opened 2026-03-10
Handle target="_blank" links in authorization WebView, Fixes AB#3535641

somalaya:
**Problem**
Total loss recovery is a feature provided by VID to recover the user's account. This gets initiated from inside of the webview. See below screenshot for reference. Once the user clicks on "Recover your account", the flow goes through a set of redirects - sometimes to 3P identity verifiers where they may handle button clicks differently. They may contain links with target="_blank" (ideally used in websites where they want to open a new tab upon a button or link click)(e.g., Terms of Use, privacy links), the Android WebView silently drops these navigations. Users tap the link and nothing happens — no browser opens, no feedback is given.

**Solution**
Override onCreateWindow to intercept target="_blank" navigations and open them in the device's default external browser via intent. 
Note : This will ONLY be enabled for TLR links. The check is made in OnPageLoadedCallback method of WebViewAuthorizationFragment. 
- **How are we gating it only for TLR URLs?** : When a page is loaded, we know its url, we check if it matches the TLR url pattern.. If it does, we enable multiple windows support in webview by calling the method setSupportMultipleWindows(true). Else, it is the same as today (set to false).
- **Would onCreateWindow method be called even when setSupportMultipleWindows is false?** : No, it completely needs setSupportMultipleWindows to be true. So, the code in onCreateWindow need not be behind a flight as the code triggering it is already behind the flight.

**Feature gating**
The change is gated behind ENABLE_WEBVIEW_MULTIPLE_WINDOWS
CommonFlight: default false (safe for MSAL/non-broker callers)
BrokerFlight: overrides to true (enabled in broker context)

**Telemetry**
The change also involves adding a new span with default sampling rate. But we expect this span to be emitted in very low numbers (10 to 20 per day) once Total loss recovery feature is enabled. Once this feature is stabilized, we can remove this span.

 Fixes [AB#3535641](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3535641)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -46,8 +46,10 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.webkit.PermissionRequest;
 import android.webkit.WebChromeClient;
+import android.webkit.WebResourceRequest;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
+import android.webkit.WebViewClient;
 import android.widget.ProgressBar;
 
 import androidx.activity.result.ActivityResultLauncher;
@@ -87,8 +89,13 @@ import java.util.Map;
 
 import static com.microsoft.identity.common.java.AuthenticationConstants.OAuth2.UTID;
 
+import com.microsoft.identity.common.java.opentelemetry.OTelUtility;
+import com.microsoft.identity.common.java.opentelemetry.SpanName;
 
+import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Scope;
 
 /**
  * Authorization fragment with embedded webview.
@@ -352,6 +359,13 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
         mWebView.getSettings().setUseWideViewPort(true);
         mWebView.getSettings().setBuiltInZoomControls(webViewZoomControlsEnabled);
         mWebView.getSettings().setSupportZoom(webViewZoomEnabled);
+
+        // Enable multiple windows so that target="_blank" links trigger onCreateWindow
+        // instead of being silently dropped by the WebView - controlled by flight.
+        final boolean multipleWindowsEnabled = CommonFlightsManager.INSTANCE.getFlightsProvider()
+                .isFlightEnabled(CommonFlight.ENABLE_WEBVIEW_MULTIPLE_WINDOWS);
+        mWebView.getSettings().setSupportMultipleWindows(multipleWindowsEnabled);
+
         mWebView.setVisibility(View.INVISIBLE);
         mWebView.setWebViewClient(webViewClient);
         mWebView.setWebChromeClient(new WebChromeClient() {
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -376,6 +392,64 @@ public Bitmap getDefaultVideoPoster() {
...
+                        ? ((AuthorizationActivity) requireActivity()).getSpanContext() : null;
+                final Span span = OTelUtility.createSpanFromParent(
+                        SpanName.WebViewTargetBlankNavigation.name(), parentSpanContext);
+                boolean windowHandled = false;
+                try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
+                    Logger.info(methodTag, "onCreateWindow: intercepting target=_blank navigation.");
+                    final WebView interceptorWebView = new WebView(view.getContext());
+                    interceptorWebView.setWebViewClient(new WebViewClient() {
+                        @Override
+                        public boolean shouldOverrideUrlLoading(WebView v, WebResourceRequest request) {
```
melissaahn, line 418: so iiuc, the url that actually needs to load in the browser is going to be a redirect from the one initially loaded into the interceptor webview? And the interceptor webview gets destroyed right afterwards.
Does our code wait on something in the external browser to finish? Wondering what might get displayed to the user if maybe something isn't loading correctly, whether the link can't be opened properly or something fails in the browser (not sure if the latter is really our problem tho)

melissaahn (commented):
I have one comment and I see some checks are failing, so I'll approve after those pass.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -376,6 +392,64 @@ public Bitmap getDefaultVideoPoster() {
...
+                            try {
+                                final String url = request.getUrl().toString();
+                                final String scheme = request.getUrl().getScheme();
+                                if (!"https".equalsIgnoreCase(scheme) && !"http".equalsIgnoreCase(scheme)) {
+                                    Logger.warn(methodTag, "onCreateWindow: ignoring non-http(s) scheme: " + scheme);
+                                    return true;
+                                }
+                                Logger.info(methodTag, "onCreateWindow: opening target=_blank URL in external browser.");
+                                final Intent browserIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+                                view.getContext().startActivity(browserIntent);
```
p3dr0rv: Is this running in UI thread or we need to launch intent via view.post()?

somalaya: shouldOverrideUrlLoading
 is guaranteed by Android to be called on the UI (main) thread — all callbacks run on the main thread. So both startActivity() and loadUrl() are safe here without needing view.post()

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
somalaya: Let me actually add some context on the scenario.
1. Total loss recovery (TLR)  is a way to recover an account from inside of a sign in page. User clicks on "Recover your account" link in sign in page and is redirected to this flow.
2. Once recovery starts in browser(this part of pushing out to browser is what we are supporting in this PR), user has to do a bunch of things like completing face check, attaching gov id card etc and then they will be redirected to AuthApp for issuing a credential. 
3. At this point, the user is actually in AuthApp and gets a TAP and they have to go to mysignins page using that TAP to finally setup their authentication methods (PSI or passkey or whatever)
4. In the end, the user still has to manually go back to the productivity app from which they initially triggered the recovery. So, we will still have the webview open as it is on the screen where "recover your account" option was shown. --- **This part should answer your question.**
5. User manually has to click on back button to finally go to the productivity app's sign in page and use the recovered auth mechanism.

Now coming to your question on interceptor webview, we are not doing much with, it is only needed so that we get hold of the URL in shouldOverrideUrlLoading method. Once we get that URL, we don't really need this interceptor webview. Our main webview is still up and running.

melissaahn: This helps me understand the scenario. Thank you so much for writing it out.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -376,10 +396,115 @@ public Bitmap getDefaultVideoPoster() {
                 // We will return a 10x10 empty image, instead of the default grey playback image. #2424
                 return Bitmap.createBitmap(10, 10, Bitmap.Config.ARGB_8888);
             }
+
+            @Override
+            public boolean onCreateWindow(final WebView view, boolean isDialog,
+                                          boolean isUserGesture, final Message resultMsg) {
```
melissaahn, line 402: Question: would it provide an extra layer of protection if we were to also make use of isUserGesture? I know that you already have a check in place to make sure that it's a TLR url, but wondering if it wouldn't hurt to also check that this was user initiated or not (I'm thinking for any url we probably don't want to allow popups that weren't initiated by users)

melissaahn (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
somalaya: Sure, works with isUserGesture for this specific scenario and it tightens the scenario further.

p3dr0rv (approved):

melissaahn (approved):

PR #3013, opened 2026-03-12
Handle openid-vc urls in webview, Fixes AB#3535641

somalaya:
**Context** : Total loss recovery is a feature introduced by VID team to recover an account. This feature will be triggered from inside of a webview when the user clicks on "Recover your account" link. After this a 3P identity verifier will verify the user and provide a credential. Once a credential is available, a specific redirect URI is triggered **openid-vc://**? request_uri=https://verifier.example.com/request/123, which will invoke wallet library in Authenticator app --- All this is functionality not connected to broker except the fact that the URLs are all being loaded in our webview.

**What is openid-vc** ? : The openid-vc:// scheme is part of the OpenID4VC family of specifications, which includes:
OpenID4VCI (Verifiable Credential Issuance) — how a wallet receives credentials from an issuer
OpenID4VP (Verifiable Presentations) — how a wallet presents credentials to a verifier

**Problem** : When a web page loaded in the authentication WebView navigates to an openid-vc:// URI (used by OpenID Verifiable Credentials flows), the WebView cannot handle this non-HTTP custom scheme natively. The navigation fails, blocking the VC issuance/presentation flow.

**Solution**  : Intercept **openid-vc://** navigations in AzureActiveDirectoryWebViewClient.handleUrl() and delegate them to the Android system so a registered wallet app can handle the request.

Fixes [AB#3535641](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3535641)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -928,19 +928,29 @@ public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     private void processOpenIdVcRequest(@NonNull final WebView view, @NonNull final String url) {
         final String methodTag = TAG + ":processOpenIdVcRequest";
         view.stopLoading();
-        try {
+        final Span span = createSpanWithAttributesFromParent(SpanName.ProcessOpenIdVcRequest.name());
+        try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
             final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
             if (intent.resolveActivity(getActivity().getPackageManager()) != null) {
                 getActivity().startActivity(intent);
                 Logger.info(methodTag, "Launched external handler for OpenID VC request.");
+                span.setAttribute(AttributeName.is_openid_vc_handler_found.name(), true);
+                span.setStatus(StatusCode.OK);
             } else {
                 Logger.warn(methodTag, "No application found to handle openid-vc:// URI.");
+                span.setAttribute(AttributeName.is_openid_vc_handler_found.name(), false);
+                span.setStatus(StatusCode.ERROR, "No handler found for openid-vc:// URI");
                 returnError(ErrorStrings.ACTIVITY_NOT_FOUND, "No application found to handle the OpenID Verifiable Credentials request.");
             }
         } catch (final ActivityNotFoundException e) {
             Logger.error(methodTag, "Failed to launch handler for openid-vc:// URI.", e);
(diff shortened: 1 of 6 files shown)


rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -901,6 +917,33 @@ private void processAmazonAppUri(@NonNull final String url) {
...
+     */
+    private void processOpenIdVcRequest(@NonNull final WebView view, @NonNull final String url) {
+        final String methodTag = TAG + ":processOpenIdVcRequest";
+        view.stopLoading();
+        try {
+            final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+            if (intent.resolveActivity(getActivity().getPackageManager()) != null) {
+                getActivity().startActivity(intent);
+                Logger.info(methodTag, "Launched external handler for OpenID VC request.");
```
mohitc1, line 937: Do you want keep the webview up in this case? likewise, do you want to terminate entire flow if activity not found?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -901,6 +917,33 @@ private void processAmazonAppUri(@NonNull final String url) {
...
+     * Handles an openid-vc:// URL by stopping the WebView and launching an
+     * {@link Intent#ACTION_VIEW} intent so the system can route it to the
+     * registered wallet application.
+     *
+     * @param view The WebView that intercepted the navigation.
+     * @param url  The original (non-lowercased) openid-vc:// URL.
+     */
+    private void processOpenIdVcRequest(@NonNull final WebView view, @NonNull final String url) {
+        final String methodTag = TAG + ":processOpenIdVcRequest";
+        view.stopLoading();
```
mohitc1, line 930: any telemetry needed?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: This will terminate the entire flow. Without these changes, the flow was anyway getting terminated with an unprotected SSL error.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Telemetry in this flow is not super important for us (broker) but I just added it in latest commit for capturing the worst case scenario.

mohitc1 (approved):

mohitc1 (approved):

PR #3022, opened 2026-03-17
Add WebView file upload support, Fixes AB#3545256

somalaya:
**Summary**
Adds file upload support to the embedded WebView used during interactive authentication flows. When a web page (e.g., eSTS) presents a file chooser (<input type="file">), the WebView now handles it by launching the system file picker.

**Why** : This is needed for verfied id flows where eSTS redirects the user to some identity verification website which may ask the user to upload a profile picture.

**Changes**

- Implemented onShowFileChooser in the WebChromeClient to handle file upload requests

- Registered an ActivityResultLauncher in onCreate to receive file picker results (single file, multi-file, or cancellation)

- Extracted handleFileUploadRequest() as a @VisibleForTesting method for testability

- Added cleanup in onDestroy (cancel pending callback, unregister launcher)

- Entire feature gated behind CommonFlight.ENABLE_WEBVIEW_FILE_UPLOAD (default: off)

-  Added WebViewFileUpload span for telemetry

Fixes [AB#3545256](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3545256)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
@@ -440,8 +440,9 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
                     final WebView webView,
                     final ValueCallback<Uri[]> filePathCallback,
                     final FileChooserParams fileChooserParams) {
-                final SpanContext parentSpanContext = requireActivity() instanceof AuthorizationActivity
-                        ? ((AuthorizationActivity) requireActivity()).getSpanContext() : null;
+                final FragmentActivity host = getActivity();
+                final SpanContext parentSpanContext = host instanceof AuthorizationActivity
+                        ? ((AuthorizationActivity) host).getSpanContext() : null;
                 return handleFileUploadRequest(filePathCallback, fileChooserParams, parentSpanContext);
             }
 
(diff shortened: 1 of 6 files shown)


fadidurah (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -515,6 +575,86 @@ boolean isTlrUrl(@Nullable final String url) {
...
+
+        final Span span = OTelUtility.createSpanFromParent(
+                SpanName.WebViewFileUpload.name(), parentSpanContext);
+
+        // Cancel any existing callback to avoid a dangling reference.
+        if (mFileUploadCallback != null) {
+            mFileUploadCallback.onReceiveValue(null);
+        }
+        // Clear any previous callback reference before handling the new request.
+        mFileUploadCallback = null;
```
wzhipan: Why not put it in the if block above?

wzhipan (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
somalaya: If you are referring to if (mFileUploadCallback != null)  block, then we only need to do this when we actually need it. It's not needed when flight is disabled.

wzhipan (approved):

PR #3027, opened 2026-03-18
[Common] Bleu cloud support, Fixes AB#3536002

mohitc1:
Fixes [AB#3536002](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3536002)

Adds support for new sovereign clouds — Bleu (France), Delos (Germany), and SovSG (Singapore) — which are not yet returned by the global instance discovery endpoint.

**Changes:**

**common4j — `AzureActiveDirectoryCloud.java`**
- Added static `AzureActiveDirectoryCloud` instances for Bleu, Delos, and SovSG with pre-seeded metadata (host, cache host, aliases)
- Added host string constants for all known clouds (`PUBLIC_CLOUD_HOST`, `PPE_CLOUD_HOST`, `CHINA_CLOUD_HOST`, `US_GOV_CLOUD_HOST`, `BLEU_CLOUD_HOST`, `DELOS_CLOUD_HOST`, `SOVSG_CLOUD_HOST`)

**common4j — `AzureActiveDirectory.java`**
- Pre-seed sovereign cloud metadata into the cloud cache at class initialization, so sovereign authorities are recognized as "known" without a network call
- Added `KNOWN_CLOUD_DISCOVERY_HOSTS` — an immutable set of all known cloud hosts used to route discovery requests to each cloud's own endpoint instead of always going through the global endpoint
- Added `isKnownCloudDiscoveryHost()` to check membership
- Updated `ensureCloudDiscoveryForAuthority()` with cache-aware routing: cache hit → no-op; known host → per-cloud discovery; unknown host → global fallback
- Removed deprecated `isInitialized()` method; callers now use `ensureCloudDiscoveryComplete()`

**Bug fixes:**
- **Authority recognition:** `getKnownAuthorityResult()` now wraps discovery in try-catch — if discovery fails, it still checks hardcoded metadata and developer configuration instead of immediately returning "unknown authority"
- **Thread safety:** Iteration over `knownAuthorities` is now synchronized to prevent `ConcurrentModificationException`

**common (Android) — `AzureActiveDirectoryWebViewClientTest.java`**
- Replaced `isInitialized() + performCloudDiscovery()` with `ensureCloudDiscoveryComplete()`

**Tests (Kotlin):**
- `AzureActiveDirectoryTest.kt` — sovereign host detection, pre-seeding verification, cache-aware routing, case insensitivity
- `AzureActiveDirectoryCloudTest.kt` — cloud constant and instance validation
- `AuthorityKnownAuthorityTest.kt` — sovereign authority recognition via pre-seeded metadata
- `AzureActiveDirectoryAuthorityCloudTest.kt` — `isSameCloudAsAuthority` for sovereign clouds


mohitc1:
@copilot Add PR description

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -204,12 +253,23 @@ public static synchronized String getDefaultCloudUrl() {
...
+        public static synchronized void performCloudDiscoveryForCloudUrl(@NonNull final String cloudUrl)
+            throws ClientException {
+        final String methodName = ":performCloudDiscoveryForCloudUrl";
         final URI instanceDiscoveryRequestUri;
         try {
-            instanceDiscoveryRequestUri = new CommonURIBuilder(getDefaultCloudUrl() + AAD_INSTANCE_DISCOVERY_ENDPOINT)
+            instanceDiscoveryRequestUri = new CommonURIBuilder(cloudUrl + AAD_INSTANCE_DISCOVERY_ENDPOINT)
                     .setParameter(API_VERSION, API_VERSION_VALUE)
-                    .setParameter(AUTHORIZATION_ENDPOINT, AUTHORIZATION_ENDPOINT_VALUE)
+                    .setParameter(AUTHORIZATION_ENDPOINT, cloudUrl + "/common/oauth2/v2.0/authorize")
```
shahzaibj: Could this part be defined in a constant?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -242,18 +302,67 @@ public static synchronized void performCloudDiscovery()
...
-            sIsInitialized = true;
         }
     }
 
     /**
-     * Ensures that cloud discovery has been completed. If not, it will perform cloud discovery.
+     * Ensures that cloud discovery has been completed using the default global endpoint.
+     * Delegates to {@link #ensureCloudDiscoveryForAuthority(URL)} with the default cloud URL.
      */
     public static synchronized void ensureCloudDiscoveryComplete() throws ClientException {
```
shahzaibj: The other method below also seems to ensure that it's "complete" but doesn't have the word "complete" in its name which seems to imply they do different things. Even though this one just calls that other method so they certainly do the same thing

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -242,18 +302,67 @@ public static synchronized void performCloudDiscovery()
...
+     *   <li>For known cloud hosts — queries that cloud's own discovery endpoint.</li>
+     *   <li>For unknown hosts — falls back to the default global discovery endpoint.</li>
+     * </ul>
+     *
+     * @param authorityUrl The authority URL whose host determines the discovery endpoint, or null to use the default.
+     */
+    public static synchronized void ensureCloudDiscoveryForAuthority(@Nullable final URL authorityUrl)
+            throws ClientException {
+        if (authorityUrl == null) {
+            ensureCloudDiscoveryComplete();
```
shahzaibj: Kinda seems circular to me

- Someone could calls ensureCloudDiscoveryForAuthority(null URL)
- This calls ensureCloudDiscoveryComplete() - no params
- That then calls ensureCloudDiscoveryForAuthority(URL) with default cloud URL

So we started in this method, branched out to another method only to come back here? Why can't this method just proceed itself by taking the default cloud URL?


Seems like we're adding unnecessary complexity and reducing readability.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -242,18 +302,67 @@ public static synchronized void performCloudDiscovery()
...
+            return;
+        }
+        final String host = authorityUrl.getHost();
+        if (host == null) {
+            return;
+        }
+        final String hostLower = host.toLowerCase(Locale.US);
```
shahzaibj, line 402: Saw this toLower in another place as well. I believe this depends on the data in sAadClouds being lowercased as well to begin with. (not sure if that's already happening, it probably is)

But maybe a better approach is to simply compare ignore case?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -242,18 +302,67 @@ public static synchronized void performCloudDiscovery()
...
+        // Already in cache — no discovery needed.
+        if (sAadClouds.containsKey(hostLower)) {
+            return;
+        }
+        if (isKnownCloudDiscoveryHost(hostLower)) {
+            // Known cloud host not yet in cache — discover from its own endpoint.
```
shahzaibj: This would be one of the new sov clouds?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryCloud.java
```diff
@@ -115,4 +116,35 @@ public boolean isValidated() {
...
+    public static final String BLEU_CLOUD_HOST = "login.sovcloud-identity.fr";
+    public static final String DELOS_CLOUD_HOST = "login.sovcloud-identity.de";
+    public static final String SOVSG_CLOUD_HOST = "login.sovcloud-identity.sg";
+
+    // These are added without network calls because they are not part of instance discovery yet.
+    /** Bleu sovereign cloud (France). */
+    public static final AzureActiveDirectoryCloud BLEU = new AzureActiveDirectoryCloud(
+            BLEU_CLOUD_HOST,
+            BLEU_CLOUD_HOST,
+            Collections.singletonList(BLEU_CLOUD_HOST)
```
shahzaibj, line 134: All same?

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
```diff
@@ -102,48 +100,4 @@ public void validate() throws ArgumentException {
             // This logic should also apply to CIAM authorities
         }
     }
-
-    /**
-     * Note - this method may throw a variety of RuntimeException if we cannot perform cloud
-     * discovery to determine the set of cloud aliases.
-     * @return true if the authority matches the cloud environment that the account is homed in.
-     */
-    private boolean authorityMatchesAccountEnvironment() {
```
shahzaibj, line 111: Are we no longer matching against account environment?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectoryCloud.java
mohitc1: yes, referring to other MSALs, this is the case.

common4j/src/main/com/microsoft/identity/common/java/nativeauth/commands/parameters/AcquireTokenNoFixedScopesCommandParameters.java
mohitc1: This was dead code

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -223,14 +223,16 @@ private static Authority getEquivalentConfiguredAuthority(@NonNull final String
...
             // are the same...
-            for (final Authority currentAuthority : knownAuthorities) {
-                if (!StringUtil.isNullOrEmpty(currentAuthority.mAuthorityUrlString)) {
-                    final URL currentAuthorityUrl = new URL(currentAuthority.mAuthorityUrlString);
-                    final String currentHttpAuthority = currentAuthorityUrl.getAuthority();
-
-                    if (httpAuthority.equalsIgnoreCase(currentHttpAuthority)) {
-                        result = currentAuthority;
-                        break;
+            synchronized (sLock) {
```
rpdome, line 226: do you think it's worth changing this to a read-write lock for perf?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -79,13 +92,55 @@ public class AzureActiveDirectory
...
+    )));
+
+    static {
+        preSeedSovereignClouds();
+    }
+
+    /**
+     * Pre-seeds the cloud cache with sovereign cloud metadata so they are
+     * recognized without a network call. Called at class init and after
+     * environment changes that clear the cache.
```
rpdome, line 129: nit: suggest attaching a link to your design doc/pr here for reference.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
mohitc1: yes, it was based on that toLower on keys when put in the map.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
rpdome: Why can't this method just proceed itself by taking the default cloud URL? << +1

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
mohitc1: one of the known (new ones and all old as well)

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
mohitc1: Fixed.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
mohitc1: I think I will leave as is, usages are short lived, low contention and existing. was just not being used here.

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -243,17 +345,75 @@ public static synchronized void performCloudDiscovery()
...
+            return;
+        }
+        final String hostLower = host.toLowerCase(Locale.US);
+
+        // Already in cache — no discovery needed.
+        if (sAadClouds.containsKey(hostLower)) {
+            return;
+        }
+
+        final boolean sovereignDiscoveryEnabled = CommonFlightsManager.INSTANCE.getFlightsProvider()
```
rpdome, line 409: this won't be a complete revert right? due to refactoring.

What if we create AzureActiveDirectoryNew and AzureActiveDirectoryOld, then made AzureActiveDirectory a facade that delegates calls to one of the two implementations based on the flight? (That seems like the cleanest approach I can come up with.)

mohitc1: won't that be hard to maintain, there is other functionality in the class as well. I will have separate those out.

I am keeping flight just as safety check, it won't even help msal flows. 

intention is to keep old vs new behavior and not old vs new code

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -399,23 +387,25 @@ public static KnownAuthorityResult getKnownAuthorityResult(Authority authority)
         boolean known = false;
 
         try {
-            performCloudDiscovery();
+            AzureActiveDirectory.ensureCloudDiscoveryForAuthority(authority);
+            Logger.info(TAG + methodName, "Cloud discovery complete.");
```
somalaya: nit : methodTag can be used here

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -228,7 +228,15 @@ public enum CommonFlight implements IFlightConfig {
...
-    ENABLE_OPEN_ID_VC_REDIRECT("EnableOpenIdVcRedirect", true);
+    ENABLE_OPEN_ID_VC_REDIRECT("EnableOpenIdVcRedirect", true),
+
+    /**
+     * Flight to enable sovereign cloud instance discovery routing.
+     * When enabled, discovery requests for known sovereign cloud hosts are routed
+     * through host of passed in authority if part of known cloud list.
+     * Turn off to fall back to global-only discovery, previous behavior.
+     */
+    ENABLE_SOVEREIGN_CLOUD_INSTANCE_DISCOVERY("EnableSovereignCloudInstanceDiscovery", true);
```
somalaya: We are ok with this being true always for Brokerless without a kill switch?

common4j/src/main/com/microsoft/identity/common/java/providers/microsoft/azureactivedirectory/AzureActiveDirectory.java
```diff
@@ -79,13 +94,66 @@ public class AzureActiveDirectory
     private static final String API_VERSION = "api-version";
     private static final String API_VERSION_VALUE = "1.1";
     private static final String AUTHORIZATION_ENDPOINT = "authorization_endpoint";
-    private static final String AUTHORIZATION_ENDPOINT_VALUE = "https://login.microsoftonline.com/common/oauth2/v2.0/authorize";
+    private static final String AUTHORIZE_ENDPOINT_SUFFIX = "/common/oauth2/v2.0/authorize";
```
somalaya: nit : Should we just call this AUTHORIZE_ENDPOINT_PATH?

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
mohitc1: yes, talked offline

somalaya (approved):

somalaya (approved):

rohitluthra (approved):

p3dr0rv (approved):

PR #3040, opened 2026-03-20
Fixes AB#3549188 add telemetry in the Broker Discovery flow

siddhijain:
Fixes [AB#3549188](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3549188) add telemetry in the Broker Discovery flow

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
+++ b/common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
@@ -37,6 +37,8 @@ import com.microsoft.identity.common.java.exception.ClientException
 import com.microsoft.identity.common.java.exception.ClientException.ONLY_SUPPORTS_ACCOUNT_MANAGER_ERROR_CODE
 import com.microsoft.identity.common.java.interfaces.IPlatformComponents
 import com.microsoft.identity.common.java.logging.Logger
+import com.microsoft.identity.common.java.opentelemetry.AttributeName
+import com.microsoft.identity.common.java.opentelemetry.SpanExtension
 import kotlinx.coroutines.*
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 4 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -665,4 +665,47 @@ public enum AttributeName {
...
+
+    /**
+     * Comma-separated active broker package names returned by each responding candidate.
+     */
+    broker_discovery_reported_active_brokers,
+
+    /**
+     * The discovery path taken: "cache", "ipc", "account_manager", or "account_manager_fallback".
+     */
+    broker_discovery_path,
```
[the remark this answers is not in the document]
siddhijain: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3040#discussion_r2963890194)

common/src/main/java/com/microsoft/identity/common/internal/activebrokerdiscovery/BrokerDiscoveryClient.kt
```diff
@@ -100,6 +102,11 @@ class BrokerDiscoveryClient(private val brokerCandidates: Set<BrokerData>,
 
         const val ERROR_BUNDLE_KEY = "ERROR_BUNDLE_KEY"
 
+        const val ACCOUNT_MANAGER = "ACCOUNT_MANAGER"
```
rpdome, line 105: brokerdiscoveryclient is on the SDK side (MSAL/OneAuth/WPJ API), we probably don't get telemetry from the former two. is this intentional?

siddhijain: You're right! I will have to move some of these telemetry data points away from the client side. Similar to enabled settings that might get overridden.

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -162,18 +164,60 @@ public int getUIDForPackage(final String packageName) {
...
+        try {
+            isEnabledByEnabledSetting = isEnabledForSetting(
+                    mPackageManager.getApplicationEnabledSetting(packageName),
+                    isEnabledByApplicationInfo
+            );
+        } catch (final IllegalArgumentException e) {
+            Logger.verbose(methodTag, "Unable to read enabled setting for package: " + packageName);
+        }
+
+        SpanExtension.current().setAttribute(AttributeName.is_installed_by_application_info.name(), isInstalledByApplicationInfo);
```
rpdome, line 200: is there another info we could use? (e.g. is the app in hibernation, deep sleep, etc)

PR #3044, opened 2026-03-20
Fixes AB#3549188 add helper method for BrokerDiscovery debugging

siddhijain:
Fixes [AB#3549188](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3549188) add helper method for BrokerDiscovery debugging

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
@@ -41,6 +41,8 @@ import androidx.annotation.NonNull;
 import com.microsoft.identity.common.BuildConfig;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
 import com.microsoft.identity.common.logging.Logger;
 
 import java.io.UnsupportedEncodingException;
@@ -164,8 +166,19 @@ public class PackageHelper {
         final String methodTag = TAG + ":isPackageInstalledAndEnabled";
         boolean enabled = false;
         try {
-            ApplicationInfo applicationInfo = mPackageManager.getApplicationInfo(packageName, 0);
-            if (applicationInfo != null) {
+            final ApplicationInfo applicationInfo = mPackageManager.getApplicationInfo(packageName, 0);
+            if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(
+                    CommonFlight.USE_ENABLED_SETTING_FOR_PACKAGE_CHECK)) {
+                try {
+                    enabled = isEnabledForSetting(
+                            mPackageManager.getApplicationEnabledSetting(packageName),
+                            applicationInfo.enabled
+                    );
+                } catch (final IllegalArgumentException e) {
+                    Logger.verbose(methodTag, "Unable to read enabled setting for package: " + packageName);
+                    enabled = applicationInfo.enabled;
+                }
+            } else {
                 enabled = applicationInfo.enabled;
             }
         } catch (NameNotFoundException e) {
--- a/common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
@@ -177,6 +177,60 @@ public class PackageHelper {
     }
 
     /**
+     * Returns a summary of the install/enabled state for a given package,
+     * checking via multiple PackageManager APIs.
+     *
+     * @param packageName the package name to look up.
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -176,6 +176,60 @@ public boolean isPackageInstalledAndEnabled(final String packageName) {
...
+            isEnabledBySetting = isEnabledForSetting(
+                    mPackageManager.getApplicationEnabledSetting(packageName),
+                    isEnabledByAppInfo
+            );
+        } catch (final IllegalArgumentException ignored) { }
+
+        return "appInfo=" + isInstalledByAppInfo
+                + ",appEnabled=" + isEnabledByAppInfo
+                + ",pkgInfo=" + isInstalledByPkgInfo
+                + ",enabledSetting=" + isEnabledBySetting;
```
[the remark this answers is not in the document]
siddhijain: This is fine. This is to get a general idea of what api to use to identify a package state.

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -176,6 +176,60 @@ public boolean isPackageInstalledAndEnabled(final String packageName) {
...
+        try {
+            final ApplicationInfo appInfo = mPackageManager.getApplicationInfo(packageName, 0);
+            if (appInfo != null) {
+                isInstalledByAppInfo = true;
+                isEnabledByAppInfo = appInfo.enabled;
+            }
+        } catch (final NameNotFoundException ignored) { }
+
+        try {
+            isInstalledByPkgInfo = getPackageInfo(mPackageManager, packageName) != null;
```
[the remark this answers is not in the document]
siddhijain: This is fine. The verbosity makes the code clearer to understand.

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -176,6 +176,60 @@ public boolean isPackageInstalledAndEnabled(final String packageName) {
...
+        } catch (final NameNotFoundException ignored) { }
+
+        try {
+            isEnabledBySetting = isEnabledForSetting(
```
rpdome: any plan to replace applicationInfo.enabled with these values (in case telemetry tells us it makes sense to do so?)

e.g. wire these methods into isPackageInstalledAndEnabled, and use flight to control it.

That way we can take an action right away after getting the data. no need to wait for another 2 months.

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
```diff
@@ -176,6 +176,60 @@ public boolean isPackageInstalledAndEnabled(final String packageName) {
...
+            );
+        } catch (final IllegalArgumentException ignored) { }
+
+        return "appInfo=" + isInstalledByAppInfo
+                + ",appEnabled=" + isEnabledByAppInfo
+                + ",pkgInfo=" + isInstalledByPkgInfo
+                + ",enabledSetting=" + isEnabledBySetting;
+    }
+
+    private static boolean isEnabledForSetting(final int enabledSetting,
```
rpdome, line 230: let's print out the actual enabledSetting value - so that we know what's actually happening behind the scene.

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
siddhijain: _yes - I was planning to get it one we have the data from the telemetry that it is indeed the issue is. But I can add that now._

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
siddhijain: This will be returned and emitted in the telemetry, right? You mean we should include in the logs as well?

rpdome: i mean both. 

My understanding is that COMPONENT_ENABLED_STATE_DISABLED, COMPONENT_ENABLED_STATE_DISABLED_USER and COMPONENT_ENABLED_STATE_DISABLED_UNTIL_USED are totally different scenarios. and we want to know exactly under what condition that the logic failed. (could be a subset of this, or all of them)

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
siddhijain: Added changes behind a flight

rpdome (approved):

common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java
siddhijain: Discussed offline. Done

melissaahn (approved):

PR #3055, opened 2026-03-26
Add flight-gated HTTP cancellation on ATS command-level timeout to eliminate zombie worker threads AB#3542516, Fixes AB#3542516

Prvnkmr337:
**Related Work item**: [AB#3542516](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3542516)

**Problem**:
When ATS command-level timeout (30s) fires in CommandDispatcher.submitAcquireTokenSilentSync(), the caller thread moves on but the worker thread remains blocked on HttpURLConnection.read() for up to 30+ additional seconds. No cancellation signal is sent — ResultFuture.cancel() is a no-op, and Thread.interrupt() doesn't unblock socket I/O. This creates zombie threads that hold thread pool slots, waste battery/network, and keep eSTS connections open for results nobody consumes. Production telemetry shows ~5.9M timed_out_execution events/day (AAD), with 40,700+ thread-hours/day wasted on zombie HTTP I/O.

**Solution**:
Introduce CancellationSignal — a thread-safe signal (AtomicBoolean + AtomicReference<HttpURLConnection>) shared between the caller and worker threads via ThreadLocal. On command timeout, the caller calls urlConnection.disconnect() through the signal, which throws SocketException on the blocked worker thread, unblocking it immediately (~0s vs ~30s).

**Implementation**:
- CancellationSignal.kt (NEW): Kotlin class with cancel(), registerConnection(), unregisterConnection(), ThreadLocal companion (@JvmStatic), Closeable interface, safeDisconnect() helper, fast-path + double-check in registerConnection().
- FinalizableResultFuture.java: Added CopyOnWriteArrayList<CancellationSignal> with addCancellationSignal() and cancelAllSignals(). List-based approach handles deduplicated requests (shared future) — cancelAllSignals() iterates all signals ensuring the worker's signal (with the registered HttpURLConnection) gets cancelled.
- CommandDispatcher.java:
  * submitAcquireTokenSilentSync(): Flight-gated cancelAllSignals() in TimeoutException catch (ENABLE_HTTP_CANCELLATION_ON_TIMEOUT, default: false).
  * submitSilentReturningFuture(): Signal created before mapAccessLock, addCancellationSignal() in all 3 branches (original + 2 dedup), setForCurrentThread() in worker run(), pre-start isCancelled() check, clearCurrentThread() in finally.
- UrlConnectionHttpClient.java: registerConnection() before I/O, unregisterConnection() in finally, completionCallback.accept(null) for HTTP telemetry in both cancellation paths, throws ClientException(REQUEST_CANCELLED_BY_COMMAND_TIMEOUT).
- ClientException.java: Added REQUEST_CANCELLED_BY_COMMAND_TIMEOUT error code.
- CommonFlight.java: Added ENABLE_HTTP_CANCELLATION_ON_TIMEOUT flight.

_Note: setConnectTimeout(30s) and setReadTimeout(30s) in UrlConnectionHttpClient are intentionally retained. CancellationSignal only covers ATS flows (via CommandDispatcher's ThreadLocal). Non-ATS flows (discovery, device registration, interactive auth, OneAuth) have no CancellationSignal — HTTP timeouts are their sole protection against hung connections. Additionally, connect timeout drives the retry policy (CONNECTION_TIMEOUT triggers isRetryableException), and serves as the safety net when the cancellation flight is disabled._

**Testing:**
- CancellationSignalTest.kt (9 tests): Atomic ops, ThreadLocal lifecycle, cross-thread cancel.
- FinalizableResultFutureCancellationTest.kt (3 tests): Single signal, dedup, CopyOnWriteArrayList safety.
- CommandDispatcherCancellationTest.kt (3 tests): E2E cross-thread — normal timeout, pre-start cancellation, dedup timeout. All 15 tests pass on JVM (no device required).
- Manual testing: Verified fix locally — worker thread cancelled immediately on timeout.


common4j/src/main/com/microsoft/identity/common/java/result/FinalizableResultFuture.java
```diff
@@ -34,6 +39,24 @@
...
+     * Cancelling ALL signals ensures the worker's connection gets disconnected.
+     *
+     * <p>Uses {@link CopyOnWriteArrayList} because:</p>
+     * <ul>
+     *   <li>Writes ({@link #addCancellationSignal}) are rare — typically 1-2 per command lifetime</li>
+     *   <li>Reads ({@link #cancelAll} iteration) happen once on timeout</li>
+     *   <li>No external synchronization needed for iteration — avoids
+     *       {@code ConcurrentModificationException} that {@code Collections.synchronizedList}
+     *       would require manual synchronization to prevent</li>
+     * </ul>
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed  it.

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -451,6 +453,20 @@ public static ILocalAuthenticationResult submitAcquireTokenSilentSync(@NonNull f
                 commandResult = future.get(silentTokenTimeOutMs, TimeUnit.MILLISECONDS);
             }
         } catch (final TimeoutException e) {
+            // Signal the worker thread to disconnect the active HTTP connection.
+            // cancelAll() iterates ALL signals on this future, including the
```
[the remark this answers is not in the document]
Prvnkmr337: Fixed it

common4j/src/main/com/microsoft/identity/common/java/flighting/CommonFlight.java
```diff
@@ -244,7 +244,16 @@ public enum CommonFlight implements IFlightConfig {
...
+    USE_ENABLED_SETTING_FOR_PACKAGE_CHECK("UseEnabledSettingForPackageCheck", false),
+
+    /**
+     * Enables HTTP cancellation on command-level timeout.
+     * When true, CommandDispatcher calls CancellationSignal.cancel() on TimeoutException,
+     * which disconnects the active HttpURLConnection on the worker thread.
+     *
+     * Default: false (disabled for safe rollout).
+     */
+    ENABLE_HTTP_CANCELLATION_ON_TIMEOUT("EnableHttpCancellationOnTimeout", false);
```
siddhijain: How would you monitor the flight rollout and decide whether to further the rollout or halt?

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
```diff
@@ -351,12 +351,50 @@ private static void safeCloseStream(final Closeable stream) {
...
+        // Register the connection so it can be disconnected on command-level timeout.
+        // If already cancelled (caller timed out while we were queued), abort immediately.
+        if (cancellationSignal != null
+                && !cancellationSignal.registerConnection(urlConnection)) {
+            // Emit HTTP end telemetry to match the HttpStartEvent — response is null (no I/O happened)
+            completionCallback.accept(null);
+            throw new ClientException(
+                ClientException.REQUEST_CANCELLED_BY_COMMAND_TIMEOUT,
+                "Request cancelled before HTTP execution due to command-level timeout");
+        }
```
siddhijain: How does having an exception that is separate from TIMEOUT_EXCEPTION help? The caller can’t act differently here, so why not just throw a timeout exception?
Here’s what I was thinking — at the configured ATS timeout (30s), the request times out and we send the timeout exception back to the caller. Internally, we cancel the HTTP requests and any other in‑flight requests as part of the cleanup. This would simply be a change in implementation details.
Why introduce a new exception that will start appearing in OneAuth telemetry when there is essentially no different action item for the caller?

common4j/src/main/com/microsoft/identity/common/java/net/CancellationSignal.kt
```diff
@@ -0,0 +1,152 @@
...
+ * sets this signal to cancelled, which disconnects any registered [HttpURLConnection].
+ *
+ * Lifecycle:
+ * 1. Created by CommandDispatcher BEFORE submitting to the thread pool
+ * 2. Stored on FinalizableResultFuture so the caller can trigger cancellation
+ * 3. Passed to the worker thread via ThreadLocal (set at start of run(), cleared in finally)
+ * 4. UrlConnectionHttpClient registers/unregisters the active HttpURLConnection
+ * 5. On [cancel], the registered connection is disconnected immediately
+ */
+class CancellationSignal : Closeable {
```
mohitc1: >Closeable [](http://example.com/codeflow?start=27&length=9)

why closeable?

mohitc1: >CancellationSignal [](http://example.com/codeflow?start=6&length=18)

This can be generic; not necessarily tied to HttpURLConnection

e.g. instead of registerConnection() we can registerOnCancel(runnable/task)

common4j/src/main/com/microsoft/identity/common/java/result/FinalizableResultFuture.java
```diff
@@ -53,4 +76,37 @@ public boolean isCleanedUp() {
...
+     *
+     * <p>Thread-safe: CopyOnWriteArrayList provides snapshot iteration —
+     * no ConcurrentModificationException even if addCancellationSignal()
+     * is called concurrently from a dedup request.</p>
+     *
+     * <p>Named {@code cancelAllSignals()} instead of {@code cancel()} to avoid
+     * confusion with {@link java.util.concurrent.Future#cancel(boolean)} inherited
+     * from {@link com.microsoft.identity.common.java.util.ResultFuture}.</p>
+     */
+    public void cancelAllSignals() {
```
mohitc1: >cancelAllSignals [](http://example.com/codeflow?start=16&length=16)

can we use exiting cancel() instead?

common4j/src/main/com/microsoft/identity/common/java/result/FinalizableResultFuture.java
```diff
@@ -34,6 +39,24 @@
...
+     * <p>Uses {@link CopyOnWriteArrayList} because:</p>
+     * <ul>
+     *   <li>Writes ({@link #addCancellationSignal}) are rare — typically 1-2 per command lifetime</li>
+     *   <li>Reads ({@link #cancelAllSignals()} iteration) happen once on timeout</li>
+     *   <li>No external synchronization needed for iteration — avoids
+     *       {@code ConcurrentModificationException} that {@code Collections.synchronizedList}
+     *       would require manual synchronization to prevent</li>
+     * </ul>
+     */
+    private final List<CancellationSignal> mCancellationSignals = new CopyOnWriteArrayList<>();
```
mohitc1: >List<CancellationSignal> [](http://example.com/codeflow?start=18&length=24)

why is the list needed here?

Even if there are multiple http connection going on under a Future, the can all be signaled using single CancelationSignal. CancelationSignal itself can have list of connections (or runnables in general) that can be executed on cancel.

common4j/src/main/com/microsoft/identity/common/java/result/FinalizableResultFuture.java
```diff
@@ -34,6 +39,24 @@
 public class FinalizableResultFuture<T> extends ResultFuture<T> {
```
mohitc1, line 37: >FinalizableResultFuture [](http://example.com/codeflow?start=13&length=23)

This can implicitly create a CancellationSignal. getCancelationSignal() method can then be used to get that object and attach to thread local in executor service.

This keeps logic more scoped to future and thread local storage. (Furthermore, if we had method like execute on future itself, updating thread local could have added there, but that'd be bigger change.)

common4j/src/main/com/microsoft/identity/common/java/net/UrlConnectionHttpClient.java
mohitc1: yeah.. plus whatever exception is thrown from here goes to caller layers and can flow all way up in the stack.

I think exception due to cancellation should not propagate outside CommandDispatcher which has already got TimeoutException. Probably, we catch this exception there, log and continue, instead of calling commandresult callbacks.

common4j/src/main/com/microsoft/identity/common/java/result/FinalizableResultFuture.java
Prvnkmr337: Can't use Future.cancel() here — two reasons:

1. Dedup safety: cancel() would set the future to a cancelled state, causing get() to throw CancellationException for all waiters including dedup callers. Per above the exception propagation comment, we intentionally keep the future's state unaffected so dedup callers time out naturally with correct state classification.
2. Semantic mismatch: Future.cancel() means "I don't want this result anymore" — it's a lifecycle operation on the future itself. What we need here is "disconnect the HTTP connection so the worker thread unblocks" — a network-layer cleanup operation. These are different concerns: one affects every consumer of the future, the other is an internal implementation detail between the dispatcher and the HTTP client. cancelAllSignals() keeps them separate.

(Also note: ResultFuture.cancel() currently returns false unconditionally — it's a no-op. Overriding it to add connection disconnection would change the contract for all ResultFuture consumers, not just the cancellation path.)

Let me know your thoughts.

mohitc1: I meant there's already cancel() in future, can we hook this logic in that?

common4j/src/main/com/microsoft/identity/common/java/controllers/CommandDispatcher.java
```diff
@@ -135,6 +136,18 @@ private enum ExecutorTerminationOutcome {
...
     private static ConcurrentMap<BaseCommand, FinalizableResultFuture<CommandResult>> sExecutingCommandMap = new ConcurrentHashMap<>();
 
+    /**
+     * Outcome of cancellation signal processing for a timed-out ATS request.
+     * Emitted as the {@link AttributeName#cancellation_outcome} span attribute
+     * to track cancellation effectiveness during flight rollout.
+     */
+    private enum CancellationOutcome {
+        /** Worker detected cancellation before executeCommand() — no work performed, thread slot freed immediately. */
+        skipped_before_execution,
```
siddhijain, line 146: nit - Since these changes are in common, you will have to include the attributes in Broker repo as well for them to be emitted.

siddhijain (approved):

mohitc1 (approved):
:shipit:

changelog.txt
```diff
@@ -22,6 +22,7 @@ Version 24.1.0
 - [MINOR] Add WebView file upload support (#3022)
 - [MINOR] Enhance WebAuthn telemetry for passkey registration (#3018)
 - [MINOR] Enabled opening of TLR URLs in browser by default by enabling the flight ENABLE_WEBVIEW_MULTIPLE_WINDOWS (#3042)
+- [MINOR] Add CancellationSignal to gracefully cancel in-flight HTTP requests on ATS command-level timeout, preventing zombie worker threads from holding thread pool slots (#3055)
```
mohitc1: Move this to vNext section

mohitc1 (approved):

PR #3069, opened 2026-04-06
Making AI recommended changes for Adding Authenticator to Android complete

rohitluthra:
Added these changes to correctly support Authenticator app in Android complete.

Its strcutrely only adds JVM Static function


shahzaibj (changes requested):
I believe these were already moved to AppRegistry - so why adding them here?

PR #3088, opened 2026-04-21
Add onboarding telemetry recorder, constants, and session correlation store for mobile onboarding flow, Fixes AB#3462876

wzhipan:
## Summary

Adds shared Android infrastructure for tracking mobile onboarding telemetry — the E2E flow where users hit Conditional Access (CA) blocking errors (broker install, MDM enrollment, device registration) and must remediate before completing sign-in.

Consumed by both OneAuth (non-brokered flows) and broker (brokered flows) to construct a JSON telemetry blob emitted through MATS.

Linked Feature: [AB#3462876](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3462876)
Linked PBI: [AB#3568356](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3568356)

### `OnboardingTelemetryConstants.kt` (common4j)
- JSON field key constants for the onboarding blob — all snake_case matching MATS convention (EntityStore prepends `mo_` to produce final MATS columns, e.g. `blocking_errors` → `mo_blocking_errors`)
- Step ID constants (`AuthenticationStarted`, `BrokerInstallPrompted`, `DeviceRegistrationStarted`, etc.)
- Blocking error value constants (`BROKER_INSTALLATION_TRIGGERED`, `MDM_FLOW`) — must match C++ hardcoded strings in OneAuth''s `InteractiveRequest.cpp`

### `OnboardingTelemetryRecorder.kt` (common/Android)
- Records steps, blocking errors, and domain tracking during interactive auth flows
- Constructs the blob from a seed JSON provided by C++ xplat core
- `addStep(stepId)` captures the current ISO-8601 timestamp internally — callers no longer pass a timestamp
- `finalizeBlob()` returns populated JSON only if blocking errors were recorded AND `sessionCorrelationId` is non-empty (empty string otherwise); logs at verbose/warn/error level for diagnostics
- Persists `sessionCorrelationId` to SharedPreferences on block detection for app-kill resilience

### `OnboardingSessionCorrelationStore.kt` (common/Android)
- SharedPreferences-backed persistence for session correlation IDs (per-app sandbox; same schema/file name across apps for consistency)
- Used by OneAuth''s Djinni `SessionCachePersistence` adapter (JNI bridge to C++ `SessionCorrelationIdCache`)

## Design decisions

- **Kotlin**: Implemented in Kotlin, consistent with recent additions to Common (e.g. `BrowserRedirectValidator`, `DeviceRegistrationClientApplication`). Public surface remains JVM-compatible for Java callers.
- **Snake_case blob keys**: All internal JSON keys use snake_case (`session_correlation_id`, `blocking_errors`, etc.) so EntityStore can use simple `"mo_" + key` concatenation for MATS columns — no mapping table needed, follows established MATS convention
- **Recorder, not builder**: The class records events incrementally during WebView navigation with mid-flow side effects (SharedPreferences persistence), so it''s named `OnboardingTelemetryRecorder` rather than following the builder pattern
- **No `remediationNeeded` field**: Redundant — implied by the presence of `blocking_errors`
- **`apply()` over `commit()` for persistence**: Telemetry tolerates rare loss; blocking errors leave the app alive for seconds-to-minutes of user remediation, so the async flush window is far longer than typical loss. Avoids main-thread disk I/O.

## Testing
- 20 Common Android/Robolectric unit tests pass (recorder + store)
- OneAuth C++ unit tests pass (794 tests, 0 failures) including 41 new onboarding-specific tests
- E2E tested with OneAuthTestApp: CA-blocked user triggers `BROKER_INSTALLATION_TRIGGERED`, blob emitted with correct fields and `sessionCorrelationId`

## Review feedback addressed
- Renamed `OnboardingBlobFieldKeys` → `OnboardingTelemetryConstants`
- Renamed `OnboardingSessionCachePersistence` → `OnboardingSessionCorrelationStore`
- `addStep()` now captures timestamp internally (single-arg signature)
- Added Logger calls in `finalizeBlob()` (verbose/error), seed-parse `init` catch (warn), and `persistSessionCorrelation()` catch (warn)
- `finalizeBlob()` now early-returns with a warn log if `sessionCorrelationId` is empty (avoids emitting uncorrelatable telemetry)
- Replaced `java.time.Instant` with `SimpleDateFormat` (desugaring disabled, minSdk 24)
- Extracted `FIELD_ID` constant for the cache entry key
- Clarified `OnboardingSessionCorrelationStore` Javadoc (per-app SharedPreferences sandbox)
- Updated `addBlockingError` Javadoc to reference correct constant values
- Full MIT license headers
- Converted to Kotlin

## Dependencies
- **Upstream of**: OneAuth onboarding telemetry PR (Djinni + C++ core + Android platform layer)
- **No breaking changes**: all new files, no modifications to existing code


common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,258 @@
...
+    public void addStep(@NonNull String stepId, @NonNull Instant timestamp) {
+        mStepsList.add(new StepEntry(stepId, timestamp.toString()));
+    }
+
+    /**
+     * Record a blocking error code detected during the flow.
+     * Also persists the session correlation entry to SharedPreferences immediately
+     * for app-kill resilience.
+     *
+     * @param errorCode The blocking error code (e.g., "65001", "53000")
```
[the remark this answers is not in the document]
wzhipan: updated JavaDoc

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,258 @@
...
+                cache = new JSONObject(existing);
+            } else {
+                cache = new JSONObject();
+            }
+
+            String key = mClientId + "|" + mTarget;
+            JSONObject entry = new JSONObject();
+            entry.put("id", mSessionCorrelationId);
+            entry.put("ts", System.currentTimeMillis());
+            cache.put(key, entry);
```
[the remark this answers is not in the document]
wzhipan: added guard at the beginning of the function

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,258 @@
...
+
+            prefs.edit().putString(PREFS_FILE, cache.toString()).apply();
+        } catch (JSONException e) {
```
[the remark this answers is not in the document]
wzhipan: Updated JavaDoc

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
```
Prvnkmr337: Copyright header should be same context and format as other existing files. Feedback applies to all other new files too.

Prvnkmr337:
Copyright header — Use full 22-line MIT block, not short 2-line version

Prvnkmr337:
New files should be Kotlin — If I remember right all new files need to be in Kotlin.

common4j/src/main/com/microsoft/identity/common/java/telemetry/OnboardingBlobFieldKeys.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.microsoft.identity.common.java.telemetry;
+
+/**
+ * JSON field keys for the onboarding telemetry blob.
+ * All keys use snake_case to match MATS convention — EntityStore prepends "mo_"
+ * to produce the final MATS column name (e.g., "blocking_errors" → "mo_blocking_errors").
+ * Seed creation + aggregation keys come from OnboardingBlobConstants (Djinni-generated).
+ */
+public final class OnboardingBlobFieldKeys {
```
Prvnkmr337: nit: can be renamed to OnboardingTelemetryConstants for better naming

common4j/src/main/com/microsoft/identity/common/java/telemetry/OnboardingBlobFieldKeys.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final String STEP_DEVICE_REGISTRATION_STARTED = "DeviceRegistrationStarted";
+    public static final String STEP_DEVICE_REGISTRATION_COMPLETED = "DeviceRegistrationCompleted";
+    public static final String STEP_FLOW_COMPLETED = "FlowCompleted";
+
+    // Blocking error values — must match C++ hardcoded strings in InteractiveRequest.cpp
+    public static final String BLOCKING_ERROR_BROKER_INSTALL = "BROKER_INSTALLATION_TRIGGERED";
+    public static final String BLOCKING_ERROR_MDM_FLOW = "MDM_FLOW";
+
+    // Platform-specific values
+    public static final String PROFILE_USER = "userProfile";
```
Prvnkmr337: these values are not separated by '_' is that expected? Shouldn't all the value be consistent format?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
```diff
@@ -0,0 +1,46 @@
...
+
+import androidx.annotation.NonNull;
+
+/**
+ * SharedPreferences-backed persistence for session correlation IDs.
+ * Used by both OneAuth (via JNI/Djinni) and the broker app directly.
+ * The same SharedPreferences file is written to by OnboardingTelemetryRecorder
+ * on block detection for app-kill resilience.
+ */
+public class OnboardingSessionCachePersistence {
```
Prvnkmr337: nit: Can be renamed to 'OnboardingSessionCorrelationStore' for better naming

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
```diff
@@ -0,0 +1,46 @@
...
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import androidx.annotation.NonNull;
+
+/**
+ * SharedPreferences-backed persistence for session correlation IDs.
+ * Used by both OneAuth (via JNI/Djinni) and the broker app directly.
+ * The same SharedPreferences file is written to by OnboardingTelemetryRecorder
```
Prvnkmr337: This implies the two apps share the same persisted data, but that's not how Android works. 

SharedPreferences accessed via context.getSharedPreferences(..., MODE_PRIVATE) is scoped to the calling app's sandbox.  OneAuth (running inside Teams/Outlook) and Broker (running inside Authenticator) are separate APKs with separate data directories — they each get their own isolated copy of this file at /data/data/<package_name>/shared_prefs/.

please clarify what you mean by 'same SharedPreferrences file is written'

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Load the persisted session correlation cache JSON string.
+     * @return JSON string, or empty string if nothing is persisted
+     */
+    @NonNull
+    public String load() {
+        SharedPreferences prefs = mContext.getSharedPreferences(PREFS_FILE, Context.MODE_PRIVATE);
+        String value = prefs.getString(PREFS_FILE, "");
```
Prvnkmr337: is same constant name used or Pref file and loading content?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
```diff
@@ -0,0 +1,46 @@
...
+        return value != null ? value : "";
+    }
+
+    /**
+     * Save the session correlation cache JSON string to SharedPreferences.
+     * @param json The JSON string to persist
+     */
+    public void save(@NonNull String json) {
+        SharedPreferences prefs = mContext.getSharedPreferences(PREFS_FILE, Context.MODE_PRIVATE);
+        prefs.edit().putString(PREFS_FILE, json).apply();
```
Prvnkmr337: The whole purpose of this persistence is app-kill resilience, but apply() is asynchronous — the write is scheduled to disk but can be lost if the process is killed before the disk flush completes. If durability is important, commit() (synchronous, with @SuppressLint("ApplySharedPref")) would be more appropriate. At minimum, the tradeoff should be documented in the Javadoc.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+        mClientId = clientId;
+        mTarget = target;
+        mContext = context.getApplicationContext();
+
+        // Deserialize seed JSON
+        String schemaVersion = "";
+        String sessionCorrelationId = "";
+        String onboardingMode = "";
+        try {
+            JSONObject seed = new JSONObject(seedJson);
```
Prvnkmr337: The constructor creates local variables (schemaVersion, sessionCorrelationId, onboardingMode) only to assign them to final fields after the try-catch.  A cleaner approach would be a helper parseSeedOrNull(seedJson) that returns a JSONObject or null, then assign fields directly. 

 In Kotlin, this would be trivially clean with runCatching or a let block.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+            schemaVersion = seed.optString(FIELD_SCHEMA_VERSION, "");
+            sessionCorrelationId = seed.optString(FIELD_SESSION_CORRELATION_ID, "");
+            onboardingMode = seed.optString(FIELD_ONBOARDING_MODE, "");
+        } catch (JSONException e) {
+            // Corrupted seed — use empty values
```
Prvnkmr337: The JSONException from a corrupted seed is silently swallowed. Per team convention, the Logger class should be used to log at least a warning. A corrupted seed blob means all telemetry for that entire auth flow is lost — empty session correlation ID, no persistence on block detection, and finalizeBlob() would emit empty seed fields. This is exactly the kind of failure that needs to be visible in diagnostics.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+        mContext = context.getApplicationContext();
+
+        // Deserialize seed JSON
+        String schemaVersion = "";
+        String sessionCorrelationId = "";
+        String onboardingMode = "";
+        try {
+            JSONObject seed = new JSONObject(seedJson);
+            schemaVersion = seed.optString(FIELD_SCHEMA_VERSION, "");
+            sessionCorrelationId = seed.optString(FIELD_SESSION_CORRELATION_ID, "");
```
Prvnkmr337: If sessionCorrelationId is empty (from a corrupted or incomplete seed), the constructor silently proceeds. Later, if blocking errors are recorded, finalizeBlob() will emit a blob with "session_correlation_id": "" to MATS — that's bad telemetry data with an empty correlation ID that can't be correlated to anything. The constructor should either:

Log a warning when critical seed fields are empty/missing, OR
finalizeBlob() should guard against emitting a blob with an empty correlation ID

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+        mOnboardingMode = onboardingMode;
+    }
+
+    /**
+     * Record a step in the onboarding flow.
+     *
+     * @param stepId    Step ID constant (from OnboardingBlobFieldKeys)
+     * @param isoTimestamp The time when the step occurred, as ISO 8601 string
+     */
+    public void addStep(@NonNull String stepId, @NonNull String isoTimestamp) {
```
Prvnkmr337: The recorder should know when a step was recorded — it's an implementation detail. Requiring callers to construct and pass an ISO 8601 timestamp string introduces risks: inconsistent formats, wrong timezone, stale timestamps from delayed calls. The recorder should capture the current time internally. The only valid reason for an external timestamp would be if the time originates from a different system (e.g., C++ core or server response) — please clarify if any caller needs this.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+     * Finalize the blob and return the JSON string.
+     * If no blocking errors were recorded, returns empty string (clears seed blob).
+     * Otherwise serializes the populated blob to JSON.
+     *
+     * @return Populated blob JSON string, or empty string if no blocking errors
+     */
+    @NonNull
+    public String finalizeBlob() {
+        if (mBlockingErrors.isEmpty()) {
+            return "";
```
Prvnkmr337: No logging

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+                JSONArray flows = new JSONArray();
+                for (String flow : mUxFlowUsed) {
+                    flows.put(flow);
+                }
+                blob.put(OnboardingBlobFieldKeys.UX_FLOW_USED, flows);
+            }
+
+            return blob.toString();
+        } catch (JSONException e) {
+            return "";
```
Prvnkmr337: No logging in catch — entire flow's telemetry silently lost

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Finalize the blob and return the JSON string.
+     * If no blocking errors were recorded, returns empty string (clears seed blob).
+     * Otherwise serializes the populated blob to JSON.
+     *
+     * @return Populated blob JSON string, or empty string if no blocking errors
+     */
+    @NonNull
+    public String finalizeBlob() {
```
Prvnkmr337: finalizeBlob() returns "" in two completely different scenarios:

1. No blocking errors were recorded (normal, expected)
2. JSON serialization threw an exception (abnormal, data loss)

Callers cannot distinguish these cases. Consider logging (at minimum) or returning a distinct sentinel for failure.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,264 @@
...
+            JSONObject cache;
+            if (existing != null && !existing.isEmpty()) {
+                cache = new JSONObject(existing);
+            } else {
+                cache = new JSONObject();
+            }
+
+            String key = mClientId + "|" + mTarget;
+            JSONObject entry = new JSONObject();
+            entry.put("id", mSessionCorrelationId);
```
Prvnkmr337: The "id" key in persistSessionCorrelation() is hardcoded rather than using a constant — minor consistency gap but acceptable for an internal cache schema.

Prvnkmr337 (changes requested):

common4j/src/main/com/microsoft/identity/common/java/telemetry/OnboardingBlobFieldKeys.java
wzhipan: The step IDs are values (not JSON keys) and are passed through verbatim into the telemetry blob's step_id field. They match the C++ side and the values the MATS team agreed to in the schema.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
wzhipan: The file is single-key (one JSON-encoded blob holding all entries), so the file name and the only key are the same string.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCachePersistence.java
wzhipan: Rationale documented in the updated Javadoc

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
wzhipan: guard is added

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
wzhipan: Added differentiating logs. Kept the String return to avoid rippling a nullable/Result contract to callers.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.java
wzhipan: Internalized timestamp

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+    // Populated fields
+    private val stepsList: MutableList<StepEntry> = mutableListOf()
+    private val blockingErrors: MutableList<String> = mutableListOf()
+    private var lastLoadedDomain: String? = null
+    private var profile: String? = null
+    private val uxFlowUsed: MutableList<String> = mutableListOf()
+
+    init {
+        // Deserialize seed JSON
+        var sv = ""
```
Prvnkmr337: meaningful variable names can be used here

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+        var scid = ""
+        var om = ""
+        try {
+            val seed = JSONObject(seedJson)
+            sv = seed.optString(FIELD_SCHEMA_VERSION, "")
```
Prvnkmr337: Do we really need temporary variable here?

why not just assign empty strings in catch block?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+    private data class StepEntry(val stepId: String, val timestamp: String)
+
+    /**
+     * Record a step in the onboarding flow. Captures the current time automatically.
+     *
+     * @param stepId Step ID constant (from OnboardingTelemetryConstants)
+     */
+    fun addStep(stepId: String) {
+        val isoTimestamp = SimpleDateFormat(
+            "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US
```
Prvnkmr337: Time format should be class level constant.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+     * Record a blocking error detected during the flow.
+     * Also persists the session correlation entry to SharedPreferences
+     * (best-effort, async) for app-kill resilience.
+     *
+     * @param errorCode The onboarding blocking-error identifier to record
+     *                  (e.g., [OnboardingTelemetryConstants.BLOCKING_ERROR_BROKER_INSTALL]
+     *                  or [OnboardingTelemetryConstants.BLOCKING_ERROR_MDM_FLOW]),
+     *                  not a numeric service auth error code.
+     */
+    fun addBlockingError(errorCode: String) {
```
Prvnkmr337, line 137: API name suggests it is to add blocking error. But, implementation shows it persists session correlation id too, feels design smell.

1. why shuold we persist session correlation id on adding blocking error? Or, should it be persisted else where in flow?
2. Else, we need to rename the API reflecting it like 'addBlockingErrorAndPersistSession'

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+     * Finalize the blob and return the JSON string.
+     * If no blocking errors were recorded, returns empty string (clears seed blob).
+     * Otherwise serializes the populated blob to JSON.
+     *
+     * @return Populated blob JSON string, or empty string if no blocking errors
+     */
+    fun finalizeBlob(): String {
+        if (blockingErrors.isEmpty()) {
+            Logger.verbose(TAG, "finalizeBlob: no blocking errors recorded, returning empty")
+            return ""
```
Prvnkmr337: nit: instead of just returning "" here in other places, can we define a constant like EMPTY_BLOB and assign "" ? And, use that to return from appropriate locations in code?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+        }
+        if (sessionCorrelationId.isEmpty()) {
```
Prvnkmr337, line 191: Should this be this first condition to check in the API?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
Prvnkmr337: this can be info log

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+                    for (flow in uxFlowUsed) {
+                        flows.put(flow)
+                    }
+                    put(OnboardingTelemetryConstants.UX_FLOW_USED, flows)
+                }
+            }
+
+            blob.toString()
+        } catch (e: JSONException) {
+            Logger.error(TAG, "Failed to serialize onboarding blob", e)
```
Prvnkmr337: nit: should we include corerlation id / session correlation id in log to identify ?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,277 @@
...
+     * Uses async [SharedPreferences.Editor.apply] — the in-memory write is
+     * effective immediately, and the disk flush happens shortly after. Acceptable
+     * for this use case: blocking errors leave the app alive for seconds-to-minutes
+     * of user remediation, so the flush window is far longer than typical loss.
+     * Telemetry tolerates rare loss; we avoid main-thread disk I/O.
+     * Called on block detection.
+     */
+    private fun persistSessionCorrelation() {
+        if (sessionCorrelationId.isEmpty()) {
+            return
```
Prvnkmr337, line 271: Add log here?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: The coupling is intentional. The persistence trigger condition IS the block detection. Splitting requires every caller to remember both calls; the side-effect is documented in the Javadoc.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: Order is intentional. "no blocking errors" is the expected path (most flows succeed without blocks), so it's the verbose log; the empty correlation ID is the abnormal path. Order them most-likely-first.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: This fires on every successful auth (no blocks). At info level it would spam production logs. VERBOSE keeps it visible only when verbose logging is enabled for diagnostics.

Prvnkmr337 (approved):
LGTM

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCorrelationStore.kt
```diff
@@ -0,0 +1,59 @@
...
+ * the same schema and file name are used across apps for consistency.
+ * OnboardingTelemetryRecorder also writes to this file on block detection.
+ */
+class OnboardingSessionCorrelationStore(context: Context) {
+
+    private val appContext: Context = context.applicationContext
+
+    /**
+     * Load the persisted session correlation cache JSON string.
+     * @return JSON string, or empty string if nothing is persisted
```
shahzaibj, line 40: why not null if nothing is persisted? Why empty string?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCorrelationStore.kt
```diff
@@ -0,0 +1,59 @@
...
+     * @param json The JSON string to persist
+     */
+    fun save(json: String) {
+        val prefs = appContext.getSharedPreferences(PREFS_FILE, Context.MODE_PRIVATE)
+        prefs.edit().putString(PREFS_FILE, json).apply()
+    }
+
+    companion object {
+        private const val PREFS_FILE = "com.microsoft.oneauth.session_correlation_cache"
+    }
```
shahzaibj, line 58: Could we declare the companion object at the top? I believe that's a general pattern

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+ * Records onboarding telemetry events during interactive auth flows.
+ * Called by WebView navigation fragments to track steps, blocking errors,
+ * and domain navigation. Operates on an in-memory model, converts
+ * to JSON at flow end. On block detection, persists sessionCorrelationId
+ * to SharedPreferences for app-kill resilience (best-effort, async).
+ *
+ * Lives in Android common core so both OneAuth (non-brokered) and broker apps
+ * (brokered) use the same class.
+ *
+ * @param seedJson  The seed blob JSON string from authParameters
```
shahzaibj: Can you add more details regarding seedJson?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+ * @param context   Android context for SharedPreferences access
+ */
+class OnboardingTelemetryRecorder(
+    seedJson: String,
+    private val clientId: String,
+    private val target: String,
+    context: Context
+) {
+
+    private val appContext: Context = context.applicationContext
```
shahzaibj, line 71: Is this different from context in ctor at line 55?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+            seed.optString(FIELD_SCHEMA_VERSION, ""),
+            seed.optString(FIELD_SESSION_CORRELATION_ID, ""),
+            seed.optString(FIELD_ONBOARDING_MODE, "")
+        )
+    } catch (e: JSONException) {
+        Logger.warn(
+            TAG,
+            "Failed to parse onboarding seed JSON; recorder will operate with empty fields: " + e.message
+        )
+        Triple("", "", "")
```
shahzaibj: I'm not a fan of creating these objects with empty strings. Why not return null here as opposed to this fake Triple empty strings?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+     *
+     * @param profile One of [OnboardingTelemetryConstants.PROFILE_USER] or
+     *                [OnboardingTelemetryConstants.PROFILE_WORK]
+     */
+    fun setProfile(profile: String) {
+        this.profile = profile
+    }
+
+    /**
+     * Add a UX flow variant tag.
```
shahzaibj: What is UX flow tag? Please add more details

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+     */
+    fun finalizeBlob(): String {
+        if (blockingErrors.isEmpty()) {
+            Logger.verbose(TAG, sessionCorrelationId, "finalizeBlob: no blocking errors recorded, returning empty")
+            return EMPTY_BLOB
+        }
+        if (sessionCorrelationId.isEmpty()) {
+            Logger.warn(
+                TAG,
+                "finalizeBlob: sessionCorrelationId is empty; dropping blob to avoid emitting uncorrelatable telemetry"
```
shahzaibj: What does it mean to drop it here? What exactly did we drop?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,288 @@
...
+            Logger.verbose(TAG, "persistSessionCorrelation: skipped — no sessionCorrelationId")
+            return
+        }
+
+        try {
+            val prefs = appContext.getSharedPreferences(PREFS_FILE, Context.MODE_PRIVATE)
+            val existing = prefs.getString(PREFS_FILE, "")
+            val cache = if (!existing.isNullOrEmpty()) JSONObject(existing) else JSONObject()
+
+            val key = "$clientId|$target"
```
shahzaibj: target is scopes, right?

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCorrelationStore.kt
wzhipan: The load() signature is dictated by the Djinni-generated SessionCachePersistence.load(): String interface. Because string in Djinni IDL is non-nullable, returning null would crash the JNI layer. Empty string is the contract-correct "nothing persisted" value here, and is also what the C++ side expects.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingSessionCorrelationStore.kt
wzhipan: Both placements exist in this codebase, but the official Kotlin coding convention places the companion object at the bottom to ensure that the primary focus remains on the class's properties and methods.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: updated JavaDoc

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: Yes, we need the application context here in case the recorder lives longer than the activity/fragment.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: updated

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: updated JavaDoc

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: I meant returning the empty-blob sentinel so callers don't push it through to MATS. Updated the log message.

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: Yes, it's sorted, space-joined scopes (matches the C++ SessionCorrelationIdCache::MakeKey(clientId, target) convention so the persisted entries are consistent between OneAuth's C++ cache and the Java-side recorder).

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * The tag identifies which experiment/feature variant the user was exposed to during
+     * the onboarding journey (e.g. a phased rollout cohort like `"MobileOnboardingPhase1"`,
+     * or a remediation experiment like `"MdmEnrollmentRedesign_v2"`). Multiple tags can be
+     * added when several flights apply to the same flow. The values are emitted as the
+     * `ux_flow_used` array in the populated blob and surface in MATS as `mo_ux_flow_used`,
+     * enabling per-experiment slicing of the onboarding funnel.
+     *
+     * @param flowTag Caller-defined experiment/variant identifier.
+     */
```
Veena11, line 174: For AM API flow that @somalaya implemented, the onboarding happens in same Webview, when we detect this flow in webview , could you add a new value to uxFlowUsed -> may be name it as am_api_webcp_flow

Similarly add blocking error : 530003 to blocking error list in this case, step : MDMEnrollmentStarted to the step list

Veena11 (approved):
Looks good, added a comment for AM API flow telemetry

common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
wzhipan: This PR only provides the API to record the fields for both brokered and non-brokered flows, and the recording of AM API will happen in Broker. No change is needed in this PR. I'll have a few other PRs for brokered flows which don't need to be in this hotfix release.

Veena11: sounds good, i thought webview navigation is detected in common core ?
Could you make sure we account for this change if it is in brokered flow?

Veena11: AM API is supported in non brokered too , i guess @somalaya

PR #3100, opened 2026-04-28
Fixes AB#3588022 Optimize load() and getIdTokensForAccountRecord() with filter-then-clone

siddhijain:
Fixes [AB#3588022](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3588022) Apply filter-then-clone optimization to MsalOAuth2TokenCache.load() and getIdTokensForAccountRecord() methods behind the existing ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE flight.

When the flight is enabled, these methods now call the direct (non-input-list) getCredentialsFilteredBy overloads which filter on uncloned in-memory references first, then clone only the matching credentials. This avoids cloning all cached credentials on every AcquireTokenSilent cache-hit path.

Changes:
- Add new getCredentialsFilteredBy overload with kid parameter to IAccountCredentialCache, SharedPreferencesAccountCredentialCache, and SharedPreferencesAccountCredentialCacheWithMemoryCache
- Gate load() to skip preload pattern and use direct filtered lookups when flight is enabled
- Gate getIdTokensForAccountRecord() with same optimization pattern
- Add tests for new kid overload (flight enabled/disabled)
- Add MsalOAuth2TokenCache-level tests for flight-gated load() and getIdTokensForAccountRecord() paths

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [PATCH] Extend filter-then-clone optimization to load() and getIdTokensForAccountRecord() in MsalOAuth2TokenCache: when ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE flight is enabled, skip clone-all preload and call direct flight-gated overloads that clone only matching credentials; add new getCredentialsFilteredBy overload with kid support (#TBD)
 - [PATCH] Move Multiple Listening apps check to the authorization layer (#3070)
 - [PATCH] Add support for Authenticator app activation links in WebView, enabling account pairing/MFA flows to launch Microsoft Authenticator directly instead of redirecting to the Play Store (#3090)
 - [PATCH] Fix: WPJ's BrokerDiscovery cache crash due to shared predefined encryption key with MSAL (#3081)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 vNext
 ----------
-- [PATCH] Extend filter-then-clone optimization to load() and getIdTokensForAccountRecord() in MsalOAuth2TokenCache: when ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE flight is enabled, skip clone-all preload and call direct flight-gated overloads that clone only matching credentials; add new getCredentialsFilteredBy overload with kid support (#TBD)
+- [PATCH] Extend filter-then-clone optimization to load() and getIdTokensForAccountRecord() in MsalOAuth2TokenCache: when ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE flight is enabled, skip clone-all preload and call direct flight-gated overloads that clone only matching credentials; add new getCredentialsFilteredBy overload with kid support (#3100)
 - [PATCH] Move Multiple Listening apps check to the authorization layer (#3070)
 - [PATCH] Add support for Authenticator app activation links in WebView, enabling account pairing/MFA flows to launch Microsoft Authenticator directly instead of redirecting to the Play Store (#3090)
 - [PATCH] Fix: WPJ's BrokerDiscovery cache crash due to shared predefined encryption key with MSAL (#3081)
--- a/common/src/test/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
+++ b/common/src/test/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
@@ -29,6 +29,7 @@ import androidx.test.core.app.ApplicationProvider;
 
 import com.microsoft.identity.common.components.AndroidPlatformComponentsFactory;
 import com.microsoft.identity.common.components.MockPlatformComponentsFactory;
+import com.microsoft.identity.common.internal.mocks.MockCommonFlightsManager;
 import com.microsoft.identity.common.internal.platform.AndroidPlatformUtil;
 import com.microsoft.identity.common.java.interfaces.IPlatformComponents;
 import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Strategy;
@@ -40,6 +41,7 @@ import com.microsoft.identity.common.java.cache.ICacheKeyValueDelegate;
 import com.microsoft.identity.common.java.cache.ICacheRecord;
 import com.microsoft.identity.common.java.cache.MsalOAuth2TokenCache;
 import com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCache;
+import com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredentialCacheWithMemoryCache;
 import com.microsoft.identity.common.java.dto.AccessTokenRecord;
 import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.common.java.dto.Credential;
@@ -47,6 +49,9 @@ import com.microsoft.identity.common.java.dto.CredentialType;
 import com.microsoft.identity.common.java.dto.IdTokenRecord;
 import com.microsoft.identity.common.java.dto.PrimaryRefreshTokenRecord;
 import com.microsoft.identity.common.java.dto.RefreshTokenRecord;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightsManager;
+import com.microsoft.identity.common.java.flighting.IFlightsProvider;
 import com.microsoft.identity.common.java.interfaces.INameValueStorage;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftAccount;
 import com.microsoft.identity.common.java.providers.microsoft.MicrosoftRefreshToken;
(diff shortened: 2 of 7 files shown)


common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -657,45 +659,138 @@ public ICacheRecord load(@NonNull final String clientId,
...
-                        : account.getRealm(),
-                isMultiResourceCapable
-                        ? null // wildcard (*)
-                        : target,
-                null, // not applicable
-                allCredentials
-        );
+        final boolean useFilterThenClone = CommonFlightsManager.INSTANCE
+                .getFlightsProvider()
+                .isFlightEnabled(CommonFlight.ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE);
```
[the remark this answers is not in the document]
siddhijain: This is fine as the in_memory_cache flight for accounts and credentials are enabled on 100% of the devices.

common/src/test/java/com/microsoft/identity/common/MsalOAuth2TokenCacheTest.java
```diff
@@ -1607,4 +1613,155 @@ public void testGetFamilyRefreshTokenForHomeAccountIdNoAccountWithHomeAccountId(
...
+        );
+        final IAccountCredentialCache memoryCache = new SharedPreferencesAccountCredentialCacheWithMemoryCache(
+                keyValueDelegate,
+                storage
+        );
+        return new MsalOAuth2TokenCache<>(
+                components,
+                memoryCache,
+                mockCredentialAdapter
+        );
```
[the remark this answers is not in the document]
siddhijain: This is fine since SharedPreferencesAccountCredentialCacheWithMemoryCache flight is enabled on 100% of the devices.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -880,44 +947,85 @@ public List<IdTokenRecord> getIdTokensForAccountRecord(@Nullable String clientId
                                                            @NonNull AccountRecord accountRecord) {
         final List<IdTokenRecord> result = new ArrayList<>();
 
-        // Load all the credentials to inspect once, such that we don't need to requery the cache
-        // pass these into the new getCredentialsFilteredBy overload, rather than hit disk again
-        final List<Credential> allCredentials = mAccountCredentialCache.getCredentials();
+        final boolean useFilterThenClone = CommonFlightsManager.INSTANCE
+                .getFlightsProvider()
+                .isFlightEnabled(CommonFlight.ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE);
```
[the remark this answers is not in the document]
siddhijain: Same as above. The in-memory cache implementation is enabled on 100% of the devices.

common/src/test/java/com/microsoft/identity/common/SharedPreferencesAccountCredentialCacheWithMemoryCacheTest.java
```diff
@@ -2843,4 +2843,91 @@ public void removeCredential_doesNotCallKeySetOrGetAll() throws Exception {
...
+        at.setCredentialType(CredentialType.AccessToken.name());
+        at.setHomeAccountId(HOME_ACCOUNT_ID);
+        at.setEnvironment(ENVIRONMENT);
+        at.setClientId(CLIENT_ID);
+        at.setRealm(REALM);
+        at.setTarget(TARGET);
+        at.setCachedAt(CACHED_AT);
+        at.setExpiresOn(EXPIRES_ON);
+        at.setSecret(SECRET);
+        mSharedPreferencesAccountCredentialCache.saveCredential(at);
```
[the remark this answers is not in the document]
siddhijain: @copilot help me take care of this comment

changelog.txt
```diff
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [PATCH] Extend filter-then-clone optimization to load() and getIdTokensForAccountRecord() in MsalOAuth2TokenCache: when ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE flight is enabled, skip clone-all preload and call direct flight-gated overloads that clone only matching credentials; add new getCredentialsFilteredBy overload with kid support (#TBD)
```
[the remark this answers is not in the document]
siddhijain: @copilot help me take care of this comment by updating the change log with the right PR number 3100.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -880,44 +947,85 @@ public List<IdTokenRecord> getIdTokensForAccountRecord(@Nullable String clientId
...
-                        clientId,
-                        null, //wildcard (*)
-                        null, //wildcard (*)
-                        accountRecord.getRealm(),
-                        null, // wildcard (*)
-                        null, // not applicable
-                        allCredentials
-                )
-        );
+            v1IdTokens = new ArrayList<>(
```
mohitc1: >new ArrayList<> [](http://example.com/codeflow?start=25&length=15)

why new ArrayList()?

mohitc1: Should we just keep original style of having one list idTokens, and loop over just that list at the end?

common4j/src/main/com/microsoft/identity/common/java/cache/SharedPreferencesAccountCredentialCache.java
```diff
@@ -570,6 +570,41 @@ public List<Credential> getCredentialsFilteredBy(
...
+                        applicationIdentifier,
+                        mamEnrollmentIdentifier,
+                        realm,
+                        target,
+                        authScheme,
+                        requestedClaims,
+                        kid,
+                        false
+                )
+        );
```
mohitc1, line 603: ... new ArrayList<>(result.addAll(
                getCredentialsFilteredByInternal(
                        allCredentials,
                        homeAccountId,
                        environment,
                        credentialType,
                        clientId,
                        applicationIdentifier,
                        mamEnrollmentIdentifier,
                        realm,
                        target,
                        authScheme,
                        requestedClaims,
                        kid,
                        false
                )
        ))

direct initialization in constructor?

siddhijain: yeah makes sense. got this fixed in the latest commit.

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
siddhijain: yeah makes sense. got it updated in the latest commit.

mohitc1 (approved):
:shipit:

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
```diff
@@ -657,45 +659,138 @@ public ICacheRecord load(@NonNull final String clientId,
...
+        final boolean useFilterThenClone = CommonFlightsManager.INSTANCE
+                .getFlightsProvider()
+                .isFlightEnabled(CommonFlight.ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE);
+
+        final List<Credential> accessTokens;
+        final List<Credential> idTokens;
+        final List<Credential> v1IdTokens;
+        List<Credential> refreshTokens;
+
+        if (useFilterThenClone) {
```
rpdome, line 678: I suspected the original implementation went the "snapshot everything" route because if any write (from another thread) lands in the middle of the four getCredentialsFilteredBy calls, those calls will be executed against the "snapshot" = the output is unaffected.

While the new getCredentialsFilteredBy uses mCacheLock, the lock is released between calls. So another thread could still write (add/update/remove) to the cache 
e.g. after `accessTokens = mAccountCredentialCache.getCredentialsFilteredBy`. but before `refreshTokens = mAccountCredentialCache.getCredentialsFilteredBy`.

Could there be a side effect from that?

siddhijain: You're right - the old snapshot approach gave a consistent view, and the new path releases the lock between calls.
 
 I believe it's acceptable because (1) identical commands are deduped via sExecutingCommandMap, (2) worst case is we read old AT + new RT (or vice versa), both functionally valid.
 
 That said, if you'd prefer preserving snapshot consistency, we could wrap all four calls under a single lock acquisition. The perf benefit (clone only matches vs clone-all) would still be preserved — filtering is cheap, cloning is what was expensive. Let me know your thoughts.

rpdome: Let's look for a way to cleanly wrap them.

My concern is more around the "removal" logic - especially the "remove stale token" logic which caused the "AT is null" concurrency issue in broker (which you've already fixed IIRC)

rpdome (approved):

common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java
siddhijain: Added locks in the latest commit.

PR #3103, opened 2026-04-30
Add StressTest annotation, Fixes AB#3582859

rpdome:
[AB#3582859](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3582859)

We're introducing stress test to the pipeline. To make sure we're not causing a cascade failure (due to this test eating the "time quota" of existing runs). We will create an exclusive stage for stresstest only.

This annotation will be used to mark which test should be in that stage.


siddhijain (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/rules/RetryTestRule.java
```diff
@@ -57,6 +58,20 @@ public void evaluate() throws Throwable {
...
+                // attempts floor) when the test or its class is a @StressTest.
+                final boolean isStressTest =
+                        description.getAnnotation(StressTest.class) != null
+                                || description.getTestClass().getAnnotation(StressTest.class) != null;
+
+                if (isStressTest) {
+                    Logger.i(TAG, "StressTest detected - running a single attempt with no retries");
+                    base.evaluate();
+                    return;
+                }
```
fadidurah, line 73: Good call, we definitely dont want this running multiple times lol

fadidurah (approved):

PR #3115, opened 2026-05-14
common4j: don't tie sourcesJar/javadocJar to assemble

fadidurah:
## Why

`common4j` (the first library built in our pipeline) currently takes ~52 minutes per build. Investigation surfaced that we run javadoc generation on *every* CI flow because it's wired into the `assemble` task graph, even when the flow only needs to compile, run tests, or run spotbugs. Generating javadoc for common4j is one of the slower individual tasks in the build.

## What

Drop `assemble.dependsOn sourcesJar, javadocJar` from `common4j/build.gradle`.

These two tasks are already wired into the **publish** graph through `components.java` + the `maven-publish` plugin, so they still run automatically as part of `publish` (which is what creates them in our published artifact). They just no longer run when we only `assemble`.

## Risk

Very low. Local developers who explicitly want the source/javadoc jars on disk can still run `./gradlew sourcesJar javadocJar`. The published artifact is unchanged.

## Companion PR

1ES-Pipelines (ADO PR — link to be added) adds `Cache@2` tasks for `~/.gradle/caches`, `--build-cache --parallel` flags, and combines spotbugs + publish into a single Gradle invocation.


fadidurah:
Folding into existing PR #3113 instead.

PR #3111, opened 2026-05-12
Broker IPC plumbing for onboarding telemetry blob, Fixes AB#3568357

wzhipan:
## Summary

Adds the IPC plumbing for the onboarding telemetry blob to flow client↔broker. Independent of PR #3088 (which adds the recorder/store/constants); both are needed together for end-to-end brokered onboarding telemetry.

Linked Feature: [AB#3462876](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3462876)
Linked PBI: [AB#3568357](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3568357)

## Changes

| File | Change |
|---|---|
| `BrokerRequest.java` | Add `onboarding_seed_json` field (client → broker) |
| `BrokerResult.java` | Add `onboarding_blob` field + builder + getter (broker → client) |
| `InteractiveTokenCommandParameters.java` | Add `onboardingSeedJson` field on the params builder |
| `AcquireTokenResult.java` | Add `onboardingBlob` field for carrying the populated blob through the result chain |
| `MsalBrokerRequestAdapter.java` | Serialize `onboardingSeedJson` from command parameters into `BrokerRequest` |
| `MsalBrokerResultAdapter.java` | Extract `onboardingBlob` from `BrokerResult` into `AcquireTokenResult` |

All additions are pure data fields + adapter wiring. No behavior change for existing callers (fields default to empty/null when not set).

## Design

- **Direction `client → broker`**: OneAuth/MSAL builds the seed JSON (containing `sessionCorrelationId`, `onboardingMode`, `schema_version`) and attaches it to the interactive request. The broker reads it via the new field on `BrokerInteractiveTokenCommandParameters`, constructs its own `OnboardingTelemetryRecorder` from the seed, and uses the same correlation ID for its onboarding telemetry events.
- **Direction `broker → client`**: When the broker emits the populated onboarding blob (after `finalizeBlob()`), it places it into `BrokerResult.onboarding_blob`. The client extracts it into `AcquireTokenResult.onboardingBlob`, where downstream OneAuth code reads it and emits the blob through MATS.

See full design: [Mobile Onboarding Telemetry Design](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3462876) §6.6 (persistence and propagation) and §11.2 (broker → OneAuth blob delivery).

## Dependencies

[...]

PR #3123, opened 2026-05-18
Propagate onboarding telemetry blob through broker success and failure paths, Fixes AB#3462876

wzhipan:
Makes `MsalBrokerResultAdapter` carry the onboarding telemetry blob symmetrically across **every** broker outcome (success, failure, cancel), closing the failure-path gap Veena flagged on #3111 and unblocking the broker recorder-lifecycle work.

**What this adds:**

### Failure path (was the original A3 scope)
1. `BaseException.onboardingBlob` field with getter/setter (common4j).
2. `MsalBrokerResultAdapter.bundleFromBaseException` now serializes `exception.onboardingBlob` into the result bundle (symmetric with the existing `ClientDataInfo` serialization right above it).
3. `MsalBrokerResultAdapter.getBaseExceptionFromBundle` extracts the blob from the deserialized `BrokerResult` and attaches it to the reconstructed exception via `setOnboardingBlob`.

### Success path (added in second commit)
4. MSAL-only overloads on `MsalBrokerResultAdapter`:
   - `bundleFromAuthenticationResult(result, onboardingBlob, version)`
   - `buildBrokerResultFromAuthenticationResult(result, onboardingBlob, version)`

   The broker calls these from `AccountChooserActivity.returnSuccessToCallingActivity` (upcoming PR) to ship the finalized recorder blob to the client. Existing overloads are preserved and delegate to the new ones with `null`, so no behavior change for existing callers (including `AdalBrokerResultAdapter` which doesn't need this).

After this, OneAuth / Common can emit onboarding telemetry on the brokered **success**, **failure**, and **cancel** outcomes via:
- `acquireTokenResult.getOnboardingBlob()` (success — already wired in #3111)
- `exception.getOnboardingBlob()` (failure/cancel — new)

**Why now:** Required by upcoming broker recorder-lifecycle work in PR #163 follow-ups, which writes the recorder blob on every termination path.

**Tests:** `MsalBrokerResultAdapterTests` adds 4 new round-trip tests (2 failure-path, 2 success-path). All 6 `testOnboardingBlob_*` tests pass locally:
`./gradlew :common:testLocalDebugUnitTest --tests *MsalBrokerResultAdapterTests*testOnboardingBlob*`

**Safety:** Telemetry-only; never affects auth logic. No flight gate (blob is only present if a prior PR seeded it).

Fixes [AB#3462876](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3462876)


Prvnkmr337:
Overall change LGTM, ensure the build and tests are passing. Ensure on oneAuth side when reading blob check for Null or empty condition before operating on it.

wzhipan:
Rebased onto latest dev and resolved changelog conflicts. Please re-review when you get a chance.

Prvnkmr337 (approved):
Please check the coverage check failure, seems like false positive.

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Record a step in the onboarding flow. The implementation captures a timestamp
+     * for each step internally.
+     *
+     * @param stepId Step ID constant from
+     *               {@link com.microsoft.identity.common.java.telemetry.OnboardingTelemetryConstants}
+     *               (e.g. {@code STEP_AUTHENTICATION_STARTED}).
+     */
+    void addStep(@NonNull String stepId);
```
shahzaibj, line 52: I feel we should define the steps as enums

shahzaibj (approved):

Prvnkmr337 (approved):

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
wzhipan: I'll make the change as a follow-up

PR #3135, opened 2026-06-02
Disabling GetCookies API from webapps APIs, Fixes AB#3624552

melissaahn:
### Summary
We are disabling the GetCookies API as Edge will use getAllSsoTokens instead. There shouldn't be any dependencies here as Edge nor anyone else has actually been calling this API yet.

[AB#3624552](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3624552)


melissaahn:
Assemble Consumers (Validate Broker) is expected to fail here because a broker unit test depends on the contract value. This change will be made in the broker repo. Otherwise, other tests pass.
image

mohitc1 (approved):

p3dr0rv (approved):

PR #3136, opened 2026-06-03
Security: tighten redirect URI matching to prevent prefix-confusion attacks (FireWatch c1bf88bd), Fixes AB#3665141

wzhipan:
Fixes [AB#3665141](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3665141) — redirect URI prefix-confusion hardening.

## Summary

Tighten redirect URI matching in `AzureActiveDirectoryWebViewClient.isRedirectUrl` to use **structured scheme + authority + path comparison** instead of `String#startsWith`, eliminating a prefix-confusion attack vector. Gated by an ECS kill switch.

## Why

`isRedirectUrl` used `url.startsWith(mRedirectUrl.toLowerCase(Locale.US))`. Because `startsWith` is a prefix match, a URL such as

```
msauth://com.outlook/HASH=.attacker.com/x?code=<AUTH_CODE>
```

passed the check for the registered redirect URI `msauth://com.outlook/HASH=`, and the OAuth2 auth code was handed to `processRedirectUrl()` on a URL controlled by an attacker. eSTS performs **exact** redirect URI validation server-side, but the client check should be exact too (defense-in-depth, MSRC Ceiling Rule 3).

## What changes

**`isRedirectUrl()`** — replaced the prefix match with `Uri.parse`-based comparison requiring scheme + authority (host[:port]) + path to all match (case-insensitive). Query and fragment are ignored (they carry `code`/`state` on legit redirects). A single trailing slash on the path is normalized, so a registered `/auth` still matches an incoming `/auth/`. Any other path difference is rejected. Unparseable URLs fail closed (`false`).

**Opaque redirect URIs (e.g. the broker OOB redirect `urn:ietf:wg:oauth:2.0:oob`)** — opaque URIs have null authority/path, so a naive scheme+authority+path comparison would degenerate to scheme-only and accept any same-scheme `urn:*`. These are now compared by scheme-specific part (minus any appended `?query`/`#fragment`), and a spoofed URI must be opaque on both sides. Parse-failure logging records the exception **type only** (never the throwable message, which could embed the URL and thus the auth code).

**ECS kill switch** — `ENABLE_STRICT_REDIRECT_URI_MATCHING` CommonFlight (default **on**). When disabled via ECS, `isRedirectUrl` reverts to the historical `String#startsWith` behavior, allowing a config-only rollback without a code change.

## ⚠️ Known limitation — msauth:// install path (tracked as follow-up)

For **msauth:// redirect URIs (the broker case)**, this fix alone is **not** sufficient. After `isRedirectUrl` correctly rejects a spoofed `msauth://…` URL, the very next branch in `handleUrl` — `isInstallRequestUrl` (`BROWSER_EXT_INSTALL_PREFIX = "msauth://"`) — matches **any** `msauth://` URL, and `processInstallRequest` also calls `RawAuthorizationResult.fromRedirectUri(url)`, delivering the code to the completion callback. So for msauth the prefix-confusion outcome is unchanged at the callback.

The fix **is** effective for non-msauth (e.g. `https://`, custom-scheme) redirect URIs, which is why the new tests use an `https://` redirect URI to isolate and prove the `isRedirectUrl` behavior.

**Behavior delta for msauth (no regression):** only the *routing* of a spoofed/suffixed `msauth://` URL changes — it now falls through from `isRedirectUrl` to `isInstallRequestUrl` and is handled by `processInstallRequest` instead of `processRedirectUrl`. The auth-code outcome at the callback is identical (both deliver the same `COMPLETED` result), so **no legitimate redirect or install request changes behavior**. The only observable difference is for attack/malformed traffic: it is now logged and counted as a broker-install step (`STEP_BROKER_INSTALL_PROMPTED`) rather than a redirect — cosmetic, and fully reverted by the `ENABLE_STRICT_REDIRECT_URI_MATCHING` kill switch.

**Plan:** closing the msauth path (tightening `isInstallRequestUrl` / `processInstallRequest` so a generic `msauth://…?code=…` without `app_link`/`wpj` is not treated as an install request) will be handled as a **separate follow-up PR**, to keep this defense-in-depth change focused and independently shippable.

## Tests

Assert the real security property via the completion callback (not just the `shouldOverrideUrlLoading` return value), using `https://` and opaque `urn:` redirect URIs:

| Test | Asserts |
|---|---|
| `testStrictMatching_acceptsLegitimateHttpsRedirect` | Legit redirect -> code delivered once |
| `testStrictMatching_acceptsTrailingSlashPathDifference` | `/auth` registered, `/auth/` incoming -> still delivered |
| `testStrictMatching_acceptsLegitimateOpaqueOobRedirect` | Legit `urn:ietf:wg:oauth:2.0:oob?code=…` -> code delivered (COMPLETED) |
| `testStrictMatching_rejectsSpoofedOpaqueOobRedirect` | `urn:…:oobstolen?code=…` -> never delivered as COMPLETED |
| `testStrictMatching_rejectsSpoofedRedirectWithSuffixHost` | `…/auth.attacker.com/x?code=…` -> callback never invoked |
| `testStrictMatching_rejectsSpoofedRedirectWithPathSuffix` | `…/authstolen?code=…` -> callback never invoked |
| `testKillSwitch_disablesStrictMatching_revertsToPrefixMatch` | Flag off -> prefix match restored, spoof delivered again |
| `testUrlOverrideHandlesRedirectUriString` / `…WithFragment` | msauth legit redirect (query / fragment) still handled |

## References

- FireWatch finding `c1bf88bd-5fce-454c-a028-cbfe176639e0`
- IcM [31000000624712](https://portal.microsofticm.com/imp/v5/incidents/details/31000000624712/summary)
- TSG: https://aka.ms/IDSEC_AMCConfirmedVulnerabilites

## Validation

Local Gradle build needs ADO credential setup unavailable here; PR pipelines run the full unit-test suite. Branch is up to date with `dev`.

---
Opened by Copilot CLI on behalf of @zhipanwang for IcM 31000000624712.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -492,8 +492,79 @@ private boolean isPasskeyUrl(@NonNull final String url) {
...
+                return false;
+            }
+            // Single trailing slash is normalized so "/auth" matches "/auth/".
+            return normalizePath(actual.getPath())
+                    .equalsIgnoreCase(normalizePath(expected.getPath()));
+        } catch (final Throwable t) {
+            // Fail closed on unparseable URLs rather than risk leaking the code.
+            Logger.warn(TAG, "Failed to parse URL for redirect URI comparison: " + t);
+            return false;
+        }
```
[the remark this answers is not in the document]
wzhipan: Good catch — confirmed and fixed in 1cb637eb8. `urn:ietf:wg:oauth:2.0:oob` is exactly `AuthenticationConstants.Broker.BROKER_REDIRECT_URI`, and being opaque it returns null authority/path, so the comparison degenerated to scheme-only and would accept any `urn:...?code=STOLEN`.

Fix: when either URI is opaque, require both to be opaque and compare the **scheme-specific part** (minus appended query/fragment) instead of authority/path. Also stopped logging the `Throwable` on the parse-failure path since its message could embed the URL/auth code — now logs a static message.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -350,6 +368,134 @@ public void testUrlOverrideHandlesRedirectUriString() {
...
+    @Test
+    public void testStrictMatching_rejectsSpoofedRedirectWithSuffixHost() throws ClientException {
+        assertSpoofedRedirectNotDelivered(HTTPS_REDIRECT_SPOOFED_SUFFIX_HOST);
+    }
+
+    @Test
+    public void testStrictMatching_rejectsSpoofedRedirectWithPathSuffix() throws ClientException {
+        assertSpoofedRedirectNotDelivered(HTTPS_REDIRECT_SPOOFED_PATH_SUFFIX);
+    }
+
```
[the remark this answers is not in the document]
wzhipan: Added in 1cb637eb8:
- `testStrictMatching_acceptsLegitimateOpaqueOobRedirect` — configured `urn:ietf:wg:oauth:2.0:oob`, incoming `urn:ietf:wg:oauth:2.0:oob?code=AUTH_CODE&state=xyz` -> delivered as a `COMPLETED` auth result.
- `testStrictMatching_rejectsSpoofedOpaqueOobRedirect` — incoming `urn:ietf:wg:oauth:2.0:oobstolen?code=STOLEN` -> not delivered as a `COMPLETED` result (it falls through to the SSL-protection error path; asserted via the captured result code).

mohitc1:
>             } else if (isWebsiteRequestUrl(formattedURL)) {

in isSwitchBrowserRequest() case, it's formattedUrl is <redirecturi>/switch_browser. Now, isRedirectUrl() check would fail for this right?

You might want to separate branch for isSwitchBrowserRedirect()

---
Refers to: common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java:358 in 9507dc9. [](commit_id = 9507dc965ed73b1914e79d2e845fa68deea2fbeb, deletion_comment = False)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -492,8 +492,108 @@ private boolean isPasskeyUrl(@NonNull final String url) {
...
+                }
+                return stripQueryAndFragment(actual.getSchemeSpecificPart())
+                        .equalsIgnoreCase(stripQueryAndFragment(expected.getSchemeSpecificPart()));
+            }
+
+            if (!equalsIgnoreCaseNullSafe(actual.getAuthority(), expected.getAuthority())) {
+                return false;
+            }
+            // Single trailing slash is normalized so "/auth" matches "/auth/".
+            return normalizePath(actual.getPath())
```
shahzaibj, line 562: The description argues the msauth fallthrough is safe because processRedirectUrl and processInstallRequest both deliver the same COMPLETED result. I don't think that holds for switch_browser though.

switch_browser comes in as `{mRedirectUrl}/switch_browser?code=...&action_uri=...`. Before, that matched isRedirectUrl (prefix) and then the nested isSwitchBrowserRequest on line 351 sent it to processSwitchBrowserRequest, which launches the system browser. It never went through processRedirectUrl.

Now the exact path match here rejects it (`/hash=/switch_browser` != `/hash=`, normalizePath only drops one trailing slash), so it falls through to isInstallRequestUrl -> processInstallRequest. No app_link/wpj/cancel on a switch_browser url so it comes back COMPLETED and we just forward it to the callback and stopLoading().

So the browser handoff never fires and we hand the switch_browser continuation code back as if it were the final auth code. That's a legit flow regressing, on by default. Is switch_browser meant to be covered by the isInstallRequestUrl follow-up, or should it still work through here?

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -350,6 +376,190 @@ public void testUrlOverrideHandlesRedirectUriString() {
         assertTrue(mWebViewClient.shouldOverrideUrlLoading(mMockWebView, TEST_REDIRECT_URL));
```
shahzaibj: Doesn't this pass now even if isRedirectUrl is broken? TEST_REDIRECT_URL is msauth:// now, so if isRedirectUrl returned false it'd just fall through to isInstallRequestUrl and shouldOverrideUrlLoading comes back true either way. The new https tests pin isRedirectUrl itself, but nothing asserts the msauth url actually goes through processRedirectUrl and not processInstallRequest. Maybe capture the callback result here and assert the COMPLETED came from the redirect path?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
wzhipan: Good catch — fixed. You're right that strict matching pushed `switch_browser` through the install path. I moved the `switch_browser` check into its own branch ahead of `isRedirectUrl` in `handleUrl`, so it's routed correctly regardless of the strict-matching flag. Added a regression test that fails without the fix.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
wzhipan: Agreed — updated. The test now captures the callback result and asserts the redirect path was actually taken (not the install fallthrough), so a broken `isRedirectUrl` can no longer pass.

wzhipan:
@mohitc1 Done — `switch_browser` now has its own branch in `handleUrl`, evaluated before `isRedirectUrl`, so the exact-path match no longer hides it. Added a regression test covering this routing.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -344,16 +344,18 @@ private boolean handleUrl(final WebView view, final String url) {
...
-                    Logger.info(methodTag,"It is a redirect request.");
-                    processRedirectUrl(view, url);
-                }
+            }
+            // A switch_browser request arrives as {redirectUrl}/switch_browser?code=...&action_uri=...,
+            // whose path no longer equals the registered redirect URI under strict matching. Detect it
+            // here explicitly instead of nesting it behind isRedirectUrl; otherwise it would fall through
+            // to isInstallRequestUrl/processInstallRequest and the switch_browser continuation code would
+            // be delivered to the completion callback as if it were the final auth code.
+            else if (mSwitchBrowserProtocolCoordinator.isSwitchBrowserRequest(formattedURL, mRedirectUrl)) {
```
shahzaibj, line 354: switch_browser is checked ahead of isRedirectUrl now, so it's the first matcher every url hits. But isSwitchBrowserRedirectUrl is still `url.startsWith("$redirectUrl/switch_browser", ignoreCase = true)`, the same prefix match we're removing right below in isRedirectUrl. So `{redirectUrl}/switch_browser.evil.com/x?code=..&action_uri=..` still gets routed into processSwitchBrowserRequest.

validateActionUri pins action_uri to a valid AAD host so the code isn't going anywhere attacker controlled, so I don't think it's an actual leak. But the whole point here is to match exactly on the client even though eSTS already validates. By that logic is switch_browser in scope for the same tightening, or deliberately left on startsWith?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -491,8 +493,108 @@ private boolean isPasskeyUrl(@NonNull final String url) {
...
+     * match that let attacker-controlled path suffixes through
+     * (FireWatch c1bf88bd / IcM 31000000624712); defense-in-depth, since eSTS
+     * already validates the redirect URI exactly server-side.
+     *
+     * @param url lowercased URL from {@link #handleUrl}; comparison is
+     *            case-insensitive so mixed-case registered URIs still match.
+     */
     private boolean isRedirectUrl(@NonNull final String url) {
-        return url.startsWith(mRedirectUrl.toLowerCase(Locale.US));
+        if (mRedirectUrl == null || mRedirectUrl.isEmpty()) {
```
mohitc1: >mRedirectUrl == null  [](http://example.com/codeflow?start=12&length=21)

is this null check needed?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
wzhipan: Good catch — tightened it. `isSwitchBrowserRedirectUrl` now does the same structured scheme + authority + path match as `isRedirectUrl` (behind the same `ENABLE_STRICT_REDIRECT_URI_MATCHING` kill switch), so a suffixed `{redirect}/switch_browser.evil.com/x` is no longer routed as a switch_browser request. Legit exact URLs (with query) still match. Added tests for the suffix spoofs and the kill-switch revert.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
wzhipan: Good catch — it's not needed. `mRedirectUrl` is `final` and set from a `@NonNull` constructor arg, so it can't be null. Annotated the field `@NonNull` and dropped the `== null` check, keeping the `isEmpty()` guard.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
```diff
@@ -224,7 +241,56 @@ object SwitchBrowserUriHelper {
...
+                !expectedScheme.equals(actual.scheme, ignoreCase = true) -> false
+                // Opaque URIs (e.g. urn:...) have null authority/path, so compare the
+                // scheme-specific part instead, minus any appended query/fragment.
+                expected.isOpaque || actual.isOpaque ->
+                    expected.isOpaque == actual.isOpaque &&
+                        stripQueryAndFragment(actual.schemeSpecificPart)
+                            .equals(stripQueryAndFragment(expected.schemeSpecificPart), ignoreCase = true)
+                !actual.authority.equals(expected.authority, ignoreCase = true) -> false
+                else -> normalizePath(actual.path).equals(normalizePath(expected.path), ignoreCase = true)
+            }
```
mohitc1, line 273: Can this be simplified like

// .clearQuery() and .fragment(null) safely strip trailing components 

        Uri cleanUri1 = uri1.buildUpon().clearQuery().fragment(null).build();
        Uri cleanUri2 = uri2.buildUpon().clearQuery().fragment(null).build();

     then cleanUri1.equals(cleanUri2) ?


Instead of doing yourself.

mohitc1: Same in main redirect matching too

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -491,8 +494,111 @@ private boolean isPasskeyUrl(@NonNull final String url) {
...
+        }
+
+        try {
+            final Uri actual = Uri.parse(url);
+            final Uri expected = Uri.parse(mRedirectUrl);
+            final String expectedScheme = expected.getScheme();
+
+            // Scheme-less configured URI: fall back to strict equality.
+            if (expectedScheme == null || expectedScheme.isEmpty()) {
+                return url.equalsIgnoreCase(mRedirectUrl);
```
shahzaibj: isSwitchBrowserRedirectUrl strips query/fragment in its scheme-less branch, but here we compare the full `url` (still carries `?code=...`) against `mRedirectUrl`, so a scheme-less redirect carrying a code would match there but not here. Probably dead today since all our redirect URIs have a scheme.. but these two are meant to mirror each other so the divergence stood out. intentional?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
```diff
@@ -46,6 +46,18 @@ object SwitchBrowserUriHelper {
             .isFlightEnabled(CommonFlight.SWITCH_BROWSER_PROTOCOL_REQUIRES_STATE)
     }
 
+    /**
+     * Whether strict (structured) switch_browser redirect matching is enabled. Shares the
```
shahzaibj, line 50: This is a flight specifically for switch browser redirects?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
wzhipan: Good idea — I tried that, but `buildUpon().clearQuery().fragment(null)` + `Uri.equals()` regresses three cases the explicit comparison handles on purpose (confirmed with a quick Robolectric check):

1. **Trailing slash** — `Uri.equals` is exact, so `…/auth` vs `…/auth/` (and the path-less `""` vs `/` case) compare unequal, and a legitimate redirect gets rejected.
2. **Case** — `Uri.equals` is case-sensitive; the incoming URL is lowercased but the registered redirect URI can be mixed-case, so mixed-case registrations would stop matching.
3. **Opaque `urn:` (broker OOB redirect)** — for opaque URIs the `?code` lives in the scheme-specific part, not the query component, so `clearQuery()` doesn't strip it cleanly. In my test it collapsed `urn:…:oob?code=…` down to just `urn:`, which makes any two urns compare equal — reintroducing the scheme-only opaque match the earlier review flagged.

So the manual scheme/authority/path (and ssp for opaque) comparison plus `normalizePath` is deliberate, and the same reasoning applies to `isRedirectUrl` — that's why both matchers share the same shape. Happy to add a short comment noting these cases if it helps readability.

mohitc1: Thanks for trying. Yes, a comment would be good. Also, consider moving method to common place/util, to avoid duplicacy.

mohitc1 (approved):
:shipit:

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
wzhipan: Good catch — real divergence, and you're right they should mirror. Aligned the Java scheme-less branch to `stripQueryAndFragment(url).equalsIgnoreCase(mRedirectUrl)` in db2c542f0, matching the Kotlin `isSwitchBrowserRedirectUrl`. It's currently unreachable (all our redirect URIs carry a scheme) and fails closed, so no behavior change in practice — just consistency. Added a test that delivers a scheme-less redirect as a COMPLETED result (verified it fails against the pre-fix branch).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserUriHelper.kt
wzhipan: It's not switch-browser-specific — it reads the same `ENABLE_STRICT_REDIRECT_URI_MATCHING` flight that gates `isRedirectUrl`. This property is just a local accessor for that shared kill switch, so the main redirect matcher and the switch_browser matcher tighten (or roll back to `String#startsWith`) together. I reused the existing flight deliberately rather than adding a separate switch-browser one.

shahzaibj (approved):

PR #3138, opened 2026-06-05
Fix MSA login timeout: use configurable timeout for 'Other ways to si…, Fixes AB#3622375

somalaya:
The 'Other ways to sign in' UI element check in handlePasswordField() was hardcoded to FIND_UI_ELEMENT_TIMEOUT_SHORT (10s), which was insufficient for the WebView to load login.live.com through the IPC + network chain. This caused intermittent TimeoutExceptions in MSA broker tests.

Changed to use mFindLoginUiElementTimeout (25s default, 30s for NAA tests) which is consistent with other UI element checks in the same handler.

Fixes: TestCase2688459 (NAA ATS), TestCase2637829 (MSA Prompt.Login)

Fixes [AB#3622375](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3622375)


mohitc1 (approved):
:shipit:

cacosta33 (approved):

PR #3145, opened 2026-06-09
Add swtich browser flag to interactive result propertybag, Fixes AB#359061

mohitc1:
This PR adds an explicit signal to the interactive auth result broadcast so downstream consumers (OneAuth) can detect whether the switch_browser protocol flow was initiated during the session. Currently, we don't have standard way to relaying back telemetry from Common to OneAuth, when Common is executing interactive auth. This change will enable OneAuth to track if switch browser flow was initiated both in success and failure cases, when it receives AuthorizationResult via broadcast to RETURN_AUTHORIZATION_REQUEST_RESULT

Changes:

- Add a new IS_SWITCH_BROWSER_FLOW constant to AuthenticationConstants.LocalBroadcasterFields.
- Track “switch browser flow initiated” state in the switch-browser handler/coordinator and propagate it into the authorization result PropertyBag.
- Refactor AuthorizationFragment to centralize PropertyBag construction via an overridable hook.

Fixes [AB#359061](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/359061)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SwitchBrowserRequestHandler.kt
```diff
@@ -121,6 +129,7 @@ class SwitchBrowserRequestHandler(
             activity.startActivity(switchBrowserIntent)
             span.setStatus(StatusCode.OK)
             isSwitchBrowserChallengeActive = true
+            wasSwitchBrowserFlowInitiated = true
             span.end()
```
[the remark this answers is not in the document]
mohitc1: This is existing problem, out of scope current change.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/SwitchBrowserRequestHandler.kt
```diff
@@ -57,6 +57,14 @@ class SwitchBrowserRequestHandler(
 
     var isSwitchBrowserChallengeActive: Boolean = false
 
+    /**
+     * Indicates that the switch browser flow was initiated during this session.
+     * Set when [processChallenge] successfully launches the browser activity.
+     * Unlike [isSwitchBrowserChallengeActive], this flag is never reset.
+     */
+    var wasSwitchBrowserFlowInitiated: Boolean = false
```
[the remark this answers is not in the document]
mohitc1: addressed

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -771,6 +773,17 @@ private SwitchBrowserProtocolCoordinator getSwitchBrowserCoordinator() {
         return mSwitchBrowserProtocolCoordinator;
     }
 
+    @NonNull
+    @Override
+    protected PropertyBag propertyBagFromAuthorizationResult(@NonNull final RawAuthorizationResult result) {
+        final PropertyBag propertyBag = super.propertyBagFromAuthorizationResult(result);
+        if (mSwitchBrowserProtocolCoordinator != null
+                && mSwitchBrowserProtocolCoordinator.getWasSwitchBrowserFlowInitiated()) {
```
shahzaibj, line 786: Is `wasSwitchBrowserFlowInitiated` being read cross-thread now?

Both fields touched here are plain `var`s: `mSwitchBrowserProtocolCoordinator` (this file line 160) and the handler's `wasSwitchBrowserFlowInitiated` (SwitchBrowserRequestHandler.kt:65). Neither has `@Volatile` or `@GuardedBy`.

Writers are main:
- `processChallenge` sets the handler flag, runs on UI via the WebViewClient
- `getSwitchBrowserCoordinator()` lazy-init sets the fragment field, also UI

But `sendResult` -> `propertyBagFromAuthorizationResult` can run on the broadcast executor, not main:
- `CommandDispatcher.java:1014` -> `LocalBroadcaster.broadcast(CANCEL_AUTHORIZATION_REQUEST, ...)`
- `LocalBroadcaster.java:71` dispatches `mCancelRequestReceiver.onReceive` on `sBroadcastExecutor` (single-thread pool declared on line 38, not main)
- receiver -> `cancelAuthorization` -> `sendResult` -> override here

So this is now a main-thread writer + broadcast-executor reader on plain fields.

Failure mode: main writes `true`, executor reads `false`, key is omitted from the bag, OneAuth never sees the signal. Bounded. Not a crash, not an auth break. But it silently under-reports the new flag in exactly the cancel-during-init race window this PR is trying to instrument.

`@Volatile` on `wasSwitchBrowserFlowInitiated`? Or post the read back onto main before building the bag?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -771,6 +773,17 @@ private SwitchBrowserProtocolCoordinator getSwitchBrowserCoordinator() {
...
+            propertyBag.put(IS_SWITCH_BROWSER_FLOW, true);
```
shahzaibj, line 787: Handler tests cover the `wasSwitchBrowserFlowInitiated` flag itself. But what OneAuth actually keys off is what ends up in this PropertyBag. The override here isn't tested.

Easy to silently break in a refactor: rename the key, flip the conditional, drop the null check. Handler-level tests wouldn't catch any of those.

Worth a test that mocks the coordinator and asserts:
- key present and true when coordinator says initiated
- key absent when coordinator says not initiated
- key absent when coordinator is null

changelog.txt
```diff
@@ -1,5 +1,7 @@
 vNext
 ----------
+- [MINOR] Add IS_SWITCH_BROWSER_FLOW flag to PropertyBag so OneAuth can detect when the switch browser protocol was used during interactive auth (#3145)
+- [MINOR] OnboardingTelemetryRecorder.finalizeBlob() emits the populated blob whenever a valid sessionCorrelationId is present (previously returned empty when blockingErrors was empty)
```
shahzaibj: Looks like a stray copy of line 6 (already added by #3148). Drop this one?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
mohitc1: Added @Volatile to wasSwitchBrowserFlowInitiated for safe cross-thread visibility. In practice the write happens on UI thread (processChallenge) and the primary read path is also UI thread (onChallengeResponseReceived -> sendResult), but @Volatile guards defensively against the cancelAuthorization path that dispatches on sBroadcastExecutor. Thanks!

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -159,6 +161,11 @@ public class WebViewAuthorizationFragment extends AuthorizationFragment {
     // This is used by the switch browser protocol to handle the resume of the flow.
     private SwitchBrowserProtocolCoordinator mSwitchBrowserProtocolCoordinator = null;
```
shahzaibj: The handler flag is `@Volatile` now but this field is still plain. Same publication race on the cancel-during-init path: executor reads `mSwitchBrowserProtocolCoordinator != null`, could see stale null even after main's lazy init, conditional fails, key gets omitted.

Mark it `volatile` too?

p3dr0rv (approved):

shahzaibj (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
mohitc1: Addressed.

shahzaibj (approved):

PR #3147, opened 2026-06-10
Add GET_DEVICE_TOKEN_V1 protocol with clientId, redirectUri and full token response, Fixes AB#3658756

mohitc1:
Adds a `GET_DEVICE_TOKEN_V1` device-token protocol that takes the caller's `clientId` and `redirectUri`, and returns the full token response instead of only the access token.

### Why

`getDeviceToken` had no way for a caller to identify itself — the broker always stamped its own `BROKER_CLIENT_ID` and redirect URI into the device token JWT request. Callers also only got back the raw access token, with no expiry, device claims, or token type.

### Changes

- New `GET_DEVICE_TOKEN_V1` protocol constant, `GetDeviceTokenV1Parameters` (required `clientId` + `redirectUri`), and `GetDeviceTokenV1Response`.
- New `DeviceTokenResult` carrying every field the token endpoint returns: `access_token`, `device_info`, `token_type`, `resource`, `expires_in`, `ext_expires_in`, `expires_on`, `not_before`.
- New `DeviceRegistrationClientApplication.getDeviceTokenResult(record, resources, correlationId, clientId, redirectUri, scope?)` returning `DeviceTokenResult`.
- Deprecated `getDeviceToken` (returns `String`, unchanged behaviour), pointing at `getDeviceTokenResult`.
- Unit tests for V1 parameter serialization and full-result round-trip through the IPC packer.

### Design notes

**Why a new protocol version rather than changing V0.** V0 stays on the wire untouched, so existing callers and already-shipped brokers are unaffected. Enforcement of the new required fields lives in the V1 broker executor.

**Why a new method name rather than an overload.** The V1 method was initially an overload of `getDeviceToken`, differing only in arity. That made a silent mis-binding possible: a Java caller writing `getDeviceToken(record, resources, correlationId, clientId)` binds to the deprecated V0 method with `clientId` passed as the `scope`. It compiles, and fails at runtime against the token endpoint. Distinct names remove the overload set, so the mistake becomes a compile error.

**Why `DeviceTokenResult` rather than returning `GetDeviceTokenV1Response`.** The response type is a wire type — it carries `serialize()`, `getProtocolName()`, and a version-bound name. Returning it would recouple the public API to the protocol version, which is the exact coupling V1 exists to avoid, and every other method on this class already unwraps to a domain type.

**Why boxed `Long` for the numeric fields.** Distinguishes "server omitted the field" from "server sent 0", matching `TokenResponse.mExpiresIn` and `MicrosoftTokenResponse.mExtendedExpiresIn`.

**Why all fields are `@Nullable` except `accessToken`.** Only `accessToken` is enforced — the broker throws `DEVICE_TOKEN_EMPTY_OR_NULL` if it's missing. The rest are passthroughs of what the token endpoint returned, and this crosses an IPC boundary where broker and client version independently.

**Why `DeviceTokenResult` carries no `@ToString`/`@Data`.** Either would put the access token into logs. There is a comment on the class saying so.

### Companion PR

Requires the broker-side V1 executor. This PR is a no-op against brokers without it.

[AB#3658756](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3658756)


common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
```diff
@@ -309,18 +311,48 @@ class DeviceRegistrationClientApplication {
...
+        message = "Use the overload that accepts clientId parameter.",
+        replaceWith = ReplaceWith("getDeviceToken(deviceRegistrationRecord, resources, correlationId, clientId, scope)")
+    )
+    @Throws(BaseException::class)
+    fun getDeviceToken(
+        deviceRegistrationRecord: IDeviceRegistrationRecord,
+        resources: String,
+        correlationId: UUID,
+        scope: String?
+    ): String {
```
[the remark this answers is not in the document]
mohitc1: Fixed. Thanks for pointing.

common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
```diff
@@ -309,9 +311,41 @@ class DeviceRegistrationClientApplication {
...
+        resources: String,
+        correlationId: UUID,
+        clientId: String,
+        redirectUri: String,
+        scope: String? = null
+    ): String {
+        val methodTag = "$TAG:getDeviceToken"
+        Logger.info(methodTag, "GetDeviceToken (V1) started. CorrelationId: $correlationId")
+        val responseSerialized = mController.execute(
+            GetDeviceTokenV1Parameters(correlationId, deviceRegistrationRecord, resources, clientId, redirectUri, scope)
```
shahzaibj, line 332: what happens when this V1 overload is called against a broker that doesn't have the V1 executor yet?

i looked locally and neither DeviceRegistrationProtocolExecutorFactory nor the WpjLegacy one has a GET_DEVICE_TOKEN_V1 case, so the call would throw BrokerUpdateRequiredException. is the plan to ship the broker companion first and let it age in field, or have callers catch and fall back to V0?

common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
```diff
@@ -309,9 +311,41 @@ class DeviceRegistrationClientApplication {
...
      * @param scope                    optional scope.
      */
     @Throws(BaseException::class)
+    fun getDeviceToken(
+        deviceRegistrationRecord: IDeviceRegistrationRecord,
+        resources: String,
+        correlationId: UUID,
+        clientId: String,
+        redirectUri: String,
+        scope: String? = null
```
shahzaibj, line 327: @JvmOverloads on this one too? the deprecated overload has it. java callers wouldn't be able to omit scope otherwise.

common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
mohitc1: The new API for new callers specifically MDE (defender) for now. It will be requirement that they should have broker updated to this version. Nothing breaks for caller of existing public method.

common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
```diff
@@ -320,7 +356,7 @@ class DeviceRegistrationClientApplication {
         scope: String? = null
     ): String {
         val methodTag = "$TAG:getDeviceToken"
-        Logger.info(methodTag, "GetDeviceToken started. CorrelationId: $correlationId")
+        Logger.info(methodTag, "GetDeviceToken (V0, deprecated) started. CorrelationId: $correlationId")
```
p3dr0rv, line 359: log warn?

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/deviceregistration/api/DeviceRegistrationClientApplication.kt
mohitc1: Kept at `info` — it's a per-call trace line and V0 is still supported, so warn would fire on every existing call for something that isn't a fault. `@Deprecated` covers the caller who can act.

logdog16 (approved):

shahzaibj (approved):

PR #3148, opened 2026-06-10
Onboarding telemetry: emit blob when sessionCorrelationId is present, AB#3462876, Fixes AB#3462876

wzhipan:
## What

`OnboardingTelemetryRecorder.finalizeBlob()` now returns the populated blob whenever a valid `sessionCorrelationId` is present, instead of returning an empty string when `blockingErrors` is empty.

## Why

The previous early-return on empty `blockingErrors` dropped the session-correlation chain on smooth-success flows. Consumers (broker → OneAuth → xplat) need the populated blob in those flows so they can decide MATS emission based on blob content (e.g. `IsReused` is enough to emit a retry-recovery row even when no blocking error occurred). The decision of whether to emit to MATS now belongs to the consumer.

## What changed

- `OnboardingTelemetryRecorder.finalizeBlob()`: removed the `if (blockingErrors.isEmpty()) return EMPTY_BLOB` early return; kept the `sessionCorrelationId.isEmpty()` integrity check that still returns empty.
- `LAST_BLOCKING_ERROR` field is now guarded by `if (blockingErrors.isNotEmpty())` so it is only emitted when there is actually a blocking error.
- Test renamed `testFinalizeBlob_NoBlockingErrors_ReturnsEmpty` → `testFinalizeBlob_NoBlockingErrors_StillEmitsBlobWithSeedFields`; assertions inverted for the new behavior.
- Changelog entry under `vNext`.

## Risk

Low and isolated:

- **Brokered smooth success**: the populated blob flows broker → OneAuth → xplat. xplat's emission gate is `HasBlockingError(blob) || IsReused(blob)`. On a clean first attempt both predicates are false, so MATS still skips — no regression. On a retry within the 15-min TTL window `IsReused=true` and the row emits, which is the intended recovery telemetry.
- **Brokered failure with a blocking error**: unchanged (`HasBlockingError=true` → emit).
- **Non-brokered flows**: same logic in `EmbeddedBrowser` + xplat's identical gate. No regression.
- **Silent flows**: do not initiate onboarding telemetry at all (no seed creation in `BackgroundRequest`), so unaffected.

[AB#3462876](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3462876)


common/src/main/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorder.kt
```diff
@@ -185,10 +185,6 @@ class OnboardingTelemetryRecorder(
      * @return Populated blob JSON string, or empty string if no blocking errors
```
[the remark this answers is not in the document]
wzhipan: Fixed

Prvnkmr337 (approved):

common/src/test/java/com/microsoft/identity/common/internal/telemetry/OnboardingTelemetryRecorderTest.kt
```diff
@@ -73,8 +73,22 @@ class OnboardingTelemetryRecorderTest {
...
+        // When a valid seed was provided but no blocking errors occurred (smooth-success
+        // flow), the recorder still emits a populated blob so consumers (OneAuth) can
+        // correlate the session and count it toward smooth-success metrics. The decision
+        // to forward to MATS belongs to the consumer based on blob content, not to the
+        // broker / common layer.
+        val result = recorder.finalizeBlob()
+        Assert.assertFalse(result.isEmpty())
+
+        val blob = JSONObject(result)
+        Assert.assertEquals("1.0.0", blob.getString("schema_version"))
```
Prvnkmr337, line 86: nit: Instead of String literals here, the getString() can use actual String Constant defined in source code, if accessible.

wzhipan: Good point. The seed-field keys at this line — `schema_version`, `session_correlation_id`, `onboarding_mode` — are `private const val` in `OnboardingTelemetryRecorder`'s companion (kept private to avoid a dependency on the Djinni-generated `OnboardingBlobConstants` in Common), so they aren't accessible from the test, hence the literals here. The blocking-error keys/values (`blocking_errors`, `last_blocking_error`, `BROKER_INSTALLATION_TRIGGERED`, `MDM_FLOW`) do have public `OnboardingTelemetryConstants` — happy to adopt those in a follow-up so this already-approved PR stays focused.

shahzaibj (approved):

PR #3150, opened 2026-06-10
fix tbca prompt on pixel, Fixes AB#3648301

rpdome:
[AB#3648301](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3648301)

The original change worked on Samsung S23, but not on the pipeline (Pixel)

Made a change to accommodate that, validated via emulator.

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AadLoginComponentHandler.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/interaction/microsoftsts/AadLoginComponentHandler.java
@@ -233,15 +233,22 @@ public class AadLoginComponentHandler implements IMicrosoftStsLoginComponentHand
         // a Keyguard biometric fallback, or an OEM-specific banner) that happens to share the
         // same positive-button resource id.
         //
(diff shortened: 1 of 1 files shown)


fadidurah (approved):

fadidurah:
LGTM, i recently did a test run on the monthly pipeline to validate my lab cleanup changes https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1648534&view=results. I think this fix will address those two tests, right?

rpdome:
> LGTM, i recently did a test run on the monthly pipeline to validate my lab cleanup changes https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1648534&view=results. I think this fix will address those two tests, right?

@fadidurah this is different. The error in your run was cause by the CA not being triggered (so there was no prompt when expected)

I did a fix for this yesterday by changing the CA scope to **_all apps_**. (we were using graph's user.read to test, and it seems like that is now excluded from exchange/sharepoint CA scope)

It no longer regresses for me, but i'm not sure how propagation on eSTS side works. what you saw might be due to that.

My run did see a prompt, though.
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1648761&view=results

somalaya (approved):

PR #3158, opened 2026-06-16
Add resiliency for duplicate PIN screens in handleAppProtectionPolicy, Fixes AB#3667378

mohitc1:
Due to a known issue, the Enter PIN and Confirm PIN screens can appear twice in succession during app protection policy setup. This change extracts PIN entry logic into a helper method and retries if the PIN screen reappears unexpectedly.

Fixes [AB#3667378](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3667378)

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1653057&view=results


uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -397,10 +397,32 @@ public void handleAppProtectionPolicy() {
...
+        final UiObject secondPinField = UiAutomatorUtils.obtainUiObjectWithResourceId(
+                PIN_ENTRY_RESOURCE_ID
+        );
+
+        if (secondPinField.waitForExists(TimeUnit.SECONDS.toMillis(5))) {
+            Logger.i(TAG, "PIN screen appeared a second time, handling again..");
+            enterAndConfirmPin(device);
+        } else {
+            Logger.i(TAG, "PIN screen did not reappear again - fine, continuing..");
+        }
```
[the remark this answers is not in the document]
mohitc1: Fixed.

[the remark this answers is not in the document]
mohitc1: Fixed

fadidurah (approved):
LGTM, only place we can fiddle with is how long to wait for the pin prompt (currently 5 seconds), may need to bump this up if ui is slow

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -397,10 +397,39 @@ public void handleAppProtectionPolicy() {
...
+        );
+        pinFieldAfterFirstEntry.waitUntilGone(TimeUnit.SECONDS.toMillis(5));
+
+        // Due to a known issue, the PIN entry and confirmation screens can appear twice in
+        // succession. If it happens, handle the second occurrence; if not, continue normally.
+        final UiObject secondTimePinField = UiAutomatorUtils.obtainUiObjectWithResourceId(
+                PIN_ENTRY_RESOURCE_ID
+        );
+
+        if (secondTimePinField.waitForExists(TimeUnit.SECONDS.toMillis(5))) {
```
shahzaibj: In the no-dup case (which is the common one) don't we always eat the full 5s on `waitForExists` here before moving on? Just making sure we're ok adding ~5s to every enrollment run to catch a dup that only shows up occasionally.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -397,10 +397,39 @@ public void handleAppProtectionPolicy() {
...
+        // Wait for the PIN field to disappear after the first entry, so we don't confuse the
+        // still-visible first screen with a second occurrence.
+        final UiObject pinFieldAfterFirstEntry = UiAutomatorUtils.obtainUiObjectWithResourceId(
+                PIN_ENTRY_RESOURCE_ID
+        );
+        pinFieldAfterFirstEntry.waitUntilGone(TimeUnit.SECONDS.toMillis(5));
+
+        // Due to a known issue, the PIN entry and confirmation screens can appear twice in
+        // succession. If it happens, handle the second occurrence; if not, continue normally.
+        final UiObject secondTimePinField = UiAutomatorUtils.obtainUiObjectWithResourceId(
```
shahzaibj, line 410: iirc, this method internally already calls waitForExists

can you double check?

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
mohitc1: It's choice between reliability vs extra 5 sec. 

Moreover, the test has 2 re-runs, often it passes in 3rd attempt - going by the video.  So, current success already costs > 5 seconds on frequently. If this change addresses the issue, then we are net good.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
mohitc1: yes. Fixed. Now checking just ".exists()"

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -397,10 +397,39 @@ public void handleAppProtectionPolicy() {
         // get access screen - continue
         UiAutomatorUtils.handleButtonClick("com.microsoft.windowsintune.companyportal:id/positive_button");
 
-        // handle PIN
+        // handle PIN entry and confirmation
+        enterAndConfirmPin(device);
+
+        // Wait for the PIN field to disappear after the first entry, so we don't confuse the
+        // still-visible first screen with a second occurrence.
+        final UiObject pinFieldAfterFirstEntry = UiAutomatorUtils.obtainUiObjectWithResourceId(
```
shahzaibj: This obtain has no timeout arg, so the internal `waitForExists` uses the 25s `FIND_UI_ELEMENT_TIMEOUT` default (line 412 passes 5s, this one doesn't). If the pin field is already gone by the time we hit this line (confirm got accepted quickly), don't we block the full 25s waiting for it to show up before `waitUntilGone` even runs? We only need the handle for `waitUntilGone` here, so maybe pass a short/0 timeout: `obtainUiObjectWithResourceId(PIN_ENTRY_RESOURCE_ID, 0)`

shahzaibj (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
mohitc1: Addressed. Didn't realize default was 25 sec. I have rather just used waitUntilGone method by having a utility method now.

shahzaibj (approved):

PR #3159, opened 2026-06-17
Fix WebView stuck on server (5xx) error during login (AB#3408586), Fixes AB#3408586

fadidurah:
## Summary
Fixes **[AB#3408586](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3408586)** — *User should not be stuck on the login if server is unavailable.*

When the embedded WebView auth flow receives a server-side HTTP error (5xx) on the **main frame**, `onReceivedHttpError` previously only logged and tracked the error. It never dismissed the loading spinner or invoked an error callback, so `onPageFinished` (which hides the spinner via `onPageLoaded`) was never reached and the user was stuck on the login screen indefinitely.

This is asymmetric with network errors (`onReceivedError`), which already call `sendErrorToCallback`. HTTP 5xx now does the same.

## Root cause
- `WebViewAuthorizationFragment.launchWebView()` shows the progress bar.
- On a 5xx, the page never finishes loading, so `OAuth2WebViewClient.onPageFinished` -> `onPageLoaded()` (hides spinner) is never called.
- Old `onReceivedHttpError` only called `mUrlLoadTracker.updateLatestUrlStatus(...)`, leaving the spinner up forever.

## Fix
- `AzureActiveDirectoryWebViewClient.onReceivedHttpError`: for **main-frame** requests with `statusCode >= 500`, log a warning and call `sendErrorToCallback` (stops loading + sends a `ClientException` through the completion callback).
- Restricted to main-frame 5xx only — 4xx flows (e.g. 401/403 handled via redirects) and sub-resource errors are unaffected.
- `OAuth2WebViewClient.sendErrorToCallback`: changed `private` -> `protected` so the subclass can invoke it.
- Gated behind a new **default-on** flight `FailWebViewFlowOnServerHttpError` (`CommonFlight`) so it can be disabled via ECS if it causes regressions.

## Tests
Added 4 unit tests to `AzureActiveDirectoryWebViewClientTest`:
- main-frame 5xx -> callback invoked with `ClientException`
- main-frame 4xx -> no callback
- sub-resource 5xx -> no callback
- 5xx with flight disabled -> no callback

All tests pass (`:common:testLocalDebugUnitTest`, JDK 17).

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
+++ b/common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
@@ -95,6 +95,7 @@ import com.microsoft.identity.common.java.util.StringUtil;
 import com.microsoft.identity.common.logging.Logger;
 
 import java.net.MalformedURLException;
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1393,9 +1394,30 @@ public void onReceivedSslError(final WebView view,
...
+        final int statusCode = errorResponse != null ? errorResponse.getStatusCode() : 0;
+
         // Track HTTP error for the URL
-        if (mUrlLoadTracker != null && request.isForMainFrame()) {
-            mUrlLoadTracker.updateLatestUrlStatus("HTTP Error Code: " + errorResponse.getStatusCode(), null);
+        if (mUrlLoadTracker != null) {
+            mUrlLoadTracker.updateLatestUrlStatus("HTTP Error Code: " + statusCode, null);
+        }
+
+        // A server-side error (5xx) on the main frame means the page will never finish loading,
```
fadidurah: @copilot validate this, is this correct?

PR #3165, opened 2026-06-22
Use exact host matching for known-authority validation, Fixes AB#3607019

rpdome:
[AB#3607019](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3607019)

Harden `Authority.isKnownAuthority()` to compare developer-configured known authorities by exact, case-insensitive parsed host, consistent with `getEquivalentConfiguredAuthority()`. This replaces the previous substring comparison, which let an untrusted host pass the gate when it was a substring of a configured authority URL.

Adds unit tests in `AuthorityKnownHostTest`.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 vNext
 ----------
 - [PATCH] Add moshi-adapters dependency to Common module so downstream consumers like OneAuth do not have to add it explicitly (#3168)
+- [PATCH] Harden Authority.isKnownAuthority() to use exact, case-insensitive host matching for developer-configured known authorities; gated behind CommonFlight.ENABLE_KNOWN_AUTHORITY_HOST_EXACT_MATCH (default on)
 
 Version 24.4.0
 ----------
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 vNext
 ----------
 - [PATCH] Add moshi-adapters dependency to Common module so downstream consumers like OneAuth do not have to add it explicitly (#3168)
-- [PATCH] Harden Authority.isKnownAuthority() to use exact, case-insensitive host matching for developer-configured known authorities; gated behind CommonFlight.ENABLE_KNOWN_AUTHORITY_HOST_EXACT_MATCH (default on)
+- [PATCH] Harden Authority.isKnownAuthority() to use exact, case-insensitive host matching for developer-configured known authorities (#3165)
 
 Version 24.4.0
 ----------
--- a/common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
+++ b/common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
@@ -217,14 +217,9 @@ public abstract class Authority {
      * Returns the developer-configured known authority whose parsed host matches (case-insensitively)
      * the host of {@code authorityStr}, or {@code null} if none matches.
      * <p>
-     * Matching is performed by parsed host only; port and path are intentionally ignored. The trust
-     * boundary for a developer-configured known authority is the host (DNS/TLS trust is per-host), so
-     * {@code https://example.com}, {@code https://example.com:443} and {@code https://example.com:8443}
-     * are all treated as the same trusted host. This deliberately mirrors the comparison performed by
-     * the known-authority gate in {@link #isKnownAuthority(Authority)} (both delegate to
-     * {@link #matchesConfiguredHost}) so that the gate and authority resolution
-     * ({@link #getAuthorityFromAuthorityUrl(String, String)}) never disagree about whether a URL is
-     * developer-configured.
+     * Matching is by parsed host only; port and path are ignored (trust is per-host). This uses the
+     * same {@link #matchesConfiguredHost} comparison as the gate in {@link #isKnownAuthority(Authority)},
+     * so resolution and the gate never disagree about whether a URL is developer-configured.
      *
      * @param authorityStr the candidate authority URL.
      * @return the matching configured {@link Authority}, or {@code null} if none matches.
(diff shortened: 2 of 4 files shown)


rpdome:
@copilot add changelog pls.

rpdome:
@copilot also remove methodtag as a param of isKnownToDeveloperByExactHost. that method should have its own methodtag.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -384,6 +384,67 @@ public static boolean isKnownAuthority(Authority authority) {
...
+    private static boolean isKnownToDeveloperByExactHost(@NonNull final URL authorityUrl) {
+        final String methodTag = TAG + ":isKnownToDeveloperByExactHost";
+        final String candidateHost = authorityUrl.getAuthority();
+        synchronized (sLock) {
+            for (final Authority currentAuthority : knownAuthorities) {
+                if (currentAuthority.mAuthorityUrlString != null && candidateHost != null) {
+                    try {
+                        final URL knownAuthorityUrl = new URL(currentAuthority.mAuthorityUrlString);
+                        if (candidateHost.equalsIgnoreCase(knownAuthorityUrl.getAuthority())) {
+                            return true;
```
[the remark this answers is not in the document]
rpdome: updated the code to check host only.

common4j/src/test/com/microsoft/identity/common/java/authorities/AuthorityKnownHostTest.java
```diff
@@ -0,0 +1,294 @@
...
+        Assert.assertFalse("Substring host must not be treated as known", result.getKnown());
+        Assert.assertNotNull("A blocking exception must be produced for the controller to throw",
+                result.getClientException());
+        Assert.assertEquals(ClientException.UNKNOWN_AUTHORITY,
+                result.getClientException().getErrorCode());
+    }
+
+    /** Test 7 - Silent: a substring-host authority is denied by the gate before the refresh token is sent. */
+    @Test
+    public void silentGateRejectsSubstringHostBeforeRefreshTokenSubmission() throws IOException {
```
somalaya: Tests 6 & 7 are identical - same setup, same method called, same assertions

somalaya (approved):

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -358,17 +360,15 @@ public static boolean isKnownAuthority(Authority authority) {
...
-                    knownToDeveloper = true;
-                    break;
-                }
-            }
-        }
+        // Determine which known-authority matching strategy to use. The exact-host comparison is the
+        // secure default; the legacy substring comparison is retained only behind a kill-switch flight
+        // so it can be re-enabled via ECS if the stricter matching ever rejects a legitimate authority.
+        final boolean useExactHostMatch = CommonFlightsManager.INSTANCE.getFlightsProvider()
+                .isFlightEnabled(CommonFlight.ENABLE_KNOWN_AUTHORITY_HOST_EXACT_MATCH);
```
mohitc1: Developer provided known authority is only effective on non-broker flow - MSAL, remote flighting would not be effective.

mohitc1 (approved):
:shipit:

changelog.txt
```diff
@@ -1,6 +1,7 @@
 vNext
 ----------
 - [PATCH] Add moshi-adapters dependency to Common module so downstream consumers like OneAuth do not have to add it explicitly (#3168)
+- [PATCH] Harden Authority.isKnownAuthority() to use exact, case-insensitive host matching for developer-configured known authorities; gated behind CommonFlight.ENABLE_KNOWN_AUTHORITY_HOST_EXACT_MATCH (default on)
```
mohitc1: PR#

common4j/src/test/com/microsoft/identity/common/java/authorities/AuthorityKnownHostTest.java
rpdome: Good catch — they were identical (both reduced to the same `getKnownAuthorityResult()` chokepoint that the ROPC and silent flows consult before any credential POST). I merged them into a single `gateRejectsSubstringHostBeforeCredentialSubmission` and regrouped the suite by purpose (security regression guards vs. non-regression guards vs. full-gate guards).

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
rpdome: Good point — since the developer-configured known-authority path is MSAL-only and standalone MSAL has no ECS flighting (it falls back to the compile-time default), the remote kill-switch wasn't effective where it mattered. I removed the flight entirely; `isKnownAuthority()` now always uses exact, case-insensitive host matching (secure by default).

changelog.txt
rpdome: Added the PR reference — the changelog entry now ends with `(#3165)`.

common4j/src/main/com/microsoft/identity/common/java/authorities/Authority.java
```diff
@@ -384,6 +374,51 @@ public static boolean isKnownAuthority(Authority authority) {
...
+        final String candidateHost = authorityUrl.getHost();
+        if (StringUtil.isNullOrEmpty(candidateHost)) {
+            return false;
+        }
+        synchronized (sLock) {
+            for (final Authority currentAuthority : knownAuthorities) {
+                if (currentAuthority.mAuthorityUrlString != null) {
+                    try {
+                        final URL knownAuthorityUrl = new URL(currentAuthority.mAuthorityUrlString);
+                        if (candidateHost.equalsIgnoreCase(knownAuthorityUrl.getHost())) {
```
shahzaibj: javadoc says this matches the host-equality intent of `getEquivalentConfiguredAuthority`, but that one compares `getAuthority()` (host:port) while this compares `getHost()` only, so the two actually diverge on port. A candidate like `https://trusted.com:8443/...` now clears this gate (port ignored) but `authorityIsKnownFromConfiguration` still treats it as not-configured, so the gate says known while `getAuthorityFromAuthorityUrl` falls through to UnknownAuthority for the same url. Host being the trust boundary makes sense to me, just should `getEquivalentConfiguredAuthority` move to host-only too so the two paths stop disagreeing?

rpdome: addressed.

mohitc1 (approved):

PR #3166, opened 2026-06-22
Add wait on Get Access screen to handle duplicate APP flow trigger, Fixes AB#3667378

mohitc1:
Due to a known issue in Company Portal, the App Protection Policy flow can trigger twice. The second instance only appears while still on the Get Access screen. By waiting 10s before clicking Continue, we absorb the potential second instance so the PIN flow only runs once.
Fixes [AB#3673330](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3673330)

### DIFF
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -402,7 +402,8 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
         try {
             Thread.sleep(TimeUnit.SECONDS.toMillis(10));
         } catch (final InterruptedException e) {
-            // Ignored
+            Thread.currentThread().interrupt();
+            Logger.w(TAG, "Interrupted while waiting on Get Access screen", e);
         }
 
         // Click Continue on Get Access
--- a/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
+++ b/uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
@@ -387,36 +387,29 @@ public class BrokerCompanyPortal extends AbstractTestBroker implements ITestBrok
 
         final UiDevice device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation());
 
-        // get access screen
+        // Wait for "Get Access" screen
         final UiObject getAccessScreen = UiAutomatorUtils.obtainUiObjectWithText("Get Access");
         Assert.assertTrue(
                 "CP - Get Access screen appears",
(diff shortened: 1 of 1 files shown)


fadidurah (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -387,36 +387,30 @@ public void handleAppProtectionPolicy() {
...
+            Thread.currentThread().interrupt();
+            Logger.w(TAG, "Interrupted while waiting on Get Access screen", e);
+        }
+
+        // Click Continue on Get Access
         UiAutomatorUtils.handleButtonClick("com.microsoft.windowsintune.companyportal:id/positive_button");
 
-        // handle PIN entry and confirmation
+        // Handle PIN entry and confirmation
         enterAndConfirmPin(device);
```
shahzaibj, line 413: The block right after this used to handle the PIN screen showing up twice (the "PIN entry and confirmation screens can appear twice in succession" check). We're dropping that and leaning entirely on the 10s Get Access wait to absorb the duplicate before we click Continue. Is the duplicate guaranteed to only ever surface on Get Access, and is 10s always enough for it to merge? If the second instance can still land after Continue (or takes longer than 10s), we're back to the original flake with no fallback now.

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
```diff
@@ -387,36 +387,30 @@ public void handleAppProtectionPolicy() {
...
         );
 
-        // get access screen - continue
+        // Due to a known issue in Company Portal, the App Protection Policy flow can trigger
+        // twice. The second instance only appears while still on the "Get Access" screen.
+        // Once we click Continue, it won't reappear. By waiting here, we allow the potential
+        // second instance to arrive and merge, so the PIN flow only needs to run once.
+        Logger.i(TAG, "Waiting on Get Access screen to absorb potential second instance..");
+        try {
+            Thread.sleep(TimeUnit.SECONDS.toMillis(10));
```
shahzaibj, line 403: Hmm. We're not really waiting on Get Access screen here, are we? We seem to just wait 10 seconds regardless

somalaya (approved):

shahzaibj (approved):

uiautomationutilities/src/main/java/com/microsoft/identity/client/ui/automation/broker/BrokerCompanyPortal.java
mohitc1: I have manually executed test case several times. In my attempts, I found that, 2nd instance lands only once, it can happen at any point in PIN setting flow. By 10s, I am making the second instance come on "Get Access" screen itself, rather than in between Set PIN flows, where it gets hard to determine, what should be done next. This simplifies the expectation.

I'll iterate this is what I experience in my manual run of the test case several times.

PR #3167, opened 2026-06-22
Fix dormant KeyStore key-gen regression (explodes on API <= 30 from broker 16.1.0), Fixed AB#3673639

p3dr0rv:
## Summary

This fixes a **dormant key-generation defect** in the KeyStore-backed secret-key provider that only **detonated at scale starting with broker 16.1.0** (common 24.2.0). It is **not exclusive to 16.1.0** — the same defect already leaked in **16.0.1**, persists in **16.2.0 / 16.3.0**, and is not fully exclusive to API &lt;= 30 either.

On devices running **Android &lt;= 11 (API &lt;= 30)**, storage secret-key generation (`fn_SecretKeyGeneration`) fails at a **~58-67% device-level rate** across broker **16.1.0 / 16.2.0 / 16.3.0**. The dominant error is `All key generation attempts failed. Total failures: 3`.

## Root cause

When the provider generates a **fresh** RSA wrapping keypair on a **pre-Keystore-2.0 keymaster (API &lt;= 30)**, every spec it tries fails:

- The modern `KeyGenParameterSpec` specs request features these keymasters reject — `PURPOSE_WRAP_KEY` (API 28+), SHA-512 digests, OAEP/MGF1 padding — and throw `java.security.ProviderException: Failed to load generated key pair from keystore`.
- The provider **does** fall back through its prioritized spec list (`generateKeyPair()` iterates with a per-spec `runCatching`) ending in the deprecated `android.security.KeyPairGeneratorSpec`. But on these devices **that legacy fallback also fails** — so all three attempts throw, producing `Total failures: 3`.

Per-spec outcome on the `KeyPairGeneration` span (broker 16.1.0, API &lt;= 30, 7-day window):

| Spec that succeeded | Devices |
|---------------------|---------|
| `modern_spec_with_wrap_key`                 | 24,655 |
| `modern_spec_without_wrap_key`              | 3,865  |
| `legacy_key_gen_spec` (deprecated fallback) | 5,742  |
| **all specs failed (`Total failures: 3`)**  | **375,604** |

The deprecated legacy fallback rescues only **5,742** of the ~381k devices where the modern specs fail (~1.5%). So this is **not** a case of a missing fallback — the fallback is present and exercised; the problem is that **neither the modern specs nor the deprecated fallback succeed** on these keymasters. The fix adds a *new* conservative spec these keymasters do accept.

## Why it was dormant, and why it exploded at 16.1.0

The key-gen code is **identical** between 16.0.1 and 16.1.0 — only the *volume of fresh keypair generations* changed:

1. **Why it stayed hidden:** `generateNewSecretKey()` does `readKey(alias) ?: generateKeyPair()` — it **reuses the existing wrapping keypair** and only generates a fresh one when none exists. Most of the installed base already had a usable keypair, so the broken **fresh-generation** path was rarely exercised (only the oldest API 26-28 stragglers leaked).
2. **What lit the fuse (broker 16.1.0):** the `useKeystoreOnly = true` routing change (common 24.2.0) began sending **first-time-enrolling devices** (no existing keypair) through fresh KeyStore key generation — creating a **high-volume stream of fresh generations** on the vulnerable API &lt;= 30 population.

The cleanest proof is **Company Portal** (`com.microsoft.windowsintune.companyportal`), same host, same OS band, 14-day window:

| API | 16.0.1 devices | 16.0.1 fail % | 16.1.0 devices | 16.1.0 fail % |
|-----|----------------|---------------|----------------|---------------|
| 30  | 23,834         | **0.2%**      | 45,650         | **66.1%**     |
| 29  | 2,781          | **5.6%**      | 9,384          | **67.7%**     |
| 28  | 583            | **51.1%**     | 933            | 31.9%         |
| 27  | 156            | **26.9%**     | 910            | 30.9%         |
| 26  | 21             | **14.3%**     | 99             | 15.2%         |

- **The defect already leaked on 16.0.1.** The oldest bands (API 26-28) were already failing at 14-51% — the capability gap existed before 16.1.0, it just hit a small population.
- **The 16.1.0 routing change detonated the mainstream legacy bands.** API 29-30 — tens of thousands of devices — was **healthy on 16.0.1 (0.2-5.6%)** and jumped to **66-68%** on 16.1.0 on the *same OS band*. That is the routing change exposing a previously-unexercised path, not a new bug.

## Not exclusive to API &lt;= 30 either

The regression also **roughly doubled** failures on API &gt;= 31: e.g. **API 33 rose from ~4.7% to ~10-12%** on 16.1.0/16.2.0, driven by a long tail of non-conformant devices (CUBOT, DOOGEE, nubia, realme, OnePlus EB2101, ChromeOS ARC++ boards) that reject the advanced spec even on Keystore 2.0. The fix benefits these devices too.

On API &lt;= 30 the failure is **universal across mainstream OEMs** (Google Pixel — the AOSP reference device — fails at ~61.7%), confirming an Android-version capability gap rather than an OEM defect.

## Change

`CryptoParameterSpecFactory.getPrioritizedKeyGenParameterSpecs()` (flighted via `ENABLE_CONSERVATIVE_KEY_GEN_SPEC_FOR_LEGACY_DEVICES`, default on):

1. **Gate the advanced specs to API &gt;= 31** — `PURPOSE_WRAP_KEY` / SHA-512 / OAEP-MGF1 are only reliably supported from Keystore 2.0 (API 31) onward; on API &lt;= 30 they only ever throw.
2. **Add a conservative spec** for API 23-30: RSA / `ENCRYPT|DECRYPT` / SHA-256 / PKCS1. Narrower than the advanced specs and more reliably introspectable than the deprecated `KeyPairGeneratorSpec`, so a compatible cipher spec is always found when wrapping (this also resolves the secondary `No compatible cipher specs found` class).
3. **Keep the legacy spec** as the final fallback (unchanged from current behavior).

Resulting priority order:

| API level | Key-gen spec order |
|-----------|--------------------|
| &gt;= 31     | wrap-key -&gt; without-wrap-key (flight-gated) -&gt; conservative -&gt; legacy |
| 23 - 30   | conservative -&gt; legacy |
| &lt; 23      | legacy |

When the flight is turned **off** via ECS, the factory restores the prior behavior exactly (advanced specs at their original API gates, no conservative spec) — an instant, deploy-free kill-switch. Behavior on API &gt;= 31 (already healthy) is otherwise unchanged except for the added conservative fallback that also helps the non-conformant tail.

> **Not yet production-validated.** The conservative spec's effectiveness rests on the rationale above, not yet on field telemetry: the fix flight is not enabled anywhere, so there are currently **0** `conservative_spec_for_api_30_and_below` successes recorded. Post-rollout telemetry (below) will confirm it.

## Why not revert the routing?

The `useKeystoreOnly` routing introduced in 24.2.0 is intentional (decoupling device registration from the MSAL/ADAL predefined-key lifecycle for OneAuth), and a device-registration component may legitimately have no predefined key. The real requirement is that KeyStore key generation must work on API &lt;= 30 — this PR delivers that rather than masking it.

## Validation

Full telemetry investigation, scoping queries, per-broker-version / per-host monitoring queries, and the post-deployment validation query are documented in `ICM-investigation/keygen-api30-regression-findings.md`. After rollout, the `API &lt;= 30` cells for 16.1.0+ should drop from ~58-67% toward the 16.0.1 baseline (Company Portal API 30 was 0.2%), the API &gt;= 31 tail (e.g. API 33) should fall back toward ~4-5%, and a non-zero `conservative_spec_for_api_30_and_below` success population should appear on API &lt;= 30.

image

[AB#3673639](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3673639)

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
@@ -243,9 +243,9 @@ class CryptoParameterSpecFactory(
     /**
      * Original key generation spec selection (fix flight OFF) - unchanged behaviour.
      *
-     * The advanced specs (PURPOSE_WRAP_KEY / SHA-512 / OAEP-MGF1) are attempted on every API level.
-     * This is what causes pre-Keystore 2.0 (API &lt;= 30) keymasters to reject every spec; see
-     * [getLegacyDeviceSafeKeyGenSpecs] for the fixed path.
+     * The advanced specs (PURPOSE_WRAP_KEY / SHA-512 / OAEP-MGF1) are attempted only on API levels
+     * that support them (e.g., wrap-key spec on API 28+; non-wrap spec on API 23+) when their
+     * corresponding feature flags are enabled. See [getLegacyDeviceSafeKeyGenSpecs] for the fixed path.
      */
     private fun getDefaultKeyGenSpecs(): List<IKeyGenSpec> {
         val specs = mutableListOf<IKeyGenSpec>()
@@ -292,9 +292,7 @@ class CryptoParameterSpecFactory(
         // Primary viable spec on API 23..30 keymasters; a safe fallback on API 31+.
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             specs.add(keyGenParamSpecConservative)
-        }
-
-        // Always include legacy spec as last resort fallback (API < 23).
+        // Always include legacy spec as last resort fallback.
         specs.add(keyGenParamSpecLegacy)
         return specs
     }
--- a/common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
@@ -218,71 +218,84 @@ class CryptoParameterSpecFactory(
      */
     fun getPrioritizedKeyGenParameterSpecs(): List<IKeyGenSpec> {
         val methodTag = "$TAG:getPrioritizedKeyGenParameterSpecs"
-        val specs = mutableListOf<IKeyGenSpec>()
         val conservativeFixEnabled = conservativeKeyGenForLegacyDevices
 
-        // The advanced specs below request features (PURPOSE_WRAP_KEY, SHA-512 digests and
-        // OAEP/MGF1 padding) that are only reliably supported by the hardware keymaster starting
-        // with Keystore 2.0 (API 31). On API <= 30 many hardware-backed keymasters - especially on
-        // rugged/enterprise devices - reject these specs, which previously caused every attempt
-        // (including the deprecated legacy KeyPairGeneratorSpec fallback) to fail with
-        // "All key generation attempts failed". When the legacy-device fix flight is enabled the
-        // advanced specs are restricted to API 31+; otherwise the previous behaviour is preserved.
-        val advancedSpecsAllowed =
-            !conservativeFixEnabled || Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
+        // Record whether the fix flight was on for this attempt so we can correlate, post-deploy,
+        // against the existing DeviceInfo_OsVersion and key_pair_gen_description telemetry.
+        SpanExtension.current().setAttribute(
+            AttributeName.key_gen_conservative_spec_flight_enabled.name,
+            conservativeFixEnabled
+        )
 
-        if (advancedSpecsAllowed) {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && keySpecWithWrapPurposeKey) {
-                // First priority: API 28+ with PURPOSE_WRAP_KEY if enabled
-                specs.add(keyGenParamSpecWithPurposeWrapKey)
-            }
-
-            if (keySpecWithoutWrapPurposeKey) {
-                // Second priority: API 23+ without PURPOSE_WRAP_KEY
-                specs.add(keyGenParamSpecWithoutPurposeWrapKey)
-            }
+        // The fix is an ECS kill switch. When the flight is OFF we run the exact original spec
+        // selection; when it is ON we run the legacy-device-safe variant. Keeping the two paths
+        // separate makes the change trivial to reason about and to revert via ECS.
+        val specs = if (conservativeFixEnabled) {
+            getLegacyDeviceSafeKeyGenSpecs()
+        } else {
+            getDefaultKeyGenSpecs()
         }
 
-        // Conservative, hardware-friendly spec (RSA, ENCRYPT|DECRYPT, SHA-256, PKCS1 only).
-        // Only added when the legacy-device fix is enabled; it is the primary viable spec on
-        // API 23..30 keymasters and a safe fallback on API 31+ before the deprecated legacy spec.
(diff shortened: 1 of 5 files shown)


common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -626,6 +626,12 @@ public enum AttributeName {
      */
     key_pair_gen_elapsed_time,
 
+    /**
+     * Indicates whether the conservative key generation spec for legacy devices (API &lt;= 30)
+     * flight is enabled. Used to validate the rollout/effectiveness of the legacy-device fix.
+     */
+    key_gen_conservative_spec_flight_enabled,
```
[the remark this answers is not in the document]
p3dr0rv: minor no -fix

common/src/test/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactoryTest.kt
```diff
@@ -208,8 +209,166 @@ class CryptoParameterSpecFactoryTest {
...
+        Assert.assertEquals("legacy_key_gen_spec", specs[1].description)
+        Assert.assertFalse(
+            "API 30 must not attempt the advanced wrap-key spec when the fix is on",
+            specs.any { it.description == "modern_spec_with_wrap_key" }
+        )
+        Assert.assertFalse(
+            specs.any { it.description == "modern_spec_without_wrap_key" }
+        )
+
+        // Reset to the class-level Robolectric SDK (API 28) to avoid leaking into other tests.
```
somalaya: You can move this to @After method to reliably reset instead of doing it in every test

somalaya (approved):

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
```diff
@@ -192,10 +260,39 @@ class CryptoParameterSpecFactory(
...
+
+        // Advanced specs are only reliable from Keystore 2.0 (API 31) onward.
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            if (keySpecWithWrapPurposeKey) {
+                specs.add(keyGenParamSpecWithPurposeWrapKey)
+            }
+            if (keySpecWithoutWrapPurposeKey) {
+                specs.add(keyGenParamSpecWithoutPurposeWrapKey)
+            }
+        }
```
mohitc1, line 291: Why we don't even attempt modern specs for >=P and <S ? is it 100% fail for all devices?

I was thinking if our logic should same as getDefaultKeyGenSpecs() with conservative spec inserted before deprecated path

common/src/test/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactoryTest.kt
p3dr0rv: done

common/src/main/java/com/microsoft/identity/common/crypto/CryptoParameterSpecFactory.kt
p3dr0rv: yeah, make sense, the failure rate is not 100% is around 50% so is better that at least those devices get a more modern encryption.
DONE

mohitc1 (approved):
:shipit:

p3dr0rv:
@copilot add changelog entry keep it concise and do not forget the PR ID

somalaya (approved):

PR #3168, opened 2026-06-22
Add moshi adapter dependency, Fixes AB#3673534

mohitc1:
Add moshi-adapters dependency to Common module so downstream consumers like OneAuth do not have to add it explicitly. 

Moshi is used device registration API IPC serialization. OneAuth is soon going call device registration APIs.

Fixes [AB#3673534](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3673534)

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -233,6 +233,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$rootProject.ext.appCompatVersion"
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "com.squareup.moshi:moshi:$rootProject.ext.moshiVersion"
+    implementation "com.squareup.moshi:moshi-adapters:$rootProject.ext.moshiAdaptersVersion"
     implementation "androidx.browser:browser:$rootProject.ext.browserVersion"
     implementation "androidx.core:core:$rootProject.ext.androidxCoreVersion"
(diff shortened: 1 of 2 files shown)


mohitc1:
@copilot fix the PR number in changelog

common/build.gradle
```diff
@@ -233,6 +233,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$rootProject.ext.appCompatVersion"
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     implementation "com.squareup.moshi:moshi:$rootProject.ext.moshiVersion"
+    implementation "com.squareup.moshi:moshi-adapters:$rootProject.ext.moshiAdaptersVersion"
```
p3dr0rv, line 236: is the version already defined here?
do we need to update LTW?

p3dr0rv (approved):

common/build.gradle
mohitc1: Due to broker, it's already there in LTW and other brokers. 
For context, while integrating OneAuth I saw NoClassDef found error, due to adapters not present. common4j as dependency but common removes transitive dependencies. So, either OneAuth adds it explicitly or we can add in Common. I believe doing latter is right approach, as right module is bringing the dependency.

rpdome (approved):

PR #3170, opened 2026-06-23
Add flighted validation in WebView intent handling, Fixes AB#3674205

cacosta33:
## Summary

Adds a flight-gated validation step to the AAD WebView client's broker-install `intent://` handling path. When the flight is enabled, the parsed intent's component/selector are cleared and its target package must be the Google Play Store before the activity is launched. When the flight is disabled, behavior is **byte-for-byte identical to today** (kill-switch / rollback).

Work item: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3674205

## Feature flag & rollout

- Gated behind a new ECS flight `EnableBrokerInstallIntentValidation` (`CommonFlight`), **default OFF**.
- Flight OFF means the legacy launch path runs unchanged, so a field regression is a config flip, not a redeploy.
- `common` ships to broker + MSAL + downstream consumers, so this is intended to ramp **progressively** (e.g. 1% then 10% then 100%) while watching install-flow success telemetry.

## Changes

- `CommonFlight.java` — new `ENABLE_BROKER_INSTALL_INTENT_VALIDATION` flight (default `false`).
- `AzureActiveDirectoryWebViewClient.java` — inside the flighted branch only: clear any explicit component/selector on the parsed intent and require the resolved package to be on the allow-list (`com.android.vending`) before `startActivity`.
- Tests — added coverage for: non-allow-listed package is not launched; an explicit component is cleared (Play Store still launches); legacy behavior preserved when the flight is off; and a fixed pre-existing test that was missing `@Test` (so it never ran).

## Validation

`:common:testLocalDebugUnitTest` for the affected client passed (BUILD SUCCESSFUL), all tests pass (4 intent-path tests confirmed executing).

## Do not close the linked work item until confirmed

A couple of points cannot be verified from this library's code alone and may affect priority — please confirm before resolving the linked item (the **code hardening here is safe to merge/ramp regardless**):

- Whether externally-influenced content can reach this navigation path inside the hosting auth WebView.
- Real-device behavior of the previously-broader launch path.

[AB#3674205](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3674205)

## Notes for reviewers

- No behavioral change for any consumer while the flight is OFF (default).
- Does not touch `OneAuthSharedFunctions` or any 1P IPC surface.


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1160,28 +1167,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+        // flighted logic and telemetry is contained here, so the flight-off path above is unaffected.
+        // Falls back to a no-op span if OpenTelemetry is not initialized.
+        final Span span = createSpanWithAttributesFromParent(SpanName.ProcessBrokerInstallIntent.name());
+        try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
             final Intent intent = Intent.parseUri(intentUrl, Intent.URI_INTENT_SCHEME);
             if (intent != null && intent.getPackage() != null) {
+                // Clear any explicit component or selector carried by the parsed intent so that
+                // activity resolution is driven solely by the validated package.
+                intent.setComponent(null);
+                intent.setSelector(null);
```
shahzaibj: we clear the selector here too, but I only see tests for the explicit-component case and the non-allowlisted package. nothing exercises a selector. that's kinda the trickier redirect vector - base package = com.android.vending so it passes the getPackage() check, but a `SEL` selector pointing at another app is what startActivity would actually resolve. can we add a case with a selector to a non-Play-Store target and assert it still doesn't launch?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+            // the else branch below is the original (flight-off) behavior, byte-for-byte unchanged.
+            // Falls back to a no-op span if OpenTelemetry is not initialized.
+            final Span span = createSpanWithAttributesFromParent(SpanName.ProcessBrokerInstallIntent.name());
+            try (final Scope scope = SpanExtension.makeCurrentSpan(span)) {
+                final Intent intent = Intent.parseUri(intentUrl, Intent.URI_INTENT_SCHEME);
+                if (intent != null && intent.getPackage() != null) {
+                    // Clear any explicit component or selector carried by the parsed intent so that
+                    // activity resolution is driven solely by the validated package.
+                    intent.setComponent(null);
+                    intent.setSelector(null);
```
shahzaibj: we clear the selector here too, but I only see tests for the explicit-component case and the non-allowlisted package. nothing exercises a selector. that's kinda the trickier redirect vector - the routing gate and the new check both look at the base package, so setting base package = com.android.vending passes, and a `SEL` selector pointing at another app is what startActivity would actually resolve. can we add a case with a selector to a non-Play-Store target and assert it still doesn't launch?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+
+                    final String targetPackage = intent.getPackage();
+                    if (!isAllowedBrokerInstallIntentTarget(targetPackage)) {
+                        Logger.warn(methodTag,
+                                "Blocking intent request to non-allow-listed package: " + targetPackage);
+                        span.setAttribute(AttributeName.is_broker_install_intent_blocked.name(), true);
+                        span.setStatus(StatusCode.OK);
```
shahzaibj: on a block we set the span OK and just return, so handleUrl falls through to true and the webview keeps going with nothing launched and no completion callback. processOpenIdVcRequest does the opposite for its "no handler" case, StatusCode.ERROR + returnError. I'm guessing the silent drop here is deliberate (firing returnError would let an injected intent kill the auth session), just making sure the OK + continue is intentional and not defaulting to the happy path.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -140,6 +141,12 @@
 public class AzureActiveDirectoryWebViewClient extends OAuth2WebViewClient {
     private static final String TAG = AzureActiveDirectoryWebViewClient.class.getSimpleName();
 
+    /**
+     * Package name of the Google Play Store, the legitimate launch target for a broker-install
+     * {@code intent://} request.
+     */
+    private static final String GOOGLE_PLAY_STORE_PACKAGE_NAME = "com.android.vending";
```
shahzaibj, line 148: isIntentRequestToInstallBrokerApp still inlines this same value as `;package=com.android.vending;` for its gate check. Could we build that substring off this constant so the gate and the allow-list can't drift apart later?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+                    final String targetPackage = intent.getPackage();
+                    if (!isAllowedBrokerInstallIntentTarget(targetPackage)) {
+                        Logger.warn(methodTag,
+                                "Blocking intent request to non-allow-listed package: " + targetPackage);
+                        span.setAttribute(AttributeName.is_broker_install_intent_blocked.name(), true);
+                        span.setStatus(StatusCode.OK);
+                        return;
+                    }
+
+                    view.getContext().startActivity(intent);
```
shahzaibj: we null the component and selector, but the rest of what the attacker put in the #Intent fragment still rides into startActivity - launchFlags, data uri, categories, extras. with the package pinned to com.android.vending it's not really a leak (a surviving FLAG_GRANT_READ_URI_PERMISSION would grant the store, not the page). the standard webview intent handling also masks off the FLAG_GRANT_* bits and adds CATEGORY_BROWSABLE before launching though. is pinning the package enough on its own here?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+                final Intent intent = Intent.parseUri(intentUrl, Intent.URI_INTENT_SCHEME);
+                if (intent != null && intent.getPackage() != null) {
+                    // Clear any explicit component or selector carried by the parsed intent so that
+                    // activity resolution is driven solely by the validated package.
+                    intent.setComponent(null);
+                    intent.setSelector(null);
+
+                    final String targetPackage = intent.getPackage();
+                    if (!isAllowedBrokerInstallIntentTarget(targetPackage)) {
+                        Logger.warn(methodTag,
```
melissaahn: This is a suggestion but non-blocking: in the case of us blocking an install intent for a non-allowlisted package, do we maybe want to throw an exception here? Or is it because we explicitly want to silently handle this type of error? Will users just see a blank page if that happens?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+                        span.setStatus(StatusCode.OK);
+                        return;
+                    }
+
+                    view.getContext().startActivity(intent);
+                    Logger.info(methodTag, "Intent request sent to launch the app: " + intent.getPackage());
+                    span.setAttribute(AttributeName.is_broker_install_intent_blocked.name(), false);
+                    span.setStatus(StatusCode.OK);
+                } else {
+                    Logger.warn(methodTag, "Unable to parse the intent URI");
```
melissaahn: Similar comment to the one above; I know this is just following the existing logic, but what happens when the intent doesn't actually get started here? Should we throw an exception, or was the thinking that it would be a worse experience if we just close the webview at that point?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
melissaahn: is there any needs to clear extras and flags, or is that certain to be empty?

changelog.txt
```diff
@@ -1,5 +1,6 @@
 vNext
 ----------
+- [PATCH] Add a gate-only OpenTelemetry span (ProcessBrokerInstallIntent) around the flighted broker-install intent handling path so the validation outcome (launched / blocked / error) can be confirmed in telemetry when the flight is enabled (#3170)
```
melissaahn: nit: I think this changlog should be updated to be more about the intent validation vs all about telemetry

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+    /**
+     * Checks whether the parsed broker-install intent targets the allow-listed package. The only
+     * supported launch target for this path is the Google Play Store, which opens the broker app's
+     * store listing, so any other package is not launched.
+     *
+     * @param packageName The target package declared by the parsed intent.
+     * @return {@code true} if the package is allow-listed, {@code false} otherwise.
+     */
+    private boolean isAllowedBrokerInstallIntentTarget(@Nullable final String packageName) {
+        return GOOGLE_PLAY_STORE_PACKAGE_NAME.equals(packageName);
```
melissaahn, line 1391: maybe this is out of scope, but what if we're dealing with devices which don't support gms?

melissaahn (approved):

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -524,8 +532,73 @@ public void testProcessCloudRedirectAndPrtHeaderInternalException() {
...
+    }
+
+    private void setBrokerInstallIntentValidationFlight(final boolean enabled) {
+        final IFlightsProvider mockFlightsProvider = Mockito.mock(IFlightsProvider.class);
+        when(mockFlightsProvider.isFlightEnabled(CommonFlight.ENABLE_BROKER_INSTALL_INTENT_VALIDATION))
+                .thenReturn(enabled);
+        final MockCommonFlightsManager mockCommonFlightsManager = new MockCommonFlightsManager();
+        mockCommonFlightsManager.setMockCommonFlightsProvider(mockFlightsProvider);
+        CommonFlightsManager.INSTANCE.initializeCommonFlightsManager(mockCommonFlightsManager);
     }
```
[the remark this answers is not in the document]
cacosta33: Done in 4bb6af3. The helper now installs a Mockito default answer returning each flight's real `getDefaultValue()` for any unstubbed flight, and overrides only `ENABLE_BROKER_INSTALL_INTENT_VALIDATION`. Tests no longer rely on Mockito's `false` boolean default, since the WebViewClient now consults additional flights. Validated locally with `:common:testLocalDebugUnitTest` (BUILD SUCCESSFUL).

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: By design - this mirrors the legacy flight-off path, which also silently no-ops (logs a warning) when the install intent can't proceed; the WebView stays on the current page. The blocked outcome is recorded via the `is_broker_install_intent_blocked=true` span attribute. Throwing here would be a behavior change relative to the rollback-match path we're intentionally preserving.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: By design - the null-intent / no-package branch is byte-for-byte the original behavior, kept so that flight-off vs flight-on differ only by the added target validation and telemetry, nothing else.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: By design - after validation the target is constrained to the trusted Play Store package (component and selector are already cleared). Clearing flags risks stripping `FLAG_ACTIVITY_NEW_TASK`, and the remaining extras are low-risk, so we intentionally leave them intact.

changelog.txt
cacosta33: Done in 4bb6af3 - reworded the changelog entry to describe the allow-list validation of broker-install intent targets rather than only the OTel span.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Out of scope for this PR and already handled: a device without GMS throws `ActivityNotFoundException`, which is caught and mapped to `returnError(ACTIVITY_NOT_FOUND, ...)`. This change only narrows which package may be launched; it doesn't alter the missing-Play-Store path.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1267,96 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
-            if (intent != null && intent.getPackage() != null) {
-                view.getContext().startActivity(intent);
-                Logger.info(methodTag, "Intent request sent to launch the app: " + intent.getPackage());
-            } else {
-                Logger.warn(methodTag, "Unable to parse the intent URI");
+
+        if (CommonFlightsManager.INSTANCE.getFlightsProvider()
+                .isFlightEnabled(ENABLE_BROKER_INSTALL_INTENT_VALIDATION)) {
+            // Flight ON (new behavior): validate the intent target and record the outcome to a
+            // telemetry span so the fix's behavior (launched / blocked / error) can be confirmed from
```
shahzaibj: Do we really need a dedicated span for something that runs within the webview and a small part of the overall acquire token interactive flows?

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -703,6 +703,13 @@ public enum AttributeName {
      * Indicates if an external handler was found to handle the openid-vc:// URI.
      */
     is_openid_vc_handler_found,
+
+    /**
+     * Indicates whether a broker-install {@code intent://} request was blocked because its target
+     * package was not the allow-listed store. Set on the {@link SpanName#ProcessBrokerInstallIntent}
+     * span (emitted only when the validation flight is enabled).
+     */
+    is_broker_install_intent_blocked,
```
shahzaibj, line 719: Looks like you are going to have just 1 attribute in the new span you are creating. This defeats the whole purpose of a span. Why not emit the attribute on the existing span?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Done — the gate now builds the match string from the shared constant (`";package=" + GOOGLE_PLAY_STORE_PACKAGE_NAME + ";"`) so this check and the allow-list validation can't drift apart. (commit 33e238005)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Added direct unit tests for the sanitizer. Note a selector can't be injected through the `intent://` URL scheme (Android doesn't (de)serialize a selector via `parseUri`/`toUri`), and on Android a top-level package and a selector are mutually exclusive — so a selector-bearing intent always has a null top-level package. `testSanitizeAndValidateBrokerInstallIntent_clearsSelectorAndBlocks` verifies the selector is nulled on the intent and the request is then blocked; the explicit-component path stays covered by the existing component test. (commit 33e238005)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Reworked as discussed: removed the dedicated `ProcessBrokerInstallIntent` child span and now emit `is_broker_install_intent_blocked` directly on the current WebView-processing span via `SpanExtension.current()` (matching the existing single-attribute pattern in this class). Dropped the now-unused `SpanName.ProcessBrokerInstallIntent` entry and updated the `AttributeName` Javadoc. (commit 33e238005)

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
cacosta33: See above — the single boolean is now set on the current span rather than a dedicated child span, so there's no longer a span created solely to carry one attribute. (commit 33e238005)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Hardened the sanitizer: it now strips `FLAG_GRANT_READ/WRITE/PERSISTABLE/PREFIX_URI_PERMISSION` and adds `CATEGORY_BROWSABLE`, mirroring the platform's standard WebView intent handling so a parsed intent can't carry an unexpected URI grant into the store app. Covered by `testSanitizeAndValidateBrokerInstallIntent_stripsGrantFlagsAndAddsBrowsable`. (commit 33e238005)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: This is intentional. On a blocked intent we set the span attribute and return without firing `returnError`; `handleUrl` then falls through so the WebView keeps its current page and no activity is launched. Firing `returnError` here would let an injected/malicious intent tear down the auth session (a DoS lever), so we deliberately drop it silently rather than surfacing an error to the caller.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -703,6 +703,16 @@ public enum AttributeName {
...
      */
     is_openid_vc_handler_found,
+
+    /**
+     * Indicates whether a broker-install {@code intent://} request was blocked because its target
+     * package was not the allow-listed store. Set on the current WebView-processing span (emitted
+     * only when the broker-install intent validation flight is enabled).
+     * <p>
+     * Note: this attribute must also be kept in sync with the {@code AttributeName} enum in the
+     * broker repository (AzureAD/ad-accounts-for-android) for cross-repo telemetry consistency.
```
shahzaibj: isn't the broker repo `identity-authnz-teams/ad-accounts-for-android`, not `AzureAD/...`? and this note is load-bearing, not just bookkeeping - broker's exporter drops any attribute name that doesn't resolve in its own `AttributeName` enum, so if this one ever rides a span in the broker process it silently won't export until it's mirrored there too.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1268,119 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+            } catch (final URISyntaxException e) {
+                Logger.error(methodTag, "Failed to parse the intent URI due to invalid syntax.", e);
+                returnError(ErrorStrings.URI_SYNTAX_ERROR, e.getMessage());
+            } catch (final ActivityNotFoundException e) {
+                Logger.error(methodTag, "No activity found to handle the intent.", e);
+                returnError(ErrorStrings.ACTIVITY_NOT_FOUND, e.getMessage());
+            } catch (final Throwable throwable) {
+                Logger.error(methodTag, "An unexpected error occurred while processing the intent URI.", throwable);
+                returnError(ErrorStrings.UNEXPECTED_ERROR, throwable.getMessage());
+            }
```
shahzaibj: Maybe can move to its own method

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1260,28 +1268,119 @@ private void processInvalidRedirectUri(@NonNull final WebView view,
...
+                returnError(ErrorStrings.URI_SYNTAX_ERROR, e.getMessage());
+            } catch (final ActivityNotFoundException e) {
+                Logger.error(methodTag, "No activity found to handle the intent.", e);
+                returnError(ErrorStrings.ACTIVITY_NOT_FOUND, e.getMessage());
+            } catch (final Throwable throwable) {
+                Logger.error(methodTag, "An unexpected error occurred while processing the intent URI.", throwable);
+                returnError(ErrorStrings.UNEXPECTED_ERROR, throwable.getMessage());
             }
-        } catch (final URISyntaxException e) {
-            Logger.error(methodTag, "Failed to parse the intent URI due to invalid syntax.", e);
```
shahzaibj, line 1274: Maybe can move to its own method

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
cacosta33: Good catch — fixed in `3c6404dde`. Corrected the repo to `identity-authnz-teams/ad-accounts-for-android` and reworded the note so it's clear the mirror is *functionally required* (the broker exporter drops any attribute name that doesn't resolve in broker's own `AttributeName` enum), not just for consistency.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Done — the new validated flow now lives in its own method, `launchValidatedBrokerInstallIntent(...)` (1c053bcc). The flight-OFF path stays byte-for-byte unchanged from dev.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
cacosta33: Done in 1c053bcc — `processIntentToInstallBrokerApp` now gates on the flight and delegates the new behavior to `launchValidatedBrokerInstallIntent(...)`, leaving the original flight-OFF body untouched.

melissaahn (approved):

PR #3171, opened 2026-06-24
Display request origin in HTTP auth dialog (flighted), Fixes AB#3674206

cacosta33:
## Summary

Adds an optional, flight-gated row to the HTTP authentication dialog that shows the requesting **host** and **realm**, so a user can see which origin is asking for credentials. Defaults off; when the flight is disabled the dialog is byte-for-byte unchanged.

Work item: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3674206

## Feature flag & rollout
- Gated behind a new ECS flight `EnableHttpAuthOriginDisplay` (`CommonFlight`), **default OFF**.
- Flight OFF means the dialog renders exactly as today (the origin row stays `GONE`), so this can ramp progressively via ECS.
- `common` is consumed by broker + MSAL, hence default-OFF + progressive ramp.

## Changes
- `NtlmChallengeHandler.java` - read the challenge host/realm and populate the origin row when the flight is enabled.
- `http_auth_dialog.xml` - add a `gone`-by-default `TextView` for the origin row.
- `strings.xml` - add host/realm label strings.
- `CommonFlight.java` - new `ENABLE_HTTP_AUTH_ORIGIN_DISPLAY` flight (default `false`).
- Tests - origin-text formatting (host + realm) and the flight gate (on/off).

## Validation
`:common:testLocalDebugUnitTest` for the handler - **BUILD SUCCESSFUL**, 3/3 tests pass.

[AB#3674206](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3674206)

## Notes for reviewers
- No behavioral change while the flight is OFF (default).
- Does not touch `OneAuthSharedFunctions` or any 1P IPC surface.


common/src/main/res/values/strings.xml
```diff
@@ -6,6 +6,8 @@
     <string name="http_auth_dialog_username">Username</string>
     <string name="http_auth_dialog_password">Password</string>
     <string name="http_auth_dialog_title">Enter your credentials</string>
+    <string name="http_auth_dialog_origin_host">Host: %1$s</string>
```
somalaya, line 9: Do we have to localize these?

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NtlmChallengeHandler.java
```diff
@@ -99,6 +124,92 @@ public void onCancel(DialogInterface dialog) {
...
+            originText.append(mActivity.getString(R.string.http_auth_dialog_origin_realm, realm));
+        }
+
+        return originText.toString();
+    }
+
+    /**
+     * Sanitizes a server-supplied origin value (host or realm) before it is rendered in the dialog.
+     * <p>
+     * The realm is taken verbatim from the {@code WWW-Authenticate} response header and is therefore
```
somalaya: The realm is also fully server-controlled, so a malicious origin could pair a suspicious host with a reassuring realm line (e.g. "Microsoft Corporate Login…") sitting right next to it. Might be worth confirming with the security owner whether realm display is intended, or whether host-only is safer.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NtlmChallengeHandler.java
```diff
@@ -99,6 +124,92 @@ public void onCancel(DialogInterface dialog) {
...
+                .replaceAll("[\\p{Cc}\\p{Zl}\\p{Zp}]", " ")
+                .replaceAll("\\s+", " ")
+                .trim();
+
+        if (sanitized.length() > MAX_ORIGIN_VALUE_LENGTH) {
+            sanitized = sanitized.substring(0, MAX_ORIGIN_VALUE_LENGTH);
+        }
+
+        return sanitized;
+    }
```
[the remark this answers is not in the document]
cacosta33: Addressed in bbc699f8a.

`sanitizeOriginValue` now strips Unicode **format** characters (`\p{Cf}`, incl. U+202E RIGHT-TO-LEFT OVERRIDE and the zero-width marks) alongside the existing `\p{Cc}\p{Zl}\p{Zp}`, closing the bidi/homograph spoofing gap. It also **pre-caps** the untrusted input to `PRE_SANITIZE_CAP` (4x the display cap) *before* the regex passes, so a pathologically large header value can't cause unnecessary UI-thread work; the final display cap is applied after. Added `testSanitizeOriginValue_stripsUnicodeFormatChars`.

Note: with the host-only change (see the realm thread) the sanitized value is now just the host, but the hardening still applies since a WebView can be navigated to a malicious origin.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NtlmChallengeHandlerTest.java
```diff
@@ -0,0 +1,201 @@
...
+        assertFalse(sanitized.contains("\r"));
+    }
+
+    @Test
+    public void testSanitizeOriginValue_stripsUnicodeLineSeparators() {
+        // U+2028 LINE SEPARATOR / U+2029 PARAGRAPH SEPARATOR are not matched by \s but must be removed.
+        final String sanitized = NtlmChallengeHandler.sanitizeOriginValue("a\u2028b\u2029c");
+
+        assertEquals("a b c", sanitized);
+    }
```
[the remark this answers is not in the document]
cacosta33: Done in bbc699f8a - added `testSanitizeOriginValue_stripsUnicodeFormatChars` covering U+202E RIGHT-TO-LEFT OVERRIDE and U+200B ZERO WIDTH SPACE, asserting both are neutralized. Full suite passes 12/12.

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NtlmChallengeHandlerTest.java
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Builds a mock {@link Activity} whose origin-string lookups echo their argument, so origin-text
+     * formatting can be verified without resolving Android string resources.
+     */
+    private Activity stubActivity() {
+        final Activity activity = Mockito.mock(Activity.class);
+        when(activity.getString(eq(R.string.http_auth_dialog_origin_host), any()))
+                .thenAnswer(invocation -> "Host: " + invocation.getArgument(1));
+        when(activity.getString(eq(R.string.http_auth_dialog_origin_realm), any()))
+                .thenAnswer(invocation -> "Realm: " + invocation.getArgument(1));
```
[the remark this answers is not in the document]
cacosta33: Verified this isn't reproducible on the repo's Mockito (5.11.0). Mockito 5.x flattens varargs, so `invocation.getArgument(1)` returns the first vararg element (the host string), not the `Object[]`. A minimal repro of the exact stub pattern printed `RESULT=[Host: login.example.com]` (no array leak), and the suite passes 12/12 including the host-formatting assertion `assertEquals("Host: " + TEST_HOST, ...)` which would fail outright if an array leaked. Resolving as not-an-issue. (The realm stub was also removed as part of the host-only change.)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/challengehandlers/NtlmChallengeHandler.java
cacosta33: Good call - went **host-only for v1** in bbc699f8a. `getOriginText` no longer reads or renders the realm (removed the now-unused `http_auth_dialog_origin_realm` string too), so there's no attacker-controlled free-text line sitting next to the host. Added a Javadoc note that realm display is deferred pending security review. The host is the higher-signal, harder-to-spoof field, so v1 shows just that.

common/src/main/res/values/strings.xml
cacosta33: Yes — localized in f3a642811. Added `http_auth_dialog_origin_host` to all 44 locale `strings.xml` files, matching the sibling `http_auth_dialog_*` strings which are already translated in-repo.

Translations use the **network/server** "host" sense from the official **Microsoft Terminology Collection** (not the "event host"/"organizer" sense) — e.g. de `Host`, fr `Hôte`, ru `Узел`, zh `主机`, ja `ホスト`, ar `مضيف`. A few locales use Microsoft's canonical "host computer" compound where that's the approved term (e.g. sv `Värddator`, gl `Computador servidor`, hr `Glavno računalo`). `:common:mergeLocalDebugResources` passes. Happy to tweak any specific locale if loc has a preferred term.

somalaya (approved):

melissaahn (approved):

melissaahn (approved):

melissaahn (approved):

PR #3172, opened 2026-06-25
Refine copilot-instructions.md to suppress recurring false-positive review patterns, Fixes AB#3660483

shahzaibj:
### What

Adds a new **§14 Recurring False-Positive Patterns — Do NOT Flag These** section to `.github/copilot-instructions.md` (placed at the end of the rules, before the appendices), plus reciprocal cross-references in the Testing section (§7.1 and §7.2).

### Why

The Copilot Code Review Effectiveness analysis (Run 3, Apr 20 – Jun 9, 2026) flagged 26 comments as **Confirmed Not Helpful** — engineers explicitly dismissed them with a stated reason. The dismissals cluster into a small set of recurring patterns. `copilot-instructions.md` is our mechanism for pushing back on these globally; this tunes the **common** repo to reduce the false-positive rate in the next measurement cycle.

### Rules added (patterns 1–5)

A cross-cutting principle heads the section: **don''t reason from names** — don''t infer behavior, type relationships, or a field''s meaning/source from an identifier name; verify against the code.

1. **Exception-handler inheritance** — verify the class hierarchy before flagging a "missing" catch (e.g. `UiRequiredException extends ServiceException`, so a `ServiceException` catch already covers it).
2. **Hypothetical configurations / flight rollout state** — don''t block on flight/config combinations you can''t see in the diff; at most a single non-blocking `Sanity check:` note asking the author to confirm the combination is reachable in production.
3. **Telemetry / instrumentation attachment paths** — don''t demand unit tests for the bare act of attaching attributes/values or calling `recordException(...)`. (Boundary: *whether* `recordException` fires on the correct error path / with the correct status is correctness logic and stays test-worthy per §7.2.)
4. **Cross-PR / cross-repo scope creep** — suppress unrelated "while you''re here" suggestions, but *do* surface genuine cross-repo dependencies the change requires or breaks (e.g. a new Common `AttributeName`/`SpanName` that must be mirrored in Broker) as a clearly-labeled, non-blocking follow-up. The test is necessity, not location.
5. **Field / domain semantics** — don''t assert what a field/source means or where it''s populated from without evidence; state it as an explicit assumption if unverifiable.

### Acceptance criteria

- [x] `copilot-instructions.md` in **common** updated with rules for patterns 1–5
- [ ] Rules reviewed by the team in a single async review (this PR)
- [ ] Effect measured in the next Check-In (Run 4)

### Notes

- Docs-only change to `.github/`; no source, no changelog entry required.
- Follow-up tracked in [AB#3678095](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3678095): a compression pass to reduce the file''s overall size / review-pass context footprint (this PR is scoped to the false-positive tuning only).
- Out of scope: disabling Copilot review on any file; changes to the effectiveness-analysis pipeline; the response-rate problem.

Fixes [AB#3660483](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3660483)


.github/copilot-instructions.md
```diff
@@ -17,6 +17,32 @@ This file is self-contained; it includes full baseline & appendices replicated f
...
+   > Real dismissal: *"UiRequiredException extends ServiceException, so retry logic applies."*
+
+2. **Hypothetical configurations / flight rollout state — don't warn on premises you can't see.**
+   Do not raise a concern whose premise is a flight/configuration *combination* that may not exist in production (e.g. "if flight X is enabled together with cache implementation Y, then…"). You have no visibility into rollout state. Only flag a flag/flight branch when the diff itself introduces an unsafe default or a real, reachable code path. Known production state: the in-memory cache flights for accounts and credentials (e.g. `ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE`) are enabled on 100% of devices, so concerns predicated on a non-memory cache implementation are moot.
+   > Real dismissal: *"This is fine as the in_memory_cache flight for accounts and credentials are enabled on 100% of the devices."*
```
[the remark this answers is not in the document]
shahzaibj: Good catch - fixed in 6d84c0a. Dropped the hard `100% of devices` claim (it's external rollout state, not verifiable from the repo and prone to going stale). Reframed Rule 2 as a durable principle - don't gate feedback on a configuration you can't verify from the diff, and the in-repo enum default isn't the deployed value either - while keeping the dismissal quote for grounding.

.github/copilot-instructions.md
```diff
@@ -17,6 +17,32 @@ This file is self-contained; it includes full baseline & appendices replicated f
...
+   > Real dismissal: *"this change is external to this pr."*
+
+5. **Field / domain semantics — don't assert meaning you can't verify; never hallucinate sources.**
+   Do not assert what a field, parameter, or data source "means", or which endpoint it comes from, unless the diff or quoted context supports it — reasoning from a name is not evidence. If a claim depends on a domain fact you cannot confirm from the changed code, either omit it or state it explicitly as an assumption ("Assumption: … If incorrect, disregard."). Known fact in this codebase: `clientDataInfo` / `BaseException.getClientDataInfo()` can be populated **both** from the `x-ms-clientdata` header **and** from an `/authorize` redirect's `clientdata` parameter — do not claim it originates from only one source.
+   > Real dismissal: *"not correct, both fields are populated from x-ms-clientdata."*
```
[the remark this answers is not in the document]
shahzaibj: Agreed, and this one is a bit ironic - my `Known fact` re-asserted the exact source semantics the engineer had corrected. Removed it in 6d84c0a. Rule 5 now stays generic (verify a field's actual population logic before claiming its source) and keeps the dismissal quote verbatim.

.github/copilot-instructions.md
```diff
@@ -437,6 +463,8 @@ Flag when new code:
 - Adds feature flag branching without tests for each state.
 - Adds new public API methods without tests for expected behavior.
 
+Do NOT flag (see §0.1 rule 3): code whose only new behavior is attaching attributes/values to telemetry spans, recording exceptions on spans, or populating an instrumentation field. These attachment paths are considered low-risk and a missing test for them is not a defect.
```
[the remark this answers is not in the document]
shahzaibj: Fixed in 6d84c0a. Scoped the 7.2 telemetry-tests bullet to higher-risk instrumentation (span lifecycle, error-status/recordException correctness, conditional emission) and added an explicit deferral to 0.1 rule 3, so it no longer reads as contradicting 7.1.

.github/copilot-instructions.md
```diff
@@ -591,6 +593,27 @@ Flag:
...
 
+## 14. Recurring False-Positive Patterns — Do NOT Flag These
+
+These rules are calibrated from the team's Copilot Code Review Effectiveness analysis: each pattern below is a comment category that engineers explicitly dismissed as *Confirmed Not Helpful* (wrong, irrelevant, or by-design). Treat them as hard suppression rules. When you are tempted to post a comment matching one of these shapes, re-verify against the rule first; if the premise still holds, stay silent. A missed nit costs the team far less than a confirmed false positive.
+
+1. **Exception-handler inheritance — verify the type hierarchy before flagging a "missing" catch.**
+   Before suggesting an added `catch` block, or claiming a thrown type is not handled/retried, confirm the inheritance chain of the type that is already caught. A `catch` (or instanceof check) on a supertype already covers every subtype. Do **not** infer type relationships from class-name similarity — verify against the actual class hierarchy. Known relationship in this codebase: `UiRequiredException extends ServiceException`, so logic guarding on `ServiceException` (including retry logic) already applies to `UiRequiredException`.
```
somalaya: Could you change this to "For example (verify against current code): `UiRequiredException extends ServiceException`, so…"

.github/copilot-instructions.md
```diff
@@ -591,6 +593,27 @@ Flag:
...
+
+3. **Telemetry / instrumentation attachment paths — do not demand unit tests for them.**
+   Do not request unit tests whose sole purpose is to cover instrumentation: setting span attributes, attaching values to spans, recording exceptions on spans, or populating an instrumentation field (e.g. asserting `getClientDataInfo()` is non-null when a given header/parameter is present). The team classifies these attachment paths as low-risk. This **narrows** §7 — keep asking for tests around branching logic, parsing/serialization, error/retry/fallback handling, concurrency, and public API behavior; just not for the bare act of attaching an attribute or field to telemetry.
```
fadidurah: I feel this may be already covered above? Higher word count may emphasize a particular point, but increasing overall verbosity of this instructions file may lead to context bloat, and causing copilot review agent to suffer accuracy loss across the board for other instructions.

Let's consider if we need to restate this one, or if we can be more concise to save on context for the instructios file.

.github/copilot-instructions.md
```diff
@@ -591,6 +593,27 @@ Flag:
...
+
+5. **Field / domain semantics — don't assert meaning you can't verify; never hallucinate sources.**
+   Do not assert what a field, parameter, or data source "means", or which endpoint/source it is populated from, unless the diff or quoted context supports it — reasoning from a name is not evidence. If a claim depends on a domain fact you cannot confirm from the changed code, either omit it or state it explicitly as an assumption ("Assumption: … If incorrect, disregard."). For example, before describing where a value such as `clientDataInfo` / `BaseException.getClientDataInfo()` originates, verify it against the field's actual population logic rather than its name.
```
fadidurah: It may be a good idea to dissuade the agent from using attribute/field names to ascertain use case in general, seems to be a bad habit of ai agents across the board. this would also be helpful for your exception handling point as well, in which we tell agents not to assume try-catch coverage based on exception class-name, rather to look at the inheritance heirarchy.

.github/copilot-instructions.md
```diff
@@ -437,11 +437,13 @@ Flag when new code:
 - Adds feature flag branching without tests for each state.
 - Adds new public API methods without tests for expected behavior.
 
+Do NOT flag (see §14 rule 3): code whose only new behavior is attaching attributes/values to telemetry spans, recording exceptions on spans, or populating an instrumentation field. These attachment paths are considered low-risk and a missing test for them is not a defect.
```
fadidurah: I think all-in-all this is a fine suggestion. although i do have some concerns when we introduce automated code coverage. Scenario i'm thinking of is user adds a new telemetry attribute to an existing code path, agent ignores it due to this instruction, code covergae flags it as a drop in overlal coverage (once it's more fully implemented). I think this is actually fine in that scenario, we still catch the drop, but the user is left to figure it out on their own rather than ai being able to flag it and come up with a fix. I think a better long term solution is using ai to somehow setup attribute population unit testing in our code, so that we can validate these lines rather than ignore them. Excepting them from the code coverage check will probably be more difficult.

.github/copilot-instructions.md
```diff
@@ -591,6 +593,27 @@ Flag:
...
+
+2. **Hypothetical configurations / flight rollout state — don't warn on premises you can't see.**
+   Do not raise a concern whose premise is a flight/configuration *combination* that may not exist in production (e.g. "if flight X is enabled together with cache implementation Y, then…"). You have no visibility into rollout state, and the in-repo enum/flag default is not the deployed value — do not assume either extreme. Only flag a flag/flight branch when the diff itself introduces an unsafe default or a real, reachable code path. For example, the in-memory cache flights for accounts and credentials (e.g. `ENABLE_FILTER_THEN_CLONE_IN_MEMORY_CACHE`) may already be fully rolled out, making a concern about a non-memory cache implementation moot; rollout state changes over time, so the durable rule is simply: don't gate review feedback on a configuration you cannot verify from the diff.
```
fadidurah: How often is this happening? I think there is some value in the agent at least flagging this with a "Sanity check:" prefix, basically just asking the user to double check if the flight mismatch possibility is actually there in production. If it's generating a lot of noise, i'm fine with this suppression though

.github/copilot-instructions.md
shahzaibj: Agree this is fine for now - the rule only governs whether the reviewer nags for a test, not coverage gating, so an automated coverage drop would still surface independently. I like the longer-term idea of AI-authored attribute-population tests over excepting these lines from coverage; worth a separate item. No change here.

.github/copilot-instructions.md
shahzaibj: Agreed - trimmed in b8f026123. The substantive 'do not flag' note already lives in 7.1, so Rule 3 is now a single line with the duplicated example dropped to save context.

.github/copilot-instructions.md
shahzaibj: Frequency: in Run 3 it was concentrated - 3 instances, all on one PR (#3100), all dismissed - so noisy rather than widespread. Went with your middle ground in b8f026123: Rule 2 no longer hard-suppresses; the agent may post at most one non-blocking `Sanity check:` note asking the author to confirm the flight combo is reachable in prod, capped at one to avoid the repeat-on-same-PR noise.

.github/copilot-instructions.md
shahzaibj: Good call - hoisted it. Added a cross-cutting 'don't reason from names' principle to the section intro (don't infer behavior/relationships/meaning from a class/field/attribute/flag name; verify against the code) and trimmed the now-redundant restatements in Rules 1 and 5. Rule 1 keeps a short 'don't assume coverage from the class name' nudge. b8f026123.

.github/copilot-instructions.md
shahzaibj: Done in b8f026123 - reworded to `For example (verify against current code): UiRequiredException extends ServiceException, so...` Framing it as verify-first rather than a fixed assertion fits the section's spirit better. Thanks!

.github/copilot-instructions.md
```diff
@@ -437,11 +437,13 @@ Flag when new code:
...
 - Adds new public API methods without tests for expected behavior.
 
+Do NOT flag (see §14 rule 3): code whose only new behavior is attaching attributes/values to telemetry spans, recording exceptions on spans, or populating an instrumentation field. These attachment paths are considered low-risk and a missing test for them is not a defect.
+
 ### 7.2 Test Types & Expectations
 - Unit tests: pure logic & edge cases.
 - Integration tests: IPC strategies, cache updates, multi-layer token acquisition.
 - Concurrency tests: stress loops or use deterministic virtual time.
-- Telemetry tests: assert span creation & attribute presence (mock or capture exporter).
+- Telemetry tests: for higher-risk instrumentation (span lifecycle, error-status / `recordException` correctness, conditional emission) assert span creation & status (mock or capture exporter). Do NOT require tests merely for attribute/field attachment (see §14 rule 3).
```
cacosta33: 🟡 **Suggestion**

"Recording exceptions on spans" lands in **both** buckets, which will reproduce the inconsistent reviewer behavior this PR is trying to remove:

- **§7.1 new line + §14 rule 3** → low-risk, *no* test: *"…recording exceptions on spans, or populating an instrumentation field."*
- **This line (§7.2)** → higher-risk, *test required*: *"error-status / `recordException` correctness…"*

A model that sees a `span.recordException(e)` change can''t tell which rule wins. I think the intended split is *merely calling* `recordException` (attachment → low-risk) vs. *whether it''s invoked on the correct failure path / with the right status* (correctness → test-worthy) — but both use the identical phrase. Could we disambiguate, e.g. rule 3 → "recording an exception **on a span** (the attachment call itself)" and keep §7.2 as "**`recordException` invoked on the correct error path / with correct status**"? Span-lifecycle and conditional-emission stay test-worthy as written.

shahzaibj: Great catch - that collision is exactly the inconsistency this PR is meant to kill, so it had to go. Fixed in 1f8ae6602 with your suggested split:

- **Attachment (exempt, no test)** - 7.1 + 14 rule 3 now say *the bare `recordException(...)` attachment call itself*.
- **Correctness (test-worthy)** - 7.2 now reads *whether `recordException` is invoked on the correct error path / with the correct status*.

I also added an explicit boundary note in rule 3 pointing at 7.2, so a model seeing a `span.recordException(e)` change has an unambiguous tiebreaker: bare call = exempt, correct-path/correct-status = test-worthy. Span lifecycle and conditional emission stay test-worthy as before. Thanks!

.github/copilot-instructions.md
```diff
@@ -591,6 +593,29 @@ Flag:
 
 --------------------------------------------------------------------------------
 
+## 14. Recurring False-Positive Patterns — Do NOT Flag These
+
+These rules are calibrated from the team's Copilot Code Review Effectiveness analysis: each pattern below is a comment category that engineers explicitly dismissed as *Confirmed Not Helpful* (wrong, irrelevant, or by-design). Treat them as strong suppression rules — default to staying silent. When you are tempted to post a comment matching one of these shapes, re-verify against the rule first; if the premise still holds, stay silent (a rule may carve a narrow, clearly-labeled exception). A missed nit costs the team far less than a confirmed false positive.
```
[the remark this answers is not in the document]
shahzaibj: Good catch on the mismatch. §14 is the intended placement - the section was deliberately moved to the end of the rules (before the appendices) during review so it doesn't sit in front of the domain primer. Fixed the other side: updated the PR description to refer to §14 and refreshed the stale rule summaries (Rules 2/4/5 had drifted from the current text). The §7.1/§7.2 cross-references already point at §14.

fadidurah (approved):

cacosta33 (approved):

PR #3173, opened 2026-06-29
SwitchBrowser: move cloud discovery off main thread; consolidate handlers, Fixes AB#3679357

mohitc1:
## Issue

`SwitchBrowserUriHelper.buildResumeUri` / `buildProcessUri` trigger AAD cloud discovery — a synchronous HTTPS call on a cold cache. When invoked from a UI thread (`AzureActiveDirectoryWebViewClient.shouldOverrideUrlLoading` or `WebViewAuthorizationFragment.onResume`) the resume/redirect crashes with `NetworkOnMainThreadException`.

## Solution

- **Added async entry points on `SwitchBrowserProtocolCoordinator`:** `processSwitchBrowserRedirectAsync` and `processSwitchBrowserResumeAsync` hop to `Dispatchers.IO` for cloud discovery and resume on `Main` for UI work.
- **Consolidated the SwitchBrowser handler classes** (3 → 1): `SwitchBrowserChallenge` and `SwitchBrowserRequestHandler` merged into `SwitchBrowserProtocolCoordinator`.
- **Sync `processSwitchBrowserResume`** now returns `Pair<Uri, HashMap<String, String>>` and is `@VisibleForTesting internal` (was public, took a callback).
- **WebViewClient** disables the WebView before the async redirect to swallow further URL loads; **Fragment** shows the spinner around the async resume and hides it on error.
- **`isSwitchBrowserChallengeActive`** marked `@Volatile` (cross-thread reads).

## Validation

| Suite | Count | Result |
|---|---|---|
| `SwitchBrowserProtocolCoordinatorTest` | 23 | ✅ |
| `AzureActiveDirectoryWebViewClientTest` | 57 | ✅ |
| `WebViewAuthorizationFragmentTest` | 3 | ✅ |
| broker `AuthTabLaunchStrategy` (downstream) | 1 | ✅ |

## ADO

[AB#3679357](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3679357) — child of Feature [AB#3590610](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3590610) (`[SWAG - M3] DUNA Impl for Sign-in-with-Google`).


common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
```diff
@@ -25,38 +25,91 @@ package com.microsoft.identity.common.internal.ui.webview.switchbrowser
...
+
+    /** Span for the outbound challenge phase. */
+    private val processSpan: Span by lazy {
+        OTelUtility.createSpanFromParent(SpanName.SwitchBrowserProcess.name, spanContext)
+    }
+
+    /** Span for the inbound resume phase. Lazy so test code that never resumes does not allocate it. */
+    private val resumeSpan: Span by lazy {
+        OTelUtility.createSpanFromParent(SpanName.SwitchBrowserResume.name, spanContext)
+    }
```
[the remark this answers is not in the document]
mohitc1: This is the behavior of existing code, nothing new introduced. But I think it's good to address it. Updated.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
```diff
@@ -81,6 +134,166 @@ class SwitchBrowserProtocolCoordinator(
...
+            SwitchBrowserUriHelper.statesMatch(authorizationUrl, state)
+
+            // Select a browser to handle the switch browser challenge
+            val browser = browserSelector.selectBrowser(
+                BrowserDescriptor.getBrowserSafeListForSwitchBrowser(),
+                null
+            ) ?: throw ClientException(
+                ClientException.NO_BROWSERS_AVAILABLE,
+                "No browser found for SwitchBrowserChallenge."
+            )
```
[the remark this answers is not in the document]
mohitc1: Fixed.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
```diff
@@ -25,38 +25,81 @@ package com.microsoft.identity.common.internal.ui.webview.switchbrowser
...
+
+    /** Convenience constructor that uses [AndroidBrowserSelector] by default. */
+    constructor(activity: Activity, spanContext: SpanContext?) : this(
+        activity,
+        AndroidBrowserSelector(activity.applicationContext),
+        spanContext
+    )
+
+    @Volatile
+    var isSwitchBrowserChallengeActive: Boolean = false
```
shahzaibj, line 91: `isSwitchBrowserChallengeActive` is a public var but `wasSwitchBrowserFlowInitiated` right below is `private set`. The class only flips this one internally (launchSwitchBrowserActivity + resetChallengeState), so I think the public setter is only there so tests can force it true. Can we make it private set too? Otherwise anyone holding the coordinator can flip the resume state from outside.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -651,29 +650,33 @@ private void processRedirectUrl(@NonNull final WebView view, @NonNull final Stri
...
-            final String errorCode;
-            if (throwable instanceof IErrorInformation) {
-                errorCode = ((IErrorInformation) throwable).getErrorCode();
-            } else {
-                errorCode = UNKNOWN_ERROR;
-            }
-            returnError(errorCode, throwable.getMessage());
-        }
+        // The handler runs cloud discovery (sync HTTPS on cold cache) off the main thread
+        view.setEnabled(false);
```
shahzaibj: On the resume side the fragment puts up mProgressBar around the async call, but here on the redirect side we just disable the WebView. buildProcessUri hits cloud discovery on a cold cache, so for that window the user is looking at a frozen page with no spinner. Should the redirect leg show progress too?

p3dr0rv:
did you run a E2E validation with happy path and user cancellation or browser dismiss?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -253,28 +253,49 @@ public void onResume() {
...
+                    // Keep the spinner up: launchWebView re-shows it and onPageLoaded hides it
+                    // once the resumed page finishes loading. Hiding here would only flicker.
+                    launchWebView(switchBrowserResumeUri.toString(), switchBrowserResumeHeaders);
+                    return null;
+                },
+                throwable -> {
+                    Logger.error(methodTag, "Error processing switch browser resume", throwable);
+                    mWebView.setEnabled(true);
+                    mProgressBar.setVisibility(View.INVISIBLE);
+                    final ClientException ce = (throwable instanceof ClientException)
```
p3dr0rv: nit: we do not need this, RawAuthorizationResult.fromException() should handle this

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -253,28 +253,49 @@ public void onResume() {
...
         }
+        Logger.info(methodTag, "Resuming switch browser flow");
+        // buildResumeUri may hit the network (cloud discovery on cold cache), so use the
+        // async handler to avoid NetworkOnMainThreadException.
+        mProgressBar.setVisibility(View.VISIBLE);
+        getSwitchBrowserCoordinator().processSwitchBrowserResumeAsync(
+                mAuthorizationRequestUrl,
+                switchBrowserBundle,
+                (switchBrowserResumeUri, switchBrowserResumeHeaders) -> {
+                    mWebView.setEnabled(true);
```
p3dr0rv: onSuccess, do we need to make  mProgressBar.setVisibility(View.INVISIBLE) again? (similar to OnError case)

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
```diff
@@ -136,33 +339,91 @@ class SwitchBrowserProtocolCoordinator(
...
+        resetChallengeState()
+        asyncScope.launch {
+            try {
+                // buildResumeUri can hit the network (cold cache), so build off the main thread.
+                val (uri, headers) = withContext(Dispatchers.IO) {
+                    processSwitchBrowserResume(authorizationRequest, extras)
+                }
+                onSuccessAction(uri, headers)
+            } catch (ce: CancellationException) {
+                throw ce
```
p3dr0rv, line 403: why here we throw vs onError?

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
mohitc1: It get's changed again launchWebView to VISIBLE. Hence didn't change here. 

Let me see, if code can be made better here. May add comment or set progress bar visibility for consistency.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
mohitc1: CancellationException -> failure of infra.
other case, we are communicating for error to previous layer. 
More of convention. do you prefer onError() always?

p3dr0rv: is fine, just documenting it would help

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
mohitc1: Yes, that was the limitation. Refactored code more, so all UI updartes are done in the Fragment, and coordinator provides callbacks.

mohitc1:
> did you run a E2E validation with happy path and user cancellation or browser dismiss?

Yes. Happy path, cancel and MFA after, I addressed the comments.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/switchbrowser/SwitchBrowserProtocolCoordinator.kt
```diff
@@ -81,6 +136,163 @@ class SwitchBrowserProtocolCoordinator(
...
+     *
+     * Returns immediately. [SwitchBrowserUriHelper.buildProcessUri] runs on [Dispatchers.IO];
+     * state validation, browser selection, and [Activity.startActivity] run on Main.
+     *
+     * Status is reported to [SwitchBrowserStatusCallback] (started / failed). Success is observed
+     * via the side effect of [Activity.startActivity].
+     *
+     * The handler does **not** serialize concurrent invocations. Callers are expected to
+     * prevent re-entry themselves — e.g. [AzureActiveDirectoryWebViewClient] disables the
+     * WebView before calling this method, which suppresses additional `shouldOverrideUrlLoading`
```
shahzaibj: The doc here says AzureActiveDirectoryWebViewClient disables the WebView before calling this, but it doesn't anymore. processSwitchBrowserRequest just calls straight through, and the disable actually happens in onSwitchBrowserStarted().

So the thing stopping a second shouldOverrideUrlLoading isn't the caller, it's onSwitchBrowserStarted() running synchronously before this method returns. And that only holds because asyncScope is Main.immediate. Switch it to plain Dispatchers.Main and the disable gets posted instead, so the re-entry guard quietly goes away.

Can we call the Main.immediate dependency out in the comment? Feels load bearing and right now the doc points at the wrong thing.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -705,6 +704,11 @@ public void run() {
     public void onDestroy() {
         super.onDestroy();
         final String methodTag = TAG + ":onDestroy";
+        // Cancel any in-flight switch-browser async work so its continuation does not resume
+        // against this destroyed fragment (touching mWebView / mProgressBar / sendResult).
+        if (mSwitchBrowserProtocolCoordinator != null) {
+            mSwitchBrowserProtocolCoordinator.cancel();
```
shahzaibj, line 710: cancel() is only in onDestroy, but the callbacks all gate on isAdded(), which stays true until onDetach. So if an IO continuation lands in the onDestroyView -> onDetach window, isAdded() is still true and onSwitchBrowserFailed would sendResult + finish with an error instead of the flow just getting cancelled.

Standard teardown runs onDestroyView and onDestroy back to back so it's probably fine in practice, but any reason not to cancel in onDestroyView for a view scoped thing like this? Just making sure isAdded() is actually enough here.

shahzaibj (approved):

p3dr0rv (approved):

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
mohitc1: Moved cancel() to onDestroyView() and null the coordinator field there -> making it coupled to view than the fragment.

common/src/main/java/com/microsoft/identity/common/internal/providers/oauth2/WebViewAuthorizationFragment.java
```diff
@@ -699,6 +698,19 @@ public void run() {
...
 
+    @Override
+    public void onDestroyView() {
+        // The switch-browser coordinator's callbacks touch view state (mWebView / mProgressBar) and
+        // the terminal result, so tear it down with the view. Cancel in-flight async work so its
+        // continuation does not resume against a destroyed view, and null it so a fresh coordinator
+        // is created if the view is recreated (cancel() permanently ends the scope).
+        if (mSwitchBrowserProtocolCoordinator != null) {
+            mSwitchBrowserProtocolCoordinator.cancel();
+            mSwitchBrowserProtocolCoordinator = null;
```
shahzaibj: cancel() + null here ties the coordinator's lifetime to the view, but isSwitchBrowserChallengeActive and wasSwitchBrowserFlowInitiated both live on it and neither is persisted.

So if the view gets torn down and recreated while we're mid flow (fragment instance survives, still waiting on the browser to come back), onResume -> getSwitchBrowserCoordinator() hands back a fresh coordinator -> isExpectingSwitchBrowserResume() is false -> we fall into the else and clearSwitchBrowserBundle(). The static bundle is still sitting there but we just dropped it, so the resume silently dies.

wasSwitchBrowserFlowInitiated resets to false in that case too, so the terminal result loses its IS_SWITCH_BROWSER_FLOW tag. That one was meant to be a never-reset latch, there's even a test asserting it survives resetChallengeState.

configChanges on AuthorizationActivity takes rotation out of it, and a full activity destroy loses this on dev anyway, so it's a narrow window. But did we think about the view-only recreation case? feels like the resume gate and that initiated latch want to outlive the view, not die with it.

mohitc1: > So if the view gets torn down and recreated while we're mid flow.

Is that real scenario here? I'd imagine that would already be an issue for more popular scenarios. 

Anyway, I feel coordinator tied fragment lifecycle was better notion anyway, ties to interactive request. So, I'd rather revert to previous state and keep small edge case of onDestroy() -> onDetach window unaddressed. It does not lead to loss of functionality.

shahzaibj (approved):

PR #3174, opened 2026-07-02
Record root-cause telemetry on KeyStore secret-key wipes to diagnose cache-wipe data loss, Fixes AB#3679391

rpdome:
[AB#3679391](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3679391)

## Summary

`KeyStoreBackedSecretKeyProvider.readSecretKeyFromStorage()` wipes the KeyStore keypair **and** the wrapped-key file on **any** `ClientException` (and on orphaned/empty key-file paths). Because the encrypted material becomes undecryptable once the keypair is gone, this can silently discard recoverable data and force mass re-auth. The `ClientException` error code conflates transient and permanent causes, so today we can't tell **why** a wipe actually happened.

This PR adds **telemetry-only** diagnostics (**no behavior change**) so we can classify these wipes before designing the real fix (retry/caps/classifier).

## What's recorded

On genuine wipes / silent re-keys of **existing** material, on the `SecretKeyRetrieval` span:

| Attribute | Meaning |
|-----------|---------|
| `secret_key_wipe_reason` | which check triggered the wipe (`load_error`, `keystore_key_absent_orphaned_file`, `empty_key_file_rekey`) |
| `secret_key_read_root_cause` | the **true root cause** (type + trimmed message), walked down the cause chain — not the generic `ClientException` wrapper |
| `keystore_numeric_error_code` | exact `android.security.KeyStoreException` numeric code (API 33+) |

## Why root cause (not the wrapper, not a stack trace)

- The `ClientException` wrapper's message is often generic (`"Error while reading..."`); the real reason (e.g. `No space left on device`, a specific `KeyStoreException`) lives further down the chain.
- A full stack trace was considered and rejected: the Aria/1DS span exporter emits span **attributes** but drops the `recordException` stacktrace event, and stack traces print the root cause **last**, so any backend field-size truncation would drop exactly the part we need. The explicit root-cause attribute is captured up-front and is groupable/queryable.

## Safety

- Legitimate **first-time** reads (no keystore key **and** no file, or a simply-absent file — e.g. Office `sharedUserId`) are intentionally **not** recorded.
- `recordWipe` swallows any failure (logs a warning), so telemetry can **never** break the read/wipe path.
- Wipe/delete behavior is unchanged; only `ClientException` triggers a wipe, exactly as before.

## Testing

- `:common:compileLocalDebugJavaWithJavac` — BUILD SUCCESSFUL
- `:common:testLocalDebugUnitTest` (`*WrappedSecretKey*`, `*AndroidKeyStoreUtil*`) — passing

> Note: the companion `AriaSpanDataAdapter` fix (so the exporter records the cause's message, not its class name) is a separate PR in the broker repo.

### DIFF
--- a/common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
+++ b/common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
@@ -36,9 +36,12 @@ import com.microsoft.identity.common.java.opentelemetry.SpanExtension
 import com.microsoft.identity.common.java.opentelemetry.SpanName
 import com.microsoft.identity.common.java.util.FileUtil
 import com.microsoft.identity.common.logging.Logger
+import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.StatusCode
 import java.io.File
 import java.security.KeyPair
+import java.util.Collections
+import java.util.IdentityHashMap
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 import javax.crypto.SecretKey
@@ -74,10 +77,38 @@ class KeyStoreBackedSecretKeyProvider(
         @VisibleForTesting
         const val KEY_FILE_SIZE: Int = 1024
 
+        /** [AttributeName.secret_key_wipe_reason]: wrapped-key file present but its keystore key is gone. */
+        private const val WIPE_REASON_KEYSTORE_KEY_ABSENT_ORPHANED_FILE: String =
+            "keystore_key_absent_orphaned_file"
+
+        /** [AttributeName.secret_key_wipe_reason]: unrecoverable error while reading existing key material. */
+        private const val WIPE_REASON_LOAD_ERROR: String = "load_error"
+
+        /** [AttributeName.secret_key_wipe_reason]: wrapped-key file exists but is empty (silent re-key). */
+        private const val WIPE_REASON_EMPTY_KEY_FILE_REKEY: String = "empty_key_file_rekey"
+
         /**
          * SecretKey cache. Maps wrapped secret key file path to the SecretKey.
          */
         private val sKeyCacheMap: ConcurrentMap<String, SecretKey> = ConcurrentHashMap()
+
+        /**
+         * Walks [throwable]'s cause chain down to its root cause and returns it. Identity tracking
+         * guarantees termination even for self-referential or multi-node cause cycles, so no depth
+         * cap is needed.
+         */
+        @VisibleForTesting
+        fun findRootCause(throwable: Throwable): Throwable {
+            val seen = Collections.newSetFromMap(IdentityHashMap<Throwable, Boolean>())
+            var current = throwable
+            seen.add(current)
+            var next = current.cause
+            while (next != null && seen.add(next)) {
+                current = next
+                next = current.cause
+            }
+            return current
+        }
     }
 
     override val keyTypeIdentifier = KEY_TYPE_IDENTIFIER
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +248,120 @@ class KeyStoreBackedSecretKeyProvider(
...
-            // All tokens with previous SecretKey are not possible to decrypt.
-            Logger.warn(
-                methodTag, "Error when loading key from Storage, " +
-                        "wipe all existing key data "
-            )
-            deleteSecretKeyFromStorage()
-            throw e
+        } catch (e: Exception) {
+            Logger.warn("$TAG:recordWipe", "Failed to record wipe telemetry: ${e.message}")
         }
```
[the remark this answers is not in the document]
rpdome: this should be okay since it's in log file.

common4j/src/main/com/microsoft/identity/common/java/opentelemetry/AttributeName.java
```diff
@@ -676,6 +676,21 @@ public enum AttributeName {
      */
     secret_key_serialization_duration,
 
+    /**
+     * Indicates which check in the secret-key read path triggered a wipe of existing key material
+     * (e.g. an unrecoverable load error, or an orphaned wrapped-key file whose keystore key is gone).
+     * Legitimate first-time reads (no keystore key and no wrapped-key file) are intentionally not recorded.
+     */
+    secret_key_wipe_reason,
```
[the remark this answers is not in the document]
rpdome: addressed in https://github.com/identity-authnz-teams/ad-accounts-for-android/pull/219

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +248,122 @@ class KeyStoreBackedSecretKeyProvider(
...
+    private fun recordWipe(span: Span, reason: String, exception: ClientException? = null) {
+        try {
+            span.setAttribute(AttributeName.secret_key_wipe_reason.name, reason)
+            if (exception != null) {
+                // Record the true root cause, not the generic ClientException wrapper.
+                val rootCause = findRootCause(exception)
+                val rootMessage = rootCause.message?.take(MAX_CAUSE_MESSAGE_LENGTH) ?: "null"
+                span.setAttribute(
+                    AttributeName.secret_key_read_root_cause.name,
+                    "${rootCause.javaClass.simpleName}: $rootMessage"
```
shahzaibj: This one's the exported span attribute, not the log file. `rootCause.message` can be arbitrary - the KDoc example here is literally `IOException: /data/.../key (No space left on device)`, so file paths (and whatever else ends up in a nested cause message) get shipped to the telemetry backend. Truncating to 256 caps the length but not the content. Isn't the class name + numeric error code enough to tell transient from permanent apart, without emitting the raw message?

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +248,122 @@ class KeyStoreBackedSecretKeyProvider(
...
+                )
+                recordWipe(span, WIPE_REASON_LOAD_ERROR, e)
+                span.setStatus(StatusCode.ERROR)
+                span.recordException(e)
                 deleteSecretKeyFromStorage()
-                return null
+                throw e
+            } catch (e: Exception) {
+                // Non-ClientException failures are recorded but not wiped (only ClientException wipes).
+                Logger.error(methodTag, "Unexpected error while reading secret key from storage", e)
```
shahzaibj: `secret_key_read_root_cause`'s doc says it's captured "unconditionally so the real reason is never lost", but recordWipe only runs on the ClientException path. `WrappedSecretKey.deserialize` throws IllegalArgumentException / IOException (unsupported serializer id, corrupt/truncated file), not ClientException - those land here in the generic catch and never call recordWipe, so no root cause attribute for a corrupt-file read failure. Is that intentional? feels like exactly the undecryptable-data case we'd want the root cause for. recordException(e) still fires but that's the top-level exception, not the walked-down root cause.

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +261,108 @@ class KeyStoreBackedSecretKeyProvider(
...
+                    methodTag, "Error when loading key from Storage, " +
+                            "wipe all existing key data "
+                )
+                recordWipe(span, WIPE_REASON_LOAD_ERROR, e)
+                span.setStatus(StatusCode.ERROR)
+                span.recordException(e)
                 deleteSecretKeyFromStorage()
-                return null
+                throw e
+            } catch (e: Exception) {
```
shahzaibj, line 313: `secret_key_read_root_cause`'s doc says it's captured "unconditionally so the real reason is never lost", but recordWipe only runs on the ClientException branch above. `WrappedSecretKey.deserialize` throws IllegalArgumentException (unsupported serializer id) / IOException on a corrupt or truncated wrapped-key file, not ClientException, so those land here in the generic catch and never hit recordWipe. So no secret_key_read_root_cause for a corrupt-file read failure, which feels like exactly the undecryptable-data case you'd want the root cause for. recordException(e) still fires but that's the top-level throwable, not the walked-down root cause. Intentional that the root cause only gets recorded on the wipe path?

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +261,108 @@ class KeyStoreBackedSecretKeyProvider(
...
+     */
+    private fun recordWipe(span: Span, reason: String, exception: ClientException? = null) {
+        try {
+            span.setAttribute(AttributeName.secret_key_wipe_reason.name, reason)
+            if (exception != null) {
+                // Record the true root cause, not the generic ClientException wrapper.
+                val rootCause = findRootCause(exception)
+                span.setAttribute(
+                    AttributeName.secret_key_read_root_cause.name,
+                    "${rootCause.javaClass.simpleName}: ${rootCause.message ?: "null"}"
```
shahzaibj, line 341: This one's the exported span attribute, not the log file Copilot's `${e.message}` comment was about. `rootCause.message` is arbitrary and untrimmed here. the KDoc example right above is literally `IOException: /data/.../key (No space left on device)`, so file paths (and whatever ends up in a nested cause message) get shipped to the exporter, not just the local log. We've already got keystore_numeric_error_code + keystore_error_transience as the low-card groupable signal. isn't class name + numeric code enough to tell transient from permanent apart, without emitting the raw message?

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +261,108 @@ class KeyStoreBackedSecretKeyProvider(
     @Throws(ClientException::class)
     fun readSecretKeyFromStorage(): SecretKey? {
         val methodTag = "$TAG:readSecretKeyFromStorage"
-        try {
-            // Load KeyPair from Android KeyStore
-            val keyPair = AndroidKeyStoreUtil.readKey(alias) ?: run {
-                Logger.info(methodTag, "key does not exist in keystore")
+        val span = OTelUtility.createSpanFromParent(
+            SpanName.SecretKeyRetrieval.name,
```
shahzaibj, line 265: Is this a new span?

rpdome: no. i reused an existing one (it was originally one level lower - in `deserializeAndUnwrapSecretKey`)

I asked @p3dr0rv whether i could bring it up a level as opposed to introducing a new span, and he agreed)

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
rpdome: For now, the scope is limited to the case where the key is wiped.

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
rpdome: keystore_numeric_error_code and keystore_error_transience are API 33+
(I think of them as good-to-have. They would help us categorize errors better, but we would still need the message as the baseline)

We also want to capture any non-keystore error (e.g. I/O error on key file)

common/src/main/java/com/microsoft/identity/common/crypto/KeyStoreBackedSecretKeyProvider.kt
```diff
@@ -230,39 +261,108 @@ class KeyStoreBackedSecretKeyProvider(
...
+                )
+                // Exact platform error code, when a KeyStoreException is present (API 33+).
+                AndroidKeyStoreUtil.getKeyStoreExceptionNumericErrorCode(exception)?.let { numericCode ->
+                    span.setAttribute(AttributeName.keystore_numeric_error_code.name, numericCode)
+                }
+                // Whether the failure is transient/permanent (API 33+) so a retryable error is
+                // distinguishable from genuine data loss.
+                span.setAttribute(
+                    AttributeName.keystore_error_transience.name,
+                    AndroidKeyStoreUtil.getKeyStoreErrorTransience(exception).name
```
shahzaibj, line 351: `keystore_error_transience` here reports NOT_KEYSTORE_ERROR (and numeric code comes back empty) even when the underlying failure IS a keystore error, on the cipher-spec path.

getKeyPairEncryptionPaddings swallows everything: catch(Exception) -> warn -> emptyList(). so if KeyFactory.getKeySpec throws InvalidKeyException(cause=android.security.KeyStoreException) - readKey passed, key goes bad right after - that KeyStoreException dies right there. then:
- unwrapSecretKey sees an empty spec list -> throw ClientException(UNKNOWN_CRYPTO_ERROR, "No compatible cipher specs found...") with no cause
- that bubbles up to this catch as the wipe trigger
- findRootCause gives back the ClientException itself, getKeyStoreErrorTransience -> NOT_KEYSTORE_ERROR, numeric code absent

so a genuine, maybe-transient keystore fault ends up looking like a non-keystore permanent one and we wipe anyway. kinda the exact transient/permanent confusion this whole change is trying to remove, just moved up a method. is preserving the cause on that throw (or not swallowing it in getKeyPairEncryptionPaddings) in scope here, or a followup?

rpdome: good catch. 
I'm intentionally avoiding modifying existing code since it's in the middle of a rollout, but I could perhaps add an attribute in getKeyPairEncryptionPaddings. Thought? @shahzaibj @p3dr0rv 

Edit: synced with @p3dr0rv offline. he suggested we can throw an error instead of swallowing it.

p3dr0rv (approved):

p3dr0rv:
nit: Validate the new telemetry emitted, share a snapshot

common/src/main/java/com/microsoft/identity/common/internal/util/AndroidKeyStoreUtil.java
```diff
@@ -555,7 +614,8 @@ private static AttributesBuilder createAttributesBuilderFromInvalidKeyException(
      * @return List of supported padding names (e.g., "PKCS1", "OAEP"),
      *         or empty list on API < 23 or if retrieval fails
```
shahzaibj, line 615: This `@return` still says empty list on API < 23 or on failure, but we throw in both those cases now. Can we update it and add a `@throws ClientException`?

shahzaibj (approved):

PR #3177, opened 2026-07-06
add intune as an exception for browser validation rule, Fixes AB#3686034

fadidurah:
Add an exception for intune package from the "Multiple-listening apps" check, as this blocks COBO/COPE/AM API BYOD flows from using CCT Browser as their authorization agent, and causing an additional interactive flow as we fallback to webview.

[AB#3686034](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3686034)


common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -264,4 +264,15 @@ public static String getPackageName(final Context context, final int callingAppU
             return null;
         }
     }
+
+    /**
+     * Helper method to validate in Broker that the calling package is Microsoft Intune
+     * to allow System Browser Support.
+     */
```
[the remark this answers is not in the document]
fadidurah: Good catch, method doc is unclear

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -161,17 +162,24 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
             Logger.warn(methodTag,
                     "Application context is null; skipping multiple-app URL scheme validation.");
         } else {
-            final String redirectUri = intent.getStringExtra(
-                    AuthenticationConstants.AuthorizationIntentKey.REDIRECT_URI);
-            if (redirectUri == null) {
+            // Intune is a special case, this is a locked-down scenario for COBO / COPE. This check fails because authenticator is also installed and is also listening in for the broker redirect, so this conflict will occur.
+            // We skip this check to not force COBO / COPE customers to authenticate an additional time. Check the usage of isCallingPackageIntune() for more details on Intune's special handling in broker code.
+            if (PackageUtils.isCallingPackageIntune(appContext.getPackageName())) {
```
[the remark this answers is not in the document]
fadidurah: This was added later

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -264,4 +264,15 @@ public static String getPackageName(final Context context, final int callingAppU
...
+    public static boolean isCallingPackageIntune(@NonNull final String packageName) {
+        final String methodName = ":isCallingPackageIntune";
+        final String intunePackageName = "com.microsoft.intune";
+        Logger.info(TAG + methodName, "Calling package name : " + packageName);
+        return intunePackageName.equalsIgnoreCase(packageName);
+    }
```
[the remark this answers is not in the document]
fadidurah: This is existing logging from the broker layer. We would need this to catch a regression in cobo/cope scenarios

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -161,17 +162,24 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
                 Logger.warn(methodTag,
-                        "Redirect URI is null in the intent; skipping multiple-app URL scheme validation.");
+                        "Calling package is Intune; skipping multiple-app URL scheme validation.");
```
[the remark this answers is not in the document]
fadidurah: Adjusted to use info instead

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -161,17 +162,24 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
                     "Application context is null; skipping multiple-app URL scheme validation.");
         } else {
-            final String redirectUri = intent.getStringExtra(
-                    AuthenticationConstants.AuthorizationIntentKey.REDIRECT_URI);
-            if (redirectUri == null) {
-                Logger.warn(methodTag,
-                        "Redirect URI is null in the intent; skipping multiple-app URL scheme validation.");
+            // Intune is a special case, this is a locked-down scenario for COBO / COPE. This check fails because authenticator is also installed and is also listening in for the broker redirect, so this conflict will occur.
+            // We skip this check to not force COBO / COPE customers to authenticate an additional time. Check the usage of isCallingPackageIntune() for more details on Intune's special handling in broker.
+            if (PackageUtils.isCallingPackageIntune(appContext.getPackageName())) {
```
somalaya: Is it safe to skip it directly for Intune? If a genuinely malicious/unknown app were registered on the redirect scheme during an Intune flow, the exemption would let it through undetected 

I was assuming we still check for the redirect uri to be broker uri and that the activity listening to it is a broker activity (BrokerBrowserRedirectActivity)

somalaya: I think in brokered flows appContext.getPackageName() returns broker's own package name right?

Have we manually tested this fix with a test app before passing the build to Julia?

fadidurah: Yup, it returns the active package name, this has been changed

fadidurah: Discussing this one, i also added a check to make sure we're in auth process before we skip

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+                            redirectUri,
+                            LibraryConfiguration.getInstance().isAuthorizationInCurrentTask()
+                    );
+                }
             }
         }
         super.launchIntent(intent);
     }
 
+    private boolean checkIfRedirectUriIsValidAppLinkOrMsauth(@NonNull final String redirectUri) {
```
somalaya: nit : isValidBrokerRedirectUri()

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+        Logger.info(TAG, "Checking if redirect URI is a valid app link or msauth: " + redirectUri);
+
+        return redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.LTW_APP_PACKAGE_NAME) ||
```
somalaya: Why do we need startsWith here? Isn't it the whole redirect uri?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
-                        appContext,
-                        redirectUri,
-                        LibraryConfiguration.getInstance().isAuthorizationInCurrentTask()
-                );
+                // COBO/COPE Conditions to skip this check:
+                // 1. Request is running in the auth process (Broker request)
+                // 2. Request has a redirect uri that is a valid app link or msauth
+                // These are exclusively COBO / COPE scenarios, since only those have a browser authorization agent under broker requests
+                // (Check MsalAndroidBrokerCommandParameterAdapter).
+                if (ProcessUtil.isRunningOnAuthService(appContext) && checkIfRedirectUriIsValidAppLinkOrMsauth(redirectUri)) {
```
somalaya: Let's keep this behind a flight with default value as true?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+                    Logger.info(methodTag,
+                            "Broker flow (most likely COBO / COPE); skipping multiple-app URL scheme validation.");
```
somalaya: COBO/COPE are not the only flows with this case. This log might be confusing. We may also see this in AM API BYOD flows

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
```diff
@@ -154,4 +156,46 @@ class AndroidAuthorizationStrategyValidationTest {
...
+     * Browser flow with a competing app registered but Intune as the calling package →
+     * multiple-app URL scheme validation is skipped, so no [ClientException] is thrown.
+     * This validates the COBO / COPE exemption where Authenticator also listens for the
+     * broker redirect and would otherwise trigger a false conflict.
+     */
+    @Test
+    fun `launchIntent browser flow skips competing app validation for Intune package`() {
+        registerCompetingApp()
+        val intuneContext = object : ContextWrapper(context) {
+            override fun getPackageName(): String = AuthenticationConstants.Broker.INTUNE_APP_PACKAGE_NAME
```
shahzaibj: This mocks `getPackageName()` -> Intune, but launchIntent doesn't skip on the package at all. the skip is `isRunningOnAuthService(appContext) && checkIfRedirectUriIsValidAppLinkOrMsauth(redirectUri)`.

the redirect half is actually fine here, `buildIntent()` uses `msauth://...` which hits the `REDIRECT_PREFIX` branch. but `isRunningOnAuthService` is false under Robolectric (the running process name won't be `com.microsoft.intune:auth`), so the skip never fires and we fall into `validateNoMultipleAppsListening`. registerCompetingApp() registered a competitor, so this should throw. does `launchIntent browser flow skips competing app validation for Intune package` actually pass for you? I'd expect it to be failing.

and the non-Intune test passes for the same reason the Intune one should fail (skip never fires), so neither is really exercising the exemption, the package name mock is a no-op on this path. can we drive it through the real condition, stub `isRunningOnAuthService` true so the test actually covers the skip branch?

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
```diff
@@ -264,4 +265,14 @@ public static String getPackageName(final Context context, final int callingAppU
             return null;
         }
     }
+
+    /**
+     * Returns true if the provided package name is Microsoft Intune.
+     */
+    public static boolean isCallingPackageIntune(@NonNull final String packageName) {
```
shahzaibj: `isCallingPackageIntune` (and the `PackageUtils` import you added to BrowserAuthorizationStrategy) don't look called anywhere. the skip in launchIntent goes through `isRunningOnAuthService` + `checkIfRedirectUriIsValidAppLinkOrMsauth`, not this. was it meant to be wired into the skip condition and then replaced by the redirect-uri check?

also broker already has an `isCallingPackageIntune` in BrokerUtils (MsalAndroidBrokerCommandParameterAdapter uses it). are we duplicating it in common on purpose, or can we drop this one?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
 
+    private boolean checkIfRedirectUriIsValidAppLinkOrMsauth(@NonNull final String redirectUri) {
+        Logger.info(TAG, "Checking if redirect URI is a valid app link or msauth: " + redirectUri);
+
+        return redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.LTW_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.BROKER_HOST_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.INTUNE_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX);
```
shahzaibj: the last branch here is `redirectUri.startsWith("msauth")`, so this returns true for any msauth redirect, not just the 5 app-link packages listed above. with the `isRunningOnAuthService` gate that means we skip `validateNoMultipleAppsListening` for every msauth redirect handled in the auth process, not only Intune/broker apps. the 5 app-link entries above are basically moot for the custom-scheme case because of it.

is that intended? the changelog says exempt the Intune package specifically but this is a lot wider. if we only mean the broker MS apps, should the msauth branch be package-scoped too instead of a blanket prefix?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +160,46 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
     }
 
+    private boolean checkIfRedirectUriIsValidAppLinkOrMsauth(@NonNull final String redirectUri) {
+        Logger.info(TAG, "Checking if redirect URI is a valid app link or msauth: " + redirectUri);
+
+        return redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.LTW_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.BROKER_HOST_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME) ||
+                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.INTUNE_APP_PACKAGE_NAME) ||
```
shahzaibj: Intune app is not a broker

shahzaibj: Rather than going through each one, why can't you just construct it based on active broker's pkg name? We shouldn't ever receive a redirect to a broker with redirect URI of an **in**-active broker.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
shahzaibj: All Android apps integrated with Entra have redirect URI starting with msauth

common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java
fadidurah: This is a relic of previous implementation. I'm actually going to move this back to broker, no need to put in common. It wasn't being duplicated, the broker PR deleted the isCallingPackageIntune in there, in order to use the more-accessible common counterpart of the method. anyway, will be reverting this.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: I totally agree, and to be honest this is part of where my initial confusion comes from. If we are running in auth process, what's the benefit of validating the redirect uri? Won't it always be a valid broker redirect uri, and this pass becomes unnecessary?

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
fadidurah: This test is failing :)

It was added after the initial implementation and i haven't gotten a chance to fix it yet, apologies for missing that.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: makes sense!

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: Sure thing! better safe than sorry

fadidurah: @copilot add a flight gate for this check

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: @copilot update the method name

fadidurah:
@copilot reset your context of these PRs, some changes have gone in

fadidurah:
@copilot there is also the broker PR, do you have visual on it?
[add intune as an exception for browser validation rule, Fixes AB#3686034 by fadidurah_microsoft · Pull Request #222 · identity-authnz-teams/ad-accounts-for-android](https://github.com/identity-authnz-teams/ad-accounts-for-android/pull/222)

fadidurah:
@copilot We should revert the isCallingPackageIntune changes in both PRs, since we are no longer using the method in common, so the changes are no longer needed

fadidurah:
@copilot remove the import in common/src/main/java/com/microsoft/identity/common/internal/util/PackageUtils.java

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: @copilot update this to include AM API BYOD in the comment, not just COBO/COPE

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: Yes, this is correct, @copilot remove the intune package name from this check

fadidurah:
@copilot also update to latest dev

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: So, reason i use startsWith is because some of our tests contain valid uris that include the switch browser modifier after the package name. using startWith satisfied both conditions without having to introduce uri parsing, which can cause issues.

I'm not sure how applicable switch browser is to this, maybe we can ignore it, if its beyond the scope of AMI API and the other intune flows

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: @copilot let's implement this, construct based on the active broker package name. with support for both the app link prefix and msauth for that specific package, and only that one

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -193,15 +193,48 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
-                redirectUri.startsWith(AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + AuthenticationConstants.Broker.INTUNE_APP_PACKAGE_NAME) ||
-                redirectUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX);
+        final String methodTag = TAG + ":isValidBrokerRedirectUri";
+        Logger.info(methodTag, "Checking if redirect URI is valid for active broker: " + redirectUri);
+
+        final Context appContext = getApplicationContext();
+        if (appContext == null) {
+            Logger.warn(methodTag, "Application context is null; cannot determine active broker package.");
+            return false;
+        }
```
fadidurah: This check is redundant, already in launchIntent @copilot

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -193,15 +193,48 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+
+        final String brokerPackageName = appContext.getPackageName();
+        Logger.info(methodTag, "Active broker package name: " + brokerPackageName);
+
+        // Construct expected redirect URI prefixes for the active broker:
+        // 1. App link format: https://login.microsoftonline.com/androidbroker/<packageName>
+        final String expectedAppLinkPrefix = AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + brokerPackageName;
+        
+        // 2. msauth format: msauth://<packageName>
+        final String expectedMsauthPrefix = AuthenticationConstants.Broker.REDIRECT_PREFIX + "://" + brokerPackageName;
```
fadidurah: 3. what if we get msauth://Microsoft.AAD.BrokerPlugin as the redirect? @copilot

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +161,88 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+        }
+
+        // Construct expected redirect URI prefixes for the active broker:
+        // 1. App link format: https://login.microsoftonline.com/androidbroker/<packageName>
+        final String expectedAppLinkPrefix = AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + brokerPackageName;
+        
+        // 2. msauth format: msauth://<packageName>
+        final String expectedMsauthPrefix = AuthenticationConstants.Broker.REDIRECT_PREFIX + "://" + brokerPackageName;
+
+        final boolean matchesAppLink = redirectUri.startsWith(expectedAppLinkPrefix);
```
fadidurah: @copilot just have the statsWith checks in the if statement, that way if the first one passes, we don't have to do the second comparison

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: I've since completely reworked this check, please take a look

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
```diff
@@ -154,4 +199,63 @@ class AndroidAuthorizationStrategyValidationTest {
...
+    fun `launchIntent skips competing app validation when in auth service with msauth redirect`() {
+        registerCompetingApp()
+        // Simulate running in the broker auth service process.
+        stubIsRunningOnAuthService(true)
+
+        val strategy = TestBrowserAuthorizationStrategy(context, activity)
+
+        // Competing app is registered, but because we're in the auth service with a valid
+        // msauth redirect, validation is skipped and launchIntent completes without throwing.
+        strategy.testLaunchIntent(buildIntent())
```
shahzaibj, line 233: All 3 skip tests use the default `msauth://org.robolectric.default/redirect`, which matches the robolectric package name. So they only cover the case where the redirect is the active broker's own, i.e. the skip firing.

The scoping itself isn't covered: in auth service, flight on, competing app registered, but the redirect is for a different package (say `msauth://com.example.otherapp/...`) -> should still throw MULTIPLE_APPS_LISTENING. That's the actual guarantee `isValidBrokerRedirectUri` gives us now. Can we add that one? Otherwise if the prefix check ever loosens nothing here catches the exemption leaking.

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +161,86 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+        if (redirectUri.startsWith(AuthenticationConstants.Broker.NEW_BROKER_REDIRECT_URI)) {
+            Logger.info(methodTag, "Redirect URI matches legacy broker format.");
+            return true;
+        }
+
+        // Construct expected redirect URI prefixes for the active broker:
+        // 1. App link format: https://login.microsoftonline.com/androidbroker/<packageName>
+        final String expectedAppLinkPrefix = AuthenticationConstants.Broker.APP_LINK_PREFIX + "/" + brokerPackageName;
+        
+        // 2. msauth format: msauth://<packageName>
```
shahzaibj: The package name won't always be `brokerPackageName` right? I mean for interrupt flow the redirect URI passed is the caller's redirect URI but the flow goes through the Broker. Can you evaluate how that works with your changes?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,25 +161,86 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+        final String methodTag = TAG + ":isValidBrokerRedirectUri";
+        Logger.info(methodTag, "Checking if redirect URI is valid for active broker: " + redirectUri);
+
+        final Context appContext = getApplicationContext();
+        final String brokerPackageName = appContext.getPackageName();
+        Logger.info(methodTag, "Active broker package name: " + brokerPackageName);
+
+        // Check for legacy broker redirect URI (msauth://Microsoft.AAD.BrokerPlugin)
+        if (redirectUri.startsWith(AuthenticationConstants.Broker.NEW_BROKER_REDIRECT_URI)) {
+            Logger.info(methodTag, "Redirect URI matches legacy broker format.");
```
shahzaibj: Why is it called legacy? The const is called New, right?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
shahzaibj: I think we could consider removing the redirect URI check. I'm not sure why it was added in the first place

fadidurah: replaced with broker validator call

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: redirect uri check is removed

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
fadidurah: redirect uri check is removed

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
fadidurah: redirect uri check is removed

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,20 +162,34 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+                // Setup validator
+                final BrokerValidator validator = new BrokerValidator(appContext);
+
+                // Conditions to skip this check (brokered flows only):
+                // 1. Flight SKIP_MULTIPLE_APP_VALIDATION_IN_AUTH_SERVICE is enabled (default: true)
+                // 2. Request is running in the auth process (Broker request)
+                // 3. the caller is a valid broker package
+                if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_MULTIPLE_APP_VALIDATION_IN_AUTH_SERVICE) &&
+                        ProcessUtil.isRunningOnAuthService(appContext) &&
+                        validator.isValidBrokerPackage(appContext.getPackageName())) {
```
fadidurah: @copilot i've removed the redirect uri check, it was not needed, and replaced it with isValidBrokerPackage. update this PR to remove the changes related to the redirect uri check, such as the imports in this class, the constant we added, comments throughout, and the tests we added

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,20 +162,34 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
+                Logger.info(methodTag,
                         "Redirect URI is null in the intent; skipping multiple-app URL scheme validation.");
             } else {
-                BrowserRedirectValidator.validateNoMultipleAppsListening(
-                        appContext,
-                        redirectUri,
-                        LibraryConfiguration.getInstance().isAuthorizationInCurrentTask()
-                );
+                // Setup validator
+                final BrokerValidator validator = new BrokerValidator(appContext);
```
fadidurah: this should be inside the flight gate as well @copilot

common/src/main/java/com/microsoft/identity/common/adal/internal/AuthenticationConstants.java
```diff
@@ -1267,6 +1267,8 @@ public static String computeMaxHostBrokerProtocol() {
          */
         public static final String REDIRECT_PREFIX = "msauth";
 
+        public static final String APP_LINK_PREFIX = "https://login.microsoftonline.com/androidbroker";
```
fadidurah: @copilot remove this

changelog.txt
```diff
@@ -6,6 +6,7 @@ vNext
 - [PATCH] Fix KeyStore key generation failures on API <= 30 by adding a conservative RSA/PKCS1/SHA-256 spec (ECS kill-switch flighted) (#3167)
 - [PATCH] Add moshi-adapters dependency to Common module so downstream consumers like OneAuth do not have to add it explicitly (#3168)
 - [PATCH] Harden Authority.isKnownAuthority() to use exact, case-insensitive host matching for developer-configured known authorities (#3165)
+- [PATCH] Exempt the Intune package from the multiple-apps-listening browser redirect validation to support COBO/COPE locked-down scenarios (#3177)
```
fadidurah: @copilot update this message to mention AM API BYOD as well

somalaya (approved):

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
```diff
@@ -154,4 +199,45 @@ class AndroidAuthorizationStrategyValidationTest {
...
+    fun `launchIntent skips competing app validation when in auth service`() {
+        registerCompetingApp()
+        // Simulate running in the broker auth service process.
+        stubIsRunningOnAuthService(true)
+
+        val strategy = TestBrowserAuthorizationStrategy(context, activity)
+
+        // Competing app is registered, but because we're in the auth service and the running
+        // process is a valid broker package, validation is skipped and launchIntent completes without throwing.
+        strategy.testLaunchIntent(buildIntent())
```
shahzaibj, line 233: for the skip to fire here we need `isValidBrokerPackage(getPackageName())` to return true, but you only stub `isRunningOnAuthService`. under robolectric the package is `org.robolectric.default` (that's what the REDIRECT_URI_VALUE host is), which isn't a known broker, so the package filter in BrokerValidator returns empty before the signature check even runs and isValidBrokerPackage is false. so the skip branch never fires, we fall into `validateNoMultipleAppsListening`, and registerCompetingApp() already put a competitor on that msauth redirect, so this should throw MULTIPLE_APPS_LISTENING. there's no try/catch so I'd expect this to be failing for you, and either way it's not actually covering the exemption. can we mock BrokerValidator (mockkConstructor or inject one) so isValidBrokerPackage returns true, same way you stub isRunningOnAuthService?

common/src/main/java/com/microsoft/identity/common/internal/ui/browser/BrowserAuthorizationStrategy.java
```diff
@@ -158,20 +162,40 @@ protected void launchIntent(@NonNull final Intent intent) throws ClientException
...
-                BrowserRedirectValidator.validateNoMultipleAppsListening(
-                        appContext,
-                        redirectUri,
-                        LibraryConfiguration.getInstance().isAuthorizationInCurrentTask()
-                );
+                // Conditions to skip this check (brokered flows only):
+                // 1. Flight SKIP_MULTIPLE_APP_VALIDATION_IN_AUTH_SERVICE is enabled (default: true)
+                // 2. Request is running in the auth process (Broker request)
+                // 3. the caller is a valid broker package
+                if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(CommonFlight.SKIP_MULTIPLE_APP_VALIDATION_IN_AUTH_SERVICE) &&
```
shahzaibj, line 178: the `validateNoMultipleAppsListening(...)` call is copy-pasted in both else branches. could we just collapse this into one skip boolean and drop the nesting? something like

```java
final boolean skip = CommonFlightsManager.INSTANCE.getFlightsProvider()
        .isFlightEnabled(CommonFlight.SKIP_MULTIPLE_APP_VALIDATION_IN_AUTH_SERVICE)
        && ProcessUtil.isRunningOnAuthService(appContext)
        && new BrokerValidator(appContext).isValidBrokerPackage(appContext.getPackageName());
if (skip) {
    Logger.info(methodTag, "Running in broker auth process; skipping multiple-app URL scheme validation.");
} else {
    BrowserRedirectValidator.validateNoMultipleAppsListening(appContext, redirectUri,
            LibraryConfiguration.getInstance().isAuthorizationInCurrentTask());
}
```

short-circuit still keeps the validator out of the non-broker path so you don't lose the instantiation guard you added.

fadidurah: Good catch, this occurred after we added the flight, no reason for this many layers in the if-else chain

common/src/test/java/com/microsoft/identity/common/internal/providers/oauth2/AndroidAuthorizationStrategyValidationTest.kt
fadidurah: @copilot implement this

somalaya (approved):

shahzaibj (approved):

PR #3180, opened 2026-07-09
Promote finalizeBlob() onto IOnboardingTelemetryRecorder interface, Fixes AB#3647677

wzhipan:
Add String finalizeBlob() to IOnboardingTelemetryRecorder so callers can finalize the onboarding telemetry blob through the interface instead of downcasting to the concrete OnboardingTelemetryRecorder. Mark the Kotlin impl's finalizeBlob() as override, refresh the interface Javadoc, add a unit test covering interface-typed access, and add a [MINOR] changelog entry.

Prerequisite for the sibling NoOpOnboardingTelemetryRecorder (Null-Object) follow-up.

[AB#3647677](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3647677)


wzhipan:
Updated with latest `dev` (now includes the 24.5.0 release + `OnboardingBlockingErrorParser` #3122); changelog entry placed under `vNext` with this PR number.

This is the **prerequisite** for the paired broker change **identity-authnz-teams/ad-accounts-for-android#226** (AB#3647677), which drops an `instanceof` downcast and calls `finalizeBlob()` through this interface. Once this Common PR is merged & released, the broker submodule is bumped to that release and #226 goes green.

Verified locally: `:common4j:compileJava` + `:common:testLocalDebugUnitTest` → **16/16 recorder tests pass** (incl. new `testFinalizeBlob_AccessibleViaInterfaceType` exercising interface-typed finalization).

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
```diff
@@ -60,4 +62,21 @@ public interface IOnboardingTelemetryRecorder {
...
+     * this lives on the interface, callers finalize through the {@code IOnboardingTelemetryRecorder}
+     * view without downcasting to the concrete recorder, so every implementation — including a
+     * no-op / null-object recorder — participates in finalization explicitly rather than being
+     * silently skipped.
+     *
+     * @return The populated blob JSON string, or an empty string when there is nothing worth
+     *         emitting (e.g. the session correlation id is missing). Never {@code null}.
+     */
+    @NonNull
+    String finalizeBlob();
```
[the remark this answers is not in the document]
wzhipan: Good catch — addressed in c553ac6e5.

You're right that `IOnboardingTelemetryRecorder` is a **published** interface (it shipped in common4j **24.3.0** via #3123), so adding an *abstract* method would be a binary-incompatible change (`AbstractMethodError` for any downstream impl compiled against an earlier version) and would arguably make this a MAJOR rather than MINOR change.

I changed `finalizeBlob()` to a **Java 8 `default` method returning the empty-blob sentinel** (`""`), which:
- keeps the interface binary-compatible (existing implementers keep working; the empty string is the already-documented "nothing worth emitting" return), so the **[MINOR]** classification stays accurate;
- follows the existing common4j `default`-method convention (e.g. `IBrokerTokenCommandParameters.isRequestFromBroker()`, `IPopManagerSupplier.getDefaultDevicePopManager()`);
- doesn't weaken the change's intent — the concrete `OnboardingTelemetryRecorder` still `override`s it, and the broker call-site now finalizes through the interface for every implementation.

Added a unit test (`testFinalizeBlob_DefaultMethod_ReturnsEmptyBlobForMinimalImplementer`) that verifies a minimal implementer inherits the empty-blob default. Updated the changelog entry to note it's a binary-compatible default method.

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
```diff
@@ -60,4 +62,31 @@ public interface IOnboardingTelemetryRecorder {
...
+     * {@link AbstractMethodError} at runtime) and would make this a MAJOR rather than a MINOR
+     * change. The concrete {@code OnboardingTelemetryRecorder} overrides this to emit the real
+     * blob; the default simply yields the documented "nothing worth emitting" result.
+     *
+     * @return The populated blob JSON string, or an empty string when there is nothing worth
+     *         emitting (e.g. the session correlation id is missing). Never {@code null}.
+     */
+    @NonNull
+    default String finalizeBlob() {
+        return "";
```
shahzaibj: Why not null?

shahzaibj (approved):

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
Prvnkmr337: **Suggestion: promote the empty-blob sentinel onto the interface itself.**

Because `finalizeBlob()` is `@NonNull`, the "nothing to emit" value is part of the contract, not an implementation detail — every caller relies on `blob.isEmpty()` (or equivalent) to decide whether to skip MATS emission, which only works if every implementation agrees the vacuum value is exactly `""`. Right now that agreement is implicit: the interface default returns a literal `""`, and `OnboardingTelemetryRecorder.kt` privately re-declares `private const val EMPTY_BLOB = ""` and uses it at two return sites. Any future override (the sibling `NoOpOnboardingTelemetryRecorder`, test doubles, mocks) that wants to signal "nothing to emit" has to re-discover the magic string — and a subtle drift (`" "`, `"{}"`, `"null"`) would silently break every `isEmpty()`-based caller with no compiler help.

Proposal — small, in this PR:

```java
public interface IOnboardingTelemetryRecorder {
    /** Sentinel returned by {@link #finalizeBlob()} when there is nothing worth emitting. */
    @NonNull String EMPTY_BLOB = "";

    @NonNull
    default String finalizeBlob() {
        return EMPTY_BLOB;
    }
    // …
}
```

Then in `OnboardingTelemetryRecorder.kt`: delete the private `EMPTY_BLOB` constant and replace the two `return EMPTY_BLOB` sites with `return IOnboardingTelemetryRecorder.EMPTY_BLOB`. Optionally, tighten the `common4j` test's minimal-implementer case to assert against `IOnboardingTelemetryRecorder.EMPTY_BLOB` rather than a literal `""`, so the constant's identity — not just its current value — is pinned by tests.

Net effect: one canonical source of truth for the vacuum value, bound to the contract that defines it; no duplication across the interface default and the concrete recorder; a clear anchor for future implementations to reference rather than re-invent.

Prvnkmr337:
Overall LGTM, just a minor comment regarding the constant.

common4j/src/main/com/microsoft/identity/common/java/telemetry/IOnboardingTelemetryRecorder.java
wzhipan: Done in f3d2dded4. Promoted the sentinel onto the interface as `@NonNull String EMPTY_BLOB = "";` and had the default `finalizeBlob()` return it. In `OnboardingTelemetryRecorder.kt` I removed the private `EMPTY_BLOB` const and both return sites now use `IOnboardingTelemetryRecorder.EMPTY_BLOB`. Also tightened the common4j test to assert against `IOnboardingTelemetryRecorder.EMPTY_BLOB` (pinning the constant's identity, not just its value), per your optional suggestion. One canonical source of truth for the vacuum value now — thanks for the catch, this is a good anchor for the sibling `NoOpOnboardingTelemetryRecorder`.

wzhipan: Two reasons for `""` over `null` here:

1. **Contract**: `finalizeBlob()` is `@NonNull` (and now returns the shared `EMPTY_BLOB` sentinel). Every caller decides whether to emit to MATS via `StringUtil.isNullOrEmpty(blob)` / `blob.isEmpty()`, and the concrete recorder already returns `""` (not `null`) for the "nothing worth emitting" cases (empty `sessionCorrelationId`, serialization failure). Keeping the interface default `""` makes it consistent with the existing implementation and the "empty means skip" convention already in use.
2. **Kotlin interop / safety**: a `@NonNull` return lets Kotlin implementers use a non-nullable `String` (as `OnboardingTelemetryRecorder` does) with no platform-type ambiguity, and callers never have to null-check the sentinel path separately from the empty-string path.

So `null` would either violate the `@NonNull` contract or force every call site to handle two "no blob" representations instead of one. Happy to revisit if you'd prefer a nullable contract, but the empty-string sentinel keeps it uniform.

Prvnkmr337 (approved):

PR #3181, opened 2026-07-10
Add on-device retry rule for flaky instrumented-tests, Fixes AB#3678341

p3dr0rv:
## What

Adds a decoupled JUnit4 retry mechanism for **instrumented** (androidTest) tests. The Gradle `org.gradle.test-retry` plugin only retries JVM/Robolectric unit tests (`tasks.withType(Test)`); it does nothing for `connectedAndroidTest`.

## Changes
- `FlakyTestRetryRule` (`TestRule`) + `@RetryFlakyTest` annotation in **testutils**. Opt-in, honours `AssumptionViolatedException` (skips aren't retried). Names are intentionally distinct from the UI-automation `RetryTestRule`/`RetryOnFailure` to prevent silent mis-imports.
- Wire `testutils` into `common` androidTest with `transitive = false` (pulls only the rule classes; avoids dragging testutils' transitive graph e.g. a second BouncyCastle line).
- `RetryRuleValidationTest` — sanity check that the rule re-runs and passes on a later attempt.
- `pr-common-instrumented.yml` — PR-validation pipeline running `common:connectedLocalDebugAndroidTest` on an emulator.

[AB#3678341](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3678341)

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1665439&view=results

## Usage
```kotlin
@get:Rule val flakyTestRetryRule = FlakyTestRetryRule()

@Test
@RetryFlakyTest(retryCount = 2) // up to 3 attempts
fun someFlakyTest() { ... }
```n
> Retry is a mitigation, not a fix — it stays opt-in so flaky tests remain visible and get fixed.

### CODE REVIEW

File: azure-pipelines/pull-request-validation/pr-common-instrumented.yml [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+
+          # Confirm KVM availability
+          sudo apt-get install -y cpu-checker
+          sudo /usr/sbin/kvm-ok
+
+          echo "Starting emulator"
+          sudo nohup $ANDROID_HOME/emulator/emulator -avd android_emulator -no-window -no-audio -wipe-data -no-snapshot -no-boot-anim > /dev/null 2>&1 &
+
+          echo "Wait for device to finish booting"
+          sudo $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
```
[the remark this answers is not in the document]
p3dr0rv: implemented

testutils/src/main/java/com/microsoft/identity/internal/testutils/rules/FlakyTestRetryRule.kt
```diff
@@ -0,0 +1,122 @@
...
+class FlakyTestRetryRule : TestRule {
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                // Prefer a method-level annotation; fall back to a class-level annotation.
+                val retryFlakyTest = description.getAnnotation(RetryFlakyTest::class.java)
+                    ?: description.testClass?.getAnnotation(RetryFlakyTest::class.java)
+
+                if (retryFlakyTest == null) {
```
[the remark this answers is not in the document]
p3dr0rv: implemented

common/build.gradle
```diff
@@ -278,6 +278,13 @@ dependencies {
     testAnnotationProcessor "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
 
     // Instrumented dependencies
+    // testutils is pulled in solely for FlakyTestRetryRule + @RetryFlakyTest (retrying flaky
+    // instrumented tests; the org.gradle.test-retry plugin only covers JVM/Robolectric unit tests).
+    // transitive = false keeps only the rule's classes and avoids dragging testutils' whole
+    // transitive graph onto the androidTest classpath.
+    androidTestImplementation(project(path: ':testutils')) {
+        transitive = false
```
shahzaibj, line 290: `transitive = false` doesn't keep only the rule's classes though. it still puts ALL of testutils on the androidTest classpath (the keyvault/labapi/lab helpers, everything), it just drops testutils' transitive deps (gson, nimbus, robolectric, etc). so the whole graph is still there, just now half of it can't be used - any common instrumented test that touches those other testutils classes would fail to compile / NoClassDefFoundError since their deps are gone.

if the intent is just to keep robolectric and the JVM-only stuff off the instrumented classpath then fine, but the comment isn't accurate. do these 2 classes even need to live in testutils, or could they go somewhere lighter that common's androidTest can depend on cleanly?

testutils/src/main/java/com/microsoft/identity/internal/testutils/rules/FlakyTestRetryRule.kt
```diff
@@ -0,0 +1,122 @@
...
+                    base.evaluate()
+                    return
+                }
+
+                val totalAttempts = 1 + retryFlakyTest.retryCount.coerceAtLeast(0)
+                var lastError: Throwable? = null
+
+                for (attempt in 1..totalAttempts) {
+                    try {
+                        base.evaluate()
```
shahzaibj, line 84: IIUC this re-runs base on the same test instance, so @Before/@After re-run each attempt but any field initialized at declaration (not in @Before) carries over between attempts. The validation tests lean on that with their counters.

The KDoc pitches this as equivalent to the gradle test-retry plugin, but that one re-runs with a fresh instance I believe. So a stateful flaky test that mutates a field before it fails won't get a clean slate on retry here.

can we note that in the annotation KDoc so people writing flaky tests don't get surprised?

testutils/src/main/java/com/microsoft/identity/internal/testutils/rules/FlakyTestRetryRule.kt
```diff
@@ -0,0 +1,122 @@
...
+                        if (attempt > 1) {
```
shahzaibj, line 85: when a test passes on retry we just Log.i to logcat and return, so in the test report a test that flaked twice then passed on the 3rd attempt looks identical to one that passed first try. doesn't that defeat the "stays opt-in so flaky tests remain visible" goal? the flakiness ends up buried in logcat and the result is a clean green.

is something downstream parsing these logs to surface the tests that are still flaking?

.github/workflows/validate-pr-ab-id.yml
```diff
@@ -10,152 +10,83 @@ on:
...
-              echo "Job succeeded or was skipped."
-          else
-              echo "Job failed or was cancelled."
-            exit 1
+          # Consider the run successful if either title was updated or the PR was skipped due to missing AB ID
+          if [ "${{ steps.decide.outputs.update_required }}" == "true" ]; then
+            echo "PR title updated to include AB ID."
+            exit 0
+          fi
+          if [ "${{ steps.extract-ab.outputs.abid }}" == "" ]; then
```
shahzaibj: Bit confused on the no-AB# path. `Abort if no AB ID found` (line 37) does `exit 1` when abid is empty, which fails the job. But this branch treats an empty abid as `skipped by design` and exits 0. By the time we get here the abort already failed the run, so this exit 0 can't rescue it. So is a missing AB# meant to fail the PR or pass? The two steps disagree.

common/build.gradle
p3dr0rv: Good catch — the comment was inaccurate. Fixed in f56ec2a7b.

You're right: `transitive = false` only drops testutils' **transitive dependencies** (gson, nimbus, robolectric, etc.); all of testutils' own classes remain on the androidTest classpath. I rewrote the comment to say exactly that, and added a note that if common's androidTest ever starts using other testutils helpers whose deps are excluded here, they'll `NoClassDefFoundError` at runtime and this exclusion needs revisiting.

Re: moving the two classes somewhere lighter — agreed that's cleaner long-term. Keeping them in testutils for now to avoid spinning up a new module; I'll track a follow-up to extract a slim `test-retry` module that common/msal/broker androidTest can depend on cleanly.

testutils/src/main/java/com/microsoft/identity/internal/testutils/rules/FlakyTestRetryRule.kt
p3dr0rv: Done in f56ec2a7b. Added a note to the `@RetryFlakyTest` KDoc:

> Retries reuse the **same** test instance (unlike the Gradle `test-retry` plugin, which builds a fresh instance per attempt). `@Before`/`@After` still run every attempt, but any field initialised at its declaration is **not** reset between attempts and its mutated state carries over — reset mutable state in `@Before`, not at field declaration.

testutils/src/main/java/com/microsoft/identity/internal/testutils/rules/FlakyTestRetryRule.kt
p3dr0rv: Honest answer: not from a plain `TestRule`. A rule only wraps the `Statement` and the JUnit result is binary (pass/fail) — it has no `RunNotifier`, so it can't emit a "flaked then passed" status the way the Gradle `test-retry` plugin does (that plugin retries at the Gradle *task* level and owns the executor). True flaky **status** reporting on instrumented tests would need a custom `Runner`/`RunListener` or Test-Orchestrator-level retry — a bigger change than this rule.

As a pragmatic middle ground (f56ec2a7b) I changed the on-retry pass from `Log.i` to a distinctive **WARN** marker so flakiness isn't buried:

`FLAKY_TEST_PASSED_AFTER_RETRY: <class>#<method> passed on attempt N of M (flaked K time(s))`

Nothing parses it automatically yet — it's greppable in logcat/CI today. I can file a follow-up to have the instrumented pipeline surface these (fail-soft annotation / dashboard). Would that work for you, or do you'd prefer we go straight for a Runner-based approach?

.github/workflows/validate-pr-ab-id.yml
p3dr0rv: You're right, the two steps disagreed. Intent: a missing AB# should **fail** — the escape hatch is the `skip AB ID validation` label, not an empty AB#. The `Abort if no AB ID found` step (`exit 1`) is the source of truth; the `skipped by design → exit 0` branch in *Final status reporting* was dead/misleading because `if: always()` runs it *after* the abort already failed the job, so it could never rescue the run.

Removed that branch in f56ec2a7b and updated the comment so the reporting step never treats an empty AB# as success.

.github/workflows/validate-pr-ab-id.yml
```diff
@@ -10,152 +10,81 @@ on:
...
-    needs: validation
-    outputs:
-      abid: ${{ steps.get.outputs.id }}
-    steps:
-  # Extracts the AB ID from the pull_request body using regex and store the result using outputs 
-      - name: Get AB ID from comment
-        id: get
+      - name: Validate PR is linked to a work item
+        uses: danhellem/github-actions-pr-is-linked-to-work-item@main
+        continue-on-error: true
```
shahzaibj: `continue-on-error: true` here means the danhellem link check can't fail this gate anymore. So the only hard gate left is the Abort step on line 37, which just greps for an `AB#<digits>` in the body. But the link bot posted lcc-416 on this very PR ("contains AB#... but the Bot could not link it to an Azure Boards work item"). So a PR whose AB# doesn't resolve to a real work item now passes? seems like we quietly went from "linked to a work item" to "has AB# text somewhere".

.github/workflows/validate-pr-ab-id.yml
```diff
@@ -10,152 +10,81 @@ on:
...
-  # 3. The title was already updated.
-  status_report:
-    name: AB ID validation Check
-    runs-on: ubuntu-latest
-    needs: [update_title, check_if_title_needs_to_be_updated, validation]
-    if: always()
-    steps:
-      - name: Check status of job1
-        id: check_status
+          curl -s -X PATCH -H "Authorization: token $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/${{ github.repository }}/pulls/${{ env.PR_ID }} -d '{"title":"'"${{ env.NEW_TITLE }}"'"}'
```
fadidurah: ${{ env.NEW_TITLE }}  is expanded by the Actions runner into the script text before bash runs. Since  NEW_TITLE  derives from the attacker-controllable PR title, a title containing backticks,  $(...) , or a single quote breaks out of the quoting / triggers command substitution → code execution on the runner. Same risk exists on the earlier  PR_TITLE / clean  handling but those go through  env: , which is safe; this one interpolates directly.

Fix: reference the env var and build JSON with jq instead of  ${{ }}  interpolation:

env:
  NEW_TITLE: ${{ steps.decide.outputs.new_title }}
run: |
  body=$(jq -n --arg t "$NEW_TITLE" '{title:$t}')
  curl -s -X PATCH -H "Authorization: token $GITHUB_TOKEN" \
    "https://api.github.com/repos/${{ github.repository }}/pulls/${{ env.PR_ID }}" -d "$body"

This is pre-existing, but since the file is being rewritten it's the right time to close it

.github/workflows/validate-pr-ab-id.yml
```diff
@@ -10,152 +10,81 @@ on:
...
-          PR_TITLE: ${{ needs.get_pr_title.outputs.prtitle }}
+          PR_TITLE: ${{ steps.get-title.outputs.title }}
+          AB_ID: ${{ steps.extract-ab.outputs.abid }}
         run: |
-          if echo "$PR_TITLE" | grep -i -q '\bfix\(ed\|es\)\? AB#[0-9]'; then
-            echo "Pattern found in PR title!, no need to update."
+          if [ -z "${AB_ID}" ]; then
+            echo "update_required=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
```
fadidurah: Isn't this already covered by the abort step earlier in the action? Seems like dead code.

fadidurah (approved):
Approved with one comment

.github/workflows/validate-pr-ab-id.yml
fadidurah: +1 here, what scenario are we bypassing by adding this continue-on-error condition?

.github/workflows/validate-pr-ab-id.yml
p3dr0rv: removed

.github/workflows/validate-pr-ab-id.yml
p3dr0rv: removed

.github/workflows/validate-pr-ab-id.yml
p3dr0rv: removed

shahzaibj (approved):

PR #3184, opened 2026-07-15
Add OpenID-VC PendingIntent return-to-caller support, Fixes AB#3615936

somalaya:
## What
Adds OpenID-VC PendingIntent **return-to-caller** support. When an eSTS CA-block interrupt redirects the auth WebView to an `openid-vc://` URL, the request is handed off to Microsoft Authenticator (the VID wallet). This lets Authenticator bring the **original caller's task** back to the foreground once the VID flow completes, so the user isn't stranded on the wallet.

> Note: the `openid-vc://` redirect detection (`isOpenIdVcUrl` / `processOpenIdVcRequest` / `ENABLE_OPEN_ID_VC_REDIRECT`) is pre-existing. This PR adds the return-to-caller increment on top of it.

## Flighted (kill-switch)
Gated behind **`ENABLE_OPEN_ID_VC_RETURN_TO_CALLER`** (default **ON**). If disabled, the pre-existing behavior applies unchanged — the `openid-vc://` handler is launched **without** a return `PendingIntent`.

## How
- **`ReturnToCallerActivity`** (new): transparent, no-history, non-exported trampoline, manifest-merged into consuming apps. The return `PendingIntent` uses **`FLAG_ACTIVITY_NEW_TASK`** so it reuses and foregrounds the caller's **existing** task (via the host app's `taskAffinity`) rather than creating a duplicate, plus **`FLAG_ACTIVITY_NO_ANIMATION`** for a clean, non-jarring return. It reveals the caller's task then `finish()`es immediately.
- **`AzureActiveDirectoryWebViewClient`**: when the flight is on, if **Microsoft Authenticator** is among the `openid-vc://` handlers **and** passes `BrokerValidator` signature-pinning, the launch is **pinned to Authenticator** (`setPackage`) and the return `PendingIntent` is attached. Uses `queryIntentActivities` rather than `resolveActivity`, which returns the system chooser (`package="android"`) when multiple apps claim the scheme — so the return works even with multiple wallets installed, and **no wallet chooser is shown** for a request only Authenticator can fulfill. If Authenticator isn't installed/verified, the pre-existing dispatch runs without a return `PendingIntent`.
- **`AuthUxJavaScriptInterface`**: rejects scheme-only (null-host) URIs instead of throwing.
- `styles.xml` / `AndroidManifest.xml`: transparent no-animation theme + activity declaration.

## Security
- The return `PendingIntent` is attached only for signature-pinned Authenticator (`com.azure.authenticator`) — not Company Portal or other brokers.
- It is `FLAG_IMMUTABLE`; the launch intent is `setPackage`-pinned to Authenticator.
- The trampoline carries no result data and **never** implies VID/auth success; success is still validated through the normal trusted continuation path.

## Tests
- Null-host `openid-vc://` URI is rejected by `AuthUxJavaScriptInterface`.
- Return `PendingIntent` **is** attached for verified Authenticator (flight on, mocked `BrokerValidator`).
- Return `PendingIntent` is **not** attached for an untrusted handler, nor when the flight is off.
- **Multi-handler:** with two wallets claiming the scheme, the launch is pinned to Authenticator (no chooser) and the return `PendingIntent` is attached.
- `ReturnToCallerActivity` finishes immediately on launch.

> The actual task foregrounding is a windowing behavior that can't be unit-tested — it must be validated on real devices (Samsung + Pixel, API 29→34), including the dead-caller-task edge.

## Pairing
Requires the Authenticator-side counterpart that reads `return_pending_intent` and fires it on the VID presentation success screen's **Ok** action. Authenticator side of the PR is yet to be raised.

Fixes [AB#3615936](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3615936)


common/src/main/res/values/styles.xml
```diff
@@ -57,6 +57,19 @@
...
+        translucent parent so the trampoline can extend android.app.Activity without requiring
+        AppCompat inflation, and suppresses all window animation so the return is instantaneous.
+    -->
+    <style name="Theme.Transparent.NoAnimation" parent="@android:style/Theme.Translucent.NoTitleBar">
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:windowContentOverlay">@null</item>
+        <item name="android:windowIsTranslucent">true</item>
+        <item name="android:backgroundDimEnabled">false</item>
+        <item name="android:windowAnimationStyle">@null</item>
+    </style>
```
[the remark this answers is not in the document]
somalaya: Done

common/src/main/java/com/microsoft/identity/common/internal/ui/ReturnToCallerActivity.java
```diff
@@ -0,0 +1,78 @@
...
+ * consuming (MSAL) app does not need to declare it. The consuming app creates an explicit,
+ * immutable {@code PendingIntent} targeting this class and passes it to Authenticator as part
+ * of the VID launch intent.</p>
+ *
+ * <p><b>This activity is purely a return / navigation signal.</b> It does NOT mark VID or auth
+ * as successful just because it was invoked. VID/auth success must be validated through the
+ * normal trusted continuation state / broker / server validation. This activity only reveals
+ * the previous activity in the caller's task and finishes immediately.</p>
+ */
+public final class ReturnToCallerActivity extends Activity {
```
mohitc1: >ReturnToCallerActivity [](http://example.com/codeflow?start=19&length=22)

Have a better name like OpenIdVcReturnActivity. Although, this pattern can be used for any such flow and such case a general name like this is fine, but right now the documentation and constant anyway couple it with VID, so, I believe, it should have more appropriate name.

mohitc1: Also, make it kotlin

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1168,6 +1197,92 @@ private void processOpenIdVcRequest(@NonNull final WebView view, @NonNull final
...
+            // cut with no jarring flash (and, if the caller's task no longer exists, the transparent
+            // no-history trampoline simply finishes without a visible artifact).
+            returnIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+            returnIntent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
+
+            final PendingIntent returnPendingIntent = PendingIntent.getActivity(
+                    context,
+                    0,
+                    returnIntent,
+                    PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE);
```
mohitc1: >FLAG_UPDATE_CURRENT [](http://example.com/codeflow?start=34&length=19)

Why do you need this? FLAG_ONE_SHOT should be right fit here.

somalaya: Done

common/src/main/java/com/microsoft/identity/common/internal/ui/ReturnToCallerActivity.java
somalaya: Done

mohitc1 (approved):
:shipit:

p3dr0rv (approved):

p3dr0rv (approved):

somalaya:
@copilot resolve the merge conflicts in this pull request

common/src/test/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClientTest.java
```diff
@@ -269,6 +274,117 @@ public void testUrlOverrideHandlesOpenIdVcUrl() {
...
+                AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME, started.getPackage());
+        assertTrue("Trusted wallet must receive the return-to-caller PendingIntent",
+                started.hasExtra(OpenIdVcReturnActivity.RETURN_PENDING_INTENT_EXTRA));
+    }
+
+    @Test
+    public void testOpenIdVcUrl_UntrustedHandler_DoesNotAttachReturnPendingIntent() {
+        // Arrange: the return-to-caller flight is on, but a non-Authenticator app claims the scheme.
+        enableOpenIdVcReturnToCallerFlight();
+        registerOpenIdVcHandler("com.example.malicious");
```
shahzaibj, line 306: This negative case uses `com.example.malicious`, so it short-circuits at `isAuthenticatorOpenIdVcHandler` before BrokerValidator is ever consulted. All three attach tests mock `isValidBrokerPackage` to true. So the one branch that matters here never gets exercised: package IS `com.azure.authenticator` but fails signature verification (a re-signed / repackaged Authenticator). That's the whole point of the sig gate. Can we add a test that registers `com.azure.authenticator` as the handler, mocks `isValidBrokerPackage` to false, and asserts we neither pin the package nor attach the PendingIntent?

common/src/main/AndroidManifest.xml
```diff
@@ -51,6 +51,18 @@
             android:taskAffinity=""
             android:exported="false" />
 
+        <!-- Return-to-caller trampoline. Manifest-merged into consuming (MSAL) apps so they do
+             not need to declare it. Authenticator invokes an explicit, immutable PendingIntent
+             targeting this activity after VID completes to reveal the caller's task.
+             Transparent, no-history, not exported; finishes immediately. It is ONLY a navigation
+             signal and never marks VID/auth as successful. -->
+        <activity
+            android:name="com.microsoft.identity.common.internal.ui.OpenIdVcReturnActivity"
```
shahzaibj, line 60: This return leans on the default taskAffinity (host app package) + NEW_TASK to land back in the caller's task. Doesn't that open a StrandHogg-style window? Another installed app declares `taskAffinity="<host pkg>"`, keeps a task alive, and when Authenticator fires the (legit) return PendingIntent the reveal surfaces the attacker's task instead of the real caller. User thinks they're back in the app after VID but it's someone else's screen.

The SwitchBrowser `taskAffinity=""` pattern doesn't help here since that isolates into a fresh task and you want the opposite (reuse the caller's). So I'm not sure there's a clean fix, and the reveal behavior is OEM/API dependent anyway. But since it's a trust-boundary reveal, was this considered? e.g. reactivating the caller by a recorded task id / known root component rather than affinity-string matching.

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
```diff
@@ -1154,7 +1158,32 @@ private void processOpenIdVcRequest(@NonNull final WebView view, @NonNull final
...
+            // pre-existing behavior (launch the openid-vc handler without a return PendingIntent).
+            if (CommonFlightsManager.INSTANCE.getFlightsProvider().isFlightEnabled(ENABLE_OPEN_ID_VC_RETURN_TO_CALLER)) {
+                // The Microsoft VID CA-block flow can only be completed by Microsoft Authenticator,
+                // so target it explicitly when it is an installed openid-vc:// handler. We do NOT
+                // rely on resolveActivity() here: when more than one app claims the scheme it returns
+                // the system chooser (package "android"), which would drop the return PendingIntent
+                // even if the user then picked Authenticator. Pinning to a verified Authenticator
+                // fixes that and avoids showing a wallet chooser for a request only it can fulfill.
+                if (isAuthenticatorOpenIdVcHandler(intent)
+                        && isTrustedVcWalletPackage(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
```
shahzaibj, line 1180: IIUC this whole path runs on the WebView UI thread (shouldOverrideUrlLoading -> handleUrl -> processOpenIdVcRequest). `isTrustedVcWalletPackage` goes through BrokerValidator, which is a getPackageInfo binder hop + an x509 parse + SHA-512 over Authenticator's certs, plus the queryIntentActivities on the line above. Before, this was a single resolveActivity. shouldLaunchCompanyPortal already does a sig check on this same thread so maybe it's fine for consistency. But the sig result is stable for the process lifetime (signatures can't change while the app is installed), so we could cache it and skip the recompute on every redirect. Do we want this off the UI thread?

common/src/main/java/com/microsoft/identity/common/internal/ui/OpenIdVcReturnActivity.kt
```diff
@@ -0,0 +1,78 @@
...
+ * activity in the caller's task and finishes immediately.
+ */
+class OpenIdVcReturnActivity : Activity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        handleReturn()
+    }
+
+    override fun onNewIntent(intent: Intent?) {
```
shahzaibj: Is `onNewIntent` ever reached here? launchMode is standard (nothing declared), the return PendingIntent is NEW_TASK without SINGLE_TOP, and the activity is noHistory, so every send comes in through a fresh onCreate and no instance sticks around to get a re-delivered intent. Kinda seems like dead code unless I'm missing a path.

mohitc1 (approved):

common/src/main/java/com/microsoft/identity/common/internal/ui/webview/AzureActiveDirectoryWebViewClient.java
somalaya: Two reasons I've kept it synchronous on the WebView UI thread for now:

1. **Frequency** — `processOpenIdVcRequest` only runs on the openid-vc redirect itself (the CA-block VID hand-off), which is effectively once per auth flow, not a per-navigation hot path. So it's a single `BrokerValidator` call per flow, not per redirect frame.
2. **Precedent/consistency** — the device-CA branch (`processDeviceCaRequest` -> `shouldLaunchCompanyPortal`) already does a synchronous SHA-512 signature check on this same thread, so a sig check here is in line with existing behavior in `shouldOverrideUrlLoading`.

You're right that the result is stable for the process lifetime (an installed app's signature can't change), so memoizing `isTrustedVcWalletPackage("com.azure.authenticator")` is a clean, safe optimization. Given the once-per-flow frequency the win is marginal today, so I'd prefer to keep this PR focused and file a follow-up for the cache rather than expand scope here -- happy to add it now if you feel strongly.

common/src/main/java/com/microsoft/identity/common/internal/ui/OpenIdVcReturnActivity.kt
somalaya: Removed it

common/src/main/AndroidManifest.xml
somalaya: Addressed in 0a4ddfeb1. OpenIdVcReturnActivity now declares android:taskAffinity="", so its mandatory NEW_TASK launch spins up a fresh isolated task instead of resolving into a forgeable affinity-matched one — combined with the explicit + immutable PendingIntent, the launch target is fully pinned and can't be hijacked.

I also scoped return-to-caller to the brokered flow - the path we validated on-device - and reduced the trampoline to a bare finishAndRemoveTask, which also removes the dead onNewIntent you flagged.

somalaya:
@copilot resolve the merge conflicts in this pull request

shahzaibj (approved):

PR #3185, opened 2026-07-15
Add NoOpOnboardingTelemetryRecorder null-object for IOnboardingTelemetryRecorder, Fixes AB#3647678

wzhipan:
## Summary

Adds `NoOpOnboardingTelemetryRecorder`, a stateless **Null-Object** implementation of `IOnboardingTelemetryRecorder` in `common4j`:

- `addStep(...)` / `addBlockingError(...)` are no-ops
- `finalizeBlob()` returns `IOnboardingTelemetryRecorder.EMPTY_BLOB` — the shared `@NonNull` "nothing worth emitting" sentinel added by #3180, exactly what the concrete `OnboardingTelemetryRecorder` returns when there is no joinable session-correlation id

Exposed as a Kotlin `object` → Java-visible `NoOpOnboardingTelemetryRecorder.INSTANCE` singleton, so broker code can hold `mOnboardingRecorder` as `@NonNull` and drop the scattered `if (recorder != null)` guards (see the companion broker PR).

## Why

Follow-up from PR-A review feedback ([ad-accounts-for-android#178](https://github.com/identity-authnz-teams/ad-accounts-for-android/pull/178)). The recorder is a textbook Null-Object fit — *"when absent, everything stays a no-op."*

## Base

Rebased directly onto `dev` (the prerequisite #3180 — [AB#3647677](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3647677), which promotes `finalizeBlob()` and the shared `EMPTY_BLOB` sentinel onto the interface — has merged). The diff here is only the NoOp change (3 files, +112).

## Cross-repo

Paired with the broker PR in `identity-authnz-teams/ad-accounts-for-android` that converts `mOnboardingRecorder` to `@NonNull` and consumes `NoOpOnboardingTelemetryRecorder.INSTANCE`. The broker's `common` submodule bump is **deferred to release** (matching #3180's pattern).

## Testing

- `NoOpOnboardingTelemetryRecorderTest` (3 tests): `finalizeBlob()` returns `IOnboardingTelemetryRecorder.EMPTY_BLOB`, and `addStep` / `addBlockingError` are no-ops that leave the blob empty.
- `./gradlew :common4j:test` green locally (JDK 17).
- `[MINOR]` changelog entry added under vNext.

## Review feedback addressed

- Return the canonical `IOnboardingTelemetryRecorder.EMPTY_BLOB` sentinel instead of a bare `""` literal (single source of truth, consistent with the concrete recorder and the interface default).
- Dropped the tautological `instance_isStatelessSharedSingleton` test — a Kotlin `object` is a compiler-guaranteed singleton, so `assertSame` could not regress.

[AB#3647678](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3647678)


common4j/src/test/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorderTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+        assertEquals("", recorder.finalizeBlob())
+    }
+
+    @Test
+    fun instance_isStatelessSharedSingleton() {
+        val first: IOnboardingTelemetryRecorder = NoOpOnboardingTelemetryRecorder
+        val second: IOnboardingTelemetryRecorder = NoOpOnboardingTelemetryRecorder
+
+        assertSame("NoOp recorder is stateless and shared through a single INSTANCE", first, second)
```
shahzaibj: isn't this always true regardless of the impl? It's a Kotlin `object` so `first` and `second` are the same INSTANCE by language guarantee, assertSame can't fail here. So it's not really testing statelessness, just that `object` is a singleton which the compiler already gives us. I'd either drop it or assert something that could actually regress.

common4j/src/main/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    override fun addStep(stepId: String) {
+        // No-op: this recorder intentionally discards onboarding steps.
+    }
+
+    override fun addBlockingError(errorCode: String) {
+        // No-op: this recorder intentionally discards blocking errors.
+    }
+
+    override fun finalizeBlob(): String = ""
```
shahzaibj: What is the obsession with empty strings?

shahzaibj (dismissed):

common4j/src/test/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorderTest.kt
wzhipan: Agreed — dropped `instance_isStatelessSharedSingleton`. You're right that a Kotlin `object` is a compiler-guaranteed singleton, so `assertSame` on two references to it can never fail; it tested the language, not our code. The remaining tests (`addStep` / `addBlockingError` stay no-ops and leave the blob empty) cover the behavior that could actually regress. Rebased onto latest `dev` and applied in b40c443a7.

common4j/src/main/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorder.kt
wzhipan: Fair point. Since #3180 merged, the interface now owns the sentinel as `@NonNull String EMPTY_BLOB = ""` and its default `finalizeBlob()` returns it. I changed the NoOp to return `IOnboardingTelemetryRecorder.EMPTY_BLOB` instead of a bare literal, so there's a single source of truth for the "nothing to emit" value (matching the concrete recorder and the interface default), and updated the tests to assert against the constant. Applied in b40c443a7.

common4j/src/main/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    override fun addStep(stepId: String) {
+        // No-op: this recorder intentionally discards onboarding steps.
+    }
+
+    override fun addBlockingError(errorCode: String) {
+        // No-op: this recorder intentionally discards blocking errors.
+    }
+
+    override fun finalizeBlob(): String = IOnboardingTelemetryRecorder.EMPTY_BLOB
```
shahzaibj: Interface default `finalizeBlob()` already returns `EMPTY_BLOB`, so this override is doing the exact same thing. Do we need it? addStep/addBlockingError are abstract so those have to stay, but this one we'd just inherit the identical value.

wzhipan: Good call — dropped the override. It now inherits the interface default, so only the abstract addStep/addBlockingError are overridden. Done in 084106f97.

common4j/src/test/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorderTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private val recorder: IOnboardingTelemetryRecorder = NoOpOnboardingTelemetryRecorder
+
+    @Test
+    fun finalizeBlob_returnsEmptyBlobSentinel() {
+        assertEquals(IOnboardingTelemetryRecorder.EMPTY_BLOB, recorder.finalizeBlob())
+    }
+
+    @Test
+    fun addStep_isNoOp_andLeavesBlobEmpty() {
```
shahzaibj: Same kind of thing as the singleton test we dropped I think. finalizeBlob() isn't overridden anymore so it returns EMPTY_BLOB unconditionally, and the object is stateless, so these two assertions can't fail no matter what addStep/addBlockingError do. Even if one of them mutated something, finalizeBlob wouldn't reflect it. So they're not really proving the no-op behavior, just that the calls don't throw. I'm not sure they buy us much over finalizeBlob_returnsEmptyBlobSentinel.

changelog.txt
```diff
@@ -4,6 +4,7 @@ vNext
 - [MINOR] Promote finalizeBlob() onto the IOnboardingTelemetryRecorder interface (as a binary-compatible default method) so callers finalize the onboarding telemetry blob through the interface instead of downcasting to the concrete OnboardingTelemetryRecorder (#3180)
 - [PATCH] Security: validate the target of WebView broker-install intent-scheme redirects against an allow-list (Google Play Store only), clearing any explicit component/selector and stripping URI-permission grant flags so resolution is driven solely by the validated package; gated by ENABLE_BROKER_INSTALL_INTENT_VALIDATION CommonFlight (default off) and recording the launched / blocked outcome via the is_broker_install_intent_blocked span attribute (#3170)
 - [PATCH] Rename WebAppError 'error' field to 'code' in WebApps API error response (#3192)
+- [MINOR] Add NoOpOnboardingTelemetryRecorder, a null-object implementation of IOnboardingTelemetryRecorder (addStep/addBlockingError no-op, finalizeBlob() returns an empty string), so callers can hold the recorder as @NonNull and drop the scattered null guards (#3185)
```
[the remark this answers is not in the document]
wzhipan: Done in 90537fa30 — updated the entry to say finalizeBlob() inherits the interface default and returns the EMPTY_BLOB sentinel.

common4j/src/main/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorder.kt
```diff
@@ -0,0 +1,54 @@
...
+ * "nothing worth emitting" result (an empty blob, exactly what the concrete recorder returns when
+ * it has no joinable session correlation id).
+ *
+ * This lets callers treat the recorder as always-present (`@NonNull`) and drop the scattered
+ * `if (recorder != null)` guards: when a request carries no onboarding seed JSON (older clients,
+ * standalone broker flows), `INSTANCE` is substituted for `null` and every step / blocking-error
+ * emission and blob finalization stays a silent no-op.
+ *
+ * The recorder is stateless, so a single shared [INSTANCE] is exposed (accessed from Java as
+ * `NoOpOnboardingTelemetryRecorder.INSTANCE`).
```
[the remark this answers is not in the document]
wzhipan: Done in 90537fa30 — rendered it as a code span (`INSTANCE`) instead of a KDoc link.

common4j/src/test/com/microsoft/identity/common/java/telemetry/NoOpOnboardingTelemetryRecorderTest.kt
wzhipan: You're right — those two assertions couldn't fail, and the `_andLeavesBlobEmpty` names promised a guarantee the test wasn't providing. I reworked it rather than dropping it, because the value isn't statelessness, it's that the calls stay safely *callable*:

The broker half of this work converts `mOnboardingRecorder` to `@NonNull` and deletes every `if (recorder != null)` guard, so production call sites now invoke `addStep`/`addBlockingError` directly on `INSTANCE`. That makes "a no-op body never throws" an actual contract — if someone later swaps a body for `TODO()` or `UnsupportedOperationException`, it's a crash in the brokered flow instead of a silent no-op.

So the test is now `recordingCalls_doNotThrow_whenInvokedUnguarded`, the comment states that rationale instead of apologizing for having nothing to assert, and it ends on `assertEquals(EMPTY_BLOB, finalizeBlob())` so it isn't assertion-free. I sanity-checked that it can actually fail: replacing a no-op body with a throw does fail it.

Applied in ab2ec87de, also rebased onto latest `dev`.

Prvnkmr337 (approved):

wzhipan:
@copilot resolve the merge conflicts in this pull request

